PR #13, opened 2020-02-18
Converts most instances of foo << browse() to /datum/browser.

MistakeNot4892:
- [x] Convert `browse()` calls to user `/datum/browser`.
- [ ] Convert `/Topic()` on popup-using atoms to `/OnTopic()`.
- [ ] Convert `/attack_hand()` on above to `/interact()`.
- [ ] Remove extraneous `set_machine()` calls.


MistakeNot4892:
will need updating

PR #21, opened 2020-02-23
Mech punches produce attack logs.

mikomyazaki:


code/modules/mechs/mech_interaction.dm
```diff
@@ -171,6 +171,9 @@
 		return attack_self(user)
 	else if(adj)
 		setClickCooldown(arms ? arms.action_delay : 15)
+		if(ismob(A))
+			var/mob/target = A
```
afterthought2: Don't think you need this cast.

MistakeNot4892: Should be handled by the arg types in the proc call.

MistakeNot4892 (dismissed):

code/modules/mechs/mech_interaction.dm
```diff
@@ -171,6 +171,7 @@
 		return attack_self(user)
 	else if(adj)
 		setClickCooldown(arms ? arms.action_delay : 15)
+		admin_attack_log(user, target, "Attacked using \a [arms]", "Was attacked with \a [arms]", "used \a [arms] to attack")
```
MistakeNot4892: target => A

mikomyazaki: Oops.

MistakeNot4892 (approved):

Issue #20, opened 2020-02-22
Possibly broken antag and away maps


quardbreak:
Hello.. I'm not sure what you like when someone disturb you, but I would like to help with mention. Some files (maps) have a leftover references to already non-exist content (solgov uniform, vendors, etc...).

afterthought2:
It's possible they are currently not unit-tested properly.

afterthought2:
Fixed in #22.

PR #22, opened 2020-02-23
Map templates are now globally unit tested.

MistakeNot4892:
Should identify specific details of #20. Will also tack on a commit to fix any identified map failures to this PR.

### DIFF
--- a/code/_helpers/logging.dm
+++ b/code/_helpers/logging.dm
@@ -208,14 +208,11 @@
 /atom/movable/get_log_info_line()
 	var/turf/t = get_turf(src)
 	if(t)
-		var/obj/effect/overmap/visitable/O = map_sectors["[z]"]
+		var/obj/effect/overmap/visitable/O = map_sectors["[t.z]"]
 		if(istype(O))
 			return "[..()] ([t]) ([t.x],[t.y],[t.z] - [O.name]) ([t.type])"
-		else
-			return "[..()] ([t]) ([t.x],[t.y],[t.z]) ([t.type])"
-
-	else
-		return "[..()] (NULL) ([t ? "[t.x],[t.y],[t.z]" : "0,0,0"]) (NULL)"
+		return "[..()] ([t]) ([t.x],[t.y],[t.z]) ([t.type])"
+	return "[..()] ([loc?.name || "NULL" ]) (NULL) ([loc?.type || "NULL"])"
 
 /mob/get_log_info_line()
 	return ckey ? "[..()] ([ckey])" : ..()
--- a/code/_helpers/logging.dm
+++ b/code/_helpers/logging.dm
@@ -199,11 +199,23 @@
 	return "[..()] ([isnum(z) ? "[x],[y],[z]" : "0,0,0"])"
 
 /turf/get_log_info_line()
-	return "[..()] ([x],[y],[z]) ([loc ? loc.type : "NULL"])"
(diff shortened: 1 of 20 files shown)


afterthought2:
Suggest placing them during init, before the testing starts, so that map tests will run on them consistently. Not sure why it hangs.

PsiOmegaDelta:
It's probably preferable to load any map templates in `/datum/controller/subsystem/unit_tests/proc/load_map_templates()` for this reason.

code/_helpers/logging.dm
```diff
@@ -199,11 +199,23 @@
...
-	return "[..()] ([t ? t : "NULL"]) ([t ? "[t.x],[t.y],[t.z]" : "0,0,0"]) ([t ? t.type : "NULL"])"
+	if(t)
+		var/obj/effect/overmap/visitable/O = map_sectors["[z]"]
+		if(istype(O))
+			return "[..()] ([t]) ([t.x],[t.y],[t.z] - [O.name]) ([t.type])"
+		else
+			return "[..()] ([t]) ([t.x],[t.y],[t.z]) ([t.type])"
+
+	else
+		return "[..()] (NULL) ([t ? "[t.x],[t.y],[t.z]" : "0,0,0"]) (NULL)"
```
PsiOmegaDelta: `	return "[..()] (NULL) ("0,0,0") (NULL)"`
The `else` can be dropped as well, the if-statement above will never fall through as is.

code/controllers/subsystems/mapping.dm
```diff
@@ -32,6 +32,17 @@ SUBSYSTEM_DEF(mapping)
...
+/datum/controller/subsystem/mapping/proc/includeTemplate(var/datum/map_template/map_template, var/list/banned_maps)
+	if(!initial(map_template.id))
+		return FALSE
+	var/datum/map_template/MT = new map_template()
+	if(banned_maps)
+		for(var/mappath in MT.mappaths)
+			if(banned_maps.Find(mappath))
+				return FALSE
+	map_templates[MT.name] = MT
+	. = MT
```
PsiOmegaDelta: This proc is mixing returning reference and value types (a template or `FALSE`)
Should either return instance/`null` or `TRUE`/`FALSE`

PsiOmegaDelta: Line 43-44 should probable be replaced with `return TRUE`

https://github.com/NebulaSS13/Nebula/pull/22/files#diff-c2e5790416598e06d4b9215742ea8c70R58-R59 should be replaced with
```
if(includeTemplate(item, banned_maps)
	map_templates[MT.name] = MT
	. = MT
else
	continue
```
That way https://github.com/NebulaSS13/Nebula/pull/22/files#diff-de1774595a6a972d2ecda2c7215a5aacR83-R86 can simply be changed to
```
/datum/controller/subsystem/mapping/includeTemplate(var/datum/map_template/map_template, var/banned_maps)
	return TRUE
```

PsiOmegaDelta, line 44: This proc is mixing returning reference and value types (map template and `FALSE`)
Should either return the map template and null or `TRUE`/`FALSE`

MistakeNot4892: Removed `FALSE` return values on earlier returns.

afterthought2 (approved):

PR #27, opened 2020-02-24
Adds preferences for sound player

quardbreak:
Ports feature which will help manually mute jukebox music for self.

### DIFF
--- a/code/datums/sound_player.dm
+++ b/code/datums/sound_player.dm
@@ -23,11 +23,11 @@ GLOBAL_DATUM_INIT(sound_player, /decl/sound_player, new)
 
 
 //This can be called if either we're doing whole sound setup ourselves or it will be as part of from-file sound setup
-/decl/sound_player/proc/PlaySoundDatum(var/atom/source, var/sound_id, var/sound/sound, var/range, var/prefer_mute)
+/decl/sound_player/proc/PlaySoundDatum(var/atom/source, var/sound_id, var/sound/sound, var/range, var/prefer_mute, var/datum/client_preference/preference)
 	var/token_type = isnum(sound.environment) ? /datum/sound_token : /datum/sound_token/static_environment
-	return new token_type(source, sound_id, sound, range, prefer_mute)
+	return new token_type(source, sound_id, sound, range, prefer_mute, preference)
 
-/decl/sound_player/proc/PlayLoopingSound(var/atom/source, var/sound_id, var/sound, var/volume, var/range, var/falloff = 1, var/echo, var/frequency, var/prefer_mute)
+/decl/sound_player/proc/PlayLoopingSound(var/atom/source, var/sound_id, var/sound, var/volume, var/range, var/falloff = 1, var/echo, var/frequency, var/prefer_mute, var/datum/client_preference/preference)
 	var/sound/S = istype(sound, /sound) ? sound : new(sound)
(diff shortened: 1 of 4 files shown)


code/modules/client/preference_setup/global/preferences.dm
```diff
@@ -108,6 +108,14 @@ var/list/_client_preferences_by_type
 		sound_to(preference_mob, sound(null, repeat = 0, wait = 0, volume = 0, channel = GLOB.lobby_sound_channel))
 		sound_to(preference_mob, sound(null, repeat = 0, wait = 0, volume = 0, channel = GLOB.ambience_sound_channel))
 
+/datum/client_preference/play_jukeboxes
+	description = "Play jukeboxes"
+	key = "SOUND_JUKEBOXES"
+
+/datum/client_preference/play_boomboxes
+	description = "Play boomboxes"
+	key = "SOUND_BOOMBOXES"
```
PsiOmegaDelta: Seems excessive to have two preferences.
If someone another kind of player should that have a custom preference as well?

quardbreak: We have boombox player which can used to walkie-play music and I thought what this is a good reason to add this preference.

You may mute jukebox which probably can be only in bar, or you can mute boombox which probably can be anywhere.

PsiOmegaDelta: Just as an example: If someone adds a vehicle with a sound system, which may or may not be mobile, which of these two preferences should be used? Should a third be created just for that?

MistakeNot4892: Having one preference for music players is probably the way to go IMO, boomboxes and jukeboxes are pretty much the same conceptually.

code/datums/sound_player.dm
```diff
@@ -190,6 +193,11 @@ datum/sound_token/proc/Mute()
 	PrivUpdateListeners()
 
 datum/sound_token/proc/PrivAddListener(var/atom/listener)
+	if(preference) // Checking if client want to hear it
```
MistakeNot4892:
````
	if(preference)
		var/mob/M = listener
		if(istype(M))
			if(M.get_preference_value(preference) != GLOB.PREF_YES)
````
should really be in a proc rather than copypasted here and in `PrivUpdateListener()`.

File: code/datums/sound_player.dm [hunk 3]
```diff
@@ -239,6 +245,10 @@ datum/sound_token/proc/PrivAddListener(var/atom/listener)
 		PrivUpdateListener(listener)
 
 /datum/sound_token/proc/PrivUpdateListener(var/listener, var/update_sound = TRUE)
+	if(preference && !check_preference(listener))
```
MistakeNot4892: You check `preference` in the first line of `check_preference()` already.

quardbreak:
Done, I did some research when messed with same idea on one repo. Should work fine now.

CrimsonShrike:
Bit late, but there was already a setup of sorts using a type override here https://github.com/NebulaSS13/Nebula/blob/dev/code/modules/synthesized_instruments/sound_token.dm#L49

Used by synthesized instruments alone. Either way may want to either convert it too or use same system I suppose.

quardbreak:
Psi said what he need (?) to rewrite sound player, so I prefer to keep it as it right now.

quardbreak:
Gonna test this thing again. I need change code.

MistakeNot4892 (approved):

PR #28, opened 2020-02-24
Fixes airtight plasticflaps

SodiumShine:
Makes them airtight on initialization as they should be.

### DIFF
--- a/code/game/objects/structures/plasticflaps.dm
+++ b/code/game/objects/structures/plasticflaps.dm
@@ -16,7 +16,7 @@
 		/mob/living/simple_animal/mouse,
 		/mob/living/silicon/robot/drone
 		)
-	var/airtight = 0
+	var/airtight = FALSE
 
 /obj/structure/plasticflaps/CanPass(atom/A, turf/T)
 	if(istype(A) && A.checkpass(PASS_FLAG_GLASS))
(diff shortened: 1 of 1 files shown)


code/game/objects/structures/plasticflaps.dm
```diff
@@ -81,3 +81,7 @@
 
 /obj/structure/plasticflaps/airtight // airtight defaults to on 
 	airtight = 1
+
+/obj/structure/plasticflaps/airtight/Initialize()
+	. = ..()
+	become_airtight()
```
PsiOmegaDelta:
```
if(airtight)
	become_airtight()
```

PsiOmegaDelta (changes requested):

code/game/objects/structures/plasticflaps.dm
SodiumShine: You want that for regular plastic flaps? Only, the airtight child should always be airtight no?

PsiOmegaDelta: Ah, this should indeed instead be
```
/obj/structure/plasticflaps/Initialize()
	. = ..()
	if(airtight)
		become_airtight() 
```

SodiumShine: Actually on the subject, should the destroy() be checking for airtightness? I don't know what other things cause a turf to block air.

MistakeNot4892 (approved):

comma:
why do flaps even need to touch the turf? Can't they set their own `atmos_canpass` to `CANPASS_NEVER`?

SodiumShine:
This should be good to go now.

MistakeNot4892 (approved):

PR #34, opened 2020-02-25
Adds shortcut chat toggles

quardbreak:
This PR adds the chat toggles for input window. On click, you can select where you will send message.
Currently there is four toggles and one redesigned  Hot key toggle:
image

Looks cute, what you think about it?

### DIFF
--- a/interface/skin.dmf
+++ b/interface/skin.dmf
@@ -947,60 +947,108 @@ window "mainwindow"
 		size = 640x440
 		anchor1 = none
 		anchor2 = none
+		background-color = none
 		is-default = true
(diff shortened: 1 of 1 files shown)


MistakeNot4892:
What is the purpose of having OOC and LOOC checks when you use different verbs to speak on them already? 

I am not sure that I am a fan of this change. It makes the chat bar look a bit crowded and quite garish.

quardbreak:
Some people prefer to use input window as I remember. This is only argument which I can say right now.
Actually yes, this looks strange, but this is only that came on my 'le desigh' mind.

PR #35, opened 2020-02-25
Adds Discord button

quardbreak:
Adds the Discord server button for general use. Can be helpful for downstream servers which use only discord server as replacement for forum. Here probably should be a special button for it, but I'll add it later.

### DIFF
--- a/interface/interface.dm
+++ b/interface/interface.dm
@@ -3,48 +3,60 @@
 	set name = "Wiki"
 	set desc = "Visit the wiki."
 	set hidden = 1
-	if( config.wikiurl )
+	if(config.wikiurl)
 		if(alert("This will open the wiki in your browser. Are you sure?",,"Yes","No")=="No")
 			return
 		send_link(src, config.wikiurl)
 	else
-		to_chat(src, "<span class='warning'>The wiki URL is not set in the server configuration.</span>")
+		to_chat(src, SPAN_WARNING("The wiki URL is not set in the server configuration."))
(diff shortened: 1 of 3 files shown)


interface/interface.dm
```diff
@@ -3,50 +3,63 @@
...
+/client/verb/discord()
+	set name = "Discord"
+	set desc = "Visit the Discord server."
+	set hidden = 1
+	if(config.forumurl)
+		if(alert("This will open the Discord invition link in your browser. Are you sure?",,"Yes","No")=="No")
+			return
+		send_link(src, config.discordurl)
+	else
+		to_chat(src, SPAN_WARNING("The forum URL is not set in the server configuration."))
```
PsiOmegaDelta: Discord URL?

quardbreak: Oh, thanks for pointing out!

interface/interface.dm
```diff
@@ -3,48 +3,60 @@
...
 	else
-		to_chat(src, "<span class='warning'>The forum URL is not set in the server configuration.</span>")
+		to_chat(src, SPAN_WARNING("The forum URL is not set in the server configuration."))
+	return
+
+/client/verb/discord()
+	set name = "Discord"
+	set desc = "Visit the Discord server."
+	set hidden = 1
+	if(config.forumurl)
```
quardbreak: Ree

MistakeNot4892 (approved):

PR #36, opened 2020-02-25
Returns Bearcat back for future purposes

quardbreak:
We have only a example map, would be better to have at least one playable map.
Ported with all replaced path.


MistakeNot4892:
One problem with Bearcat - the away mission uses exactly the same area definitions, so it can cause access inheritance problems. If you could rename the areas on the full-map bearcat to be `/area/ship/bearcat` or something that would address the problem.

Ref - https://github.com/ScavStation/ScavStation/pull/403

quardbreak:
Okay, will fix it later, thanks for noticing this.

MistakeNot4892 (approved):

PR #38, opened 2020-02-26
Broke up clothing.dm into the individual subtype folders.

MistakeNot4892:
Didn't actually change any logic, just the location of code.


afterthought2 (approved):

MistakeNot4892:
````
invidiaul
````

PR #52, opened 2020-02-27
Get stacktrace if deleting uninitialized instance

PsiOmegaDelta:


PsiOmegaDelta:
Will have to ponder how to deal with ladders. Most likely make them LateInitialize.

PsiOmegaDelta:
Ladders have been fixed, now the problem is away sites changing/deleting the same turf multiple times before Init runs
However, I'm actually somewhat unsure if that is a problem in practice

MistakeNot4892:
That sounds like it could conceivably happen if two turf-altering templates overlap.

PsiOmegaDelta:
Travis finally approves

MistakeNot4892 (approved):

code/modules/multiz/structures.dm
```diff
@@ -26,9 +26,12 @@
...
 		find_connections()
 
-/obj/structure/ladder/Initialize(var/ml, var/_mat)
-	. = ..()
-	if(ml)
+/obj/structure/ladder/Initialize(maploading, material)
+	..()
+	return INITIALIZE_HINT_LATELOAD
+
+/obj/structure/ladder/LateInitialize(maploading, material)
```
afterthought2, line 33: While it doesn't matter here, I feel it's correct to always call parent in this proc.

MistakeNot4892: Parent proc doesn't do anything though.

afterthought2: Yes, but if something is added to it (and with the New -> Initialize enforcement that's going to become more and more likely) this would then not see that. Of course here it doesn't matter; I'm suggesting that "Always call parent in LateInitialize" should be something to enforce generally, like "Always return `..()` in examine".

PsiOmegaDelta: I don't mind either way but unlike `/New()` and `/Initialize()` the base of `/LateInitialize()` doesn't do anything of importance

PR #53, opened 2020-02-28
Prevents fake connection spam from crashing the server 

deathride58:
This PR patches one of several major DoS vectors that were exploited during the first few weeks of January of this year. Spamming connection packets on a server that has the DB enabled without proper mitigations such as what's contained in this PR results in the server eventually crashing due to the sheer amount of DB queries that can be generated in a single tick without this. Since there's a few mainstream servers that still don't have this patched, I won't be detailing reproduction steps for crashing servers with the exploit in question.

This PR is based off of the following TGStation PRs: tgstation/tgstation#48529 and tgstation/tgstation#48583

### DIFF
--- a/code/modules/admin/IsBanned.dm
+++ b/code/modules/admin/IsBanned.dm
@@ -1,32 +1,51 @@
 //Blocks an attempt to connect before even creating our client datum thing.
-world/IsBanned(key,address,computer_id)
-	if(ckey(key) in admin_datums)
+/world/IsBanned(key, address, computer_id, type)
+	var/static/key_cache = list()
+	if(type == "world")
+		return ..()
+
+	if(key_cache[key] >= world.time)
+		return list("reason"="concurrent connection attempts", "desc"="You are attempting to connect too fast. Try again.")
+	key_cache[key] = world.time + 10 //This proc shouldn't be runtiming. But if it does, then the expiry time will cover it to ensure genuine connection attempts don't get trapped in limbo.
+
+	var/ckeytext = ckey(key)
+
+	if(ckeytext in admin_datums)
+		key_cache[key] = 0
 		return ..()
 
 	//Guest Checking
 	if(!config.guests_allowed && IsGuestKey(key))
 		log_access("Failed Login: [key] - Guests not allowed")
 		message_admins("<span class='notice'>Failed Login: [key] - Guests not allowed</span>")
+		key_cache[key] = 0
 		return list("reason"="guest", "desc"="\nReason: Guests not allowed. Please sign in with a byond account.")
+
+	var/client/C = GLOB.ckey_directory[ckeytext]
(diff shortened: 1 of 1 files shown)


code/modules/admin/IsBanned.dm
```diff
@@ -1,32 +1,51 @@
...
+	if(type == "world")
+		return ..()
+
+	if(key_cache[key] >= REALTIMEOFDAY)
+		return list("reason"="concurrent connection attempts", "desc"="You are attempting to connect too fast. Try again.")
+	key_cache[key] = REALTIMEOFDAY + 10 //This proc shouldn't be runtiming. But if it does, then the expiry time will cover it to ensure genuine connection attempts don't get trapped in limbo.
+
+	var/ckeytext = ckey(key)
+
+	if(ckeytext in admin_datums)
```
Lohikar: `admin_datums` is assoc.

code/modules/admin/IsBanned.dm
```diff
@@ -1,32 +1,51 @@
...
+		key_cache[key] = 0
 		return ..()
 
 	//Guest Checking
 	if(!config.guests_allowed && IsGuestKey(key))
 		log_access("Failed Login: [key] - Guests not allowed")
 		message_admins("<span class='notice'>Failed Login: [key] - Guests not allowed</span>")
+		key_cache[key] = 0
```
Lohikar, line 21: Why null this for guests?

code/modules/admin/IsBanned.dm
```diff
@@ -1,32 +1,51 @@
...
 		return list("reason"="guest", "desc"="\nReason: Guests not allowed. Please sign in with a byond account.")
 
+	var/client/C = GLOB.ckey_directory[ckeytext]
+	//If this isn't here, then topic call spam will result in all clients getting kicked with a connecting too fast error.
+	if (C && ckeytext == C.ckey && address == C.address && computer_id == C.computer_id)
+		key_cache[key] = 0
```
Lohikar, line 27: Why null this for non-admin players? They shouldn't be logging in more than once every second anyways.

deathride58: If this isn't nulled here, then internal byond bans will result in every client getting kicked for "connecting too fast", since byond's internal ban system calls isbanned on every single connected client, and as a result, checks the ckey against the cache and records it. We found this to be a hard necessity for the server to actually function when trying to patch the exploit over at Citadel.

code/modules/admin/IsBanned.dm
deathride58: Mainly so that guest ckeys don't get stuck in limbo when getting bounced off a guest wall. Less of an issue since key_cache was given expiry times in this implementation, but still worth keeping into account regardless.

MistakeNot4892 (approved):

PR #59, opened 2020-02-28
Adds method to create textbooks in game

SodiumShine:
* Requires Master Literacy (and gives it function)
* Blank textbooks must be printed via an autobinder and then written on
* Codex entry on details included
* Only two textbooks per person, per round
* Also, made textbooks in general smaller


SodiumShine:
This PR is probably messy. Would appreciate suggestions and corrections.

code/game/objects/items/books/skill_book.dm
```diff
@@ -23,28 +23,31 @@
...
 
 	var/decl/hierarchy/skill/skill // e.g. SKILL_LITERACY
 	var/skill_req //The level the user needs in the skill to benefit from the book, e.g. SKILL_PROF
 	var/reading = FALSE //Tto check if the book is actively being used
-	var/custom = FALSE //To bypass init stuff, for player made textbooks and weird books
+	var/custom = FALSE //To bypass init stuff, for player made textbooks and weird books. If true must have details manually set
 	var/ez_read = FALSE //Set to TRUE if you can read it without basic literacy skills
 
 	var/skill_name = "missing skill name"
+	var/progress = 6 // used to track the progress of making a custom book. defaults to 6 so most books count as 'finished'
```
afterthought2: Best practice is to use defines for state like this.

code/game/objects/items/books/skill_book.dm
```diff
@@ -633,11 +642,143 @@ MEDICAL
...
+/obj/item/book/skill/custom/yellow
+	icon_state = "book3"
+/obj/item/book/skill/custom/blue
+	icon_state = "book4"
+/obj/item/book/skill/custom/green
+	icon_state = "book5"
+/obj/item/book/skill/custom/purple
+	icon_state = "book6"
+/obj/item/book/skill/custom/light_blue
+	icon_state = "book7"
```
afterthought2: Haven't looked at the icons, but may be simpler to use color var + overlays.

code/game/objects/items/books/skill_book.dm
```diff
@@ -633,11 +642,143 @@ MEDICAL
...
+	if(istype(W, /obj/item/pen))
+
+		if(!user.skill_check(SKILL_LITERACY, SKILL_BASIC))
+			to_chat(user, SPAN_WARNING("You can't even read, yet you want to write a whole educational textbook?"))
+			return
+		if(!user.skill_check(SKILL_LITERACY, SKILL_PROF))
+			to_chat(user, SPAN_WARNING("You have no clue as to how to write an entire textbook in a way that is actually useful. Maybe a regular book would be better?"))
+			return
+
+		var/choice = input("What would you like to change?","Textbook editing") as null|anything in list("Title", "Author", skill_option_string)
```
afterthought2: Should include `user` at least, in input. Same for the other inputs.

code/game/objects/items/books/skill_book.dm
```diff
@@ -633,11 +642,143 @@ MEDICAL
...
+		switch(choice)
+
+			if("Title")
+				var/newtitle = reject_bad_text(sanitizeSafe(input("Write a new title:")))
+				if(!newtitle)
+					to_chat(usr, "The title is invalid.")
```
afterthought2: Use `user` rather than `usr`.

code/game/objects/items/books/skill_book.dm
```diff
@@ -633,11 +642,143 @@ MEDICAL
...
+					if(user.skill_check(S.type, SKILL_BASIC))
+						skill_choices += S
+				var/decl/hierarchy/skill/skill_choice = input(usr, "What subject does your textbook teach?", "Textbook skill selection") as null|anything in skill_choices
+				if(!skill_choice)
+					to_chat(user, SPAN_WARNING("Textbook skill selection cancelled."))
+					return
+				var/newskill = skill_choice.type
+
+				//Choosing the level
+				var/list/skill_levels = list("Basic" = SKILL_BASIC, "Trained" = SKILL_ADEPT, "Experienced" = SKILL_EXPERT, "Master" = SKILL_PROF)
```
afterthought2: These are best read off from the skill datum's `level` var; there is no guarantee that the skill uses these default names.

code/game/objects/items/books/skill_book.dm
```diff
@@ -633,11 +642,143 @@ MEDICAL
...
+
+		if(!user.skill_check(SKILL_LITERACY, SKILL_BASIC))
+			to_chat(user, SPAN_WARNING("You can't even read, yet you want to write a whole educational textbook?"))
+			return
+		if(!user.skill_check(SKILL_LITERACY, SKILL_PROF))
+			to_chat(user, SPAN_WARNING("You have no clue as to how to write an entire textbook in a way that is actually useful. Maybe a regular book would be better?"))
+			return
+
+		var/choice = input("What would you like to change?","Textbook editing") as null|anything in list("Title", "Author", skill_option_string)
+		switch(choice)
```
afterthought2: After an `input` is resolved, an unpredictable amount of time will have passed and game state can be completely different from what it was before. If any interactivity constraints are expected (e.g. pen in hands, book close to user, etc.) they need to be made here, after every `input`. The most common ones are `CanInteract` and `CanPhysicallyInteract`.

afterthought2:
Also noting that there is a kind of implicit state check in the big switch on the `progress` var, which is then not repeated after `input`. It's possible that the skill book is being interacted with by several mobs at once, so have to be consistent about that.

code/game/objects/items/books/skill_book.dm
SodiumShine: I'm vaguely aware fabricators can pick colours, as seen with pipes, but I wasn't super sure if that was suitable for this.

code/game/objects/items/books/skill_book.dm
SodiumShine: Can you elaborate or give an example?

code/game/objects/items/books/skill_book.dm
SodiumShine: Woops missed that one. Bad copy paste

code/game/objects/items/books/skill_book.dm
SodiumShine: Oh. That's a good point.
Though, I'm not sure how to use the skill's levels list and have them correspond to values though, since the associated values are strings. I'm not actually very on top of using lists at all.

code/game/objects/items/books/skill_book.dm
SodiumShine: Okay, seems simple enough. Just check if they can physically interact after every input.

SodiumShine: Or, wait, should I be checking specifically for the pen/book somehow? CanPhysicallyInteractWith?

SodiumShine:
> 
> 
> Also noting that there is a kind of implicit state check in the big switch on the `progress` var, which is then not repeated after `input`. It's possible that the skill book is being interacted with by several mobs at once, so have to be consistent about that.

I'm sorry I don't understand this at all, apart from the fact I should probably stop multiple mobs from interacting at the same time.

code/game/objects/items/books/skill_book.dm
MistakeNot4892: You'd define something like 
````
#define TEXTBOOK_PROGRESS_NONE 0
...
#define TEXTBOOK_PROGRESS_FINISHED 6
````
then rather than using magic numbers in your code you'd use the defines.

File: code/game/objects/items/books/skill_book.dm (hunk 6 above)
Comment (MistakeNot4892) [reply code/game/objects/items/books/skill_book.dm]: You could check if the pen is still the active hand of the holder and that the book is still close enough (`W == user.get_active_hand() && user.CanPhysicallyInteract(src)`).

File: code/game/objects/items/books/skill_book.dm [hunk 7]
```diff
@@ -633,11 +642,143 @@ MEDICAL
...
+
+			if("Title")
+				var/newtitle = reject_bad_text(sanitizeSafe(input("Write a new title:")))
+				if(!newtitle)
+					to_chat(usr, "The title is invalid.")
+					return
+				else
+					newtitle = usr.handle_writing_literacy(usr, newtitle)
```
MistakeNot4892: More `usr`s!

code/game/objects/items/books/skill_book.dm
MistakeNot4892: You'd just need to use `skill_choice.levels` instead of declaring your own, as `skill_choice` has a list defined on it.

code/game/objects/items/books/skill_book.dm
```diff
@@ -633,11 +642,143 @@ MEDICAL
...
+					to_chat(user, SPAN_WARNING("Textbook skill selection cancelled."))
+					return
+				var/newskill = skill_choice.type
+
+				//Choosing the level
+				var/list/skill_levels = list("Basic" = SKILL_BASIC, "Trained" = SKILL_ADEPT, "Experienced" = SKILL_EXPERT, "Master" = SKILL_PROF)
+				for(var/SL in skill_levels)
+					if(!user.skill_check(newskill, skill_levels[SL]))
+						LAZYREMOVE(skill_levels, SL)
+				var/newskill_level = input(usr, "What level of education does it provide?","Textbook skill level") as null|anything in skill_levels
```
MistakeNot4892: You'll probably want/need to check CanPhysicallyInteract and the presence of the pen every time you do `input()`.

code/game/objects/items/books/skill_book.dm
MistakeNot4892: You'll also need to check `progress` as it might have changed.

SodiumShine: Think I got this one.

code/game/objects/items/books/skill_book.dm
SodiumShine: Okay, think I got this one.

SodiumShine:
Gonna work on this later, this might sit here for a bit. DNM of course.

afterthought2:
The menu the user sees is determined by 
```
var/choice = input("What would you like to change?","Textbook editing") as null|anything in list("Title", "Author", skill_option_string)
```
Here, `skill_option_string` is tracking state on the book, and determines which menus you see. That means that within each menu, after input or do_after or whatever, you need to make sure that that state is still valid, i.e. the user is supposed to still have access to that menu.

SodiumShine:
> 
> 
> The menu the user sees is determined by
> 
> ```
> var/choice = input("What would you like to change?","Textbook editing") as null|anything in list("Title", "Author", skill_option_string)
> ```
> 
> Here, `skill_option_string` is tracking state on the book, and determines which menus you see. That means that within each menu, after input or do_after or whatever, you need to make sure that that state is still valid, i.e. the user is supposed to still have access to that menu.

Okay I understand now. I think though it only needs to be checked for "Skill" since all other scenarios shouldn't matter (changing the author or title isn't related to skill content). You can't change back from "Continue writing" and that already has checks of its own. Though on this note I think i should have it change to "Cancel" rather than giving a dumb failure message

code/game/objects/items/books/skill_book.dm
SodiumShine: Having the weirdest issue with this. I've done what I thought would be the best way to do this... and the skill datums are being deleted, and not just removed from what I thought I made as a temporary list. Viewing the skill panel shows the skill levels that get removed from selection lists as simply not being present.
```//Choosing the skill
var/list/skill_choices = list()
for(var/decl/hierarchy/skill/S in GLOB.skills)
	if(user.skill_check(S.type, SKILL_BASIC))
		LAZYADD(skill_choices, S)
var/decl/hierarchy/skill/skill_choice = input(user, "What subject does your textbook teach?", "Textbook skill selection") as null|anything in skill_choices
if(!can_write(user,W) || progress == SKILLBOOK_PROG_FINISH)
	return
if(!skill_choice)
	to_chat(user, SPAN_WARNING("Textbook skill selection cancelled."))
	return
var/newskill = skill_choice.type

//Choosing the level
var/list/skill_levels = skill_choice.levels
for(var/SL in skill_levels)
	if(!user.skill_check(newskill, skill_levels.Find(SL)))
		LAZYREMOVE(skill_levels, SL)
	else
		skill_levels[SL] = skill_levels.Find(SL) + 1
LAZYREMOVE(skill_levels,skill_levels[1])
var/newskill_level = input(user, "What level of education does it provide?","Textbook skill level") as null|anything in skill_levels
```

code/game/objects/items/books/skill_book.dm
SodiumShine: Okay, got this working.

code/game/objects/items/books/skill_book.dm
SodiumShine: Woops nevermind. Good now.

SodiumShine:
Still not done, I wanna tidy it up some more but just updating the PR all the same.

SodiumShine:
All done now, I hope!

MistakeNot4892:
Travis is being weird, trying to force him to tidy up.

MistakeNot4892 (approved):
Minor question @ you in Discord, repeating here in case: do skillbooks add themselves to persistence tracking? Will probably need some kind of exception for that if they do.

SodiumShine:
> 
> 
> Minor question @ you in Discord, repeating here in case: do skillbooks add themselves to persistence tracking? Will probably need some kind of exception for that if they do.

Stating here as well for posterity: Skillbooks are `unique` so they should not persist.

PR #61, opened 2020-02-29
Sound and emotions related changes

quardbreak:
DNM, Work in progress, didn't test yet.
* Cleanups sound.dm file, moves defines and global variables to define and globvar folders
* Adds quiet var to playsound() in case if sound don't must go through walls
* Moves emotions sound feature to core level so you can make claps have a sound
   * Makes emotions support gender
* Makes species speech support gender check


code/modules/emotes/emote_define.dm
```diff
@@ -78,18 +80,29 @@
...
-			M.audible_message(message = use_3p, self_message = use_1p, deaf_message = emote_message_impaired, checkghosts = /datum/client_preference/ghost_sight)
-		else
-			M.visible_message(message = use_3p, self_message = use_1p, blind_message = emote_message_impaired, checkghosts = /datum/client_preference/ghost_sight)
+	if(message_type == AUDIBLE_MESSAGE)
+		if(isliving(user))
+			var/mob/living/L = user
+			if(!L.silent)
+				user.audible_message(message = use_3p, self_message = use_1p, deaf_message = emote_message_impaired, checkghosts = /datum/client_preference/ghost_sight)
+			else
+				user.visible_message(message = "[user] noiselessly opens mouth!", self_message = "You cannot say anything!", blind_message = emote_message_impaired, checkghosts = /datum/client_preference/ghost_sight)
```
MistakeNot4892: `[user] opens their mouth silently` would be better grammar.

code/modules/emotes/emote_define.dm
```diff
@@ -78,18 +80,29 @@
...
 
 /decl/emote/proc/do_extra(var/atom/user, var/atom/target)
 	return
 
+/decl/emote/proc/do_sound(var/atom/user)
+	if(emote_sound)
+		if(islist(emote_sound))
+			if(emote_sound[user.gender])
+				return playsound(user.loc, pick(emote_sound[user.gender]), 50, 0)
+		return playsound(user.loc, pick(emote_sound), 50, 0)
```
MistakeNot4892: It looks like this should be tabbed in one step.
You can also remove `if(emote_sound)` because `islist()` will fail if it is null/false.

MistakeNot4892 (approved):

PR #65, opened 2020-03-01
Removing 'as mob|obj|turf' from procs.

MistakeNot4892:
This was definitely a good use of my morning.


afterthought2 (approved):

deathride58 (approved):
image
This is the very first time I've used the viewed files feature. Mad respect, hot dam. This PR looks good after glancing at all the changes, it just removes the `as obj` stuff and replaces a few `AM as mob|obj` args with `atom/movable/AM` args.
LGTM

PR #67, opened 2020-03-01
Wall machines to construct decls.

afterthought2:
Currently air and fire alarms. Still need to do apcs, switches, lights, etc.

If building a new one from scratch, it will require you to install stock components like the power one, the keyboard, the screen, which you'll have to print yourself. I'm open to having printed frames come with those already and install them automatically, so long as frame construction is moved from stack recipes to autolathe.


code/game/machinery/alarm.dm
```diff
@@ -129,16 +133,8 @@
 	unregister_radio(src, frequency)
 	return ..()
 
-/obj/machinery/alarm/Initialize(mapload, var/dir, atom/frame)
-	. = ..(mapload, dir)
-
-	if(istype(frame))
-		buildstage = 0
-		wiresexposed = 1
-		pixel_x = (dir & 3)? 0 : (dir == 4 ? -21 : 21)
```
comma, line 138: wouldn't removing this hackery break the mapped alarms offsets?

afterthought2: It's also set in `update_icon`.

afterthought2:
Feature complete but needs testing.

comma:
- Air / fire alarm circuitboards have no costs in lathe
- Cannot build air alarm from frame. Frame ignores being hit with circuitboard. Also doesn't set offset until alarm is finished it seems.
- Cannot build fire alarm from frame. Ignores circuitboard too.
- When bulding air alarm from kit, it doesn't automatically update icon to ready state, looks unfinished and all the wall. Otherwise works fine.
- No issues when building fire alarm from kit.

comma (approved):

PR #72, opened 2020-03-02
[DNM] Upgrades webhook info

quardbreak:
* Adds new webhook - a round preparation. Calls when Master fully initialized, so players can know when they're can join before round start. 
  * Now can click on the server link to join!
* Upgrades round end webhook. Now it show some information like players amount and round duration, plus, little format design.

Require some help to test this. I dunno how to test it in local environment.

### DIFF
--- a/code/modules/webhooks/webhook_roundprep.dm
+++ b/code/modules/webhooks/webhook_roundprep.dm
@@ -1,0 +1,15 @@
+/decl/webhook/roundprep
+	id = WEBHOOK_ROUNDPREP
+
+// Data expects a "url" field pointing to the current hosted server and port to connect on.
+/decl/webhook/roundprep/get_message(var/list/data)
+	. = ..()
+	var/desc = "The server has been started!\n"
+	desc += "Map: **[station_name()]**\n"
+	desc += "Address: <[get_world_url()]>"
+
+	.["embeds"] = list(list(
+		"title" = "New round preparation.",
+		"description" = desc,
+		"color" = COLOR_WEBHOOK_DEFAULT
+	))
--- a/code/modules/webhooks/webhook_roundstart.dm
+++ b/code/modules/webhooks/webhook_roundstart.dm
@@ -4,13 +4,14 @@
 // Data expects a "url" field pointing to the current hosted server and port to connect on.
 /decl/webhook/roundstart/get_message(var/list/data)
 	. = ..()
-	var/desc = "A new round is starting"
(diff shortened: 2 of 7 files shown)


code/modules/webhooks/webhook_roundstart.dm
```diff
@@ -4,13 +4,11 @@
 // Data expects a "url" field pointing to the current hosted server and port to connect on.
 /decl/webhook/roundstart/get_message(var/list/data)
 	. = ..()
-	var/desc = "A new round is starting"
-	if(data && data["url"])
-		desc += " on [data["url"]]"
-	desc += "."
+	var/desc = "Gamemode: [SSticker.master_mode]\n"
+	desc += "Players: [GLOB.player_list.len]"
```
MistakeNot4892, line 8: No URL?

code/modules/webhooks/webhook_submap.dm
```diff
@@ -7,10 +7,10 @@
 	var/submap_name = data && data["name"]
 	if(submap_name)
 		submap_name = ", [submap_name],"
-	var/desc = "A submap[submap_name] is currently available."
+	var/desc = "A submap[submap_name] is currently available"
```
MistakeNot4892: I can more or less agree with removing the periods from titles, but I really don't think removing them from the contents is good. It's a full sentence, it should be grammatically correct.

code/modules/webhooks/webhook_roundprep.dm
```diff
@@ -0,0 +1,15 @@
+/decl/webhook/roundprep
+	id = WEBHOOK_ROUNDPREP
+
+// Data expects a "url" field pointing to the current hosted server and port to connect on.
+/decl/webhook/roundprep/get_message(var/list/data)
+	. = ..()
+	var/desc = "The server has been started!\n"
+	desc += "Map: **[station_name()]**\n"
+	desc += "Adress: <[get_world_url()]>"
```
MistakeNot4892: Address.

code/modules/webhooks/webhook_roundstart.dm
quardbreak: ~~Not sure what people want to join in game from roundend webhook. I made special hook where you can just click and join. Can add if you want.~~ Sorry, didn't see the context. 
Btw, think this PR need some discussion how those webhooks must look like.

MistakeNot4892:
Feel free to reopen when ready for merge.

PR #74, opened 2020-03-03
Converts unarmed attacks to decls, adds per-species natural attack tracking.

MistakeNot4892:
This was sorta kinda in previously, not sure where it went.
Re-implements https://github.com/NebulaSS13/NebulaFeatureRequests/issues/6.
- Unarmed attacks are now properly singletons and are passed around and tracked as types except when needed for actual processing.
- The state of your bodyparts and the associated species will determine which unarmed attacks you have available.

### DIFF
--- a/code/modules/mob/living/simple_animal/simple_animal.dm
+++ b/code/modules/mob/living/simple_animal/simple_animal.dm
@@ -257,12 +257,13 @@
 		if(I_HURT)
 			var/dealt_damage = harm_intent_damage
 			var/harm_verb = response_harm
-			if(ishuman(M) && M.species)
-				var/datum/unarmed_attack/attack = M.get_unarmed_attack(src)
-				dealt_damage = attack.damage <= dealt_damage ? dealt_damage : attack.damage
-				harm_verb = pick(attack.attack_verb)
-				if(attack.sharp || attack.edge)
-					adjustBleedTicks(dealt_damage)
+			if(ishuman(M))
+				var/decl/natural_attack/attack = M.get_unarmed_attack(src)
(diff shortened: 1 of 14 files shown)


MistakeNot4892:
Tested:
- selecting and using default attacks
- trying to select a default attack after amputating both my hands.

I identified a bug (#75) with trying to use an unarmed attack with two broken hands, but otherwise this appears to be working fine.

code/modules/mob/living/simple_animal/simple_animal.dm
```diff
@@ -257,12 +257,13 @@
...
 			var/dealt_damage = harm_intent_damage
 			var/harm_verb = response_harm
-			if(ishuman(M) && M.species)
-				var/datum/unarmed_attack/attack = M.get_unarmed_attack(src)
-				dealt_damage = attack.damage <= dealt_damage ? dealt_damage : attack.damage
-				harm_verb = pick(attack.attack_verb)
-				if(attack.sharp || attack.edge)
-					adjustBleedTicks(dealt_damage)
+			if(ishuman(M))
+				var/decl/natural_attack/attack = M.get_unarmed_attack(src)
```
comma, line 261: wots happening here?
simple animal is getting victim's unarmed attack?.. and pretending to attack with it?.. but `get_unarmed_attack` doesn't take args?..

MistakeNot4892: `M` is the person attacking not the victim. Not sure why it's passing an arg though.

comma (approved):

PR #77, opened 2020-03-05
Made catwalks deconstructable.

MistakeNot4892:


deathride58 (approved):

code/game/objects/structures/catwalk.dm
```diff
@@ -36,8 +38,12 @@
 	update_icon()
 
 /obj/structure/catwalk/Destroy()
+	var/turf/oldloc = loc
 	redraw_nearby_catwalks()
-	return ..()
+	. = ..()
+	if(istype(oldloc))
+		for(var/atom/movable/AM in oldloc)
+			oldloc.Entered(AM, oldloc)
```
afterthought2: There is code that assumes this is only called if `oldloc != src`. I'd just call `fall` on every movable instead, but if there is some reason that's not possible, you'll need to go through `Entered` overrides and insert `oldloc != src` checks, as well as moved event subscribers (possibly; those probably don't make this assumption).

MistakeNot4892: fall() turned out to work better than expected.

afterthought2 (approved):

code/game/objects/structures/lattice.dm
```diff
@@ -38,8 +38,10 @@
 /obj/structure/lattice/Destroy()
 	var/turf/old_loc = get_turf(src)
 	. = ..()
-	if(old_loc)
+	if(istype(old_loc))
```
Lohikar, line 41: Minor note: technically `isturf()` is faster than istype, but it doesn't matter here.

PR #80, opened 2020-03-05
Ports Individual Say Log panel from OnyxStation

quardbreak:
Adds a panel which contain (currently) say logs and ooc logs. 
They're temporary and deletes with mob.
You can add new types in list by making define in list and use it with special proc
image

### DIFF
--- a/code/game/verbs/ooc.dm
+++ b/code/game/verbs/ooc.dm
@@ -3,6 +3,7 @@
 	set category = "OOC"
 
 	sanitize_and_communicate(/decl/communication_channel/ooc, src, message)
+	mob.log_individual_message(sanitize(message), INDIVIDUAL_OOC_LOG, key_name(src, include_name = FALSE))
 
 /client/verb/looc(message as text)
 	set name = "LOOC"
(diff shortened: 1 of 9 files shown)


code/game/verbs/ooc.dm
```diff
@@ -3,6 +3,7 @@
 	set category = "OOC"
 
 	sanitize_and_communicate(/decl/communication_channel/ooc, src, message)
+	mob.log_individual_message(sanitize(message), INDIVIDUAL_OOC_LOG, key_name(src, include_name = FALSE))
```
PsiOmegaDelta, line 6: Communication channel logging is handled by setting https://github.com/NebulaSS13/Nebula/blob/dev/code/datums/communication/channel.dm#L8 approparitely

PsiOmegaDelta (changes requested):

code/game/verbs/ooc.dm
quardbreak: Don't really get how this work

quardbreak:
This is pretty fast but I will close for now. I'm not sure what I can handle more than 3 PR's at this moment and maintain they all so maybe reopen when I get more attention to adapt this without mess.

PR #81, opened 2020-03-05
Machinery components damage

comma:
Machines now take damage from emp/explosions/bullets 
They pass that damage to their components.
First, shielding components are damaged
If there's still damage, non-circuitboard components are damaged
And lastly circuitboard is damaged.

Components that cannot be taken out / replaced are not damaged.

Shielding components are special ones that have no other role but to soak up damage.

Damaged components will make machine consider them missing for purposes of setting broken state.
Axed computer parts damage system in lieu of this generic one
SMES does a little boom if it has above 50% charge and one of coils fails, removed bigger effects and its separate damage system.

Most construction states now let you see what components (and in what state) are installed, a-la sci HUD.

Made ship sensors properly constructible so they can have components and get damaged by overheating.


code/game/machinery/_machines_base/machinery_damage.dm
```diff
@@ -0,0 +1,77 @@
+/obj/machinery/proc/take_damage(amount, damtype, silent = TRUE)
+	//Let's not bother initializing all the components for nothing
+	if(amount <= 0)
+		return
+	if(damtype != BRUTE && damtype != BURN && damtype != ELECTROCUTE)
+		return
+	if(!silent)
+		var/hitsound = 'sound/effects/hit_on_shattered_glass.ogg'
+		if(damtype == ELECTROCUTE)
+			hitsound = "sparks"
```
afterthought2, line 10: Shouldn't it be a file path?

code/game/machinery/_machines_base/machinery_damage.dm
```diff
@@ -0,0 +1,77 @@
...
+		if(istype(component, /obj/item/stock_parts/circuitboard))
+			continue
+		// Don't damage what can't be repaired
+		if(!(component.part_flags & PART_FLAG_HAND_REMOVE))
+			continue
+		if(component.is_functional())
+			return component
+	for(var/path in uncreated_component_parts)
+		if(uncreated_component_parts[path])
+			return force_init_component(path)
```
afterthought2: Probably need while loops in here.

code/game/machinery/_machines_base/machinery_damage.dm
```diff
@@ -0,0 +1,77 @@
...
+	
+/obj/machinery/proc/get_damageable_component()
+	var/list/victims = component_parts.Copy()
+	victims = shuffle(victims)
+	for(var/obj/item/stock_parts/component in victims)
+		// Circuitboards are handled separately
+		if(istype(component, /obj/item/stock_parts/circuitboard))
+			continue
+		// Don't damage what can't be repaired
+		if(!(component.part_flags & PART_FLAG_HAND_REMOVE))
```
afterthought2: I suggest just a part flag for damageable; put it on the board. Note that boards are not hand-remove, by the way.

code/game/machinery/_machines_base/machinery_damage.dm
```diff
@@ -0,0 +1,77 @@
...
+
+/obj/machinery/ex_act(severity)
+	take_damage(100/severity, BRUTE)
+
+/obj/machinery/bullet_act(obj/item/projectile/P, def_zone)
+	. = ..()
+	take_damage(P.damage, P.damage_type)
+
+/obj/machinery/attackby(obj/item/W, mob/user)
+	..()
```
afterthought2, line 73: Should really return `..()`.

code/game/machinery/_machines_base/machinery_damage.dm
```diff
@@ -0,0 +1,79 @@
...
+	if(!(W.item_flags & ITEM_FLAG_NO_BLUDGEON) && W.force > 5)
+		user.setClickCooldown(W.attack_cooldown + W.w_class)
+		take_damage(W.force, W.damtype)
```
afterthought2, line 77: Maybe harm only? There are a billion machine interactions, so you'll have otherwise necessary interactions doing damage (think insertion into vendor, etc).

On the other hand, there are some unarmed attack things that could do damage from dumb species; see APCs for the current handling.

code/game/machinery/_machines_base/machinery_damage.dm
comma: Nah, if string is passed, it uses some proc that gets a random sound from a corresponding global list.

code/game/machinery/_machines_base/machinery_damage.dm
comma: It just needs first damageable one, and since uncreated ones are created right here and now, they usually have health to spare.
Just realized if I do flag for being damageable they'll need a check here, will add that

code/game/machinery/_machines_base/machinery_damage.dm
comma: yeah will do that

code/game/machinery/_machines_base/machinery_damage.dm
comma: well usually things only call ..() in attackby if they didn't do any special interactions, since it'll end at 'Urist hits thing with rock!' handling.
Will check if it gets in the way anywhere, can slap intent check.

afterthought2: Just think it will be annoying when you mess up some interaction and it falls through to here. When you get that red message, at least you don't have to fix the machine afterwards.

code/game/machinery/_machines_base/machinery_damage.dm
comma: done

code/game/machinery/_machines_base/machinery_damage.dm
comma: done

code/game/machinery/_machines_base/machinery_damage.dm
comma: done

code/modules/overmap/ships/computers/sensors.dm
```diff
@@ -193,7 +153,7 @@
 			s.set_up(3, 1, src)
 			s.start()
 
-			take_damage(rand(10,50))
+			take_damage(rand(10,50), BURN)
```
comma: TODO: make sure sensors aren't instadying from overheating

code/modules/power/smes.dm
```diff
@@ -332,7 +332,7 @@
 	if(!output_attempt)
 		outputting = 0
 
-/obj/machinery/power/smes/proc/take_damage(var/amount)
+/obj/machinery/power/smes/take_damage(var/amount)
```
comma: TODO: Do something about this shit

code/game/machinery/_machines_base/machinery_damage.dm
comma: Done

afterthought2 (approved):

PR #87, opened 2020-03-06
Ports Bluespace RPED

quardbreak:
Ports bluespace version of rped from Infinity Station (ported from tgstation).
Adds two new sounds from tgstation for rped part replace act


code/modules/research/part_replacer.dm
```diff
@@ -13,6 +13,34 @@
 	max_w_class = ITEM_SIZE_NORMAL
 	max_storage_space = 100
 
+	var/replace_sound = 'sound/items/rped.ogg'
+	var/remote_interaction = FALSE
+
+/obj/item/storage/part_replacer/proc/part_replacement_sound()
+	playsound(src, replace_sound, 40, 1)
+
+/obj/item/storage/part_replacer/afterattack(obj/machinery/machine, mob/living/carbon/human/user, flag, params)
```
afterthought2: Please no; this is handled in machine/attackby already on machines that support it. Forward the ranged interaction there if need be, or directly to the construct state's attackby if you prefer.

MistakeNot4892: `attackby` won't be called on range click - isn't this the only way to handle ranged interactions without doing it at the mob level?

afterthought2: I meant, either call `attackby` from within this proc (if ranged), or call the decl's attackby from this proc, or call another new proc on the decl which forwards to attackby or does the required interaction.

afterthought2:
I mean, this still bypasses construct decls entirely.

afterthought2 (approved):

PR #97, opened 2020-03-08
Adds a chem conflict unit test.

### LINKED ISSUE #84: Chem making conflicts
#### Description of issue

Attempting to make Lubricant also makes Oxygel

Making Lubricant should make lubricant, and making Oxygel should make oxygel. not both.

#### Steps to reproduce

Make lubricant. You'll get oxygel and lubricant

Chemlab dispencer

At least a session

??

#### Issue bingo

afterthought2:
Am open to other recipe changes.

Closes #84.

### DIFF
--- a/code/modules/reagents/reactions/reaction_drinks.dm
+++ b/code/modules/reagents/reactions/reaction_drinks.dm
@@ -47,6 +47,7 @@
 	required_reagents = list(/datum/reagent/drink/milk = 5, /datum/reagent/nutriment/coco = 1)
 	result_amount = 5
 	mix_message = "The solution thickens into a creamy brown beverage."
+	maximum_temperature = 70 CELSIUS
 
 /datum/chemical_reaction/recipe/nothing
 	name = "Nothing"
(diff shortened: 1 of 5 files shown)


code/modules/reagents/reactions/reaction_drinks.dm
```diff
@@ -47,6 +47,7 @@
 	required_reagents = list(/datum/reagent/drink/milk = 5, /datum/reagent/nutriment/coco = 1)
 	result_amount = 5
 	mix_message = "The solution thickens into a creamy brown beverage."
+	maximum_temperature = 70 CELSIUS
```
MistakeNot4892, line 50: I think you may have wanted minimum temperature for this?

afterthought2: This one is max, other is min, so they don't overlap. If you want chocolate, heat it up; if  you want chocolate milk, do it cold.

MistakeNot4892: Ah fair enough.

MistakeNot4892 (approved):

PR #104, opened 2020-03-09
Generalized sparring/boxing gloves.

MistakeNot4892:
Slowly dissecting the mess that is human attack_hand.


afterthought2 (approved):
Not super sure about the AND/OR logic with multiple bodyparts and only some padded, but probably doesn't matter much.

MistakeNot4892:
We really need a way to unify bodypart coverage and limb IDs.

PR #113, opened 2020-03-09
End Round button

quardbreak:
Adds the round end button. The difference between reboot - it makes game end smoothly with credits and round result print.
Formatted some button names and removed admin name duplication for some logs.
Adds coordinate output log for the Create Object panel, moves the spawn button to right from the Number of matched field: 
image


code/modules/admin/admin.dm
```diff
@@ -848,6 +848,24 @@ var/global/floorIsLava = 0
...
 
+/datum/admins/proc/endnow()
+	set category = "Server"
+	set desc = "Ending game round"
+	set name = "End Round"
+	if(!usr.client.holder || !check_rights(R_ADMIN))
+		return
+
+	if(GAME_STATE != RUNLEVEL_GAME)
+		to_chat(usr, "<span class='bigdanger'>You can use this only with the active round.</span>")
```
MistakeNot4892: Suggest swapping this to `The round has not started yet!`.

code/modules/admin/admin.dm
```diff
@@ -907,12 +925,12 @@ var/global/floorIsLava = 0
...
 
 	config.ninjas_allowed = !config.ninjas_allowed
-	log_and_message_admins("toggled Space Ninjas [config.ninjas_allowed ? "on" : "off"].")
+	log_and_message_admins("toggled space ninjas [config.ninjas_allowed ? "on" : "off"].")
 	SSstatistics.add_field_details("admin_verb","TSN") //If you are copy-pasting this, ensure the 2nd parameter is unique to the new proc!
 
 /datum/admins/proc/delay()
 	set category = "Server"
-	set desc="Delay the game start/end"
+	set desc="Delay the Game start/end"
```
MistakeNot4892: Why make this uppercase?

code/modules/admin/admin.dm
```diff
@@ -931,23 +949,23 @@ var/global/floorIsLava = 0
 
 /datum/admins/proc/adjump()
 	set category = "Server"
-	set desc="Toggle admin jumping"
+	set desc="Toggle Admin Jumping"
```
MistakeNot4892: I don't understand why you made the ones above lowercase and then made these ones uppercase.

MistakeNot4892 (approved):

code/modules/admin/topic.dm
```diff
@@ -1713,7 +1713,7 @@
 									var/mob/M = O
 									M.real_name = obj_name
 
-		log_and_message_admins("created [number] [english_list(paths)]")
+		log_and_message_admins("created [number] [english_list(paths)] at ([target.x],[target.y],[target.z])")
```
PsiOmegaDelta, line 1716: `log_and_message_admins()` already print X/Y/Z information if supplied a valid target

PR #115, opened 2020-03-09
Various tweaks.

MistakeNot4892:


code/game/machinery/pipe/construction.dm
```diff
@@ -24,7 +24,7 @@ Buildable meters
 /obj/item/pipe/Initialize(var/mapload, var/obj/machinery/atmospherics/P)
 	. = ..(mapload, null)
 	set_extension(src, /datum/extension/parts_stash)
-	if(!P)
+	if(!istype(P))
```
PsiOmegaDelta, line 27: Do we want to silently ignore inappropriate instances?

MistakeNot4892: I could throw in a `crash_with` I suppose.

MistakeNot4892: Turns out silently failing is the only way any of this code actually works.

comma (approved):

deathride58 (approved):

PR #116, opened 2020-03-09
Allows in-hand crafting recipes to generate their material costs from their product.

MistakeNot4892:
- Crafting recipes with no `req_amount` set will try to generate one from the matter of their product.
- All recipes will pass material information. Stuff that doesn't care about material information should handle it gracefully anyway.


code/game/objects/items/stacks/stack.dm
```diff
@@ -355,6 +363,15 @@
 		difficulty +=  material.construction_difficulty
 	if(reinforce_material)
 		use_reinf_material = reinforce_material
+	if(result_type && isnull(req_amount))
+		req_amount = 0
+		var/obj/item/I = new result_type
```
comma: wot about non-items like structures?

MistakeNot4892:
Tested, ready for review.

code/modules/materials/_recipe.dm
```diff
@@ -0,0 +1,74 @@
...
+		use_material = material.type
+		difficulty +=  material.construction_difficulty
+	if(reinforce_material)
+		use_reinf_material = reinforce_material
+	if(result_type && isnull(req_amount))
+		req_amount = 0
+		var/obj/I = new result_type
+		var/list/building_cost = I.building_cost()
+		for(var/path in building_cost)
+			req_amount += building_cost[path]
```
afterthought2: Maybe worth adding a unit test that basically says: if you build a stack recipe, and then you retrieve its building cost, you don't wind up with (a) materials you didn't use in building it or (b) more of materials you used to build it than you started with. I don't think the logic here guarantees that such a test would always pass.

MistakeNot4892: Added!

MistakeNot4892:
Hnnnnnn it's passing Travis now but still not actually working. Materials are all ending up null.

MistakeNot4892:
Closing in on finishing this one. R&D is working, just need to determine why random structures are getting a cost of 1 sheet despite having 40k steel in their matter list.

MistakeNot4892:
Too low:
- steel closet (1 sheet)
- steel Pressure Tank (1 sheet)
- steel cannon frame (1 sheet)
- steel coffin (1 sheet)
- steel meat hook (1 sheet)
- steel modular tablet frame (1 sheet)
- steel table frame (1 sheet)

Too high:
- padded beige steel chair (50 sheets)
- dark office chair (50 sheets)
- beige comfy office chair (50 sheets)
- beige steel comfy chair (50 sheets)
- beige steel armchair (50 sheets)

MistakeNot4892:
I'm reigning the scope of this back and just committing the code, not making anything actually use it for now. Material costs are all over the place and matter/material setting chains are wildly inconsistent. Will address them in another PR.

comma (dismissed):

deathride58 (approved):

PR #117, opened 2020-03-10
Deleted actors.

MistakeNot4892:
We don't have them mapped and the verb bypasses huge chunks of the antag system - not sure if it ever worked in its current state.


afterthought2:
Feel like adding a map var or landmark check would be sufficient here. It does work fine on maps with the thing mapped (bay).

MistakeNot4892:
The proc the verb calls skips checking for the landmark entirely. It just creates a naked human mob.
EDIT: I would be interested in seeing if it does actually work on Bay, and if it does, checking how the code differs.

MistakeNot4892:
Never mind, spotted a line I missed that makes this code work overall.

PR #118, opened 2020-03-12
Adds a pipenet unit test.

afterthought2:
My suspicion is that this usually fails.

Note that unlike the pipeline test just above it, which is a mapping test for the most part, this is testing the network creation backend.

### DIFF
--- a/code/unit_tests/atmospherics_tests.dm
+++ b/code/unit_tests/atmospherics_tests.dm
@@ -293,6 +293,26 @@
 		pass("All pipes belonged to a unique pipeline.")
 	return 1
 
+/datum/unit_test/pipelines_shall_belong_to_unique_pipenets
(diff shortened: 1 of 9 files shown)


code/unit_tests/atmospherics_tests.dm
```diff
@@ -293,6 +293,26 @@
...
+
+/datum/unit_test/pipelines_shall_belong_to_unique_pipenets/start_test()
+	var/list/checked_pipelines = list()
+	var/list/bad_pipenets = list()
+	for(var/datum/pipe_network/network)
+		for(var/thing in network.line_members)
+			if(!checked_pipelines[thing])
+				checked_pipelines[thing] = network
+				continue
+			LAZYDISTINCTADD(bad_pipenets[network], thing)
```
PsiOmegaDelta: Any particular reason you're using the LAZY macros with an initialized list?

comma (approved):

PR #120, opened 2020-03-12
Makes airlock deny sound play even without secure wires

quardbreak:
* Moves all airlock subtypes from airlock core file to subtype file
* Removes secured wires variable requirement to play access denied sound
  * Added speaker wire for airlocks
* ~~Adds coordinate output for door access unit test~~


code/game/machinery/doors/door.dm
```diff
@@ -86,7 +86,7 @@
 	health = maxhealth
 #ifdef UNIT_TEST
 	if(autoset_access && length(req_access))
-		crash_with("A door with mapped access restrictions was set to autoinitialize access.")
+		crash_with("A door with mapped access restrictions was set to autoinitialize access at [x], [y], [z].")
```
afterthought2: Coordinates should already appear in the runtime error's header on `src`.

comma (approved):
yolo

PR #126, opened 2020-03-13
Adds simply fullscreen mode

quardbreak:
image

### DIFF
--- a/code/modules/client/preference_setup/global/preferences.dm
+++ b/code/modules/client/preference_setup/global/preferences.dm
@@ -191,7 +191,8 @@ var/list/_client_preferences_by_type
 	default_value = GLOB.PREF_NO
 
 /datum/client_preference/fullscreen_mode/changed(mob/preference_mob, new_value)
-	toggle_fullscreen(preference_mob.client, new_value == GLOB.PREF_YES)
+	if(preference_mob.client)
+		preference_mob.client.toggle_fullscreen(new_value == GLOB.PREF_YES)
 
 /datum/client_preference/autohiss
 	description = "Autohiss"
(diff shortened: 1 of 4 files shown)


code/modules/client/preference_setup/global/preferences.dm
```diff
@@ -223,6 +223,14 @@ var/list/_client_preferences_by_type
 	key = "EXAMINE_MESSAGES"
 	options = list(GLOB.PREF_SHOW, GLOB.PREF_HIDE)
 
+/datum/client_preference/fullscreen_mode
+	description = "Toggle Fullscreen Mode"
+	key = "TOGGLE_FULLSCREEN"
+	options = list(GLOB.PREF_NO, GLOB.PREF_YES)
```
PsiOmegaDelta: Any particular reason for switching Yes/No around? It makes it inconsistent with other settings using the same options.
Should instead be able to instead do `default_value = GLOB.PREF_NO`

code/modules/client/preference_setup/global/preferences.dm
```diff
@@ -181,6 +185,15 @@ var/list/_client_preferences_by_type
 	key = "BROWSER_STYLED"
 	options = list(GLOB.PREF_FANCY, GLOB.PREF_PLAIN)
 
+/datum/client_preference/fullscreen_mode
+	description = "Fullscreen Mode"
+	key = "FULLSCREEN"
+	options = list(GLOB.PREF_YES, GLOB.PREF_NO)
```
PsiOmegaDelta: This is the same as the default `options` value

MistakeNot4892 (approved):

PR #127, opened 2020-03-13
Moves admin login/logout message to client path

quardbreak:
Adds login message so admins can know who connected to server with rights. 
Not really sure if here difference between mob and client log/out, but I moved all process in client since we grant admin powers in client's login.

### DIFF
--- a/code/modules/client/client_procs.dm
+++ b/code/modules/client/client_procs.dm
@@ -168,6 +168,7 @@
 	if(holder)
 		GLOB.admins += src
 		holder.owner = src
+		handle_staff_login()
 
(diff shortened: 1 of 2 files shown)


MistakeNot4892 (approved):

PsiOmegaDelta:
Mob logout is triggered whenever a client switches mob or disconnects/shuts down BYOND
Client logout is sometimes triggered when the client disconnects/shuts down BYOND, sometimes not

afterthought2:
I guess the upside is that this avoids redundant login/logout messages when admins are transferring mobs. I don't believe `mob/Logout` is much more consistent then `client/Logout`, and in any case, the message here is really not mission-critical.

PR #128, opened 2020-03-14
Ports emergency dehumidifier from EuropaStation

quardbreak:
Fast port of the dehumidifier created by EmperorJon (https://github.com/Yonaguni/EuropaStation/pull/817) 
It allows to drain water from a fluid sources (work like portable drain)


code/game/machinery/dehumidifier.dm
```diff
@@ -0,0 +1,123 @@
...
+
+		T.remove_fluid(ceil(fluid_here*drainage))
+		cell.use(drainage*5000*CELLRATE)
+		T.show_bubbles()
+		if(world.time > last_gurgle + 80)
+			last_gurgle = world.time
+			playsound(T, pick(SSfluids.gurgles), 50, 1)
+	else
+		on = 0
+		power_change()
```
afterthought2: Never call this proc on self.

code/game/machinery/dehumidifier.dm
```diff
@@ -0,0 +1,123 @@
...
+/obj/machinery/dehumidifier/Process()
+	if(!on)
+		return
+
+	if(cell && cell.charge)
+		var/turf/T = get_turf(src)
+		if(!istype(T)) return
+		var/fluid_here = T.get_fluid_depth()
+		if(fluid_here <= 0)
+			cell.use(drainage*1000*CELLRATE)
```
afterthought2: Use the battery component instead; if it has a battery component, you can use `use_power_oneoff` here. See other machines, like e.g. portable atmospherics, for how to set it all up.

code/game/machinery/dehumidifier.dm
```diff
@@ -0,0 +1,123 @@
...
+		return
+	else
+		..()
+
+/obj/machinery/dehumidifier/attack_hand(mob/user)
+	interact(user)
+
+/obj/machinery/dehumidifier/interact(mob/user)
+	if(!CanPhysicallyInteract(user))
+		return
```
afterthought2: Use the standard machine API. That means stuff that requires this check goes into `physical_attack_hand`, other stuff goes into `interface_interact`, and `interact` is used only for creating legacy or datum/browser UIs.

code/game/machinery/dehumidifier.dm
```diff
@@ -0,0 +1,123 @@
...
+			return
+	else if(isScrewdriver(I))
+		panel_open = !panel_open
+		user.visible_message(SPAN_NOTICE("[user] [panel_open ? "opens" : "closes"] the hatch on [src]."), SPAN_NOTICE("You [panel_open ? "open" : "close"] the hatch on [src]."))
+		update_icon(TRUE)
+		return
+	else
+		..()
+
+/obj/machinery/dehumidifier/attack_hand(mob/user)
```
afterthought2: This proc generally should not be overridden.

code/game/machinery/dehumidifier.dm
```diff
@@ -0,0 +1,123 @@
...
+				cell = I
+				I.forceMove(src)
+				user.visible_message(SPAN_NOTICE("[user] inserts a power cell into [src]"), SPAN_NOTICE("You insert the power cell into [src]."))
+				power_change()
+				return
+		else
+			to_chat(user, "The hatch must be open to insert a power cell.")
+			return
+	else if(isScrewdriver(I))
+		panel_open = !panel_open
```
afterthought2: Use construct decls to handle all of your construct stuff. That + using the battery component means you can remove 80% of the code in this file.

code/game/machinery/dehumidifier.dm
```diff
@@ -0,0 +1,123 @@
...
+	icon_state = "dhum-[on ? "on" : "off"]"
+	if(rebuild_overlay)
+		overlays.Cut()
+		if(panel_open)
+			overlays.Add("dhum-open")
+
+/obj/machinery/dehumidifier/examine(mob/user)
+	. = ..(user)
+	if (!.)
+		return
```
afterthought2: I believe this is an incorrect use of `examine` parent calls; you should do `. = ..()` (no truncating arguments) and then not check the result.

code/game/machinery/dehumidifier.dm
```diff
@@ -0,0 +1,123 @@
...
+	var/drainage = TRUE
+	var/last_gurgle = 0
+	atom_flags = ATOM_FLAG_CLIMBABLE
+	clicksound = "switch"
+
+/obj/machinery/dehumidifier/Initialize()
+	if(ispath(cell))
+		cell = new cell(src)
+	. = ..()
+	update_icon()
```
afterthought2: Usually preferred to `queue_icon_update` in `Initialize`.

MistakeNot4892:
This was written before the machine refactors, it will need rewriting.

code/game/machinery/dehumidifier.dm
```diff
@@ -0,0 +1,105 @@
...
+			return
+
+		if(!do_after(user, 20, src))
+			return
+
+		playsound(src, 'sound/items/Ratchet.ogg', 100, 1)
+		to_chat(user, SPAN_NOTICE("You [anchored ? "un" : ""]anchor the [src]."))
+		anchored = !anchored
+	else
+		..()
```
afterthought2: Either have a `..()` call or `component_attackby`; having both is redundant.

code/game/machinery/dehumidifier.dm
```diff
@@ -0,0 +1,105 @@
...
+/obj/machinery/dehumidifier/physical_attack_hand(mob/user)
+	if(!panel_open)
+		if(stat & NOPOWER)
+			to_chat(user, SPAN_WARNING("You try to switch [src] on, but nothing happens."))
+			return TRUE
+		set_active(!active)
+		user.visible_message(
+			SPAN_NOTICE("[user] switches [active ? "on" : "off"] the [src]."),
+			SPAN_NOTICE("You switch [active ? "on" : "off"] the [src]."))
+	return TRUE
```
afterthought2: Should probably tab this.

code/game/machinery/dehumidifier.dm
```diff
@@ -0,0 +1,105 @@
...
+
+/obj/machinery/dehumidifier/Process()
+	if(!active)
+		return
+
+	if(!(stat & NOPOWER) || !use_power)
```
afterthought2: Isn't this check backwards?

code/game/objects/items/weapons/circuitboards/machinery/portable_atmospherics.dm
```diff
@@ -49,4 +49,21 @@
...
+	board_type = "machine"
+	build_path = /obj/machinery/dehumidifier
+	origin_tech = "{'" + TECH_ENGINEERING + "':4,'" + TECH_POWER + "':4}"
+	req_components = list(
+		/obj/item/stock_parts/capacitor = 2,
+		/obj/item/stock_parts/matter_bin = 2,
+		/obj/item/pipe = 2)
+	additional_spawn_components = list(
+		/obj/item/stock_parts/console_screen = 1,
+		/obj/item/stock_parts/keyboard = 1,
```
afterthought2, line 65: If you include these, also set `stat_immune = 0` on the machine.

code/game/objects/items/weapons/circuitboards/machinery/portable_atmospherics.dm
```diff
@@ -49,4 +49,21 @@
...
+		/obj/item/stock_parts/power/apc/buildable = 1,
```
afterthought2: Because you set its power channel to `LOCAL` in the machine, I believe it will never use this.

code/game/machinery/dehumidifier.dm
```diff
@@ -0,0 +1,103 @@
...
+
+	var/active = FALSE
+	var/drainage = TRUE
+	var/last_gurgle = 0
+
+/obj/machinery/dehumidifier/Initialize()
+	. = ..()
+	queue_icon_update()
+
+/obj/machinery/dehumidifier/on_update_icon(rebuild_overlay = FALSE)
```
afterthought2: Nothing will ever call this with `TRUE` there, so it will never get the overlay. Just make it add it always; not like this is expensive.

code/game/machinery/dehumidifier.dm
```diff
@@ -0,0 +1,103 @@
...
+		if(panel_open)
+			to_chat(user, "The power cell is [cell ? "installed" : "missing"].")
+		else
+			to_chat(user, "The charge meter reads [cell ? round(cell.percent(), 1) : 0]%")
+
+/obj/machinery/dehumidifier/proc/set_active(new_active)
+	if(active != new_active)
+		active = new_active
+		update_use_power(active ? POWER_USE_ACTIVE : POWER_USE_OFF)
+		update_icon()
```
afterthought2: Maybe `queue` here as well, or possibly omit it. The above proc `update_use_power` also updates icon if a change is made. Also note that I'm not sure you actually _want_ to set use_power here; I don't think it does anything, since you aren't setting `active_power_usage`.

code/game/machinery/dehumidifier.dm
```diff
@@ -0,0 +1,103 @@
...
+		if(is_space_turf(get_turf(src)))
+			to_chat(user, SPAN_WARNING("You can't anchor something to empty space."))
+			return
+
+		if(!do_after(user, 20, src))
+			return
+
+		playsound(src, 'sound/items/Ratchet.ogg', 100, 1)
+		to_chat(user, SPAN_NOTICE("You [anchored ? "un" : ""]anchor the [src]."))
+		anchored = !anchored
```
afterthought2: Doesn't `OBJ_FLAG_ANCHORABLE` just do this, now that I think about it?

afterthought2 (approved):

PR #136, opened 2020-03-14
Fullscreen mode+

quardbreak:
* Adds option "Full" which will hide menu and statusbar in fullscreen mode
* Adds chat input position setting, which can alter input field under output text

image

### DIFF
--- a/code/modules/client/client_procs.dm
+++ b/code/modules/client/client_procs.dm
@@ -216,8 +216,8 @@
 	if(!winexists(src, "asset_cache_browser")) // The client is using a custom skin, tell them.
 		to_chat(src, "<span class='warning'>Unable to access asset cache browser, if you are using a custom skin file, please allow DS to download the updated version, if you are not, then make a bug report. This is not a critical issue but can cause issues with resource downloading, as it is impossible to know when extra resources arrived to you.</span>")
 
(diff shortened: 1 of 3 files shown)


MistakeNot4892 (approved):
Willing to give this a shot. Fuck me if that skin code isn't ugly as Hell though.

deathride58 (approved):
[Not hiding the verb menu? You are tiny grasshopper.](https://i.redd.it/s6j7q7eefw941.png) (/s)

Jokes aside though, the code looks alright, even with it being a little messy. However, you might actually be able to make use of splitters here to make this a lot more elegant. You can do this by turning the command bar into it's own control, adding a horizontal splitter layer to both the main control and the chat control, then have the code side of things simply give the command bar control to the appropriate bottom splitter (and of course, scale the splitter so that the command bar retains the proper size). This would remove the need for having completely separate controls for both layouts, which would reduce maintenance overhead for things that require touching interface controls directly (for example, if we wanted to make the command bar able to show typing indicators, or wanted to bring back the classic "GLORF" when you're hit with a strong weapon while typing)

I'll be giving this 24 hours from this comment being posted before I merge it, in case any other maintainers wants to make my idea detailed above a requested change, or in case you want to rework it yourself regardless.

PR #139, opened 2020-03-14
Vent hum is now sound played by vents

comma:
Removed it from area ambience
Made turfs raise opacity change event when their opacity counter is updated (otherwise sound players never reacted to doors opening)
Implements https://github.com/NebulaSS13/NebulaFeatureRequests/issues/17


code/modules/atmospherics/components/unary/vent_pump.dm
```diff
@@ -554,6 +557,19 @@
...
+/obj/machinery/atmospherics/unary/vent_pump
+	
+/obj/machinery/atmospherics/unary/vent_pump/power_change()
+	. = ..()
+	if(.)
+		update_sound()
+
+/obj/machinery/atmospherics/unary/vent_pump/proc/update_sound()
+	if(can_pump())
+		sound_token = GLOB.sound_player.PlayLoopingSound(src, "[type]_[x]_[y]_[z]", 'sound/machines/vent_hum.ogg', 15, range = 14, falloff = 4)
```
PsiOmegaDelta: Due to BYOND-limitations we can only have 1024 unique sound-id sources and vents would take up a big part of those
Generally objects playing the same sound should all utilize the same sound id

comma: Changed it to only enforce uniqueness inside zlevel. As far as people can hear, they're all unique.

code/modules/atmospherics/components/unary/vent_pump.dm
```diff
@@ -554,6 +558,21 @@
 	. = ..()
 	air_contents.volume = ATMOS_DEFAULT_VOLUME_PUMP + 500 //meant to match air injector
 
+/obj/machinery/atmospherics/unary/vent_pump
```
MistakeNot4892: Random path?

comma: fixed

afterthought2 (approved):
Not wild about the implementation, but don't have a reasonable alternative suggestion. The turf opacity thing is also a bit screwy; may be worth adding a comment about it in the opacity set file.

PR #140, opened 2020-03-14
Radial menu, Tooltips

quardbreak:
General changes:
-
* Ports the radial menu from tgstation. Can be used for short visible actions.
  * Currently used for robot icon pickup only.
* Ports tooltip system from tgstation. 
  * Currently used in radial menu only.

image


code/_helpers/icons.dm
```diff
@@ -927,3 +927,10 @@ proc/generate_image(var/tx as num, var/ty as num, var/tz as num, var/range as nu
 
 	return cap
 
+/atom/proc/add_overlay(var/overlay)
```
Lohikar: Why though? Just use native overlay manipulation.

quardbreak: Author posted something about this in his PR description:
> Lets add_overlay use images instead of just icon_states, to let add_overlay work with icon files that aren't the src object's icon.

code/_helpers/lists.dm
```diff
@@ -178,6 +178,12 @@ Checks if a list has the same entries and values as an element of big.
 		return picked
 	return null
 
+//Returns the first element from the list and removes it from the list
+/proc/popleft(list/L)
+	if(L.len)
```
MistakeNot4892: `if(length(L))` would make this proc safe for null args.

MistakeNot4892 (approved):
I didn't read this in great detail, but nothing leapt out as badwrong, so this is a tentative approval.

PR #142, opened 2020-03-15
Holopad update

quardbreak:
* Adds the holoray overlay which tracks with a hologram
* Holopad's holograms no longer restricted only on station records
* Shifts up hologram overlay a bit
* Adds lights to holopad, hologram and holoray

Sorta tested, need to kill one runtime which I get at icon generation I guess
Runtime log: https://hastebin.com/nuqiverihe.js


code/game/machinery/hologram.dm
```diff
@@ -286,9 +288,24 @@ For the other part of the code, check silicon say.dm. Particularly robot talk.*/
...
-	hologram.set_light(1, 0.1, 2)	//hologram lighting
-	hologram.color = color //painted holopad gives coloured holograms
-	set_light(1, 0.1, 2)			//pad lighting
+
+	//painted holopad gives coloured holograms
+	var/new_color = holopadType == HOLOPAD_LONG_RANGE ? "#e1df7d" : "#7db4e1"
+
+	//pad lighting
+	light_color = new_color
+	set_light(1, 0.1, 2)
```
afterthought2, line 297: I believe you can supply color to the proc, which should be preferred.

code/game/objects/effects/overlays.dm
```diff
@@ -53,3 +53,14 @@
 	. = ..()
 	pixel_x += rand(-10, 10)
 	pixel_y += rand(-10, 10)
+
+/obj/effect/overlay/holoray
```
afterthought2, line 57: Is there a reason you can't do this with an image overlay?

MistakeNot4892:
Going to go ahead and close this as your situation was discussed on Discord, please feel free to reopen if you end up deciding to keep going with it.

PR #143, opened 2020-03-15
Misc floor changes & Sticky paper fix

quardbreak:
* Adds footstep sounds for dirt, asteroid and snow flooring decls
* Fixes sticky paper placement. Should be above window layer now
* Prevents crowbar actions on flooring with harm intent
* Small skill integration: 
 Makes flooring removal with crowbar use a construction skill as delay (higher skill = less time). Unskilled acts already pretty quickly.

### DIFF
--- a/code/game/turfs/flooring/flooring.dm
+++ b/code/game/turfs/flooring/flooring.dm
@@ -72,6 +72,7 @@
 	has_base_range = 3
 	damage_temperature = T0C+80
 	can_engrave = FALSE
+	footstep_type = /decl/footsteps/grass
 
 /decl/flooring/asteroid
 	name = "coarse sand"
@@ -81,6 +82,7 @@
 	flags = TURF_HAS_EDGES | TURF_REMOVE_SHOVEL
 	build_type = null
 	can_engrave = FALSE
+	footstep_type = /decl/footsteps/asteroid
 
 /decl/flooring/carpet
 	name = "brown carpet"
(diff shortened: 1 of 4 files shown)


code/game/turfs/flooring/flooring.dm
```diff
@@ -356,3 +358,14 @@
 	build_type = null
 	flags = TURF_ACID_IMMUNE | TURF_CAN_BREAK | TURF_REMOVE_CROWBAR
 	color = "#00ffe1"
+
+/decl/flooring/snow
+	name = "snow"
+	desc = "Let is sno-ow... Let is snow..."
```
Lohikar: Spelling.

code/modules/paperwork/paper_sticky.dm
```diff
@@ -78,6 +78,8 @@
 	icon = 'icons/obj/stickynotes.dmi'
 	color = COLOR_YELLOW
 	slot_flags = 0
+	layer = ABOVE_WINDOW_LAYER
+	pixel_z = 0
```
Lohikar, line 82: Why are you zeroing this out here? Isn't it already 0?

quardbreak: As I understand it, this is necessary so that the icons do not go beyond the boundaries of a wall. This fix not mine, I just ported it.

code/game/turfs/flooring/flooring.dm
MistakeNot4892: 'Iet it snow, let it snow, let it snow' is the lyric in English.

code/modules/paperwork/paper_sticky.dm
MistakeNot4892: This won't do anything, `pixel_z` is already 0 on this object definition. It might do something you you put it in logic or something, but not here.

MistakeNot4892:
Happy to approve this if the `pixel_z` setting is moved/removed.

PR #145, opened 2020-03-15
Switching worth system to an automatic value generation proc.

MistakeNot4892:
DNM, WIP

Basically just making items generate their own worth when examined based on the actual use and purpose of the item.

Kills `worths_list.dm`.


MistakeNot4892:
Numbers seem to be closing in on more or less reasonable for now (except for guns and weapons). Will work out some proc to dump the entire worth list to CSV or something so we can discuss it another day.

code/game/machinery/_machines_base/machinery.dm
```diff
@@ -457,4 +457,9 @@ Class Procs:
 		var/obj/item/stock_parts/part = get_component_of_type(path)
 		var/list/part_costs = part.building_cost()
 		for(var/key in part_costs)
-			.[key] += part_costs[key] * component_types[path]
\ No newline at end of file
+			.[key] += part_costs[key] * component_types[path]
+
+/obj/machinery/get_monetary_worth()
```
comma: maybe count value of all components?

afterthought2:
Arguably it can even iterate contents and add in the worth of everything there. Not entirely sure that's reasonable, though.

MistakeNot4892:
It already does that.

afterthought2 (approved):
Read through it; looks reasonable. Machinery is already counting components.

PR #150, opened 2020-03-16
Updates vote panel interface

quardbreak:
Ports Infinity Station vote panel changes.
Allow to turn off percent vote, highlight leading choose and etc..

Multiple|Single
-|-
image|image


code/controllers/subsystems/vote.dm
```diff
@@ -117,7 +117,9 @@ SUBSYSTEM_DEF(vote)
 	if(active_vote)
 		win_x = active_vote.win_x
 		win_y = active_vote.win_y
-	show_browser(user, interface(user.client),"window=vote;size=[win_x]x[win_y]")
+	var/datum/browser/popup = new(user, "vote", "Voting Panel", win_x, win_y)
+	popup.set_content(interface(user.client))
+	popup.open(use_onclose = TRUE)
 	onclose(user, "vote", src)
 
 /datum/controller/subsystem/vote/proc/close_panel(mob/user)
```
afterthought2, line 124: On the line below this, you should delete the browser instead.

code/controllers/subsystems/vote.dm
```diff
@@ -117,7 +117,9 @@ SUBSYSTEM_DEF(vote)
 	if(active_vote)
 		win_x = active_vote.win_x
 		win_y = active_vote.win_y
-	show_browser(user, interface(user.client),"window=vote;size=[win_x]x[win_y]")
+	var/datum/browser/popup = new(user, "vote", "Voting Panel", win_x, win_y)
+	popup.set_content(interface(user.client))
+	popup.open(use_onclose = TRUE)
```
afterthought2: As this autoupdates, it's possible you should store the browser datums and update their html instead of recreating them.

code/datums/vote/vote.dm
```diff
@@ -47,13 +51,14 @@
 
 /datum/vote/proc/start_vote()
 	start_time = world.time
+	time_set = (time_set ? time_set : config.vote_period) SECONDS
```
afterthought2: Confusing; this means that at compile time this var is in seconds, but then is multiplied by 10 at runtime?

code/datums/vote/vote.dm
```diff
@@ -11,17 +11,21 @@
 
 	var/start_time
 	var/time_remaining
+	var/time_set
 	var/status = VOTE_STATUS_PREVOTE
 
 	var/list/result                // The results; format is list(choice = votes).
-	var/results_length = 3         // How many choices to show in the result. Setting to -1 will show all choices.
+	var/results_length = 3         // How many choices to show in the result. Setting to INFINITY will show all choices.
```
afterthought2, line 18: Is this an actual change? I don't see the relevant code anywhere.

quardbreak: This is info update because -1 doesn't work iirc

code/controllers/subsystems/vote.dm
MistakeNot4892: I don't think it really matters in this case. The browser GCs on its own after the HTML is sent to the client, as far as I am aware.

afterthought2 (approved):

code/datums/vote/vote.dm
```diff
@@ -153,7 +158,9 @@
 	else
 		. += "<h2>Vote: [capitalize(name)]</h2>"
 	. += "Time Left: [time_remaining] s<hr>"
-	. += "<table width = '100%'><tr><td align = 'center'><b>Choices</b></td><td colspan='3' align = 'center'><b>Votex</b></td><td align = 'center'><b>Votes</b></td>"
+	. += "<div class='statusDisplay'>"
+	var/votes_count =  show_votes_count || check_rights(R_INVESTIGATE, 0, user) ? "<td align = 'center'><b>Votes</b></td>" : null
+	. += "<table width = '100%'><tr><td align = 'center'><b>Choices</b></td><td colspan='[priorities.len]' align = 'center'><b>Votex</b></td>[votes_count]"
```
comma: `Votex`?

quardbreak: Haha It was before my changes!

PR #158, opened 2020-03-16
Nabbed some GC return value fixes from BoH.

MistakeNot4892:
Yoinked from https://github.com/BoHBranch/BoH-Bay/pull/530.

Also moved some logic to `/turf/simulated/Destroy()` from `ChangeTurf()`.

### DIFF
--- a/code/controllers/subsystem.dm
+++ b/code/controllers/subsystem.dm
@@ -78,7 +78,8 @@
 	can_fire = 0
 	flags |= SS_NO_FIRE
 	Master.subsystems -= src
-	return ..()
+	..()
+	return QDEL_HINT_HARDDEL_NOW
 
 //Queue it to run.
 //	(we loop thru a linked list until we get to the end or find the right point)
--- a/code/game/objects/items/weapons/tanks/jetpack.dm
+++ b/code/game/objects/items/weapons/tanks/jetpack.dm
@@ -20,8 +20,8 @@
 	ion_trail.set_up(src)
 
 /obj/item/tank/jetpack/Destroy()
-	qdel(ion_trail)
-	..()
+	QDEL_NULL(ion_trail)
+	. = ..()
 
 /obj/item/tank/jetpack/examine(mob/living/user)
 	. = ..()
--- a/code/game/turfs/simulated.dm
+++ b/code/game/turfs/simulated.dm
@@ -165,3 +165,12 @@
 	if(GAME_STATE >= RUNLEVEL_GAME)
 		fluid_update()
 	. = ..()
+
+/turf/simulated/Destroy()
+	if (zone)
+		if (can_safely_remove_from_zone())
+			c_copy_air()
+			zone.remove(src)
+		else
+			zone.rebuild()
+	. = ..() 
--- a/code/modules/power/breaker_box.dm
+++ b/code/modules/power/breaker_box.dm
@@ -19,11 +19,6 @@
 	var/RCon_tag = "NO_TAG"
 	var/update_locked = 0
 
-/obj/machinery/power/breakerbox/Destroy()
-	..()
-	for(var/datum/nano_module/rcon/R in world)
-		R.FindDevices()
-
 /obj/machinery/power/breakerbox/activated
 	icon_state = "bbox_on"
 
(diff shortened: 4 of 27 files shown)


code/controllers/subsystem.dm
```diff
@@ -78,7 +78,7 @@
 	can_fire = 0
 	flags |= SS_NO_FIRE
 	Master.subsystems -= src
-	return ..()
+	. = ..()
```
afterthought2: ??

code/game/objects/items/weapons/tanks/jetpack.dm
```diff
@@ -21,7 +21,7 @@
 
 /obj/item/tank/jetpack/Destroy()
 	qdel(ion_trail)
```
afterthought2: `QDEL_NULL`

code/modules/power/breaker_box.dm
```diff
@@ -20,9 +20,9 @@
 	var/update_locked = 0
 
 /obj/machinery/power/breakerbox/Destroy()
-	..()
 	for(var/datum/nano_module/rcon/R in world)
 		R.FindDevices()
```
afterthought2: Please just kill this. It autoupdates and rebuilds the list on autoupdate anyway.

afterthought2 (commented):
Thought we added a unit test for this?

code/controllers/subsystem.dm
MistakeNot4892: ???
It's functionally the same, just more consistent.

afterthought2: I wasn't sure if you were making a stylistic change or if you wanted them to return a different GC hint but didn't include the full change. I'm not sure what GC hint is appropriate here.

code/modules/power/breaker_box.dm
MistakeNot4892: Any reason not to kill the SMES one as well?

afterthought2: Kill that too, I feel.

MistakeNot4892:
I don't recall a unit test, but atoms not returning a proper GC hint should cause a stack trace at runtime. I guess maybe none of these items are mapped, or have been blown up recently.

code/controllers/subsystem.dm
MistakeNot4892: The PR I am stealing from returns `QDEL_HARRDEL_NOW`. Was going to ask @Lohikar but Discord is crapping out at the moment.

afterthought2:
I was referring to #51

afterthought2 (approved):

PsiOmegaDelta:
#51 just enforces that `..()` is called, not returned.
The linter can enforce return values but, AFAIK, only that they adhere to a specific type (which would exclude simple numbers)

code/controllers/subsystem.dm
```diff
@@ -78,7 +78,7 @@
 	can_fire = 0
 	flags |= SS_NO_FIRE
 	Master.subsystems -= src
-	return ..()
+	. = QDEL_HINT_HARDDEL_NOW
```
Lohikar:
```suggestion
	..()
	return QDEL_HINT_HARDDEL_NOW
```

Lohikar (changes requested):

afterthought2:
May be worth changing the GC hint defines so as to function with the linter's typecheck, then, as the actual define values are arbitrary. They have to be datum instances?

PsiOmegaDelta:
Looks like it's possible to enforce specific paths, and not having to resort to instances.
In either case, should be easy to map the existing macros to either.

MistakeNot4892:
I am fine with this PR as-is, changing how GC works is out of scope for what was meant to be a quick fix PR.

code/game/turfs/simulated.dm
```diff
@@ -165,3 +165,12 @@
 	if(GAME_STATE >= RUNLEVEL_GAME)
 		fluid_update()
 	. = ..()
+
+/turf/simulated/Destroy()
```
Lohikar, line 169: This isn't a GC return value fix. Document if intentional, pls.

MistakeNot4892: ok, it's in the PR message now.

Lohikar (approved):

PR #159, opened 2020-03-16
Enforces no sleeping in process procs.

afterthought2:
Needs newer build of SpacemanDMM, or building from source.

A bit of an experiment. Don't think it's correct to do it at subsystem/fire level.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ env:
 matrix:
     include:
     - name: "DreamChecker"
-      env: SPACEMAN_DMM_VERSION=suite-1.2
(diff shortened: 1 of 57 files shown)


afterthought2:
I'll reopen once SpacemanDMM updates to the next version. Could build from source in Travis, but I think it's not worth it and bad practice.

MistakeNot4892:
Seems like https://github.com/Baystation12/Baystation12/pull/28271 is needed?

afterthought2:
This is ready. If you hate the simple animal thing, can discuss alternatives.

There are a couple of minor functional changes (autopsy scanner).

comma (approved):

PR #160, opened 2020-03-17
Forced 48x48 on client join, removed stretch to fit.

MistakeNot4892:


afterthought2 (approved):
Don't think you need to remove stretch necessarily, but OK.

PR #167, opened 2020-03-17
Makes PDA off from start

quardbreak:
Adds variable for enabling modular computers at spawn


MistakeNot4892 (approved):

MistakeNot4892 (approved):

afterthought2:
I don't get it; this isn't a functional change, right? Is this downstream support?

quardbreak:
This doesn't change anything in game but moves call to parent with new check so we can make consoles and laptops turned on from start

PR #170, opened 2020-03-18
Apcs to construct states.

afterthought2:
Migration procedure: replace `/obj/item/module/power_control` with `/obj/item/stock_parts/circuitboard/apc`. If maps still have removed vars set, remove them (you can do more sophisticated stuff, but it's unlikely to be worth it).

### DIFF
--- a/code/modules/power/apc.dm
+++ b/code/modules/power/apc.dm
@@ -46,31 +46,34 @@
 	lighting = 0
 	equipment = 0
 	environ = 0
-	locked = 0
-	coverlocked = 0
+	locked = FALSE
 
 /obj/machinery/power/apc/critical
 	is_critical = 1
 
 /obj/machinery/power/apc/high
-	cell_type = /obj/item/cell/high
+	uncreated_component_parts = list(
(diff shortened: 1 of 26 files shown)


code/modules/power/apc.dm
```diff
@@ -420,250 +392,69 @@
...
-				if(50 to 70)
-					to_chat(user, "<span class='notice'>You manage to bash the lid open!</span>")
-					opened = 1
-				if(90 to 100)
-					to_chat(user, "<span class='warning'>There's a nasty sound and \the [src] goes cold...</span>")
-					set_broken(TRUE)
-			queue_icon_update()
-	playsound(get_turf(src), 'sound/weapons/smash.ogg', 75, 1)
+		var/roulette = rand(1,100)
+		switch(roulette)
```
comma: maybe only leave cover here?
breaking lock / something vital to break machine should probably be just from natural damage procs so all damaging is going through same pathways.

MistakeNot4892 (dismissed):

comma (approved):

PR #174, opened 2020-03-18
Implements scaling glide_size, mostly for mobs and grabbed atoms.

MistakeNot4892:
Smooths movement out based on expected move delays.

### DIFF
--- a/code/game/atoms_movable.dm
+++ b/code/game/atoms_movable.dm
@@ -131,6 +131,7 @@
 
 /atom/movable/proc/throw_at(atom/target, range, speed, mob/thrower, spin = TRUE, datum/callback/callback) //If this returns FALSE then callback will not be called.
 	. = TRUE
+	glide_size = world.icon_size / max((speed - GLIDE_SIZE_CONSTANT), world.tick_lag) * world.tick_lag
 	if (!target || speed <= 0 || QDELETED(src) || (target.z != src.z))
 		return FALSE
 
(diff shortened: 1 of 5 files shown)


code/game/atoms_movable.dm
```diff
@@ -131,6 +131,7 @@
 
 /atom/movable/proc/throw_at(atom/target, range, speed, mob/thrower, spin = TRUE, datum/callback/callback) //If this returns FALSE then callback will not be called.
 	. = TRUE
+	glide_size = world.icon_size / max((speed - GLIDE_SIZE_CONSTANT), world.tick_lag) * world.tick_lag
```
afterthought2, line 134: Possibly in the non-movement handler cases, one should cache the old glide size, do the operation, and then reset it? Not sure about mobs exactly, but for objs there are many things which call step. And at least some mobs, like simple animals, use walk and such.

MistakeNot4892: There's no clean point to reset glide_size, which is the issue.

afterthought2: Well, in this particular case, you can cache the value on the throw datum and reset when the throw is fully resolved in the SS. But in general, yes, that's a valid point.

Lohikar:
I don't really know enough about how bay movement works to review this, probably best left for Psi.

MistakeNot4892:
I'm going to revisit this when I get through my other PRs, will park it till then.

PR #178, opened 2020-03-19
Fixes runtime with card slot qdel in nullspace.

MistakeNot4892:


code/modules/modular_computers/hardware/card_slot.dm
```diff
@@ -116,7 +116,8 @@
 	usage_flags = PROGRAM_PDA
 
 /obj/item/stock_parts/computer/card_slot/Destroy()
-	loc.verbs -= /obj/item/stock_parts/computer/card_slot/proc/verb_eject_id
+	if(loc)
+		loc.verbs -= /obj/item/stock_parts/computer/card_slot/proc/verb_eject_id
```
PsiOmegaDelta, line 120: Have to say, this makes me a wonder if there's a possibility for verbs to remain added to objects if the card slot is ever removed via non-deconstruction means

PsiOmegaDelta (approved):

code/modules/modular_computers/hardware/card_slot.dm
afterthought2: Sort of unclear; this part is PDA-only, I believe, and not really removable. Could use closer examination, though.

PR #179, opened 2020-03-19
Removing sort_string and id from research datums.

MistakeNot4892:
Breaking changes from #116 out into a separate PR.

- Removed `sort_string` and made recipe sorting use `name`.
- Removed `id` and made the R&D interfaces use a reference to the datum instead.

### DIFF
--- a/code/modules/research/research.dm
+++ b/code/modules/research/research.dm
@@ -91,16 +91,8 @@ research holder datum.
 	return
 
 /datum/research/proc/AddDesign2Known(var/datum/design/D)
-	if(!length(known_designs))
-		known_designs.Add(D)
-		return
-	for(var/i = 1 to known_designs.len)
-		var/datum/design/A = known_designs[i]
-		if(A.name != D.name && "[A.name]" > "[D.name]")
-			known_designs.Insert(i, D)
-			return
-	known_designs.Add(D)
-	return
+	if(!(D in known_designs))
+		ADD_SORTED(known_designs, D, /proc/cmp_name_asc)
 
 //Refreshes known_tech and known_designs list
 //Input/Output: n/a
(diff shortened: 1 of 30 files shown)


MistakeNot4892:
Ready to go.

code/modules/research/research.dm
```diff
@@ -91,14 +91,14 @@ research holder datum.
...
 
 /datum/research/proc/AddDesign2Known(var/datum/design/D)
-	if(!known_designs.len) // Special case
+	if(!length(known_designs))
 		known_designs.Add(D)
 		return
 	for(var/i = 1 to known_designs.len)
 		var/datum/design/A = known_designs[i]
-		if(A.id == D.id) // We are guaranteed to reach this if the ids are the same, because sort_string will also be the same
+		if(A.name == D.name)
```
PsiOmegaDelta: Wouldn't `A.type == D.type` work here?

code/modules/research/research.dm
```diff
@@ -91,14 +91,14 @@ research holder datum.
...
 			return
-		if(A.sort_string > D.sort_string)
+		if("[A.name]" > "[D.name]")
```
PsiOmegaDelta: Doesn't `if(A.name > D.name)` work?
Seems a little odd to create a new string out of what should already be a string

PsiOmegaDelta: Feels like this should have the same functionality
```
/datum/research/proc/AddDesign2Known(var/datum/design/D)
    if(D.name in known_designs)
        return
    ADD_SORTED(known_designs, D.name, /proc/cmp_text_asc)
    known_designs[D.name] = D
```

MistakeNot4892: It runtimes on a null - I actually had name initialize and forgot to reverse this.

PsiOmegaDelta (approved):

PR #186, opened 2020-03-19
Prevent overloading /New()

PsiOmegaDelta:
Unless it's a straight /datum/ subtype

Awaiting linter errors


PsiOmegaDelta:
Probably cannot use the no-override-setting in general like this because some of the `/New()` overrides are necessary

PR #189, opened 2020-03-20
Z-Sound - Adds the ability for sound to propagate across multiple connected z-levels

deathride58:
This more or less does exactly what the title suggests it does on the tin. [A short demo showing this in action can be found here.](https://streamable.com/etrrr)

This effect can be tweaked via the ZSOUND_DRYLOSS_PER_Z and ZSOUND_DISTANCE_PER_Z defines in `code/__defines/sound.dm`. The former affects how muffled a sound gets the further the distance between z-levels, and the latter affects how much physical distance gets added per z-level. The defaults are tweaked to make quieter sounds nearly inaudible, while still allowing beefier sounds (such as gunshots and melee attacks) to reverberate throughout the ship/station/etc.

Sound behavior can also be modified via new args on playsound and playsound_local. For the former, the args are zrange (affects how many z-levels up and down it can travel, default is 2), override_env (disables zsound's effects for better manual control over sounds), envdry (affects the dry channel (ECHO_DIRECT) of the environmental effects, this is what zsound adjusts to give that muffled effect), envwet (affects the wet channel (ECHO_ROOM) of the environmental effects). For the latter, the args added are override_env, envdry, and envwet.

This also adds new defines for all of the variables modifiable via the echo list. I have done some testing, and it seems the only two args that do anything are ECHO_DIRECT and ECHO_ROOM, though I have included defines for all echo list index positions for the sake of completeness.

Potential future improvements include disabling or reducing the muffle effect if either the source or target of a given sound can see each other from an openspace, though the implementation has been kept relatively simple for now.

### DIFF
--- a/code/_global_vars/sound.dm
+++ b/code/_global_vars/sound.dm
@@ -110,4 +110,4 @@ GLOBAL_LIST_INIT(glasscrack_sound,\
 		'sound/effects/glass_crack4.ogg'))
 
 //This is a global list in interest of optimization; list init is expensive enough for this to be worth doing, even with playsound() not being very hot.
-GLOBAL_LIST_INIT(echo_list, list(null, null, null, null, null, null, null, null, null, null, null, null, null, null, null, null, null, null))
+GLOBAL_LIST_INIT(echo_list, new(18))
--- a/code/_global_vars/sound.dm
+++ b/code/_global_vars/sound.dm
@@ -108,6 +108,3 @@ GLOBAL_LIST_INIT(glasscrack_sound,\
 		'sound/effects/glass_crack2.ogg',\
 		'sound/effects/glass_crack3.ogg',\
 		'sound/effects/glass_crack4.ogg'))
-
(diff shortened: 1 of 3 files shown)


code/_global_vars/sound.dm
```diff
@@ -108,3 +108,6 @@ GLOBAL_LIST_INIT(glasscrack_sound,\
 		'sound/effects/glass_crack2.ogg',\
 		'sound/effects/glass_crack3.ogg',\
 		'sound/effects/glass_crack4.ogg'))
+
+//This is a global list in interest of optimization; list init is expensive enough for this to be worth doing, even with playsound() not being very hot.
+GLOBAL_LIST_INIT(echo_list, new(18))
```
Lohikar: Remove this. `playsound_local()` is not remotely hot enough to require optimizations like this.

Lohikar (dismissed):

comma (approved):

PR #193, opened 2020-03-20
Enforces minimum matter amounts for obj.

MistakeNot4892:
More split-off stuff from the recipes PR.


MistakeNot4892:
Unit test isn't working.

code/game/machinery/_machines_base/machinery.dm
```diff
@@ -115,6 +115,11 @@ Class Procs:
 	var/processing_flags         // What is being processed
 
 /obj/machinery/Initialize(mapload, d=0, populate_parts = TRUE)
+
+	// TODO - temporary - convert machinery to use the material var.
+	if(!matter)
+		LAZYSET(matter, MAT_STEEL, max(matter[MAT_STEEL], ceil(w_class * INITIAL_MATTER_W_CLASS_MULT * SHEET_MATERIAL_AMOUNT)))
```
afterthought2: I believe that the `matter[MAT_STEEL]` there will always be `0`, given how it's gated behind `!matter`.

code/game/machinery/doors/simple.dm
```diff
@@ -18,12 +18,17 @@
...
 /obj/machinery/door/unpowered/simple/Initialize(mapload, var/material_name, var/locked)
-	. = ..(mapload)
 	if(!material_name)
 		material_name = MAT_STEEL
 	material = SSmaterials.get_material_datum(material_name)
 	if(!material)
 		return INITIALIZE_HINT_QDEL
+
+	if(!matter)
+		LAZYSET(matter, material.type, max(matter[material.type], ceil(w_class * INITIAL_MATTER_W_CLASS_MULT * SHEET_MATERIAL_AMOUNT)))
```
afterthought2: Similarly here.

code/unit_tests/material_tests.dm
```diff
@@ -0,0 +1,32 @@
...
+
+/datum/unit_test/crafting_recipes_shall_not_have_inconsistent_materials/start_test()
+	var/list/failed_designs = list()
+	var/list/passed_designs = list()
+	for(var/owner_mat in SSmaterials.materials_by_name)
+		var/material/mat_datum = SSmaterials.get_material_datum(owner_mat)
+		for(var/datum/stack_recipe/recipe in mat_datum.recipes)
+			var/obj/product = recipe.spawn_result()
+			var/failed
+			if(product.matter && (recipe.use_material || recipe.use_reinf_material))
```
afterthought2: Is the `&&` there correct? If the recipe uses neither material nor reinf_material, we presumably want to have the result fail the test if it has anything in `matter` at all.

code/unit_tests/material_tests.dm
```diff
@@ -0,0 +1,32 @@
...
+				if(recipe.use_material && product.matter[recipe.use_material] >= recipe.req_amount)
```
afterthought2: Isn't matter in units like 2000 or something but `req_amount` is in sheets?

MistakeNot4892: Yeah that's meant to have a `/SHEET_MATERIAL_AMOUNT` on it.

code/game/machinery/doors/simple.dm
```diff
@@ -18,12 +18,17 @@
...
 	if(!material_name)
 		material_name = MAT_STEEL
 	material = SSmaterials.get_material_datum(material_name)
 	if(!material)
 		return INITIALIZE_HINT_QDEL
+
+	if(!matter)
+		LAZYSET(matter, material.type, ceil(w_class * INITIAL_MATTER_W_CLASS_MULT * SHEET_MATERIAL_AMOUNT))
+
+	. = ..(mapload)
```
PsiOmegaDelta: Must always call parent even if you're returning a different initialize hint

PsiOmegaDelta: That or manually set the initialize flag

MistakeNot4892: It is calling parent.

PsiOmegaDelta: https://github.com/NebulaSS13/Nebula/pull/193/files#diff-5ad140ad80ac034ecfb5d029a22d973bR25 returns early

code/unit_tests/material_tests.dm
```diff
@@ -0,0 +1,40 @@
...
+				else
+					for(var/mat in product.matter)
+						if(mat != recipe.use_material && mat != recipe.use_reinf_material)
+							failed = "extra material type ([mat])"
+			if(failed)
+				failed_designs += "[owner_mat] - [recipe.type] - [failed]"
+			else
+				passed_designs += recipe
+			if(!QDELETED(product))
+				qdel(product)
```
afterthought2, line 34: Maybe just don't qdel it? Who cares?

MistakeNot4892: Exposing and then ignoring an error is pretty bad practice.

MistakeNot4892:
This is going to depend on standardizing Initialize() for /obj so will return to it after that.

PR #197, opened 2020-03-21
Makes TP portals qdel on forceMove.

afterthought2:
Mostly to avoid some shuttle exploits.


PsiOmegaDelta (dismissed):
Also includes if somehow moved non-forced

PsiOmegaDelta (approved):

PR #212, opened 2020-03-27
Moving cult/wiz/deity antagonists into a content package.

MistakeNot4892:
TODO:
- Return the various creatures to xenoarch.
- Work out some day to add the cult stuff to traders at runtime, or move traders into the package.
- Return cult runes to surgery checks.
- General cleanup of excised code.


MistakeNot4892:
This is going to need its own thing, will revisit.

PR #216, opened 2020-03-28
Xenoarch machinery/tools cleanup, tweaks, resprites

comma:
Changed scan time to be ticking instead of using world.time. Cut the scan duration so it's more or less same length as it was.
Made interface NanoUI, since old one was gross and needed refresh for scan progress etc.
Broke up the gross ass hardcoded scan results proc, moved em into obj-level proc. Kinda snowflake but I think it's ok.
Anomaly scanner no longer anchors anomalies at the start of scan. Was super prone to bugging out and leaving anomalies stuck there forever whenever something went wrong.
Added new icon for it (definetly not a recolored protolathe)
image

Gave similar treatment to anomaly harvesting machines and anomaly power utilizer devices.
Made excavation picks actually differ in size (2px diff from neighbor)


MistakeNot4892 (dismissed):

MistakeNot4892 (dismissed):

MistakeNot4892 (approved):
I have given this a look over but honestly I am not familiar with xenoarch. Nothing leaps out as wrong or bad and the worst-case scenario is that xenoarch breaks, which isn't too serious.

PR #219, opened 2020-03-28
Anomaly effects cleanup

comma:
Refactors the shit out of this mess.
First of all, max range for anomalies is now 40 tiles rather than 200 because fuck range(200).
All instances of effects looking for shit in 200 tiles radius were replaced with their effect range value.
Gas generation effects generalized. Buffed aura slightly because 0.1 mole every few ticks is fucking nothing.
Temp manipulation effects generalized. Now using fancy gas mix procs instead of directly changing temperature var. Should preserve some sanity like not breaking absolute zero.
Other effects mostly un-copypasted by moving similar shit into separate procs.
Some huge string lists were made global for type.


code/modules/xenoarcheaology/effects/celldrain.dm
```diff
@@ -8,41 +8,31 @@
 	if(user)
 		if(istype(user, /mob/living/silicon/robot))
 			var/mob/living/silicon/robot/R = user
-			for (var/obj/item/cell/D in R.contents)
-				D.charge = max(D.charge - rand() * 100, 0)
-				to_chat(R, "<span class='warning'>SYSTEM ALERT: Energy drain detected!</span>")
+			var/obj/item/cell/C = R.get_cell()
+			C.use(100)
```
MistakeNot4892: `get_cell()` can return null.

MistakeNot4892: As with all the other get_cell calls.

comma: Fixed

MistakeNot4892 (approved):
Jesus Christ that is a lot of awful old code.

PR #220, opened 2020-03-28
[WIP] Machinery explosion tweaks

comma:
Now machinery will take more damage to components on higher explosion seveity, and has a chance to dismantle.
Removed a bunch of ex_act overrides in buildable machines, they should be using base one. Adjusted ones with special effects to call parent after they're done.

### DIFF
--- a/code/game/machinery/vending/_vending.dm
+++ b/code/game/machinery/vending/_vending.dm
@@ -124,22 +124,9 @@
 	
 
 /obj/machinery/vending/ex_act(severity)
-	switch(severity)
-		if(1.0)
-			qdel(src)
-			return
-		if(2.0)
-			if (prob(50))
-				qdel(src)
-				return
-		if(3.0)
-			if (prob(25))
-				spawn(0)
-					src.malfunction()
-					return
-				return
-		else
-	return
+	if(severity == 3 && prob(25))
+		malfunction()
+	..()
 
 /obj/machinery/vending/emag_act(var/remaining_charges, var/mob/user)
 	if (!emagged)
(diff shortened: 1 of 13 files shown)


code/modules/power/lighting.dm
```diff
@@ -395,15 +395,13 @@
...
 	switch(severity)
-		if(1)
-			qdel(src)
-			return
 		if(2)
 			if (prob(75))
 				broken()
 		if(3)
 			if (prob(50))
 				broken()
```
afterthought2, line 403: Doesn't really make sense: at 2 it breaks, but at 1 nothing happens?

code/game/machinery/vending/_vending.dm
```diff
@@ -124,22 +124,9 @@
...
-		if(3.0)
-			if (prob(25))
-				spawn(0)
-					src.malfunction()
-					return
-				return
-		else
-	return
+	if(severity == 3 && prob(25))
+		malfunction()
```
afterthought2, line 128: also here

code/modules/power/lighting.dm
comma: They call parent at the end, broken() is lightswitch-specific thing to break bulb

code/game/machinery/vending/_vending.dm
comma: Ditto, it does malfunction check then runs parent which should handle the wholesale destruction.

comma:
As per discorb need to look into if non-constructible machines doing dismantle is okay

PR #222, opened 2020-03-29
Refactors anomaly protection calculation

comma:
Now instead of gross ass typechecks it's a value of armor, special key just for that.


MistakeNot4892 (dismissed):

code/modules/xenoarcheaology/effect.dm
```diff
@@ -118,29 +118,10 @@
 	if(!istype(H))
 		return 1
 
-	var/protected = 0
+	. = 1
+	for(var/obj/item/I in H.get_equipped_items())
+		var/datum/extension/armor/armor = get_extension(I, /datum/extension/armor)
+		if(armor)
+			. -= armor.get_value("anomaly")/100
```
afterthought2: I mean, should at least run it through `get_blocked` or whatever, and allow the key to be fetched. Defeats the point of armor datums like this.

comma: made it item var instead

MistakeNot4892 (approved):

PR #223, opened 2020-03-29
Converted languages to decls.

MistakeNot4892:
Languages were previously used as singletons with their name as an index to a global list. This changes them to pass the paths around directly (meaning the decl repository handles it all and there are no weird "LANGUAGE_FOO" entries in lists if a macro is typo'd).


comma (approved):

afterthought2:
I'll go on the record as saying I'm not convinced this is a great idea. But as language keys need to be set at compile time anyway and are quite limiting, I don't have a great argument against it.

PR #237, opened 2020-03-31
Fixes all-terrain wheelchairs

### LINKED ISSUE #235: Wheelchairs allow passing through machinery, unhindered.
power

Doors seem to work fine though

deathride58:
Fixes #235 
This PR does exactly as it suggests on the tin. It fixes the bug where wheelchairs are able to pass through machinery and other dense objects that don't override canpass(). This is done by introducing the `MOVABLE_FLAG_NONDENSE_COLLISION` movable flag (by default, non-dense movers can pass through all dense movables, such as machinery. this flag disables that behavior) and adding buckle checks to the canpass() procs of /mob and /obj (this prevents wheelchairs from colliding with the mob that's buckled to it and vice-versa).


MistakeNot4892 (approved):

TheKnightofAura:
Did this fix going through the floor while in a wheelchair? I don't think it's been merged to my downstream yet, but I just went through the floor and was stranded in space due to, what I assume, this.

PR #239, opened 2020-03-31
Reagent-driven fluids.

MistakeNot4892:
This is absurdly costly currently but is otherwise working pretty well.
- Fluids are now driven by a reagent holder in each overlay.
- Fluids will not spread on exoplanet turfs.
- `water_act()` is now `fluid_act()`.
- Fluid pushing and reagent transfer has been moved to `SSobjs`.
- Fluids will call `touch_obj()`, `touch_mob()` and transfer to `touching` in `fluid_act()`.

### DIFF
--- a/code/__defines/fluids.dm
+++ b/code/__defines/fluids.dm
@@ -13,15 +13,6 @@
 #define ADD_ACTIVE_FLUID(F)           SSfluids.active_fluids[F] = TRUE
 #define REMOVE_ACTIVE_FLUID(F)        SSfluids.active_fluids -= F
 
-// Expects /obj/effect/fluid for F, int for amt.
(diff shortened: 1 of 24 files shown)


MistakeNot4892:
This is as good as it will get. I think it will likely be fine in practice but if it ever goes to Bay or another downstream with a large active map it may cause problems.

MistakeNot4892:
Realized while half asleep at 3am that the contact transfer proc doesn't check coverage/permeability. Will need to fix that.

MistakeNot4892:
This can go straight in after #377 is merged and I rebase, performance is improved by literally 100x by the reagent refactor.

MistakeNot4892:
Ready for review.

afterthought2 (approved):

PR #240, opened 2020-04-02
Refactors artifact item finds

comma:
Datumized finds and digsite types, got rid of huge switch statements with defines.
Split up finds_spawning and pretty much rewrote how they work, now it's a series of tubes that apply bunch of modifications to appearance if desired. Got rid of a lot of copypaste with generalizing spawn like that.


MistakeNot4892 (dismissed):
I only skimmed this, but I can't imagine it being worse.

afterthought2 (dismissed):

afterthought2 (approved):

PR #241, opened 2020-04-03
Made geological data stored in an extension

comma:
Previously it was scattered in vars across class tree, resulting in some ugly stuff when you had to get ti from a random item.
Also axed bunch of arbitrary stuff from geodata generation, like 'ages' on materials that meant nothing.

### DIFF
--- a/code/modules/xenoarcheaology/tools/anomaly_scanner.dm
+++ b/code/modules/xenoarcheaology/tools/anomaly_scanner.dm
@@ -25,16 +25,10 @@
 		var/nearestSimpleTargetDist = -1
 		var/turf/cur_turf = get_turf(src)
 
-		for(var/A in SSxenoarch.artifact_spawning_turfs)
-			var/turf/simulated/mineral/T = A
-			if(T.density && T.artifact_find)
-				if(T.z == cur_turf.z)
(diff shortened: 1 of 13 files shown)


MistakeNot4892 (dismissed):

MistakeNot4892 (approved):

code/modules/xenoarcheaology/tools/anomaly_scanner.dm
```diff
@@ -25,16 +25,10 @@
...
-					var/cur_dist = get_dist(cur_turf, T) * 2
-					if(nearestTargetDist < 0 || cur_dist < nearestTargetDist)
-						nearestTargetDist = cur_dist + rand() * 2 - 1
-						nearestTargetId = T.artifact_find.artifact_id
-			else
-				SSxenoarch.artifact_spawning_turfs.Remove(T)
+		var/list/artifact = SSxenoarch.get_nearest_artifact(cur_turf)
+		if(artifact)
+			nearestTargetId = artifact[1]
+			nearestTargetDist = artifact[2]
```
afterthought2, line 31: This will be randomized every time. Doesn't it make more sense to set the extension on a scanned turf, and then get these variables from the cached values on the extension?

PR #243, opened 2020-04-03
Splitting species changes out of /decl/species PR.

MistakeNot4892:
Going to split stuff out until it's just the /decl changes and hopefully isolate where the unit tests start failing.


MistakeNot4892:
Going to need to test a bunch of this before a merge.

comma (commented):
so here we compare result of that proc to species
```
if(H.species.get_root_species_name() != SPECIES_MANTID_ALATE)
```
but here to bodytypes 
```
        var/bodytype = user.species.get_root_species_name(user)
        if(bodytype in list(BODYTYPE_MANTID_SMALL, BODYTYPE_MANTID_LARGE, BODYTYPE_SNAKE))
```
whch one is real

code/modules/species/species_shapeshift.dm
```diff
@@ -13,57 +13,64 @@ var/list/wrapped_species_by_ref = list()
 
 	var/list/valid_transform_species = list()
 	var/monochromatic
-	var/default_form = SPECIES_HUMAN
+	var/default_form
+
+/datum/species/shapeshifter/New()
+	default_form = GLOB.using_map.default_species
+	valid_transform_species |= default_form
+	..()
```
comma, line 22: shapeshifters don't fake their root_species_name is that intended?

MistakeNot4892: Fixed.

MistakeNot4892:
Bodytype was the intended call, fixed.

comma (dismissed):

comma (approved):

Issue #82, opened 2020-03-06
being prone moves head, and some body, overlays one tile left of you.


TheKnightofAura:
#### Description of issue
See title. being prone moves head, and some body, overlays (such as helmets, and backpack bands) one tile left of you.

#### Difference between expected and actual behavior
When I go prone, I expect my clothes and hat to stay on me. (Perhaps the hat is a little idealistic.) but they do not. 
I definately don't expect them to phase into a wall, and move around with me.

#### Steps to reproduce

Rest. Your stuff will float to the side

#### Length of time in which bug has been known to occur

Dunno, honestly. Only found out yesterday

#### Issue bingo

- [x ] Issue could be reproduced at least once
- [ x] Issue could be reproduced by different players
- [ x] Issue could be reproduced in multiple rounds
- [ x] Issue happened in a recent (less than 7 days ago) round
- [x ] [Couldn't find an existing issue about this](https://github.com/NebulaSS13/Nebula/issues)

MistakeNot4892:
Can't reproduce this on Nebula since it doesn't have any aliens that use icon offsets. Can you post if you were a scav and what clothes you were using when it happened?

TheKnightofAura:
Was a scav. Almost any clothing does it, but off the top of my head, backpacks, toolbelt, and welding mask all slid when I went down.

afterthought2:
Well, it's "Fixed." More attention may be warranted in the future, as now species with extra offsets may have some issues that the original PR was trying to fix (not sure), but at any rate should not recur in its present form.

PR #245, opened 2020-04-04
Revert offset calc PR.

MistakeNot4892:
@mikomyazaki pinging you in just so you're informed.

This appears to be causing surgery and mob clothing offsets to be broken (#82). Reverting it caused no issues when tested with a mob with species offsets, and with surgery.


afterthought2 (approved):

afterthought2:
I have no real idea as to what is going on here, but whatever.

PR #258, opened 2020-04-07
Update mapmerge2

PsiOmegaDelta:
Updating didn't help with the newline issue but otherwise appears to work.

Main addition is `update_paths.py` which is a way to automatically replace paths, etc. See the script itself as well as examples over at https://github.com/tgstation/tgstation/tree/master/tools/mapmerge2/map_scripts

### DIFF
--- a/tools/hooks/install.sh
+++ b/tools/hooks/install.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+set -e
 shopt -s nullglob
 cd "$(dirname "$0")"
 for f in *.hook; do
@@ -9,4 +10,8 @@ for f in *.merge; do
 	echo Installing merge driver: ${f%.merge}
 	git config --replace-all merge.${f%.merge}.driver "tools/hooks/$f %P %O %A %B %L"
 done
+
+echo "Installing Python dependencies"
+./python.sh -m pip install -r ../mapmerge2/requirements.txt
+
 echo "Done"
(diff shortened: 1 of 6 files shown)


tools/hooks/install.sh
```diff
@@ -9,4 +10,11 @@ for f in *.merge; do
 	echo Installing merge driver: ${f%.merge}
 	git config --replace-all merge.${f%.merge}.driver "tools/hooks/$f %P %O %A %B %L"
 done
+
+echo "Installing tgui hooks"
```
comma: we dun have TGUI

PsiOmegaDelta: This is indeed true

comma (approved):

tools/mapmerge2/requirements.txt
```diff
@@ -1,3 +1,3 @@
-pygit2>=0.27.2
-bidict>=0.13.1
-Pillow>=5.1.0
+pygit2==1.0.1
+bidict==0.13.1
+Pillow==7.0.0
```
afterthought2, line 3: What happens if you have higher versions installed?

PsiOmegaDelta: Good question apparently without a direct answer, just based on a quick search. I'll get back to you

deathride58: The py script installs and uses the specific version listed in requirements.txt, and pip is responsible for keeping the specific version separate from whatever you've got installed.

deathride58 (approved):

PR #259, opened 2020-04-07
Adding exonet. Removing ntnet.

biscuitWizard:
Features:
  * Completely removes NTNet
  * NTNet is no longer global. Exonets do not broadcast beyond z-levels.
  * Exonet relays act as repeaters, allowing exonets cross Z.
  * Signal simulation is in, actually having an effect now in gameplay.
  * No more email globals. Emails require PLEXUS connections to be sent between exonets, otherwise all emails are internal.
  * Everything is a file. Reports. Logs. Emails. Accounts. Permissions (Grants). Access Records.
  * Added RFID Programmer to Modular Computers, allowing the programming of embedded devices. (Embedded Controller, Exonet Lock)
  * Added an Exonet Lock component, which can be added to any machine to lock its req_access by exonet/grant. Can be programmed by RFID Programmer.
  * Added a "Device Editor" modular program. Full UI for editing.
  * Added an Access Directory console which can be used to edit access roles, grants, link access roles and grants, and print or edit exonet ID cards.
  * Added exonet ID cards, which are wireless ID cards that connect to exonet for permissions.
  * Added router, which broadcasts and acts as a hub for exonet. If this goes down, everything goes down.
  * Added Mainframe, which accepts hard_drives and acts as a file storage server.
  * Made file explorer capable of browsing network drives like mainframes.
  * Removed file_servers. Programs now reside on mainframes, which can be targeted by ntdownload to install programs on modular computers.


comma:
pls don't delete bearcat map we're using it

biscuitWizard:
> pls don't delete bearcat map we're using it

oh no. I think there was a script I was supposed to run to make it merge friendly?

code/modules/modular_computers/terminal/terminal_commands.dm
```diff
@@ -109,153 +116,153 @@ Subtypes
...
+	if(network.intrusion_detection_alarm)
 		. += "NETWORK INCURSION DETECTED"
 
-/datum/terminal_command/locate
-	name = "locate"
-	man_entry = list("Format: locate nid", "Attempts to locate the device with the given nid by triangulating via relays.")
-	pattern = "locate"
-	req_access = list(access_network)
-	skill_needed = SKILL_PROF
+// /datum/terminal_command/locate
```
comma: what's with those comented out commands?

code/game/machinery/_machines_base/stock_parts/exonet_lock.dm
```diff
@@ -0,0 +1,83 @@
...
+	var/list/resulting_grants = list()
+	for(var/datum/computer_file/data/grant_record/grant in grants)
+		for(var/obj/machinery/computer/exonet/mainframe/MF in network.mainframes)
+			if(grant in MF.stored_files)
+				resulting_grants.Add(uppertext("[ennid].[grant.stored_data]"))
+				break
+	return resulting_grants
+
+/obj/item/stock_parts/exonet_lock/attackby(obj/item/I, mob/user)
+	if(istype(I, /obj/item/modular_computer))
```
comma: can skip this, kinda ruins the point of ntos being extension, it's supposed to work same on any object you stick it on.

code/modules/modular_computers/file_system/programs/generic/deck_management.dm
```diff
@@ -326,8 +325,8 @@
 		var/crew = selected_mission.flight_plan.manifest.get_value(in_line = 1)
 		var/time = selected_mission.flight_plan.planned_depart.get_value()
 		if(!crew || !time)
-			to_chat(user, "<span class='warning'>Please fill in the crew manifest and departure time first.</span>")
 			return 1
+			to_chat(user, "<span class='warning'>Please fill in the crew manifest and departure time first.</span>")
```
comma: was this intended?
this code will never run

code/modules/modular_computers/ntos/ntos.dm
```diff
@@ -63,10 +64,10 @@
 		if(hard_drive)
 			if(prob(10))
 				hard_drive.visible_message("<span class='warning'>[src] emits some ominous clicks.</span>")
-				hard_drive.take_damage(0.5 * hard_drive.health)
+				hard_drive.take_damage(20)
```
comma: why hard numbers here?
health of components hadn't been really tested in live env so might end up being tweaked, I think it's better to leave stuff like this with multipliers

code/modules/modular_computers/hardware/rfid_programmer.dm
```diff
@@ -0,0 +1,24 @@
...
+	critical = 0
+	icon_state = "netcard_basic"
+	hardware_size = 1
+	origin_tech = list(TECH_DATA = 4)
+	// usage_flags = PROGRAM_PDA & PROGRAM_TABLET
+
+	var/obj/linked_device			// What device we're editing the permissions for.
+
+/obj/item/stock_parts/computer/rfid_programmer/proc/link_device(var/mob/usr, var/obj/target)
+	linked_device = target
```
comma: this kinda stuff can result in GC failures because device holds the reference to the target.
I think here the easiest way would be to use weakrefs
so here it'd be `weakref(target)`
and in get_device() you'd do 
var/obj/result = linked_device.resolve()
then check sanity/distance and return 'result'

code/modules/modular_computers/hardware/network_card.dm
```diff
@@ -1,119 +1,112 @@
...
-	matter = list(MAT_STEEL = 500, MAT_GLASS = 200)
 
 /obj/item/stock_parts/computer/network_card/wired
-	name = "wired NTNet network card"
-	desc = "An advanced network card for usage with standard NTNet frequencies. This one also supports wired connection."
+	name = "wired EXONET network card"
+	desc = "An advanced network card for usage with standard EXONET frequencies. This one also supports wired connection."
 	ethernet = 1
-	origin_tech = "{'" + TECH_DATA + "':5,'" + TECH_ENGINEERING + "':3}"
+	origin_tech = list(TECH_DATA = 5, TECH_ENGINEERING = 3)
```
comma: no bad they're jsons now

code/modules/modular_computers/hardware/network_card.dm
```diff
@@ -1,119 +1,112 @@
...
+		identification_id = new_guid()
 
 /obj/item/stock_parts/computer/network_card/advanced
-	name = "advanced NTNet network card"
-	desc = "An advanced network card for usage with standard NTNet frequencies. It's transmitter is strong enough to connect even when far away."
+	name = "advanced EXONET network card"
+	desc = "An advanced network card for usage with standard EXONET frequencies. It's transmitter is strong enough to connect even when far away."
 	long_range = 1
-	origin_tech = "{'" + TECH_DATA + "':4,'" + TECH_ENGINEERING + "':2}"
+	origin_tech = list(TECH_DATA = 4, TECH_ENGINEERING = 2)
```
comma: ditto

code/modules/modular_computers/hardware/network_card.dm
```diff
@@ -1,119 +1,112 @@
...
 	if(long_range)
 		. += "511.n WFS/HB (Subspace) - Wide Frequency Spread/High Bandiwdth"
 	if(ethernet)
 		. += "OpenEth (Physical Connection) - Physical network connection port"
 
-/obj/item/stock_parts/computer/network_card/Initialize()
-	. = ..()
-	identification_id = ntnet_card_uid
-	ntnet_card_uid++
+/obj/item/stock_parts/computer/network_card/New(var/l)
```
comma: why New override instead of initialize()?

code/modules/modular_computers/file_system/programs/generic/camera.dm
```diff
@@ -149,7 +148,7 @@
 			L.tracking_cancelled()
 	current_camera = null
 
-/datum/nano_module/camera_monitor/check_eye(var/mob/user)
+/datum/nano_module/camera_monitor/check_eye(var/mob/user as mob)
```
comma: why this?

code/modules/modular_computers/file_system/programs/generic/camera.dm
```diff
@@ -94,7 +93,7 @@
 			return
 		if(!(current_network in C.network))
 			return
-		if(!ARE_Z_CONNECTED(get_z(C), get_z(host)) && !(get_z(C) in GLOB.using_map.admin_levels))
+		if(!AreConnectedZLevels(get_z(C), get_z(host)) && !(get_z(C) in GLOB.using_map.admin_levels))
```
comma: ARE_Z_CONNECTED is new supposedely faster check

code/modules/modular_computers/file_system/programs/engineering/rcon_console.dm
```diff
@@ -125,10 +125,10 @@
...
 	known_SMESs = new /list()
 	for(var/obj/machinery/power/smes/buildable/SMES in SSmachines.machinery)
-		if(ARE_Z_CONNECTED(get_host_z(), get_z(SMES)) && SMES.RCon_tag && (SMES.RCon_tag != "NO_TAG") && SMES.RCon)
+		if(AreConnectedZLevels(get_host_z(), get_z(SMES)) && SMES.RCon_tag && (SMES.RCon_tag != "NO_TAG") && SMES.RCon)
 			known_SMESs.Add(SMES)
 
 	known_breakers = new /list()
 	for(var/obj/machinery/power/breakerbox/breaker in SSmachines.machinery)
-		if(ARE_Z_CONNECTED(get_host_z(), get_z(breaker)) && breaker.RCon_tag != "NO_TAG")
+		if(AreConnectedZLevels(get_host_z(), get_z(breaker)) && breaker.RCon_tag != "NO_TAG")
```
comma: ditto

code/modules/modular_computers/file_system/programs/antagonist/access_decrypter.dm
```diff
@@ -78,7 +77,7 @@
 			return
 
 		var/access = href_list["PRG_execute"]
-		var/obj/item/card/id/id_card = RFID.stored_card
+		var/obj/item/weapon/card/id/id_card = RFID.stored_card
```
comma: whole `/weapon` branch was folded into `/item/`

code/modules/modular_computers/computers/subtypes/preset_pda.dm
```diff
@@ -14,7 +14,7 @@
 	..()
 
 	hard_drive.store_file(new /datum/computer_file/program/email_client())
-	hard_drive.store_file(new /datum/computer_file/program/crew_manifest())
+//	hard_drive.store_file(new /datum/computer_file/program/crew_manifest())
```
comma: ??

code/modules/modular_computers/computers/subtypes/dev_telescreen.dm
```diff
@@ -16,12 +16,12 @@
 	broken_damage = 150
 	w_class = ITEM_SIZE_HUGE
 
-/obj/item/modular_computer/telescreen/Initialize()
-	. = ..()
+/obj/item/modular_computer/telescreen/New()
```
comma: again random init->new and 'as stuff' regressions

code/modules/modular_computers/computers/subtypes/dev_pda.dm
```diff
@@ -25,13 +27,13 @@
 		..()
 
 // PDA box
-/obj/item/storage/box/PDAs
+/obj/item/weapon/storage/box/PDAs
```
comma: /weapon is gone

code/modules/modular_computers/computers/modular_computer/core.dm
```diff
@@ -50,11 +50,12 @@
 		install_default_programs()
 	if(scanner)
 		scanner.do_after_install(null, src)
+	if(network_card)
+		network_card.ennid = initial_ennid
+		network_card.keydata = initial_keydata
 	update_icon()
 	update_verbs()
 	update_name()
-	if(enabled_by_default)
```
comma: seems to be reverting the smoll change, this var and enabling only if it's set

code/modules/exonet/access_directory.dm
```diff
@@ -0,0 +1,250 @@
...
+			to_chat(user, SPAN_WARNING("There appears to already be a card inserted into \the [src]."))
+			return
+		if(!user.unEquip(W, src))
+			return
+		visible_message(SPAN_NOTICE("\The [user] inserts \a [W] into \the [src]."))
+		stored_card = W
+		return
+	return ..()
+
+/obj/machinery/computer/exonet/access_directory/Topic(href, href_list)
```
afterthought2: Use `OnTopic` for this kind of stuff.

code/modules/exonet/access_directory.dm
```diff
@@ -0,0 +1,250 @@
...
+	if(href_list["PRG_back"])
+		error = null
+		editing_user = null
+
+	if(href_list["PRG_changefileserver"])
+		var/old_value = file_server
+		var/list/file_servers = list()
+		for(var/obj/machinery/computer/exonet/mainframe/mainframe in exonet.get_mainframes())
+			LAZYDISTINCTADD(file_servers, exonet.get_network_tag(mainframe))
+		file_server = sanitize(input(usr, "Choose a fileserver to view access records on:", "Select File Server") as null|anything in file_servers)
```
afterthought2: Must check CanInteract or similar after every input.

code/game/machinery/computer/computer.dm
```diff
@@ -52,6 +67,11 @@
 					verbs -= x
 				set_broken(TRUE)
 
+/obj/machinery/computer/bullet_act(var/obj/item/projectile/Proj)
```
comma: re-axe it, there's generic machinery damage stuff now

code/game/machinery/computer/computer.dm
```diff
@@ -21,14 +21,29 @@
...
+
+/obj/machinery/computer/LateInitialize()
+	. = ..()
+	if(!identification_id)
+		identification_id = new_guid()
+	if(ennid)
+		var/datum/extension/exonet_device/exonet = get_extension(src, /datum/extension/exonet_device)
+		exonet.connect_network(null, ennid, NETWORKSPEED_ETHERNET, keydata)
 
-/obj/machinery/computer/get_codex_value()
```
comma, line 29: probably unintended removal

code/game/machinery/_machines_base/stock_parts/_stock_parts.dm
```diff
@@ -9,6 +9,7 @@
 	var/status = 0             // Flags using PART_STAT defines.
 	var/base_type              // Type representing parent of category for replacer usage.
 	health = 50
+	var/tightened			   // Tightened parts cannot be uninstalled.
```
comma: if you want part to be non-removable, remove HAND_REMOVE from its flags

code/modules/modular_computers/hardware/network_card.dm
```diff
@@ -1,119 +1,112 @@
...
-			var/obj/item/stock_parts/computer/network_card/network_card = comp.get_component(PART_NETWORK)
-			if(network_card)
-				return network_card.get_network_tag(routed_through)
+	// if(proxy_id && !(src in routed_through))
+	// 	var/datum/extension/interactive/ntos/comp = ntnet_global.get_os_by_nid(proxy_id)
+	// 	if(comp) // If not we default to exposing ourselves, but it means there was likely a logic error elsewhere.
+	// 		LAZYADD(routed_through, src)
+	// 		var/obj/item/stock_parts/computer/network_card/network_card = comp.get_component(PART_NETWORK)
+	// 		if(network_card)
+	// 			return network_card.get_network_tag(routed_through)
```
afterthought2: Should just axe all proxy code if it doesn't work anymore.

code/game/machinery/_machines_base/machinery_components.dm
```diff
@@ -162,6 +162,8 @@ GLOBAL_LIST_INIT(machine_path_to_circuit_type, cache_circuits_by_build_path())
 		return
 
 	if(istype(part))
+		if(part.tightened)
+			return
```
afterthought2: Extremely unsafe; this proc is expected to always work. Use the HAND_REMOVE flag on the part instead if needed.

code/game/machinery/_machines_base/stock_parts/exonet_lock.dm
```diff
@@ -0,0 +1,83 @@
...
+	var/datum/exonet/network = exonet.get_local_network()
+	if(!network)
+		return get_default_access()
+	var/list/resulting_grants = list()
+	for(var/datum/computer_file/data/grant_record/grant in grants)
+		for(var/obj/machinery/computer/exonet/mainframe/MF in network.mainframes)
+			if(grant in MF.stored_files)
+				resulting_grants.Add(uppertext("[ennid].[grant.stored_data]"))
+				break
+	return resulting_grants
```
afterthought2: Confusing usage; this return value is not in standard access form.

code/modules/modular_computers/file_system/programs/engineering/device_editor.dm
```diff
@@ -0,0 +1,142 @@
...
+	if(href_list["key"])
+		var/new_key = sanitize(input(usr, "Enter exonet keypass or leave blank to cancel:", "Change key"))
+		if(!new_key)
+			return 1
+		lock.keydata = new_key
+		lock.grants = list()
+	if(href_list["PRG_disable"])
+		lock.tightened = FALSE
+	if(href_list["PRG_enable"])
+		lock.tightened = TRUE
```
afterthought2: Not a fan of this being software controlled. If you want to restrict access to it, install a hardware lock as well.

afterthought2 (commented):
Looks like there are some merge artifacts.

Not really into how the grants are handled, and in particular how they are always combined via AND. Should probably just be access datums that you create and give a unique sequential key, that have a name so they properly show up on various access modification systems already in the game, and should allow OR combining in standard access format. Yes, I realize that by adding an arbitrary number of access grants you can simulate the same set of access combinations as currently, but it's cumbersome and requires nonlocal (and in many cases not straightforward) changes each time you want to add several accesses with OR.

code/game/machinery/_machines_base/stock_parts/exonet_lock.dm
biscuitWizard: Expected format is not a list of strings?

code/modules/modular_computers/file_system/programs/engineering/device_editor.dm
biscuitWizard: Not a restricted access thing. It's to prevent circumventing req_access-level control by removing the lock. Balance issue.

Followup evolve on this implementation is removing req_access entirely, which is redundant, and this needing the same-ish circumvention gameplay pathways.

code/modules/modular_computers/hardware/network_card.dm
biscuitWizard: Good idea. Proxies have no gameplay value.

code/modules/exonet/access_directory.dm
biscuitWizard: Difference between Topic and OnTopic?

biscuitWizard:
> Looks like there are some merge artifacts.
> 
> Not really into how the grants are handled, and in particular how they are always combined via AND. Should probably just be access datums that you create and give a unique sequential key, that have a name so they properly show up on various access modification systems already in the game, and should allow OR combining in standard access format. Yes, I realize that by adding an arbitrary number of access grants you can simulate the same set of access combinations as currently, but it's cumbersome and requires nonlocal (and in many cases not straightforward) changes each time you want to add several accesses with OR.

Major change. Requires UI changes, too. I'm personally not a fan. KISS. Like you said, just make new grants. If anything, I would change it so it's all OR instead of AND. I assume this has something to do with this "standard access format".

code/modules/exonet/access_directory.dm
afterthought2: One calls the other after making some checks; OnTopic does not need to call parent unless you specifically need the parent override. If you do call Topic, you should return `..()`, not `TOPIC_HANDLED`.

code/game/machinery/_machines_base/stock_parts/exonet_lock.dm
afterthought2: Expected format is a list of strings or lists, with every string corresponding to an access datum.

code/modules/modular_computers/file_system/programs/engineering/device_editor.dm
afterthought2: I understand how it works, I just don't like it from a design perspective. I think a pathway for hardware circumvention should exist.

biscuitWizard: EMAG? Could add an attack by welder that shunts it into default mode, whatever the user set it to.

biscuitWizard: There's also still wires, for airlocks.

afterthought2: I think having it be removable and adding a hardware lock if you want it to not be is sufficient. @comma can weigh in if he feels differently.

comma: There's parts already existing for preventing hardware access, don't see why this is needed.

biscuitWizard: What do you mean preventing hardware access? Exonet lock doesn't prevent you from accessing the hardware, only the lock itself?

biscuitWizard: Unrelated question, if the access_lock prevents opening the access panel, how does an access_lock get the permissions to unlock/lock the access lock if the permissions to operate the device are on the exolock?

comma: pretty sure that means it's not preventing jack shit then because if you got to hatch-open stage you can just do full deconstruction step with a crowbar and rebuild machine without the exonet lock part.

biscuitWizard: Ugh. That's true.

code/game/machinery/_machines_base/stock_parts/exonet_lock.dm
comma: see https://github.com/NebulaSS13/Nebula/wiki/Access

biscuitWizard: Reviewed. My recommendation is to remove req_access entirely, change it to mapping_access or something, and I can set exonet to generate access datums and grant files automatically for "station_ennid" or something on Init().

biscuitWizard:
Confirmed Changes To Make:
* Fix access list to be ORs not ANDs
* Remove tightened
* Make sure access_lock respects exonet configuration.
* Merge Artifacts Begone
* Mainframe needs a connection/config UI
* Remove terminal program. This thing is neat in concept and dumb in practice.
* Add a WHOIS program for listing all devices on a network. Add ban/unban button if network_administrator.

code/modules/exonet/access_directory.dm
biscuitWizard: What's CanInteract?

code/modules/modular_computers/hardware/rfid_programmer.dm
```diff
@@ -0,0 +1,24 @@
...
+/obj/item/stock_parts/computer/rfid_programmer/proc/link_device(var/mob/usr, var/obj/target)
+	linked_device = weakref(target)
+	if(target)
+		to_chat(usr, "\The [src] chirps in acknowledgement as it's linked to \a [target].")
+
+
+/obj/item/stock_parts/computer/rfid_programmer/proc/get_device()
+	if(!linked_device)
+		return
+	if(get_dist(get_turf(src), get_turf(linked_device)) > 1)
```
comma: you're trying to get a turf of a datum here.
this proc should be
```
        var/obj/result = linked_device.resolve() 
	if(!result)
		return
	if(get_dist(get_turf(src), get_turf(result)) > 1)
                return
	return result 
```

code/modules/exonet/access_directory.dm
```diff
@@ -0,0 +1,251 @@
+/obj/machinery/computer/exonet/access_directory
+	name = "EXONET Access Controller"
```
MistakeNot4892: `\improper`

code/modules/assembly/signaler.dm
```diff
@@ -62,7 +62,7 @@
 	return
 
 
-/obj/item/assembly/signaler/Topic(href, href_list, state = GLOB.physical_state)
+/obj/item/assembly/signaler/Topic(var/mob/user, href_list, state = GLOB.physical_state)
```
MistakeNot4892: `Topic` does not have a `user` arg.

code/modules/exonet/_exonet_machine.dm
```diff
@@ -0,0 +1,138 @@
...
+	data["power_usage"] = active_power_usage / 1000
+
+	. = data
+
+/obj/machinery/computer/exonet/interface_interact(mob/user)
+	ui_interact(user)
+	return TRUE
+
+/obj/machinery/computer/exonet/Destroy()
+	..()
```
MistakeNot4892: Must return value of parent.

code/modules/exonet/_exonet_machine.dm
```diff
@@ -0,0 +1,138 @@
...
+	if(..())
+		return TOPIC_HANDLED
+	. = TOPIC_REFRESH
+	var/datum/extension/exonet_device/exonet = get_extension(src, /datum/extension/exonet_device)
+	
+	if(href_list["options"])
+		clear_errors()
+		current_ui_template = "exonet_options.tmpl"
+		rebuild_ui = TRUE
+		return TOPIC_HANDLED
```
MistakeNot4892, line 82: Wouldn't you want to return `TOPIC_REFRESH` here?

code/modules/modular_computers/file_system/programs/generic/library.dm
```diff
@@ -0,0 +1,190 @@
...
+	if(href_list["uploadbook"])
+		if(!scanner || !scanner.anchored)
+			scanner = null
+			error_message = "Hardware Error: No scanner detected. Unable to access cache."
+			return 1
+		if(!scanner.cache)
+			error_message = "Interface Error: Scanner cache does not contain any data. Please scan a book."
+			return 1
+
+		var/obj/item/weapon/book/B = scanner.cache
```
MistakeNot4892: Path removal.

code/modules/modular_computers/file_system/programs/generic/library.dm
```diff
@@ -0,0 +1,190 @@
...
+			error_message = "Software Error: Unable to print; book not found."
+			return 1
+
+		//PRINT TO BINDER
+		if(!nano_host())
+			return 1
+		for(var/d in GLOB.cardinal)
+			var/obj/machinery/bookbinder/bndr = locate(/obj/machinery/bookbinder, get_step(nano_host(), d))
+			if(bndr && bndr.anchored)
+				var/obj/item/weapon/book/B = new(bndr.loc)
```
MistakeNot4892: Path removal.

code/modules/modular_computers/file_system/programs/generic/library.dm
```diff
@@ -0,0 +1,190 @@
...
+		data["scanner"] = istype(scanner)
+
+	ui = SSnano.try_update_ui(user, src, ui_key, ui, data, force_open)
+	if (!ui)
+		ui = new(user, src, ui_key, "library.tmpl", "Library Program", 575, 700, state = state)
+		ui.auto_update_layout = 1
+		ui.set_initial_data(data)
+		ui.open()
+
+/datum/nano_module/library/Topic(href, href_list)
```
MistakeNot4892: This whole module is going to have problems as the library stuff got removed on Neb in favour of a persistence category.

code/modules/modular_computers/file_system/programs/generic/news_browser.dm
```diff
@@ -0,0 +1,128 @@
+/datum/computer_file/program/newsbrowser
```
MistakeNot4892: If this is the `SPACE` HTML loader thing, it also got removed on Neb,

code/modules/modular_computers/hardware/network_card.dm
```diff
@@ -1,25 +1,24 @@
...
-	desc = "A basic network card for usage with standard NTNet frequencies."
+/obj/item/stock_parts/computer/network_card/
+	name = "basic EXONET network card"
+	desc = "A basic network card for usage with standard EXONET frequencies."
 	power_usage = 50
 	origin_tech = "{'" + TECH_DATA + "':2,'" + TECH_ENGINEERING + "':1}"
 	critical = 0
 	icon_state = "netcard_basic"
 	hardware_size = 1
-	matter = list(MAT_STEEL = 250, MAT_GLASS = 100)
```
MistakeNot4892, line 11: Unintended removal?

code/modules/modular_computers/hardware/network_card.dm
```diff
@@ -28,92 +27,84 @@ var/global/ntnet_card_uid = 1
...
-	name = "wired NTNet network card"
-	desc = "An advanced network card for usage with standard NTNet frequencies. This one also supports wired connection."
+	name = "wired EXONET network card"
+	desc = "An advanced network card for usage with standard EXONET frequencies. This one also supports wired connection."
 	ethernet = 1
 	origin_tech = "{'" + TECH_DATA + "':5,'" + TECH_ENGINEERING + "':3}"
 	power_usage = 100 // Better range but higher power usage.
 	icon_state = "netcard_ethernet"
 	hardware_size = 3
-	matter = list(MAT_STEEL = 2500, MAT_GLASS = 400)
```
MistakeNot4892, line 53: Unintentional removal here and Destroy() below?

code/modules/exonet/_exonet_machine.dm
biscuitWizard: Why would I want to?

code/modules/modular_computers/hardware/network_card.dm
biscuitWizard: Destroy removal was appropriate. It is not needed.

code/modules/exonet/_exonet_machine.dm
biscuitWizard: Sorry, that was a rhetorical leading question. The answer is no: I don't want to because the UI is destroyed immediately after this. See what rebuild_ui does.

code/datums/exonet.dm
```diff
@@ -0,0 +1,226 @@
...
+	if(router.keydata != keydata)
+		return 0 // Authentication failed.
+
+	if(istype(device, /obj/machinery/computer/exonet/mainframe))
+		mainframes |= device
+	else if(istype(device, /obj/machinery/computer/exonet/broadcaster))
+		broadcasters |= device
+	else if(istype(device, /obj/machinery/computer/exonet/uplink))
+		modems |= device
+	else if(istype(device, /obj/machinery/computer/exonet/broadcaster/router) && !router)
```
comma, line 104: this will never be true because proc will quit early if `router` is null.

code/datums/exonet.dm
```diff
@@ -0,0 +1,226 @@
...
+		modems -= device
+	network_devices -= device
+
+/datum/exonet/proc/set_router(var/device)
+	if(router)
+		// THERE CAN BE ONLY ONE!!!
+		LAZYREMOVE(network_devices, router)
+		LAZYREMOVE(broadcasters, router)
+	router = device
+	LAZYADD(network_devices, device)
```
comma, line 124: maybe LAZYDISTINCTADD here and for broadcasters?
I didn't read to ther spot where it's used, but would make it safe to set router that's already in the network.

code/datums/exonet.dm
```diff
@@ -0,0 +1,226 @@
...
+	logfile.stored_data = jointext(logs, "\[br\]")
+
+/datum/exonet/proc/get_signal_strength(var/obj/device, var/netspeed)
+	var/best_signal = -1
+	var/turf/device_turf = get_turf(device)
+	if(!device_turf)
+		return best_signal
+	for(var/obj/machinery/computer/exonet/broadcaster/broadcaster in broadcasters)
+		var/turf/b_turf = get_turf(broadcaster)
+		if(b_turf.z != device_turf.z || !broadcaster.operable())
```
comma, line 158: are broadcasters intended to be one per z-level, rather than one per connected-zlevels?

code/datums/exonet.dm
```diff
@@ -0,0 +1,226 @@
...
+/datum/exonet/New(var/new_ennid)
+	if(new_ennid)
+		ennid = new_ennid
+		GLOB.exonets[new_ennid] = src
+
+/datum/exonet/proc/change_ennid(var/new_ennid)
+	// aw god. pls never call this...
+	for(var/datum/network_device in network_devices)
+		if("ennid" in network_device.vars)
+			network_device.vars["ennid"] = new_ennid
```
comma, line 36: jesus on a stick
this is what extensions are for.
this var is for mapping etc right? Do you actually need to change it during runtime, instead of just using for initial extension setup and letting extension do the things?

code/datums/extensions/exonet.dm
```diff
@@ -0,0 +1,161 @@
...
+
+	var/datum/exonet/old_exonet = GLOB.exonets[ennid]
+	if(old_exonet)
+		old_exonet.remove_device(holder)
+		ennid = null
+	return FALSE // This is a success.
+
+/datum/extension/exonet_device/proc/broadcast_network(var/b_ennid, var/key)
+	// Broadcasts an ENNID. If the network doesn't exist, it will create it.
+	// If the network does exist, this will attempt to be added as a relay.
```
comma, line 99: `will attempt to be added as a relay` seems to be just added like a normal device tho?

code/datums/extensions/exonet.dm
```diff
@@ -0,0 +1,161 @@
...
+/datum/extension/exonet_device/proc/get_network_tag(var/obj/device)
+	// Gets a friendly, unique name for a device on a local network.
+	var/datum/exonet/network = get_local_network()
+	if(!network)
+		return
+	if(istype(device, /obj/machinery/computer/exonet))
+		var/obj/machinery/computer/exonet/exonet_machine = device
+		if(exonet_machine.net_tag)
+			return exonet_machine.net_tag
+	var/index = network.network_devices.Find(device)
```
comma: that seems like there's possibility of collision
1.device that's before this one in list is is removed 
2.device of same type is added, now occupying this device's spot number-wise
3.same tag
Should probably just use sequential_id(device.type) or something IMO

code/datums/extensions/exonet.dm
```diff
@@ -0,0 +1,161 @@
...
+	var/datum/exonet/network = get_local_network()
+	if(!network)
+		return
+	for(var/obj/machinery/computer/exonet/exonet_machine in network.network_devices)
+		if(!exonet_machine.net_tag)
+			continue
+		if(exonet_machine.net_tag == net_tag)
+			return exonet_machine
+	var/list/tokens = splittext(net_tag, ".")
+	var/index = text2num(tokens[length(tokens)])
```
comma: that looks fragile as fuck, unless every time a device is removed form network you loop over every single device and update their network tag to update their indext in the list.

comma: Also it seems these can be changed by user, who will probably fail to set right list index.

code/datums/exonet.dm
```diff
@@ -0,0 +1,226 @@
...
+
+
+/datum/exonet/New(var/new_ennid)
+	if(new_ennid)
+		ennid = new_ennid
+		GLOB.exonets[new_ennid] = src
+
+/datum/exonet/proc/change_ennid(var/new_ennid)
+	// aw god. pls never call this...
+	for(var/datum/network_device in network_devices)
```
comma, line 34: datums?..

code/__defines/misc.dm
```diff
@@ -112,17 +112,17 @@
...
-#define NTNET_COMMUNICATION 3		// Communication (messaging)
-#define NTNET_SYSTEMCONTROL 4		// Control of various systems, RCon, air alarm control, etc.
-
-// NTNet transfer speeds, used when downloading/uploading a file/program.
-#define NTNETSPEED_LOWSIGNAL 0.5	// GQ/s transfer speed when the device is wirelessly connected and on Low signal
-#define NTNETSPEED_HIGHSIGNAL 1	// GQ/s transfer speed when the device is wirelessly connected and on High signal
-#define NTNETSPEED_ETHERNET 2		// GQ/s transfer speed when the device is using wired connection
-#define NTNETSPEED_DOS_AMPLIFICATION 5	// Multiplier for Denial of Service program. Resulting load on NTNet relay is this multiplied by NTNETSPEED of the device
+										// EXONET module-configuration values. Do not change these. If you need to add another use larger number (5..6..7 etc)
+#define	NETWORK_SOFTWAREDOWNLOAD 	1 	// Downloads of software from EXONET
```
MistakeNot4892, line 116: Tiny tiny note, you put a tab here instead of a space.

code/__defines/misc.dm
```diff
@@ -141,14 +141,14 @@
 #define PROG_OFFICE  	"Office Work"
 #define PROG_COMMAND  	"Command"
 #define PROG_SUPPLY  	"Supply and Shuttles"
-#define PROG_ADMIN  	"NTNet Administration"
+#define PROG_ADMIN  	"EXONET Administration"
```
MistakeNot4892, line 144: I almost feel like the actual name of the network should be a map-level var, with 'Network' used in general. Always bugged me that NTNet and NTOS were hard-baked into things and couldn't be easily removed or changed.

Not really a note for this PR, more one for the future.

code/_helpers/guid.dm
```diff
@@ -0,0 +1,10 @@
+// I'm tired of seeing everyone litter their UIDs everywhere. That's fine for round-based play where everything is reset,
+// but for any sort of persistent play, it's absolutely fucking hell. It'll also one day just break catastrophically.
+//
+// do better.
+//
+// Use GUIDs like people fucking expect in sane environments.
+
+//#define NEW_GUID() "[num2hex(rand(1,255))][num2hex(rand(1,255))][num2hex(rand(1,255))][num2hex(rand(1,255))]-[num2hex(rand(1,255))][num2hex(rand(1,255))]-[num2hex(rand(1,255))][num2hex(rand(1,255))]-[num2hex(rand(1,255))][num2hex(rand(1,255))]-[num2hex(rand(1,255))][num2hex(rand(1,255))][num2hex(rand(1,255))][num2hex(rand(1,255))][num2hex(rand(1,255))][num2hex(rand(1,255))]";
```
MistakeNot4892, line 8: These comments are pretty useless and should probably get trimmed.

code/_helpers/guid.dm
```diff
@@ -0,0 +1,10 @@
...
+/proc/new_guid()
+	return "[num2hex(rand(1,255))][num2hex(rand(1,255))][num2hex(rand(1,255))][num2hex(rand(1,255))]-[num2hex(rand(1,255))][num2hex(rand(1,255))]-[num2hex(rand(1,255))][num2hex(rand(1,255))]-[num2hex(rand(1,255))][num2hex(rand(1,255))]-[num2hex(rand(1,255))][num2hex(rand(1,255))][num2hex(rand(1,255))][num2hex(rand(1,255))][num2hex(rand(1,255))][num2hex(rand(1,255))]"
```
MistakeNot4892, line 10: Couldn't this be a define as the comment above? Slightly faster. Not sure how hot this proc is anyway.

code/datums/exonet.dm
```diff
@@ -0,0 +1,226 @@
...
+		if("ennid" in network_device.vars)
```
MistakeNot4892, line 35: Seconding @comma, this is really not kosher, it's `:` with extra steps.

MistakeNot4892 (commented):
Didn't get very far in this before I needed to take a break, agree with chinsky's comments so far though.

code/modules/modular_computers/file_system/programs/generic/configurator.dm
```diff
@@ -35,6 +66,30 @@
...
+			data["signal_strength"] = "Not Connected"
+		else
+			var/signal_strength = network.get_signal_strength(network_card, network_card.get_netspeed())
+			if(signal_strength <= 0)
+				data["signal_strength"] = "Not Connected"
+			else if(signal_strength <= 6)
+				data["signal_strength"] = "Low Signal"
+			else
+				data["signal_strength"] = "High Signal"
+		if(network_card.ennid)
```
comma: I think these params should be pulled out of the extension instead, these seems to be this weird duplication of state between objects and extension they hold

code/modules/modular_computers/file_system/programs/engineering/exonet_monitor.dm
```diff
@@ -0,0 +1,106 @@
...
+	if(..())
+		return TOPIC_HANDLED
+
+	if(!user.skill_check(SKILL_COMPUTER, SKILL_BASIC))
+		return TOPIC_HANDLED
+
+	var/datum/exonet/network = get_network()
+	if(program.error)
+		return TOPIC_HANDLED
+	// if(href_list["toggleWireless"])
```
comma, line 68: don't leave commented out code in

code/modules/modular_computers/file_system/programs/generic/ntdownloader.dm
```diff
@@ -64,26 +68,26 @@
 
 	return 1
 
-/datum/computer_file/program/ntnetdownload/proc/hide_file_info(datum/computer_file/file, skill)
-	server = (file in ntnet_global.available_station_software) ? "NTNet Software Repository" : "unspecified server"
+/datum/computer_file/program/exonetdownload/proc/hide_file_info(datum/computer_file/file, skill)
+	//server = (file in ntnet_global.available_station_software) ? "NTNet Software Repository" : "unspecified server"
```
comma, line 72: What's this for? 
Kinda getting tired of these random commented out bits, either reimplement or remove cleanly. Don't leave half-broken stuff in with a TODO.

code/modules/modular_computers/file_system/programs/generic/ntdownloader.dm
```diff
@@ -136,18 +140,27 @@
...
-/datum/nano_module/program/computer_ntnetdownload
+/datum/nano_module/program/computer_exonetdownload
 	name = "Network Downloader"
 
-/datum/nano_module/program/computer_ntnetdownload/ui_interact(mob/user, ui_key = "main", var/datum/nanoui/ui = null, var/force_open = 1, var/datum/topic_state/state = GLOB.default_state)
+/datum/nano_module/program/computer_exonetdownload/proc/get_functional_network_card()
+	var/datum/extension/interactive/ntos/os = get_extension(nano_host(), /datum/extension/interactive/ntos)
+	var/obj/item/stock_parts/computer/network_card/network_card = os && os.get_component(/obj/item/stock_parts/computer/network_card)
+	if(!network_card || !network_card.check_functionality())
+		// error = "Error establishing connection. Are you using a functional and NTOSv2-compliant device?"
```
comma, line 150: why is this commented out?

code/game/machinery/_machines_base/stock_parts/exonet_lock.dm
```diff
@@ -0,0 +1,81 @@
...
+/obj/item/stock_parts/exonet_lock/examine(mob/user)
+	. = ..()
+	if(emagged && user.skill_check_multiple(list(SKILL_FORENSICS = SKILL_EXPERT, SKILL_COMPUTER = SKILL_EXPERT)))
+		to_chat(user, SPAN_WARNING("On close inspection, there is something odd about the interface. You suspect it may have been tampered with."))
+
+/obj/item/stock_parts/exonet_lock/buildable
+	part_flags = PART_FLAG_HAND_REMOVE
+	matter = list(MAT_STEEL = 400, MAT_GLASS = 200)
+
+/decl/stock_part_preset/exonet_lock
```
comma, line 76: this preset seems pretty useless since it's setting req access that isn't actually used in lieu of a proc, is it really needed?

code/modules/exonet/mainframe.dm
```diff
@@ -0,0 +1,162 @@
...
+/obj/machinery/computer/exonet/mainframe/proc/get_log_file()
+	var/datum/computer_file/data/logfile/file = find_file_by_name("network_log")
+	if(!file)
+		file = new()
+		file.filename = "network_log"
+		store_file(file)
+	return file
+
+// Use this proc to add file to the drive. Returns 1 on success and 0 on failure. Contains necessary sanity checks.
+/obj/machinery/computer/exonet/mainframe/proc/store_file(var/datum/computer_file/F)
```
comma, line 49: wouldn't whole thing be easier if it'd be using ntos for file stuff? Seems like it's pretty much reimplementing all the filessystem stuff from there.

code/datums/exonet.dm
biscuitWizard: Yes.

code/datums/exonet.dm
biscuitWizard: This is not for mapping. You can change the ennid on the router and that calls this.

code/datums/extensions/exonet.dm
biscuitWizard: add_device in exonet datum does the check to add it as a broadcaster.

code/datums/exonet.dm
biscuitWizard: /datum because I need access to .vars.

code/modules/exonet/mainframe.dm
biscuitWizard: Actually is a copy of hard_drive. Reason that is, is because this is an equivalent LVM. There's a TODO for future to add in some RAID-type simulation and actually write down to the HDD level but the interface needed exposed on a machine level.

code/datums/exonet.dm
biscuitWizard: the link between broadcasters is wuju space transmitters, able to link to each other across vast distances in space. That tech is bulky. The wifi transmitters fittable in portable devices cannot do that reasonably, or whatever. And thus their range is smaller.

You could put a relay on an exoplanet and connect it to Torch's network, for example.

code/_helpers/guid.dm
biscuitWizard: Would just a link to http://guid.one/guid be better?

code/datums/exonet.dm
```diff
@@ -213,7 +219,7 @@ GLOBAL_LIST_EMPTY(exonets)
 		user.StoreMemory("Your email account address has been changed to [new_login].", /decl/memory_options/system)
 	if(issilicon(user))
 		var/mob/living/silicon/S = user
-		var/datum/nano_module/email_client/my_client = S.get_subsystem_from_path(/datum/nano_module/email_client)
+		var/datum/nano_module/program/email_client/my_client = S.get_subsystem_from_path(/datum/nano_module/program/email_client)
```
comma, line 222: making it a /program module makes it impossible for silicons to use, since it requires a NTOS device to run.

code/modules/modular_computers/file_system/programs/generic/email_client.dm
```diff
@@ -177,22 +178,22 @@
...
-/datum/nano_module/email_client/ui_interact(mob/user, ui_key = "main", var/datum/nanoui/ui = null, var/force_open = 1, var/datum/topic_state/state = GLOB.default_state)
+/datum/nano_module/program/email_client/ui_interact(mob/user, ui_key = "main", var/datum/nanoui/ui = null, var/force_open = 1, var/datum/topic_state/state = GLOB.default_state)
 	var/list/data = host.initial_data()
 
 	// Password has been changed by other client connected to this email account
 	if(current_account)
 		if(current_account.password != stored_password)
 			if(!log_in())
 				log_out()
-				error = "Invalid Password"
```
comma, line 189: why are you changing this everywhere to program's var?

biscuitWizard: There was copy pasting everywhere. That's why I moved the whole namespace to /program, because I made error at the program level rather than the nano module level.

code/datums/exonet.dm
biscuitWizard: Not a note for you specifically,  but that really feels like something important that should be in a comment or document somewhere in /program, because I had no idea and it has now wasted a significant chunk of time I spent fixing what I thought was outdated code.

comma:
I poked about at making a skeletal version of this and realized there's some important considerations for making ntnet actually non-global.

How does it handle crew records? You say they're all files now, but I don't see any special handling for generation of records for joining people. How does it handle the stuff like secHUDs querying arrest status? As a note, a lot of OOC things seem to be using crew records, so probably will have to do something about them or keep a copy of records for them.

What about cargo program? With how it is now, people would be able to order stuff for main cargo shuttle from any network, because cargo is global.

biscuitWizard:
> I poked about at making a skeletal version of this and realized there's some important considerations for making ntnet actually non-global.
> 
> How does it handle crew records? You say they're all files now, but I don't see any special handling for generation of records for joining people. How does it handle the stuff like secHUDs querying arrest status? As a note, a lot of OOC things seem to be using crew records, so probably will have to do something about them or keep a copy of records for them.
> 
> What about cargo program? With how it is now, people would be able to order stuff for main cargo shuttle from any network, because cargo is global.

Should follow the same policies as the email system -- check for a map datum "station_ennid" which is defined, and use that against the GLOB list of all exonets to find a mainframe mapped in as the records server on the exonet datum.

Create all crew records on that.

All these programs and the crew records system should already be doing that. The email system definitely is, and I'm 97% sure the other systems are too, but I could be wrong. But I definitely made sure email does that in the mob proc.

biscuitWizard:
It looks like I reverted the changes from all_crew_records a while ago. That does need changing to the paradigm I listed above.

station_ennid is the replacement for that system, as the necessity of auto-generated crew records is solely a design consideration for round based play and not persistent play.

biscuitWizard:
Initial design proposal for crew records:
Change all GLOB.all_crew_records to pull from the device's connected exonet, (or if a non-device check like in human_helpers, or names.dm, feel free to use the station_ennid network as appropriate.) The check should be similar to get_email_accounts(), where the network is checked and if the local network plexus is checked, all other mainframes of connected networks should also be checked.

This can be changed later in further iterations for further sophistication.

/obj/item/uplink_service/fake_crew_announcement should have a new verb added to change the target network. default to station_ennid

Remove despawn_occupant's entry for crew records in cryopod.

Add a configuration for holopad to link caller_id to an exonet.

For /obj/item/uplink add a verb for changing the target network.

secHUD needs to connect to ennid. Add the extension to that device.

Same for medHUD.

fully_replace_character_name in human_helpers should remove the crew_record update for now.

mob/living/silicon/ai should pull from station_ennid crew records.

Everything else is program related and should pull from local exonet.

Final Design:
Some kind of PLEXUS-based social security record enhancement for crew records for basic essential information exists on space internet, and is enhanced by crew records, with a central law enforcement records agency impl. Out of scope.

PR #262, opened 2020-04-07
Reworks radio filter usage.

afterthought2:
May be (pick one or both):

- Drastically more performant than current usage (divide cost by O(radio enabled machines))
- hilariously broken


comma:
Docking airlock for shuttle on beercat has all buttons in disabled state
EDIT: nvm im dumb that's not how docking airlocks work, needed either to override or just move shuttle

comma (dismissed):

comma (approved):

PR #266, opened 2020-04-08
Made planets overmap icons 64x64

comma:
Just visual change (and I guess being able to reach it from more tiles?)
Done because Pawn wouldn't stop bitching about how im wasting the big sprites by not using them.
image


MistakeNot4892:
This would look nicer if you offset them by -16,-16 so the main bulk sat in a turf :(

MistakeNot4892 (approved):

PR #268, opened 2020-04-09
Probably fixes medical side effects

quardbreak:
Opening for diff. Ports some medical side effects changes from Infinitystation

### DIFF
--- a/code/modules/mob/living/carbon/human/MedicalSideEffects.dm
+++ b/code/modules/mob/living/carbon/human/MedicalSideEffects.dm
@@ -11,11 +11,11 @@
 /datum/medical_effect/proc/manifest(mob/living/carbon/human/H)
 	for(var/R in cures)
 		if(!H.reagents || H.reagents.has_reagent(R))
-			return 0
+			return FALSE
 	for(var/R in triggers)
 		if(H.reagents.get_reagent_amount(R) >= triggers[R])
-			return 1
-	return 0
+			return TRUE
+	return FALSE
 
 /datum/medical_effect/proc/on_life(mob/living/carbon/human/H, strength)
 	return
@@ -24,10 +24,9 @@
 	for(var/R in cures)
 		if(H.reagents.has_reagent(R))
 			if (cure_message)
-				to_chat(H, "<span class='notice'>[cure_message]</span>")
-			return 1
-	return 0
-
+				to_chat(H, SPAN_NOTICE("[cure_message]"))
+			return TRUE
+	return FALSE
 
 // MOB HELPERS
 // ===========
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/human/MedicalSideEffects.dm
```diff
@@ -40,23 +39,21 @@
...
-	var/T = side_effects[name]
-	if (!T)
-		return
-
-	var/datum/medical_effect/M = new T
-	if(M.name == name)
-		M.strength = strength
-		M.start = life_tick
-		side_effects += M
+	var/list/L = subtypesof(/datum/medical_effect)
```
MistakeNot4892:
```
	for(var/T in subtypesof(/datum/medical_effect))
		var/datum/medical_effect/M = T
		if(initial(M.name) == name)
			M = new T
			M.strength = strength
			M.start = life_tick
			side_effects += M
			break
```
Either this, or you should be deleting `M` if it isn't valid.

code/modules/mob/living/carbon/human/MedicalSideEffects.dm
```diff
@@ -40,23 +39,21 @@
...
 /mob/living/carbon/human/proc/handle_medical_side_effects()
 	//Going to handle those things only every few ticks.
 	if(life_tick % 15 != 0)
-		return 0
+		return FALSE
 
-	var/list/L = typesof(/datum/medical_effect)-/datum/medical_effect
+	var/list/L = subtypesof(/datum/medical_effect)
 	for(var/T in L)
 		var/datum/medical_effect/M = new T
```
MistakeNot4892, line 58: As above, these need to use `initial()` or clean up after being instantiated. Preferrably they should be cached somewhere for reference.

code/modules/mob/living/carbon/human/MedicalSideEffects.dm
```diff
@@ -87,15 +84,21 @@
...
 
 /datum/medical_effect/headache/on_life(mob/living/carbon/human/H, strength)
-	var/obj/item/organ/external/head/head = H.get_organ("head")
-	if(istype(head))
+	var/obj/item/organ/external/head/head = H.get_organ(BP_HEAD)
+	if(head)
 		switch(strength)
 			if(1 to 10)
-				H.custom_pain("You feel a light pain in your head.",0, affecting = head)
+				H.custom_pain("You feel a light pain in your head.", 5, affecting = head)
```
MistakeNot4892: Maybe `in your [head.name]` for these, for species without 'heads'?

code/modules/mob/living/carbon/human/MedicalSideEffects.dm
```diff
@@ -106,13 +109,15 @@
...
-			H.custom_pain("You feel a bit light around the stomach.",0)
-		if(11 to 30)
-			H.custom_pain("Your stomach hurts.",0)
-		if(31 to INFINITY)
-			H.custom_pain("You feel sick.",1)
+	var/obj/item/organ/external/head/groin = H.get_organ(BP_GROIN) //INF
+	if(groin)
+		switch(strength)
+			if(1 to 10)
+				H.custom_pain("You feel a bit light around the stomach.", 10, affecting = groin)
```
MistakeNot4892: As above, not all species have stomachs.

MistakeNot4892:
Do these actually get used anywhere?

comma:
They're definetly not, not since like 2013 or earlier

code/modules/mob/living/carbon/human/MedicalSideEffects.dm
MistakeNot4892: If you just add `else qdel(M)` at line 64 this will be ok.

MistakeNot4892 (changes requested):
Just add the qdel and I'll approve this, it's unused code anyway so improvement is harmless.

MistakeNot4892 (approved):

PR #270, opened 2020-04-09
Alt Chat & Fullscreen Fixes

quardbreak:
I belive I didn't broke anything. Should fix some issues.
Also removes automatical OnResize from fullscreen and alt chat.
Adds fit viewport verb. Works funky with extended client view but work with default value.

### DIFF
--- a/code/modules/client/client_procs.dm
+++ b/code/modules/client/client_procs.dm
@@ -534,3 +534,49 @@ client/verb/character_setup()
 		winset(src, "mainwindow.mainvsplit", "pos=3x0")
 	winset(src, "mainwindow", "is-maximized=true")
 	OnResize()
+
+/client/verb/fit_viewport()
+	set name = "Fit Viewport"
+	set category = "OOC"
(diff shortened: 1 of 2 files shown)


Lohikar:
What exactly is this doing to chat?

quardbreak:
Changes some _alt's macro references to _alt type so _alt buttons will only change _alt ones, not original one, I believe
I should re-test it twice

code/modules/client/client_procs.dm
```diff
@@ -533,4 +534,49 @@ client/verb/character_setup()
 		winset(src, "mainwindow", "menu=menu;statusbar=true")
 		winset(src, "mainwindow.mainvsplit", "pos=3x0")
 	winset(src, "mainwindow", "is-maximized=true")
-	OnResize()
```
MistakeNot4892, line 536: Why remove OnResize()?

quardbreak: I don't really see a reason to have it here :?

MistakeNot4892 (approved):

PR #271, opened 2020-04-09
Fixes secbot and ed209 constructon

quardbreak:
Makes possible use subtypes instead a hardcoded path

### DIFF
--- a/code/modules/crafting/crafting_recipes/bot_crafting/crafting_ed209.dm
+++ b/code/modules/crafting/crafting_recipes/bot_crafting/crafting_ed209.dm
@@ -44,6 +44,9 @@
 	completion_trigger_type = /obj/item/clothing/head/helmet
 	item_icon_state = "ed209_5"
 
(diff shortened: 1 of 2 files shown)


code/modules/crafting/crafting_recipes/bot_crafting/crafting_ed209.dm
```diff
@@ -44,6 +44,9 @@
 	completion_trigger_type = /obj/item/clothing/head/helmet
 	item_icon_state = "ed209_5"
 
+/decl/crafting_stage/ed209_helmet/is_appropriate_tool(obj/item/thing)
+	. = istype(thing) && thing in typesof(completion_trigger_type)
```
afterthought2: `istype(thing, completion_trigger_type)`. In fact I think the operator precedence is wrong here; the && is higher than `in`.

MistakeNot4892 (approved):

PR #272, opened 2020-04-10
Overmap Radio Beacons

comma:
Not actually used anywhere yet but all is working.
Basically just a small overmap object that can attach to another object and move with it, and give a message when scanned via sensors consoles etc.
Added an item for spawning those, but item itself is not accessible ingame.
Could be used for SOS and crap but lazy.
image


code/modules/overmap/radio_beacon.dm
```diff
@@ -0,0 +1,52 @@
...
+	icon = 'icons/obj/radio.dmi'
+	icon_state = "walkietalkie"
+	var/obj/effect/overmap/radio/signal
+
+/obj/item/radio_beacon/attack_self(mob/user)
+	var/obj/effect/overmap/visitable/O = map_sectors["[get_z(src)]"]
+	if(!O)
+		to_chat(user, SPAN_WARNING("You cannot deploy [src] here."))
+		return
+	var/message = sanitize(input("What should it broadcast") as message|null)
```
MistakeNot4892: Needs a question mark.

comma: Added

MistakeNot4892 (approved):

MistakeNot4892:
Only thing I can think of is that you can't turn the beacon off again afterwards.

comma:
It's more or less a debug item anyway, can just del it to clean up the signal

code/modules/overmap/radio_beacon.dm
```diff
@@ -0,0 +1,52 @@
...
+	if(!O)
+		to_chat(user, SPAN_WARNING("You cannot deploy [src] here."))
+		return
+	var/message = sanitize(input("What should it broadcast?") as message|null)
+	
+	if(!signal)
+		signal = new()
+	
+	signal.message = message
+	signal.set_origin(O)
```
afterthought2, line 50: I don't get it; what if I take this obj once deployed and move it to a z level with a different overmap thing associated with it?

comma: Then it doesn't follow. It's just a debug item, it's not very robust

MistakeNot4892: Can be removed if desired, it's not on the map or craftable so doesn't seem like an issue.

PR #276, opened 2020-04-12
Cash and vending machine refactor.

MistakeNot4892:
- Cash appearance is now defined per map.
- Currencies are defined as `/decl/currency` and are at least nominally supported by mechanics.
- Cash in general has been refactored to be less internally bizarre. No more `/bundle`.
- New cash colours are based on the Australian banknote spectrum, no I will not change this.
- Charge cards are now their own object type.
- Material coins now have an associated currency and value (but aren't legal tender except to traders).
- Vendor prices are now generated from item worth using a new repository, which also holds matter info.


comma (dismissed):

MistakeNot4892:
Seems to work ok now.

MistakeNot4892:
This is spiralling out of control. Will try not to add more shit.

MistakeNot4892:
Should be all good. I probably missed a few vendors that shouldn't have prices, but that can be fixed as people report them.

MistakeNot4892:
Ok properly done now.

image

MistakeNot4892:
Ok ok properly done for real.
image

Techhead0:
Pretty sure /obj/machinery/vending/hydronutrients and /obj/machinery/vending/hydroseeds were free before, but if the change is intentional then go ahead

comma (approved):

PR #280, opened 2020-04-12
Applies workaround for instaghosting when joining as AI.

MistakeNot4892:
This started happening with robits shortly after the Initialize() changes went in but I haven't been able to isolate a proper fix.


MistakeNot4892:
Doesn't actually appear to work, on Scav at least.

PR #283, opened 2020-04-12
Ports cloaks

quardbreak:
Ports cloaks from Aurora or Polaris station... don't really remember where I take it.
Allows to band suits and spacesuits with ambards.

There's some problems with clothing which I can't fix:

- [x] Layering issue: accessory and suit layers are different, I need to make cloaks be higher than backpack or etc layer


code/game/objects/items/item_icon_experimental.dm
```diff
@@ -33,7 +33,7 @@
 	return
 
 /obj/item/proc/experimental_mob_overlay(var/mob/user_mob, var/slot)
-	var/bodytype = lowertext(user_mob?.get_bodytype() || BODYTYPE_HUMANOID)
+	var/bodytype = lowertext(user_mob?.get_bodytype()) || BODYTYPE_HUMANOID
```
MistakeNot4892: Need to lowertext() BODYTYPE_HUMANOID, the icon states are all lowertext.

MistakeNot4892 (approved):
I worked on these so I probably shouldn't be approving but I think they're cool.

PR #287, opened 2020-04-13
Ports SSoverlays

quardbreak:
DNM, opening for diff. Rude ported from Europa. Would be nice to get any help with this.

### DIFF
--- a/code/__defines/subsystem-priority.dm
+++ b/code/__defines/subsystem-priority.dm
@@ -6,7 +6,7 @@
 #define SS_PRIORITY_DEFAULT 50         // Default priority for all processes levels
 
 // SS_TICKER
-#define SS_PRIORITY_OVERLAY        200 // Applies overlays. May cause overlay pop-in if it gets behind.
+#define SS_PRIORITY_OVERLAY        100 // Applies overlays. May cause overlay pop-in if it gets behind.
 #define SS_PRIORITY_TIMER          20
 #define SS_PRIORITY_ICON_UPDATE    20  // Queued icon updates. Mostly used by APCs and tables.
 
--- a/code/__defines/subsystem-priority.dm
+++ b/code/__defines/subsystem-priority.dm
@@ -3,11 +3,12 @@
 // SS_BACKGROUND handles high server load differently than Normal and SS_TICKER do.
 // Higher priority also means a larger share of a given tick before sleep checks.
 
-#define SS_PRIORITY_DEFAULT 50          // Default priority for all processes levels
+#define SS_PRIORITY_DEFAULT 50         // Default priority for all processes levels
 
 // SS_TICKER
+#define SS_PRIORITY_OVERLAY        200 // Applies overlays. May cause overlay pop-in if it gets behind.
 #define SS_PRIORITY_TIMER          20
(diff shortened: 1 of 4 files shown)


code/__defines/subsystem-priority.dm
```diff
@@ -3,11 +3,12 @@
 // SS_BACKGROUND handles high server load differently than Normal and SS_TICKER do.
 // Higher priority also means a larger share of a given tick before sleep checks.
 
-#define SS_PRIORITY_DEFAULT 50          // Default priority for all processes levels
+#define SS_PRIORITY_DEFAULT 50         // Default priority for all processes levels
 
 // SS_TICKER
+#define SS_PRIORITY_OVERLAY        200 // Applies overlays. May cause overlay pop-in if it gets behind.
```
afterthought2: I believe this is an unreasonably large number.

code/controllers/subsystems/overlays.dm
```diff
@@ -0,0 +1,223 @@
...
+	var/init_p_len = LAZYLEN(cached_priority)  //starter pokemon
+
+	LAZYREMOVE(cached_overlays, overlays)
+	if(priority)
+		LAZYREMOVE(cached_priority, overlays)
+
+	if(NOT_QUEUED_ALREADY && ((init_o_len != LAZYLEN(cached_priority)) || (init_p_len != LAZYLEN(cached_overlays))))
+		QUEUE_FOR_COMPILE
+
+/atom/proc/add_overlay(list/overlays, priority = FALSE)
```
afterthought2: Confusing arg name; use `_overlays` or something slightly different.

code/controllers/subsystems/overlays.dm
```diff
@@ -0,0 +1,223 @@
...
+
+//TODO: Better solution for these?
+/image/proc/add_overlay(x)
+	overlays += x
+
+/image/proc/cut_overlay(x)
+	overlays -= x
+
+/image/proc/cut_overlays(x)
+	overlays.Cut()
```
afterthought2, line 218: What's the point of these?

code/controllers/subsystems/overlays.dm
```diff
@@ -0,0 +1,223 @@
...
+
+/atom
+	var/tmp/list/our_overlays	//our local copy of (non-priority) overlays without byond magic. Use procs in SSoverlays to manipulate
+	var/tmp/list/priority_overlays	//overlays that should remain on top and not normally removed when using cut_overlay functions, like c4.
```
afterthought2, line 222: But these are not tmp?

Lohikar: They're not meant to be saved anywhere or mapped in, so they're tmp.

code/__defines/subsystem-priority.dm
Lohikar: O7 sets this to 100.

code/controllers/subsystems/overlays.dm
Lohikar: Preferably the latter, params shouldn't be underscore-prefixed.

afterthought2 (approved):
Seems OK to me. Would be good if someone else also reviewed.

MistakeNot4892 (approved):
Seems fine to me as well.

Lohikar (approved):

PR #297, opened 2020-04-14
Animated coin flip

quardbreak:
https://streamable.com/84k8nu

### DIFF
--- a/code/game/objects/items/weapons/material/coins.dm
+++ b/code/game/objects/items/weapons/material/coins.dm
@@ -100,6 +100,7 @@
 		animate(src, transform = flipit, time = 2, easing = QUAD_EASING)
 
 	playsound(src, 'sound/effects/coin_flip.ogg', 75, 1)
+
 	if(user && !thrown)
 		user.visible_message(
 			SPAN_NOTICE("[user] flips \the [src] into the air."),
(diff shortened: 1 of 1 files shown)


MistakeNot4892 (dismissed):

MistakeNot4892:
Code is more or less fine but the actual animation is really quite ugly I think. Not sold on it.

MistakeNot4892 (approved):
Eh why not.

code/game/objects/items/weapons/material/coins.dm
```diff
@@ -58,29 +58,69 @@
...
+	flipit.Scale(0.2, 1)
+	animate(src, transform = flipit, time = 2, easing = QUAD_EASING)
+
+	flipit.Scale(5, 1)
+	if(pick(0, 1))
+		flipit.Invert()
+	flipit.Turn(rand(1, 359))
+	animate(src, transform = flipit, time = 2, easing = QUAD_EASING)
+
+	if (prob(0.1) || rigged)
```
MistakeNot4892, line 97: Doesn't this mean a coin will now always land on the edge forever?

quardbreak: nope, not always. `rigged` turned off by default and needed for special actions like cheating :^)

MistakeNot4892: Oh, this is for a future feature? I didn't see `rigged` was an argument not a member variable.

PR #300, opened 2020-04-15
Removing req_amount from crafting recipes for items and structures.

MistakeNot4892:
All the important parts of the original PR got split off into their own PRs so I'm going to use it for this small one. Removes preset amounts from item/structure recipes so they will autogenerate from matter.

### DIFF
--- a/code/game/objects/items/stacks/rods.dm
+++ b/code/game/objects/items/stacks/rods.dm
@@ -14,7 +14,7 @@
 	max_amount = 100
 	attack_verb = list("hit", "bludgeoned", "whacked")
 	lock_picking_level = 3
-	matter_multiplier = 0.5
+	matter_multiplier = 0.3
 	material_flags = USE_MATERIAL_COLOR
 	stacktype = /obj/item/stack/material/rods
 	material = MAT_STEEL
--- a/code/game/objects/items/stacks/stack.dm
+++ b/code/game/objects/items/stacks/stack.dm
@@ -143,9 +143,9 @@
 			to_chat(user, "<span class='warning'>You waste some [name] and fail to build \the [recipe.display_name()]!</span>")
 			return
 		var/atom/O = recipe.spawn_result(user, user.loc, produced)
-		O.add_fingerprint(user)
-
-		user.put_in_hands(O)
+		if(!QDELETED(O)) // In case of stack merger.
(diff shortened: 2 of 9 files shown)


code/game/machinery/doors/simple.dm
```diff
@@ -73,13 +70,15 @@
 /obj/machinery/door/unpowered/simple/close(var/forced = 0)
 	if(!can_close(forced))
 		return
-	playsound(src.loc, material.dooropen_noise, 100, 1)
+	if(material)
```
comma: not sure if there's any point checking for material considering it frekaing qdels self if it hasn't got one

code/game/machinery/kitchen/smartfridge.dm
```diff
@@ -241,10 +241,11 @@
 	overlays += I
 
 /obj/machinery/smartfridge/dismantle()
-	for(var/datum/stored_items/I in item_records)
-		while(I.amount > 0)
-			I.get_product(get_turf(src)) // They'd get dumped anyway, but this makes things GC properly.
-	..()
+	if(!dismantled)
```
comma: isn't this var set in the base call, why check it first?

code/game/objects/obj_materials.dm
```diff
@@ -0,0 +1,55 @@
...
+		desc = base_desc
+
+/obj/proc/update_material_colour(var/override_colour)
+	var/base_colour = override_colour || initial(color)
+	if(istype(material))
+		color = material.icon_colour
+	else
+		color = base_colour
+
+/obj/on_update_icon()
```
comma: I think a /lot/ of on_update_icon() calls don't call parent, I'm like 40% that not worrying about what parent might have was the point of making it a separate proc. 
Maybe it's something that should go into obj update_icon() call so it's reliably called?

MistakeNot4892: Well it'd give update_icon() an actual purpose at least I guess.

comma:
gib generation results

MistakeNot4892:
[message (29).txt](https://github.com/NebulaSS13/Nebula/files/4519676/message.29.txt)

afterthought2 (dismissed):

afterthought2 (approved):

PR #302, opened 2020-04-15
Adds unique sound for unarmed light strike

quardbreak:
https://streamable.com/97zaca
Sounds from tgstation


code/_global_vars/sound.dm
```diff
@@ -108,3 +108,8 @@ GLOBAL_LIST_INIT(glasscrack_sound,\
 		'sound/effects/glass_crack2.ogg',\
 		'sound/effects/glass_crack3.ogg',\
 		'sound/effects/glass_crack4.ogg'))
+
+GLOBAL_LIST_INIT(light_strike_sound,\
```
MistakeNot4892, line 112: `GLOB` is deprecated, you can just do `var/list/light_strike_sound` and then refer to `global.light_strike_sound` instead of `GLOB`.

MistakeNot4892 (approved):
`GLOB` comment is for future reference, not blocking.

PR #306, opened 2020-04-16
Swapping matter numbers out for defines and a multiplier in Initialize().

MistakeNot4892:
God bless regex.
The actual logic in this PR is on `/obj/Initialize()`, `/obj/item/Initialize()` and `/obj/structure/Initialize()`.


code/game/objects/structures/windoor_assembly.dm
```diff
@@ -25,7 +25,7 @@
 	var/facing_left           //Does the windoor open to the left?
 	var/secure                //Whether or not this creates a secure windoor
 
-/obj/structure/windoor_assembly/Initialize()
+/obj/structure/windoor_assembly/Initialize(var/ml, var/d, )
```
afterthought2: Is this intended?

afterthought2 (approved):

code/game/objects/objs.dm
```diff
@@ -16,11 +16,21 @@
...
+		for(var/mat in matter)
+			matter[mat] = round(matter[mat] * get_matter_amount_modifier())
+	UNSETEMPTY(matter)
+
 /obj/Destroy()
 	STOP_PROCESSING(SSobj, src)
 	return ..()
 
+/obj/proc/get_matter_amount_modifier()
+	. = ceil(w_class * 0.25)
```
Techhead0, line 30: Should this be powers of two instead of linear (like other w_class interactions), or does that break things?

MistakeNot4892: w_class is pretty arbitrary, doesn't really matter. Power of two might end up producing some excessively large values.

PR #309, opened 2020-04-16
Fixes infinite recursion in fingerprint addition

comma:
It was checking for gloves, and if gloves existed it added print to gloves which was checking for gloves, and if gloves existed it added print to gloves which was checking for gloves, and if gloves existed it added print to gloves which was checking for gloves, and if gloves existed it added print to gloves which was checking for gloves, and if gloves existed it added print to gloves
So whole thing crashed silently, preventing any code after fingerprint adding from running.


MistakeNot4892 (dismissed):

MistakeNot4892 (approved):

mattmatterson111:
Oh this is why I couldn't pick up stuff with hands being covered?

comma:
da

PR #311, opened 2020-04-16
Converting simple doors to structures, collapsing damage code into /obj/structure.

MistakeNot4892:
- Simple doors are now a structure.
- Structures now handle health, damage and destruction at a basic level.
- Health for a lot of previously destroyable structures has been boosted as during testing it felt very low.

Upshots of this are basically that you can now bash open locked lockers or crates, and that material doors aren't airlocks anymore.

**TODO**:
- [x] Readd structure-specific damage strings.
- [x] Readd inflatables puncturing.
- [x] Craploads of debugging.


MistakeNot4892:
Ready for review.

code/game/objects/structures/displaycase.dm
```diff
@@ -41,19 +42,16 @@
...
-			playsound(src, "shatter", 70, 1)
-			update_icon()
-	else
-		playsound(src.loc, 'sound/effects/Glasshit.ogg', 75, 1)
+/obj/structure/displaycase/destroyed()
+	if(destroyed)
+		return
+	set_density(0)
+	destroyed = TRUE
+	new /obj/item/material/shard(loc)
```
afterthought2: Should probably reduce `matter` when you perform this kind of action, unless it's fully qdel'd or something?

code/game/objects/structures/doors/_door.dm
```diff
@@ -0,0 +1,195 @@
...
+
+/obj/structure/door/Initialize()
+	. = ..()
+	if(!istype(material))
+		return INITIALIZE_HINT_QDEL
+	if(lock)
+		lock = new /datum/lock(src, lock)
+	if(!icon_base)
+		icon_base = material.door_icon_base
+	update_nearby_tiles(need_rebuild=TRUE)
```
afterthought2, line 26: Sometimes for this type of obj, it's useful to call open() or close() or similar to make sure that the state it's in is actually a valid state, unless all permutations of the state vars are valid. They are sometimes set inconsistently in the map.

code/game/objects/structures/doors/_door.dm
```diff
@@ -0,0 +1,195 @@
...
+	set waitfor = 0
+	if(!can_close())
+		return FALSE
+	flick("[icon_base]closing", src)
+	playsound(src.loc, material.dooropen_noise, 100, 1)
+
+	changing_state = TRUE
+	sleep(1 SECOND)
+	set_density(TRUE)
+	set_opacity(initial(opacity) && material.opacity > 0.5)
```
afterthought2: If I understand the logic correctly, this means that I can make steel doors which are transparent when closed, but not glass doors which are opaque when closed? Seems kind of odd.

code/game/objects/structures/doors/_door.dm
```diff
@@ -0,0 +1,195 @@
...
+			var/obj/item/material/lock_construct/L = I
+			lock = L.create_lock(src,user)
+		return
+
+	if(density)
+		open()
+	else
+		close()
+
+/obj/structure/door/CanPass(atom/movable/mover, turf/target, height=0, air_group=0)
```
afterthought2, line 136: May need some flag set for this to work properly with ZAS; I have no clue about it, though.

MistakeNot4892: It seems to work correctly in testing, and this was copied from inflatables.

code/game/objects/structures/doors/_door.dm
MistakeNot4892: I suppose I can add a flag for non-opaque doors. Nothing will actually use it though.

code/game/objects/structures/doors/_door.dm
MistakeNot4892: The only state var it has is `density` unless you include the 'do not mess with this door' flag. If someone is setting it on the map that should really fail to work so they go and fix it.

code/game/objects/structures/displaycase.dm
MistakeNot4892: I don't believe it is qdel'd, this is code I just moved. I'll modify.

afterthought2 (approved):

PR #312, opened 2020-04-16
[WIP] Proc call mode for buildmode

quardbreak:
Adds new buildmode mode. Allows to click-and-call on objects with pre-set call and arguments.


afterthought2:
I feel that this needs to use code from the VV callproc handler. Allowing calling procs like this is unsafe without blacklists; it's possible (among other things) to bootstrap further admin rights by certain procs, and the checks against that should remain centralized.

MistakeNot4892:
Yeah I am not a big fan of this just because of how many admin I've known who will use a tool like this while having no idea how to do so safely. Doesn't seem safe/necessary.

deathride58:
This happens to be a feature on TG, and it's extremely useful for debugging and adminbus in a live environment, since it allows more fine-grain control over mass proc calling than SDQL does (... speaking of, SDQL here does have the proper wrappers and all, yeah?).
Needs basic safety in the form of a proc call wrapper, though. Using call()() directly without a wrapper is fairly dangerous even for a debug tool.

afterthought2:
I have no problems with this as a feature. It should just use current, existing, VV callproc handling instead of duplicating it but without safety checks.

MistakeNot4892:
Feel free to reopen when ready for review.

PR #314, opened 2020-04-16
Mobile Ladders

quardbreak:
Mobile ladders from Aurora. Can be placed by engineers for temp multiz movement.


code/modules/multiz/mobile_ladder.dm
```diff
@@ -0,0 +1,100 @@
...
+		if (!below_loc || (istype(/turf/space, below_loc)))
+			to_chat(user, SPAN_NOTICE("Why would you do that?! There is only infinite space there..."))
+			return
+		user.visible_message(SPAN_WARNING("[user] begins to lower \the [src] into \the [A]."),
+			SPAN_WARNING("You begin to lower \the [src] into \the [A]."))
+		if (!handle_action(A, user))
+			return
+
+		var/obj/structure/ladder/mobile/downer  = new(below_loc)
+		var/obj/structure/ladder/mobile/current = new(A)
```
comma: pls add fingerprint transfers for changes between forms

code/modules/multiz/mobile_ladder.dm
```diff
@@ -0,0 +1,104 @@
...
+	name = "mobile ladder"
+	desc = "A lightweight deployable ladder, which you can use to move up or down. Or alternatively, you can bash some faces in."
+	icon = 'icons/obj/multiz_items.dmi'
+	icon_state = "mobile_ladder"
+	throw_range = 3
+	force = 10
+	w_class = ITEM_SIZE_LARGE
+	slot_flags = SLOT_BACK
+
+/obj/item/mobile_ladder/proc/place_ladder(atom/A, mob/user)
```
MistakeNot4892, line 11: I haven't tested the code below, but this whole proc could be compressed.

````
	var/above = FALSE
	var/turf/target
	if(isopenspace(A))
		target = GetBelow(A)
		if (!target || (istype(/turf/space, target)))
			to_chat(user, SPAN_WARNING("There is nothing below you."))
			return FALSE
	else if (istype(A, /turf/simulated/floor) || istype(A, /turf/unsimulated/floor))
		target = GetAbove(A)
		if (!target || !isopenspace(target))
			to_chat(user, SPAN_WARNING("There is something above you. You can't deploy \the [src]!"))
			return FALSE
		above = TRUE

	if(above)
		user.visible_message(
			SPAN_NOTICE("\The [user] begins deploying \the [src] on \the [A]."),
			SPAN_NOTICE("You begin to deploy \the [src] on \the [A].")
		)
	else
		user.visible_message(
			SPAN_NOTICE("\The [user] begins to lower \the [src] into \the [A]."),
			SPAN_WARNING("You begin to lower \the [src] into \the [A].")
		)

	if (!handle_action(A, user))
		return

	var/obj/structure/ladder/mobile/ladder = new (target)
	transfer_fingerprints_to(ladder)
	ladder = new(get_turf(A))
	transfer_fingerprints_to(ladder)

	user.drop_from_inventory(src,get_turf(src))
	qdel(src)
````

File: code/modules/multiz/mobile_ladder.dm [hunk 3]
```diff
@@ -0,0 +1,104 @@
...
+	draw_shadow = FALSE
+	tool_interaction_flags = 0
+
+/obj/structure/ladder/mobile/verb/fold()
+	set name = "Fold Ladder"
+	set category = "Object"
+	set src in oview(1)
+
+	if(usr.incapacitated() || !usr.check_dexterity(DEXTERITY_GRIP, silent = TRUE) || !ishuman(usr))
+		to_chat(usr, SPAN_WARNING("You can't do that right now!"))
```
MistakeNot4892: Dexterity should really be separate as it prints it own specific message that is more informative.

code/modules/multiz/mobile_ladder.dm
```diff
@@ -0,0 +1,104 @@
...
+		return
+
+	var/mob/living/carbon/human/H = usr
+	H.visible_message(SPAN_NOTICE("[H] starts folding up [src]."),
+		SPAN_NOTICE("You start folding up [src]."))
+
+	if(!do_after(H, 30, src))
+		to_chat(H, SPAN_WARNING("You are interrupted!"))
```
MistakeNot4892: Not sure if it needs a message, not a lot of other do_afters bother.

code/modules/multiz/mobile_ladder.dm
```diff
@@ -0,0 +1,104 @@
...
+	if(QDELETED(src))
+		return
+
+	var/obj/item/mobile_ladder/R = new(get_turf(H))
+	transfer_fingerprints_to(R)
+
+	H.visible_message(SPAN_NOTICE("[H] folds [src] up into [R]!"),
+		SPAN_NOTICE("You fold [src] up into [R]!"))
+
+	if(target_down)
```
MistakeNot4892: Will need to typecheck target_down/target_up as the ladder code doesn't care about type and will happily connect these ones to existing ladders.

code/modules/multiz/mobile_ladder.dm
```diff
@@ -0,0 +1,104 @@
...
+		current.target_down = downer
+
+		transfer_fingerprints_to(downer)
+		transfer_fingerprints_to(curent)
+		user.drop_from_inventory(src,get_turf(src))
+		qdel(src)
+
+	else if (istype(A, /turf/simulated/floor) || istype(A, /turf/unsimulated/floor))	//Place onto Floor
+		var/turf/upper_loc = GetAbove(A)
+		if (!upper_loc || !isopenspace(upper_loc))
```
MistakeNot4892: Should check for pre-existing ladders here and in the above rather than spawning a double.

code/modules/multiz/mobile_ladder.dm
```diff
@@ -0,0 +1,104 @@
...
+		return
+
+	if(QDELETED(src))
+		return
+
+	var/obj/item/mobile_ladder/R = new(get_turf(H))
+	transfer_fingerprints_to(R)
+
+	H.visible_message(SPAN_NOTICE("[H] folds [src] up into [R]!"),
+		SPAN_NOTICE("You fold [src] up into [R]!"))
```
MistakeNot4892: 'You fold the mobile ladder up into the mobile ladder' is a bit verbose.

code/modules/multiz/mobile_ladder.dm
```diff
@@ -0,0 +1,103 @@
...
+		return FALSE
+	return TRUE
+
+/obj/structure/ladder/mobile
+	base_icon = "mobile_ladder"
+	climb_time = 4 SECONDS
+	draw_shadow = FALSE
+	tool_interaction_flags = 0
+
+/obj/structure/ladder/mobile/verb/fold()
```
MistakeNot4892: Would be nice if `AltClick` also called this verb.

code/modules/multiz/mobile_ladder.dm
```diff
@@ -0,0 +1,103 @@
...
+		SPAN_NOTICE("[H] starts folding up \the [src]."),
+		SPAN_NOTICE("You start folding up \the [src]."))
+
+	if(!do_after(H, 30, src))
+		return
+
+	if(QDELETED(src))
+		return
+
+	var/obj/item/mobile_ladder/R = new(get_turf(H))
```
MistakeNot4892: Throwing a `H.put_in_hands(R)` after this would feel nicer than having it just fall onto the turf.

MistakeNot4892 (commented):
I pulled this down and had a play, other than a couple of minor comments about UX it works fine, happy to approve it if you want to implement the tweaks (or just post to say you don't want to).

code/modules/multiz/mobile_ladder.dm
```diff
@@ -0,0 +1,115 @@
...
+	set src in oview(1)
+	fold(usr)
+
+/obj/structure/ladder/mobile/AltClick(mob/user)
+	fold(user)
+
+/obj/structure/ladder/mobile/proc/fold(mob/user)
+	set name = "Fold Ladder"
+	set category = "Object"
+	set src in oview(1)
```
PsiOmegaDelta: Not sure why these `set`s are made as this proc is never used as a verb directly (it's called via a verb and AltClick but that's different)

code/modules/multiz/mobile_ladder.dm
```diff
@@ -0,0 +1,115 @@
...
+
+	if(user)
+		if(!CanPhysicallyInteract(user) || !ishuman(user))
+			to_chat(user, SPAN_WARNING("You can't do that right now!"))
+			return
+
+		if(!user.check_dexterity(DEXTERITY_GRIP))
+			return
+
+		var/mob/living/carbon/human/H = user
```
PsiOmegaDelta: Why the cast to human? `visible_message()` is a mob-level proc (technically /atom but well..)

code/modules/multiz/mobile_ladder.dm
```diff
@@ -0,0 +1,115 @@
...
+		H.visible_message(
+			SPAN_NOTICE("[H] starts folding up \the [src]."),
+			SPAN_NOTICE("You start folding up \the [src]."))
+
+		if(!do_after(H, 30, src))
+			return
+
+		if(QDELETED(src))
```
PsiOmegaDelta: The `do_after()` above already checks for and fails if `src` is QDELETED

code/modules/multiz/mobile_ladder.dm
quardbreak: Woops forgot to remove.

code/modules/multiz/mobile_ladder.dm
```diff
@@ -0,0 +1,107 @@
...
+/obj/item/mobile_ladder/afterattack(atom/A, mob/user,proximity)
+	if (!proximity)
+		return
+
+	place_ladder(A,user)
+
+/obj/item/mobile_ladder/proc/handle_action(atom/A, mob/user)
+	if (!do_after(user, 30, user))
+		return FALSE
+	if (!A || QDELETED(src) || QDELETED(user))
```
PsiOmegaDelta: `user` is already QDELETED-checked in `do_after()`
I wonder if `A` shouldn't be QDELETED-checked as well (and/or checked before the `do_after()`)

deathride58 (approved):
Deliberate typechecks outside of checks for openspace seems a little messy (is there no isfloor() define or anything like that?), but aside from that, the code looks fine.

PR #316, opened 2020-04-17
Networking overhaul

comma:
Based and heavily copypasted from Exonet by bitMuse. #259 

Makes networks non-global, makes several things respect network availability.
Networks are based on network device extensions, so any object can be hooked up to the network with not that much code. Subtypes implement special behaviour like being a mainframe or a router.
Routers create a network with their ID, or join existing one. If network's router fails, next connected router picks up the slack. If none left, network is gone.
Mainframes expose disk storage to the network, and can have sevral roles like record storage or software repository. Roles are mostly used as filter for programs and network when looking for files, e.g. only downloader iwll only look for programs on repository servers.
Mainframes and routers can have network ID and passkey pre-mapped via their vars, and will init accordingly. If they're not set, router will generate a random ID and no passkey, and mainframe will try to join whatever passwordless network it can find. Most devices try to join any netowkr they can on spawn too.

Silicons changes:
Silicons now spawn with set of stock_parts needed for computing, and an NTOS extension. Since many programs are now netowkr reliant, it was easier to give silicons full on NTOS ability rather than trying to make it work via silicon subsystems. Think of it as inbuilt PDA.
Silicons now also create crew records. AIs create it after picking name, borgs create it after picking a module. It was needed for manifest generation to handle them without special exceptions.

HUDs:
SecHUD and MedHUD are now network reliant. They won't show status or ID icon for sechud if the HUD device's (borg or glasses) network has no crew records on the target. Glasses' network settings can be changed with a verb.
Sadly since mobs generate their own HUD images, the actual status icons are decided via global hidden record storage, and wouldn't care if different networks had different ones.

Crew manifest:
It's now generated based on crew records available on your device's network. 

File browser:
Can now select what are you browsing: HDD, removable disk, or remote server. Can connect to different remove fileservers by picking a network tag. Currently only only pick mainframes with fileserver role enabled.
Added ability to copy between mediums, e.g. HDD->disk or remote->HDD and other combos. Generalized that greatly, by creating datum interfaces for local/network file API and datums for generic transfer tracking.
Removed p2p program in lieu of this.

Programs:
Programs are now stored on the mainframes with software repository role, and downloader app will look for them there. Currently mainframes spawn with an HDD containing all software that was previously available for download.

Antag programs:
Removed DDoS one since can't figure how to make it work with current system.
Since antag programs aren't available for download now, adds a disk containing them to uplink for 5 TCs

Didn't implement some things done in Exonet PR to keep this PR smaller and more review-able:
- Signal simulation. For now just being in connected zlevels with network's router is enough. Speed is fully decided by your network card.
- Access grants system. It's great but can be implemented after since network doesn't really rely on it.

Things left to do:
Make machines buildable.
Look into making apps not die if they can't reach network, just fail gracefully at networking parts.
Testing dunno? I tested it as I went seemed mostly working.

Things I'd probably leave to other PR:
Renaming all stuff to generic-ify names away from NTnet. Want to keep diff readable for this one.


comma:
update:
crew records are now stored on network's record servers
manifests, SEC/MED huds (both glasses and all silicons) use ones on their networks
gun auth is now network-respecting too, secure guns have to be on your net for you to authorize modes
warrants are almost done, split into separate PR
all in all crew records usages are now either OOC (like credit rolls) or going through the appropriate networking
cargo program blocked from being used on networks that do not have supply shuttles going to their (routers) zlevel.

All in all this is about as 'done' as it gets, other important stuff like access system revamp, signal simulation, inter-network interactions and some sort of cybersecurity can be added later
Just waiting on warrants PR, after that will finish em and start removing NTnet and dealing with whatever pops up during that

comma:
Updaed OP, this is more or less done, ready for review

code/modules/mob/living/silicon/robot/modules/_module.dm
```diff
@@ -35,6 +35,7 @@
 	var/list/equipment = list()
 	var/list/synths = list()
 	var/list/skills = list() // Skills that this module grants. Other skills will remain at minimum levels.
+	var/list/software = list() // Apps to preinstall on robot's inbiult computer
```
biscuitWizard, line 38: Typo. Did you mean inbuilt?

code/modules/modular_computers/NTNet/NTNRC/conversation.dm
```diff
@@ -1,24 +1,20 @@
...
-
 /datum/ntnet_conversation/
 	var/id = null
 	var/title = "Untitled Conversation"
 	var/datum/computer_file/program/chatclient/operator // "Administrator" of this channel. Creator starts as channel's operator,
 	var/list/messages = list()
 	var/list/clients = list()
 	var/password
-	var/source_z
+	var/datum/computer_network/network
```
biscuitWizard, line 8: Would this be a better weakref?

code/modules/mob/living/silicon/robot/modules/_module.dm
comma: ye

code/modules/modular_computers/NTNet/NTNRC/conversation.dm
comma: yeah, networks can get deleted after all

code/modules/modular_computers/NTNet/device_types/_network_device.dm
```diff
@@ -0,0 +1,180 @@
...
+
+/datum/extension/network_device/New(datum/holder, n_id, n_key, c_type, autojoin = TRUE)
+	..()
+	network_id = n_id
+	key = n_key
+	if(c_type)
+		connection_type = c_type
+	address = uppertext(NETWORK_MAC)
+	var/obj/O = holder
+	network_tag = "[uppertext(replacetext(O.name, " ", "_"))]-[num2hex(sequential_id(type))]"
```
biscuitWizard, line 19: Not user friendly imo. In my branch I made default network tags procedural. Review that code to see what I meant.

net_tag wasn't ever assigned to because it was an "override" variable, if net_tag was null, no net_tag was assigned, so it always got the most up-to-date procedural tag which was always the device holder's name + the MAC address. So device name changes were reflected instantly.

Only if net_tag was assigned by a user (and thus overriding the procedural net_tag) would it display that, and even still the proc added the mac address onto the end, making sure it was completely and utterly unique no matter what the user set.

code/modules/modular_computers/NTNet/device_types/_network_device.dm
```diff
@@ -0,0 +1,180 @@
...
+	if(check_connection())
+		return GLOB.computer_networks[network_id]
+
+/datum/extension/network_device/proc/add_log(text)
+	var/datum/computer_network/net = get_network()
+	if(!net)
+		return FALSE
+	return net.add_log(text, network_tag)
+
+/datum/extension/network_device/proc/connect_to_any()
```
biscuitWizard, line 86: Connect_to_any shouldn't connect to a network if a key is set on that network, imo.

comma: it won't, it'll fail at the connect()

code/modules/modular_computers/NTNet/NTNRC/conversation.dm
comma: actually made networks qdel their convos when deleted so should be okay now, no lingering references left

code/modules/modular_computers/NTNet/device_types/_network_device.dm
comma: I dunno if instantly checking network tag when machine name changes is desirable since I'm using it as 'address' for remote file browser, kinda weird that I can e.g. drop fileserver users by slapping it with a hand labeler.

code/__defines/computers.dm
```diff
@@ -0,0 +1,50 @@
+#define RANDBYTE  num2hex(rand(1,255))
+#define NETWORK_MAC  "[RANDBYTE]-[RANDBYTE]-[RANDBYTE]-[RANDBYTE]"
```
afterthought2: What if there are clashes?

code/__defines/computers.dm
```diff
@@ -0,0 +1,50 @@
...
+#define NTNET_COMMUNICATION 	2	// Communication (messaging)
+#define NTNET_SYSTEMCONTROL 	4	// Control of various systems, RCon, air alarm control, etc.
+#define NTNET_ALL_FEATURES		(NTNET_SOFTWAREDOWNLOAD|NTNET_COMMUNICATION|NTNET_SYSTEMCONTROL)
+
+// NTNet transfer speeds, used when downloading/uploading a file/program.
+#define NTNETSPEED_LOWSIGNAL 0.5	// GQ/s transfer speed when the device is wirelessly connected and on Low signal
+#define NTNETSPEED_HIGHSIGNAL 1	// GQ/s transfer speed when the device is wirelessly connected and on High signal
+#define NTNETSPEED_ETHERNET 2		// GQ/s transfer speed when the device is using wired connection
+#define NTNETSPEED_DISK 	10		// GQ/s transfer speed when the device is transferring between hard drives
+#define NTNETSPEED_DOS_AMPLIFICATION 5	// Multiplier for Denial of Service program. Resulting load on NTNet relay is this multiplied by NTNETSPEED of the device
```
afterthought2: Thought this was gone.

code/controllers/subsystems/jobs.dm
```diff
@@ -499,6 +487,14 @@ SUBSYSTEM_DEF(jobs)
 			H.buckled.forceMove(H.loc)
 			H.buckled.set_dir(H.dir)
 
+	if(rank != "Robot" && rank != "AI")		//These guys get their emails later.
+		var/domain = "freemail.net"
+		if(H.char_branch?.email_domain)
+			domain = H.char_branch.email_domain
+		var/datum/computer_network/network = get_local_network_at(get_turf(H))
+		if(network)
+			network.create_email(H, H.real_name, domain, rank)
```
afterthought2, line 496: You checked that this works properly with latejoins? I vaguely recall there's some insane logic branching involved between latejoin mob placement and roundstart mob placement.

code/modules/clothing/glasses/glasses.dm
```diff
@@ -237,3 +237,11 @@
 	icon_state = "rwelding-g"
 	item_state = "rwelding-g"
 	tint = TINT_MODERATE
+
+/obj/item/clothing/glasses/proc/network_setup()
+	set name = "Setup HUD Network"
+	set category = "Object"
+	set src in usr
+
+	var/datum/extension/network_device/D = get_extension(hud ? hud : src, /datum/extension/network_device)
```
afterthought2: `hud || src`. Though I must say, weird, weird implementation for hud glasses.

code/modules/mob/living/carbon/human/examine.dm
```diff
@@ -330,16 +334,35 @@
 
 //Helper procedure. Called by /mob/living/carbon/human/examine() and /mob/living/carbon/human/Topic() to determine HUD access to security and medical records.
 /proc/hasHUD(mob/M, hudtype)
+	return !!getHUDsource(M, hudtype)
+
+/proc/getHUDsource(mob/M, hudtype)
```
afterthought2: Feels like the saner pattern is to make this a mob proc.

code/modules/mob/living/carbon/human/examine.dm
```diff
@@ -330,16 +334,35 @@
...
+/proc/getHUDnetwork(mob/M, hudtype)
+	var/obj/O = getHUDsource(M, hudtype)
+	if(!O)
+		return
+	if(istype(M, /mob/living/carbon/human))
+		var/datum/extension/network_device/D = get_extension(O, /datum/extension/network_device)
+		return D.get_network()
+	else if(istype(M, /mob/living/silicon))
+		var/mob/living/silicon/S = M
+		return S.get_computer_network()
```
afterthought2: Feel like these should be unified, possibly by having the above proc return the borg itself instead of the borg sight. If really not viable, could at least be a mob proc.

code/modules/modular_computers/NTNet/_network.dm
```diff
@@ -0,0 +1,142 @@
...
+	return ARE_Z_CONNECTED(get_z(router.holder), get_z(D.holder))
+
+/datum/computer_network/proc/get_signal_strength(datum/extension/network_device/D)
+	if(check_connection(D))
+		return D.connection_type
+
+/datum/computer_network/proc/get_device_by_tag(nettag)
+	for(var/datum/extension/network_device/D in devices)
+		if(D.network_tag == nettag)
+			return D
```
afterthought2: Feels like this is worth optimizing by having a tag -> devices list. Also from what I can tell so far, there's nothing to prevent multiple devices from sharing a tag; this is intentional?

code/modules/modular_computers/NTNet/_network.dm
```diff
@@ -0,0 +1,142 @@
...
+		
+	add_log("Mainframe roles updated, now: [english_list(M.roles)]", M.network_tag)
+
+/datum/computer_network/proc/get_os_by_nid(nid)
+	for(var/datum/extension/network_device/D in devices)
+		if(D.address == uppertext(nid))
+			var/datum/extension/interactive/ntos/os = get_extension(D.holder, /datum/extension/interactive/ntos)
+			if(!os)
+				var/atom/A = D.holder
+				os = get_extension(A.loc, /datum/extension/interactive/ntos)
```
afterthought2, line 141: Seems questionable; if this is some borg bullshit, may be worth abstracting to a proc on the D extension, which should know where to look for the os extension.

code/modules/modular_computers/NTNet/device_types/_network_device.dm
```diff
@@ -0,0 +1,180 @@
+/datum/extension/network_device
+	base_type = /datum/extension/network_device
+	expected_type = /atom/movable
+	flags = EXTENSION_FLAG_IMMEDIATE
+	var/network_id
+	var/key
+	var/address
+	var/network_tag
```
afterthought2: Please add some comments explaining what these mean; even if I can figure them out now, in a year both of us will have forgotten.

code/modules/modular_computers/NTNet/device_types/_network_device.dm
```diff
@@ -0,0 +1,180 @@
...
+	return net.get_signal_strength(src)
+	
+/datum/extension/network_device/proc/get_signal_wordlevel()
+	var/datum/computer_network/network = get_network()
+	if(!network)
+		return "Not Connected"
+	var/signal_strength = network.get_signal_strength(src)
+	if(signal_strength <= 0)
+		return "Not Connected"
+	else if(signal_strength <= 6)
```
afterthought2: Wait, how does it get up to 6? It seems like it's between 0 and 2. In any case, best to avoid the magic number.

code/modules/modular_computers/NTNet/device_types/_network_device.dm
```diff
@@ -0,0 +1,180 @@
...
+/datum/extension/network_device/proc/connect_to_any()
+	var/list/nets = get_nearby_networks()
+	for(var/net in nets)
+		network_id = net
+		if(connect())
+			return
+
+/datum/extension/network_device/proc/do_change_id(var/mob/user)
+	var/new_id = sanitize(input(user, "Enter network ID:", "Network ID", network_id) as text|null)
+	if(!new_id)
```
afterthought2: This probably needs interactivity checks after the input, though I have no idea what interactivity checks. May be best to handle the `input` on the holder. Similar for the other procs below.

code/modules/modular_computers/NTNet/device_types/_network_device.dm
```diff
@@ -0,0 +1,180 @@
...
+	ui = SSnano.try_update_ui(user, src, ui_key, ui, data, force_open)
+	if (!ui)
+		var/atom/A = holder
+		ui = new(user, src, ui_key, "network_machine_settings.tmpl", capitalize(A.name), 380, 500)
+		ui.set_initial_data(data)
+		ui.open()
+		ui.set_auto_update(1)
+	
+/datum/extension/network_device/Topic(href, href_list)
+	. = ..()
```
afterthought2, line 174: Must check `.` here. You may also want to throw in a CanUseTopic call to holder here, though you'll have to check which topic state is reasonable. This ties in with the previous comment. There's an extension subtype (`interactive` I believe) which handles topic stuff for you that you may want to look over.

code/modules/modular_computers/NTNet/device_types/router.dm
```diff
@@ -0,0 +1,44 @@
...
+		if(net.router == src)
+			to_chat(user, SPAN_NOTICE("Hosting the network '[network_id]'."))
+		else
+			to_chat(user, SPAN_NOTICE("Broadcasting on the network '[network_id]'."))
+
+/datum/extension/network_device/router/set_new_key(new_key, user)
+	if(is_router())
+		var/datum/computer_network/net = GLOB.computer_networks[network_id]
+		key = new_key
+		net.network_key = new_key
```
afterthought2, line 41: Isn't there some issue with doing this where the stuff already on the network isn't grandfathered in properly, possibly resulting in a state where members don't have the proper key?

code/modules/modular_computers/NTNet/machinery/_network_machine.dm
```diff
@@ -0,0 +1,55 @@
...
+	var/network_device_type =  /datum/extension/network_device
+	var/error
+	
+	var/initial_network_id
+	var/initial_network_key
+	var/lateload
+
+/obj/machinery/network/Initialize()
+	. = ..()
+	set_extension(src, network_device_type, initial_network_id, initial_network_key, 2, !lateload)
```
afterthought2: The 2 should be a define, right?

code/modules/modular_computers/file_system/programs/generic/reports.dm
```diff
@@ -170,7 +171,8 @@
 		return 1
 	if(href_list["get_report"])
 		var/uid = text2num(href_list["report"])
-		for(var/datum/computer_file/report/report in ntnet_global.fetch_reports(get_access(user)))
+		var/datum/computer_network/net = program.computer.get_network()
+		for(var/datum/computer_file/report/report in net.fetch_reports(get_access(user)))
```
afterthought2, line 176: This file doesn't seem to be doing the `if(net)` thing that all the others do.

code/modules/modular_computers/ntos/components.dm
```diff
@@ -33,6 +33,17 @@
...
+/datum/extension/interactive/ntos/proc/get_network_speed(var/specific_action = 0)
+	switch(get_ntnet_status(NTNET_SOFTWAREDOWNLOAD))
+		if(0)
+			return 0
+		if(1)
+			return NTNETSPEED_LOWSIGNAL
+		if(2)
+			return NTNETSPEED_HIGHSIGNAL
+		if(3)
+			return NTNETSPEED_ETHERNET
```
afterthought2, line 45: Is there a reason it doesn't just always use speed defines? I think I made a comment above based on a misunderstanding of the relationship between these numbers and these defines.

code/modules/modular_computers/ntos/files.dm
```diff
@@ -87,4 +87,190 @@
...
 	var/datum/computer_file/C = F.clone(0)
-	return disk_to.store_file(C)
\ No newline at end of file
+	return disk_to.store_file(C)
+
+// Generic file storage interface
+/datum/file_storage
+	var/name = "Generic File Interface"
+	var/datum/extension/interactive/ntos/os
+/datum/file_storage/New(ntos)
```
afterthought2, line 97: What's with the weird whitespace use?

maps/bearcat/bearcat_define.dm
```diff
@@ -20,7 +20,7 @@
 	allowed_spawns = list("Cryogenic Storage")
 	default_spawn = "Cryogenic Storage"
 	use_overmap = 1
-	num_exoplanets = 2
+	num_exoplanets = 0
```
afterthought2: Debugging code left in?

afterthought2 (commented):
Looks OK for the most part, but there are a lot of non-localized changes via proc argument additions and such which I'm assuming you've systematically searched for.

For the Topic stuff, I'm guessing that because you've set nano_host properly, you can just check CanInteract(user, GLOB.default_state), but worth checking that it properly works with tablets and such.

code/modules/modular_computers/NTNet/device_types/router.dm
comma: That's kinda intended, if painful way, I mean why change password in the first place if whoever got it will get updated version.

code/modules/modular_computers/NTNet/_network.dm
comma: It's not just borgs, computers too, since extension lives on the network card

code/modules/clothing/glasses/glasses.dm
comma: don

code/modules/modular_computers/NTNet/_network.dm
comma: Added such list and also networks now will enforce unique tags, modifying device's tag if needed.

code/modules/modular_computers/ntos/components.dm
comma: remains of / setup for signal strength stuff, connection type is used as sole factor because im too lazy to add that now.

code/__defines/computers.dm
comma: I'm willing to take that risk.
TBH these are only used by terminals, and for banning, most of interactions use network tags as 'address'

code/controllers/subsystems/jobs.dm
comma: Yep it works

code/modules/modular_computers/ntos/files.dm
comma: it's a bit of a haze, but i tihnk i didn't like how there's so many empty functions
does look nicer with whitespaces tho, changed

maps/bearcat/bearcat_define.dm
comma: yeah, gonna remove in the end, but it does speed up the init a lot

code/modules/mob/living/carbon/human/examine.dm
comma: If borg is returned would still need special handing cause network device is not on borg but on its network card. Made it mob proc with override by silicons

code/modules/mob/living/carbon/human/examine.dm
comma: yeah did that

code/modules/modular_computers/NTNet/device_types/_network_device.dm
comma: Repathed to interactive and done checks

code/modules/modular_computers/NTNet/device_types/_network_device.dm
comma: Added the interactivity check.

code/modules/modular_computers/NTNet/device_types/_network_device.dm
comma: nevermind running into some issues re

code/modules/modular_computers/NTNet/device_types/_network_device.dm
comma: nevermind running into some issues re

comma: was caused by stuff fixed in #346 good now

code/modules/modular_computers/NTNet/device_types/_network_device.dm
comma: fixed

afterthought2 (approved):

PR #318, opened 2020-04-17
Adds looping sound for shower

quardbreak:
as always thanks tgstation for these sounds


code/game/objects/structures/watercloset.dm
```diff
@@ -223,15 +223,47 @@
...
 
+	var/sound_id
+	var/datum/sound_token/sound_token
+
+//add heat controls? when emagged, you can freeze to death in it?
+
 /obj/structure/hygiene/shower/Initialize()
 	. = ..()
 	create_reagents(50)
+	sound_id = "[type]_[sequential_id(type)]"
```
PsiOmegaDelta: As I keep repeating, sound ids should be re-used to the greatest possible degree because the limit is 1024 unique ids.
`sound_id = /obj/structure/hygiene/shower` should suffice.

comma (approved):

PR #320, opened 2020-04-17
Inertial damper

quardbreak:
Ports https://github.com/Baystation12/Baystation12/pull/27899 with small fix and improvement.
By default, all overmap ships don't require inertial damper. It can be set by var to enable damper requirement.
Credits to mikomyazaki for feature and help to fix bound problem.

image


MistakeNot4892:
80% sure this sprite is from the original /tg/ gravity generator, not Eris.

code/_onclick/adjacent.dm
```diff
@@ -26,36 +26,44 @@
...
 */
 /turf/Adjacent(var/atom/neighbor, var/atom/target = null)
-	var/turf/T0 = get_turf(neighbor)
-	if(T0 == src)
-		return 1
-	if(!T0 || T0.z != z)
-		return 0
-	if(get_dist(src,T0) > 1)
-		return 0
+	var/list/turf/Ts = get_turf(neighbor)
```
MistakeNot4892, line 29: the `/turf` here doesn't do anything as far as I know.

MistakeNot4892: Also for people who get confused like me, this proc is nearly identical to the old one, just tabbed in one step and with a `return TRUE` condensed.

code/_onclick/adjacent.dm
```diff
@@ -84,7 +92,6 @@ Quick adjacency (to turf):
 	if(neighbor == loc || (neighbor.loc == loc)) return 1
 	if(!isturf(loc)) return 0
 	for(var/turf/T in locs)
-		if(isnull(T)) continue
```
MistakeNot4892, line 87: Any confidence in this being unneeded? Could do `if(T &&` below instead.

code/game/area/areas.dm
```diff
@@ -281,6 +281,24 @@ var/list/mob/living/forced_ambiance_list = new
 				H.AdjustWeakened(3)
 			to_chat(mob, "<span class='notice'>The sudden appearance of gravity makes you fall to the floor!</span>")
 
+/area/proc/throw_unbuckled_occupants(var/maxrange, var/speed, var/direction = null)
+	for(var/mob/M in src)
```
MistakeNot4892, line 285: This feels dangerous. Isn't there some consideration with looping on area contents weirdly expensive?

code/modules/events/inertial_damper.dm
```diff
@@ -0,0 +1,29 @@
...
+		I.damping_modifier += -5 //Gm/h
+		I.was_reset = FALSE
+
+/datum/event/inertial_damper/end()
+	var/display_announcement = FALSE
+	for(var/obj/machinery/inertial_damper/I in SSmachines.machinery)
+		I.damping_modifier = initial(I.damping_modifier)
+		if(!I.was_reset)
+			display_announcement = TRUE
+
```
MistakeNot4892, line 28: Could break here.

code/modules/inertial_damper/inertial_damper.dm
```diff
@@ -0,0 +1,204 @@
...
+	var/name = "ship inertial damper"
+	var/obj/machinery/holder
+
+/datum/ship_inertial_damper/proc/get_damping_strength(var/reliable)
+	return 0
+
+/datum/ship_inertial_damper/New(var/obj/machinery/_holder)
+	..()
+	holder = _holder
+	ship_inertial_dampers += src
```
MistakeNot4892: `global.ship_inertial_dampers`.

MistakeNot4892: Likewise elsewhere.

code/modules/inertial_damper/inertial_damper.dm
```diff
@@ -0,0 +1,204 @@
...
+			GLOB.global_announcer.autosay("WARNING: Inertial dampening level dangerously low! All crew must be secured before firing thrusters!", "inertial damper Monitor")
+	else
+		delta = initial(delta) * 5 // rate of dampening strength decay is higher if we have no power
+		target_strength = 0
+		if(!damping_strength)
+			damping_modifier = initial(damping_modifier)
+			was_reset = TRUE
+		change_power_consumption(0, POWER_USE_OFF)
+
+	var/overlay_state = null
```
MistakeNot4892: All of this overlay logic should be in `on_update_icon`.

code/modules/inertial_damper/inertial_damper.dm
```diff
@@ -0,0 +1,204 @@
...
+
+	if(href_list["toggle"])
+		toggle()
+		return TOPIC_REFRESH
+
+	if(href_list["set_strength"])
+		var/new_strength = input("Enter a new damper strength between 0 and [max_strength] Gm/h", "Modifying damper strength", get_damping_strength(TRUE)) as num
+		if(!(new_strength in 0 to max_strength))
+			to_chat(user, SPAN_WARNING("That's not a valid damper strength."))
+			warned = FALSE
```
MistakeNot4892, line 179: Shouldn't this be set on success as well?

code/modules/overmap/ships/ship.dm
```diff
@@ -100,6 +104,13 @@ var/const/OVERMAP_SPEED_CONSTANT = (1 SECOND)
 /obj/effect/overmap/visitable/ship/proc/adjust_speed(n_x, n_y)
 	CHANGE_SPEED_BY(speed[1], n_x)
 	CHANGE_SPEED_BY(speed[2], n_y)
+	var/magnitude = norm(n_x, n_y)
+	var/inertia_dir = magnitude >= 0 ? turn(fore_dir, 180) : fore_dir
+	var/inertia_strength = magnitude * 1e3
+	if(needs_dampers && damping_strength < inertia_strength)
+		for(var/area/A in world)
```
MistakeNot4892: `*scream`
Isn't there an area list somewhere?

afterthought2: `area_repository` has a bunch of helpers for quickly finding areas matching criteria.

MistakeNot4892:
Note to self to pull this down and test, yell at me if I haven't commented again in a day or so.

code/_onclick/adjacent.dm
comma: locs is magical inbuilt BYOND list, and we do check loc to be turf before, so this should be always turfs

MistakeNot4892:
Pulled this down and tested it, all works fine except that I can interact with the damper from 2 turfs away on the northeast corner. I assume it is because you set bounds but also offset the icon.

It would also be nice if turning the thing off and on had some noises to go with it.

MistakeNot4892 (approved):

PR #325, opened 2020-04-18
Nerfs reagent values into the ground.

MistakeNot4892:
They are producing some absolutely mental monetary worth values in the cargo PR.


MistakeNot4892:
Needs some work.

afterthought2 (approved):

PR #327, opened 2020-04-18
Moved cortical borers into a modpack.

MistakeNot4892:
Also added some symbiote job slot stuff from an old Bay PR.

Borers in general are quite under-maintained, I'm going to rewrite them soon I think, this PR is to make that easier.


mods/borers/mob/overrides.dm
```diff
@@ -0,0 +1,58 @@
+#define HAS_BRAIN_WORMS(HOST) locate(/mob/living/simple_animal/borer) in HOST.contents
```
Techhead0: Wrap in (), just in case, due to the `in` operator's low precedence

code/_helpers/medical_scans.dm
```diff
@@ -70,6 +69,7 @@
 		O["brute_dam"] = E.brute_dam
 		O["burn_dam"] = E.burn_dam
 		O["scan_results"] = E.get_scan_results(tag)
+		O["tumors"] = E.has_growths()
```
Techhead0, line 72: Looking forward to seeing Chinsky add Space Cancer

comma: It's been on my todo for ages!

mods/psionics/system/psionics/mob/borer_power.dm
```diff
@@ -2,6 +2,11 @@
 	var/image/aura_image
```
comma, line 3: Maybe this file should be included if borer mod is included only

MistakeNot4892: I can't see a way to reliably do that since the include order is not predictable. :(

comma (approved):

PR #330, opened 2020-04-19
Added some butchery calc to mob worth.

MistakeNot4892:


MistakeNot4892:
Merging this in with the Cargo PR.

PR #335, opened 2020-04-19
Z-Mimic Fixes Pack 2

Lohikar:
Reworks layering a bit so it actually works. I couldn't get large turfs to play nice, so they get a flag that changes (read: breaks) the layering so they don't look bad.

~~DNM: Untested. It compiles and the O7 ver works, that's all I know.~~ Seems to work.

### DIFF
--- a/code/controllers/subsystems/zcopy.dm
+++ b/code/controllers/subsystems/zcopy.dm
@@ -17,6 +17,10 @@ SUBSYSTEM_DEF(zcopy)
 	var/openspace_overlays = 0
 	var/openspace_turfs = 0
 
+	// Highest Z level in a given Z-group for absolute layering used by FIX_BIGTURF.
+	// zstm[zlev] = group_max
+	var/list/zstack_maximums = list()
+
 // for admin proc-call
 /datum/controller/subsystem/zcopy/proc/update_all()
 	disable()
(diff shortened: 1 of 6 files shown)


code/controllers/subsystems/zcopy.dm
```diff
@@ -77,10 +81,23 @@ SUBSYSTEM_DEF(zcopy)
 	..("Q:{T:[queued_turfs.len - (qt_idex - 1)]|O:[queued_overlays.len - (qo_idex - 1)]} T:{T:[openspace_turfs]|O:[openspace_overlays]}")
 
 /datum/controller/subsystem/zcopy/Initialize(timeofday)
+	calculate_zstack_limits()
 	// Flush the queue.
 	fire(FALSE, TRUE)
 	return ..()
 
+// If you add a new Zlevel or change Z-connections, call this.
+/datum/controller/subsystem/zcopy/proc/calculate_zstack_limits()
```
Lohikar, line 90: Not familiar with away site loading et-al, so not sure where this should get called.

code/controllers/subsystems/zcopy.dm
```diff
@@ -127,28 +142,35 @@ SUBSYSTEM_DEF(zcopy)
...
+				if (!T.below.bound_overlay)
+					T.below.bound_overlay = new(T)
+				var/atom/movable/openspace/turf_overlay/TO = T.below.bound_overlay
+				TO.appearance = T.below
+				TO.name = T.name
+				TO.gender = T.gender	// Need to grab this too so PLURAL works properly in examine.
+				TO.opacity = FALSE
+				TO.plane = t_target
+		else if (T.below.bound_overlay)
+			QDEL_NULL(T.below.bound_overlay)
```
afterthought2, line 177: Maybe cleaner to do
```
if(T.z_flags & ZM_MIMIC_OVERWRITE)
    // This openturf doesn't care about its icon, so we can just overwrite it.
    etc
else if(T.below.z_eventually_space)
    QDEL_NULL(T.below.bound_overlay)
else
    // Some openturfs have icons, so we can't overwrite their appearance.
    etc
```
though I could be misreading the logic branching here.

code/controllers/subsystems/zcopy.dm
afterthought2: Seems like they now go through `INCREMENT_WORLD_Z_SIZE`, so at least not hard to find. I'm not super clear on the ordering requirements for this; the connectivity is set in landmark/map_data/New, but if maps are being loaded via dmm_suite and queuing zmimic updates and then sleeping, there's no guarantee that will have been created before things start being queued here, unless you're already handling that. If you are, then can e.g. call this from both that macro and from `map_data/New`.

code/controllers/subsystems/zcopy.dm
Lohikar: That should work too; this is structured like this as the code it was extracted from had another branch above the bound_overlay check.

code/controllers/subsystems/zcopy.dm
Lohikar: This should only matter for things mutating the world size after Z-Mimic init, if away sites are loaded earlier than that they're probably fine.

afterthought2: Well, away sites (including multiz ones) can be loaded via admin verb. There are a couple of other things that add blank z levels.

afterthought2 (commented):
As discussed on discord, should refresh the `zstack` stuff in `New` of the map_data landmark. Otherwise looks good.

afterthought2 (approved):

PR #340, opened 2020-04-20
Generalizing/removing old SS13 jobs.

MistakeNot4892:
Splitting this off the Exodus PR. These jobs are only ever used for access inheritance and populating the example map job list. Map already have to override antags etc to add their own roles, might as well lop off the outdated ones.


afterthought2 (approved):

code/game/gamemodes/objectives/objective_heist.dm
```diff
@@ -11,10 +11,11 @@
 	for(var/datum/mind/possible_target in SSticker.minds)
 		if(possible_target != owner && ishuman(possible_target.current) && (possible_target.current.stat != DEAD) && (!possible_target.special_role))
 			possible_targets += possible_target
-			for(var/datum/job/role in SSjobs.get_by_path(roles))
-				if(possible_target.assigned_role == role.title)
-					priority_targets += possible_target
-					continue
+			if(length(roles))
+				for(var/datum/job/role in SSjobs.get_by_path(roles))
+					if(possible_target.assigned_role == role.title)
```
comma, line 16: `possible_target.assigned_role == /datum/job/chief_engineer.title`
needs initial maybe?

MistakeNot4892: `get_by_path` returns a datum.

comma: derp didn't see that

PR #341, opened 2020-04-20
Misc fixes

quardbreak:
Port some fixes from other stations ~~DMN till don't test everything~~

### DIFF
--- a/code/game/antagonist/antagonist_helpers.dm
+++ b/code/game/antagonist/antagonist_helpers.dm
@@ -1,6 +1,8 @@
 /datum/antagonist/proc/can_become_antag(var/datum/mind/player, var/ignore_role)
 
 	if(player.current)
+		if(isliving(player.current) && player.current.stat)
+			return 0
 		if(jobban_isbanned(player.current, id))
 			return 0
 		if(player.current.faction != MOB_FACTION_NEUTRAL)
--- a/code/modules/mob/observer/ghost/ghost.dm
+++ b/code/modules/mob/observer/ghost/ghost.dm
@@ -515,7 +515,7 @@ This is the proc mobs get to turn into a ghost. Forked from ghostize due to comp
 /mob/observer/ghost/MayRespawn(var/feedback = 0, var/respawn_time = 0)
 	if(!client)
 		return 0
-	if(mind?.current && (mind.current in GLOB.living_mob_list_) && (can_reenter_corpse in list(CORPSE_CAN_REENTER, CORPSE_CAN_REENTER_AND_RESPAWN)))
+	if(mind?.current && mind.current.stat != DEAD && (can_reenter_corpse in list(CORPSE_CAN_REENTER, CORPSE_CAN_REENTER_AND_RESPAWN)))
 		if(feedback)
 			to_chat(src, "<span class='warning'>Your non-dead body prevents you from respawning.</span>")
 		return 0
--- a/code/modules/mob/observer/ghost/ghost.dm
+++ b/code/modules/mob/observer/ghost/ghost.dm
@@ -515,7 +515,7 @@ This is the proc mobs get to turn into a ghost. Forked from ghostize due to comp
 /mob/observer/ghost/MayRespawn(var/feedback = 0, var/respawn_time = 0)
 	if(!client)
 		return 0
-	if(mind && mind.current && mind.current.stat != DEAD && can_reenter_corpse == CORPSE_CAN_REENTER)
+	if(mind?.current && (mind.current in GLOB.living_mob_list_) && (can_reenter_corpse in list(CORPSE_CAN_REENTER, CORPSE_CAN_REENTER_AND_RESPAWN)))
 		if(feedback)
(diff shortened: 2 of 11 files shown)


code/modules/mob/observer/ghost/ghost.dm
```diff
@@ -515,7 +515,7 @@ This is the proc mobs get to turn into a ghost. Forked from ghostize due to comp
 /mob/observer/ghost/MayRespawn(var/feedback = 0, var/respawn_time = 0)
 	if(!client)
 		return 0
-	if(mind && mind.current && mind.current.stat != DEAD && can_reenter_corpse == CORPSE_CAN_REENTER)
+	if(mind?.current && (mind.current in GLOB.living_mob_list_) && (can_reenter_corpse in list(CORPSE_CAN_REENTER, CORPSE_CAN_REENTER_AND_RESPAWN)))
```
afterthought2: I get the second part, but why `(mind.current in GLOB.living_mob_list_)` instead of the stat check?

code/game/antagonist/antagonist_helpers.dm
```diff
@@ -1,6 +1,8 @@
 /datum/antagonist/proc/can_become_antag(var/datum/mind/player, var/ignore_role)
 
 	if(player.current)
+		if(isliving(player.current) && player.current.stat)
+			return 0
```
afterthought2, line 5: Should probably check for the override mob flag in here, as antags which override mobs may as well pull from any kind of mob.

code/modules/xenoarcheaology/tools/ano_device_battery.dm
```diff
@@ -156,6 +156,8 @@
 	else if(href_list["shutdown"])
 		shutdown_emission()
 		. = TOPIC_REFRESH
+	else if(href_list["refresh"])
+		interact(user)
```
afterthought2: This is using nanoui, so should return `TOPIC_REFRESH` instead.

code/modules/security levels/keycard_authentication.dm
```diff
@@ -30,15 +31,15 @@
...
+				if(event_source && initial_card != ID)
 					event_source.confirmed = 1
-					event_source.event_confirmed_by = usr
+					event_source.event_confirmed_by = user
 				else
 					to_chat(user, "<span class='warning'>Unable to confirm, DNA matches that of origin.</span>")
 			else if(screen == 2)
-				event_triggered_by = usr
+				event_triggered_by = user
+				initial_card = ID
```
afterthought2: Should either use a weakref or register for the deletion events.

code/modules/security levels/keycard_authentication.dm
```diff
@@ -3,6 +3,12 @@
 	desc = "This device is used to trigger functions which require more than one ID card to authenticate."
 	icon = 'icons/obj/monitors.dmi'
 	icon_state = "auth_off"
+
+	anchored = 1.0
```
afterthought2: `1`, but I believe that's the default value anyway.

code/modules/mob/observer/ghost/ghost.dm
quardbreak: Because body can be destroyed/gibbed, we checking is it still persist in living list.

afterthought2: `mind.current` should never be a `QDELETED` mob, and any non-QDELETED living mob will be in the living mob list. Therefore if `(mind.current in GLOB.living_mob_list_)` is false, so is `mind?.current`. So this effectively just removes the `mind.current.stat != DEAD` check.

code/modules/security levels/keycard_authentication.dm
```diff
@@ -116,33 +120,39 @@
...
 /obj/machinery/keycard_auth/proc/broadcast_request()
 	icon_state = "auth_on"
 	for(var/obj/machinery/keycard_auth/KA in world)
-		if(KA == src) continue
+		if(KA == src)
+			continue
 		KA.reset()
-		spawn()
-			KA.receive_request(src)
+		KA.receive_request(src, initial_card)
```
afterthought2: This is sending a weakref, but that proc seems to want an actual ID.

code/modules/security levels/keycard_authentication.dm
```diff
@@ -116,33 +120,39 @@
...
+	initial_card = null
 
 /obj/machinery/keycard_auth/proc/broadcast_request()
 	icon_state = "auth_on"
 	for(var/obj/machinery/keycard_auth/KA in world)
-		if(KA == src) continue
+		if(KA == src)
+			continue
 		KA.reset()
-		spawn()
```
afterthought2, line 125: Removing this spawn doesn't work, as the called proc sleeps.

code/modules/xenoarcheaology/tools/ano_device_battery.dm
```diff
@@ -156,6 +156,9 @@
...
```
afterthought2: Returning `TOPIC_REFRESH` should be instead of calling `interact`. Calling `interact` does nothing on this machine.

afterthought2 (approved):

PR #342, opened 2020-04-20
Attempting to optimize fluids a bit.

MistakeNot4892:
- Fluids will stop processing if their fluids do not change during their process iteration.
- Fluids keep track of neighbors automatically via Initialize() and Destroy().
- Fluids will not equalize if a minimum delta is not reached with at least one neighbor.
- Each edge turf will only be fully checked once for spreading.
- Various other tweaks.

### DIFF
--- a/code/controllers/subsystems/fluids.dm
+++ b/code/controllers/subsystems/fluids.dm
@@ -18,8 +18,6 @@ var/datum/controller/subsystem/fluids/SSfluids
 
 	var/tmp/active_fluids_copied_yet = FALSE
 	var/af_index = 1
-	var/downward_fluid_overlay_position = 1 // Bit of an odd hack, set in fluid spread code to determine which overlay
-	                                        // in the list is 'down'. More maintainer-friendly than hardcoding it.
 	var/list/fluid_images = list()
 
 	var/list/gurgles = list(
@@ -41,14 +39,34 @@ var/datum/controller/subsystem/fluids/SSfluids
 		active_fluids_copied_yet = FALSE
 		af_index = 1
 
-	var/flooded_a_neighbor // Not used, required by FLOOD_TURF_NEIGHBORS.
 	var/list/curr_sources = processing_sources
+	var/list/checked = list()
 	while (curr_sources.len)
-		var/turf/T = curr_sources[curr_sources.len]
 		curr_sources.len--
-
-		FLOOD_TURF_NEIGHBORS(T, FALSE)
-
+		var/flooded_a_neighbor
+		var/turf/T = curr_sources[curr_sources.len]
+		UPDATE_FLUID_BLOCKED_DIRS(T)
+		for(var/spread_dir in GLOB.cardinal)
+			if(T.fluid_blocked_dirs & spread_dir) 
+				continue
+			var/turf/next = get_step(T, spread_dir)
+			if(!istype(next) || next.flooded) 
+				continue
+			UPDATE_FLUID_BLOCKED_DIRS(next)
+			if((next.fluid_blocked_dirs & GLOB.reverse_dir[spread_dir]) || !next.CanFluidPass(spread_dir) || checked[next])
+				continue
+			checked[next] = TRUE
+			flooded_a_neighbor = TRUE
+			var/obj/effect/fluid/F = locate() in next
+			if(!F)
+				F = new /obj/effect/fluid(next)
+				var/datum/gas_mixture/GM = T.return_air()
+				if(GM) F.temperature = GM.temperature
+			if(F)
+				if(F.fluid_amount < FLUID_MAX_DEPTH)
+					SET_FLUID_DEPTH(F, FLUID_MAX_DEPTH)
+		if(!flooded_a_neighbor)
+			REMOVE_ACTIVE_FLUID_SOURCE(T)
 		if (MC_TICK_CHECK)
 			return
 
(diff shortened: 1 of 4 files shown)


MistakeNot4892:
Not sure how to profile this. It does properly sleep when the change delta isn't hit, so it's an improvement on existing code.

code/controllers/subsystems/fluids.dm
```diff
@@ -41,14 +39,36 @@ var/datum/controller/subsystem/fluids/SSfluids
...
+	var/list/checked = list()
 	while (curr_sources.len)
 		var/turf/T = curr_sources[curr_sources.len]
 		curr_sources.len--
-
-		FLOOD_TURF_NEIGHBORS(T, FALSE)
-
+		var/flooded_a_neighbor
+		for(var/spread_dir in GLOB.cardinal)
+			UPDATE_FLUID_BLOCKED_DIRS(T)
```
afterthought2: Why do this 4 times rather than once outside the loop?

code/controllers/subsystems/fluids.dm
```diff
@@ -41,14 +39,36 @@ var/datum/controller/subsystem/fluids/SSfluids
...
+				continue
+			UPDATE_FLUID_BLOCKED_DIRS(next)
+			if((next.fluid_blocked_dirs & GLOB.reverse_dir[spread_dir]) || !next.CanFluidPass(spread_dir) || checked[next])
+				continue
+			checked[next] = TRUE
+			flooded_a_neighbor = TRUE
+			var/obj/effect/fluid/F = locate() in next
+			if(!F)
+				F = new /obj/effect/fluid(next)
+				var/datum/gas_mixture/GM = T:return_air()
```
afterthought2: `.` seems fine here.

code/controllers/subsystems/fluids.dm
```diff
@@ -41,14 +39,36 @@ var/datum/controller/subsystem/fluids/SSfluids
...
+				if(GM) F.temperature = GM.temperature
+			if(F)
+				if(F.fluid_amount < FLUID_MAX_DEPTH)
+					SET_FLUID_DEPTH(F, FLUID_MAX_DEPTH)
+		if(!flooded_a_neighbor)
+			REMOVE_ACTIVE_FLUID_SOURCE(T)
+		else
+			ADD_ACTIVE_FLUID_SOURCE(T)
```
afterthought2: Isn't this redundant; i.e. if we've reached this line, aren't we in the relevant list already?

code/controllers/subsystems/fluids.dm
```diff
@@ -57,137 +77,132 @@ var/datum/controller/subsystem/fluids/SSfluids
...
-				UPDATE_FLUID_BLOCKED_DIRS(T)
-				if((T.fluid_blocked_dirs & coming_from) || !T.CanFluidPass(coming_from))
+				var/turf/neighbor_turf = get_step(T, spread_dir)
+				if(!istype(neighbor_turf) || neighbor_turf.flooded)
 					continue
-				var/obj/effect/fluid/other = locate() in T.contents
-				if(other && (QDELETED(other) || other.fluid_amount <= FLUID_DELETING))
+				var/coming_from = GLOB.reverse_dir[spread_dir]
+				UPDATE_FLUID_BLOCKED_DIRS(neighbor_turf)
+				if((neighbor_turf.fluid_blocked_dirs & coming_from) || !neighbor_turf.CanFluidPass(coming_from) || checked[neighbor_turf])
```
afterthought2: Should move the `checked[]` condition to before the one with the proc call.

code/controllers/subsystems/fluids.dm
```diff
@@ -57,137 +77,132 @@ var/datum/controller/subsystem/fluids/SSfluids
...
+		var/equalize_avg_temp = 0
+		F.last_flow_strength = 0
 
-			for(var/obj/effect/fluid/other in F.equalizing_fluids)
-				if(!istype(other) || QDELETED(other) || other.fluid_amount <= FLUID_DELETING)
-					F.equalizing_fluids -= other
-					continue
-				F.equalize_avg_depth += other.fluid_amount
-				F.equalize_avg_temp += other.temperature
+		for(var/thing in F.neighbors + F)
```
afterthought2: Why add + F here? It will never pass the check below as far as I can tell, and it costs you a list op here.

code/controllers/subsystems/fluids.dm
```diff
@@ -57,137 +77,132 @@ var/datum/controller/subsystem/fluids/SSfluids
...
+			var/equalize_count = 0
+			for(var/thing in F.neighbors + F)
+				var/obj/effect/fluid/other = thing
 				var/flow_amount = F.fluid_amount - other.fluid_amount
-				if(F.flow_amount < flow_amount && flow_amount >= FLUID_PUSH_THRESHOLD)
-					F.flow_amount = flow_amount
+				equalize_count++
+				equalize_avg_depth += other.fluid_amount
+				if(F.last_flow_strength < flow_amount && flow_amount >= FLUID_PUSH_THRESHOLD)
+					F.last_flow_strength = flow_amount
```
afterthought2, line 171: Wait, so it can never decrease? If last_flow_strength is >= flow_amount, it doesn't change.

code/controllers/subsystems/fluids.dm
```diff
@@ -57,137 +77,132 @@ var/datum/controller/subsystem/fluids/SSfluids
...
 
+		if(sufficient_delta)
+			var/setting_dir = 0
+			var/equalize_count = 0
+			for(var/thing in F.neighbors + F)
+				var/obj/effect/fluid/other = thing
 				var/flow_amount = F.fluid_amount - other.fluid_amount
-				if(F.flow_amount < flow_amount && flow_amount >= FLUID_PUSH_THRESHOLD)
-					F.flow_amount = flow_amount
+				equalize_count++
```
afterthought2: Feel like just do a length(F.neighbors) + 1 outside the loop.

code/controllers/subsystems/fluids.dm
```diff
@@ -57,137 +77,132 @@ var/datum/controller/subsystem/fluids/SSfluids
...
+				equalize_avg_depth += other.fluid_amount
+				if(F.last_flow_strength < flow_amount && flow_amount >= FLUID_PUSH_THRESHOLD)
+					F.last_flow_strength = flow_amount
 					setting_dir = get_dir(F, other)
```
afterthought2, line 172: I'm not really following what's going on with this var. It's used outside the loop, so its value seems to depend on the order neighbors is iterated.

afterthought2 (commented):
Some of the comments may be dumb, as I don't know the details of this system.

code/controllers/subsystems/fluids.dm
MistakeNot4892: It's finding the strongest flow for use in a loop at the end of the process call - I may have removed the line that zeroes it out before this loop.

code/controllers/subsystems/fluids.dm
MistakeNot4892: Not every neighbor equalizes.

code/controllers/subsystems/fluids.dm
MistakeNot4892: It is used to set the direction of the overlay in the flow direction of the strongest flow this iteration.

code/controllers/subsystems/fluids.dm
MistakeNot4892: `F` has to equalize as well - will check my logic, I might have broken it.

code/controllers/subsystems/fluids.dm
MistakeNot4892: Might be passing it the wrong argument, it's meant to update the turf we're stepping to.

code/controllers/subsystems/fluids.dm
afterthought2: Ah, didn't realize it was a temporary thing.

code/controllers/subsystems/fluids.dm
afterthought2: In that case it seems there's a logic issue, as this will be called exactly that many times here.

MistakeNot4892:
Updating broke equalizing so DNM.

MistakeNot4892:
Ok working again.

afterthought2 (approved):

PR #344, opened 2020-04-21
Adds in letterman jackets

PlayerDeer:
They are customizable in loadouts.
fxS5
There are four base colours. Red, blue, brown, and baby puke green.

Thank you everyone in the discord who helped me  : )


code/modules/clothing/suits/miscellaneous.dm
```diff
@@ -446,3 +446,47 @@
 
 /obj/item/clothing/suit/hospital/pink
 	color = "#ffb7db"
+
+/obj/item/clothing/suit/letterman/Initialize()
```
babydoll, line 458: Put these procs (`Initialize`, `on_update_icon`, `get_mob_overlay`) after the type definition on line 466, but before the subtype definitions beginning on 474

MistakeNot4892 (approved):

PR #349, opened 2020-04-22
Adds floating chat messages

comma:
AKA runescape chat
Stolen from beestation PR which is stolen from goon (in a license friendly way)
https://github.com/BeeStation/BeeStation-Hornet/pull/1603

Differencies to bee PR include name=color setup, and some code reshuffling to reduce copypaste.
It's shown if you have preference enabled, defaults to off

### DIFF
--- a/code/modules/mob/floating_message.dm
+++ b/code/modules/mob/floating_message.dm
@@ -1,0 +1,63 @@
+// Thanks to Burger from Burgerstation for the foundation for this
+var/list/floating_chat_colors = list()
+
+/atom/movable
+	var/list/stored_chat_text
+
+/atom/movable/proc/animate_chat(message, decl/language/language, small, list/show_to, duration)
+	set waitfor = FALSE
+	
+	var/style	//additional style params for the message
+	var/fontsize = 6
+	if(small)
+		fontsize = 5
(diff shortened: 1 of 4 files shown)


code/modules/mob/floating_message.dm
```diff
@@ -0,0 +1,74 @@
...
+	seed = hex2num(seed) % 10000
+	rand_seed(seed)
+	var/list/base_colors = list("#83c0dd","#8396dd","#9983dd","#c583dd","#dd83b6","#dd8383","#83dddc","#83dd9f","#a5dd83","#ddd983","#dda583","#dd8383")
+	var/base_color = base_colors[seed % base_colors.len]
+	var/list/components = GetHexColors(base_color)
+	for(var/i in 1 to 3)
+		var/sign = seed % 2 ? -1 : 1
+		components[i] += sign * rand(60)
+		components[i] = Clamp(components[i], 0, 255)
+	return rgb(components[1], components[2], components[3])
```
afterthought2: Why not use a normal `rand` call and cache the color in a global name -> color list? Would avoid global de-randomization exploits and would be more performant.

MistakeNot4892 (approved):

PR #350, opened 2020-04-22
Adds shuttle landing

NataKilar:
Adds manual shuttle landing for high-skill pilots, allowing for arbitrary placement of temporary landing landmarks. Code adapted from /tg/, particularly the code for displaying the landing images. Currently restricted to exoplanets.

On the backend, moves the current freelook (currently cult, AI, and camera connection spell) eyes over to a subtype, so that the parent type can be used for features that do not require visualnet style obfuscation.


code/modules/overmap/exoplanets/_exoplanet.dm
```diff
@@ -55,7 +56,6 @@
 	var/list/possible_themes = list(
 		/datum/exoplanet_theme = 30,
 		/datum/exoplanet_theme/mountains = 100,
-		/datum/exoplanet_theme/radiation_bombing = 10,
```
comma, line 58: ???

code/modules/mob/living/deity/deity.dm
```diff
@@ -16,7 +16,8 @@
 	bone_material = null
 	bone_amount = 0
 
-	var/eye_type = /mob/observer/eye/cult
+	var/eye_type = /mob/observer/eye/freelook/cult
+	var/mob/observer/eye/freelook/cult/eye
```
comma: why adding this var instead of using existing eyeobj? You're setting eyeobj anyway

NataKilar: eyeobj is of type mob/observer/eye, rather than /mob/observer/eye/freelook, so it doesn't have the vars associated with visualnets. I figured this was cleaner than type-casting it every time I needed to access a visual net specific var, although thinking about it now it could just hold its visualnet datum instead, which I'll probably do, unless there's a better solution.

code/modules/overmap/exoplanets/_exoplanet.dm
NataKilar: Missed the removal part of a whitespace change. Will fix.

code/modules/mob/living/deity/deity.dm
comma: Typecasting is fine, IMO better than keeping second reference to existing var
Also it used to have visualnet var didn't it, just keep em.

code/modules/mob/observer/eye/landing_eye/landing_eye.dm
```diff
@@ -0,0 +1,93 @@
...
+			. = FALSE // Cannot land on a dense turf.
+			continue
+		if(!A || A != get_area(src))
+			img.icon_state = "red"
+			. = FALSE // Cannot cross between two areas.
+			continue
+		if(!istype(A, /area/space) && !istype(A, /area/exoplanet)) // Can only land in space or outside.
+			img.icon_state = "red"
+			. = FALSE
+			continue
```
afterthought2, line 70: There's a proc (or at least proc piece) for checking shuttle collision at a turf when landing. Should just use the same proc here.

code/modules/overmap/ships/computers/shuttle.dm
```diff
@@ -25,7 +45,7 @@
 			"fuel_span" = fuel_span
 		)
 
-/obj/machinery/computer/shuttle_control/explore/handle_topic_href(var/datum/shuttle/autodock/overmap/shuttle, var/list/href_list)	
+/obj/machinery/computer/shuttle_control/explore/handle_topic_href(var/datum/shuttle/autodock/overmap/shuttle, var/list/href_list, var/mob/usr)
```
afterthought2: Better to rename it to `user` and replace; quite confusing otherwise.

code/modules/overmap/ships/computers/shuttle.dm
```diff
@@ -39,8 +59,108 @@
...
+
+		var/list/available_sectors = list()
+		for(var/obj/effect/overmap/visitable/S in range(get_turf(current_sector), shuttle.range))
+			if(S.free_landing)
+				available_sectors += S
+
+		if(LAZYLEN(available_sectors))
+			target_sector = input("Choose sector to land in.", "Sectors") as null|anything in available_sectors
+
+	if(target_sector && CanUseTopic(user))
```
afterthought2: CanInteract

code/modules/shuttles/landmarks.dm
```diff
@@ -122,6 +124,43 @@
...
+	timer_id = addtimer(CALLBACK(src, .proc/check_for_shuttle), 1 MINUTE, TIMER_STOPPABLE | TIMER_LOOP)
+	. = ..()
+
+/obj/effect/shuttle_landmark/temporary/Destroy()
+	SSshuttle.unregister_landmark(landmark_tag)
+	return ..()
+
+/obj/effect/shuttle_landmark/temporary/shuttle_departed(datum/shuttle/shuttle)
+	deltimer(timer_id)
+	qdel_self(src)
```
afterthought2: `qdel(src)`. `qdel_self` is a helper for when you're passing procs around to listeners; `qdel(src)` is preferred.

code/modules/shuttles/landmarks.dm
```diff
@@ -122,6 +124,43 @@
...
+	if(!shuttle_restricted)	// Temporary landmarks *must* be restricted to a single shuttle, so if this occurs, it should be deleted.
+		qdel_self()
+		return
+
+	var/datum/shuttle/autodock/shuttle = SSshuttle.shuttles[shuttle_restricted]
+	if(shuttle.current_location != src) // Something has gone wrong, or the shuttle has changed its destination, and thus this landmark should be deleted.
+		if(shuttle.next_location == src)
+			if(shuttle.current_location == shuttle.landmark_transition) // The shuttle is on the way, do not delete the landmark.
+				return
+		shuttle.next_location = null
```
afterthought2: Probably should be indented one tab further; I may have changed my destination from this to something else. Also, if `shuttle.next_location == src`, I feel we shouldn't delete; maybe they are just waiting for a bit.

Incidentally, it may be better to have a hook for that and delete if it triggers.

code/modules/shuttles/landmarks.dm
```diff
@@ -122,6 +124,43 @@
...
+	flags = SLANDMARK_FLAG_AUTOSET
+	var/timer_id
+
+/obj/effect/shuttle_landmark/temporary/Initialize(var/mapload, var/shuttle_name)
+	shuttle_restricted = shuttle_name
+	. = ..(mapload)
+
+/obj/effect/shuttle_landmark/temporary/Initialize()
+	landmark_tag += "-[random_id("landmarks",1,9999)]"
+	timer_id = addtimer(CALLBACK(src, .proc/check_for_shuttle), 1 MINUTE, TIMER_STOPPABLE | TIMER_LOOP)
```
afterthought2: I feel it would be more graceful to avoid the checks entirely; delete if our shuttle leaves or if it clears its destination from this landmark (add a hook) without being in transit to here or here already. If you do want to keep it, seems simpler to `START_PROCESSING(SSprocessing, src)` instead.

code/modules/shuttles/landmarks.dm
```diff
@@ -122,6 +124,43 @@
...
 
+//Used for custom landing locations. Self deletes after a shuttle leaves.
+/obj/effect/shuttle_landmark/temporary
+	name = "Landing Point"
+	landmark_tag = "landing"
+	flags = SLANDMARK_FLAG_AUTOSET
+	var/timer_id
+
+/obj/effect/shuttle_landmark/temporary/Initialize(var/mapload, var/shuttle_name)
+	shuttle_restricted = shuttle_name
```
afterthought2: While you can set it if you like, note that your landmark won't be added to the sector it's attached to (it's added if it's in the sector's list of initial waypoints to find, or if it's /automatic), and that's the only thing this var is used for. If you do wish for it to be added, manually call `add_landmark` in here (after ..()).

code/modules/mob/living/deity/deity.dm
NataKilar: Addressing the second half, I meant that the deity would hold its visualnet datum. The vast majority of the references to the visual net itself through linked.eye.visualnet etc, so that will end up cleaner regardless.

code/modules/shuttles/landmarks.dm
NataKilar: I'll go ahead and cut it, there's no real need to add it to the sector.

NataKilar:
Landmarks now have additional hooks for selection and deselection to make the temporary landmark removal process cleaner, as afterthought suggested. In the future, these could probably be used to create landing warning-outlines for those on the ground, but I'll leave that for another PR.

afterthought2:
Shuttle stuff looks fine to me. Not an expert on the eye/image stuff.

code/modules/mob/observer/eye/landing_eye/landing_eye.dm
```diff
@@ -0,0 +1,93 @@
...
+	// Generates the overlay of the shuttle on turfs.
+	var/turf/origin = get_turf(src)
+	for(var/area/A in shuttle.shuttle_area)
+		for(var/turf/T in A)
+			var/image/I = image('icons/effects/alphacolors.dmi', origin, "red")
+			// Record the offset of the turfs from the eye. The eye is where the shuttle landmark will be placed, so the resultant images will reflect actual landing.
+			var/x_off = T.x - origin.x
+			var/y_off = T.y - origin.y
+
+			I.loc = locate(origin.x + x_off, origin.y + y_off, origin.z)
```
MistakeNot4892, line 21: Should check world bounds when doing this, otherwise you're generating images nobody will ever see.

afterthought2 (approved):

code/modules/mob/observer/eye/landing_eye/landing_eye.dm
```diff
@@ -0,0 +1,93 @@
...
+			. = FALSE // Cannot land on two different types of turfs.
+			continue
+		if(check_collision(origin.loc, list(T))) // Checking for density or multi-area overlap.
+			img.icon_state = "red"
+			. = FALSE
+			continue
+		if(!istype(A, /area/space) && !istype(A, /area/exoplanet)) // Can only land in space or outside.
+			img.icon_state = "red"
+			. = FALSE
+			continue
```
MistakeNot4892, line 74: Possibly compress this block to...
````
		if(!T || !T.loc || !origin || !origin.loc)
			. = FALSE
		else if((T.x < TRANSITIONEDGE || T.x > world.maxx - TRANSITIONEDGE) || (T.y < TRANSITIONEDGE || T.y > world.maxy - TRANSITIONEDGE))
			. = FALSE // Cannot land past the normal world boundaries.
		else if(!istype(T, origin))
			. = FALSE // Cannot land on two different types of turfs.
		else if(check_collision(origin.loc, list(T))) // Checking for density or multi-area overlap.
			. = FALSE
		else if(!istype(A, /area/space) && !istype(A, /area/exoplanet)) // Can only land in space or outside.
			. = FALSE

		if(!.)
			img.icon_state = "red"
````

MistakeNot4892:
Compression comment can be done in another PR if you feel like it.

PR #353, opened 2020-04-23
Add maximum atom content configuration.

biscuitWizard:
* Adds a configuration variable for limiting the maximum number of /obj that can be in a single turf (recursively).
* Default value is unlimited.
* Turfs that have reached the configured variable limit will not allow new items to be dropped or thrown in that turf.
* Further improvements could be made like limiting conveyors, or teleporters with the same limit.

* For persistence, this discourages hoarding behaviour on single turfs in a user friendly way. (And allows a much tighter limit.)
* Other servers may wish to use the limit for performance reasons.

### DIFF
--- a/code/game/atoms.dm
+++ b/code/game/atoms.dm
@@ -165,8 +165,27 @@
 			found += A.search_contents_for(path,filter_path)
 	return found
 
-
-
+/*
+ * Recursively searches all atom contents (including contents contents and so on.)
+ * 
+ * ARGS: paths (array) - only count atoms of one the provided paths.
+ * RETURNS: A number of all contents found.
+*/
+/atom/proc/get_all_contents_count(var/paths)
+	var/count = 0
+	for(var/atom/A in src.contents)
+		if(paths)
+			var/found = FALSE
+			for(var/path in paths)
+				if(istype(A, path))
+					found = TRUE
+					break
+			if(!found)
+				continue
+		count++
+		if(A.contents.len)
+			count += A.get_all_contents_count()
+	return count
 
 /*
 Beam code by Gunbuddy
--- a/code/modules/mob/inventory.dm
+++ b/code/modules/mob/inventory.dm
@@ -181,14 +181,16 @@ var/list/slot_equipment_priority = list( \
 			qdel(grab)
 			. = TRUE
 		return
+	if(!is_space_available(get_active_hand(), Target ? get_turf(Target) : get_turf(loc)))
+		return
+	return hand ? drop_l_hand(Target) : drop_r_hand(Target)
 
-	var/obj/item/W = hand ? l_hand : r_hand
-	if(W && config.atom_content_limit)
-		var/turf/T = Target ? get_turf(Target) : get_turf(loc)
-		if(T.get_all_contents_count(list(/obj/structure, /obj/item)) >= config.atom_content_limit)
+/mob/proc/is_space_available(var/obj/item/W, var/atom/target)
+	if(W && target && config.atom_content_limit)
+		if(target.get_all_contents_count(list(/obj/structure, /obj/item)) >= config.atom_content_limit)
 			to_chat(src, SPAN_WARNING("There is no room to put that there."))
 			return 0
-	return hand ? drop_l_hand(Target) : drop_r_hand(Target)
+	return 1
 
 /*
 	Removes the object from any slots the mob might have, calling the appropriate icon update proc.
@@ -243,11 +245,8 @@ var/list/slot_equipment_priority = list( \
 /mob/proc/unEquip(obj/item/I, var/atom/target)
 	if(!canUnEquip(I))
 		return
-	if(I && config.atom_content_limit && target)
-		var/turf/T = get_turf(target)
-		if(T.get_all_contents_count(list(/obj/structure, /obj/item)) >= config.atom_content_limit)
-			to_chat(src, SPAN_WARNING("There is no room to put that there."))
-			return 0
+	if(target && !is_space_available(I, get_turf(target)))
+		return
 	drop_from_inventory(I, target)
 	return 1
 
--- a/code/modules/mob/inventory.dm
+++ b/code/modules/mob/inventory.dm
@@ -181,6 +181,13 @@ var/list/slot_equipment_priority = list( \
 			qdel(grab)
 			. = TRUE
 		return
+
+	var/obj/item/W = hand ? l_hand : r_hand
+	if(W && config.atom_content_limit)
+		var/turf/T = Target ? get_turf(Target) : get_turf(loc)
+		if(T.get_all_contents_count(list(/obj/structure, /obj/item)) >= config.atom_content_limit)
+			to_chat(src, SPAN_WARNING("There is no room to put that there."))
+			return 0
 	return hand ? drop_l_hand(Target) : drop_r_hand(Target)
 
 /*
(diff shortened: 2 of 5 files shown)


code/game/atoms.dm
```diff
@@ -165,8 +165,27 @@
...
+			var/found = FALSE
+			for(var/path in paths)
+				if(istype(A, path))
+					found = TRUE
+					break
+			if(!found)
+				continue
+		count++
+		if(A.contents.len)
+			count += A.get_all_contents_count()
```
PsiOmegaDelta, line 187: BYOND doesn't do too well with recursion, should instead unroll into a loop

biscuitWizard: That should be true too for line 165. I can update both, can I get an example of what that would look like? I'm not super familiar with how to flatten that data.

code/modules/mob/inventory.dm
```diff
@@ -181,6 +181,13 @@ var/list/slot_equipment_priority = list( \
 			qdel(grab)
 			. = TRUE
 		return
+
+	var/obj/item/W = hand ? l_hand : r_hand
```
PsiOmegaDelta: `var/obj/item/W = get_active_hand()`

PsiOmegaDelta (changes requested):

code/modules/mob/inventory.dm
```diff
@@ -236,6 +243,11 @@ var/list/slot_equipment_priority = list( \
 /mob/proc/unEquip(obj/item/I, var/atom/target)
 	if(!canUnEquip(I))
 		return
+	if(I && config.atom_content_limit && target)
+		var/turf/T = get_turf(target)
+		if(T.get_all_contents_count(list(/obj/structure, /obj/item)) >= config.atom_content_limit)
+			to_chat(src, SPAN_WARNING("There is no room to put that there."))
+			return 0
```
PsiOmegaDelta: This is almost straight copy-paste of https://github.com/NebulaSS13/Nebula/pull/353/files#diff-b0373f944c6d476ad32d44bf2693e238R186-R190

code/modules/mob/living/carbon/carbon.dm
```diff
@@ -268,6 +268,10 @@
 
 	if(!item) return
 
+	if(config.atom_content_limit && get_turf(target).get_all_contents_count(list(/obj)) >= config.atom_content_limit)
```
PsiOmegaDelta: Multiple `get_turfs()` of the same object (`get_turf(target)`)
The latter case also double-checks that turf isn't null, this line does not

code/modules/mob/inventory.dm
PsiOmegaDelta: Shouldn't this check be in `canUnEquip()`?

PsiOmegaDelta (changes requested):

code/modules/mob/inventory.dm
biscuitWizard: Felt bleedy. canUnEquip is called by other stuff. Could have side-effects.

code/game/atoms.dm
biscuitWizard: Line 513 of code/_helpers/unsorted.dm uses recursion for GetAllContents()
Line 165 of code/game/atoms.dm uses recursion for path searching contents.

Just additional places where this critique applies, and maybe out of scope of this PR for this requested change to be enacted. Especially because I don't know how to "unroll" it without fancy features present in modern languages.

code/modules/mob/inventory.dm
```diff
@@ -236,6 +245,8 @@ var/list/slot_equipment_priority = list( \
 /mob/proc/unEquip(obj/item/I, var/atom/target)
 	if(!canUnEquip(I))
 		return
+	if(target && !is_space_available(I, get_turf(target)))
```
MistakeNot4892, line 248: This proc moves `I` into `target`, not onto `target`'s turf.

code/modules/mob/inventory.dm
```diff
@@ -181,8 +181,17 @@ var/list/slot_equipment_priority = list( \
 			qdel(grab)
 			. = TRUE
 		return
+	if(!is_space_available(get_active_hand(), Target ? get_turf(Target) : get_turf(loc)))
+		return
 	return hand ? drop_l_hand(Target) : drop_r_hand(Target)
 
+/mob/proc/is_space_available(var/obj/item/W, var/atom/target)
```
MistakeNot4892, line 188: Wouldn't it make more sense to define this on the atom itself rather than only when a mob is involved?

MistakeNot4892 (commented):
I definitely feel like this is way too high level as a solution for the problem you described on Discord. There are a ton of edge cases and ways for atoms to pile up on turfs or inside other atoms beyond when dropped by a mob. You could consider overriding `/atom/Enter()` to fail if the contents bound is exceeded, for example, and handle that gracefully in `dropInto()` to avoid breaking mob drop/unequip calls.

code/game/atoms.dm
MistakeNot4892:
````
	var/count = 0
	var/list/counting_atoms = contents.Copy() + src
	while(length(counting_atoms))
		var/atom/movable/thing = counting_atoms[1]
		counting_atoms -= thing
		counting_atoms |= thing.contents
		for(var/path in paths)
			if(istype(thing, path))
				count++
				break
````

code/modules/mob/inventory.dm
biscuitWizard: As designed.

code/modules/mob/inventory.dm
biscuitWizard: > I definitely feel like this is way too high level as a solution for the problem you described on Discord. There are a ton of edge cases and ways for atoms to pile up on turfs or inside other atoms beyond when dropped by a mob. You could consider overriding `/atom/Enter()` to fail if the contents bound is exceeded, for example, and handle that gracefully in `dropInto()` to avoid breaking mob drop/unequip calls.

Counterpoint: I think handling all edge cases would be a massive undertaking, and a brittle solution. The effort required to handle all edge cases would exceed the value of the solution itself as well. Handling just the happy-path is efficient enough. That's my cost-benefit analysis.

I've looked at a few exceptions, things like mobs or containers getting Destroyed can immediately drop all contents into their occupying turf. Same for deconstruction and the like. In this way you can still bypass the atom/turf content limit, but it is *much* harder than it was before. I think that fits my design goals.

biscuitWizard:
> I definitely feel like this is way too high level as a solution for the problem you described on Discord. There are a ton of edge cases and ways for atoms to pile up on turfs or inside other atoms beyond when dropped by a mob. You could consider overriding `/atom/Enter()` to fail if the contents bound is exceeded, for example, and handle that gracefully in `dropInto()` to avoid breaking mob drop/unequip calls.

Counterpoint: I think handling all edge cases would be a massive undertaking, and a brittle solution. The effort required to handle all edge cases would exceed the value of the solution itself as well. Handling just the happy-path is efficient enough. That's my cost-benefit analysis.

I've looked at a few exceptions, things like mobs or containers getting Destroyed can immediately drop all contents into their occupying turf. Same for deconstruction and the like. In this way you can still bypass the atom/turf content limit, but it is *much* harder than it was before. I think that fits my design goals.

biscuitWizard:
User story:
I am a miner on a planet. I mine a bunch of ore. Instead or processing more ore, I put it into a locker.

I keep putting it into a locker. For two days.

There are 1,000 stacks of ore in that locker.

This PR fixes that specific scenario. Replacing ore with any item.

It does not fix systemic aggregation of resources by machines. Example: Conveyor belts.

code/modules/mob/inventory.dm
biscuitWizard: The check is at the turf level to prevent bypassing content restrictions by nesting contents in contents /w containers.

code/modules/mob/inventory.dm
biscuitWizard: The only time it's called is when a mob is involved.

code/game/atoms.dm
biscuitWizard: This only goes one level deep and would break my code.

MistakeNot4892: No it doesn't?

MistakeNot4892:
This would not, for example, fix someone dragging a locker onto the pile of ore and closing it.
Lockers also already have a contents size restriction.
Your solution is significantly more brittle than an override at a fundamental level.

PR #358, opened 2020-04-23
Fixes shuttle range calculation with multi-tile sectors

NataKilar:


afterthought2 (approved):
Should probably look into using pixel distance instead.

PR #359, opened 2020-04-23
Pipe leak and stasis clamp changes

quardbreak:
Stolen from https://github.com/BoHBranch/BoH-Bay/pull/695
Might fix issues with pipe leak. 
Also allows use clamps on 4-way pipes and non-simply pipes.

### DIFF
--- a/code/game/machinery/atmoalter/clamp.dm
+++ b/code/game/machinery/atmoalter/clamp.dm
@@ -7,18 +7,17 @@
 	icon = 'icons/atmos/clamp.dmi'
 	icon_state = "pclamp0"
 	anchored = 1.0
-	var/obj/machinery/atmospherics/pipe/simple/target = null
+	var/obj/machinery/atmospherics/pipe/target = null
 	var/open = 1
 
-	var/datum/pipe_network/network_node1
-	var/datum/pipe_network/network_node2
+	var/list/datum/pipe_network/network_nodes
 
-/obj/machinery/clamp/Initialize(mapload, var/obj/machinery/atmospherics/pipe/simple/to_attach = null)
+/obj/machinery/clamp/Initialize(mapload, var/obj/machinery/atmospherics/pipe/to_attach = null)
 	. = ..(mapload)
 	if(istype(to_attach))
 		target = to_attach
 	else
-		target = locate(/obj/machinery/atmospherics/pipe/simple) in loc
+		target = locate(/obj/machinery/atmospherics/pipe) in loc
 	if(target)
 		update_networks()
 		set_dir(target.dir)
@@ -27,14 +26,9 @@
 	if(!target)
 		return
 	else
-		var/obj/machinery/atmospherics/pipe/node1 = target.node1
-		var/obj/machinery/atmospherics/pipe/node2 = target.node2
-		if(istype(node1))
-			var/datum/pipeline/P1 = node1.parent
-			network_node1 = P1.network
-		if(istype(node2))
-			var/datum/pipeline/P2 = node2.parent
-			network_node2 = P2.network
+		for(var/obj/machinery/atmospherics/pipe/node in target.pipeline_expansion())
+			var/datum/pipeline/PL = node.parent
+			LAZYADD(network_nodes, PL.network)
 
 /obj/machinery/clamp/physical_attack_hand(var/mob/user)
 	if(!target)
@@ -47,8 +41,9 @@
 	return TRUE
 
 /obj/machinery/clamp/Destroy()
-	if(!open)
-		spawn(-1) open()
+	if(!open && !QDELING(target))
+		open()
+	target = null
 	. = ..()
 
 /obj/machinery/clamp/proc/open()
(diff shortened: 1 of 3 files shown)


afterthought2:
What issue is this supposed to fix?

quardbreak:
See PR link above.

code/game/machinery/atmoalter/clamp.dm
```diff
@@ -80,39 +69,31 @@
...
-		P1.build_pipeline(node1)
-		qdel(P1)
-	if(istype(node2) && node2.parent)
-		var/datum/pipeline/P2 = node2.parent
-		P2.build_pipeline(node2)
-		qdel(P2)
-//  P1.build_network()
-//  P2.build_network()
+	for(var/obj/machinery/atmospherics/pipe/node in target.pipeline_expansion())
+		if(!istype(node))
```
Techhead0: FYI this is extraneous since a `for(var/path/foo in var)` loop will only find things of that part. NBD though since this is a small list and not called often

code/modules/atmospherics/pipes.dm
```diff
@@ -454,6 +459,11 @@
 	if(node3)
 		node3.disconnect(src)
 		node3 = null
+	if(in_stasis)
+		var/obj/machinery/clamp/C = locate() in get_turf(src)
+		if(C.target == src)
+			C.open()
+			C.removal()
```
Techhead0: Should this block be at level `/obj/machinery/atmospherics/pipe`?

MarinaGryphon: OH SHIT THANKS YEAH

MarinaGryphon: Since it calls parent moving it to `/obj/machinery/atmospherics/pipe/Destroy()` causes no issues.

Techhead0 (dismissed):
As the one who reviewed the original stasis clamp PR, looks good to me

MarinaGryphon:
> As the one who reviewed the original stasis clamp PR, looks good to me

We got some runtimes on Hestia and I'm unsure if my latest commit has fixed them.

MistakeNot4892:
Could you share the runtimes @MarinaGryphon?

code/game/machinery/atmoalter/clamp.dm
```diff
@@ -80,39 +69,29 @@
...
-		qdel(P1)
-	if(istype(node2) && node2.parent)
-		var/datum/pipeline/P2 = node2.parent
-		P2.build_pipeline(node2)
-		qdel(P2)
-//  P1.build_network()
-//  P2.build_network()
+	for(var/obj/machinery/atmospherics/pipe/node in target.pipeline_expansion())
+		node.build_network()
+		network_nodes += node
```
MarinaGryphon:
```suggestion
		LAZYADD(network_nodes, node)
```

code/game/machinery/atmoalter/clamp.dm
```diff
@@ -27,14 +26,9 @@
...
-		var/obj/machinery/atmospherics/pipe/node2 = target.node2
-		if(istype(node1))
-			var/datum/pipeline/P1 = node1.parent
-			network_node1 = P1.network
-		if(istype(node2))
-			var/datum/pipeline/P2 = node2.parent
-			network_node2 = P2.network
+		for(var/obj/machinery/atmospherics/pipe/node in target.pipeline_expansion())
+			var/datum/pipeline/PL = node.parent
+			network_nodes += PL.network
```
MarinaGryphon:
```suggestion
			LAZYADD(network_nodes, PL.network)
```

code/game/machinery/atmoalter/clamp.dm
```diff
@@ -80,39 +69,29 @@
...
-	if(network_node1)
-		qdel(network_node1)
-	if(network_node2)
-		qdel(network_node2)
-
-	var/obj/machinery/atmospherics/pipe/node1 = null
-	var/obj/machinery/atmospherics/pipe/node2 = null
+	for(var/datum/pipe_network/netnode in network_nodes)
+		qdel(netnode)
+	network_nodes.Cut()
```
MarinaGryphon:
```suggestion
	network_nodes = null
```

MarinaGryphon (changes requested):
Needs changes to prevent runtimes, I think. Will update if this fixes them on Hestia.

Techhead0 (approved):
Yeah this looks good too me. Nothing new since last time but runtime safety

code/game/machinery/atmoalter/clamp.dm
```diff
@@ -80,39 +69,29 @@
...
-		P2.build_pipeline(node2)
-		qdel(P2)
-//  P1.build_network()
-//  P2.build_network()
+	for(var/obj/machinery/atmospherics/pipe/node in target.pipeline_expansion())
+		node.build_network()
+		LAZYADD(network_nodes, node)
+		var/datum/pipeline/P = node.parent
+		P.build_pipeline(node)
+		qdel(P)
```
afterthought2, line 81: Is this some hack to distribute air around? I don't get it. Why are we deleting P here?

afterthought2 (commented):
Seems like this will just break if the target's state changes (e.g. adjacent pipe disconnected). Would make more sense to me if instead clamps were dummy devices that modified logic on pipes. But it's also logic-equivalent, except that it can go on non-simple pipes now, so idk.

afterthought2 (approved):
I guess it's hard to argue this is worse than current implementation.

MistakeNot4892:
Logic seems equivalent either way.

PR #360, opened 2020-04-23
Added signal strength simulation.

biscuitWizard:
* Adds signal strength simulation to new network code.
* Router now takes network_card.
* long_range will double router's range.
* base range is 25 distance from tile.

Relays will be added in a future commit.

### DIFF
--- a/code/__defines/computers.dm
+++ b/code/__defines/computers.dm
@@ -44,6 +44,7 @@
 #define NETWORK_CONNECTION_WIRELESS			1
 #define NETWORK_CONNECTION_STRONG_WIRELESS	2
 #define NETWORK_CONNECTION_WIRED			3
+#define NETWORK_BASE_BROADCAST_STRENGTH		25
 
 // Caps for network logging. Less than 10 would make logging useless anyway, more than 500 may make the log browser too laggy. Defaults to 100 unless user changes it.
 #define MAX_NETWORK_LOGS 100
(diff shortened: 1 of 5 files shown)


code/modules/modular_computers/NTNet/_network.dm
```diff
@@ -98,8 +98,12 @@ GLOBAL_LIST_EMPTY(computer_networks)
 	return ARE_Z_CONNECTED(get_z(router.holder), get_z(D.holder))
 
 /datum/computer_network/proc/get_signal_strength(datum/extension/network_device/D)
-	if(check_connection(D))
-		return D.connection_type
+	if(!check_connection(D))
+		return 0
+	var/broadcast_strength = router.get_broadcast_strength()
```
PsiOmegaDelta: The only correct pattern of acquiring extensions is:
`var/datum/extension/network_device/router/router = get_extension(src, /datum/extension/network_device/router)`

PsiOmegaDelta (dismissed):

code/modules/modular_computers/NTNet/_network.dm
biscuitWizard: This is outside of the scope for this PR. Recommend making it an issue for a new PR.

code/modules/modular_computers/NTNet/_network.dm
```diff
@@ -98,8 +98,12 @@ GLOBAL_LIST_EMPTY(computer_networks)
...
+	var/distance = get_dist(router.holder, D.holder)
```
comma: I think it's better do get_turf() on both because not sure what get_dist would do to nested things (e.g. network card is holder, but its inside a PDA carried by a mob)

biscuitWizard: it is done

comma (approved):

PR #364, opened 2020-04-24
Adding a mining application for generating money out of nothing.

biscuitWizard:
Small application that when running can "donate processor cycles for money". The amount of money generated is insignificant, and the time it takes to do so ranges from 15-30 minutes on average. (Though it's faster with faster processors. Half time with the best processor.)

This application is just an anti-deadlock for persistence, where vending machines provide basic gear but cost money. The presence of this application prevents a deadlock where players run out of money from losing all their gear and thus cannot play anymore.

In round-based play, it's just a silly and fun application that adds a bit of flavor to the game without unbalancing it.

### DIFF
--- a/code/modules/modular_computers/file_system/programs/generic/folding.dm
+++ b/code/modules/modular_computers/file_system/programs/generic/folding.dm
@@ -1,0 +1,119 @@
+GLOBAL_LIST_INIT(science_strings, list(
+	"Extruding Mesh Terrain", "Virtualizing Microprocessor",
+	"Reticulating Splines", "Inserting Chaos Generator",
+	"Reversing Polarity", "Unfolding Proteins",
+	"Simulating Alien Abductions", "Scanning Pigeons",
+	"Iterating Chaos Array", "Abstracting Supermatter",
+	"Adjusting Social Network", "Recalculating Clown Principle"
+))
+
+#define MINIMUM_SCIENCE_INTERVAL 450
+#define MAXIMUM_SCIENCE_INTERVAL 900
+#define MINIMUM_FOLDING_EVENT_INTERVAL 50
+#define MAXIMUM_FOLDING_EVENT_INTERVAL 100
+#define SCIENCE_MONEY_PER_MINUTE 0.08 // So little money.
+
+/datum/computer_file/program/folding
+	filename = "fldng"
+	filedesc = "FOLDING@SAPCE"
+	extended_desc = "This program uses processor cycles for science, in exchange for money."
+	program_icon_state = "generic"
+	program_key_state = "generic_key"
+	size = 6
+	available_on_network = 1
+	usage_flags = PROGRAM_ALL
+	nanomodule_path = /datum/nano_module/program/folding
+	category = PROG_UTIL
+	requires_network = 1
+
+	var/started_on 			= 0 	// When the program started some science.
+	var/current_interval 	= 0		// How long the current interval will be.
+	var/next_event 			= 0 	// in world timeofday, when the next event is scheduled to pop.
+	var/crashed				= FALSE	// Program periodically needs a restart.
+	var/crashed_at			= 0		// When the program crashed.
+
+/datum/computer_file/program/folding/Topic(href, href_list)
+	if(..())
+		return TOPIC_HANDLED
+	. = TOPIC_REFRESH
+
+	if(href_list["fix_crash"] && crashed)
+		current_interval += world.timeofday - crashed_at
+		crashed = FALSE
+
+	if(href_list["start"] && started_on == 0)
+		started_on = world.timeofday
+		current_interval = rand(MINIMUM_SCIENCE_INTERVAL, MAXIMUM_SCIENCE_INTERVAL) SECONDS
+		// Better processors can do this faster. Normal processors are slower.
+		var/obj/item/stock_parts/computer/processor_unit/processor = computer.get_component(PART_CPU)
+		current_interval = (processor.processing_power / 2) * current_interval
+		next_event = (rand(MINIMUM_FOLDING_EVENT_INTERVAL, MAXIMUM_FOLDING_EVENT_INTERVAL) SECONDS) + world.timeofday
+
+	if(href_list["collect"] && started_on > 0 && !crashed)
+		var/obj/item/card/id/I = usr.GetIdCard()
+		if(!I)
+			to_chat(usr, SPAN_WARNING("Unable to locate ID card for transaction."))
+			return TOPIC_HANDLED
+		var/datum/money_account/account = get_account(I.associated_account_number)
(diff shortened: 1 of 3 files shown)


MistakeNot4892:
In before people think that this is literally adding bitcoin mining like the /tg/ PR.

code/modules/modular_computers/file_system/programs/generic/folding.dm
```diff
@@ -0,0 +1,119 @@
...
+		current_interval = (processor.processing_power / 2) * current_interval
+		next_event = (rand(MINIMUM_FOLDING_EVENT_INTERVAL, MAXIMUM_FOLDING_EVENT_INTERVAL) SECONDS) + world.timeofday
+
+	if(href_list["collect"] && started_on > 0 && !crashed)
+		var/obj/item/card/id/I = usr.GetIdCard()
+		if(!I)
+			to_chat(usr, SPAN_WARNING("Unable to locate ID card for transaction."))
+			return TOPIC_HANDLED
+		var/datum/money_account/account = get_account(I.associated_account_number)
+		var/earned = current_interval * SCIENCE_MONEY_PER_MINUTE
```
Techhead0: Making it based directly on `current_interval` means that more processing power doesn't actually make you more money. Should probably average the min/max science interval

code/modules/modular_computers/file_system/programs/generic/folding.dm
```diff
@@ -0,0 +1,119 @@
...
+				if(2)
+					to_chat(holder, SPAN_WARNING("The [computer] gets scaldingly hot, burning you!"))
+					holder.burn_skin(0.45)
+				if(3)
+					to_chat(holder, SPAN_WARNING("The [computer] pings an error chime."))
+					crashed = TRUE
+					crashed_at = world.timeofday
+		else
+			crashed = TRUE
+			crashed_at = world.timeofday
```
Techhead0, line 88: So, it *always* crashes if you're not holding it and only sometimes crashes if you are?

Techhead0 (commented):
Also, simply hiding the Collect button when not-done and not confirming that it's actually done leaves it vulnerable to a href exploit, although with such little money IDK if it actually matters

code/modules/modular_computers/file_system/programs/generic/folding.dm
biscuitWizard: Yes. It's the balance of trying to escape consequences by putting it down.

code/modules/modular_computers/file_system/programs/generic/folding.dm
```diff
@@ -0,0 +1,122 @@
...
+		var/datum/money_account/account = get_account(I.associated_account_number)
+		var/obj/item/stock_parts/computer/processor_unit/processor = computer.get_component(PART_CPU)
+		var/earned = current_interval * (SCIENCE_MONEY_PER_MINUTE * processor.processing_power)
+		account.deposit(earned, "Completed FOLDING@SPACE project.")
+		to_chat(usr, SPAN_NOTICE("Transferred [earned] to your account."))
+		started_on = 0
+		current_interval = 0
+
+	if(.)
+		SSnano.update_uis(NM)
```
afterthought2: No need for this; returning TOPIC_REFRESH will refresh anyway.

code/modules/modular_computers/file_system/programs/generic/folding.dm
```diff
@@ -0,0 +1,122 @@
...
+
+	var/started_on 			= 0 	// When the program started some science.
+	var/current_interval 	= 0		// How long the current interval will be.
+	var/next_event 			= 0 	// in world timeofday, when the next event is scheduled to pop.
+	var/crashed				= FALSE	// Program periodically needs a restart.
+	var/crashed_at			= 0		// When the program crashed.
+
+/datum/computer_file/program/folding/Topic(href, href_list)
+	if(..())
+		return TOPIC_HANDLED
```
afterthought2: Really weird pattern that preempts base checks from closing the UI or refreshing it. Is this a conscious choice?

biscuitWizard: Yep. It's done on a significant number of all Topic() implementations.

code/modules/modular_computers/file_system/programs/generic/folding.dm
biscuitWizard: This was copy pasted code where it's done elsewhere in the source. Copy was from scanner.dm. Just delete line 67 and 68?

afterthought2: I believe it's not necessary in scanner.dm either, except that that file doesn't use Topic return value defines (it will actually refresh on any truthy reply, though, despite the names).

code/modules/modular_computers/file_system/programs/generic/folding.dm
afterthought2: Well, the typical line would be
```
. = ..()
if(.)
	return
```
Not everything uses Topic return value defines properly, but those that do should try and remain consistent about it.

code/modules/modular_computers/file_system/programs/generic/folding.dm
```diff
@@ -0,0 +1,120 @@
...
+	if(href_list["fix_crash"] && crashed)
+		current_interval += world.timeofday - crashed_at
+		crashed = FALSE
+
+	if(href_list["start"] && started_on == 0)
+		started_on = world.timeofday
+		current_interval = rand(MINIMUM_SCIENCE_INTERVAL, MAXIMUM_SCIENCE_INTERVAL) SECONDS
+		// Better processors can do this faster. Normal processors are slower.
+		var/obj/item/stock_parts/computer/processor_unit/processor = computer.get_component(PART_CPU)
+		current_interval = (processor.processing_power / 2) * current_interval
```
Techhead0: Should be divided by processing_power if you want faster processors to be faster
Also, `*=` is handy

afterthought2: Comment seems accurate.

afterthought2 (approved):

code/modules/modular_computers/file_system/programs/generic/folding.dm
```diff
@@ -0,0 +1,117 @@
...
+
+#define MINIMUM_SCIENCE_INTERVAL 450
+#define MAXIMUM_SCIENCE_INTERVAL 900
+#define MINIMUM_FOLDING_EVENT_INTERVAL 50
+#define MAXIMUM_FOLDING_EVENT_INTERVAL 100
+#define SCIENCE_MONEY_PER_MINUTE 0.08 // So little money.
+
+/datum/computer_file/program/folding
+	filename = "fldng"
+	filedesc = "FOLDING@SAPCE"
```
MistakeNot4892, line 18: `SAPCE`

PR #365, opened 2020-04-24
Adds material sheet vendomat

comma:
Anticipating more material usage in future, it helps to make sure players always have all the kinds they need at start, so this is an easier way of ensuring that compared to mapping sheets.

Also mapped it on bearcat


afterthought2:
Kind of confusing to reuse the engineering vendor sprite, unless it's not what I think it is.

MistakeNot4892:
I sorta wonder if a 'default materials' landmark would be better than using a vendor.

MistakeNot4892 (approved):

PR #368, opened 2020-04-24
Makes holograms not use crew record picture

CrimsonShrike:
Dont really see point of it, it just makes people's hologram not representative of their actual status.


PsiOmegaDelta:
Do holograms necessarily have to be a present representation?
Could just be a stored version, animated using space motion-capture.

CrimsonShrike:
Guess it's a stylistic choice, but since holograms transmit a lot of things that are definitely real time such as messages about actions, I figure they are realy scanning person and sorroundings.

MistakeNot4892 (approved):

PR #372, opened 2020-04-26
Playing cards QoL changes

quardbreak:
Stole some qol changes from Infinity Station:
* Added compact type
* Added click on gambling table to deal with cards
* Some misc output messages like deck examine.

### DIFF
--- a/code/modules/games/cards.dm
+++ b/code/modules/games/cards.dm
@@ -52,9 +52,14 @@
 	desc = "A simple deck of playing cards."
 	icon_state = "deck"
 
-/obj/item/deck/cards/Initialize()
+/obj/item/deck/Initialize()
 	. = ..()
+	generate_cards()
 
+/obj/item/deck/proc/generate_cards()
+	return
+
+/obj/item/deck/cards/generate_cards()
 	var/datum/playingcard/P
 	for(var/suit in list("spades","clubs","diamonds","hearts"))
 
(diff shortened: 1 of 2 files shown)


code/modules/games/cards.dm
```diff
@@ -54,7 +54,9 @@
 
 /obj/item/deck/cards/Initialize()
 	. = ..()
+	return INITIALIZE_HINT_LATELOAD
 
+/obj/item/deck/cards/LateInitialize()
```
afterthought2: Why the move to lateload?

code/modules/games/cards.dm
```diff
@@ -78,18 +80,59 @@
...
 		P.name = "joker"
 		P.card_icon = "joker"
 		cards += P
 
+/obj/item/deck/compact
+	name = "compact deck of cards"
+	desc = "A deck of playing cards. Looks like this one hasn't numbers from two to five, and jokers."
+	icon_state = "deck"
+
+/obj/item/deck/compact/LateInitialize()
```
afterthought2: If it's to do a hard override without calling parent, just use a custom proc that you call from Initialize instead.

code/modules/games/cards.dm
```diff
@@ -185,16 +232,20 @@
...
-	if(!Adjacent(usr) || !over.Adjacent(usr)) return // should stop you from dragging through windows
-
-	if(!ishuman(over) || !(over in viewers(3))) return
-
-	if(!cards.len)
-		to_chat(usr, "There are no cards in the deck.")
-		return
-
-	deal_at(usr, over)
+	if(over == usr && !usr.restrained() && !usr.stat && (usr.contents.Find(src) || in_range(src, usr)))
```
afterthought2: `incapacitated` should cover some of these checks.

code/modules/tables/tables.dm
```diff
@@ -130,6 +130,16 @@
...
 		if(H.cards && H.cards.len == 1)
 			usr.visible_message("\The [user] plays \the [H.cards[1].name].")
+
+	if(istype(W, /obj/item/deck)) //playing cards
+		if(user.a_intent == I_GRAB)
+			var/obj/item/deck/D = W
+			if(!D.cards.len)
+				to_chat(usr, "There are no cards in the deck.")
+				return
+			D.deal_at(usr, src)
```
afterthought2: `user`. Same two lines above.

code/modules/tables/tables.dm
```diff
@@ -130,6 +130,16 @@
 		var/obj/item/hand/H = W
 		if(H.cards && H.cards.len == 1)
 			usr.visible_message("\The [user] plays \the [H.cards[1].name].")
+
+	if(istype(W, /obj/item/deck)) //playing cards
+		if(user.a_intent == I_GRAB)
```
afterthought2, line 135: Wait, why?

quardbreak: Every attackby will trigger this. We make it check grab intent so it won't call deal proc.

afterthought2 (approved):

PR #375, opened 2020-04-27
Add mandatory parent calls

PsiOmegaDelta:
Probably won't succeed the first run due to some `/examine()` overrides


code/game/atoms.dm
```diff
@@ -276,6 +279,7 @@ its easier to just keep the beam vertical.
 		icon_state = new_icon_state
 
 /atom/proc/update_icon()
+	SHOULD_CALL_PARENT(TRUE)
```
MistakeNot4892, line 282: Nothing else overrides or calls `update_icon()` though. It's even tested by Travis that nothing overrides it.

PsiOmegaDelta: I can confirm that this is not true

MistakeNot4892: I stand corrected. Thanks relative pathing.

code/modules/clothing/masks/monitor.dm
```diff
@@ -45,8 +45,8 @@
 		)
 
 /obj/item/clothing/mask/monitor/set_dir()
-	dir = SOUTH
```
MistakeNot4892, line 48: This is deliberate, it has null states for some directions and will be invisible if dragged/discarded while not equipped.

PsiOmegaDelta: `dir` should never be changed outside of the `set_dir()`-proc and even VV is re-routed through it, so with this override as long as `dir` was initially set to `SOUTH` it will remain so.

Anything that directly sets `dir` should be replaced with a `set_dir()`-call, due to event handling if nothing else.

PsiOmegaDelta: I see now that there are some cases where `dir` is manually set, bypassing `set_dir()`, looking rather deliberate. I'll arrange something.

afterthought2: I feel it's fine to not call parent in this proc specifically if the dir is left unchanged, or to modify what the given arguments are before calling parent. In this example, for instance, you could instead do `..(SOUTH)`.

PsiOmegaDelta: The problem, as I see it, there are pieces of code doing `dir = DIRECTION` and overriding `set_dir()` to prevent others from changing said direction.
I've got a solution for it.

code/modules/multiz/zmimic/mimic_common.dm
```diff
@@ -9,11 +9,11 @@
...
-/turf/update_icon()
+/turf/on_update_icon()
 	..()
 	if (above)
 		update_above()
 
-/atom/movable/update_icon()
+/atom/movable/on_update_icon()
 	..()
 	UPDATE_OO_IF_PRESENT
```
afterthought2: This defeats the purpose: everything hard-overrides on_update_icon, so these updates will never fire except on atoms where they do nothing.

code/modules/organs/external/_external_icons.dm
```diff
@@ -1,7 +1,10 @@
 var/list/limb_icon_cache = list()
 
-/obj/item/organ/external/set_dir()
-	return
+/obj/item/organ/external/set_dir(var/direction, var/forced)
```
PsiOmegaDelta, line 3: This allows external organs to forcefully set a direction when desired, while preventing normal attempts

code/modules/multiz/zmimic/mimic_common.dm
PsiOmegaDelta: Original `update_icon()`-calls restored.

code/modules/clothing/masks/monitor.dm
PsiOmegaDelta: Solution implemented for the override I noted was bypassed via `dir = direction`, see external organs

MistakeNot4892 (approved):

PR #376, opened 2020-04-27
Enforces calling parent for some program procs

comma:
They do important stuff there and it breaks stuff if you don't.


PsiOmegaDelta (dismissed):
Assuming it passes

PsiOmegaDelta (approved):

PR #377, opened 2020-04-27
Converting reagents to a decl system.

MistakeNot4892:
**Changes**
- Converts `/datum/reagent` to `/decl/reagent`. Reagents are now stored on two lists within `/datum/reagents`, `reagent_volumes` and `reagent_data`.
- Refactored a bunch of blood data code for the above.
- Refactored a bunch of reagent handling code for the above.
- Added code to SSchemistry to ensure reagent changes to atoms only happen once per tick even if multiple things call it (this is mostly for the reagent fluids PR).
- Removed a hard dependance on the blood reagent type path from human vessel interactions. Should currently have no effect, but will allow for non-blood vessel contents in the future hopefully.
- Removed the old readme file from the reagents module. It's woefully outdated and unmaintained, and honestly isn't very useful. Partial recreation found at: https://github.com/NebulaSS13/Nebula/wiki/Reagents
- Paint has been drastically simplified. It will now use the colour of the reagents mixed with it and does not weight the colour at all itself.

Ideally the outcome of this PR will be cheaper reagent procs with a much smaller memory footprint. No more creation or deletion of reagent datums, a number of unnecessary for loops removed, and generally nicer code.


code/controllers/subsystems/chemistry.dm
```diff
@@ -60,9 +60,9 @@ SUBSYSTEM_DEF(chemistry)
 			return
 
 /datum/controller/subsystem/chemistry/proc/get_prototype(given_type, temperature)
-	if(!ispath(given_type, /datum/reagent/random))
+	if(!ispath(given_type, /decl/reagent/random))
 		return
-	var/datum/reagent/random/prototype = random_chem_prototypes[given_type]
+	var/decl/reagent/random/prototype = random_chem_prototypes[given_type]
```
MistakeNot4892: TODO refactor this.

MistakeNot4892:
As far as I can tell this all works. I have specifically tested:
- Blood samples and blood mixing.
- Injections into mobs and objects.
- ChemMaster and chem dispenser.
- Pill making.
- Reactions in general.
- Chem grenades (weedkiller grenade specifically).
- Blood splatters.
- Eating and drinking.

Also added some unit tests to ensure basic `trans_to` functions still work.

comma (dismissed):

deathride58 (approved):
TIL force pushes reset the files viewed thing, qq

~~On a more serious note, undocumented changes in large PRs are kind of my pet peeve. It's generally a good idea to document as many notable changes as possible in the PR desc so that it's possible for contribs to figure out the origin of a given change by simply searching for key words via github's search feature, instead of a given change having to be manually tracked down via git blame, or in some cases, git history. The contents of this approval comment should make the PR show up in the searches for those given keywords, though, so NBD here honestly. Changes not documented in the PR desc include~~

- ~~Chemistry-Readme.dm was removed (tbf it's hilariously out of date in the context of baycode, but it'd be worth mentioning regardless just in case a future contrib wonders where it went)~~
- ~~Paint was made to use pigment reagents instead of being static per-object colors~~
- ~~Things that interact with blood vessels no longer ignore the presence of non-blood reagents~~

~~I've skimmed over a good chunk of the PR's contents, since the majority of it is just a mass replacement of `/datum/reagent` with `/decl/reagent`, so there may be a thing or two I've missed in terms of potentially undocumented changes, but aside from that, there's no flaws or noticeable bugs that stand out.~~

(EDIT: above was added to PR desc!)

LGTM

MistakeNot4892:
````Things that interact with blood vessels no longer ignore the presence of non-blood reagents```` Other two are fair enough, but this isn't a significant change as in theory nothing else should be using the blood vessel.

deathride58:
Aye, true, true, but still worth a mention

PR #379, opened 2020-04-27
Makes light switch turn on on the example map

quardbreak:
Less pain to debug things in dark.


PsiOmegaDelta (approved):
Less pain to debug things in the light I assume

PR #382, opened 2020-04-28
Replaces tracksuit sprites with Deer's cooler version

comma:
Removed green and white ones so they all look sameish, and axed some baylore remnant.
Also made jackets not cover your legs.
image
image


deathride58 (approved):
очень круто

PR #386, opened 2020-04-29
Generalize fabricators output direction

quardbreak:
So we can make directional subtypes without overriding proc. Useful when we need output something on conveyors.


code/modules/fabrication/fabricator_build.dm
```diff
@@ -16,6 +16,9 @@
 
 /obj/machinery/fabricator/proc/do_build(var/datum/fabricator_recipe/recipe, var/amount)
 	. = recipe.build(get_turf(src), amount)
+	if(output_dir)
```
MistakeNot4892, line 19: Might be worth adding a button in the interface to set or clear this.

MistakeNot4892 (approved):

PR #393, opened 2020-04-29
Removed uncompiled icon files.

MistakeNot4892:
Deleted the icons folder and checked out icons in the error log until it clean compiled.


PsiOmegaDelta (approved):
The icon diff bot is struggling

PR #395, opened 2020-04-30
Adds a check for when network machine is bothered too early

comma:
It seems that during init something can try to change its power state before it's inited properly and gets a network device extension.


MistakeNot4892 (dismissed):
Assuming it passes.

biscuitWizard (approved):
yum guard checks

MistakeNot4892 (approved):

PR #399, opened 2020-04-30
Nar'Sie Spawn Animation from vgstation

quardbreak:
Credits to a original author of this animation and PR (vgstation-coders/vgstation13#13801)


MistakeNot4892 (approved):
Not sure we're even keeping cult but I suppose there's no harm in this.

PR #401, opened 2020-04-30
Network machines now produce heat as a byproduct of functioning.

biscuitWizard:
Kinda just farting around. Copy+pasted some code from telecomms over to networking.

Network machines produce heat now, I made the amount quite low. If the environment the machines are in gets too hot they stop working.

Very simple. IMO all machines should just do this, but maybe a different PR?

### DIFF
--- a/code/modules/modular_computers/networking/machinery/_network_machine.dm
+++ b/code/modules/modular_computers/networking/machinery/_network_machine.dm
@@ -5,16 +5,44 @@
 	var/main_template = "network_mainframe.tmpl"
 	var/network_device_type =  /datum/extension/network_device
 	var/error
-	
+
 	var/initial_network_id
 	var/initial_network_key
 	var/lateload
+
+	var/produces_heat = TRUE	// If true, produces and is affected by heat.
+	var/inefficiency = 0.12		// How much power is waste heat.
+	var/heat_threshold = 358.15	// At what temperature the machine will lock up.
 
 /obj/machinery/network/Initialize()
 	. = ..()
 	set_extension(src, network_device_type, initial_network_id, initial_network_key, NETWORK_CONNECTION_WIRED, !lateload)
 	if(lateload)
 		return INITIALIZE_HINT_LATELOAD
+
+/obj/machinery/network/operable()
+	var/turf/simulated/L = loc
+	if(istype(L))
+		var/datum/gas_mixture/env = L.return_air()
+		if(env.temperature >= heat_threshold)
+			return FALSE
+	return ..()
+
+/obj/machinery/network/proc/produce_heat()
+	if (!produces_heat || !use_power || !operable())
+		return
+	var/turf/simulated/L = loc
+	if(istype(L))
+		var/datum/gas_mixture/env = L.return_air()
+		var/transfer_moles = 0.25 * env.total_moles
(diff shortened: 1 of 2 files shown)


MistakeNot4892:
Would like relays to default to not producing any heat just so I don't have to remap them all on Exodus (as most of them are in areas without climate control).

MistakeNot4892 (dismissed):

code/modules/modular_computers/networking/machinery/_network_machine.dm
```diff
@@ -5,17 +5,45 @@
...
 	if(lateload)
 		return INITIALIZE_HINT_LATELOAD
 
+/obj/machinery/network/operable()
+	var/turf/simulated/L = loc
+	if(istype(L))
+		var/datum/gas_mixture/env = L.return_air()
+		if(env.temperature >= heat_threshold)
+			return FALSE
+	return ..()
```
afterthought2: Nothing checks operable unless you're checking it manually someplace. Are you sure this does anything?

code/modules/modular_computers/networking/machinery/_network_machine.dm
```diff
@@ -5,17 +5,45 @@
...
 	var/error
-	
+
 	var/initial_network_id
 	var/initial_network_key
 	var/lateload
 
+	var/produces_heat = TRUE	// If true, produces and is affected by heat.
+	var/inefficiency = 0.12		// How much power is waste heat.
+	var/heat_threshold = 358.15	// At what temperature the machine will lock up.
```
afterthought2: Please use temperatures of the form `T0C + X` or `X CELSIUS` to improve readability.

code/modules/modular_computers/networking/machinery/_network_machine.dm
```diff
@@ -5,17 +5,45 @@
...
+		var/datum/gas_mixture/env = L.return_air()
+		var/transfer_moles = 0.25 * env.total_moles
+		var/datum/gas_mixture/removed = env.remove(transfer_moles) // Air is moved through computer vents.
+		if(removed)
+			removed.add_thermal_energy(idle_power_usage * inefficiency)
+		env.merge(removed)
+
+/obj/machinery/network/Process()
+	. = ..()
+	produce_heat()
```
afterthought2, line 47: Should check that power is on and not broken here.

code/modules/modular_computers/networking/machinery/_network_machine.dm
```diff
@@ -5,17 +5,45 @@
...
+	if(istype(L))
+		var/datum/gas_mixture/env = L.return_air()
+		var/transfer_moles = 0.25 * env.total_moles
+		var/datum/gas_mixture/removed = env.remove(transfer_moles) // Air is moved through computer vents.
+		if(removed)
+			removed.add_thermal_energy(idle_power_usage * inefficiency)
+		env.merge(removed)
+
+/obj/machinery/network/Process()
+	. = ..()
```
afterthought2: The default return value is `PROCESS_KILL`, so this will fire exactly once.

afterthought2:
Having all machines do this is expensive. On large maps there are tens of thousands of machines, and this forces them to process every tick (that's intentionally avoided by default, even for power). If you want to do this at a global level, I'd move the logic to areas instead, based on power consumption levels (or apcs if you prefer).

afterthought2 (approved):

PR #402, opened 2020-04-30
Added NTOS-v2 cyberbrains.

biscuitWizard:
As the title says.


comma:
Need to add attackby on augment to handle removing/adding those stock parts because otherwise it'll be unrepairable. Unless that's the intention so that whole augment must be swapped out completely.
also kitbashed a sprot 
image

code/modules/augment/active/cyberbrain.dm
```diff
@@ -0,0 +1,36 @@
...
+	var/list/starting_stock_parts = list(
+		/obj/item/stock_parts/computer/processor_unit/small,
+		/obj/item/stock_parts/computer/hard_drive/silicon,
+		/obj/item/stock_parts/computer/network_card
+	)
+
+/obj/item/organ/internal/augment/active/cyberbrain/Initialize()
+	set_extension(src, /datum/extension/interactive/ntos/implant)
+	for(var/T in starting_stock_parts)
+		stock_parts += new T(src)
```
Techhead0: No parent call on Initialize? (Or Destroy, for that matter)

EDIT: Turns out this is part of why it fails Travis. Thanks, Travis

biscuitWizard:
> Need to add attackby on augment to handle removing/adding those stock parts because otherwise it'll be unrepairable. Unless that's the intention so that whole augment must be swapped out completely.
> also kitbashed a sprot
> image

I was initially not going to allow that, and instead do presets. but honestly I think the extra work really makes the cyberbrain way cooler.  Accepted.

code/modules/augment/active/cyberbrain.dm
```diff
@@ -0,0 +1,117 @@
+/obj/item/organ/internal/augment/active/cyberbrain
+	name = "cyberbrain module"
+	action_button_name = "Access cyberbrain"
+	icon_state = "cyberbrain"
+	allowed_organs = list(BP_AUGMENT_HEAD)
+	augment_flags = AUGMENTATION_MECHANIC
+	var/list/stock_parts = list()
+	var/list/starting_stock_parts = list(
+		/obj/item/stock_parts/computer/processor_unit/small,
+		/obj/item/stock_parts/computer/hard_drive/silicon,
```
comma: Should be /obj/item/stock_parts/computer/hard_drive/small, silicon subtype comes preloaded with some programs people might not have access too.

comma (approved):

PR #403, opened 2020-04-30
Auto-throw shurikens if on harm intent

comma:
Cause screw clicking throw mode button


Techhead0:
Very cool, but I'm wondering what happens to point-blank targets. I suspect it might first hit them with the star and then immediately throw it

comma:
Actually nope, afterattack isn't called if attack happened, you jsu stab em

Techhead0:
The semantic weirdities of "afterattack isn't called if attack happened" aside, nice to know

MistakeNot4892 (approved):

PR #405, opened 2020-04-30
Circuitboards now tell what parts you need on examine

comma:


MistakeNot4892 (approved):

afterthought2:
The power thing is potentially inaccurate, but whatever.

PR #406, opened 2020-04-30
Stock parts health

comma:
Added max_health var to track how damaged things are, since initial(health) is meaningless with material based health.
Added multiplier for how much of material's integrity is translated into health. Default is mimicing same values it had.

Added bit more gradation to damage descriptors when examining stock parts.
Shielding components have twice the health.
Made armor component requrie plasteel (also means it gets more health)

### DIFF
--- a/code/game/machinery/_machines_base/machinery.dm
+++ b/code/game/machinery/_machines_base/machinery.dm
@@ -387,8 +387,8 @@ Class Procs:
 		var/line = "<span class='notice'>	[C.name]</span>"
 		if(!C.health)
 			line = "<span class='warning'>	[C.name] (destroyed)</span>"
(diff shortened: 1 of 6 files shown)


maps/bearcat/bearcat_define.dm
```diff
@@ -20,7 +20,7 @@
 	allowed_spawns = list("Cryogenic Storage")
 	default_spawn = "Cryogenic Storage"
 	use_overmap = 1
-	num_exoplanets = 1
+	num_exoplanets = 0
```
Techhead0: Accidental commit?

comma: yeep

MistakeNot4892 (approved):

PR #413, opened 2020-05-01
Fixes suits fucked up mob overlays sometimes

comma:
Now we check before we shrek ourselves, instead of setting icon_state to possible null and THEN trying to revert to possibly wrong iconstate


MistakeNot4892 (approved):

afterthought2:
Wait, so: old logic was slot list state || item_state || icon_state, while new logic is slot list state || item_state || null? Is that really correct?

comma:
no see image is done in ..() with icon_state so it was
slot list state || item_state OH BUT IF item_state is null revert to icon_state
now its 
slot list state || item_state only if it's set, otherwise keep icon_state

So same result, just checking before doing something instead of checking after and reverting (badly in case ..() did different icon)

PR #415, opened 2020-05-01
Adds check for not having separate inventory icon for new icon system

comma:
Makes porting things to it much easier because no need to duplicate / rename obj icon

### DIFF
--- a/code/game/objects/items/item_icon_experimental.dm
+++ b/code/game/objects/items/item_icon_experimental.dm
@@ -1,9 +1,11 @@
 // This file is an experiment in changing the way item icons are handled.
 // Not really expecting it to work out of the box, so we'll see how it goes
 // with a handful of specific items.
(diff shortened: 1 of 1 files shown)


MistakeNot4892 (approved):

code/game/objects/items/item_icon_experimental.dm
```diff
@@ -1,9 +1,11 @@
 // This file is an experiment in changing the way item icons are handled.
 // Not really expecting it to work out of the box, so we'll see how it goes
 // with a handful of specific items.
+var/list/if_has_inventory_icon_cache = list()  // for checking if we have special in-inventory HUD state. Cached cause asking icons is expensive
```
PsiOmegaDelta, line 4: This would mean that every `/obj/item` initialize a list upon creation, even if never needed

afterthought2: This appears to be a global list.

comma: it's indoot global

deathride58 (approved):

PR #416, opened 2020-05-01
Made labcoats use colored overlays

comma:
Cuts down massively on number of icons they need


code/modules/clothing/suits/labcoat.dm
```diff
@@ -30,34 +56,22 @@
 /obj/item/clothing/suit/storage/toggle/labcoat/mad
 	name = "The Mad's labcoat"
```
MistakeNot4892, line 57: This is such a bad name. What is this even for?

code/modules/clothing/suits/labcoat.dm
```diff
@@ -30,34 +56,22 @@
...
 	desc = "It makes you look capable of konking someone on the noggin and shooting them into space."
-	icon_state = "labgreen_open"
-	icon_open = "labgreen_open"
-	icon_closed = "labgreen"
-
-/obj/item/clothing/suit/storage/toggle/labcoat/genetics
```
MistakeNot4892, line 37: rip genetics

MistakeNot4892 (approved):

code/modules/clothing/suits/labcoat.dm
afterthought2: Apparently dates back to before 2011. Primordial code.

deathride58 (approved):

PR #422, opened 2020-05-01
Cleaning up some guard checks.

biscuitWizard:
All of these checks are things that tripped up persistence game mode. Either they are too hostile to parameterless construction, or their null checks were not good enough.

Updated for friendliness.

### DIFF
--- a/code/datums/wires/wires.dm
+++ b/code/datums/wires/wires.dm
@@ -27,7 +27,6 @@ var/list/wireColours = list("red", "blue", "green", "darkred", "orange", "brown"
 	src.holder = holder
 	if(!istype(holder, holder_type))
 #ifdef DISABLE_DEBUG_CRASH
-		to_world_log("Our holder is null/the wrong type!") 
 		return ..()
 #else
 		CRASH("Our holder is null/the wrong type!")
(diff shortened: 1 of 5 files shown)


afterthought2:
Historically `to_world_log` and similar has been pretty useless, as it's hard to parse for debug information while simultaneously spamming admins. One construction that works for me is `#ifdef UNIT_TEST CRASH`, but I'm not certain others are as into it as I am.

biscuitWizard:
> Historically `to_world_log` and similar has been pretty useless, as it's hard to parse for debug information while simultaneously spamming admins. One construction that works for me is `#ifdef UNIT_TEST CRASH`, but I'm not certain others are as into it as I am.

I'm okay looking at alternatives, but crashing is unacceptable. I need to be able to construct things with parameterless New() downstream.

code/datums/wires/wires.dm
```diff
@@ -27,7 +26,12 @@ var/list/wireColours = list("red", "blue", "green", "darkred", "orange", "brown"
 	..()
 	src.holder = holder
 	if(!istype(holder, holder_type))
+#ifdef DISABLE_DEBUG_CRASH
+		to_world_log("Our holder is null/the wrong type!")
```
MistakeNot4892: CRASH_WITH supplies debugging info (stack trace), this just prints a sourceless message without saying who 'our' refers to.

biscuitWizard: we're pro engineers. we don't need stack traces. honestly I was tempted to remove the to_world_log entirely.

MistakeNot4892: That would be preferrable to the current line.

comma (approved):

PR #427, opened 2020-05-02
Refactoring ex_act() procs.

MistakeNot4892:
- Refactors `ex_act()` into `explosion_act()` which attempts to preserve a call chain all the way to `/atom`.
- Moves reagent explosions down to `/atom`.
- Moves `/obj/structure/destroyed()` to `/atom/physically_destroyed()`.
- Moves several `dump_contents()` procs to `/atom/dump_contents()` and calls it in physical destruction.
- Generalizes propagation of explosions to contents using an `/atom/get_contained_external_atoms()` proc.

Not tested yet. This was meant to be a small PR.

I have concerns that `get_contained_external_atoms()` is going to need overrides for `/obj/machinery` and anything with contents-held atoms that shouldn't be interacted with (similar to how humans previously farted out all their organs when Drop Everything was used).


afterthought2 (dismissed):

MistakeNot4892:
This still causes or exposes a bad runtime problem with proxy triggers.

deathride58 (dismissed):

deathride58:
Probably would be worth doing some further investigations on whatever's causing those runtimes. There's nothing in this PR that really touches proximity triggers (as far as I can tell at a glance, at least), so I'm leaving my approval for now, since it seems the issue may lie elsewhere. Having more eyes on this PR would probably be a very good idea, though.

afterthought2 (dismissed):
Fuck it. Can deal with proximity triggers separately.

afterthought2 (dismissed):
Well, proximity trigger updates are in. If there are proximity trigger issues caused by this, can open an issue and address from there.

afterthought2 (approved):

PR #429, opened 2020-05-02
Adding populate parts override hook to atoms subsystem.

biscuitWizard:
This isn't used in Nebula, and shouldn't change any functionality, but is helpful for overriding functionality in mods.

### DIFF
--- a/code/controllers/subsystems/atoms.dm
+++ b/code/controllers/subsystems/atoms.dm
@@ -8,6 +8,8 @@ SUBSYSTEM_DEF(atoms)
 	init_order = SS_INIT_ATOMS
 	flags = SS_NO_FIRE | SS_NEEDS_SHUTDOWN
 
+	// override and GetArguments() exists for mod-override/downstream hook functionality.
+	// Useful for total-overhaul type modifications.
 	var/override_populate_parts = FALSE
 	var/populate_parts = TRUE
 
--- a/code/controllers/subsystems/atoms.dm
+++ b/code/controllers/subsystems/atoms.dm
@@ -8,6 +8,9 @@ SUBSYSTEM_DEF(atoms)
 	init_order = SS_INIT_ATOMS
 	flags = SS_NO_FIRE | SS_NEEDS_SHUTDOWN
 
+	var/override_populate_parts = FALSE
+	var/populate_parts = TRUE
+
 	var/atom_init_stage = INITIALIZATION_INSSATOMS
 	var/old_init_stage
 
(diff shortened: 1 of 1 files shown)


code/controllers/subsystems/atoms.dm
```diff
@@ -8,6 +8,9 @@ SUBSYSTEM_DEF(atoms)
 	init_order = SS_INIT_ATOMS
 	flags = SS_NO_FIRE | SS_NEEDS_SHUTDOWN
 
+	var/override_populate_parts = FALSE
```
MistakeNot4892: Possibly a brief comment explaining that it is for downstream use rather than Neb use?

code/controllers/subsystems/atoms.dm
```diff
@@ -96,6 +100,17 @@ SUBSYSTEM_DEF(atoms)
 
 	return qdeleted || QDELING(A)
 
+/datum/controller/subsystem/atoms/proc/GetArguments(atom/A, list/mapload_arg, created=TRUE)
+	var/list/arguments = mapload_arg.Copy()
+	if(created && created_atoms[A])
+		arguments += created_atoms[A]
+	if(override_populate_parts && istype(A, /obj/machinery))
```
PsiOmegaDelta: Should probably use a proc-override to acquire additional arguments rather than `istype()`-checks.
Open-closed principle and all that.

biscuitWizard: Think it's worth removing override_populate_parts entirely and leaving it up to mod implementor to override GetArguments()?

PsiOmegaDelta: It does look like the vars wouldn't be needed with a carefully implemented `/atom/proc/AdjustInitArguments(var/list/arguments)` implementation (or similar)

On the other hand, having an `adjust_init_arguments `-var (or config-setting) would mean the proc-calls could be skipped on servers that don't use it at all.
I.e:
```
/atom/proc/AdjustInitializeArguments(list/arguments)
    // Lists are passed by reference so can simply modify the arguments list without returning it

/datum/controller/subsystem/atoms
    var/adjust_init_arguments = TRUE/FALSE

/datum/controller/subsystem/atoms/proc/GetArguments(atom/A, list/mapload_arg, created=TRUE)
	var/list/arguments = mapload_arg.Copy()
	if(created && created_atoms[A])
		arguments += created_atoms[A]
	if(adjust_init_arguments)
        A.AdjustInitializeArguments(arguments)
    return arguments
```

code/controllers/subsystems/atoms.dm
```diff
@@ -96,6 +99,19 @@ SUBSYSTEM_DEF(atoms)
 
 	return qdeleted || QDELING(A)
 
+// override and GetArguments() exists for mod-override/downstream hook functionality.
+// Useful for total-overhaul type modifications.
+/atom/proc/AdjustInitializeArguments(list/arguments)
+	// Lists are passed by reference so can simply modify the arguments list without returning it
+
+/datum/controller/subsystem/atoms/proc/GetArguments(atom/A, list/mapload_arg, created=TRUE)
```
MistakeNot4892, line 107: Shouldn't this default to just returning `mapload_arg` if `adjust_init_arguments` isn't set? Otherwise you're adding a list copy, a var check, an assoc list lookup and a list op to the init for every atom on Neb for no gain.

MistakeNot4892 (approved):

MistakeNot4892 (approved):
This is more or less the same behavior as previous when the override isn't set. I'm ok with this going in.

PR #430, opened 2020-05-02
Makes shoes use colored overlays

comma:
God that's a lot of repaths im sorry.
Culls some unused states too.


code/modules/clothing/shoes/colour.dm
```diff
@@ -1,104 +1,113 @@
...
+	var/image/ret = ..()
+	if(markings_color)
+		var/mutable_appearance/MA = new()
+		MA.icon = ret.icon
+		MA.icon_state = markings_icon
+		MA.color = markings_color
+		MA.appearance_flags = RESET_COLOR
+		MA.layer = FLOAT_LAYER
+		MA.plane = FLOAT_PLANE
+		ret.overlays += MA
```
PsiOmegaDelta: This look rather identical to https://github.com/NebulaSS13/Nebula/pull/430/files#diff-d144d7ca9389f824d315c483c1639f29R19-R28

code/modules/clothing/shoes/colour.dm
```diff
@@ -1,104 +1,113 @@
...
+	var/markings_icon = "shoes_coloring"
+	var/markings_color
+
+/obj/item/clothing/shoes/color/Initialize()
+	. = ..()
+	update_icon()
+
+/obj/item/clothing/shoes/color/on_update_icon()
+	if(markings_color)
+		overlays.Cut()
```
MistakeNot4892: Might be worth using `cut_overlays()` and `add_overlay()` since you're redefining this entirely here.

code/modules/clothing/shoes/colour.dm
comma: Yeah, sorta thinking of unifying them with some /clothing level handling

comma: Moved it to /clothing

MistakeNot4892 (approved):

PR #432, opened 2020-05-02
Added MMO-style protection areas.

biscuitWizard:
* Walls cannot be constructed/deconstructed in /area/protected where safe_zone is true
* Same for walls
* Same for anything made from a stack out of hand.
* Same for floors.
* Walls, floors, etc are all immune to damage. Machines cannot be deconstructed.

Anti-griefing tool. Could be used for timeout zones or something on a round-based server.

### DIFF
--- a/code/_helpers/protection.dm
+++ b/code/_helpers/protection.dm
@@ -1,0 +1,14 @@
+/atom/proc/isProtected(var/mob/user)
+	var/area/A
+	if(istype(src, /turf))
+		A = get_area(src)
+	else
+		A = get_area(loc)
+		
+	var/protected = A.area_flags & AREA_FLAG_PROTECTED
+	if(!protected)
+		return FALSE
+	if(!user)
+		return TRUE
+	// Check permission whitelists for this user.
+	return TRUE // TODO: actually implement.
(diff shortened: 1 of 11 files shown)


code/_helpers/protection.dm
```diff
@@ -0,0 +1,16 @@
...
+		A = get_area(loc)
+		
+	if(!A.safe_zone)
+		return FALSE
+	if(!user)
+		return TRUE
+	// Check permission whitelists for this user.
+	return TRUE // TODO: actually implement.
+
+/area/protected
```
MistakeNot4892: You don't typecheck `A`. This should really be an area flag rather than a bool.

PsiOmegaDelta:
This implementation honestly feel very brittle, with checks needed everywhere.
That "everywhere"-part isn't a very good thing either.

biscuitWizard:
> This implementation honestly feel very brittle, with checks needed everywhere.
> That "everywhere"-part isn't a very good thing either.

I don't feel this is a constructive comment.

code/_helpers/protection.dm
biscuitWizard: I wasn't aware of area_flags, I'll see about converting it, actually.

PsiOmegaDelta:
Very well: There needs to be a more centralized way to check for protection than dropping proc-checks everywhere, be that via event registration or otherwise.

code/_helpers/protection.dm
biscuitWizard: Converted.

PR #434, opened 2020-05-03
Added GUID-goodness to sequential_id.

biscuitWizard:
Need round-based security to prevent integer overflows and collisions on persistence.

Added guid implementation that takes into account epoc, 1/10 time, id sequence, and signed variance with 24 bits of additional entropy.

### DIFF
--- a/code/__defines/computers.dm
+++ b/code/__defines/computers.dm
@@ -1,5 +1,4 @@
-#define RANDBYTE  num2hex(rand(1,255))
-#define NETWORK_MAC   "[RANDBYTE]-[RANDBYTE]-[RANDBYTE]-[RANDBYTE]"
+#define NETWORK_MAC   "[sequential_id("NETWORK_MAC")]"
 									// Network allowed actions
 #define NETWORK_SOFTWAREDOWNLOAD 	1 	// Downloads of software
 #define NETWORK_COMMUNICATION 		2	// Communication (messaging)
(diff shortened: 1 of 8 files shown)


PsiOmegaDelta (changes requested):
The following lines, and anything that might depend on them need to be adjusted, due expectations of being a number or because of general user unfriendliness:
https://github.com/NebulaSS13/Nebula/blob/dev/code/modules/atmospherics/components/unary/vent_pump.dm#L247
https://github.com/NebulaSS13/Nebula/blob/dev/code/modules/atmospherics/components/unary/vent_scrubber.dm#L106
https://github.com/NebulaSS13/Nebula/blob/dev/code/modules/mob/living/simple_animal/friendly/mouse.dm#L74
https://github.com/NebulaSS13/Nebula/blob/dev/code/game/machinery/newscaster.dm#L165

You could instead consider having a third, separate, `sequential_guid()` (or `_ uuid`) implementation and use that were appropriate.

biscuitWizard:
> The following lines, and anything that might depend on them need to be adjusted, due expectations of being a number or because of general user unfriendliness:
> https://github.com/NebulaSS13/Nebula/blob/dev/code/modules/atmospherics/components/unary/vent_pump.dm#L247
> https://github.com/NebulaSS13/Nebula/blob/dev/code/modules/atmospherics/components/unary/vent_scrubber.dm#L106
> https://github.com/NebulaSS13/Nebula/blob/dev/code/modules/mob/living/simple_animal/friendly/mouse.dm#L74
> https://github.com/NebulaSS13/Nebula/blob/dev/code/game/machinery/newscaster.dm#L165
> 
> You could instead consider having a third, separate, `sequential_guid()` (or `_ uuid`) implementation and use that were appropriate.

It already is sequential.

biscuitWizard:
Requested changes denied. And those files have all been adapted already with my current implementation.

PR #435, opened 2020-05-03
alternate try at providing modularity to unique ids.

biscuitWizard:
lets modders or downstreams determine how things are generated by overriding uniqueness_generator. assumes all things returned will be strings.

### DIFF
--- a/code/datums/repositories/unique.dm
+++ b/code/datums/repositories/unique.dm
@@ -1,3 +1,5 @@
+#define RANDBYTE  num2hex(rand(1,255))
+
 var/repository/unique/uniqueness_repository = new()
 
 /repository/unique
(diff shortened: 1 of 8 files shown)


code/datums/repositories/unique.dm
```diff
@@ -20,6 +22,9 @@ var/repository/unique/uniqueness_repository = new()
 /datum/uniqueness_generator/proc/Generate()
 	return
 
+/datum/uniqueness_generator/hex/Generate()
+	return "[RANDBYTE]-[RANDBYTE]-[RANDBYTE]-[RANDBYTE]"
```
PsiOmegaDelta, line 26: While the probability is exceedingly small that a duplicate will occur I'll note that there's nothing inherent in the system that prevents them.
See `/datum/uniqueness_generator/id_random/Generate()` as an example

biscuitWizard: I don't want to change CHINSKY's code more than necessary. Resolving that problem is outside of the scope of this PR, I feel.

PsiOmegaDelta: Merely making a note

biscuitWizard: Comment rescinded. In that case I agree completely.

comma (approved):

PR #445, opened 2020-05-04
Give forensics borg a clerical gripper

HeyBanditoz:
The gripper in question can now pick up /obj/item/forensics/sample


Techhead0:
I think it can just be combined with the clerical gripper, no reason to add another hyper-specialized gripper just for this

MistakeNot4892:
Agreed with Techhead.

comma (approved):

PR #446, opened 2020-05-04
Silicon can climb like humans do

HeyBanditoz:


MistakeNot4892 (dismissed):
I wrote a chunk of this so my approval probably shouldn't count.

code/game/atoms.dm
```diff
@@ -460,20 +460,13 @@ its easier to just keep the beam vertical.
...
-		return 0
-	if (issilicon(user))
-		to_chat(user, "<span class='notice'>You need hands for this.</span>")
-		return 0
-	return 1
+/mob/proc/can_touch(var/atom/touching)
+	if(!touching.Adjacent(src) || incapacitated())
+		return FALSE
+	if(restrained() || buckled)
+		to_chat(src, SPAN_WARNING("You need your hands and legs free for this."))
```
afterthought2: Shouldn't this message be at a level that ensures that the mob has hands and legs?

HeyBanditoz: Could the message just be more abstract instead? Split up the check and say "You are restrained." or "You are buckled." maybe.

MistakeNot4892 (approved):

PR #451, opened 2020-05-04
Adding a utility frame robotic species.

MistakeNot4892:
Adds a simple machine species. Framework for merging crew robots into them later.

Default appearance is as follows, but the plating, body and various markings are colourable.
image

~~Another one I could use some eyes on, as they are insisting on rendering icons with BLEND_ADD despite debug lines in the external organ icon procs showing that nothing that uses BLEND_ADD is being called, and BLEND_MULTIPLY is being used instead.~~ For reference `BLEND_*` and `ICON_*` are not interchangeable.


MistakeNot4892:
Ready for review.

afterthought2 (approved):

PR #299, opened 2020-04-14
Nanoui misc changes

quardbreak:
DNM; Ports some nanoui changes from Eris:
* Force open optimization, window focus on open
* tgui css
* panic button to close all mob's uis

PR #453, opened 2020-05-04
tgui.css and textless icon for NanoUI

quardbreak:
#299 but without a big focus changes. This adds some tgui styles for NanoUI and one style for just icons. Both from Eris.


comma (approved):
oooh yeah the buttons without text are pretty damn ugly, that's neat

PR #454, opened 2020-05-04
Material armor

comma:
Materials now generate platonic 'armor values' based on their properties.
Items can opt to pull armor values from their material with a given multiplier to values (for things taht shouldn't be as protective as a suit of armor)
Makes crafted boots use it and cloaks (half armor).


code/modules/materials/material_armor.dm
```diff
@@ -0,0 +1,43 @@
...
+	var/list/armor = basic_armor.Copy()
+	if(coef == 1)
+		return armor
+	for(var/val in armor)
+		armor[val] = round(coef * armor[val], 3)
+		if(armor[val] < 1)
+			armor -= val
+	return armor
+
+/client/verb/test_armor()
```
MistakeNot4892: Stop committing test verbs!

comma: people have right to test!

MistakeNot4892 (dismissed):

MistakeNot4892 (approved):

PR #455, opened 2020-05-04
converted modular computer assembly state into an extension.

biscuitWizard:


biscuitWizard:
All scanners (ore, health, etc) can use this extension for batteries.

code/modules/modular_computers/computers/subtypes/dev_laptop.dm
```diff
@@ -2,25 +2,26 @@
...
-	w_class = ITEM_SIZE_NORMAL
 	var/icon_state_closed = "laptop-closed"
 	interact_sounds = list("keyboard", "keystroke")
 	interact_sound_volume = 20
 	
+
 /obj/item/modular_computer/laptop/Initialize()
 	. = ..()
 	screen_on = anchored
+	var/datum/extension/assembly/modular_computer/assembly = get_extension(src, /datum/extension/assembly)
```
comma: maybe make these subtypes of extension and give modular computers 'type of assembly' var? Feels kinda iffy to do this in init. Can just put em in same files as devices since it's just few lines of definition with no new logic

code/modules/modular_computers/computers/modular_computer/variables.dm
```diff
@@ -26,33 +20,19 @@
 	randpixel = 0											// And no random pixelshifting on-creation either.
 	var/icon_state_unpowered = null							// Icon state when the computer is turned off
 	var/max_hardware_size = 0								// Maximal hardware size. Currently, tablets have 1, laptops 2 and consoles 3. Limits what hardware types can be installed.
```
comma: is this still used?

code/modules/modular_computers/computers/modular_computer/variables.dm
```diff
@@ -5,15 +5,9 @@
 	desc = "A modular computer. You shouldn't see this."
 
 	var/enabled_by_default = FALSE							// Start enabled durint Initialization
-	var/enabled = 0											// Whether the computer is turned on.
 	var/screen_on = 1										// Whether the computer is active/opened/it's screen is on.
-	var/hardware_flag = 0									// A flag that describes this device type
-	var/last_power_usage = 0								// Last tick power usage of this computer
 	var/computer_emagged = FALSE							// Whether the computer is emagged.
 	var/apc_powered = FALSE									// Set automatically. Whether the computer used APC power last tick.
```
comma: Wasn't this moved to assembly?

code/modules/modular_computers/computers/modular_computer/core.dm
```diff
@@ -23,48 +21,53 @@
...
-		install_default_programs()
-	if(scanner)
-		scanner.do_after_install(null, src)
 	update_icon()
 	update_verbs()
 	update_name()
 	if(enabled_by_default)
 		enable_computer()
+	..()
+	return INITIALIZE_HINT_LATELOAD
```
comma, line 62: Why lateload it?

code/modules/modular_computers/computers/modular_computer/core.dm
```diff
@@ -23,48 +21,53 @@
...
+	var/datum/extension/assembly/assembly = get_extension(src, /datum/extension/assembly)
+	for(var/T in default_hardware)
+		assembly.try_install_component(null, new T(src))
 
 // Used to install preset-specific programs
 /obj/item/modular_computer/proc/install_default_programs()
-	return 1
-
-/obj/item/modular_computer/proc/install_default_programs_by_job(var/mob/living/carbon/human/H)
+	var/mob/living/carbon/human/H = get_holder_of_type(src, /mob)
```
comma, line 30: okay this is merging two things that shouldn't be really merged this way IMO
default programs by job was called only on custom loadout devices
This makes basic default programs require a human holder (early returns if no human or no job) which means that pre-mapped presets (the main customer of this proc) are allll fucked.
Job programs should be installed on TOP of default ones, not blocking default stuff from existing

code/modules/modular_computers/computers/modular_computer/assembly_computer.dm
```diff
@@ -0,0 +1,92 @@
...
+/datum/extension/assembly/modular_computer
+	var/max_hardware_size
+	var/hardware_flag
+	var/bsod
+	assembly_name = "computer"
+	max_parts = list(
+		PART_BATTERY 	= 1,
+		PART_CPU		= 1,
+		PART_NETWORK	= 1,
+		PART_HDD		= 1,
```
comma, line 17: Have you tried actually using the removable storage? I have a suspicion it might not be working because PART_DRIVE is actually subtype of PART_HDD so it might be getting blocked by the max parts check.
Low effort fix would be changing the order in list, high effort repathing them to some common storage parent. Im fine with either.

code/modules/modular_computers/computers/modular_computer/core.dm
biscuitWizard: /obj/item/modular_computer/LateInitialize()
	. = ..()
	install_default_hardware()
	install_default_programs()

Initialization for programs and hardware has to happen later because of race condition reasons.

code/modules/modular_computers/computers/modular_computer/assembly_computer.dm
biscuitWizard: You're right. It is blocked.

biscuitWizard: I have fixed this now.

comma (dismissed):

code/modules/modular_computers/ntos/files.dm
```diff
@@ -267,6 +263,52 @@
...
+
+/datum/file_storage/disk_removable/check_errors()
+	. = ..()
+	if(.)
+		return
+	var/obj/item/stock_parts/computer/drive_slot/drive_slot = os.get_component(PART_D_SLOT)
+	if(!drive_slot)
+		return "HARDWARE ERROR: No compatible device found"
+	if(!drive_slot.check_functionality())
+		return "NETWORK ERROR: [drive_slot] is non-operational"
```
comma: >network

code/datums/extensions/assembly/assembly.dm
```diff
@@ -0,0 +1,122 @@
...
+ * Basic usage:
+ *	1. Have Processing enabled on your item/machine. Call Process() on the assembly datum.
+ *	2. Call attackby on the assembly datum to handle tool interactions.
+ *	3. Recommended to also hook into ex_act, and any of the assembly_damage functions on base level for assembly to handle damage.
+ */
+
+/datum/extension/assembly
+	base_type = /datum/extension/assembly
+	var/assembly_name = "assembly"
+	var/list/parts = list()					// List of all components.
```
MistakeNot4892, line 14: Is it possible to LAZY these lists? This seems like an extension that will see a lot of use and there's at least three of them here.

MistakeNot4892 (approved):
I didn't actually read most of this PR.

PR #459, opened 2020-05-05
Borg fire and o2 icons work

HeyBanditoz:
Checks if air is safe to breathe

### DIFF
--- a/code/modules/mob/living/silicon/robot/life.dm
+++ b/code/modules/mob/living/silicon/robot/life.dm
@@ -234,9 +234,23 @@
 			else
 				src.bodytemp.icon_state = "temp-2"
 
-//Oxygen and fire does nothing yet!!
-//	if (src.oxygen) src.oxygen.icon_state = "oxy[src.oxygen_alert ? 1 : 0]"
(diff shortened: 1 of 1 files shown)


MistakeNot4892 (changes requested):
Need to gate the breath check behind the poisoning check or it makes the poison check pointless.

code/modules/mob/living/silicon/robot/life.dm
```diff
@@ -234,9 +234,22 @@
...
+				src.fire.icon_state = "fire0"
+			else
+				src.fire.icon_state = "fire1"
+	if(oxygen && environment)
+		var/datum/species/species = all_species[GLOB.using_map.default_species]
+		if(!species.breath_type || environment.gas[species.breath_type] >= species.breath_pressure)
+			src.oxygen.icon_state = "oxy0"
+			for(var/gas in species.poison_types)
+				if(environment.gas[gas])
+					src.oxygen.icon_state = "oxy1"
```
MistakeNot4892, line 250: Can `break` here.

MistakeNot4892 (approved):

PR #462, opened 2020-05-05
Adds SScomputer_networks for queueing connecttions

comma:
Final solution to the init racing we've been having with routers-relays-devices.
Now devices queue themselves in SS, and SS will try to hook them up with their desired network every once in a while.
Most of diff is just renaming of global list to ss network list

### DIFF
--- a/code/controllers/subsystems/initialization/computer_networks.dm
+++ b/code/controllers/subsystems/initialization/computer_networks.dm
@@ -1,0 +1,36 @@
+SUBSYSTEM_DEF(networking)
+	name = "Computer Networks"
+	priority = SS_PRIORITY_COMPUTER_NETS
+	flags = SS_BACKGROUND | SS_NO_INIT
+	wait = 2 SECOND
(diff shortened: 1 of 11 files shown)


afterthought2 (dismissed):

afterthought2 (dismissed):

code/controllers/subsystems/initialization/computer_networks.dm
```diff
@@ -0,0 +1,37 @@
+SUBSYSTEM_DEF(networking)
+	name = "Computer Networks"
+	init_order = SS_INIT_COMPUTER_NETS
```
afterthought2: Doesn't init, so give it the no init flag and remove this.

afterthought2 (changes requested):

code/controllers/subsystems/initialization/computer_networks.dm
comma: don

afterthought2 (approved):

PR #465, opened 2020-05-05
Moves eye creation to an extension

NataKilar:
Adds an extension for easier creation of eyes for objects that require displaced views for mobs. Handles the possession and release for mobs, as well as granting the necessary actions. Currently used for shuttle landing, and the visualnet connection for the MIU mask and two spells.

Also adds a couple hooks for applying visual effects and a click handler to the owner of the eye, to be used in the blueprints rework.


code/modules/spells/general/god_vision.dm
```diff
@@ -13,17 +13,12 @@
 
 	spell_flags = Z2NOCAST
 	hud_state = "wiz_IPC"
-	var/mob/observer/eye/freelook/vision
-	var/eye_type = /mob/observer/eye/freelook/wizard_eye
+
+	var/extension_type = /datum/extension/eye/cameranet
```
PsiOmegaDelta, line 17: This variable only appears to be written to, never read

code/modules/clothing/masks/miscellaneous.dm
```diff
@@ -98,54 +98,24 @@
...
-	if(user == eye.owner)
-		to_chat(eye.owner, "<span class='notice'>You feel disorented for a moment as your mind disconnects from the camera network.</span>")
-		eye.release(eye.owner)
-		eye.forceMove(src)
+	if(CN.current_looker)
+		CN.unlook()
+		to_chat(user, SPAN_NOTICE("You disengage \the [src]."))
+	else
+		CN.look(user)
+		to_chat(user, SPAN_NOTICE("You engage \the [src]."))
```
PsiOmegaDelta: How does the AI mask ensure that it can only be used while equipped in the appropriate slot?

code/modules/mob/observer/eye/eye.dm
```diff
@@ -14,7 +14,10 @@
 	var/acceleration = 1
 	var/owner_follows_eye = 0
 	var/living_eye = TRUE 	// Whether or not the eye uses normal living vision handling.
-
+	
+	var/use_click_handler = FALSE // Set if the eye uses special click handling. Distinct from parent mob click handling for AI etc.
+	var/click_handler_type = /datum/click_handler/eye
```
PsiOmegaDelta: Why just not set `click_hander_type` to `null` and check for that instead of having an additional var, `use_click_handler`?

code/modules/mob/observer/eye/eye.dm
```diff
@@ -133,3 +142,32 @@
...
+/mob/observer/eye/remove_visual(var/mob/M)
+	if(M != owner || !M.client)
+		return FALSE
+	return TRUE
+
+/mob/observer/eye/ClickOn(var/atom/A, var/params)
+	if(owner)
+		return ..()
+	
+	return owner.ClickOn(A, params)
```
PsiOmegaDelta: `owner` is either null in which case this line will runtime, or line 158-159 will return first

code/datums/extensions/eye/_eye.dm
```diff
@@ -0,0 +1,103 @@
...
+	return get_turf(extension_eye)
+
+/datum/action/eye/
+	action_type = AB_GENERIC
+	check_flags = AB_CHECK_STUNNED|AB_CHECK_LYING
+	var/eye_type = /mob/observer/eye/
+	var/target_type = HOLDER_TARGET // The relevant owner of the proc to be called by the action.
+
+/datum/action/eye/New(var/datum/extension/eye/eye_extension)
+	switch(target_type)
```
PsiOmegaDelta, line 86: No validation for if the incorrect type is supplied?

code/modules/overmap/ships/computers/shuttle.dm
```diff
@@ -4,25 +4,9 @@
...
-	finish_landing_action = new(src)
-	end_landing_action = new(src)
-
-/obj/machinery/computer/shuttle_control/explore/Destroy()
-	end_landing()
-	current_user = null
-	QDEL_NULL(finish_landing_action)
-	QDEL_NULL(end_landing_action)
-	. = ..()
+	set_extension(src, /datum/extension/eye/landing)
```
PsiOmegaDelta, line 9: What happens to the AIs vision if it ever uses this computer?

code/datums/extensions/eye/_eye.dm
```diff
@@ -0,0 +1,103 @@
...
+			return ..(eye_extension.extension_eye)
+		if(EXTENSION_TARGET)
+			return ..(eye_extension)
+
+/datum/action/eye/CheckRemoval(mob/living/user)
+	if(!user.eyeobj || !istype(user.eyeobj, eye_type))
+		return TRUE
+
+// Every eye created using a subtype of this extension will have this action added for manual unlooking.
+/datum/action/eye/unlook/
```
PsiOmegaDelta: Trailing /

code/datums/extensions/eye/_eye.dm
```diff
@@ -0,0 +1,103 @@
...
+	if(current_looker)
+		for(var/datum/action/A in actions)
+			A.Remove(current_looker)
+
+	current_looker = null
+
+/datum/extension/eye/proc/get_eye_turf()
+	return get_turf(extension_eye)
+
+/datum/action/eye/
```
PsiOmegaDelta: Trailing /
Repeat for further trailing /

code/modules/overmap/ships/computers/shuttle.dm
NataKilar: Currently, nothing. The look() proc on the extension returns if the user already has an eyeobj. Caching the old eyeobj instead may be possible.

code/datums/extensions/eye/_eye.dm
NataKilar: Validation for the target_type? How would it deal with an incorrect type, just a crash?

code/modules/clothing/masks/miscellaneous.dm
NataKilar: I assumed this was how the mask worked previously, but I missed a couple lines. Fixed.

code/datums/extensions/eye/_eye.dm
PsiOmegaDelta: That's the usual way, aye

code/modules/clothing/masks/miscellaneous.dm
```diff
@@ -98,54 +98,27 @@
...
-	engage_mask(user)
-
-/obj/item/clothing/mask/ai/dropped(var/mob/user)
-	..()
-	disengage_mask(user)
-
-/obj/item/clothing/mask/ai/proc/engage_mask(var/mob/user)
-	if(!active)
-		return
 	if(user.get_equipped_item(slot_wear_mask) != src)
```
PsiOmegaDelta, line 109: So now the equip-check is back but what if the mob or someone else (via stripping, magic spells, etc.) remove the mask?

NataKilar: Events hooked to the unlook proc in the extension already check for movement of both the holder and eye player, which should be sufficient for any removal of the mask, unless there's an edge case I'm not considering.

MistakeNot4892 (dismissed):

MistakeNot4892 (approved):

PR #468, opened 2020-05-06
Unifying /material and /decl/reagent as /decl/material.

MistakeNot4892:
This is solely the path unification, var unification and updates to the way materials are retrieved. It ~~doesn't touch any logic or merge any extraneous variables yet~~ yes it does I merged a bunch of vars and SSchemistry/SSmaterials. Opening to get the main body of changes out of the way before doing the rest.

EDIT ok, we're doing the whole thing in this PR apparently.


MistakeNot4892:
Marking this DNM as it's going to all have to go in at once to work properly.

MistakeNot4892:
Will get back to this in a few weeks most likely.

PR #475, opened 2020-05-07
Removes runtime caused by null lists in english_lists

Techhead0:
Null lists are common due to LAZYLIST usage
Dedicated to Alex6511


Lohikar:
Squash pls.

MistakeNot4892 (approved):

PR #483, opened 2020-05-07
MultiZ Singulo

quardbreak:
* Singulo now have a chance to travel up/down.
* Adds fancy animation for the singularity generator. Credits to Tau Ceti community.


MistakeNot4892 (approved):

code/modules/power/singularity/singularity.dm
```diff
@@ -285,32 +285,44 @@
...
 	if(target && prob(60))
 		movement_dir = get_dir(src,target) //moves to a singulo beacon, if there is one
+	else
+		var/location
+		if(prob(16))
+			location = GetAbove(src)
+			if (location)
+				src.Move(location, UP)
+				return TRUE
+		else if(prob(16))
```
PsiOmegaDelta, line 306: If I still remember how to calculate probabilities, the probability that the singulo will move down is only 0.84 * 0.16=0,1344% (i.e. the chance that it moves up takes away from the chance that it will move down).
For a fair chance in either direction you should do (semi-pseudo-code):
``` 
if(target && prob(60))
    movement_dir = get_dir(src,target)
else if(prob(16))
    if(prob(50) && GetAbove(src))
        MoveUp()
    else if(GetBelow(src))
        MoveDown()
    return TRUE
```

PR #488, opened 2020-05-09
Make planetary atmos scans allow atmos skill

SierraKomodo:
:cl:
tweak: Atmospheric readings when scanning planetoids is now reliant on atmospherics or science skill instead of only science skill (You only need one of the two).
/:cl:

### DIFF
--- a/code/modules/overmap/exoplanets/_exoplanet.dm
+++ b/code/modules/overmap/exoplanets/_exoplanet.dm
@@ -222,7 +222,7 @@
 	. = ..()
 	var/list/extra_data = list("<br>")
 	if(atmosphere)
-		if(user.skill_check(SKILL_SCIENCE, SKILL_EXPERT))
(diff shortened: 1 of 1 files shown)


MistakeNot4892 (dismissed):
Seems weird to make Atmospherics a required skill for pilots but shrug.

comma:
I think it's super non-obvious that for piloting/sciencing you need an engineering atmos skill, and that skill in general should just die off TBH.
It's grandfathered form before mechanical skills and is used in literally 1 place.

MistakeNot4892:
Relating to convo on Discord about the skill - https://github.com/NebulaSS13/Nebula/pull/490

Techhead0:
If the skill isn't being removed, another option is making it science *or* atmos skill here

SierraKomodo:
Updated to be science OR atmos, as recommended by @Techhead0

MistakeNot4892 (approved):
Compromise works for me.

PR #490, opened 2020-05-09
Kills three more or less useless skills.

MistakeNot4892:
- Anatomy is now handled by Medicine. It had no independant use outside of surgery and medical contexts covered by other skills.
- Atmospherics is now handled by Construction. The sole previous use for it was the rapid pipe dispenser.
- Exosuit Operation is now handled by Piloting.


N8-Toe:
I'm not to hot on this change, I like anatomy being a separate skill as it puts a higher point cost to be a full on doctor, and helps distinct between "is a really skilled EMT/knows a lot of first aid" and is an actual doctor. 

more ambivalent on the atmos skill

code/modules/surgery/__surgery_setup.dm
```diff
@@ -1,4 +1,4 @@
-#define  SURGERY_SKILLS_GENERIC				list(SKILL_ANATOMY = SKILL_ADEPT, SKILL_MEDICAL = SKILL_EXPERT)
-#define  SURGERY_SKILLS_DELICATE			list(SKILL_ANATOMY = SKILL_EXPERT, SKILL_MEDICAL = SKILL_EXPERT)
+#define  SURGERY_SKILLS_GENERIC				list(SKILL_MEDICAL = SKILL_EXPERT)
+#define  SURGERY_SKILLS_DELICATE			list(SKILL_MEDICAL = SKILL_EXPERT)
 #define  SURGERY_SKILLS_ROBOTIC				list(SKILL_DEVICES = SKILL_ADEPT)
-#define  SURGERY_SKILLS_ROBOTIC_ON_MEAT		list(SKILL_ANATOMY = SKILL_ADEPT, SKILL_DEVICES = SKILL_ADEPT)
```
comma, line 4: Should replace it with SKILL_MEDICAL so every engineer isn't suddenly a surgeon too

MistakeNot4892:
Needs more discussion regarding the role Anatomy previously filled as a class gate.

PR #492, opened 2020-05-10
Adds some simple department signs

SodiumShine:
* Some departments didn't have any signs. So I added some.
* Also added a bathroom sign but the European kind because I don't like the other kind.
* Removed a Scavstation sign.


SodiumShine:
why tho travis

code/game/objects/structures/signs.dm
```diff
@@ -436,3 +431,68 @@
...
+	desc = "A sign labelled FORENSICS."
+	icon_state = "forensics"
+
+/obj/structure/sign/security
+	name = "\improper SECURITY"
+	desc = "A sign labelling the area as belonging to Security."
+	icon_state = "sec_scale"
+
+/obj/structure/sign/security/alt
+	name = "\improper SECURITY"
```
MistakeNot4892: These don't need to be set again as they are inherited from the previous definition.

SodiumShine: These? Do you just mean security/alt name or am I missing another?

MistakeNot4892: Name and desc on this one, I didn't see any others.

SodiumShine: Oh just realized mobile view is really bad and didn't make it clear what line you were pointing at.

MistakeNot4892 (approved):

PR #497, opened 2020-05-10
Moved req_access on machines to initial_access /w network lock auto-c…

biscuitWizard:
Moved req_access on machines to initial_access /w network lock auto-creation.

This prevents scenarios where machines created in play by manually building them have permissions that prevent their creator from accessing the machine.

This also converts all machines to using the network authorization system. The network ACL auto-create grant list has been expanded to correlate with this change.

### DIFF
--- a/code/game/machinery/_machines_base/machine_construction/_construction.dm
+++ b/code/game/machinery/_machines_base/machine_construction/_construction.dm
@@ -18,6 +18,10 @@
 
 // Return a change state define or a fail message to block transition.
 /obj/machinery/proc/cannot_transition_to(var/state_path, var/mob/user)
+	if(ispath(state_path, /decl/machine_construction/default/deconstructed))
+		var/obj/item/stock_parts/network_lock/lock = get_component_of_type(/obj/item/stock_parts/network_lock)
+		if(istype(lock) && !allowed(user)) // Only check if we have a network_lock.
+			return MCS_BLOCK
 	return MCS_CHANGE
 
 /decl/machine_construction
--- a/code/game/machinery/_machines_base/machinery_components.dm
+++ b/code/game/machinery/_machines_base/machinery_components.dm
@@ -30,9 +30,10 @@ GLOBAL_LIST_INIT(machine_path_to_circuit_type, cache_circuits_by_build_path())
 				for(var/type in req_components)
 					uncreated_component_parts[type] += (req_components[type] || 1)
 		if(initial_access && length(initial_access) > 0)
-			// Auto-create a network lock.
-			var/obj/item/stock_parts/network_lock/lock = install_component(/obj/item/stock_parts/network_lock/buildable, refresh_parts = FALSE)
-			lock.grants = initial_access
+			for(var/access_list in initial_access)
+				// Each part is an AND component.
+				var/obj/item/stock_parts/network_lock/lock = install_component(/obj/item/stock_parts/network_lock/buildable, refresh_parts = FALSE)
+				lock.grants = access_list
 
 	// Create the parts we are supposed to have. If not full_populate, this is only hard-baked parts, and more will be added later.
 	for(var/component_path in uncreated_component_parts)
@@ -308,6 +309,9 @@ Standard helpers for users interacting with machinery parts.
 	for(var/path in types_of_component(/obj/item/stock_parts))
 		var/obj/item/stock_parts/part = path
 		if(!(initial(part.part_flags) & PART_FLAG_HAND_REMOVE))
+			continue
+		var/obj/item/stock_parts/network_lock/lock = part
+		if(istype(lock) && !allowed(user))
 			continue
 		if(components_are_accessible(path))
 			removable_parts[initial(part.name)] = path
--- a/code/game/machinery/_machines_base/machinery_components.dm
+++ b/code/game/machinery/_machines_base/machinery_components.dm
@@ -29,6 +29,10 @@ GLOBAL_LIST_INIT(machine_path_to_circuit_type, cache_circuits_by_build_path())
 				LAZYINITLIST(uncreated_component_parts)
 				for(var/type in req_components)
 					uncreated_component_parts[type] += (req_components[type] || 1)
+		if(initial_access && length(initial_access) > 0)
+			// Auto-create a network lock.
+			var/obj/item/stock_parts/network_lock/lock = install_component(/obj/item/stock_parts/network_lock, refresh_parts = FALSE)
+			lock.grants = initial_access
 
 	// Create the parts we are supposed to have. If not full_populate, this is only hard-baked parts, and more will be added later.
 	for(var/component_path in uncreated_component_parts)
--- a/code/game/machinery/_machines_base/machinery_components.dm
+++ b/code/game/machinery/_machines_base/machinery_components.dm
@@ -31,7 +31,7 @@ GLOBAL_LIST_INIT(machine_path_to_circuit_type, cache_circuits_by_build_path())
 					uncreated_component_parts[type] += (req_components[type] || 1)
 		if(initial_access && length(initial_access) > 0)
 			// Auto-create a network lock.
-			var/obj/item/stock_parts/network_lock/lock = install_component(/obj/item/stock_parts/network_lock, refresh_parts = FALSE)
+			var/obj/item/stock_parts/network_lock/lock = install_component(/obj/item/stock_parts/network_lock/buildable, refresh_parts = FALSE)
 			lock.grants = initial_access
 
 	// Create the parts we are supposed to have. If not full_populate, this is only hard-baked parts, and more will be added later.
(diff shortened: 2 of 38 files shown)


mods/psionics/machines/psimonitor.dm
```diff
@@ -6,7 +6,7 @@
 	anchored = TRUE
 	density = TRUE
 	opacity = FALSE
-	req_access = list(list(access_psychiatrist, access_captain, access_cmo, access_hos))
+	initial_access = list(access_psychiatrist, access_captain, access_cmo, access_hos)
```
comma: keep the second list, you're changing the logic otherwise

biscuitWizard: No, all network lists are OR based, not and based. This is the same functionality.

network_lock doesn't follow req_access spec.

biscuitWizard: After discussing with afterthought, we decided to remove AND functionality completely from grant-building permission sets with network_lock. It seemed superfluous when you could alternatively just create more granular grants.

code/game/machinery/_machines_base/machinery_components.dm
```diff
@@ -29,6 +29,10 @@ GLOBAL_LIST_INIT(machine_path_to_circuit_type, cache_circuits_by_build_path())
 				LAZYINITLIST(uncreated_component_parts)
 				for(var/type in req_components)
 					uncreated_component_parts[type] += (req_components[type] || 1)
+		if(initial_access && length(initial_access) > 0)
+			// Auto-create a network lock.
+			var/obj/item/stock_parts/network_lock/lock = install_component(/obj/item/stock_parts/network_lock, refresh_parts = FALSE)
+			lock.grants = initial_access
```
afterthought2: I don't get it; this just won't work as far as I can tell. This is supposed to be a list of file names, but you're potentially feeding in a list of lists.

code/game/machinery/alarm.dm
```diff
@@ -56,7 +56,7 @@
 	idle_power_usage = 80
 	active_power_usage = 1000 //For heating/cooling rooms. 1000 joules equates to about 1 degree every 2 seconds for a single tile of air.
 	power_channel = ENVIRON
-	req_access = list(list(access_atmospherics, access_engine_equip))
+	initial_access = list(list(access_atmospherics, access_engine_equip))
```
afterthought2, line 59: Here's an example where it's a list of lists.

afterthought2 (commented):
I don't like how this doesn't support both AND/OR, but replaces aspects of systems that do. Having both AND and OR for _mapped_ entities is vital, imo, and anything that touches access req lists in the standard access req format should use and support that format fully (list or list of lists). I don't care about the UI aspects; make it error when modifying something with both AND and OR for all I care. I want the backend to be fully compatible with existing systems.

I also don't like how this component it adds is trivially circumvented by removing it from the machine, while in most cases the original system is not circumvented in this way. Yes, I know you're adding a non-removable component in init, but that doesn't make it better: it means that if I deconstruct and then reconstruct the machine, I can never, under no circumstances, recover the machine's prior state, and the whole point of buildable machines was to make this generally possible. I feel that the hardware/software aspects of this access stuff haven't been thought through fully enough.

biscuitWizard:
> I don't like how this doesn't support both AND/OR, but replaces aspects of systems that do. Having both AND and OR for _mapped_ entities is vital, imo, and anything that touches access req lists in the standard access req format should use and support that format fully (list or list of lists). I don't care about the UI aspects; make it error when modifying something with both AND and OR for all I care. I want the backend to be fully compatible with existing systems.
> 
> I also don't like how this component it adds is trivially circumvented by removing it from the machine, while in most cases the original system is not circumvented in this way. Yes, I know you're adding a non-removable component in init, but that doesn't make it better: it means that if I deconstruct and then reconstruct the machine, I can never, under no circumstances, recover the machine's prior state, and the whole point of buildable machines was to make this generally possible. I feel that the hardware/software aspects of this access stuff haven't been thought through fully enough.

"Vital" but never used in that capacity -- ever. It's not a needed feature, just over-engineering for the sake of over-engineering.

code/game/machinery/alarm.dm
biscuitWizard: Oops. That's a bug. It should only be one list. :) Good catch.

code/game/machinery/_machines_base/machinery_components.dm
biscuitWizard: I'm never supposed to be feeding it in a list of lists. List of lists is invalid format.

biscuitWizard:
> if I deconstruct and then reconstruct the machine, I can never, under no circumstances, recover the machine's prior state

Why not? The req_access is stored in configuration on the component itself. That is not modified by the component's installation or removal.

biscuitWizard:
to prove my point, here is a complete list of locations where AND is used for req_access.
alarm.dm, line 127
`req_access = list(access_rd, access_atmospherics, access_engine_equip)`
This looks like a bug to me.

suit_cycler.dm, line 11
`req_access = list(access_captain, access_bridge)`
Bridge is irrelevant. Captain should be sufficient. Or this is a bug.

gibber.dm, line 9
`req_access = list(access_kitchen,access_morgue)`
This looks like a bug to me. Or access_kitchen is sufficient.

All other lines are not machines, and thus not impacted by these changes.

afterthought2:
The component you create on init isn't a buildable one, and won't be left behind when the machine deconstructs.

`"Vital" but never used in that capacity -- ever. It's not a needed feature, just over-engineering for the sake of over-engineering.` You haven't maintained maps for large traditional servers (traditional here mostly referring to them having a single dominant station map, not pop or anything). Having this flexibility in mapping is important from my bay dev experience. Posting bearcat examples doesn't really prove your point.

biscuitWizard:
> The component you create on init isn't a buildable one, and won't be left behind when the machine deconstructs.

Then that's a bug. It's supposed to be buildable.

> You haven't maintained maps for large traditional servers (traditional here mostly referring to them having a single dominant station map, not pop or anything). Having this flexibility in mapping is important from my bay dev experience. Posting bearcat examples doesn't really prove your point.

None of these come from bearcat. All req_access entries provided are on the machine definitions, which is why I posted the filename. That's extraordinarily poor design when looked at through the perspective of multi-ship map design. Your point would be valid if they were actually map definitions, but they are not.

Furthermore, req_access for machines hasn't been removed and mappers, using appropriate implementation procedures could use req_access on the mapped versions of these machines per legacy support.

biscuitWizard:
If you want my user story, I was a player on the Unishi, and I built a robotics fabricator -- and was *locked out* because I didn't have the torch role.

I was *furious* I went through all the time and effort to research and build the thing and wasn't even able to use it for nonsensical reasons. To me that's a showstopping bug. Whatever needs to happen to make that not possible, that needs to happen.

This was my solution. Making mapped machines spawn with permissions, but constructed machines not. I just used the network lock to do that. It could have literally been anything, I could even turn initial_access into a req_access setter only on populate parts. It would be an equal fix.

But I figured I'd kill two birds with one stone here and just do a proper migration over to network_lock while I did it.

By core game perspective, AND permission lists are definitively not required. There are mechanisms that exist to do AND permission lists, but this is not that mechanism. If network_locks will eventually support AND functionality that will be provided by a future PR. But by all regards core game specification of machines -- and their role -- should be more of a suggestion and first / foremost not a gameplay blocker.

Regarding your problem with network locks being able to be removed -- I could add access_locks to spawn with network_locks.

afterthought2:
Maybe to clarify, my issues with this PR and the other access stuff you've been doing can be resolved with maybe 20-30 lines of code, which is why I keep being confused at this turning into an argument. Here is what I'd be looking for:

- Allow setting `req_access` or this new `initial_access` to anything in the standard req access format (list or list of lists). If it is then set to something other than what your component can parse into grants, you may handle it however you like. Examples may be to take the first entry only, to make several components (several components, each operating on OR, replicate the nested AND / OR structure), or crashing. [Note that I'm not asking to have your internal grants system to ever support AND; I don't care about that. I'm asking for the places that it interfaces with external code to adhere fully to the external code's conventions.]

- Make buildable network access locks check the user for access to determine if they can be removed. [There are other solutions, but this is by far the simplest.]

MistakeNot4892:
Seconding afterthought that the system needs to look the same to external code after this PR.

Confused about the Unishi example, the fact that something on a submap is using main map access is irrelevant to asking for AND access to be supported by mapping.

code/game/machinery/_machines_base/machine_construction/_construction.dm
```diff
@@ -18,6 +18,9 @@
 
 // Return a change state define or a fail message to block transition.
 /obj/machinery/proc/cannot_transition_to(var/state_path, var/mob/user)
+	var/obj/item/stock_parts/network_lock/lock = get_component_of_type(/obj/item/stock_parts/network_lock)
+	if(istype(lock) && !allowed(user)) // Only check if we have a network_lock.
+		return MCS_BLOCK
```
afterthought2: Should check that state_path is `deconstructed`; otherwise it will block every state transition.

biscuitWizard:
After discussing we decided the solution was to treat network_locks as AND when there are multiple. To block deconstruction if there's an active network lock. And to block removing network_locks if there's an active network lock without sufficient user access.

These features have been added.

afterthought2 (approved):
The typecheck in `part_removal` is kind of horrible, but whatever.

PR #507, opened 2020-05-13
Overmap Flags System

Mazianni:
Originally from https://github.com/Baystation12/Baystation12/pull/25716 

I'm not aware of any issues with the system or the implementation as it was taken straight from that PR, but if there's any issues, please do point them out.

This transitions sectors to using bitflags instead of variables, and is something I figured I should send upstream since it is a perquisite for a coming modpack.

### DIFF
--- a/maps/away/blueriver/blueriver.dm
+++ b/maps/away/blueriver/blueriver.dm
@@ -3,7 +3,7 @@
 /obj/effect/overmap/visitable/sector/arcticplanet
 	name = "arctic planetoid"
 	desc = "Sensor array detects an arctic planet with a small vessle on the planet's surface. Scans further indicate strange energy levels below the planet's surface."
-	sector_flags = null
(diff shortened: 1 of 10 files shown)


Mazianni:
Missed away sites. Which aren't run locally. That's fun.

maps/away/blueriver/blueriver.dm
```diff
@@ -3,7 +3,7 @@
 /obj/effect/overmap/visitable/sector/arcticplanet
 	name = "arctic planetoid"
 	desc = "Sensor array detects an arctic planet with a small vessle on the planet's surface. Scans further indicate strange energy levels below the planet's surface."
-	in_space = 0
+	sector_flags = null
```
MistakeNot4892: `0` would be more appropriate than `null`.

MistakeNot4892 (approved):

PR #513, opened 2020-05-15
[halp] Added toggle-able default and lighting plane masters.

MistakeNot4892:
Need help with this.
- ~~At some point lighting became completely fucked despite not really changing it very much and I cannot for the life of me work out why or how.~~
- ~~Toggling darkness sets the alpha of both the lighting plane master and the darkness layer to 0, but this does not actually disables shadows, just makes them darker.~~
- ~~Space is perma-white.~~ 
- Adding plane masters for the skybox, dust and space planes causes space to be perma-white.
- The plane master objects block all clicks and prevent you from interacting with anything if they are in the client screen list.
Pls.

### DIFF
--- a/code/__defines/_planes+layers.dm
+++ b/code/__defines/_planes+layers.dm
@@ -59,7 +59,6 @@ What is the naming convention for planes or layers?
 
 	FLOAT_PLANE = -32767
 */
-
 #define CLICKCATCHER_PLANE -100
 
 #define SPACE_PLANE               -99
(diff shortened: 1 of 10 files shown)


code/__defines/_planes+layers.dm
```diff
@@ -72,10 +71,8 @@ What is the naming convention for planes or layers?
 	#define DUST_LAYER                   2
 
 // Openspace uses planes -80 through -70.
-
-#define OVER_OPENSPACE_PLANE        -3
-
-#define DEFAULT_PLANE                   0
+#define DEFAULT_PLANE                0
+	#define DEFAULT_LAYER               0
```
Lohikar, line 75: What's this for?

MistakeNot4892: Completeness.

MistakeNot4892:
Will revisit if I ever work out the bug.

PR #514, opened 2020-05-16
Add proximity unit tests

PsiOmegaDelta:
Suppose I might as well push them as I complete them, in case I'm run over by a bus


code/datums/proximity_trigger/turf_selection.dm
```diff
@@ -6,6 +6,7 @@
 	var/center = get_turf(origin)
 	if(!center)
 		return
+	. += center
```
PsiOmegaDelta, line 9: This one doesn't actually make sense, conducting additional testing

PsiOmegaDelta: After some testing it actually does make some sense

comma (approved):

PR #516, opened 2020-05-17
Refactoring holiday handling.

MistakeNot4892:
Holidays are now driven by a JSON file which creates and configures a datum (with a specific type for Christmas and any future event that needs to have ingame effects).

I removed some dead code and the Mayan end of the world as it will not fire again for ninety-two years and if SS13 is still alive at that point we will have more serious problems than missing a meme holiday.

Also added the next ten years of Easter rather than calling out to a service or something. Note to whoever is here in ten years: add more Easter dates.


MistakeNot4892:
Occurs to me you could theoretically use this to automate stuff like Goon's Ass Day.

comma (approved):

PR #519, opened 2020-05-18
Reworks ship sensors

NataKilar:
Ports and adapts https://github.com/BoHBranch/BoH-Bay/pull/566

Sensors no longer uses lighting for obfuscation, but instead casts images of overmap objects to the client screen upon identification. Currently, only ships use the full implementation of this, requiring scanning ships to remain within range until full identification is complete. How long this takes depends on the strength of the sensors machine, the speed of the ship being identified, and the distance between the two ships. Additionally, ships can be tracked with a new conspicuous looking ship tracker item, or enable a distress beacon from their sensors console. Overmap events such as ion clouds require sensors to be in range to be visible, but otherwise act as normal.


Mazianni:
Should really change the PR desc to be 'removes the flashlight taped to the ship'.

If only for humorous effect.

code/modules/overmap/contacts/_contacts.dm
```diff
@@ -0,0 +1,103 @@
...
+	var/obj/effect/overmap/effect
+
+/datum/overmap_contact/New(var/obj/machinery/computer/ship/sensors/creator, var/obj/effect/overmap/source)
+	// Update local tracking information.
+	owner =  creator
+	effect = source
+	name =   effect.name
+	owner.contact_datums[effect] = src
+
+	marker = image(loc = effect, icon = effect.icon, icon_state = effect.icon_state)
```
comma: can change `var/image/marker` to `var/mutable_apperance/marker`
and then just do
```
marker = new(effect)
marker.alpha = 0
``` 
here, won't need `name =   effect.name` and the rest of stuff

code/modules/overmap/contacts/_contacts.dm
```diff
@@ -0,0 +1,103 @@
...
+		radar.pixel_x = -2
+		radar.tag = "radar"
+		radar.filters = filter(type="blur", size = 1)
+		images += radar
+
+		var/matrix/M = matrix()
+		M.Scale(range*2.6)
+		animate(radar, transform = M, alpha = 0, time = (SENSOR_TIME_DELAY*range), 1, SINE_EASING)
+		addtimer(CALLBACK(src, .proc/reset_radar, radar), (SENSOR_TIME_DELAY *range+0.1))
+		QDEL_IN(radar, (0.25 SECONDS *range+0.3))
```
comma: I'm unsure if qdeling images is even a thing, and even if it is, since tag is set, it'll never GC so that's a hard del every time I tihnk?
IMO should just reuse same image stored in a var instead of regenerating it every time.
just reset the alpha and scale and should be as good as new
since it'll be 0 alpha at the end of animation, maybe shouldn't even bother removing it form images

code/modules/overmap/contacts/contact_sensors.dm
```diff
@@ -0,0 +1,154 @@
...
+			var/datum/overmap_contact/record = contact_datums[key]
+			if(record.effect == linked)
+				continue
+			qdel(record) // Immediately cut records if power is lost.
+
+		objects_in_view.Cut()
+		return
+
+	var/sensor_range = round(sensors.range,1)
+	linked.set_light(1, sensor_range, sensor_range+1)
```
comma: are theses set light things needed anymore? Since area isn't even dynamically lit

NataKilar: I wasn't entirely sure why it was left in in the original PR. I don't see any reason it's necessary though so I'll cut it.

code/modules/overmap/contacts/_contacts.dm
NataKilar: Per Lohikar, it doesn't sound like using a mutable appearance for this is a good idea. I went ahead and copied the appearance of the effect instead and it seems to work just as well while cutting down on a few extraneous lines.

code/modules/overmap/contacts/_contacts.dm
```diff
@@ -0,0 +1,94 @@
...
+	if(check_effect_shield())
+		var/image/shield_image = image(icon = 'icons/obj/overmap.dmi', icon_state = "shield")
+		shield_image.pixel_x = 8
+		marker.overlays += shield_image
+
+	radar.transform = null
+	radar.alpha = 255
+
+	if(range > 1)
+		images |= radar
```
comma, line 47: should also probably reset alpah and transform of the image here.

code/modules/overmap/ships/computers/sensors.dm
```diff
@@ -28,6 +44,7 @@
 	var/data[0]
 
 	data["viewing"] = viewing_overmap(user)
+	data["distress"] = linked.instant_contact
```
comma: is this still needed?

code/modules/overmap/contacts/contact_sensors.dm
```diff
@@ -0,0 +1,150 @@
...
+		return
+	
+	// Update our own marker icon regardless of power or sensor connections.
+	var/sensor_range = round(sensors.range,1)
+	var/datum/overmap_contact/self_record = contact_datums[linked]
+	self_record.update_marker_icon(sensor_range)
+	self_record.show()
+
+	// Update our 'sensor range' (ie. overmap lighting)
+	if(!sensors || !sensors.use_power || !sensors.powered())
```
afterthought2: Generally best to check `sensors.stat & NOPOWER` instead of call `powered()`.

code/modules/overmap/contacts/contact_sensors.dm
```diff
@@ -0,0 +1,150 @@
...
+
+	// Find all sectors with a tracker on their z-level. Only works on ships when they are in space.
+	for(var/obj/item/ship_tracker/tracker in trackers)
+		if(tracker.enabled)
+			var/obj/effect/overmap/visitable/tracked_effect = map_sectors["[get_z(tracker)]"]
+			if(tracked_effect && istype(tracked_effect) && tracked_effect != linked && tracked_effect.requires_contact)
+				objects_in_current_view[tracked_effect] = TRUE
+				objects_in_view[tracked_effect] = 100
+
+	for(var/obj/effect/overmap/contact in (view(sensor_range, linked)-linked))
```
afterthought2: There are some kind of dm-level optimizations for iterating view directly without doing this list op which in some cases are very dramatic (unsure about this one). I'd move it to the logic below instead of here.

comma (dismissed):

afterthought2 (commented):
So expensive. I guess it's for the greater good, though. Worth considering having it not scan by default, and needing the console to be on and having the user hit a button to start actually scanning imo (unless it already does that with the sensors.use_power?).

comma (approved):

PR #520, opened 2020-05-18
Adjust spawn count

PsiOmegaDelta:
Did someone manage to sneak in an extra spawn, or how did this occur?


MistakeNot4892:
Can we just remove the spawn count check? We will catch spawns in review anyway.

PsiOmegaDelta:
Humans are fallible, Travis is ever vigilant

PR #521, opened 2020-05-18
Extra typechecking in materials code.

biscuitWizard:
Specifically things like the materials controller error which would *never* be hit, since it'd throw a runtime before getting to the error check. :)

### DIFF
--- a/code/controllers/subsystems/initialization/materials.dm
+++ b/code/controllers/subsystems/initialization/materials.dm
@@ -58,8 +58,9 @@ SUBSYSTEM_DEF(materials)
 				alloy_components[component] = TRUE
 
 /datum/controller/subsystem/materials/proc/get_material_datum(var/mat)
-	. = materials_by_name[mat]
-	if(!.)
+	try
+		. = materials_by_name[mat]
+	catch
 		if(ispath(mat))
 			crash_with("Unable to acquire material by path '[mat]'.")
 		else
--- a/code/game/objects/structures/window.dm
+++ b/code/game/objects/structures/window.dm
@@ -345,8 +345,9 @@
 /obj/structure/window/on_update_icon()
 	//A little cludge here, since I don't know how it will work with slim windows. Most likely VERY wrong.
 	//this way it will only update full-tile ones
-	color =  material.icon_colour
-	if(reinf_material)
+	if(istype(material))
+		color =  material.icon_colour
+	if(istype(reinf_material))
 		basestate = reinf_basestate
 	else
 		basestate = initial(basestate)
--- a/code/game/turfs/simulated/walls.dm
+++ b/code/game/turfs/simulated/walls.dm
@@ -37,7 +37,7 @@
 	if(!materialtype)
 		materialtype = DEFAULT_WALL_MATERIAL
 	material = SSmaterials.get_material_datum(materialtype)
-	if(!isnull(rmaterialtype))
+	if(ispath(rmaterialtype, /material))
 		reinf_material = SSmaterials.get_material_datum(rmaterialtype)
 	if(ispath(girder_material, /material))
 		girder_material = SSmaterials.get_material_datum(girder_material)
--- a/code/modules/client/preference_setup/global/05_settings.dm
+++ b/code/modules/client/preference_setup/global/05_settings.dm
@@ -101,7 +101,7 @@
 /client/proc/get_preference_value(var/preference)
 	if(prefs)
 		var/datum/client_preference/cp = get_client_preference(preference)
-		if(cp)
+		if(cp && (cp.key in prefs.preference_values))
 			return prefs.preference_values[cp.key]
 		else
 			return null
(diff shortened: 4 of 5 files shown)


code/controllers/subsystems/initialization/materials.dm
```diff
@@ -58,8 +58,9 @@ SUBSYSTEM_DEF(materials)
 				alloy_components[component] = TRUE
 
 /datum/controller/subsystem/materials/proc/get_material_datum(var/mat)
-	. = materials_by_name[mat]
-	if(!.)
+	try
```
MistakeNot4892, line 61: This is unnecessary. `materials_by_name` is always a list, `.` will either be a material or null.

code/game/objects/structures/window.dm
```diff
@@ -345,8 +345,9 @@
 /obj/structure/window/on_update_icon()
 	//A little cludge here, since I don't know how it will work with slim windows. Most likely VERY wrong.
 	//this way it will only update full-tile ones
-	color =  material.icon_colour
-	if(reinf_material)
+	if(istype(material))
```
MistakeNot4892, line 348: As discussed earlier I am completely against putting `istype()` checks into code that should not be executing prior to init, just to avoid your code throwing runtimes downstream when you do weird shit.

code/game/turfs/simulated/walls.dm
```diff
@@ -37,7 +37,7 @@
 	if(!materialtype)
 		materialtype = DEFAULT_WALL_MATERIAL
 	material = SSmaterials.get_material_datum(materialtype)
-	if(!isnull(rmaterialtype))
+	if(ispath(rmaterialtype, /material))
```
MistakeNot4892, line 40: This will allow non-material values like text or numbers to be left in the material var.

code/controllers/subsystems/initialization/materials.dm
biscuitWizard: accessing materials_by_name by an invalid mat will throw a runtime. Bad index runtimes exist, it needs to be in a try catch.

code/game/objects/structures/window.dm
biscuitWizard: The code under if() wouldn't work if it wasn't of istype(). Regardless of its relation to my downstream, the if check as coded is incorrect when applied without context.

code/game/turfs/simulated/walls.dm
biscuitWizard: you're saying ispath(2, /material) will return true?

biscuitWizard: Or are you saying materialtype should also have an ispath check? Sorry, the github code snippet preview is still confusing to me.

code/controllers/subsystems/initialization/materials.dm
afterthought2: Bad index runtime only occurs if it's a number, in which case that runtime is already a sufficient way to handle the situation.

biscuitWizard: When I encountered the runtime, it was insufficient information to debug. The more verbose way is the more than sufficient way to handle it imo

MistakeNot4892: How is it insufficient? The example you showed was a snippet from DD after it had been called so often that DD started collapsing the runtimes. The information is still in the logs.

code/game/objects/structures/window.dm
MistakeNot4892: It isn't incorrect to assume that a variable is holding the correct value of its type when it's validated elsewhere. That is the point of having typed variables in the first place.

code/game/turfs/simulated/walls.dm
MistakeNot4892: I am saying that this check means an invalid `reinf_material` value will not runtime, it will just never try to find a datum for it and code further down the stream will assume it is valid because it isn't null. This is related to the point I made in the comments earlier that materials are validated lower in the code.

code/game/objects/structures/window.dm
biscuitWizard: Except typed variables are type abused by the original authors of this code.

code/controllers/subsystems/initialization/materials.dm
biscuitWizard: Your assumption is incorrect. That error only appeared because of my code. Before I had made that change, I just received invalid index errors and no valuable output.

code/game/turfs/simulated/walls.dm
biscuitWizard: That code should be repaired, too. In my case code down the stream is never reached because the object is destroyed or properly configured by other code later. I only cleared out errors relevant to me. But given these errors never impacted anyone anyways, it's kind of moot to argue about it.

PR #522, opened 2020-05-19
Fluid updates.

MistakeNot4892:
- Reagent dispensers are now universally unwrenchable and leak their reagents as fluids.
- Fluids are now flammable, replacing `liquid_fuel` decals.
- Fluids will now glow if radioactive, replacing greengoo decals.
- Fluids will now puddle when below a certain depth, ending their spread and eventually evaporating.


MistakeNot4892:
Ready for review.

MistakeNot4892:
Ready for review.

code/game/objects/effects/fluids.dm
```diff
@@ -47,10 +54,11 @@
 	ADD_ACTIVE_FLUID(src)
 
 /obj/effect/fluid/Destroy()
-	STOP_PROCESSING(SSobj, src)
-	var/turf/simulated/T = loc
+	var/turf/simulated/T = get_turf(src)
 	if(istype(T))
+		T.zone?.fuel_objs -= src
```
afterthought2: If `zone` is null, this will still runtime, so the `?` doesn't do what you want.

afterthought2 (approved):

PR #532, opened 2020-05-20
Changes spawn() count check from exactly to less than.

MistakeNot4892:
The only time this comes up is when people are removing spawn and it decides to conflict or start failing random PRs post-merge. We can check for spawn use in review, we don't need to make builds fail because they removed one.

### DIFF
--- a/test/check-paths.sh
+++ b/test/check-paths.sh
@@ -37,7 +37,6 @@ exactly 0 "incorrect indentations" '^( {4,})' -P
 exactly 20 "text2path uses" 'text2path'
 exactly 3 "update_icon() override" '/update_icon\((.*)\)'  -P
 exactly 1 "goto uses" 'goto '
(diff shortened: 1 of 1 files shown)


Lohikar (dismissed):

deathride58 (dismissed):
The check causes more trouble than it's worth. +support

comma:
Make it less than check rather than exact, would be looser but still there and someone bored can update count every so often

MistakeNot4892:
Being discussed on Disco

comma (dismissed):

Lohikar (changes requested):
Per maint-poll, pls just delete entirely.

Lohikar (approved):

PsiOmegaDelta:
Does mean that one PR could remove spawns and another could sneak new ones in if missed in a review

comma (approved):
I hate democracy

PR #534, opened 2020-05-20
Generalized lighting sabotage.

MistakeNot4892:
You can now sabotage lights in injecting any flammable reagent.


afterthought2 (approved):

afterthought2:
Kind of stupid that you can inject it with anything and it does nothing but eats the reagents, not even a message.

PR #537, opened 2020-05-21
Converting mining walls into natural walls.

MistakeNot4892:
- Converts the old mining turf into a material-driven simulated wall subtype.
- Adds solid and gas names to appropriate areas that use materials.
- Adds a strata system for generating rock layers per z-level.
- Updates paths appropriately.

**TODO**
- [X] Basic conversion.
- [x] Get final icons from Pawn.
- [x] Finish converting over xenoarch interactions.
- [x] Convert over weighted ore spawning turfs.
- [x] Add basalt material for volcanic turfs.
- [ ] Test/debug.


MistakeNot4892:
Still need to debug/polish but otherwise ready for review.

code/game/turfs/simulated/wall_natural.dm
```diff
@@ -0,0 +1,161 @@
+var/list/strata_by_z = list()
+var/list/strata_material_by_z = list()
```
afterthought2: Confusing name: it's not by z, it's by some key system you have to know. Though after reading it, if it's the case that every z level has a unique strata, this could be by z instead, probably?

code/game/turfs/simulated/wall_natural.dm
```diff
@@ -0,0 +1,161 @@
...
+	global.natural_walls -= src
+	. = ..()
+
+/turf/simulated/wall/natural/proc/spread_deposit()
+	if(!istype(reinf_material) || reinf_material.ore_spread_chance <= 0)
+		return
+	for(var/trydir in GLOB.cardinal)
+		if(!prob(reinf_material.ore_spread_chance))
+			continue
+			var/turf/simulated/wall/natural/target_turf = get_step(src, trydir)
```
afterthought2: This will never fire. Indent error?

code/game/turfs/simulated/wall_natural.dm
```diff
@@ -0,0 +1,161 @@
...
+	var/image/ore_overlay
+
+/turf/simulated/wall/natural/get_paint_examine_message()
+	. = SPAN_NOTICE("It has been <font color = '[paint_color]'>noticeably discoloured</font> by the elements.")
+
+/turf/simulated/wall/natural/Initialize()
+	if(!material)
+		if(!strata)
+			if(!global.strata_by_z["z"])
+				global.strata_by_z["z"] = pick(/decl/strata/sedimentary, /decl/strata/metamorphic, /decl/strata/igneous)
```
afterthought2: Feel like this should be a var on strata (in the running to be a stata_by_z of a z level) and then  you filter them here.

code/game/turfs/simulated/wall_natural.dm
MistakeNot4892: What is confusing about it? The key is the strata and the z-level. A z-level can have multiple stratas via turf mapping, the `strata_by_z` list is the default strata.

code/game/turfs/simulated/wall_natural.dm
MistakeNot4892: Would prefer to do that in a future PR, this is as complex as it needs to be for this PR and I am getting sick of conflicts with my other PRs.

code/game/turfs/simulated/wall_natural.dm
MistakeNot4892: Renamed them to be clearer.

code/game/turfs/simulated/wall_natural.dm
MistakeNot4892: Implemented barebones version.

afterthought2 (approved):

PR #541, opened 2020-05-23
Some OS/net refactors/fixes

comma:
Gives it proc for acquiring disk that is override by removable disk one.

Fixes ntos net status proc

Also fixes network transfer speed configuration.
It used to be 0-3 value, now it's signal power in around 0-100 range I think.
Speed now sorta mirrors old setup, by getting divided by base broadcasting strength.
Removed bunch of unneeded procs and vars and defines
Repaths disk drive filestream handler to child of disk one
Fixes network status tray icon not appearing also.

### DIFF
--- a/code/modules/modular_computers/ntos/files.dm
+++ b/code/modules/modular_computers/ntos/files.dm
@@ -182,7 +182,7 @@
 /datum/file_storage/network/get_transfer_speed()
 	if(check_errors())
 		return 0
-	return os.get_network_speed()
+	return os.get_network_status()
 
 /*
  * Special subclass for network machines specifically.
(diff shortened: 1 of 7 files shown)


code/modules/modular_computers/ntos/files.dm
```diff
@@ -223,91 +223,67 @@
...
 	. = ..()
 	if(.)
 		return
-	var/obj/item/stock_parts/computer/hard_drive/HDD = os.get_component(disk_type)
-	if(!HDD)
+	var/obj/item/stock_parts/computer/hard_drive/HDD = get_disk()
+	if(!istype(HDD))
 		return "HARDWARE ERROR: No compatible device found"
 	if(!HDD.check_functionality())
 		return "NETWORK ERROR: [HDD] is non-operational"
```
MistakeNot4892, line 237: Network error?

MistakeNot4892 (approved):

PR #542, opened 2020-05-23
Added generic lockable interface.

biscuitWizard:
This branch has nothing to do with money cards. Instead I completely redid the locking interface for secure briefcases, which is now in a datum extension.

Once this PR is in, I can work on adding the locking interface to charge cards.

### DIFF
--- a/code/datums/extensions/lockable.dm
+++ b/code/datums/extensions/lockable.dm
@@ -1,0 +1,169 @@
+/datum/extension/lockable
+	base_type = /datum/extension/lockable
+	expected_type = /obj/item
+	flags = EXTENSION_FLAG_IMMEDIATE
+	// Configuration variables
+	var/max_code_length = 5		// Max length of the password.
+
+	// Stateful variables
+	var/locked 			= 1		// Is it locked?
+	var/code 					// The passcode currently inputed.
+	var/l_code 			= null	// The actual code of the lock
+	var/l_set 			= 0		// Whether or not the lock has been set.
+	var/l_setshort 		= 0		// Whether or not the ability to set the lock is broken.
+	var/l_hacking 		= 0		// Whether or not the lock is being hacked.
+	var/emagged 		= 0		// What or not the lock is emagged.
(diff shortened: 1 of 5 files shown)


code/datums/extensions/lockable.dm
```diff
@@ -0,0 +1,195 @@
...
+					addtimer(CALLBACK(src, /datum/extension/lockable/proc/reset_memory), 3 SECONDS)
+					return TRUE
+				else
+					user.show_message(SPAN_WARNING("Unable to reset internal memory."), 1)
+					l_hacking = 0
+					bad_access_attempt(user)
+					return TRUE
+			else	
+				l_hacking = 0
+		//At this point you have exhausted all the special things to do when locked
```
MistakeNot4892: Comment is a bit weird, it will return FALSE even if unlocked.

code/datums/extensions/lockable.dm
```diff
@@ -0,0 +1,195 @@
...
+	l_hacking = 0
+	l_code = null
+	l_set = FALSE
+
+/datum/extension/lockable/proc/emag_act(var/remaining_charges, var/mob/user, var/feedback)
+	var/atom/movable/A = holder
+	if(!emagged)
+		emagged = 1
+		sleep(6)
+		A.overlays.Cut()
```
MistakeNot4892: You shouldn't be manipulating overkays outside of on_update_icon generally.

code/datums/extensions/lockable.dm
```diff
@@ -0,0 +1,195 @@
...
+	base_type = /datum/extension/lockable
+	expected_type = /obj/item/storage
+	var/icon_locking = "secureb"
+	var/icon_sparking = "securespark"
+	var/icon_opened = "secure0"
+
+/datum/extension/lockable/storage/open()
+	open = TRUE
+	var/atom/movable/A = holder
+	A.overlays.Cut()
```
MistakeNot4892: As with above, all this overlay manip should be on the storage atom.

PR #543, opened 2020-05-24
Update Charge Cards

biscuitWizard:
Charge cards now have grades and locks, as well as some defenses.

All cards are capable of emitting a shrill alarm when being tampered with unsuccessfully. They will also shock the tamperer to allow time for authorities to arrive.

The loudness of the alarm and strength of the shock depends on the grade of the card. The maximum amount a card can hold will also be dependent on grade.

PR #544, opened 2020-05-25
Remove ID card payment

biscuitWizard:
Converted vending machine and laptop vendor to only accept charge_cards in prep for https://github.com/NebulaSS13/Nebula/pull/543

### DIFF
--- a/code/game/machinery/vending/_vending.dm
+++ b/code/game/machinery/vending/_vending.dm
@@ -139,17 +139,13 @@
 
 /obj/machinery/vending/attackby(obj/item/W, mob/user)
 
-	var/obj/item/card/id/I = W.GetIdCard()
 	var/obj/item/charge_stick/CS = W.GetChargeStick()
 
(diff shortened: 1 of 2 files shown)


biscuitWizard:
Blocked by #543

comma:
What for? Sounds like adding pain in the ass for sake of it.
Now instead of just being able to buy some smokes after joining you have to go hunt down an ATM, get a credstick, and then finally be able to.

MistakeNot4892:
I would be in favour of this due to the decoupling of local access and bank account, so long as the on-join proc dumped half or two thirds of your money onto a credstick rather than into your account, for the usability concern @comma mentioned above. Could possibly be a toggle in chargen (`Money on hand: none [some] all`).

comma:
On-join spawn of credstick in your PDA would solve my ree yea.

biscuitWizard:
loadout prefs for credstick has been added.

comma (approved):

PR #547, opened 2020-05-27
Added reagent ices.

biscuitWizard:
Adding some fantasy ices that are combinations of reagents. Should auto-add to SSmaterials, and added a deep-oremap variant for ice as well.

Gives a source for some of these reagents that might not be obtainable without trading/supply subsystem otherwise.

### DIFF
--- a/code/modules/materials/definitions/materials_ice.dm
+++ b/code/modules/materials/definitions/materials_ice.dm
@@ -1,0 +1,69 @@
+/decl/material/ice
+	stack_type = null
+	icon_colour = "#a5f2f3"
+	heating_products = list(
+		/decl/material/gas/water 				= 1
+	)
+	ore_name = "ice"
+	ore_spread_chance = 25
+	ore_scan_icon = "mineral_common"
+	ore_icon_overlay = "lump"
+	removed_by_welder = TRUE
+	value = 0.2
+	sparse_material_weight = 2
+	rich_material_weight = 37
+	heating_point = T0C
+
+/decl/material/ice/aspium
+	heating_products = list(
+		/decl/material/chem/fuel/hydrazine 		= 0.3,
+		/decl/material/gas/water				= 0.7
+	)
+	ore_name = "aspium ice"
+	value = 0.4
+	sparse_material_weight = 2
+	rich_material_weight = 37
+
+/decl/material/ice/lukrite
+	heating_products = list(
+		/decl/material/chem/sulfur		 		= 0.4,
+		/decl/material/gas/water				= 0.2,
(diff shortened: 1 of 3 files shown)


code/modules/random_map/noise/ore.dm
```diff
@@ -124,4 +124,44 @@
...
+		/decl/material/ediroite_ice =	list(RESOURCE_MID_MIN, RESOURCE_MID_MAX),
+		MAT_GOLD =              list(RESOURCE_MID_MIN,  RESOURCE_MID_MAX),
+		MAT_SILVER =            list(RESOURCE_MID_MIN,  RESOURCE_MID_MAX),
+		MAT_URANIUM =           list(RESOURCE_MID_MIN,  RESOURCE_MID_MAX),
+		MAT_PHORON =            list(RESOURCE_MID_MIN,  RESOURCE_MID_MAX),
+		MAT_OSMIUM =            list(RESOURCE_MID_MIN,  RESOURCE_MID_MAX),
+		MAT_RUTILE =            list(RESOURCE_MID_MIN,  RESOURCE_MID_MAX)
+	)
+	deep_metals = list(
+		/decl/material/aspium_ice =		list(RESOURCE_MID_MIN, RESOURCE_MID_MAX),
```
MistakeNot4892: For consistency's sake please use a define for these paths. I am probably going to go through and remove them at some point but until then it should really not be mixed,

code/modules/materials/definitions/materials_ice.dm
```diff
@@ -0,0 +1,88 @@
...
+	ore_scan_icon = "mineral_common"
+	ore_name = "trigarite ice"
+	ore_icon_overlay = "lump"
+	removed_by_welder = TRUE
+	value = 0.5
+	sparse_material_weight = 38
+	rich_material_weight = 23
+
+/decl/material/ediroite_ice
+	display_name = "ediroite ice"
```
MistakeNot4892: If you made a `/decl/material/ice` and moved all the shared vars onto it you could make this file a lot smaller.

code/modules/materials/definitions/materials_ice.dm
```diff
@@ -0,0 +1,68 @@
+/decl/material/ice
+	stack_type = null
+	icon_colour = "#a5f2f3"
+	heating_products = list(
+		/decl/material/gas/water 				= 1
+	)
```
MistakeNot4892, line 6: Needs a heating_point (can't recall the var) set to 0C.

MistakeNot4892 (approved):

deathride58 (approved):

PR #552, opened 2020-05-27
Refactored AI mind for species.

biscuitWizard:
Added AI for nymph, monkey, and human.
Removed handle_npc() from species.
Add AI path to both mob define and species define. Species define takes precedence.
Added basic fugue state for mobs.


biscuitWizard:
Note for future PR, can way reduce the prob() checks for all AI and reduce their tick count to 5. Will result in performance boost.

code/modules/mob/mob.dm
```diff
@@ -52,6 +52,8 @@
 		move_intent = move_intents[1]
 	if(ispath(move_intent))
 		move_intent = decls_repository.get_decl(move_intent)
+	if(ispath(ai))
+		ai = new(src)
```
PsiOmegaDelta, line 57: This instance never appears to be deleted

biscuitWizard: For some reason I thought only /atoms had to be qdel'd. Datums too?

code/datums/ai/ai.dm
```diff
@@ -1,5 +1,17 @@
 /datum/ai
 	var/name
+	var/mob/living/body		// The parent mob we control.
+	var/waiting	= 	0		// The number of process ticks we've been waiting.
+	var/wait_for = 	1		// How many process ticks to wait before processing.
+	
+/datum/ai/New(var/mob/living/_body)
+	body = _body
```
PsiOmegaDelta: This reference is never cleared, preventing GC

code/datums/ai/ai.dm
```diff
@@ -1,5 +1,17 @@
 /datum/ai
 	var/name
+	var/mob/living/body		// The parent mob we control.
+	var/waiting	= 	0		// The number of process ticks we've been waiting.
+	var/wait_for = 	1		// How many process ticks to wait before processing.
+	
+/datum/ai/New(var/mob/living/_body)
```
PsiOmegaDelta: Could be
```
/datum/ai/New(var/mob/living/body)
    src.body = body
```

code/datums/ai/ai.dm
```diff
@@ -1,5 +1,17 @@
...
+	body = _body
 
-/datum/ai/proc/process()
-	return PROCESS_KILL
```
PsiOmegaDelta, line 5: Why not keep this proc as
```
/datum/ai/proc/process()
    SHOULD_NOT_OVERRIDE(TRUE)
    if(can_process())
        do_process()
```
Would prevent leaking implementation details to users

code/datums/ai/ai.dm
```diff
@@ -1,5 +1,17 @@
...
+/datum/ai/proc/can_process()
```
PsiOmegaDelta, line 10: A can-checking proc is typically pure

code/datums/ai/ai.dm
biscuitWizard: I don't understand this comment.

code/datums/ai/ai.dm
biscuitWizard: I don't understand this comment.

code/modules/mob/living/carbon/life.dm
```diff
@@ -31,5 +31,8 @@
 
 		. = 1
 
-		if(!client && !mind && species)
-			species.handle_npc(src)
+		// If we have an AI, and we're in a fugue state or without client.
+		// Give control to AI.
+		if(istype(ai) && ((stat & FUGUE) || (!client && !mind && species)))
+			if(ai.can_process())
+				ai.do_process()
```
PsiOmegaDelta, line 38: This appears to be copy-paste of https://github.com/NebulaSS13/Nebula/pull/552/files#diff-2de2df982739a4fbdb54a4b2715d6f92R84-R88

code/modules/mob/mob_defines.dm
```diff
@@ -186,3 +186,5 @@
 	var/list/additional_vision_handlers = list() //Basically a list of atoms from which additional vision data is retrieved
 
 	var/list/progressbars = null //for stacking do_after bars
+
+	var/datum/ai/ai						// Type abused. Define with path and will automagically create. Determines behaviour for clientless mobs.
```
PsiOmegaDelta, line 190: This is a mob-level var but is only called (and separately at that) by carbon and human `/Life()`?
The var should either be moved up, or the proc calls moved down.

code/modules/mob/living/carbon/life.dm
biscuitWizard: It is. Should be fine.

code/modules/mob/mob_defines.dm
biscuitWizard: No, this is fine. It will be overriden in simple_mob later in a future PR.

code/modules/mob/living/carbon/human/life.dm
```diff
@@ -81,9 +81,11 @@
 
 		handle_medical_side_effects()
 
-		if(!client && !mind)
-			species.handle_npc(src)
-
+		// If we have an AI, and we're in a fugue state or without client.
+		// Give control to AI.
+		if(istype(ai) && ((stat & FUGUE) || (!client && !mind && species)))
```
PsiOmegaDelta, line 86: Noting that having the FUGUE check here would prevent an AI from toggling off and then back on again

code/modules/mob/living/carbon/life.dm
biscuitWizard: Dunno why carbon/life has a separate life that doesn't call its predecessor life, but I didn't write the code and I have no desire to understand why that change was made. If it's not broke don't fix it.

code/__defines/mobs.dm
```diff
@@ -2,6 +2,7 @@
 #define CONSCIOUS   0
 #define UNCONSCIOUS 1
 #define DEAD        2
+#define FUGUE		4
```
PsiOmegaDelta: Inconsistent spacing, you're using tabs unlike the lines above

code/datums/ai/ai.dm
PsiOmegaDelta: Further suggestion
```
/datum/ai/proc/process(var/wait) // wait is provided by the mob system to /Life() and life could pass that on to the AI datum
    SHOULD_NOT_OVERRIDE(TRUE)
    if(!(body.stat & FUGUE))  // Noting that this still has the weakness of being unable to toggle the FUGUE-state as is
        return FALSE // No AI control, returning false as no processing was considered

    if(can_process())
        do_process()
    return TRUE // TRUE as processing was considered (but not necessarily executed)
```

PsiOmegaDelta: Callers of the AI system shouldn't have to know that can-process should be called before do-process, and preferably not having to conduct any additional checks such as FUGUE either.

code/modules/mob/living/carbon/human/life.dm
biscuitWizard: Not sure I understand why that matters since there's no toggle state machine on an AI.

code/datums/ai/ai.dm
PsiOmegaDelta: A method that checks if something can/should be done typically shouldn't alter state
Instead consider the more usual:
```
/datum/ai
    var/next_process = 0
    var/wait_for = 1

/datum/ai/proc/can_process()
    return world.time => next_process

/datum/ai/proc/process()
    if(can_process()
        next_process = world.time + wait_for
        do_process()
```

code/modules/mob/living/carbon/human/life.dm
PsiOmegaDelta: Yet. You are solidifying the inability to allow it in the future

code/datums/ai/ai.dm
biscuitWizard: process() is a terrible name for a function, given a function already exists with the name Process().

code/modules/mob/mob_defines.dm
PsiOmegaDelta: Sounds like you're proposing even more copy-paste, if you plan to continue calling AI-processing as you currently are

code/datums/ai/ai.dm
biscuitWizard: Don't use casing as a way to escape already declared methods. Rename it entirely or don't use it at all.

code/modules/mob/mob_defines.dm
biscuitWizard: I think you commented on the wrong review note, as that doesn't make sense in context here for this snippet.

code/datums/ai/ai.dm
biscuitWizard: Valid feedback.

code/__defines/mobs.dm
```diff
@@ -2,6 +2,7 @@
 #define CONSCIOUS   0
 #define UNCONSCIOUS 1
 #define DEAD        2
+#define FUGUE       4
```
Techhead0, line 5: Mob stat is not treated as a bitflag by existing code, this may be unsafe to do

MarinaGryphon, line 5: I'm not sure if there are any `stat >= DEAD` left, but it may not be intentional for fugue to be counted as "deader than dead".
Pre-posting edit: There are several instances of stat > which would probably be broken by `FUGUE > DEAD`. Since it's not exactly in-scope to change all of those, I'm not sure what to do about this.

Lohikar:
Per discussion on discord, do not add a new state to `stat`. `stat != 0` is treated as not conscious by much of the code.

PR #553, opened 2020-05-27
Refactored AI mind for species.

biscuitWizard:
As again, AI has been split out and properly refactored with previous concerns addressed.

Mobs can be overriden with AI.
This will pave the way also for simple_animal type AI splitting out.


code/datums/ai/ai.dm
```diff
@@ -1,5 +1,27 @@
...
+		return FALSE
+	if(body.stat == DEAD)
+		return FALSE
+	wait_for = world.time + run_interval
+	return TRUE
+
+/datum/ai/proc/do_process()
+	if(!can_process())
+		return FALSE
+	return TRUE
```
PsiOmegaDelta: By having the following setup things will be easier for subtypes:
```
/datum/ai/proc/process()
    if(can_process())
        do_process()
```
The reason being is that when a subtype overrides `do_process()` one does not have to remember to call `..()`, check the return value, and then return one's own appropriate value.

code/datums/ai/human.dm
```diff
@@ -0,0 +1,51 @@
...
+			else if(damaged_organ.burn_dam > damaged_organ.brute_dam && damaged_organ.organ_tag != BP_HEAD)
+				H.custom_emote("blows on [T.his] [damaged_organ.name] carefully.")
+			else
+				H.custom_emote("rubs [T.his] [damaged_organ.name] carefully.")
+
+		for(var/obj/item/organ/I in H.internal_organs)
+			if((I.status & ORGAN_DEAD) || BP_IS_PROSTHETIC(I)) continue
+			if(I.damage > 2) if(prob(2))
+				var/obj/item/organ/external/parent = H.get_organ(I.parent_organ)
+				H.custom_emote("clutches [T.his] [parent.name]!")
```
PsiOmegaDelta, line 47: If you keep your pattern can_process/do_process-pattern then all `do_process()` overrides should return TRUE, in the case of an additional additional subtyping

code/modules/mob/mob_defines.dm
```diff
@@ -186,3 +186,5 @@
 	var/list/additional_vision_handlers = list() //Basically a list of atoms from which additional vision data is retrieved
 
 	var/list/progressbars = null //for stacking do_after bars
+
+	var/datum/ai/ai						// Type abused. Define with path and will automagically create. Determines behaviour for clientless mobs.
```
PsiOmegaDelta, line 190: Can add a proc:
``` 
/mob/proc/get_ai_type()
    return ispath(ai) && ai
```
OR
```
/mob/proc/get_ai_type()
    if (ispath(ai))
        return ai
```


That way humans can override the proc via
```
/mob/living/human/get_ai_type()
    return ispath(ai) && (species ? species.ai : ..())
```
OR
```
/mob/living/human/get_ai_type()
    return ispath(ai) && (species?.ai || ..())
```
OR
```
/mob/living/human/get_ai_type()
    if (ispath(ai))
        if (species)
            return species.ai
        return ..()
```
Reasoning in the next comment.

Edit: The comments ended up out-of-order, instead see https://github.com/NebulaSS13/Nebula/pull/553#discussion_r434709292

code/modules/mob/mob.dm
```diff
@@ -52,6 +53,8 @@
 		move_intent = move_intents[1]
 	if(ispath(move_intent))
 		move_intent = decls_repository.get_decl(move_intent)
+	if(ispath(ai))
+		ai = new(src)
```
PsiOmegaDelta:
```
var/ai_type = get_ai_type()
if(ai_type)
    ai = new ai_type(src)
```

Edit: Due to comments being out of order this makes more sense by reading https://github.com/NebulaSS13/Nebula/pull/553#discussion_r434709292

PsiOmegaDelta: Even if you do not use the proc override this should instead be `ai = new ai(src)`, otherwise you'll only instantiate the type of the var `ai`, in this case `/datum/ai`, not whatever path that is assigned to it.

code/modules/mob/living/carbon/carbon.dm
```diff
@@ -6,6 +6,8 @@
 
 	if (!default_language && species_language)
 		default_language = species_language
+	if(!isnull(species) && ispath(species.ai))
+		ai = new species.ai(src)
```
PsiOmegaDelta: With the `get_ai_type()`-proc these two lines are no longer necessary, AI-instantiation can be handled in a single location.

By only having a single point of initiation we avoid the potential headaches of:
* Having to update multiple locations if AI initiation ever changes
* Having to keep in mind in which order the `/Initialize()`-overrides are called

code/datums/ai/ai.dm
```diff
@@ -1,5 +1,27 @@
...
-/datum/ai/proc/process()
-	return PROCESS_KILL
+/datum/ai/proc/can_process()
+	if((body.client || body.mind) && !(body.status_flags & ENABLE_AI))
+		return FALSE
+	if(wait_for > world.time)
+		return FALSE
+	if(body.stat == DEAD)
+		return FALSE
+	wait_for = world.time + run_interval
```
PsiOmegaDelta: A proper method that by its name implies that it only checks something (i.e.: Can Conduct Action, Is Something) should never cause a side-effects (That is: Change variable values, write to file, raise an event, etc.)

This is the expected behavior in general and one should not have to read the code of the proc to avoid confusion/surprise. It also makes it impractical to call such a method multiple times, if ever needed.

Better would be to have the proposed `/datum/ai/proc/process()` handle it via
```
/datum/ai
    var/next_process = 0
    var/run_interval = 1

/datum/ai/proc/can_process()
    if(next_process < world.time)
        return FALSE
    [...]

/datum/ai/proc/process()
    if(can_process())
        do_process()
        next_process = world.time + run_interval // Only calculate it once per process call - Set after do_process() to ensure can_process() still returns TRUE if called again (unless the AI has its mob commit suicide or the like)
```

I have to wonder why you have the `wait_for`/`run_interval`-vars though. It's already called in intervals via `/ Life()`

code/datums/ai/human.dm
```diff
@@ -0,0 +1,47 @@
...
+		var/maxdam = 0
+		var/obj/item/organ/external/damaged_organ = null
+		for(var/obj/item/organ/external/E in H.organs)
+			if(!E.can_feel_pain()) continue
+			var/dam = E.get_damage()
+			// make the choice of the organ depend on damage,
+			// but also sometimes use one of the less damaged ones
+			if(dam > maxdam && (maxdam == 0 || prob(50)) )
+				damaged_organ = E
+				maxdam = dam
```
MistakeNot4892, line 31: Probably should `break`.

code/datums/ai/ai.dm
```diff
@@ -1,5 +1,35 @@
...
+	if(!can_process())
+		return
+
+	var/time_elapsed = wait_for - world.time
+	wait_for = world.time + run_interval
+	do_process(time_elapsed)
+
+// This is the place to actually do work in the AI.
+/datum/ai/proc/do_process(var/time_elapsed)
+
```
MistakeNot4892, line 30: Aiee double newline.

code/datums/ai/ai.dm
```diff
@@ -1,5 +1,35 @@
 /datum/ai
 	var/name
+	var/mob/living/body		// The parent mob we control.
+	var/wait_for = 	0 		// The next time we can process.
+	var/run_interval = 1 	// How long to wait between processes.
+	
+/datum/ai/New(var/mob/living/body)
+	src.body = body
```
MistakeNot4892: Aiee var shadowing. Can you name the arg `new_body` or something rather than relying on `src.`?

code/datums/ai/ai.dm
```diff
@@ -1,5 +1,35 @@
...
+	wait_for = world.time + run_interval
+	do_process(time_elapsed)
+
+// This is the place to actually do work in the AI.
+/datum/ai/proc/do_process(var/time_elapsed)
+	
+
+/datum/ai/Destroy()
+	STOP_PROCESSING(SSai, src)
+	body = null
```
MistakeNot4892, line 34: `if(body.ai == src) body.ai = null`

code/datums/ai/human.dm
```diff
@@ -0,0 +1,47 @@
...
+		var/datum/gender/T = gender_datums[H.get_gender()]
+		if(damaged_organ)
+			if(damaged_organ.status & ORGAN_BLEEDING)
+				H.custom_emote("clutches [T.his] [damaged_organ.name], trying to stop the blood.")
+			else if(damaged_organ.status & ORGAN_BROKEN)
+				H.custom_emote("holds [T.his] [damaged_organ.name] carefully.")
+			else if(damaged_organ.burn_dam > damaged_organ.brute_dam && damaged_organ.organ_tag != BP_HEAD)
+				H.custom_emote("blows on [T.his] [damaged_organ.name] carefully.")
+			else
+				H.custom_emote("rubs [T.his] [damaged_organ.name] carefully.")
```
MistakeNot4892, line 41: Should these actions return afterwards or do you want them to pile up?

code/datums/ai/monkey.dm
```diff
@@ -0,0 +1,31 @@
+/datum/ai/monkey
+	name = "monkey"
+	var/list/no_touchie = list(/obj/item/mirror,
```
MistakeNot4892: Pls break this out like
````
	var/list/no_touchie = list(
		/obj/item/mirror,
		/obj/item/storage/mirror
	)
````
while you're in the neighborhood.

MistakeNot4892 (changes requested):
Other than the little tweaks suggested, looks more or less fine. I do wonder why `/datum/ai` isn't a `/decl` since it doesn't seem to be stateful - you might end up with a ton of these chilling out in memory if simple_animal logic is moved onto them for example.

biscuitWizard:
> Other than the little tweaks suggested, looks more or less fine. I do wonder why `/datum/ai` isn't a `/decl` since it doesn't seem to be stateful - you might end up with a ton of these chilling out in memory if simple_animal logic is moved onto them for example.

Things like /nymph are stateful.

Kharmaani berserk will be stateful. Would not make great decls.

code/datums/ai/human.dm
biscuitWizard: I don't want to change this logic for fear of regression.

code/datums/ai/ai.dm
biscuitWizard: Why?

code/datums/ai/human.dm
biscuitWizard: I don't want to change this logic for fear of regression.

code/datums/ai/ai.dm
MistakeNot4892: Because this will not GC while the mob holds a reference to it.

MistakeNot4892:
For reference only the GC issue is blocking, the others are suggestions.
Though I will probably have to PR a fix to that list formatting as I compulsively have to fix it when I see it.

code/datums/ai/ai.dm
biscuitWizard: It's QDEL_NULL'd in the mob.

MistakeNot4892: And if something deletes the datum rather than the mob, it won't GC.

biscuitWizard:
Changes request have been made.

MistakeNot4892 (approved):

PR #556, opened 2020-05-29
Pipe Fabricator Fixes and Adjustments

NataKilar:
Fixes pipe fabricator bugs with Heat Exchange pipes not being visible after wrenching down and many devices and pipes not being able to be properly rotated in hand until they were wrenched and unwrenched.

You can now recycle subtypes of /obj/item/pipe in the Fabricator. This does not return any of the matter used for things such as devices, but it should prevent clutter from building up.

### DIFF
--- a/code/modules/atmospherics/components/unary/outlet_injector.dm
+++ b/code/modules/atmospherics/components/unary/outlet_injector.dm
@@ -48,9 +48,9 @@
 
 	frame_type = /obj/item/pipe
 	construct_state = /decl/machine_construction/default/panel_closed/item_chassis
-	base_type = /obj/machinery/atmospherics/unary/vent_pump/buildable
+	base_type = /obj/machinery/atmospherics/unary/outlet_injector/buildable
 
-/obj/machinery/atmospherics/unary/vent_pump/buildable
+/obj/machinery/atmospherics/unary/outlet_injector/buildable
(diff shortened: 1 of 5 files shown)


code/modules/atmospherics/components/unary/outlet_injector.dm
```diff
@@ -48,7 +48,7 @@
 
 	frame_type = /obj/item/pipe
 	construct_state = /decl/machine_construction/default/panel_closed/item_chassis
-	base_type = /obj/machinery/atmospherics/unary/vent_pump/buildable
+	base_type = /obj/machinery/atmospherics/unary/outlet_injector
```
afterthought2: Probably needs to be the `/buildable` type below.

code/modules/fabrication/designs/pipe/device_pipe_datums.dm
```diff
@@ -26,11 +27,13 @@
 	build_icon_state = "uvent"
 	constructed_path = /obj/machinery/atmospherics/unary/vent_pump
 	pipe_class = PIPE_CLASS_UNARY
+	rotate_class = PIPE_ROTATE_STANDARD
```
afterthought2: Seems like these are mostly inherited?

afterthought2 (approved):

PR #561, opened 2020-05-30
Replaced skin, hair, eye and facial hair colour vars with single hex values.

MistakeNot4892:
- As title. 
- Renames several vars (`h_col`, `s_col`, `s_col_blend`) to be more immediately informative.
- Adds hex splitting to DNA instead of handling it everywhere else.


comma (approved):

code/modules/client/preference_setup/general/02_body.dm
```diff
@@ -65,24 +49,42 @@ var/global/list/valid_bloodtypes = list("A+", "A-", "B+", "B-", "AB+", "AB-", "O
 	pref.preview_icon = null
 	from_file(S["bgstate"], pref.bgstate)
 
+	// Grandfathering in older saves post colour rewrite.
```
PsiOmegaDelta, line 52: I'll note that `/datum/category_item/player_setup_item` datums have a  `/update_setup()`-proc just for this sort of thing

PR #566, opened 2020-06-01
Added zero markup to four vending machines that should be free.

### LINKED ISSUE #475: Vending Machines Charging Too Much For Supplies
#### Description of issue
Jimmy here, look, I get it. You guys need to cover overhead for the supplies we use on-station. I understand that. But 938 credits for a mixing bowl? Seriously? They can't be that hard to make. I'm sitting here, at the bar, and the bartender cant even serve me alcohol because they spent their last hundred seventy creds on the glass for my drink, and can't afford the booze to put in it. This is absurd.

It shouldn't cost the chef over 2,000 creds to make a cake, the upsell would be ridiculous. Not to mention Sec and Med not being able to do their jobs without paying for the tools.

#### Steps to reproduce
Security, Medical, and Kitchen Vending machines, check em for the prices.

The Machines in question:

- "NanoMed Plus"
- "Dinnerware"
- "Booze-O-Mat"
- "SecTech"

Unsure. At least as long as we've has this "new" station.

#### Issue bingo

MistakeNot4892:
Fixes https://github.com/ScavStation/ScavStation/issues/475, which is an issue upstream as well.

### DIFF
--- a/code/game/machinery/vending/food.dm
+++ b/code/game/machinery/vending/food.dm
@@ -122,6 +122,7 @@
 	desc = "A refrigerated vending unit for alcoholic beverages and alcoholic beverage accessories."
 	icon_state = "fridge_dark"
 	icon_deny = "fridge_dark-deny"
+	markup = 0
 	products = list(
 		/obj/item/chems/food/drinks/glass2/square = 10,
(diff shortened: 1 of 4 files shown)


comma (approved):
communism

PR #567, opened 2020-06-01
[WIP] Added DMAPI support from TGS.

biscuitWizard:
https://tgstation.github.io/tgstation-server/


code/game/world.dm
```diff
@@ -88,20 +88,24 @@
...
 	Master.Initialize(10, FALSE)
+	TgsInitializationComplete()
 
 #undef RECOMMENDED_VERSION
 
 var/world_topic_spam_protect_ip = "0.0.0.0"
 var/world_topic_spam_protect_time = world.timeofday
 
 /world/Topic(T, addr, master, key)
+	TGS_TOPIC
```
afterthought2, line 108: I feel this is potentially unsafe. 

At a minimum it needs to handle spam protection on input which fails the authentication check, in all cases (that seems to require extra logic from what is shown here). Spam protection seems to be handled via the two vars defined just above on an ad hoc basis.

Moreover, it's probably a good idea to handle the authentication step prior to the processing step, in particular prior to calling expensive stuff like db queries, json_decode on arbitrary input, file i/o, etc. Not doing so leaves you vulnerable to DOS shenanigans even without very high request count.

Would rather someone with better knowledge offered feedback, though.

code/__defines/tgs.dm
```diff
@@ -0,0 +1,35 @@
...
+#define TGS_DEFINE_AND_SET_GLOBAL(Name, Value) var/global/##Name = ##Value
+
+/// Read the value in the global variable `Name`.
+#define TGS_READ_GLOBAL(Name) global.##Name
+
+/// Set the value in the global variable `Name` to `Value`.
+#define TGS_WRITE_GLOBAL(Name, Value) global.##Name = ##Value
+
+/// Disallow ANYONE from reflecting a given `path`, security measure to prevent in-game use of DD -> TGS capabilities.
+#define TGS_PROTECT_DATUM(Path) //GENERAL_PROTECT_DATUM(##Path)
```
deathride58, line 14: The things you can do with direct var read/call access to the TGS datum are fairly benign across the board (you can't really do much remotely even if you have the secret api key of a given session), but it'd be worth looking into a protect datum define if it isn't already present.

deathride58 (commented):
In this PR's current state, it's a good enough barebones TGS implementation. TGS4 will function just fine with what's currently in this PR. However, it's currently missing the following:
- Chat integration (i recall nebcode having discord webhooks of some sort? it might be worth it to have those send broadcasts to channels with specific channel tags)
- Discord subsystem (This allows for things like allowing the server to link your ckey with discord and record that in the DB, which in turn allows use of the ingame notify verb which makes the server DM you on discord when the round ends, or even automatically granting roles if the config's set up for it)
- Basic chat commands, such as check (reports the current playercount, server address and other info), and notify (see above)

code/game/world.dm
deathride58: Actually, this entire Topic proc is currently unsafe, as the server writes to logs every single world Topic() call, spam prevention is completely missing from potentially performance-intensive calls such as `status`, and it does not take into account the potential for multiple IPs to be performing an attack simultaneously at any given time.
Aside from that, it's important that TGS is granted some form of bypass for all forms of topic spam prevention, as the watchdog is extremely chatty with the server, especially during bootup. But it'd have to be done in a way that also prevents forged/unauthorized input from abusing it to spam the server.

biscuitWizard:
Please add help wanted on this!

MistakeNot4892:
Closing a couple of stale draft PRs, please reopen when it's closer to being ready to go.

PR #578, opened 2020-06-06
Wall icon tweaks.

### LINKED ISSUE #574: Mineral walls do not update smoothing after being mined
#### Description of issue
image
dis

Expected them to update icon after loss of the sibling

#### Steps to reproduce
Go to rocks.
Mine one out with a drill.

dril

Since new mineral walls

#### Issue bingo

- [ ] [Couldn't find an existing issue about this](https://github.com/NebulaSS13/Nebula/issues)

MistakeNot4892:
- Wall icons now collect overlays to avoid appearance churn.
- Natural walls now have a shine overlay based on material reflectiveness (thanks Pawn).
- Wall refresh is now handled in `Destroy()`.
- `dismantle_wall()` is now used as a chain rather than overridden completely and returns the new turf as a result.

Fixes #574.


code/game/turfs/simulated/wall_icon.dm
```diff
@@ -66,54 +66,54 @@
...
+	var/new_overlays
 	var/material_icon_base = get_wall_state()
 	var/image/I
 	var/base_color = paint_color ? paint_color : material.icon_colour
 	if(!density)
 		I = image(icon, "[material_icon_base]fwall_open")
 		I.color = base_color
-		overlays += I
+		LAZYADD(new_overlays, I)
 		return
```
afterthought2, line 77: need to watch for early returns with this setup; this returns without setting overlays to `new_overlays`.

MistakeNot4892: Ah shoot so it does.

afterthought2 (approved):

code/game/turfs/simulated/walls.dm
```diff
@@ -57,7 +57,15 @@
 
 /turf/simulated/wall/Destroy()
 	STOP_PROCESSING(SSturf, src)
+	material = decls_repository.get_decl(MAT_PLACEHOLDER)
+	reinf_material = null
+	var/old_x = x
+	var/old_y = y
+	var/old_z = z
 	. = ..()
+	for(var/turf/simulated/wall/W in trange(1, locate(old_x, old_y, old_z)))
```
Lohikar, line 66: `RANGE_TURFS` pls.

Lohikar (changes requested):

Lohikar (approved):
Loaf has promised on penalty of teaching someone git that they'll purge the scourge of `trange` globally in a later PR.

PR #580, opened 2020-06-06
Replaced trange() with RANGE_TURFS().

MistakeNot4892:
Here you go @Lohikar.


Lohikar (approved):
Some poor sod doesn't get to learn git now. Woe.

PR #581, opened 2020-06-06
Converts all turf update_icons() to use SSoverlays.

MistakeNot4892:
@Lohikar

### DIFF
--- a/code/game/turfs/simulated/wall_icon.dm
+++ b/code/game/turfs/simulated/wall_icon.dm
@@ -58,10 +58,10 @@
 
 /turf/simulated/wall/on_update_icon()
 
-	..()
+	. = ..()
+	cut_overlays()
 
 	if(!material)
-		overlays.Cut()
 		return
 
 	if(!damage_overlays[1]) //list hasn't been populated; note that it is always of fixed length, so we must check for membership.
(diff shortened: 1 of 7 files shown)


afterthought2 (dismissed):

code/game/turfs/simulated/wall_icon.dm
```diff
@@ -73,47 +72,46 @@
 	if(!density)
 		I = image(icon, "[material_icon_base]fwall_open")
 		I.color = base_color
-		overlays = list(I)
+		add_overlay(I)
```
Lohikar, line 76: This should be `set_overlays()`, no?

code/game/turfs/turf.dm
```diff
@@ -49,6 +49,7 @@
 	update_starlight()
 
 /turf/on_update_icon()
+	cut_overlays()
```
Lohikar: It feels like this should be something done in individual turf update icon calls? Some turfs don't have overlays, and it's clearer if this is done closer to where the other overlay calls are.

Lohikar: For instance, I didn't realize until now that the above add_overlays() call is valid with this here.

code/game/turfs/simulated/wall_icon.dm
MistakeNot4892: Does it matter?

code/game/turfs/turf.dm
MistakeNot4892: I'm not sure I follow. The call chain should be preserved where possible and anything that cares about its own overlays will reapply them.

Lohikar: My complaint is about clarity, not actual function - I didn't realize that overlays were being cut elsewhere when I first read this PR. I feel like the `cut_overlays()` call should be with the rest of the appearance ops; in `on_update_icon()`.

The comment about some turfs not having overlays is a performance nit - calling this will trip an overlay compile, regardless if any overlays even exist or existed before.

Lohikar (dismissed):

Lohikar (approved):

PR #582, opened 2020-06-06
Added decompiler grenades.

MistakeNot4892:
Adds ~~recycler~~ decompiler grenades which are _completely legally distinct to the ones from Prey, for sure_. This was @Andromeda-K22's idea I just found myself tinkering with it this evening. ~~No way to get them currently, just adminspawn.~~ They have been added to the general uplink for 6 points apiece or 5 for 25.

They create a mini-singularity that will suck items into itself for ten seconds. Items that start the tick on the center turf will be decompiled into their raw matter. Mobs have a chance of having their equipped items yanked off, and humans have an additional chance of prosthetics getting chewed up.

At the end of the ten seconds the singularity will vanish and poop out a bunch of cubes of devoured matter, which will work in fabricators.


comma (dismissed):

code/game/objects/items/weapons/grenades/decompiler.dm
```diff
@@ -0,0 +1,142 @@
...
+/obj/effect/decompiler/Process()
+
+	FOR_DVIEW(var/atom/movable/thing, pull_range, loc, INVISIBILITY_LEVEL_TWO)
+		if(thing.simulated && !thing.anchored)
+			step_towards(thing, src)
+	END_FOR_DVIEW
+
+	if(world.time > expiry_time)
+		fade_out()
+		return
```
afterthought2: Feel like you should return PROCESS_KILL, as this proc is waitfor false and sleeps before qdeling, so this may fire multiple times.

afterthought2 (approved):

PR #585, opened 2020-06-07
Standardized material composition, added solvents.

MistakeNot4892:
- Grinding reagents no longer uses the `chemical_makeup` list for results, it just converts from sheet to reagent.
- Breathing reagents no longer uses `chemical_makeup`, it will just add units of that reagent to the bloodstream directly.
- Materials now have a general `solvent_power` value. This determines if they can remove ink from paper and which other materials will dissolve in them.
- Materials that previously had `chemical_makeup` now have `dissolves_into`. This is the product list used when the reagent is mixed with a strong enough solvent (acetone, ethanol, acid).
- Heating and chilling product lists now function as ratio lists like `chemical_makeup` (and now `dissolves_into`).
- Acid melting behavior is now at material level and is governed by `solvent_power` and some adjacent vars.


code/modules/codex/categories/category_materials.dm
```diff
@@ -69,12 +69,23 @@
...
+				if(mat.dissolves_in <= MAT_SOLVENT_NONE)
+					solvent_needed = "any liquid"
+				else if(mat.dissolves_in <= MAT_SOLVENT_MILD)
+					solvent_needed = "a mild solvent, like water"
+				else if(mat.dissolves_in <= MAT_SOLVENT_MODERATE)
+					solvent_needed = "a moderately strong solvent, like acetone"
+				else if(mat.dissolves_in <= MAT_SOLVENT_STRONG)
+					solvent_needed = "a strong solvent, like sulphuric acid"
+
+				material_info += "The following chemicals can be extracted from it with [solvent_needed] solvent:<br> [english_list(chems)]"
```
comma, line 97: add solvent properties of the chem itself too pls

code/modules/materials/_materials.dm
```diff
@@ -179,12 +178,19 @@
...
+
 	var/glass_icon = DRINK_ICON_DEFAULT
 	var/glass_name = "something"
 	var/glass_desc = "It's a glass of... what, exactly?"
 	var/list/glass_special = null // null equivalent to list()
 
 	// Matter state data.
+	var/dissolve_message = "dissolves in"
+	var/dissolve_sound = 'sound/effects/bubbles.ogg'
+	var/list/dissolves_in = MAT_SOLVENT_STRONG
```
Techhead0: Should not be `/list/`

MistakeNot4892: Whoops

code/modules/codex/categories/category_materials.dm
MistakeNot4892: Done.

afterthought2 (dismissed):

afterthought2 (dismissed):

afterthought2 (approved):

PR #588, opened 2020-06-07
Generalizes reagent toxicity to material level.

MistakeNot4892:


code/modules/materials/_materials.dm
```diff
@@ -77,6 +77,9 @@
 	var/mechanics_text
 	var/antag_text
 
+	var/toxicity = 0 // Organ damage from ingestion.
+	var/toxicity_targets_organ // Bypass liver/kidneys when ingested, harm this organ directly.
```
afterthought2: Helpful to indicate what class of values this should be set to, as there are a bunch of ways to refer to body parts.

code/modules/materials/_materials.dm
```diff
@@ -385,7 +388,25 @@
...
+			if(I)
+				var/can_damage = I.max_damage - I.damage
+				if(can_damage > 0)
+					if(dam > can_damage)
+						I.take_internal_damage(can_damage, silent=TRUE)
+						dam -= can_damage
+					else
+						I.take_internal_damage(dam, silent=TRUE)
+						dam = 0
+		if(dam)
```
afterthought2: Usually best to do `if(dam > 0)` or similar to avoid issues with float arithmetic.

afterthought2 (dismissed):

comma (approved):

PR #589, opened 2020-06-07
Moves dirtiness to material level, adds it to graphite, carbon, cleaner and antiseptics.

MistakeNot4892:


afterthought2 (dismissed):
Would consider adding comments by the var/ define at some point on the random oneoff vars like this, esp. as it's not obvious at a glance that this is the soap effect var.

comma (approved):

PR #594, opened 2020-06-07
Moved gas products from reagent splashes into a vapor_products list.

MistakeNot4892:


afterthought2 (approved):
Unsure that I like this behavior to begin with, but OK.

MistakeNot4892:
It'll be removed entirely when I add phase changes.

PR #616, opened 2020-06-11
Moves Fabrication SS init up in order

comma:
Roundstart lathes started with no recipies becaue fabrication is inited way too late for them to get the initial ones.


comma:
well that didn't work.

PR #622, opened 2020-06-12
Fixes lighting check to be inverted.

Andrew-Fall:
Fixes very bright light becoming black.

### DIFF
--- a/code/modules/lighting/lighting_overlay.dm
+++ b/code/modules/lighting/lighting_overlay.dm
@@ -97,7 +97,7 @@
 	var/set_luminosity = max > 1e-6
 	#endif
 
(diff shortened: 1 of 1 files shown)


afterthought2 (approved):
This is probably my fault.

afterthought2:
For reference, see this commit, where I was trying to fix the same thing, I believe: https://github.com/Baystation12/Baystation12/pull/23088/commits/495b0cd5ab9aa94eaa851392873ad6e589df1e7f

The only relevant change in it is to the sprite itself, in `lighting_overlay.dmi`.

@Lohikar in case it's relevant.

Lohikar:
Right, sorry for delay.

This looks valid given the inversion of baylights, though I haven't tested it.

Lohikar (approved):

PR #624, opened 2020-06-16
Adds support for forced spawnpoints

quardbreak:
With simply additions to maps, we can create forced spawnpoints for jobs.

### DIFF
--- a/code/game/jobs/job/_job.dm
+++ b/code/game/jobs/job/_job.dm
@@ -37,6 +37,7 @@
 
 	var/announced = TRUE                  //If their arrival is announced on radio
 	var/latejoin_at_spawnpoints           //If this job should use roundstart spawnpoints for latejoin (offstation jobs etc)
+	var/forced_spawnpoint                 //If this job should spawn on specific spawnpoint only. Requires spawnpoint name as string.
 
 	var/hud_icon						  //icon used for Sec HUD overlay
 
@@ -407,6 +408,9 @@
 	var/mob/H = C.mob
 	var/spawnpoint = C.prefs.spawnpoint
 	var/datum/spawnpoint/spawnpos
+
+	if(forced_spawnpoint)
+		spawnpoint = forced_spawnpoint
 
 	if(spawnpoint == DEFAULT_SPAWNPOINT_ID)
 		spawnpoint = GLOB.using_map.default_spawn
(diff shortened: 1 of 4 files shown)


code/game/jobs/job/_job.dm
```diff
@@ -37,6 +37,7 @@
 
 	var/announced = TRUE                  //If their arrival is announced on radio
 	var/latejoin_at_spawnpoints           //If this job should use roundstart spawnpoints for latejoin (offstation jobs etc)
+	var/forced_spawnpoint                 //If this job should spawn on specific spawnpoint only
```
comma: 1. Comment is confusing, it's implying that it's a flag, but it's a list of landmarks?
2. Nothing is actually setting it anywhere.

quardbreak: 1. It can be a list, or just text string which must reference to a real spawn point. I'll edit this message later then.
2. Because I don't want make it for tradeship right now. Conceptually we can spawn captain in his room, instead regular cryo. I can add this as feature later in this PR, just don't want mess with map right now.

comma: it's passed to pick() so it's confusing. Comment implies it's a flag either way pls change.

My point is there's no landmark /type/ that sets it, how are you going to set a /job datum var/ on map?

maps/tradeship/tradeship_spawnpoints.dm
```diff
@@ -1,13 +1,17 @@
...
 		GLOB.latejoin_cryo_two += loc
 		delete_me = 1
+	if(name == "JoinLateCryoCaptain")
+		GLOB.latejoin_cryo_captain += loc
+		delete_me = 1
 	. = ..()
 
 /datum/map/tradeship
-	allowed_spawns = list("Port Cryogenic Storage", "Starboard Cryogenic Storage", "Robot Storage")
+	allowed_spawns = list("Port Cryogenic Storage", "Starboard Cryogenic Storage", "Robot Storage", "Captain Compartment")
```
Techhead0, line 14: ~~Doesn't this make the spawn available for non-Captains?~~
EDIT: Ah, missed the `restrict_job`

code/game/jobs/job/_job.dm
```diff
@@ -37,6 +37,7 @@
 
 	var/announced = TRUE                  //If their arrival is announced on radio
 	var/latejoin_at_spawnpoints           //If this job should use roundstart spawnpoints for latejoin (offstation jobs etc)
+	var/forced_spawnpoint                 //If this job should spawn on specific spawnpoint only. Requires spawnpoint name as string.
```
MistakeNot4892, line 40: Would suggest changing the comment to say 'If set to a spawnpoint name, will use that spawn point for joining as this job' to be clearer.

MistakeNot4892:
Quard isn't around, this isn't getting touched, feel free to reopen if/when you are back.

PR #626, opened 2020-06-18
Adding some fun, silly faulty augments from CDDA.

biscuitWizard:
May use for future things. Maybe not. Was fun to code. Adds 8 faulty augments.


code/modules/augment/passive/faulty/locking_thumbs.dm
```diff
@@ -0,0 +1,39 @@
+/obj/item/organ/internal/augment/faulty/locking_thumbs
+	name = "noisemaker"
```
Geevies: needta fix this name and desc

Geevies (commented):
i really like this idea, think it's cute. might steal it, which means it'll get stolen twice in its lifetime >:D

code/modules/augment/passive/faulty/itchy.dm
```diff
@@ -0,0 +1,25 @@
+/obj/item/organ/internal/augment/faulty/itchy
+	name = "itchy metal thing"
+	desc = "An augment that was badly installed. Sometimes it itches horribly."
+	material = MAT_STEEL
+	var/timer_id
+
+/obj/item/organ/internal/augment/faulty/itchy/onInstall()
+	. = ..()
+	if(timer_id)
```
Lohikar: Might be better to just use an override timer instead of deleting it here.

code/modules/augment/passive/faulty/locking_thumbs.dm
```diff
@@ -0,0 +1,39 @@
...
+	desc = "An augment that was badly installed. Sometimes it emits a loud noise."
+	material = MAT_STEEL
+	allowed_organs = list(BP_AUGMENT_R_HAND, BP_AUGMENT_L_HAND)
+	var/timer_id
+	var/obj/item/organ/external/limb
+
+/obj/item/organ/internal/augment/faulty/locking_thumbs/onInstall()
+	. = ..()
+	limb = owner.get_organ(parent_organ)
+	if(timer_id)
```
Lohikar: ditto

code/modules/augment/passive/faulty/noisemaker.dm
```diff
@@ -0,0 +1,25 @@
+/obj/item/organ/internal/augment/faulty/noisemaker
+	name = "noisemaker"
+	desc = "An augment that was badly installed. Sometimes it emits a loud noise."
+	material = MAT_STEEL
+	var/timer_id
+
+/obj/item/organ/internal/augment/faulty/noisemaker/onInstall()
+	. = ..()
+	if(timer_id)
```
Lohikar: ditto

code/modules/augment/passive/faulty/locking_thumbs.dm
```diff
@@ -0,0 +1,18 @@
+/obj/item/organ/internal/augment/faulty/locking_thumbs
+	name = "self-locking thumbs"
+	desc = "An augment that was badly installed. Now featuring self-locking thumbs, which would otherwise be convenient but now just makes you drop things."
+	allowed_organs = list(BP_AUGMENT_R_HAND, BP_AUGMENT_L_HAND)
+
+
+/obj/item/organ/internal/augment/faulty/locking_thumbs/on_malfunction()
+	var/slot = null
+
+	if(limb.organ_tag in list(BP_L_ARM, BP_L_HAND))
```
Lohikar:
```dm
switch (limb.organ_tag)
    if (BP_L_ARM, BP_L_HAND)
        slot = slot_l_hand
    if (BP_R_ARM, BP_R_HAND)
        slot = slot_r_hand
```

code/modules/augment/passive/faulty/locking_thumbs.dm
```diff
@@ -0,0 +1,18 @@
...
+		slot = slot_l_hand
+	else if(limb.organ_tag in list(BP_R_ARM, BP_R_HAND))
+		slot = slot_r_hand
+
+	var/obj/item/thing = owner.get_equipped_item(slot)
+
+	if(thing && owner.unEquip(thing))
+		owner.visible_message("<B>\The [owner]</B> drops what they were holding, \his [limb] malfunctioning!")
```
Lohikar: Maybe add a self message here (2nd param)?

code/modules/augment/passive/faulty/noisemaker.dm
```diff
@@ -0,0 +1,6 @@
+/obj/item/organ/internal/augment/faulty/noisemaker
+	name = "noisemaker"
+	desc = "An augment that was badly installed. Sometimes it emits a loud noise."
+
+/obj/item/organ/internal/augment/faulty/noisemaker/on_malfunction()
+	owner.audible_message(SPAN_DANGER("[owner]'s body emits a loud, piezoelectric beep."), hearing_distance = 7)
```
Techhead0: You should actually play a beep noise for players

biscuitWizard: I can't find a SFX for an electric screech/hiss/beep/whine/awful noise.

Techhead0: I refuse to believe there's not a suitable one in the `sound/machines/` folder

biscuitWizard: Not one that's close enough to my mental imagination of the sound.

biscuitWizard: https://freesound.org/people/waveplay./sounds/515521/ I found something a little closer.

biscuitWizard: https://freesound.org/people/waveplay./sounds/395542/ Maybe?

biscuitWizard: this is very, very close https://freesound.org/people/waveplay./sounds/389400/

Techhead0: Okay, that sort of sound did not really come across in the description at all

biscuitWizard: I know! It's so hard to write good words.

Techhead0: Could try the airlock-crowbar noise sped up (frequency var, but it also pitch-shifts it), possibly reversed to obscure it. That'd prolly sound sufficiently grating. Failing that maybe try some of the other low mechanical noises

Kurgis:
Shouldn't these all have newlines, so there isn't an enforce newline PR down the line?

Lohikar:
I don't see any reason to enforce terminating newlines.

biscuitWizard:
If it were going to be enforced, it would just be enforced via git scripts anyways.

MistakeNot4892 (approved):
Bit pointless at the moment, but IMO would be good as a consequence of installing an implant and failing your skill check (replace the target implant with one of these).

PR #631, opened 2020-06-21
Attempting to simplify fluid<->reagent code for turf-based fluids later.

MistakeNot4892:
Infinite water bug was not fixed, this is a rather more drastic but effective solution (ha).
Probably breaks mob washing from flows but I don't really care about that right now since this is making it impossible to interact with water without causing an infinite flood.

Current plan is to make fluid overlays a visual thing only and have turfs handle flooding with their own reagents - this standardizes the fluid procs to reagent transfer to make that easier when I get to it.

### DIFF
--- a/code/game/atoms_fluids.dm
+++ b/code/game/atoms_fluids.dm
@@ -2,6 +2,7 @@
 	return
 
 /atom/proc/fluid_act(var/datum/reagents/fluids)
+	fluids.touch(src)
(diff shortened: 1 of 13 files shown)


MistakeNot4892:
Seems to work fine on Scav, definitely fixes the infinite water issue.

afterthought2 (approved):

PR #637, opened 2020-06-23
Made hands act as items.

MistakeNot4892:
This is sort of a weird idea. Essentially, when a human tries to do an unarmed click on someone, instead of just jumping to `attack_hand()` it now retrieves the associated organ and uses it as though attacking with it as an item. Hands have been changed to handle the checking and call `attack_hand()` in their `resolve_attackby()`. 

Upshots:
- Non-grasping or unwieldly limbs can now be handled by overriding the organ attackby (thinking mostly for Baxxid on Scav ie. https://github.com/ScavStation/ScavStation/issues/241).
- Augments could be checked for in the limb check rather than handled through hacking a tool item into the hand.
- Claws could be set up to handle engraving or damaging via harm intent rather than blocks of code in their associated interactions.
- Would make pizza hands more feasible in the future.

Opening as a draft to see thoughts and comments on it.

### DIFF
--- a/code/_onclick/other_mobs.dm
+++ b/code/_onclick/other_mobs.dm
@@ -9,18 +9,14 @@
 	Otherwise pretty standard.
 */
 /mob/living/carbon/human/UnarmedAttack(var/atom/A, var/proximity)
-
 	if(!..())
 		return
-
-	// Special glove functions:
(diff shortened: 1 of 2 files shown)


MistakeNot4892:
Closing off my drafts for now to clean up the PR list.

PR #639, opened 2020-06-30
Overmap movement generalization and extensionization

Mazianni:
Whew.

This PR is aimed to generalize a lot of movement-related code for overmaps to the very base of the inheritance tree ; not all objects will use this code, obviously.

A number of defines that were also shoved in ship.dm have been moved to the overmap define file and modified (just a little bit) to make them compile, namely by adding a second var to MOVING() and a third var to CHANGE_SPEED_BY().


code/modules/overmap/overmap_object.dm
```diff
@@ -8,8 +8,22 @@
...
+	var/vessel_size = SHIP_SIZE_LARGE	// arbitrary number, affects how likely are we to evade meteors
+	var/max_speed = 1/(1 SECOND)        // "speed of light" for the ship, in turfs/tick.
+	var/min_speed = 1/(2 MINUTES)       // Below this, we round speed to 0 to avoid math errors.
+
+	var/list/speed = list(0,0)          // speed in x,y direction
+	var/list/position = list(0,0)       // position within a tile.
+	var/last_burn = 0                   // worldtime when ship last acceleated
+	var/burn_delay = 1 SECOND           // how often ship can do burns
+
+	var/datum/extension/overmap_movement/movement //Movement handler.
```
MistakeNot4892: Would suggest setting this to the path and setting extension `if(ispath(movement, /datum/extension))` in `Initialize()`. `movement_handler_type` is otherwise unused, and the naming suggests it is a movement handler, which it isn't.

comma:
Why move it to an extension if it can only be set on one type? Can just move that code up to overmap object and it'll be same.

Mazianni:
I suppose you're right. I was initially recommended to move into an extension.

MistakeNot4892:
I think the actual comment was moving the pixel movement to an extension so that it could conceivably be generalized to like, projectiles, or putting things on tables/bumping things.

code/modules/overmap/overmap_object.dm
```diff
@@ -2,13 +2,28 @@
...
 
 	var/known = 1				 //shows up on nav computers automatically
 	var/scannable				 //if set to TRUE will show up on ship sensors for detailed scans, and will ping when detected by scanners.
 
 	var/requires_contact = FALSE //whether or not the effect must be identified by ship sensors before being seen.
 	var/instant_contact  = FALSE //do we instantly identify ourselves to any ship in sensors range?
+	var/halted = FALSE
+	var/can_move = FALSE
 	var/sensor_visibility = 10	 //how likely it is to increase identification process each scan.
+	var/list/consoles
```
comma: why are consoles and map_z list needed on base level? Kinda against the point of visitable sectors being a subtype.

code/modules/overmap/overmap_object.dm
```diff
@@ -2,13 +2,28 @@
...
+	var/list/map_z = list()
+
+	var/vessel_mass = 10000             // metric tonnes, very rough number, affects acceleration provided by engines
+	var/vessel_size = SHIP_SIZE_LARGE	// arbitrary number, affects how likely are we to evade meteors
```
comma: This is ship specific, no code in here uses it. Why move it up here?

code/modules/overmap/overmap_object.dm
```diff
@@ -42,6 +58,132 @@
...
+			var/delta_v = get_delta_v(TRUE, partial_power) / KM_OVERMAP_RATE
+			adjust_speed(-SIGN(speed[1]) * min(delta_v, abs(speed[1])), 0)
+		if (speed[2])
+			var/partial_power = Clamp(speed[2] / (get_delta_v() / KM_OVERMAP_RATE), 0, 1)
+			var/delta_v = get_delta_v(TRUE, partial_power) / KM_OVERMAP_RATE
+			adjust_speed(0, -SIGN(speed[2]) * min(delta_v, abs(speed[2])))
+		last_burn = world.time
+
+
+/obj/effect/overmap/proc/get_specific_wet_mass()
```
comma: this should get moved at least down to visitable, it's for wacky dynamic mass calculations, meaningless on objects without attached zlevels.

comma (approved):

PR #649, opened 2020-07-05
Refactoring tank bombs.

MistakeNot4892:
- Generalized terminology of tank bomb vars to accelerant and oxidizer instead of specifically phoron and oxygen.
- Added a hydrogen preset for tank transfer bombs for testing.
- ~~Changed hydrogen gas specific heat and molar mass values to follow phoron, to be equivalent in bombmaking.~~ Turned out to be unnecessary, they go bang just fine without these vars being set.
- Cleaned up single tank bomb code.


MistakeNot4892:
Hydrogen seems to work fine as a substitute for phoron, but the premade spawned bombs are producing absolutely pathetic explosions (0,1,2). Seems to be a wider issue than just the gas being used, so calling this PR ready to go.

afterthought2 (dismissed):

afterthought2 (approved):

biscuitWizard:
Not a note for this PR, but would like to see these merged in with /datum/extension/assembly to remove the weird one-off assembly code.

Must standardize item construction.

PR #652, opened 2020-07-05
Bumped Travis up to 513.

MistakeNot4892:
I have no idea what else needs to be fixed in the codebase before we move to 513, pls advise.


SierraKomodo:
The BS12 discord seems to feel that the main issue with 513 is string sanitization in user inputs because 513 allows characters that 512 didn't.

MistakeNot4892:
Will reopen when @Lohikar finishes the copytext_char PR or gives up.

PR #654, opened 2020-07-07
Null checking a strange error causing a lot of runtimes on Scav.

MistakeNot4892:


deathride58 (approved):

afterthought2 (approved):
Should maybe nullcheck in the move handler instead? Deleting mobs during moves might actually be a reasonable thing.

PR #655, opened 2020-07-07
Add configuring readout threshholds in supermatter monitoring

SierraKomodo:
:cl: SierraKomodo
tweak: Supermatter reading thresholds (The point at which the numbers turn yellow and red) can now be configured. The configuration is stored per supermatter crystal and syncs between all programs/subsystems looking at that supermatter. This applies to EER, Temperature, Pressure, and EPR readings. Does not apply to integrity, icon state, or the delamination radio broadcast.
/:cl:

vlc_Tvi9DfvTP5

### DIFF
--- a/code/__defines/machinery.dm
+++ b/code/__defines/machinery.dm
@@ -116,6 +116,11 @@ var/list/restricted_camera_networks = list(NETWORK_ERT, NETWORK_MERCENARY, NETWO
 #define SUPERMATTER_EMERGENCY 5		// Integrity < 25%
 #define SUPERMATTER_DELAMINATING 6	// Pretty obvious.
 
+#define SUPERMATTER_DATA_EER         "Relative EER"
+#define SUPERMATTER_DATA_TEMPERATURE "Temperature"
+#define SUPERMATTER_DATA_PRESSURE    "Pressure"
+#define SUPERMATTER_DATA_EPR         "Chamber EPR"
+
 // Scrubber modes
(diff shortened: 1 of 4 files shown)


MistakeNot4892 (dismissed):

afterthought2:
Feels really odd to store it on the crystal; wouldn't it make more sense to store it on the nanomodule, or on the hard drive in a config file if you want to be fancy?

SierraKomodo:
> Feels really odd to store it on the crystal; wouldn't it make more sense to store it on the nano module, or on the hard drive in a config file if you want to be fancy?

Storing it on the nano module or hard drive presents the issue that the settings only apply to that specific console instead of to every monitor for that crystal - The intent here is for the configured thresholds to be 'global' for anyone looking at the monitor so that you don't need to reconfigure every single tablet, laptop, console, and AI subsystem to match your settings, while also being different per crystal because the thresholds you want to watch for might be different.

SierraKomodo:
I just noticed an issue with one of the thresholds not coloring EER for some reason so going to fix that.

SierraKomodo:
Fixed the issue, although I wasn't able to rebase on `/dev` for this due to a compile error currently on dev. Ready for re-review, though the only change is the order in which the `get_threshhold_color()` proc checks the min/max vars so that max_h is applied regardless of max_l.

MistakeNot4892 (approved):

PR #663, opened 2020-07-10
Pipe mapping QoL.

Mazianni:
Y'know how north and west facing pipes are bad?

How travis just hates them?

How annoying it is to pick through a map to find them all?

WELL, SUFFER NO MORE, FOR I HAVE INVENTED A VISUAL INDICATION DEVICE!

https://cdn.discordapp.com/attachments/728782430781177936/730998022523060304/unknown.png


MistakeNot4892:
Apparently regular pipes are fine as they use a separate icon.

MistakeNot4892 (approved):

PR #665, opened 2020-07-11
Generalizing active and inactive hands.

MistakeNot4892:
Replaces the current hardcoded `l_hand` and `r_hand` slots with a datum system that is generated from your limbs - your hands create inventory slots that your mob then respects, draws a UI for, etc. Rewrites and renames a ton of backend procs.

Also messes with Serpentid and Kharmaani hands so they can be the display model for the feature.

Species now have a `limb_mapping` var which generally autopopulates but can be set up to map bodyparts from the zone selector onto extra limbs for nonstandard species. `check_zone()` will return a random entry from the mapping list, and surgery will prompt the surgeon to select the target. This isn't ideal but it's as good as I could get it for this PR.

image

**TODO:**
- [X] Main system conversion/implementation.
- [x] Secondary hand icons for other UI styles.
- [x] Reimplement Pick Up verb.
- [x] Get basic inhands for left and right hand generating again.
- [x] Add a sane/clean way for per-bodypart per-species inhand offsets to be applied while using the existing l_hand/r_hand icon states.
- [x] Add workaround for slot_l_hand/slot_r_hand being invalid now.
- [x] Add a multi-arm demo species maybe? Went with Serpentid and Kharmaani.
- [x] Add midlimb for Kharmaani.
- [x] Add genereal zone-to-bodypart species mapping proc for attacking creatures with nonstandard limbs.
- [x] Check if handcuffs still work!
- [x] Optimization/macro pass.
- [x] Tidy-up and code relocation pass.

**Bugs:**
- ~~`holding` var on held item slot is not clearing properly on drag and drop equip or regular equip.~~
- ~~Slots in general are not properly clearing in `u_equip()`.~~
- ~~`get_organ()` returning non-existent organs.~~

**Next time, Gadget:**
- Generalize `screen_loc` generation for 2+ hands. - need better handling of screenloc for all inventory UI
- Update/rename/reimplement in-hand sprites for held items and remove the lookup table workaround.


MistakeNot4892:
This is probably in a good place for initial reviews, everything is working without any visible bugs.

MistakeNot4892:
I would like to playtest this at some point but it is otherwise ready for review.

code/_onclick/hud/hud.dm
```diff
@@ -156,13 +154,61 @@
...
+
+	if(isnull(removing))
+		if(!skip_client_update)
+			mymob?.client?.screen -= hand_hud_objects
+		QDEL_NULL_LIST(hand_hud_objects)
+	else
+		for(var/bp in removing)
+			for(var/obj/screen/inventory/inv_box in hand_hud_objects)
+				if(inv_box.slot_id == bp)
+					mymob?.client?.screen -= inv_box
```
afterthought2: This kind of thing doesn't work the way you want; if there's no client you're still doing `null -= X`. There are several similar constructions in this proc.

code/modules/mob/living/carbon/human/inventory.dm
```diff
@@ -112,123 +75,132 @@ This saves us from having to call add_fingerprint() any time something is put in
...
-	W.forceMove(src)
+	. = ..()
+	if(!. || !has_organ_for_slot(slot))
+		return
 
+	// TODO: unify this block with below when inventory
+	// is rewritten to remove boilerplate horseshit.
+	u_equip(W)
 	var/obj/item/old_item = get_equipped_item(slot)
+	if(!isnum(slot))
```
afterthought2, line 193: I'm pretty lost here (apparently can be num or string, giving different function?). I appreciate that there's a todo above, but may help to give some comments explaining roughly what's going on here and what the expectations are. What do the return values mean? What is `slot` here? etc.

code/modules/mob/living/deity/phenomena/starlight.dm
```diff
@@ -70,7 +70,7 @@
 			extension.expected_helmet = helm.type //We only do by type because A. its easier to manage and B the chances of it being non-unique in a normal game is very small
 		if(form["weapon"])
 			var/w_type = form["weapon"]
-			L.equip_to_slot_or_store_or_drop(new w_type(T), slot_r_hand)
+			L.put_in_hands(new w_type(T))
```
afterthought2: I'll be honest; I've been glossing over the mass-replace stuff. But in some cases change of function does matter somewhat: e.g. for the antag item ones it's somewhat relevant that they are stored prior to dropped. May be worth checking a bit more (not really pointing to this one in particular; I can't tell from the context if it matters here or even what mob type this is).

code/modules/mob/inventory.dm
```diff
@@ -117,39 +107,45 @@ var/list/slot_equipment_priority = list( \
...
+	return
+	
 /mob/proc/put_in_active_hand(var/obj/item/W)
-	return 0 // Moved to human procs because only they need to use hands.
+	. = equip_to_slot_if_possible(W, get_active_held_item_slot())
 
 //Puts the item into our inactive hand if possible. returns 1 on success.
 /mob/proc/put_in_inactive_hand(var/obj/item/W)
-	return 0 // As above.
+	. = equip_to_slot_if_possible(W, get_empty_hand_slot())
```
afterthought2: Does this actually do what it says on the tin? Seems like it puts it in the first available empty hand slot, regardless of whether it's active.

code/modules/mob/inventory.dm
```diff
@@ -167,21 +163,23 @@ var/list/slot_equipment_priority = list( \
...
 //Drops the item in our left hand
-/mob/proc/drop_l_hand(var/atom/Target)
-	return drop_from_inventory(l_hand, Target)
+/mob/proc/drop_from_hand(var/slot, var/atom/Target)
+	return FALSE
 
-//Drops the item in our right hand
-/mob/proc/drop_r_hand(var/atom/Target)
-	return drop_from_inventory(r_hand, Target)
+/mob/living/drop_from_hand(var/slot, var/atom/Target)
```
afterthought2: Should be in `living/inventory.dm`.

code/modules/mob/inventory.dm
```diff
@@ -197,19 +195,28 @@ var/list/slot_equipment_priority = list( \
...
-	else if (W == wear_mask)
+		return TRUE
+	if(W == wear_mask)
 		wear_mask = null
 		update_inv_wear_mask(0)
-	return
+		return TRUE
+	return FALSE
+
+/mob/living/u_equip(obj/W)
```
afterthought2: Should be in `living/inventory.dm`.

code/modules/mob/inventory.dm
```diff
@@ -263,24 +270,27 @@ var/list/slot_equipment_priority = list( \
 		I.dropped(src)
 	return 1
 
+/mob/proc/drop_held()
+	unEquip(get_active_held_item())
```
afterthought2: This implementation seems redundant with `unequip_item` above and also inconsistent with the name; did you intend for it to drop all held items instead?

code/modules/mob/living/carbon/human/human_defense.dm
```diff
@@ -110,14 +110,12 @@ meteor_act
...
 			return gear
-	return null
 
 /mob/living/carbon/human/proc/check_shields(var/damage = 0, var/atom/damage_source = null, var/mob/attacker = null, var/def_zone = null, var/attack_text = "the attack")
-	for(var/obj/item/shield in list(l_hand, r_hand, wear_suit))
-		if(!shield) continue
-		. = shield.handle_shield(src, damage, damage_source, attacker, def_zone, attack_text)
-		if(.) return
-	return 0
+	for(var/obj/item/shield in (get_held_items() | list(wear_suit)))
```
afterthought2: `get_held_items()` appears to return a lazylist, which is safe to iterate but not if you do list ops on it like here. I believe you do this in multiple places, so worth searching the proc and checking.

code/modules/species/species.dm
```diff
@@ -197,6 +197,7 @@
 
 	var/list/skin_overlays = list()
 
+	var/list/limb_mapping
```
afterthought2: Suggest including a comment here about the expected format (or a comment saying this is internal).

code/modules/species/species.dm
```diff
@@ -342,18 +343,24 @@ The slots that you can use are found in items_clothing.dm and are the inventory
 			has_limbs[ltag] = list("path" = override_limb_types[ltag])
 
 	//Build organ descriptors
+	LAZYINITLIST(limb_mapping)
 	for(var/limb_type in has_limbs)
 		var/list/organ_data = has_limbs[limb_type]
 		var/obj/item/organ/limb_path = organ_data["path"]
 		organ_data["descriptor"] = initial(limb_path.name)
+		LAZYDISTINCTADD(limb_mapping[limb_type], limb_type)
```
afterthought2: Did you mean the second argument to the macro to be `organ_data`? I'm a bit lost here.

MistakeNot4892: This block is used to make sure a given tag maps to itself but I suppose I could make the default behavior to not try to map it at all. Might be clearer.

MistakeNot4892: Just noticed that the default behavior is to do this, so I will just remove the list setting from species init entirely.

code/modules/species/species.dm
MistakeNot4892: This has been split into a separate PR and a comment has been added.

code/modules/mob/inventory.dm
MistakeNot4892: Added a `get_empty_hand_slots()` and a check for held item slot.

code/modules/mob/living/carbon/human/human_defense.dm
MistakeNot4892: According to Wrench:
````
json_encode(null | list("foo")) => "[\"foo\"]"
````

Comment (MistakeNot4892) [reply code/modules/mob/living/carbon/human/human_defense.dm]: For the sake of clarity I went through and broke these ops out into assignments.

File: code/modules/mob/inventory.dm (hunk 7 above)
Comment (MistakeNot4892) [reply code/modules/mob/inventory.dm]: I have swapped `unequip_item` out for `unEquip(item)` or `drop_held` and removed it entirely, also renamed `drop_held` to `drop_held_items`.

File: code/modules/mob/living/deity/phenomena/starlight.dm (hunk 3 above)
Comment (MistakeNot4892) [reply code/modules/mob/living/deity/phenomena/starlight.dm]: I've added a `put_in_hands_or_store_or_drop` and will go through and make sure I swapped the right procs out for it.

File: code/_onclick/hud/hud.dm [hunk 11]
```diff
@@ -156,13 +154,62 @@
...
+			if(existing_box.slot_id == bp)
+				inv_box = existing_box
+				break
+		if(!inv_box)
+			inv_box = new /obj/screen/inventory()
+		var/datum/inventory_slot/inv_slot = target.held_item_slots[bp]
+		inv_box.SetName(bp)
+		inv_box.icon = ui_style
+		inv_box.icon_state = "hand_base"
+		inv_box.overlays += "hand_[bp]"
```
Lohikar: Probably should `add_overlay()` these w/ a `compile_overlays()` at the end.

MistakeNot4892:
Deps are #728 and #729.

code/modules/mob/inventory.dm
```diff
@@ -172,16 +181,12 @@ var/list/slot_equipment_priority = list( \
 	return 0
 
 //Drops the item in our left hand
```
afterthought2: Incorrect description.

afterthought2 (commented):
Only found one minor thing.

If you haven't already, I'd suggest checking that borgs work OK, as you modified a bunch of inventory API stuff at /living level and traditionally borgs override these calls. Looks fine otherwise.

MistakeNot4892:
Tested robots, everything seems to work fine with them.

afterthought2 (approved):

MistakeNot4892:
🎊

PR #674, opened 2020-07-13
Added capacitor weapons.

MistakeNot4892:
Adds a capacitor rifle and a capacitor pistol, ~~loosely based on~~ completely unrelated to the Sleeper Simulant from Destiny. Tldr for capacitor weapons - they start with a removable cell and a set of capacitors. Each shot has to charge for a short time before it actually fires the beam. Capacitor size determines the amount of power needed and the damage/armour pen of the resulting beam (basic types are 28 damage for pistol, 40 damage for rifle), capacitor rating determines how fast shots charge.

You can also dial your capacitor gun to a specific wavelength to move between damage and armour penetration. Each wavelength has an associated colour.


code/modules/projectiles/guns/energy/capacitor.dm
```diff
@@ -0,0 +1,163 @@
...
+/obj/item/gun/energy/capacitor/Destroy()
+	if(capacitors)
+		QDEL_NULL_LIST(capacitors)
+	. = ..()
+
+/obj/item/gun/energy/capacitor/Initialize()
+	var/capacitor_volume = 0
+	for(var/i = 1 to max_capacitors)
+		var/obj/item/stock_parts/capacitor/capacitor = new capacitor_type(src)
+		capacitor_volume = min(capacitor_volume, capacitor.max_charge)
```
afterthought2: This local var appears unused.

code/modules/projectiles/guns/energy/capacitor.dm
```diff
@@ -0,0 +1,163 @@
...
+	. = !charging && ..()
+	
+/obj/item/gun/energy/capacitor/attackby(obj/item/W, mob/user)
+	if(!charging)
+
+		if(isScrewdriver(W))
+			if(length(capacitors))
+				var/obj/item/capacitor = capacitors[1]
+				capacitor.dropInto(user.loc)
+				user.put_in_hands(capacitor)
```
afterthought2: Dropping into `user.loc` is redundant with `put_in_hands`, as that's the fail behavior anyway. Unsure if you meant `loc` for action-at-a-distance stuff or if it should be dropped.

code/modules/projectiles/guns/energy/capacitor.dm
```diff
@@ -0,0 +1,163 @@
...
+				LAZYREMOVE(capacitors, capacitor)
+			else if(power_supply)
+				power_supply.dropInto(loc)
```
afterthought2: Similar to above, except this one uses item's loc.

code/modules/projectiles/guns/energy/capacitor.dm
```diff
@@ -0,0 +1,163 @@
...
+		while(!QDELETED(user) && length(capacitors) && charging && user.get_active_hand() == src)
+			var/charged = TRUE
+			for(var/obj/item/stock_parts/capacitor/capacitor in capacitors)
+				if(capacitor.charge < capacitor.max_charge)
+					charged = FALSE
+					tried_charge = TRUE
+					var/use_charge_cost = min(charge_cost * capacitor.rating, round((capacitor.max_charge - capacitor.charge) / capacitor_charge_constant))
+					if(power_supply.use(use_charge_cost))
+						capacitor.charge(use_charge_cost * capacitor_charge_constant)
+						update_icon()
```
afterthought2, line 172: Do we really need to be doing this 10 times a second? Could we use the obj or the fast processing SS instead?

MistakeNot4892: As discussed on Discord, using processing would be a pain. I've lowered the number of times this will fire each second though.

afterthought2 (approved):

afterthought2 (approved):

PR #681, opened 2020-07-14
Splitting backpacks out to on_mob_icon system.

MistakeNot4892:
Also updated existing `icon_state = "world"` lines to use the state defines.


MistakeNot4892:
I did not exhaustively test all backpacks, but the 10 or so I spawned worked fine, the ERT ones use the correct overlays, and I populated the icon states with copypaste so there shouldn't be many (or any) errors. Ready for review.

comma (approved):

PR #684, opened 2020-07-15
Attempting to fix chem grenades.

MistakeNot4892:
It looks like these were never updated for scheduler-managed reactions, and also never updated for fluids. So:
- Chem grenades no longer touch every turf in view with their reagents.
- Chem grenades will now delay up to 5 seconds to allow their contents to react.
- At the end of the five seconds, the grenade will dump any remaining reagents to the turf as a fluid.


MistakeNot4892:
Tested with all chem grenades, they're all working including incendiary.

afterthought2 (approved):

PR #686, opened 2020-07-15
Sets mouse_opacity on items when entering and leaving storage.

MistakeNot4892:
QOL idea from @KathyRyals, thanks very much.


afterthought2 (approved):

MistakeNot4892:
This causes problems with overlapping small items in backpacks.

MistakeNot4892:
Going to need more discussion/work.

PR #696, opened 2020-07-17
Long range holopads can only connect to other long range holopads

SierraKomodo:
:cl:
tweak: Long-distance holopad calls now require both holopads to be long-range instead of just the calling pad. Also means AI cannot do long distance holopadding unless it's on a long-range pad.
/:cl:

### DIFF
--- a/code/game/machinery/hologram.dm
+++ b/code/game/machinery/hologram.dm
@@ -93,6 +93,8 @@ var/const/HOLOPAD_MODE = RANGE_BASED
 				var/area/area = get_area(src)
 				for(var/mob/living/silicon/ai/AI in GLOB.living_mob_list_)
 					if(!AI.client)	continue
+					if (holopadType != HOLOPAD_LONG_RANGE && !AreConnectedZLevels(AI.z, src.z))
+						continue
 					to_chat(AI, "<span class='info'>Your presence is requested at <a href='?src=\ref[AI];jumptoholopad=\ref[src]'>\the [area]</a>.</span>")
 			else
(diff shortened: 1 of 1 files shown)


code/game/machinery/hologram.dm
```diff
@@ -104,14 +106,18 @@ var/const/HOLOPAD_MODE = RANGE_BASED
...
 						if(!isnull(O))
-							zlevels |= O.map_z
+							zlevels_long |= O.map_z
 				for(var/obj/machinery/hologram/holopad/H in SSmachines.machinery)
-					if((H.z in zlevels) && H.operable())
-						holopadlist["[H.loc.loc.name]"] = H	//Define a list and fill it with the area of every holopad in the world
+					if (H.operable())
+						if(H.z in zlevels)
+							holopadlist["[H.loc.loc.name]"] = H	//Define a list and fill it with the area of every holopad in the world
+						if (H.holopadType == HOLOPAD_LONG_RANGE && (H.z in zlevels_long))
```
MistakeNot4892, line 119: Won't this still be able to target short range holopads?

SierraKomodo: No - The `H.holopadtype == HOLOPAD_LONG_RANGE` check ensures that in the if statement for line 119. Essentially, line 117 checks for holopads in connected z levels, and line 119 checks for long-range holopads in unconnected z levels.

MistakeNot4892: Alrighty.

MistakeNot4892 (approved):

PR #699, opened 2020-07-18
Moves all suits to single dmi system

comma:
Also removed some unncessary icon state overrides where base type is ported (head items)


MistakeNot4892 (approved):
Bloody hell that is a lot of .dmis.

PR #700, opened 2020-07-18
Block access to objects that are not on connected z-levels

SierraKomodo:
Port of https://github.com/Baystation12/Baystation12/pull/28889 from BS12

:cl:
rscdel: AI can no longer access objects in a non-connected z-level, including shuttles at away sites. Distant holopads are still accessible.
/:cl:

### DIFF
--- a/code/modules/nano/interaction/default.dm
+++ b/code/modules/nano/interaction/default.dm
@@ -40,7 +40,8 @@ GLOBAL_DATUM_INIT(default_state, /datum/topic_state/default, new)
 	// Prevents the AI from using Topic on admin levels (by for example viewing through the court/thunderdome cameras)
 	// unless it's on the same level as the object it's interacting with.
 	var/turf/T = get_turf(src_object)
-	if(!T || !(z == T.z || (T.z in GLOB.using_map.player_levels)))
+	var/turf/A = get_turf(src)
+	if(!A || !T || !AreConnectedZLevels(A.z, T.z))
 		return STATUS_CLOSE
 
 	// If an object is in view then we can interact with it


code/game/machinery/_machines_base/machinery.dm
```diff
@@ -91,6 +91,7 @@ Class Procs:
 	var/malf_upgraded = 0
 	var/datum/wires/wires //wire datum, if any. If you place a type path, it will be autoinitialized.
 	var/use_power = POWER_USE_IDLE
+	var/silicon_restriction = FALSE // FALSE or one of the STATUS_* flags. If set, will force the given status flag if a silicon tries to access the machine.
```
PsiOmegaDelta: Kind of odd placement, in the middle of several power based vars

nano/templates/shipsensors.tmpl
```diff
@@ -1,6 +1,15 @@
+{{if data.viewing_silicon}}
+	<div class="block">
+		<div class="item">
+			<div class="itemContent bad">
+				Remote AI systems detected. Firewall protections forbid remote AI control.
+			</div>
+		</div>
+	</div>
+{{/if}}
```
PsiOmegaDelta: These blocks, and the data-part, feel somewhat copy-pasty. Hmm

SierraKomodo: I couldn't figure out a way to handle this otherwise - NanoUI code looks like too much of a mess to try and find a way to have that inserted outside of the .tmpl files.

code/game/machinery/_machines_base/machinery.dm
SierraKomodo: Moved to the end of the vars list. Not sure why I put it there tbh.

MistakeNot4892:
I can understand the remote z-level restriction, but why restrict AI from flying the ship? Feels like a Bay drama change.

SierraKomodo:
Dropped the helm restrictions.

MistakeNot4892 (approved):

PR #702, opened 2020-07-18
Standardize borg languages and remove mech skill from borgs

SierraKomodo:
:cl:
tweak: Languages across all borg modules are now standardized: All borgs understand and speak all human languages, and can understand but not speak sign language.
tweak: Maintenance drones can understand common but cannot speak it, and do not understand any other language.
rscdel: Mech piloting skill has been removed from all borg modules.
/:cl:


SierraKomodo:
Fixed a compile error that I missed

MistakeNot4892:
One of the practical concessions made to drone/human interaction was that humans could tell a drone to stop and go away - removing that neans the only way to peacefully deal with a drone toeing the line is to deactivate it. Also feels like it makes the role much more isolated, not being able to listen to the radio. Not a fan.

SierraKomodo:
Gave common back to maintenance drones.

MistakeNot4892 (approved):

PR #706, opened 2020-07-19
Fixes Z copying floor decals

Andrew-Fall:
Fixes decals from /turf/simulated/floor
Removes /image/proc/plating_decal_layerise() and /image/proc/turf_decal_layerise()
Replaces instances of them with assignments to the layer 
Current issue:
image
Fixed:
image

### DIFF
--- a/code/__defines/_planes+layers.dm
+++ b/code/__defines/_planes+layers.dm
@@ -197,14 +197,6 @@ What is the naming convention for planes or layers?
 /atom
 	plane = DEFAULT_PLANE
 
-/image/proc/plating_decal_layerise()
-	plane = DEFAULT_PLANE
-	layer = DECAL_PLATING_LAYER
-
-/image/proc/turf_decal_layerise()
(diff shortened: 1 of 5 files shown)


MistakeNot4892 (dismissed):

PsiOmegaDelta:
Is this not a problem for the other 5 locations that also call `turf_decal_layerise()`?

Andrew-Fall:
`> Is this not a problem for the other 5 locations that also call turf_decal_layerise()?`
It probably is, should I remove setting of the plane in that proc? I assumed the other calls had a reason to use that instead of just setting the layer.

Andrew-Fall:
I tested it with natural walls with ore, which were affected by the removal of the image procs. It appears to work fine.

MistakeNot4892 (approved):
Discussion generally indicates this whole system is an artifact of the Bay superlayer-plane setup that has been removed since.

PR #710, opened 2020-07-20
Ports eye blink from Infinity Station

quardbreak:
Awful implementation which require some help.

### DIFF
--- a/code/modules/mob/living/carbon/human/eye_blink.dm
+++ b/code/modules/mob/living/carbon/human/eye_blink.dm
@@ -1,0 +1,49 @@
+/mob/living/carbon/human/proc/handle_eye_blink()
+	eye_blink()
+
+/mob/living/carbon/human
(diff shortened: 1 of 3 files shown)


MistakeNot4892 (changes requested):
This code is really, really bad, to the point that there's no individual part of it that is acceptable.

Blinking also requires rebuilding the head icon, which means rebuilding the body icon. Way too much overhead for the sake of one missing frame.

It would be better to just make the eye overlay animated with an occasional pause.

MistakeNot4892:
Going to be making a unilateral decision to close this one, sorry. As written, this PR is not going to get merged, and I think fundamentally the method you're using is not desirable as rebuilding the entire human icon literally every time they blink is not worth the immersion of having blinking spacemen.

Splitting the eyes away from the head and making them an overlay that can be added or removed easily would be much preferred. 

Feel free to open another PR with a different method.

PR #712, opened 2020-07-20
Upgrades Who verb, makes it as separate window

quardbreak:
Code is ugly, but now in separate window.
image

### DIFF
--- a/code/game/verbs/who.dm
+++ b/code/game/verbs/who.dm
@@ -1,13 +1,20 @@
-
 /client/verb/who()
 	set name = "Who"
 	set category = "OOC"
 
-	var/msg = "<b>Current Players:</b>\n"
+	var/msg = "<meta charset='UTF-8'><h2><b>Curret Players:</b></h2>\n"
 
 	var/list/Lines = list()
 
+	var/living = 0
+	var/dead = 0
+	var/observers = 0
+	var/lobby = 0
+	var/living_antags = 0
+	var/dead_antags = 0
+
 	if(check_rights(R_INVESTIGATE, 0))
+		log_admin("[key_name(usr)] checked advanced who in-round.")
 		for(var/client/C in GLOB.clients)
 			var/entry = "\t[C.key]"
 			if(!C.mob) //If mob is null, print error and skip rest of info for client.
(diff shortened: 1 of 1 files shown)


code/game/verbs/who.dm
```diff
@@ -1,63 +1,94 @@
-
 /client/verb/who()
 	set name = "Who"
 	set category = "OOC"
 
-	var/msg = "<b>Current Players:</b>\n"
+	var/msg = "<meta charset='UTF-8'><h2><b>Curret Players:</b></h2>\n"
```
Lohikar: Current.

code/game/verbs/who.dm
```diff
@@ -1,63 +1,94 @@
...
 
+	var/living = 0
+	var/dead = 0
+	var/observers = 0
+	var/lobby = 0
+	var/living_antags = 0
+	var/dead_antags = 0
+
 	if(check_rights(R_INVESTIGATE, 0))
+		log_admin("[key_name(usr)] checked advanced who in-round.")
```
Lohikar, line 17: Unlogged change.

MistakeNot4892: Also seems kinda unnecessary, why does the admin team need to know when you use who?

code/game/verbs/who.dm
```diff
@@ -1,63 +1,93 @@
...
+
+	msg += "<b>Total Players: [length(Lines)]</b><br><br>"
 
 	for(var/line in sortList(Lines))
-		msg += "[line]\n"
+		msg += "[line]<br>"
 
-	msg += "<b>Total Players: [length(Lines)]</b>"
-	to_chat(src, msg)
+	show_browser(usr, msg, "window=who;size=780x420;can_close=1")
```
MistakeNot4892, line 90: Please use `/datum/browser` for adding new static windows like this.

MistakeNot4892:
A preference for browser vs. chat window would probably be preferrable.
If it's being dumped to a window I think it could also benefit from a table to make the various fields align properly.

code/game/verbs/who.dm
Lohikar: It looks like it's specifically for admin-who, which does show stuff like antag status.

PR #713, opened 2020-07-21
Converting ghost traps to decls.

MistakeNot4892:
- Ghost traps are now singletons.
- Refactors some of the internal logic and var names, no significant exterior changes.
- Clean up the type tree, adds a pAI stub so the pref shows up in the role list (where did pAI settings go?).

### DIFF
--- a/code/modules/ghosttrap/trap.dm
+++ b/code/modules/ghosttrap/trap.dm
@@ -1,61 +1,40 @@
 // This system is used to grab a ghost from observers with the required preferences 
 // and lack of bans set. See posibrain.dm for an example of how they are called/used.
-
-var/list/ghost_traps
-
-/proc/get_ghost_trap(var/trap_key)
-	if(!ghost_traps)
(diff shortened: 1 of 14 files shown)


code/modules/ghosttrap/trap.dm
```diff
@@ -1,57 +1,40 @@
...
 	var/can_set_own_name = TRUE
 	var/list_as_special_role = TRUE	// If true, this entry will be listed as a special role in the character setup
-
 	var/list/request_timeouts
+	var/check_target_type
+	var/create_mob_if_invalid = FALSE
 
-/datum/ghosttrap/New()
+/decl/ghosttrap/New()
 	request_timeouts = list()
```
afterthought2: Could just do this in the var definition.

afterthought2 (dismissed):

MistakeNot4892:
Tested, all seems to be working.

afterthought2 (approved):

PR #715, opened 2020-07-21
Cherry-pick some flavor fixes

quardbreak:


PsiOmegaDelta (dismissed):

code/game/objects/structures/window.dm
```diff
@@ -7,7 +7,7 @@
 
 	layer = SIDE_WINDOW_LAYER
 	anchored = 1.0
-	atom_flags = ATOM_FLAG_NO_TEMP_CHANGE | ATOM_FLAG_CHECKS_BORDER
```
MistakeNot4892, line 10: This flag only exists for windows as far as I can tell - what is removing it meant to fix?

code/game/turfs/turf.dm
```diff
@@ -269,7 +269,7 @@ var/const/enterloopsanity = 100
 
 /turf/proc/remove_cleanables()
 	for(var/obj/effect/O in src)
-		if(istype(O,/obj/effect/rune) || istype(O,/obj/effect/decal/cleanable) || istype(O,/obj/effect/overlay))
```
MistakeNot4892, line 272: Why remove this?

code/modules/mob/mob.dm
```diff
@@ -461,7 +461,8 @@
 // If usr != src, or if usr == src but the Topic call was not resolved, this is called next.
 /mob/OnTopic(mob/user, href_list, datum/topic_state/state)
 	if(href_list["flavor_more"])
-		show_browser(user, "<HTML><HEAD><TITLE>[name]</TITLE></HEAD><BODY><TT>[replacetext(flavor_text, "\n", "<BR>")]</TT></BODY></HTML>", "window=[name];size=500x200")
+		var/text = "<HTML><HEAD><TITLE>[name]</TITLE></HEAD><BODY><TT>[replacetext(flavor_text, "\n", "<BR>")]</TT></BODY></HTML>"
```
MistakeNot4892, line 464: All this does is add an alloc to the logic, why change it?

code/game/turfs/turf.dm
MistakeNot4892: Nvm.

code/modules/mob/mob.dm
afterthought2: There's a dm-level issue with macro substitutions and string logic. Go figure.

See the bay PR for an extensive discussion.

code/game/objects/structures/window.dm
afterthought2: Seems to be about forbidding movement onto low walls. Probably the correct fix is to set this if `dir` is a cardinal, and unset it for full-tile windows (don't recall if they are `dir = 0` or `dir = 5`).

MistakeNot4892: Tested this locally and it doesn't seem to impact window behavior at all.

MistakeNot4892 (approved):

PR #721, opened 2020-07-23
Charge stick/starting cash refactor.

MistakeNot4892:
- Charge stick world icons now work again. Not sure what's up with the totals but they are showing `000` - don't think it's related to changes to the icon system but will give it a look.
- Vending machines now report if there's a problem with a charge stick.
- Prefs and job equip code now handle starting cash split between credstick, cash and bank account, instead of having it hidden in the depths of loadout. This will likely need to be reverted on Persistence, but the previous UX was very confusing.
- Submap jobs now get appropriate cash amounts as cash in hand.


afterthought2 (approved):
A bit suspect how you can't combine several of the choice datums, but OK.

PR #724, opened 2020-07-24
Ports custom name and description for gear from BoH

quardbreak:
Allow set gear flag for items which can have custom name and description (like medals or coins).


code/modules/client/preference_setup/loadout/gear_tweaks.dm
```diff
@@ -395,4 +395,62 @@
...
+
+/datum/gear_tweak/custom_desc/get_default()
+	return ""
+
+/datum/gear_tweak/custom_desc/get_metadata(var/user, var/metadata)
+	if(valid_custom_desc)
+		return input(user, "Choose an item description.", "Character Preference", metadata) as null|anything in valid_custom_desc
+	return sanitize(input(user, "Choose the item's description. Leave it blank to use the default description.", "Item Description", metadata) as message|null, extra = 0)
+
+/datum/gear_tweak/custom_desc/tweak_item(var/obj/item/I, var/metadata)
```
PsiOmegaDelta: Just noting that no other (correct) implementation (that exists before this PR) of this proc returns a value, and as such both the name/desc overrides can be shortened

MarinaGryphon:
The original feature was ported from Aurora, you may want to check there for fixes.

code/modules/client/preference_setup/loadout/gear_tweaks.dm
```diff
@@ -395,4 +395,62 @@
...
+
+/datum/gear_tweak/custom_desc/get_contents(metadata)
+	return "Description: [metadata]"
+
+/datum/gear_tweak/custom_desc/get_default()
+	return ""
+
+/datum/gear_tweak/custom_desc/get_metadata(user, metadata, title)
+	if(valid_custom_desc)
+		return input(user, "Choose an item description.", "Character Preference", metadata) as null|anything in valid_custom_desc
```
PsiOmegaDelta: `CHARACTER_PREFERENCE_INPUT_TITLE` instead of "Character Preference"
Goes for line 421 as well

MistakeNot4892 (dismissed):
I think we need to revisit the way layout is used generally at some point but this is fine as a feature in itself.

MistakeNot4892:
Will merge when conflict is fixed.

code/modules/client/preference_setup/loadout/gear_tweaks.dm
```diff
@@ -364,3 +364,61 @@
...
+
+/datum/gear_tweak/custom_desc/get_default()
+	return ""
+
+/datum/gear_tweak/custom_desc/get_metadata(user, metadata, title)
+	if(valid_custom_desc)
+		return input(user, "Choose an item description.", CHARACTER_PREFERENCE_INPUT_TITLE, metadata) as null|anything in valid_custom_desc
+	return sanitize(input(user, "Choose the item's description. Leave it blank to use the default description.", "Item Description", metadata) as message|null, extra = 0)
+
+/datum/gear_tweak/custom_desc/tweak_item(obj/item/I, metadata)
```
PsiOmegaDelta, line 420:
```
/datum/gear_tweak/custom_desc/tweak_item(obj/item/I, metadata)
	if(metadata)
		 I.desc = metadata
```

code/modules/client/preference_setup/loadout/gear_tweaks.dm
```diff
@@ -364,3 +364,61 @@
...
+
+/datum/gear_tweak/custom_name/get_metadata(user, metadata, title)
+	if(valid_custom_names)
+		return input(user, "Choose an item name.", CHARACTER_PREFERENCE_INPUT_TITLE, metadata) as null|anything in valid_custom_names
+	return sanitize(input(user, "Choose the item's name. Leave it blank to use the default name.", "Item Name", metadata) as text|null, MAX_LNAME_LEN, extra = 0)
+
+/datum/gear_tweak/custom_name/tweak_item(obj/item/I, metadata)
+	if(!metadata)
+		return I.name
+	I.name = metadata
```
PsiOmegaDelta:
```
/datum/gear_tweak/custom_name/tweak_item(obj/item/I, metadata)
	if(metadata)
		I.name = metadata
```

code/modules/client/preference_setup/loadout/gear_tweaks.dm
PsiOmegaDelta: The original comment is only 4 days old

MistakeNot4892: The code your comment is referring to doesn't return a value, so I am not sure if it is still relevant.

MistakeNot4892 (dismissed):

code/modules/client/preference_setup/loadout/gear_tweaks.dm
```diff
@@ -364,3 +364,59 @@
...
+/datum/gear_tweak/custom_name/get_contents(metadata)
+	return "Name: [metadata]"
+
+/datum/gear_tweak/custom_name/get_default()
+	return ""
+
+/datum/gear_tweak/custom_name/get_metadata(user, metadata, title)
+	if(valid_custom_names)
+		return input(user, "Choose an item name.", CHARACTER_PREFERENCE_INPUT_TITLE, metadata) as null|anything in valid_custom_names
+	return sanitize(input(user, "Choose the item's name. Leave it blank to use the default name.", "Item Name", metadata) as text|null, MAX_LNAME_LEN, extra = 0)
```
MistakeNot4892: `extra` arg just strips newlines and tabs, why are you manually setting it here? Descs and names don't need tabs or newlines.

MistakeNot4892 (approved):

PR #725, opened 2020-07-25
Fixes spam when dispensing items from vending machines

quardbreak:


MistakeNot4892:
What spam is this meant to fix? Should already be handled by click cooldown.

PR #729, opened 2020-07-26
Slight do_attack_animation refactor, allowed it to be supplied an item to use.

MistakeNot4892:
Splitting it out of the hands PR, should be pretty self explanatory.
I tried using a mutable appearance for this but it resulted in the user turning into a wrench. Advice on that would be appreciated @lohikar.

### DIFF
--- a/code/modules/mob/animations.dm
+++ b/code/modules/mob/animations.dm
@@ -119,7 +119,7 @@ note dizziness decrements automatically in the mob's Life() proc.
 	//reset the pixel offsets to zero
 	is_floating = 0
 
-/atom/movable/proc/do_attack_animation(atom/A)
+/atom/movable/proc/do_attack_animation(atom/A, atom/movable/weapon)
 
 	var/pixel_x_diff = 0
 	var/pixel_y_diff = 0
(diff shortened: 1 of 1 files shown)


comma (approved):

code/modules/mob/animations.dm
```diff
@@ -156,18 +156,26 @@ note dizziness decrements automatically in the mob's Life() proc.
...
-		return
+	// What are we attacking with?
+	if(!weapon)
+		weapon = get_active_hand()
+		if(!weapon)
+			return
+
+	// Create an image to show to viewers.
+	// Reset plane and layer so that it doesn't inherit the UI settings from equipped items.
+	var/image/I = new(loc = A)
```
Lohikar, line 171: What code were you using w/ MAs when you tried them?

MistakeNot4892: I don't recall exactly, I think it was more or less the same but `var/mutable_appearance/MA = new(A)`.

Lohikar: Should be:
```dm
var/mutable_appearance/MA = new(weapon)
MA.loc = A
```
(with the `appearance =` line deleted)

PR #732, opened 2020-07-26
Adds throwing sound and animation

quardbreak:
Adds throw animation. which was inspired by the goonstation throw animation.
Also adds a throw sound which was taken from tgstation's throwing bola and shortened.


MistakeNot4892 (dismissed):
Looks good. As discussed we're pretty sure this is sufficient to not be cribbing from the Goon worksheet so if the original author of the Goon version spots this and has any issue, please get in touch and it can be removed/redone.

MistakeNot4892 (approved):

PR #736, opened 2020-07-27
Refactored client view scaling.

MistakeNot4892:
Default icon size is now saved/loaded by prefs since the skin settings are unreliable.

I tried winsetting `mainwindow.on-size` to fix the 'invalid verb' issue with `OnResize` but no dice.


afterthought2 (approved):
Let's try it.

Azarak:
bless!

PR #738, opened 2020-07-27
Centralizes/removes extraneous fullscreen/chat position update calls.

MistakeNot4892:
This should reduce/remove the weird window flicker from ghosting.

### DIFF
--- a/code/modules/client/client_procs.dm
+++ b/code/modules/client/client_procs.dm
@@ -223,6 +223,9 @@
 	if(holder)
 		src.control_freak = 0 //Devs need 0 for profiler access
 
+	if(!istype(mob, world.mob))
+		prefs?.apply_post_login_preferences()
+
 	//////////////
 	//DISCONNECT//
 	//////////////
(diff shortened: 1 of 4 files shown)


code/modules/client/client_procs.dm
```diff
@@ -223,6 +223,13 @@
 	if(holder)
 		src.control_freak = 0 //Devs need 0 for profiler access
 
+	// Update our windows and chat settings.
+	if(get_preference_value(/datum/client_preference/chat_position) == GLOB.PREF_YES)
+		update_chat_position(TRUE)
+
+	if(get_preference_value(/datum/client_preference/fullscreen_mode) != GLOB.PREF_OFF)
+		toggle_fullscreen(get_preference_value(/datum/client_preference/fullscreen_mode))
```
afterthought2: It's unsafe to do it this early, as clients may be loaded before prefs are set up. The result will be that these will runtime and the toggles won't fire.

afterthought2 (approved):

PR #740, opened 2020-07-27
Removes the numerical inv slot defines, swaps them for strings.

MistakeNot4892:
Swaps the numerical slot flags for the string slot flags. They were redundant and this will make the inventory rewrite easier (and less of a huge diff).
This is largely find and replace other than spots that now refer to `global.all_inventory_slots`, and to places that refer to `slowdown_per_slot`, which is now lazylisted.

I have tested this a fair bit locally, and the slots were not used much in isolation, so this seems to work fine.


afterthought2 (approved):
Only danger is icons, but those wouldn't be using the numerical defines anyway.

MistakeNot4892:
Icons already use the string slots yeah, they all seem to work fine currently.

PR #749, opened 2020-07-28
Increases atom/movable glide_size to 8 from 4, to increase gameplay smoothiness.

Azarak:
When a character walk he won't finish his glide animation before making another step "snapping" their sprite and the users camera several pixels away, for purposes of starting the next glide, which causes a feeling of lag, and that's not good

This bumps the glide_size to 8, which considerably helps the smoothness of stuff


MistakeNot4892 (approved):
Prior to getting automatic glide size working, we should probably just outright remove any glide_size, but I don't object to this in the interim.

PR #754, opened 2020-07-28
Added maploader prefixes to modpack away sites, fixes ion engines.

MistakeNot4892:
Still need to test this locally, but should at least make Travis scream about missing paths.


code/modules/maps/reader.dm
```diff
@@ -35,8 +35,7 @@ GLOBAL_DATUM_INIT(_preloader, /dmm_suite/preloader, new)
  */
 /dmm_suite/load_map(var/dmm_file, var/x_offset, var/y_offset, var/z_offset, var/cropMap, var/measureOnly, var/no_changeturf, var/clear_contents, var/lower_crop_x, var/lower_crop_y, var/upper_crop_x, var/upper_crop_y, var/initialized_areas_by_type)
 	//How I wish for RAII
-	Master.StartLoadingMap()
```
Lohikar: DMMS can be used outside of map templates, moving this function outside of here is probably not safe.

afterthought2: Was an experiment to see if halting the icon update SS during mapload resolved some init order bugs (placing this in `load_map` means everything unhalts during the atom Initialize step later, which checks tick). Unsure why it caused the infinite looping in Travis, though.

MistakeNot4892: I've reverted that commit.

afterthought2 (approved):

deathride58 (approved):

PR #764, opened 2020-07-31
Adds a helper for \icon[foo].

MistakeNot4892:
This has to be done for any JS-driven UI, whether Goonchat or VueUI. Might as well do it now and get it out of the way.

### DIFF
--- a/code/controllers/subsystems/zcopy.dm
+++ b/code/controllers/subsystems/zcopy.dm
@@ -316,14 +316,14 @@ SUBSYSTEM_DEF(zcopy)
 		if (istype(A, /atom/movable/openspace/overlay))
 			var/atom/movable/openspace/overlay/OO = A
 			var/atom/movable/AA = OO.associated_atom
-			out += "<li>\icon[A] plane [A.plane], layer [A.layer], depth [OO.depth], associated Z-level [AA.z] - [OO.type] copying [AA] ([AA.type], eventually [OO.mimiced_type])</li>"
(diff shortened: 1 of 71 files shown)


comma (dismissed):

afterthought2:
You tested that this works? There are sometimes dm issues with stuff like this.

code/controllers/subsystems/zcopy.dm
```diff
@@ -316,14 +316,14 @@ SUBSYSTEM_DEF(zcopy)
 		if (istype(A, /atom/movable/openspace/overlay))
 			var/atom/movable/openspace/overlay/OO = A
 			var/atom/movable/AA = OO.associated_atom
-			out += "<li>\icon[A] plane [A.plane], layer [A.layer], depth [OO.depth], associated Z-level [AA.z] - [OO.type] copying [AA] ([AA.type], eventually [OO.mimiced_type])</li>"
+			out += "<li>[CHAT_ICON(A)] plane [A.plane], layer [A.layer], depth [OO.depth], associated Z-level [AA.z] - [OO.type] copying [AA] ([AA.type], eventually [OO.mimiced_type])</li>"
```
Lohikar: These aren't going to chat.

MistakeNot4892: I can find and replace it to HTML_ICON if that would make you happy. I don't think it really matters.

Lohikar: That would be fine; main thing is that its usage matches its name.

MistakeNot4892:
image

afterthought2 (approved):

PR #761, opened 2020-07-30
Makes exoplanets unsimulated.

MistakeNot4892:
I have no idea if this actually works, will test it properly in the morning. Tl;dr, exoplanets are made from `/turf/exterior`, which uses a fresh copy of a global atmosphere (or one set by the exoplanet) for `return_air`. It essentially acts like an infinite sink/source when bordering simulated zones. Upshot is that exterior exoplanet turfs do zero processing, same as space turfs. I think the only real mechanical difference will be that you will need to lay down floor tiles before you can lay cables.

Also makes mining resources into an extension.

**TODO:**
- [x] Test exoplanet spawn.
- [x] Test atmosphere interaction.
- [x] Test building interactions.
- [x] Readd burned/broken states.
- [x] Readd turf edge blending.
- [x] Readd randomly selected edge turfs.
- [x] Do the bitflag thing @comma requested.
- [x] Remove broken/burned overlays from most turfs (permafrost etc).

PR #772, opened 2020-08-03
Turf resources are now an extension.

MistakeNot4892:
Splitting this out of #761.

### DIFF
--- a/code/game/objects/items/devices/scanners/mining.dm
+++ b/code/game/objects/items/devices/scanners/mining.dm
@@ -1,5 +1,8 @@
-/turf/simulated
-	var/surveyed
+/datum/extension/buried_resources
+	base_type = /datum/extension/buried_resources
+	expected_type = /turf
+	var/list/resources
+	var/surveyed = FALSE
 
 /obj/item/scanner/mining
 	name = "ore detector"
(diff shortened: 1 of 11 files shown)


MistakeNot4892:
Ready to go.

code/game/objects/items/devices/scanners/mining.dm
```diff
@@ -1,5 +1,9 @@
-/turf/simulated
-	var/surveyed
+/datum/extension/buried_resources
+	base_type = /datum/extension/buried_resources
+	expected_type = /turf
+	flags = EXTENSION_FLAG_IMMEDIATE
```
comma: Does it really need immediate flag? It's put on a lot of places, and doesn't really do anything on its own, just set/read by others. Might shave few KB if you leave it lazy

comma (approved):

code/modules/overmap/exoplanets/turfs.dm
```diff
@@ -146,14 +145,16 @@
...
-	if(!resources)
-		resources = list()
+
+	var/datum/extension/buried_resources/resources = get_or_create_extension(src, /datum/extension/buried_resources)
 	if(prob(70))
-		resources[/decl/material/solid/mineral/graphite] = rand(3,5)
+		LAZYSET(resources.resources, /decl/material/solid/mineral/graphite, rand(3,5))
 	if(prob(5))
-		resources[/decl/material/solid/metal/uranium] = rand(1,3)
+		LAZYSET(resources.resources, /decl/material/solid/metal/uranium, rand(1,3))
```
PsiOmegaDelta, line 153: Responsibilities are a bit vague here and implementation details are leaking, with a user of the extension having to know that the `resources` list is a lazy-list
Preferable would probably be to have an `/AddResource(type, amount)` proc on the extension (and also mark the `resources` list as `VAR_PRIVATE`/`VAR_PROTECTED` as necessary)

PsiOmegaDelta: Similarly there should perhaps be a `/RemoveResource(type, amount)`-proc which returns the amount actually acquired and, if all resources have been depleted, makes the extension remove itself from the host.

MistakeNot4892: Proc overhead for this would increase init time significantly, ore maps are generated repeatedly in worldgen.

PR #774, opened 2020-08-04
Adds default client pixel offset tracking and has reset_view delay until screen shake completes.

MistakeNot4892:


comma (dismissed):

code/modules/mob/mob.dm
```diff
@@ -279,22 +279,25 @@
...
-				client.eye = client.mob
-				client.perspective = MOB_PERSPECTIVE
-			else
-				client.perspective = EYE_PERSPECTIVE
-				client.eye = loc
-	return
-
+	set waitfor = 0
+	while(shakecamera && client && !QDELETED(src))
+		sleep(1)
```
Lohikar, line 284: Probably should be `stoplag(1)` instead of sleep.

Lohikar: On second thought, these're all nullchecks - no need.

MistakeNot4892: As discussed on Discord, would prefer to leave it as `sleep(1)` for clarity.

MistakeNot4892:
Please review this.

comma (approved):

PR #776, opened 2020-08-04
Swapped to a browser based chat instead of an output panel based chat.

MistakeNot4892:
Highly WIP, figured I'd open it for input.

Swaps the `output` to a `browser`, with accompanying code to support icons and newlines being sent to the client. Also bundles a small JS feature to collapse identical lines of text into each other. This is similar to Goonchat/Vuechat but relies on no other libraries and has like 1% the features.

I copypasted `icon2html()` and accompanying procs from Snapshot/quardbreak's Goonchat port on Bay.

image

**TODO:**
- [x] Basic conversion.
- [ ] Fix up CSS so text isn't so huge.
- [ ] Add a darkmode and a way to specify CSS in the client prefs.
- [ ] Check if there's some way to automatically update the broken-link icons in the chat when an asset is sent but before the next message arrives.
- [ ] Find a less inefficient way of appending to the buffer than `innerHTML +=`
- [ ] Add a pause button.
- [ ] Add a ping indicator.

### DIFF
--- a/code/_chat.dm
+++ b/code/_chat.dm
@@ -1,0 +1,110 @@
+/proc/generate_asset_name(file)
+	return "asset.[md5(fcopy_rsc(file))]"
+	
+/proc/icon2html(thing, target, icon_state, dir, frame = 1, moving = FALSE, realsize = FALSE, class = null)
+	if (!thing)
+		return
+
+	var/key
+	var/icon/I = thing
+	if (!target)
+		return
+	if (target == world)
+		target = GLOB.clients
+
+	var/list/targets
+	if (!islist(target))
+		targets = list(target)
+	else
+		targets = target
+		if (!targets.len)
+			return
+	if (!isicon(I))
+		if (isfile(thing)) //special snowflake
+			var/name = "[generate_asset_name(thing)].png"
+			register_asset(name, thing)
+			for (var/thing2 in targets)
+				send_asset(thing2, key, FALSE)
+			return "<img class='icon icon-misc [class]' src=\"[url_encode(name)]\">"
+		var/atom/A = thing
+		if (isnull(dir))
+			dir = A.dir
+		if (isnull(icon_state))
+			icon_state = A.icon_state
+		I = A.icon
+		if(istype(thing, /mob/living/carbon/human)) // Shitty workaround for a BYOND issue.
+			var/icon/temp = I
+			I = icon()
+			I.Insert(temp, dir = SOUTH)
+			dir = SOUTH
+	else
+		if (isnull(dir))
+			dir = SOUTH
+		if (isnull(icon_state))
+			icon_state = ""
+
+	I = icon(I, icon_state, dir, frame, moving)
+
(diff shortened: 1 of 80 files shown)


Spookerton:
You might consider parsing messages with the document parser api, storing them in a document fragment, and using appendChild periodically to flush the fragment into the output. This will allow you to fail fast in an informed way on malformed payloads and only causes a partial repaint of the document, just once per batch.

ed: wrapping things in `<div>` may well also be preferable to `<br>` prefixing. Not sure about the performance implications of that, but it would help contain spills.

<https://caniuse.com/#search=DOMParser>
<https://caniuse.com/#search=DocumentFragment>
<https://caniuse.com/#search=appendChild>

It's not a sane task by scale but as a musing - sending messages as json and constructing them on the client would allow you to recreate payloads more quickly and securely by disallowing script tags and arbitrary attributes and by using textContent to add explicitly non-parsed element bodies.

<https://caniuse.com/#search=textContent>

midnight notepad napkin:
```js
var output = document.getElementById('chatOutput')
var fragment = new DocumentFragment()
var parser = new DOMParser()

setInterval(function() {
  'use strict';
  if (!fragment.children.length)
    return
  output.appendChild(fragment)
}, 5e2)

function append(msg) {
  'use strict';
  try {
    msg = parser.parseFromString('<br>' + msg, 'text/html')
    if (msg.getElementsByTagName('parsererror').length) {
      //logging maybe
      return
    }
    fragment.appendChild(msg)
  }
  catch (error) {
    //logging maybe
    return
  }
}

//--

function isObject(value) {
  return Object.prototype.toString.call(value) === '[object Object]'
}

function appendButItsJson(msg) {
  'use strict';
  try {
    msg = JSON.parse(msg)
  }
  catch (error) {
    return
  }
  if (!Array.isArray(msg) || !msg.length)
    return
  var line = document.createElement('div')
  for (var i = 0, n = msg.length; i < n; ++i) {
    var entry = msg[i]
    if (!isObject(entry))
      continue
    if (!entry.text)
      continue
    if (!entry.tag)
      entry.tag = 'span'
    else if (entry.tag === 'script')
      continue
    var chunk = document.createElement(tag)
    if (!chunk)
      continue
    chunk.textContent = entry.text
    if (entry.className)
      chunk.className = entry.className
    if (isObject(entry.style)) {
      for (var key in entry.style) {
        if (!entry.style.hasOwnProperty(key))
          continue
        chunk.style[key] = entry.style[key]
      }
    }
    line.appendChild(chunk)
  }
  if (!line.children.length)
    return
  fragment.appendChild(line)
}
```


pretty json for appendButItsJson
```json
[
  {
    "text": "Hello ",
    "className": "some-display-class"
  },
  {
    "text": "World",
    "style": {
      "color": "#f00"
    }
  }
]
```

MistakeNot4892:
Thanks @Spookerton, I will give that a proper read when I get back to this PR. I was thinking of pinging you at some point but didn't want to put more on your plate.

MistakeNot4892:

PR #778, opened 2020-08-04
Planets generate landing areas BEFORE placing ruins

comma:
Prevents the often seen case of auto-clearing landmarks chomping out most of the walls of a ruin.
Now autoclear landmarks set their affected turfs to NO RUINS

### DIFF
--- a/code/modules/overmap/exoplanets/_exoplanet.dm
+++ b/code/modules/overmap/exoplanets/_exoplanet.dm
@@ -22,6 +22,8 @@
 	var/y_size
 
 	var/landmark_type = /obj/effect/shuttle_landmark/automatic
+	var/shuttle_size = 20  		 //'diameter' of expected shuttle in turfs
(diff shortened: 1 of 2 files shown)


MistakeNot4892 (dismissed):

afterthought2:
The problem with this is that they don't generate clearing landmarks necessarily; they first try pretty hard to place a non-clearing one. But that one won't get the flag set.

I'd suggest either having the flag set in /automatic or adding another subtype.

comma:
Oh drat I assumed it was always a clearing one. Will adjust to always mark area.

comma:
Okay now always using clearing landmarks - they have no effect aside from territory marking if placed where old ones would've been.
Also number of landmarks is decided now dynamically with an arbitrary formula based on planet side (capped at 3)

MistakeNot4892 (approved):

PR #781, opened 2020-08-05
Backporting stoves/hearths/campfires from Europa.

MistakeNot4892:
Backport of https://github.com/Yonaguni/EuropaStation/pull/324 with repeated thanks to @TheEliminator16 for the campfire icons and @PlayerDeer for the stove (I think, if I am naming the wrong person please correct me).

image

Adds some basic fire sources like campfires and stoves that can be built with wood, lit with fire sources, and fed with wood, coal or flammable reagents (or extinguished with water).


code/game/objects/structures/fires.dm
```diff
@@ -0,0 +1,325 @@
...
+	var/lit = 0
+
+/obj/structure/fire_source/Initialize()
+	. = ..()
+	if(lit == FIRE_LIT && fuel > 0)
+		light(TRUE)
+	update_icon()
+	steam = new(name)
+	steam.attach(get_turf(src))
+	steam.set_up(3, 0, get_turf(src))
```
afterthought2, line 42: Need to clean `steam` in `Destroy`.

code/game/objects/structures/fires.dm
```diff
@@ -0,0 +1,325 @@
...
+		if(R.fuel_value <= 0)
+			do_steam = TRUE
+		fuel += REAGENT_VOLUME(RG, rtype) * R.fuel_value
+	RG.clear_reagents()
+	fuel = max(0, fuel)
+	if(lit == FIRE_LIT)
+		if(fuel <= 0)
+			die()
+		if(do_steam)
+			steam.start() // HISSSSSS!
```
afterthought2, line 254: Does this work OK if called more than once? Does it return steam to original state later? Does it ever try to `qdel` steam, which will fail?

I know there are often deletion problems with these effects because of how they are implemented.

code/game/objects/structures/fires.dm
```diff
@@ -0,0 +1,325 @@
...
+
+/obj/structure/fire_source/Destroy()
+	STOP_PROCESSING(SSobj, src)
+	return ..()
+
+/obj/structure/fire_source/Process()
+
+	if(!check_atmos())
+		return
```
afterthought2, line 265: Shouldn't this burn it out or something?

MistakeNot4892: Ah true it should put it out.

code/game/objects/structures/fires.dm
MistakeNot4892: Steam overrides start() and just sets up a bunch of callbacks to a proc that creates an effect and moves it around, looks fine to call repeatedly to me.

code/game/objects/structures/fires.dm
MistakeNot4892: `die()` added.

afterthought2 (approved):

PR #787, opened 2020-08-08
Overmap Solar Arrays

Mazianni:
This PR aims to make solars (on overmap enabled maps that is) track a new overmap object - a star. Stars are roughly scientifically accurate.

Solars will always try to track the system's star on the overmap - the farther from the star they are, the less power they'll generate. Some stars give a bonus to luminosity, which helps offset that distance penalty a decent amount.

I'd like to add events around stars and such, but right now overmap objects don't actually support proper movement, and I'm unsure how to code the two events I would like to add.

There is a catch. Since the ship's fore facing direction is *not* dynamic, and Solars will always try to track the sun, this will lead to strange behavior, even if you're say, west of the sun, and flying towards it: The solars will point to the west.


afterthought2:
Haven't read through it, but in re: the desc: ships do have a `fore_dir` var they are expected to set to indicate which direction they are pointing on the z level; that and/or `get_heading` may help.

Mazianni:
The issue is mainly a matter of perspective.

Mazianni:
ah fuck I forgot travis isn't 513 yet.

Mazianni:
I have no idea what that travis error is about.

code/modules/power/tracker.dm
```diff
@@ -19,6 +19,10 @@
 	. = ..(mapload)
 	Make(S)
 	connect_to_network()
+	if(GLOB.using_map.using_sun)
+		if(!istype(src, /obj/machinery/power/tracker/overmap))
+			new /obj/machinery/power/tracker/overmap(get_turf(src))
+			qdel(src)
```
PsiOmegaDelta, line 25: Do not delete self in `/Initialize()`, instead `return INITIALIZE_HINT_QDEL`

PsiOmegaDelta (changes requested):

code/modules/overmap/sectors.dm
```diff
@@ -132,6 +135,22 @@
...
+		var/sun_y
+
+		if(dimension & 1 == 0)
+			sun_x = (GLOB.using_map.overmap_size / 2) + 0.5
+			sun_y = (GLOB.using_map.overmap_size / 2) + 0.5 //add 0.5 to get the final x/y of the star's location
+		else
+			sun_x = (GLOB.using_map.overmap_size / 2)
+			sun_y = (GLOB.using_map.overmap_size / 2) //This is a rough 'center' because even maps don't have a true center.
+
+		var/turf/sun_turf = locate(sun_x, sun_y, GLOB.using_map.overmap_z)
```
MistakeNot4892, line 151: As discussed, could be:
````
var/c = Ceil(GLOB.using_map.overmap_size / 2)
var/turf/sun_turf = locate(c, c, GLOB.using_map.overmap_z)
````

File: code/modules/power/overmap_solar.dm [hunk 3]
```diff
@@ -0,0 +1,206 @@
...
+		return
+	//find the smaller angle between the direction the panel is facing and the direction of the sun (the sign is not important here)
+	var/p_angle = min(abs(adir - overmap_control.overmap_tracker.get_sun_angle()), 360 - abs(adir - overmap_control.overmap_tracker.get_sun_angle()))
+
+	if(p_angle > 90)			// if facing more than 90deg from sun, zero output
+		sunfrac = 0
+
+		return
+	//Do distance checks from the star, keep sunfrac at a min of 0 and a max of 1.
+	var/obj/effect/overmap/star/star = locate(/obj/effect/overmap/star)
```
MistakeNot4892, line 28: This might cause problems if we ever have a multi-overmap setup - also quite slow compared to keeping a list of stars and checking if they're on your overmap z-level.

code/modules/power/overmap_solar.dm
```diff
@@ -0,0 +1,206 @@
...
+	var/dist = get_dist(star, overmap_control.linked)
+	var/dist_penalty = dist * 0.1
+
+	sunfrac = cos(p_angle) ** 2
+	sunfrac -= (dist_penalty -= star.star_luminosity)
```
MistakeNot4892, line 33: Please do these operations on separate lines.

code/modules/power/overmap_solar.dm
```diff
@@ -0,0 +1,206 @@
...
+	//Opacity check.
+	var/list/turflist = getline(overmap_control.linked, star)
+	for(var/turf/T in turflist)
+		var/obj/effect/overmap/overmap_effect = locate(/obj/effect/overmap) in T
+		if(overmap_effect && overmap_effect.opacity)
+			sunfrac = 0
+			break
+	if(!star.star_luminosity)
```
MistakeNot4892, line 41: Could you check this before iterating on the turf list above?

code/modules/power/overmap_solar.dm
```diff
@@ -0,0 +1,206 @@
...
+		overmap_control.connected_panels.Remove(src)
+	overmap_control = null
+
+/obj/machinery/power/solar/overmap/Process()
+	if(stat & BROKEN)
+		return
+	if(!overmap_control) //if there's no sun or the panel is not linked to a solar control computer, no need to proceed
+		return
+
+	if(powernet)
```
MistakeNot4892, line 57: Should you also unset control if there's no powernet, or is the assumption that you don't need to unset if the powernet is already clear?

code/modules/power/overmap_solar.dm
```diff
@@ -0,0 +1,206 @@
...
+	if(SC && (get_dist(src, SC) > SOLAR_MAX_DIST))
+		return 0
+	overmap_control = SC
+	return 1
+
+//set the control of the panel to null and removes it from the control list of the previous control computer if needed
+/obj/machinery/power/solar/overmap/unset_control()
+	if(overmap_control)
+		overmap_control.connected_panels.Remove(src)
+	overmap_control = null
```
MistakeNot4892, line 78: Nit: you check if it's false and then null it regardless, could indent this one step.

code/modules/power/overmap_solar.dm
```diff
@@ -0,0 +1,206 @@
...
+
+/obj/machinery/power/tracker/overmap
+	var/obj/machinery/power/solar_control/overmap/overmap_control
+
+/obj/machinery/power/tracker/overmap/proc/get_sun_angle()
+	if(control)
+		var/obj/machinery/power/solar_control/overmap/SC = control
+		if(SC.linked)
+			var/ship = SC.linked
+			var/obj/effect/overmap/star/sun = locate()
```
MistakeNot4892, line 91: Sun may not exist if overmap is not configured to use it, would this runtime in that case?
Also same issue as the comment above regarding using a star list and checking z-level.

code/modules/power/overmap_solar.dm
```diff
@@ -0,0 +1,206 @@
...
+		linked = sector
+		return 1
+
+/obj/machinery/power/solar_control/overmap/Destroy()
+	linked = null
+	for(var/obj/machinery/power/solar/overmap/M in connected_panels)
+		M.unset_control()
+	if(overmap_tracker)
+		overmap_tracker.unset_control()
+	..()
```
MistakeNot4892, line 119: `. = ..()`

code/modules/power/solar.dm
```diff
@@ -90,12 +103,16 @@ var/list/solars_list = list()
...
-		overlays += image('icons/obj/power.dmi', icon_state = "solar_panel", layer = ABOVE_HUMAN_LAYER)
-		src.set_dir(angle2dir(adir))
+	if(!isnull(panel))
+		if(stat & BROKEN)
+			panel.icon_state = "solar_panel-b"
+		else
+			panel.icon_state = "solar_panel"
+
+		var/matrix/rot_matrix = matrix()
+		rot_matrix.Turn(adir)
```
MistakeNot4892, line 113: For reference, after the convo on Discord, it should be possible to calc the effective direction of the panels based on the overmap ship, sun direction and ship fore direction via:
````
arctan(sun) + (foredir - mapdir)
````
(thanks @NataKilar)

File: code/modules/overmap/star/star.dm [hunk 11]
```diff
@@ -0,0 +1,124 @@
...
+	star_luminosity = class.luminosity + startype.luminosity_bonus
+	filters += filter(type="blur", size = 1)
+
+/decl/star_class
+	var/star_class //What class the star is - this is used for the name.
+	var/color //What color the star is, this is used to color it's light source and the star itself.
+	var/luminosity = 0 //We use this to adjust the amount of power solar panels generate, plus some other maths.
+	var/list/possible_types = list() //Some classes of star can't be certain types, so we define what they can be in a list that's picked from.
+
+/decl/star_class/K
```
MistakeNot4892, line 32: Nit: not a big fan of uppercase lettering in type paths.

nebula.dme
```diff
@@ -3194,7 +3197,6 @@
 #include "maps\antag_spawn\wizard\wizard.dm"
 #include "maps\away\away_sites.dm"
 #include "maps\away_sites_testing\away_sites_testing_define.dm"
-#include "maps\example\example_define.dm"
```
MistakeNot4892, line 3197: Put it back!

MistakeNot4892 (changes requested):
Only other comment is that some of your changes use a ton of unnecessary newlines which look a bit messy.

MistakeNot4892:
Feel free to reopen if you get to this, I might adopt it after tcomms is done in 200 years if you don't get time/brainpower for it.

PR #788, opened 2020-08-08
Shuttle Landing Adjustments

NataKilar:
Makes some adjustments to shuttle landing and turf translation for ease of use and code cleanliness.
* Manual Shuttle Landing now only requires expert rather than professional skill in piloting.
* Turf translation now keeps track of the previous turf type, so shuttles will now leave behind whatever they landed on rather than relying on the landmarks base_turf alone.
* Manual landing now allows landing across multiple types of turf, provided they are not dense.
* Increased the view range of shuttle landing to 25.
* Cleans up the check_landing proc as suggested by Loaf when this feature was first implemented.

### DIFF
--- a/code/_helpers/turfs.dm
+++ b/code/_helpers/turfs.dm
@@ -135,13 +135,15 @@
 				transport_turf_contents(source, target)
 	//change the old turfs
 	for(var/turf/source in translation)
-		source.ChangeTurf(base_turf ? base_turf : get_base_turf_by_area(source), 1, 1)
+		var/old_turf = source.prev_type || base_turf || get_base_turf_by_area(source)
+		source.ChangeTurf(old_turf)
 
 //Transports a turf from a source turf to a target turf, moving all of the turf's contents and making the target a copy of the source.
 /proc/transport_turf_contents(turf/source, turf/target)
-
+	var/target_type = target.type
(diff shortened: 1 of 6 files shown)


afterthought2:
Not convinced `ChangeTurf` is where you want to do this. Imagine: shuttle arrives in space; previous_type now space. Shuttle is bombed and the turf is changed from wall to floor or space or whatever; previous_type is now wall. Shuttle leaves; turf reverts back to wall.

NataKilar:
In the scenario you gave, prev_type would actually remain as space; ChangeTurf() sets the new turf's prev_type to the prev_type of the former turf, not the type of the former turf itself. The only place prev_type should actually be set is in transport_turf_contents(), unless I've made a mistake somewhere.

That being said, I don't think this is too big of a concern. It would cause some weirdness if, for instance, a shuttle landed on plating that is then bombed to space, and upon leaving it would turn back to plating. I think this is still preferable however, as removing the prev_type inheritance on ChangeTurf() could lead to issues on takeoff if a shuttle's wall is removed or a new one constructed in the shuttle area.

afterthought2:
Ah, I misunderstood the `ChangeTurf` logic there.

afterthought2 (approved):

PR #793, opened 2020-08-13
Expanding inventory slot system to handle other slots.

MistakeNot4892:
Phase 2 of the inventory rewrite. After this ideally all that is left will be converting the other slots to the datum system and removing the outdated code.

**TODO:**
- [x] Work out why only lighting overlays are showing for hats.
- [x] Cause inventory slots to create/update HUD elements instead of relying on species HUD info.
- [x] Fix up inventory showing/hiding for inv icons.
- [x] Work out why the heck you can't equip hats. :(
- [x] Make inventory slots respect the toggle button (bottom left) when added.
- [x] Readd species-level UI loc for inventory slots.
- [x] Fix `wearing_wiz_garb()`.
- [x] Convert remaining single-bodypart inventory slots (shoes, gloves and storage can be done another time).
- [x] Debug.


MistakeNot4892:
This is ready for review.

MistakeNot4892:
Need to test/fix species setting updating inventory slots.

MistakeNot4892:

PR #798, opened 2020-08-15
Misc fixes.

MistakeNot4892:
Fixes for a bunch of runtimes observed during a recent round of Scav.


code/game/objects/items/item_icon_experimental.dm
```diff
@@ -6,6 +6,8 @@
 // Cached cause asking icons is expensive
 var/list/icon_state_cache = list()
 /proc/check_state_in_icon(var/checkstate, var/checkicon)
+	if(!checkstate || !istext(checkstate) || !isicon(checkicon))
```
afterthought2, line 9: Check the logic: this is always true.

afterthought2 (approved):

PR #804, opened 2020-08-16
Adds self-density-checking to /atom/movable/CanPass.

### LINKED ISSUE #148: disposals ejection chute glitch
#### Description of issue
disposals ejection chute works when a mob enters it or if an item is thrown in, but objects entering the disposals ejection chute by way of conveyor belt simply sit on the tile the chute is on without going in.  

Until recently, conveyor belts could work properly with disposals ejection chute.  

#### Steps to reproduce
use conveyor belt to push trash into the disposals chute

Client Version: 512
Server Revision: e81a0d8af562878154e6d5b52ce23c64479d15c0 - brickwork - 2020-03-15
Game ID: b6b-bfGF

#### Issue bingo

MistakeNot4892:
This is failing unit tests and I am not sure why.
Fixes #148.

````
## UNIT_TEST ##: !!! FAILURE !!! [FOUNDATION: step() shall return true on success]: step() did not return true: Mob result: 1 - Obj result: 0.
````
I don't know why the CanPass change is causing this, if anyone can advise I'd appreciate it.

### DIFF
--- a/code/game/atoms_movable.dm
+++ b/code/game/atoms_movable.dm
@@ -329,4 +329,9 @@
 	. = ..()
 
 /atom/movable/CanPass(atom/movable/mover, turf/target, height=1.5, air_group = 0)
-	. = ..() || (mover && !(mover.movable_flags & MOVABLE_FLAG_NONDENSE_COLLISION) && !mover.density)
+	. = ..()
+	if(. && mover)
+		if(!density && (mover.movable_flags & MOVABLE_FLAG_NONDENSE_COLLISION))
+			return FALSE
+		if(!mover.density && (movable_flags & MOVABLE_FLAG_NONDENSE_COLLISION))
+			return FALSE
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: code/game/atoms_movable.dm [hunk 1]
```diff
@@ -329,4 +329,9 @@
 	. = ..()
 
 /atom/movable/CanPass(atom/movable/mover, turf/target, height=1.5, air_group = 0)
-	. = ..() || (mover && !(mover.movable_flags & MOVABLE_FLAG_NONDENSE_COLLISION) && !mover.density)
+	. = ..()
+	if(. && mover)
```
afterthought2, line 333: This does actually change the logic: it used to be that if `..()` is true, we return true and continue on; the condition is an alternative success condition. The new logic is that there are additional fail conditions. In particular, in the old logic, all nondense movable movers would succeed unless the  flag is set, and that happens right here (..() will be false if src is dense, for example).

Generally, if CanPass returns false on a movable on a turf, turf/Enter will return false, which will prevent the move and step will fail.

MistakeNot4892:
I have had a play with this locally and haven't managed to get it to work - will take another crack another time.

code/game/atoms_movable.dm
MistakeNot4892: Thanks for laying that out, I was a bit confused by the flag even though I put it there.

PR #809, opened 2020-08-18
*scream into the microphone

Alceris:


MistakeNot4892:
As discussed yesterday, this code introduces the potential for people to call `audible_message` (either directly on the mob or via a more abstract type) in a way consistent with the `/atom/movable`-level definition, which will display as a cut off message. Supplying a radio message as an argument to the proc would be a cleaner way to handle it without introducing inconsistencies.

MistakeNot4892:
This needs to be rebased, something is screwy with your branch.

MistakeNot4892:
Closing this in view of the alternative PR.

PR #811, opened 2020-08-20
Fixes the mining processor.

MarinaGryphon:
Mining processor would previously make duplicate entries due to lacking a text2path. It would create a duplicate entry for `"/decl/material/solid/mineral/hematite"` in addition to the actual entry for `/decl/material/solid/mineral/hematite`.
Fun fact: `new "/decl/material/solid/mineral/hematite"()` apparently works, meaning it created a whole new decl, which is how it was able to get info like the proper `solid_name`.

I'll change check-paths.sh soon, I swear.

### DIFF
--- a/code/modules/mining/machinery/mineral_processor.dm
+++ b/code/modules/mining/machinery/mineral_processor.dm
@@ -19,9 +19,9 @@
 /obj/machinery/mineral/processing_unit/Initialize()
 	ores_processing = list()
 	ores_stored = list()
-	for(var/orename in SSmaterials.processable_ores)
-		ores_processing[orename] = 0
-		ores_stored[orename] = 0
+	for(var/orepath in SSmaterials.processable_ores)
+		ores_processing[orepath] = 0
+		ores_stored[orepath] = 0
 	. = ..()
 
 /obj/machinery/mineral/processing_unit/Process()
@@ -158,7 +158,7 @@
 			if("Smelting")    choice = ORE_SMELT
 			if("Compressing") choice = ORE_COMPRESS
 			if("Alloying")    choice = ORE_ALLOY
-		ores_processing[href_list["toggle_smelting"]] = choice
+		ores_processing[text2path(href_list["toggle_smelting"])] = choice
 		. = TRUE
 	else if(href_list["toggle_power"])
 		active = !active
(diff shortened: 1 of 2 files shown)


MarinaGryphon:
Oh, I also changed the name of `orename` to `orepath` to make it more clear that it uses a typepath, not a material name. That misconception made it much harder to track down this bug. I can revert it if needed though.

MarinaGryphon:
Oh, and no matter what this will require upping the number of `text2path`s to 20, that's unavoidable. But it would technically, maybe, be cleaner to do it in get_decl, depending on the overhead added by such? If this is an issue elsewhere it would also fix that as well.

MarinaGryphon:
Fixes #812.

code/modules/mining/machinery/mineral_processor.dm
```diff
@@ -158,7 +158,7 @@
 			if("Smelting")    choice = ORE_SMELT
 			if("Compressing") choice = ORE_COMPRESS
 			if("Alloying")    choice = ORE_ALLOY
-		ores_processing[href_list["toggle_smelting"]] = choice
+		ores_processing[text2path(href_list["toggle_smelting"])] = choice
```
PsiOmegaDelta, line 161: There is no need to text2ref. Pass the ref of the decl instance to the UI, instead of its path, and `locate()` the instance

PsiOmegaDelta (changes requested):

code/modules/mining/machinery/mineral_processor.dm
MarinaGryphon: Uuuuuuuuuuuuuuuuuuuuuuuuh is that worth it? I feel like maybe a better option altogether would be just using text2path in get_decl to avoid this issue cropping up ever again

PsiOmegaDelta: Or pass the ref of the path and locate that, if that's what the `ores_processing()`-proc expects, is just a bit redundant.

PsiOmegaDelta: text2path is easier to exploit and has been known to fail to find types despite the string matching.
\ref + locate() is simply more robust.

MistakeNot4892: Agree with Psi. `get_decl` is hot code additionally, adding checking there for behavior that is almost always a coder mistake is not a good idea.

MistakeNot4892:
This should now be redundant in view of the mining rewrite.

MistakeNot4892:
Going to close this for now, feel free to reopen if the issue needs further addressing and you update the PR.

PR #814, opened 2020-08-20
Mining fixes.

### LINKED ISSUE #812: Ore processor creates duplicate entries when toggled.
#### Description of issue
https://cdn.discordapp.com/attachments/678827927755816970/746091178386259968/unknown.png
https://cdn.discordapp.com/attachments/678827927755816970/746091188926414970/unknown.png
Images courtesy of Alceris.

Duplicate entries are created, and I think it might not process? I'm not sure.

#### Steps to reproduce
1. Load ore into ore processor.
2. Click to toggle ore.
3. A duplicate entry is created. The original entry is unaffected.

N/A

Unsure, but I have a fix PR up.

N/A, affects master branch.

#### Issue bingo

MistakeNot4892:
Alternative to #811, fixes #812. Also fixes the stacker, refactors consoles to work nicer, and fixes the tradeship mining area on the map.

### DIFF
--- a/code/game/objects/items/weapons/circuitboards/machinery/mining.dm
+++ b/code/game/objects/items/weapons/circuitboards/machinery/mining.dm
@@ -1,6 +1,11 @@
-/obj/item/stock_parts/circuitboard/mineral_processing
+/obj/item/stock_parts/circuitboard/mineral_processing_console
 	name = T_BOARD("mineral processing console")
-	build_path = /obj/machinery/computer/mining
+	build_path = /obj/machinery/computer/mining/processing
+	origin_tech = "{'programming':2,'engineering':2}"
+
+/obj/item/stock_parts/circuitboard/mineral_stacking_console
+	name = T_BOARD("mineral stacking console")
+	build_path = /obj/machinery/computer/mining/stacking
 	origin_tech = "{'programming':2,'engineering':2}"
 
 /obj/item/stock_parts/circuitboard/mining_processor
--- a/code/modules/mining/machinery/mineral_processor.dm
+++ b/code/modules/mining/machinery/mineral_processor.dm
@@ -142,7 +142,8 @@
 					status_string = "<font color='gray'>alloying</font>"
 		else
 			status_string = "<font color='red'>not processing</font>"
-		result += "<tr><td>[line]</td><td><a href='?src=\ref[src];toggle_smelting=[ore]'>[status_string]</a></td></tr>"
+		var/decl/material/ore_mat = decls_repository.get_decl(ore)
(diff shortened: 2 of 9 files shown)


code/game/objects/items/weapons/circuitboards/machinery/mining.dm
```diff
@@ -1,6 +1,11 @@
-/obj/item/stock_parts/circuitboard/mineral_processing
+/obj/item/stock_parts/circuitboard/mineral_processing_console
 	name = T_BOARD("mineral processing console")
-	build_path = /obj/machinery/computer/mining
+	build_path = /obj/machinery/computer/mining/processing
+	origin_tech = "{'programming':2,'engineering':2}"
+
+/obj/item/stock_parts/circuitboard/mineral_stacking_console
+	name = T_BOARD("mineral processing console")
```
afterthought2: Should change the name if not using buildtype_select.

afterthought2 (commented):
While you're at it, I believe there was some change/fix to their density on bay, maybe as they are mapped on walls sometime? Unsure. May be worth checking out.

code/game/objects/items/weapons/circuitboards/machinery/mining.dm
MistakeNot4892: Re: density, they are currently dense and I'm mapping them on a floor, so just going to leave them. When I redo these machines the consoles won't exist anymore.

afterthought2 (approved):

code/modules/mining/machinery/mineral_processor.dm
```diff
@@ -151,15 +152,17 @@
...
-			if("Nothing")     choice = ORE_DISABLED
-			if("Smelting")    choice = ORE_SMELT
-			if("Compressing") choice = ORE_COMPRESS
-			if("Alloying")    choice = ORE_ALLOY
-		ores_processing[href_list["toggle_smelting"]] = choice
-		. = TRUE
+		var/decl/material/ore_mat = locate(href_list["toggle_smelting"])
+		if(istype(ore_mat) && (ore_mat.type in ores_processing))
+			var/choice = input("What setting do you wish to use for processing [ore_mat.solid_name]?") as null|anything in list("Smelting","Compressing","Alloying","Nothing")
+			if(!choice) return
```
PsiOmegaDelta, line 158: Don't want to be a stickler when it's mostly rearranged code but as usual with inputs
`/obj/machinery/mineral/processing_unit/Topic(href, href_list, state)`
`if(!choice || !CanInteract(usr, state))`

MistakeNot4892: The issue is that this machine may or may not be adjacent to the user as the console is what sends the href.
I'm intending to completely rewrite these machines to remove the console after this goes in so I can sort it out then if you don't mind leaving the crappy code in for awhile.

PR #817, opened 2020-08-22
Enforce standards

PsiOmegaDelta:
And also fixes sleeps in `/Process()` that for some reason haven't been caught before.

### DIFF
--- a/code/modules/submaps/_submap.dm
+++ b/code/modules/submaps/_submap.dm
@@ -14,7 +14,6 @@
 	. = ..()
 
 /datum/submap/proc/setup_submap(var/decl/submap_archetype/_archetype)
-
 	if(!istype(_archetype))
 		testing( "Submap error - [name] - null or invalid archetype supplied ([_archetype]).")
 		qdel(src)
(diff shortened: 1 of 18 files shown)


code/modules/submaps/_submap.dm
```diff
@@ -14,7 +14,7 @@
 	. = ..()
 
 /datum/submap/proc/setup_submap(var/decl/submap_archetype/_archetype)
-
+	set waitfor = FALSE
```
PsiOmegaDelta: This particular waitfor is really concerning.

afterthought2: Some digging suggests that the thing which sleeps here (load in new z) is on a branch which is never called unless a submap landmark is spawned directly (and in nullspace or otherwise not on a turf). That might be an oversimplification, but it does seem that this won't matter in routine use.

MistakeNot4892: As discussed on Discord the branch afterthought is referring to is vestigal code for a feature that never got finished, you could just delete the entire branch after `if(!associated_z)`.

PsiOmegaDelta: Vestigial code removed, keeping the second `if(!associated_z)`-check around.

PsiOmegaDelta:
Includes a bump to the latest version of the linter, which now discovers the `/Processing()`-issues.

MistakeNot4892 (approved):
Assuming tcomms and radio work (as discussed) this looks good to me.

PsiOmegaDelta:
Tcomms and radio tested, still appears to work.

PR #822, opened 2020-08-22
QoL for display cases

quardbreak:
* Display cases can be filled during runtime if they're unlocked with ID.
* You can remove any item from inside if you unlock case.
* Added extra cooldown for hand attacking.

image

### DIFF
--- a/code/game/objects/structures/displaycase.dm
+++ b/code/game/objects/structures/displaycase.dm
@@ -3,13 +3,16 @@
 	icon = 'icons/obj/structures/displaycase.dmi'
 	icon_state = "glassbox"
 	desc = "A display case for prized possessions. It taunts you to kick it."
-	density = 1
-	anchored = 1
-	unacidable = 1//Dissolving the case would also delete the gun.
+	density = TRUE
+	anchored = TRUE
+	unacidable = TRUE //Dissolving the case would also delete the gun.
 	alpha = 150
 	maxhealth = 100
 	hitsound = 'sound/effects/Glasshit.ogg'
-	var/destroyed = 0
+	req_access = null
+
+	var/destroyed = FALSE
+	var/locked = TRUE
 
 /obj/structure/displaycase/Initialize()
 	. = ..()
@@ -19,10 +22,19 @@
 			AM.forceMove(src)
 	update_icon()
 
-/obj/structure/displaycase/examine(mob/user)
+	if(!req_access)
+		var/area/A = get_area(src)
+		if(!istype(A) || !islist(A.req_access))
+			return
+		req_access = A.req_access.Copy()
+
+/obj/structure/displaycase/examine(mob/user, distance)
 	. = ..()
 	if(contents.len)
 		to_chat(user, "Inside you see [english_list(contents)].")
+
+	if(distance <= 1)
+		to_chat(user, "It looks [locked ? "locked. You can open it with your ID card" : "unlocked"].")
 
 /obj/structure/displaycase/explosion_act(severity)
 	..()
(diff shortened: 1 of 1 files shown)


code/game/objects/structures/displaycase.dm
```diff
@@ -78,12 +84,50 @@
 
 /obj/structure/displaycase/attackby(obj/item/W, mob/user)
 	user.setClickCooldown(DEFAULT_ATTACK_COOLDOWN)
-	take_damage(W.force)
+	if(istype(W, /obj/item/card/id))
+		if(allowed(usr))
+			locked = !locked
+			to_chat(user, "\The [src]'s lock was [locked ? "enabled" : "disabled"].")
+		else
+			to_chat(user, "\The [src]'s card reader denied you.")
```
MistakeNot4892: Minor grammar/tense nitpicking here, should really be `\The [src] was [locked ? "locked" : "unlocked"].` and `\The [src]'s card reader denies you access.` or such.

code/game/objects/structures/displaycase.dm
```diff
@@ -78,12 +84,50 @@
...
+		return
+
+	if(isitem(W) && (!locked || destroyed))
+		if(!W.simulated || W.anchored)
+			return
+
+		if(user.drop_item())
```
MistakeNot4892: `if(user.unEquip(W, src)` will properly check `canremove` and will also mean you can remove the forceMove() call below the pixel offset setting.

code/game/objects/structures/displaycase.dm
```diff
@@ -78,12 +84,50 @@
...
+	if(!locked || destroyed)
+		var/obj/item/selected_item
+		var/list/options = list()
+
+		for(var/atom/movable/AM in src)
+			var/image/radial_button = image(icon = AM.icon, icon_state = AM.icon_state)
+			options[AM] = radial_button
+
+		selected_item = show_radial_menu(user, src, options, radius = 42, require_near = TRUE, use_labels = TRUE)
+		if(!selected_item)
```
MistakeNot4892: You also need to check adjacency and if the user is incapacitated here, also if the selected item is still inside the display case.

code/game/objects/structures/displaycase.dm
```diff
@@ -78,12 +84,50 @@
...
+			return
+
+		to_chat(user, "You picking up \the [selected_item] from \the [src].")
```
MistakeNot4892: `You remove \the [selected_item] from \the [src].`

MistakeNot4892: Could also use a SPAN_NOTICE() on here and the other messages.

code/game/objects/structures/displaycase.dm
```diff
@@ -78,12 +84,50 @@
...
+			return
+
+		if(user.drop_item())
+			W.pixel_x = 0
+			W.pixel_y = -7
+			W.forceMove(src)
+			update_icon()
+		return
+
+	if(user.a_intent == I_HURT)
```
MistakeNot4892: Weapon damage to structures is handled by `..()` these days, you can just remove this and call ..().

code/game/objects/structures/displaycase.dm
```diff
@@ -78,12 +84,50 @@
...
+		take_damage(W.force)
 	..()
```
MistakeNot4892: `attackby()` needs to return a value these days, `. = ..()`.

code/game/objects/structures/displaycase.dm
```diff
@@ -3,13 +3,16 @@
...
-	anchored = 1
-	unacidable = 1//Dissolving the case would also delete the gun.
+	density = TRUE
+	anchored = TRUE
+	unacidable = TRUE //Dissolving the case would also delete the gun.
 	alpha = 150
 	maxhealth = 100
 	hitsound = 'sound/effects/Glasshit.ogg'
-	var/destroyed = 0
+	req_access = list()
```
MistakeNot4892: Just an idle thought, but it might be easiest to just make the display case take access from the area like airlocks do.

code/game/objects/structures/displaycase.dm
```diff
@@ -78,12 +84,46 @@
...
+			options[AM] = radial_button
+
+		selected_item = show_radial_menu(user, src, options, radius = 42, require_near = TRUE, use_labels = TRUE)
+		if(QDELETED(selected_item) || !contents.Find(selected_item) || !Adjacent(user))
+			return
+
+		to_chat(user, SPAN_NOTICE("You remove \the [selected_item] from \the [src]."))
+		selected_item.dropInto(loc)
+		update_icon()
+		return
```
MistakeNot4892: `return TRUE`

code/game/objects/structures/displaycase.dm
```diff
@@ -78,12 +84,46 @@
...
+
+	else if(!destroyed && user.a_intent == I_HURT)
+		visible_message(SPAN_WARNING("[user] kicks \the [src]."), SPAN_WARNING("You kick \the [src]"))
+		take_damage(2)
```
MistakeNot4892, line 136: This should return TRUE as well.

code/game/objects/structures/displaycase.dm
```diff
@@ -78,12 +84,46 @@
 
 /obj/structure/displaycase/attackby(obj/item/W, mob/user)
 	user.setClickCooldown(DEFAULT_ATTACK_COOLDOWN)
-	take_damage(W.force)
-	..()
+	if(istype(W, /obj/item/card/id))
```
MistakeNot4892:
````
var/obj/item/card/id/id = W.GetIdCard()
if(istype(id))
    blah
````
This would let people use PDAs and held mobs to open the case.

File: code/game/objects/structures/displaycase.dm [hunk 11]
```diff
@@ -3,27 +3,40 @@
...
 		if(AM.simulated && !AM.anchored)
 			AM.forceMove(src)
+	
+	if(!req_access)
+		var/area/A = get_area(src)
+		if(!istype(A) || !islist(A.req_access))
+			return
+		req_access = A.req_access.Copy()
+	
 	update_icon()
```
MistakeNot4892: Might need to move this before the req_access block, or change the block to not return.

code/game/objects/structures/displaycase.dm
```diff
@@ -78,12 +90,48 @@
...
+	if(!locked || destroyed)
+		var/obj/item/selected_item
+		var/list/options = list()
+
+		for(var/atom/movable/AM in src)
+			var/image/radial_button = image(icon = AM.icon, icon_state = AM.icon_state)
+			options[AM] = radial_button
+
+		selected_item = show_radial_menu(user, src, options, radius = 42, require_near = TRUE, use_labels = TRUE)
+		if(QDELETED(selected_item) || !contents.Find(selected_item) || !Adjacent(user))
```
MistakeNot4892: `|| user.incapacitated()`

MistakeNot4892 (approved):

PR #825, opened 2020-08-23
Ports QOL updates for vitals monitors from Bay

quardbreak:
Sorry no cherry-picking today
This PR ports Bay's vital monitor changes, credits to NullSnapshot for all changes ~~which I stole~~


code/game/machinery/OpTable.dm
```diff
@@ -25,6 +26,13 @@
 			computer.table = src
 			break
 
+/obj/machinery/optable/Destroy()
+	victim = null
+	if(connected_monitor)
+		connected_monitor.update_victim()
+		connected_monitor.update_optable()
```
MistakeNot4892, line 33: You need to null `connected_monitor`.

code/game/machinery/vitals_monitor.dm
```diff
@@ -13,12 +16,31 @@
...
+	//alert stuff
+	var/read_alerts = TRUE
+	var/alert_cooldown = 50 // buffer to prevent rapid changes in state
+	var/last_alert_time = 0
+	var/list/alerts = null
+	var/list/last_alert = null
+
+/obj/machinery/vitals_monitor/Initialize()
+	. = ..()
+	alerts = new /list(3)
```
MistakeNot4892, line 31: Why this weird syntax? Vitals monitors are not exactly hot code that needs memory or init optimization.

code/game/machinery/vitals_monitor.dm
```diff
@@ -50,76 +72,157 @@
...
+		to_chat(user, SPAN_NOTICE("Connected to adjacent [connected_optable]."))
 
 /obj/machinery/vitals_monitor/Process()
 	if(QDELETED(victim))
-		victim = null
+		update_victim()
 	if(victim && !Adjacent(victim))
-		victim = null
-		update_use_power(POWER_USE_IDLE)
+		update_victim()
```
MistakeNot4892, line 82: This whole proc is calling `update_victim()` repeatedly.

code/game/machinery/vitals_monitor.dm
```diff
@@ -50,76 +72,157 @@
...
 		update_icon()
-	if(beep && victim && victim.pulse())
-		playsound(src, 'sound/machines/quiet_beep.ogg')
-	
+
+/obj/machinery/vitals_monitor/proc/update_victim(var/new_victim = null)
+	var/old_victim = victim
+	victim = new_victim
+	if(victim)
+		visible_message(SPAN_NOTICE("\The [src] is now showing data for \the [victim]."))
```
MistakeNot4892, line 92: Why does this need a public broadcast message every single time the proc is called? At the very least it should only show if the victim has changed.

code/game/machinery/vitals_monitor.dm
```diff
@@ -50,76 +72,157 @@
...
 		else
 			overlays += image(icon, icon_state = "breathing_warning")
+			if(read_alerts)
+				alerts[LUNGS_ALERT] = "Patient is not breathing!"
 	else
 		overlays += image(icon, icon_state = "breathing_warning")
 
+/obj/machinery/vitals_monitor/proc/handle_alerts()
+	if(!victim || !read_alerts) //Clear our alerts
+		alerts[PULSE_ALERT] = ""
```
MistakeNot4892: `alerts.Cut()`

code/game/machinery/vitals_monitor.dm
```diff
@@ -128,11 +231,24 @@
...
 	if(!istype(user))
 		return
-	
+
 	if(CanPhysicallyInteract(user))
 		beep = !beep
 		to_chat(user, SPAN_NOTICE("You turn the sound on \the [src] [beep ? "on" : "off"]."))
-		
+
+/obj/machinery/vitals_monitor/verb/toggle_alerts()
```
MistakeNot4892: Vitals monitor currently has no `attack_hand()` definition, could use that instead of a verb.

code/game/machinery/OpTable.dm
```diff
@@ -16,6 +16,7 @@
 	var/mob/living/carbon/human/victim = null
 	var/strapped = 0.0
 	var/obj/machinery/computer/operating/computer = null
+	var/obj/machinery/vitals_monitor/connected_monitor = null
```
MistakeNot4892, line 19: You don't actually need to do `= null` on var declarations, it's copypasted from other code.

code/game/machinery/vitals_monitor.dm
```diff
@@ -13,12 +16,31 @@
...
+	var/list/alerts = null
+	var/list/last_alert = null
+
+/obj/machinery/vitals_monitor/Initialize()
+	. = ..()
+	alerts = new /list(3)
+	last_alert = new /list(3)
+	for(dir in list(NORTH,EAST,SOUTH,WEST))
+		connected_optable = locate(/obj/machinery/optable, get_step(src, dir))
+		if (connected_optable)
```
MistakeNot4892, line 35: Should check if it already has a monitor for mapping sanity purposes.

code/game/machinery/vitals_monitor.dm
```diff
@@ -50,76 +72,157 @@
...
-		victim = null
-		update_use_power(POWER_USE_IDLE)
+		update_victim()
 	else if(ishuman(over_object))
-		victim = over_object
-		update_use_power(POWER_USE_ACTIVE)
-		visible_message(SPAN_NOTICE("\The [src] is now showing data for [victim]."))
+		update_victim(over_object)
+	else if(istype(over_object, /obj/machinery/optable/))
+		var/obj/machinery/optable/new_table_connection = over_object
```
MistakeNot4892: You don't need to cast this, just pass over_object to update_optable below.

code/game/machinery/vitals_monitor.dm
```diff
@@ -50,76 +72,157 @@
...
+		update_optable(new_table_connection)
 
 /obj/machinery/vitals_monitor/on_update_icon()
 	overlays.Cut()
```
MistakeNot4892: While you're modifying the file it looks like it would be trivial to change this to use `cut_overlays()` and `add_overlay()`.

code/game/machinery/vitals_monitor.dm
```diff
@@ -50,76 +72,157 @@
...
+		overlays += image(icon, icon_state = "pulse_warning")
 
+/obj/machinery/vitals_monitor/proc/handle_brain()
+	if(!victim)
+		return
 	var/obj/item/organ/internal/brain/brain = victim.internal_organs_by_name[BP_BRAIN]
 	if(istype(brain) && victim.stat != DEAD && !(victim.status_flags & FAKEDEATH))
 		switch(brain.get_current_damage_threshold())
 			if(0 to 2)
 				overlays += image(icon, icon_state = "brain_ok")
```
MistakeNot4892: Not sure why these are procs, they are only used to update the icon and will break overlays if called outside of it.

code/game/machinery/vitals_monitor.dm
```diff
@@ -50,76 +72,157 @@
...
+		if(alerts[BRAIN_ALERT] && alerts[BRAIN_ALERT] != last_alert[BRAIN_ALERT])
+			audible_message(SPAN_WARNING("<b>\The [src]</b> alarms, \"[alerts[BRAIN_ALERT]]\""))
+		if(alerts[LUNGS_ALERT] && alerts[LUNGS_ALERT] != last_alert[LUNGS_ALERT])
+			audible_message(SPAN_WARNING("<b>\The [src]</b> warns, \"[alerts[LUNGS_ALERT]]\""))
+		last_alert = alerts.Copy()
+		last_alert_time = world.time
+
+
+
 /obj/machinery/vitals_monitor/verb/toggle_beep()
```
MistakeNot4892: Re: the attack_hand comment below, could make this alt click instead.

MistakeNot4892 (changes requested):
See comments. Code has quite a lot of issues as it currently stands.

code/game/machinery/OpTable.dm
NullSnapshot: I realize in retrospect this is kind of jank, but update_optable nulls the connected_optable's monitor by default before starting any new connection (incl null). See vitals_monitor.dm#L105

code/game/machinery/vitals_monitor.dm
NullSnapshot: I was spacing out on how associative lists worked and instead did list indexes like an idiot.

code/game/machinery/vitals_monitor.dm
NullSnapshot: It should be noted the only time this calls update_victim() repeatedly is on QDELETED(victim), which is triggered when either victim is marked for qdel, or victim is null.

code/game/machinery/vitals_monitor.dm
NullSnapshot: In practice that's exactly how this works. update_victim() is only called repeatedly as in the QDELETED case outlined above.

code/game/machinery/vitals_monitor.dm
NullSnapshot: See comment above about the list init.

code/game/machinery/vitals_monitor.dm
NullSnapshot: update_optable would have to be partially rewritten to attempt the cast instead. Trying to pass it directly causes a runtime.

code/game/machinery/vitals_monitor.dm
NullSnapshot: 👍

code/game/machinery/vitals_monitor.dm
NullSnapshot: 👍

code/game/machinery/vitals_monitor.dm
MistakeNot4892: You declare a variable at the scope of this condition and then never reference it. The same value is being passed to update_victim and cast as the same argument regardless.

MistakeNot4892:
Closing as eckff is AWOL, feel free to reopen.

PR #826, opened 2020-08-23
Plunger cleanup

quardbreak:
Merges plungers into single type.
Kills robo-plunger, now robots use normal plungers.


MistakeNot4892 (dismissed):
Wonder why I made it a mask originally.

MistakeNot4892 (approved):

PR #830, opened 2020-08-24
Ghosts can now view the contents of backpacks

quardbreak:
Ghosts still cannot interact with objects inside, they can only see.


MistakeNot4892:
Why do ghosts need to see the inside of backpacks?

quardbreak:
I think because ghosts still can see contents of lockers and crates. This kinda inspired me to make such feature.

MistakeNot4892 (approved):

PR #832, opened 2020-08-25
Implements anon say preference

quardbreak:
WIP, opening for feedback.
Removes ghost anonsay variable and replaces with preference check.

### DIFF
--- a/code/datums/communication/looc.dm
+++ b/code/datums/communication/looc.dm
@@ -39,7 +39,11 @@
 
 /client/proc/receive_looc(var/client/C, var/commkey, var/message, var/prefix)
 	var/mob/M = C.mob
-	var/display_name = isghost(M) ? commkey : M.name
+	var/display_name
+	if(isghost(M) && C.get_preference_value(/datum/client_preference/anon_say) != GLOB.PREF_YES)
+		display_name = commkey
+	else
+		display_name = M.name
 	var/admin_stuff = holder ? "/([commkey])" : ""
 	if(prefix)
 		prefix = "\[[prefix]\] "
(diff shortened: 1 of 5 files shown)


code/modules/mob/mob_helpers.dm
```diff
@@ -441,6 +441,8 @@ var/list/intents = list(I_HELP,I_DISARM,I_GRAB,I_HURT)
 			C = M.original.client
 
 	if(C)
+		if(C.get_preference_value(/datum/client_preference/anon_say) == GLOB.PREF_YES)
+			return
```
quardbreak, line 445: This is pretty questionable. Should we allow ghosts hide their presence with anonsay pref??

code/datums/communication/looc.dm
```diff
@@ -39,7 +39,7 @@
 
 /client/proc/receive_looc(var/client/C, var/commkey, var/message, var/prefix)
 	var/mob/M = C.mob
-	var/display_name = isghost(M) ? commkey : M.name
+	var/display_name = isghost(M) ? (C.get_preference_value(/datum/client_preference/anon_say) == GLOB.PREF_YES ? M.name : commkey) : M.name
 	var/admin_stuff = holder ? "/([commkey])" : ""
```
MistakeNot4892, line 47: This is getting pretty mangled. Maybe:
````
var/display_name
if(isghost(M) && C.get_preference_value(/datum/client_preference/anon_say) != GLOB.PREF_YES)
    display_name = commkey
else
    display_name = M.name
````

MistakeNot4892 (changes requested):

code/modules/mob/mob_helpers.dm
MistakeNot4892: Seems to be more or less the current behavior.

MistakeNot4892 (approved):

PR #835, opened 2020-08-25
Adds skill check for flooring actions

quardbreak:
Tweaks skill timer for flooring actions. Adds more checks for screwdriver and wrench.


code/game/turfs/simulated/floor_attackby.dm
```diff
@@ -20,30 +20,46 @@
 	if(flooring)
 		if(isCrowbar(C) && user.a_intent != I_HURT)
 			if(broken || burnt)
+				if(!user.do_skilled(flooring.remove_timer, SKILL_CONSTRUCTION, src, 0.15))
+					return
```
MistakeNot4892: While you're changing these - I believe they should be returning TRUE.

MistakeNot4892 (approved):

PR #838, opened 2020-08-27
Refactored the smell system.

MistakeNot4892:
This is needed to support pheromones on Scav but also makes the smell system a bit more interactive.

### DIFF
--- a/code/datums/extensions/scent/_scent.dm
+++ b/code/datums/extensions/scent/_scent.dm
@@ -63,16 +63,29 @@ Scent intensity
 		return PROCESS_KILL
 	emit_scent()
 
+/datum/extension/scent/proc/check_smeller(var/mob/living/carbon/human/smeller)
+	if(!istype(smeller) || smeller.stat != CONSCIOUS || smeller.failed_last_breath)
+		return FALSE
+	if(smeller.get_equipped_item(slot_wear_mask_str))
+		return FALSE
+	var/obj/item/head = smeller.get_equipped_item(slot_head_str)
+	if(head?.permeability_coefficient < 1)
+		return FALSE
+	return TRUE
+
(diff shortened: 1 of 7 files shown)


code/datums/extensions/scent/_scent.dm
```diff
@@ -63,16 +63,30 @@ Scent intensity
...
+	if(smeller.stat != CONSCIOUS)
+		return FALSE
+	if(smeller.failed_last_breath || smeller.wear_mask)
+		return FALSE
+	if(smeller.head && smeller.head.permeability_coefficient < 1)
+		return FALSE
+	return TRUE
+
 /datum/extension/scent/proc/emit_scent()
 	for(var/mob/living/carbon/human/H in all_hearers(holder, range))
```
PsiOmegaDelta: Noting that this loop already filters out all non-human smellers

code/datums/extensions/scent/_scent.dm
```diff
@@ -63,16 +63,30 @@ Scent intensity
...
 	emit_scent()
 
+/datum/extension/scent/proc/check_smeller(var/mob/living/carbon/human/smeller)
+	if(!istype(smeller))
+		return FALSE
+	if(smeller.stat != CONSCIOUS)
+		return FALSE
+	if(smeller.failed_last_breath || smeller.wear_mask)
+		return FALSE
+	if(smeller.head && smeller.head.permeability_coefficient < 1)
```
PsiOmegaDelta: Since there is work being done on the inventory system, you may want to consider using the various inventory procs like `get_equipped_item(SLOT)`

MistakeNot4892: I will need to change those lines again anyway since the new inventory code replaces slot definitions.

comma (approved):

PR #850, opened 2020-09-05
Fix protein bar reagents reacting

quardbreak:
Adds `atom_flags = ATOM_FLAG_NO_REACT` to `/obj/item/chems/food/snacks/candy/proteinbar`


MistakeNot4892:
This is not an issue on Nebula code.

PR #863, opened 2020-09-07
Backported Ministation from Scav.

ZirconiumWit:
Just backported ministation from scav, removed all scav content (I hope)


maps/ministation/jobs/command.dm
```diff
@@ -0,0 +1,159 @@
...
+	economic_power = 20
+	ideal_character_age = 70
+	guestbanned = 1
+	must_fill = 1
+	not_random_selectable = 1
+
+/datum/job/ministation/captain/equip(var/mob/living/carbon/human/H)
+	. = ..()
+	if(H.client)
+		H.client.verbs += /client/proc/freetradeunion_rename_company
```
quardbreak, line 34: pretty sure what this should be a mob verb, not client (if we giving verb to client they can use it after ghosting)

ZirconiumWit: Thank you for bringing this to my attention, that is a very valid concern.  If this requires changing, it appears to require changing on Tradeship as well.  I'm still learning the very fundamentals of this, so I will attempt to look at this in a future PR, with apologies.

MistakeNot4892:
This is more or less an exercise in syncing the ScavStation maps with Nebula (similar to previous Tradeship PR) so fixes and adjustments can be shared so personally I'm ok with merging it with some issues like the client verb. Can be fixed post-PR along with tradeship.

MistakeNot4892 (approved):

afterthought2:
OK, PR a fix or open an issue about the client verb.

PR #874, opened 2020-09-10
Ministation rework.

MistakeNot4892:
Opening for Travis at the moment.

- Reduces the ministation map bounds to 160x160.
- Adds the groundwork for cycling airlocks to all the external points.
- Adds a singularity engine to engineering and moves the solars around.
- Remaps/updates the mining area to work with the new machines.
- Adds an elevator and stairwell, and vestigal lower and upper levels (TBD).


MistakeNot4892:

PR #880, opened 2020-09-10
Makes floating messages on by default

quardbreak:
Since anyone can turn it off, but not everyone know about this feature.


MistakeNot4892:
The floating text is distinct enough and weird enough that I don't think it should be enabled by default. Will ping for a discussion.

MistakeNot4892:
Brief discussion seemed generally positive.

MistakeNot4892 (approved):

PR #882, opened 2020-09-11
Fixes issue with trash not entering the inlet.

MistakeNot4892:


afterthought2 (approved):
you mean that's not how you do a dir check?

PR #884, opened 2020-09-11
Ports baystation runtime fixes for synth instruments

quardbreak:


code/modules/synthesized_instruments/real_instruments.dm
```diff
@@ -283,7 +283,8 @@
 
 
 /obj/item/synthesized_instrument/ui_interact(mob/user, ui_key = "instrument", var/datum/nanoui/ui = null, var/force_open = 0)
-	real_instrument.ui_call(user,ui_key,ui,force_open)
+	if (real_instrument)
+		real_instrument.ui_call(user,ui_key,ui,force_open)
```
afterthought2, line 287: This proc should not be called on deleted instruments: if it is, that's an explicit call something is making illegally. Are there other ways `real_instrument` can be null?

code/modules/synthesized_instruments/song_editor.dm
```diff
@@ -17,15 +17,14 @@
 
 
 /datum/nano_module/song_editor/proc/current_page()
-	return src.song.current_line > 0 ? Ceiling(src.song.current_line / GLOB.musical_config.song_editor_lines_per_page) : src.page
+	return src.song.current_line > 0 ? Ceiling(src.song.current_line / GLOB.musical_config.song_editor_lines_per_page) : min(src.page, pages())
```
afterthought2, line 20: Seems superficial: if `pages` is decreased, `page` should be modified accordingly instead.

code/modules/synthesized_instruments/song_editor.dm
```diff
@@ -17,15 +17,14 @@
...
 
 
 /datum/nano_module/song_editor/proc/page_bounds(page_num)
 	return list(
-		max(1 + GLOB.musical_config.song_editor_lines_per_page * (page_num-1), 1),
+		max(min(1 + GLOB.musical_config.song_editor_lines_per_page * (page_num-1), src.song.lines.len), 1),
```
afterthought2, line 25: `Clamp`, though to be honest I don't see how this is an improvement.

MistakeNot4892:
Closing as eckff is AWOL, feel free to reopen.

PR #905, opened 2020-09-15
several hooks will now update higher z-levels

### LINKED ISSUE #893: Z-mimic blood uncleanable, but still there
#### Description of issue
Blood on a lower floor had been cleaned, but was still showing on the catwalk on the floor above. All other blood on catwalk had been removed, but _this_ blood was 'too far away'
Blood on lower floor had been properly cleaned up.

Ghost of Blood Cleaned should be beaten soundly by ghostbusters before anyone notices, and removed from view

#### Steps to reproduce
Unsure. only one spot out of almost a dozen did this.

blood splatters across two floors, blood being on both layers

Unsure

#### Issue bingo

ZirconiumWit:
setname and setinvisibility will now update when viewing a scene from higher z-levels.


PsiOmegaDelta (dismissed):

PsiOmegaDelta (dismissed):

Lohikar (changes requested):
Shoulda said originally, but please use `UPDATE_OO_IF_PRESENT` (no parens) instead of `update_above()` - it'll skip the proc-call entirely for unmimicked atoms.

Lohikar:
Fixes #893.

ZirconiumWit:
Changed to UPDATE_OO_IF_PRESENT and reverted upon determining that this was not an effective approach to the problem.

Lohikar (approved):

PR #907, opened 2020-09-15
Ports tgstation attack animation

quardbreak:


MistakeNot4892 (approved):
Attack animation should probably be defined on natural attacks and items or something so that a kick doesn't pivot from the bottom like this animation. But it does look a bit more dynamic than the current animation.

PR #908, opened 2020-09-15
NTOS programs will not start when there's no network connection

quardbreak:
Credits Hubblenaut


MistakeNot4892 (approved):
`P.on_startup()` should probably add `P` to `running_programs` itself, but this is fine for a fix.

comma:
Please no, programs should be able to handle not having connection rather than hard dying / not launching.
It may have some point on bay for 'no sneak peak' but here most programs were revamped to handle not being able to connect so it's pointless.

PR #913, opened 2020-09-16
no-vermap

quardbreak:
guards against a roundstart runtime where the torch is expected to exist but the overmap is disabled
also adds fetching ships by name and type from SSshuttle

### DIFF
--- a/code/controllers/subsystems/shuttle.dm
+++ b/code/controllers/subsystems/shuttle.dm
@@ -146,5 +146,15 @@ SUBSYSTEM_DEF(shuttle)
 		var/obj/effect/overmap/visitable/ship/ship_effect = ship
 		overmap_halted ? ship_effect.halt() : ship_effect.unhalt()
 
+/datum/controller/subsystem/shuttle/proc/ship_by_name(name)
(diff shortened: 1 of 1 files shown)


code/controllers/subsystems/shuttle.dm
```diff
@@ -146,5 +146,17 @@ SUBSYSTEM_DEF(shuttle)
 		var/obj/effect/overmap/visitable/ship/ship_effect = ship
 		overmap_halted ? ship_effect.halt() : ship_effect.unhalt()
 
+/datum/controller/subsystem/shuttle/proc/ship_by_name(name)
+	for (var/obj/effect/overmap/visitable/ship/ship in ships)
+		if (ship.name == name)
+			return ship
+	return null
```
MistakeNot4892: `return null` is redundant when `.` isn't set, you can just not return at all.

comma:
Kinda confused, it seems to be only adding two unused helpers, where's the runtime thing?

quardbreak:
@comma I actually add `setup_map()` thing like on torch on my fork, but there's no helpers for it, so I though to port helpers anyway so people can use them too.

comma (approved):

PR #922, opened 2020-09-18
Added hands and feet to prosthetic limb production, removed old robolimb installation surgery.

MistakeNot4892:
- You can now build feet and hands at the robotics fabricator.
- You can no longer build chests, as they are unusable with FBPs removed.
- You can no longer install a robot-mob bodypart as a robolimb (the old method, prior to Bay organs being distinct objects).
- Default robolimbs now use the macro properly.


code/modules/fabrication/designs/robotics/designs_prosthetics.dm
```diff
@@ -1,6 +1,45 @@
+#define DEFINE_ROBOLIMB_DESIGNS(MODEL_PATH, MODEL_ID, MODEL_NAME)          \
+/datum/fabricator_recipe/robotics/prosthetic/model_##MODEL_ID {            \
+	category = MODEL_NAME + " Prosthetics";                                \
```
MistakeNot4892, line 3: Sadly if I make these indented with spaces so the `\` lines up, it trips the incorrect indentation detection. :(

comma (approved):

PR #924, opened 2020-09-18
Hide AOOC broadcasts from non-AOOC users

quardbreak:
Ports some AOOC tweaks from baystation.
Credits to SierraKomodo.


code/datums/communication/aooc.dm
```diff
@@ -1,3 +1,5 @@
+#define SPAN_AOOC(X) "<span class='ooc'><span class='aooc'>[create_text_tag("aooc", "Antag-OOC:", target)] [X]</span></span>"
```
PsiOmegaDelta, line 1: Should undef non-global defines at the end of file, as they affect compile times

comma (dismissed):

PsiOmegaDelta (approved):

PR #932, opened 2020-09-19
Removes redundant dots from fusion interfaces

quardbreak:
They're ugly and looks bad on buttons.

Here's old vs new versions:
imageimage


TheKnightofAura:
Not a fan of half the changes here. I'd say you should keep the : and remove the . from online and offline

Edit: I'm assuming leaving the Injector Control : was intentional, but I do believe if this is gonna be pushed through, it should be all or nothing.

Also the preview you've submitted shows the pictures in reverse order, but the discord feed put them in the proper arrangement, but it is still confusing.

comma (approved):

PR #934, opened 2020-09-20
Remove unnecessary else statements

MarinaGryphon:
Removes unnecessary else statements following early returns.


code/_helpers/mobs.dm
```diff
@@ -49,8 +49,7 @@ proc/random_name(gender, species)
 		var/datum/species/current_species = get_species_by_key(species)
 		if(current_species)
 			var/decl/cultural_info/current_culture = SSlore.get_culture(current_species.default_cultural_info[TAG_CULTURE])
-			if(current_culture)
-				return current_culture.get_random_name(null, gender)
+			return current_culture?.get_random_name(null, gender)
```
PsiOmegaDelta: Non-equivalent control flow.
Before:
* If current_culture != null then return culture-name
* Otherwise, return some-random name (line 54)

Now
* Return culture-name or null

code/_onclick/cyborg.dm
```diff
@@ -91,9 +91,8 @@
 			if(!resolved && A && W)
 				W.afterattack(A, src, 1, params) // 1 indicates adjacency
 			return
-		else
-			W.afterattack(A, src, 0, params)
-			return
+		W.afterattack(A, src, 0, params)
+		return
```
PsiOmegaDelta: Superfluous return

code/_onclick/hud/screen_objects.dm
```diff
@@ -320,8 +320,7 @@
 				if(R.module)
 					R.hud_used.toggle_show_robot_modules()
 					return 1
-				else
-					to_chat(R, "You haven't selected a module yet.")
+				to_chat(R, "You haven't selected a module yet.")
```
PsiOmegaDelta, line 323: Unable to easily confirm equivalent control flow

code/game/machinery/igniter.dm
```diff
@@ -133,8 +133,7 @@
 /obj/machinery/sparker/attack_ai()
 	if (anchored)
 		return ignite()
-	else
-		return
+	return
```
PsiOmegaDelta, line 136: Superfluous return

code/game/machinery/flasher.dm
```diff
@@ -48,8 +48,7 @@
 /obj/machinery/flasher/attack_ai()
 	if (src.anchored)
 		return src.flash()
-	else
-		return
+	return
```
PsiOmegaDelta, line 51: Superfluous return

code/_helpers/mobs.dm
MarinaGryphon: Ah shoot. Didn't mean to keep this change

code/_onclick/cyborg.dm
MarinaGryphon:
```suggestion
```

code/_onclick/hud/screen_objects.dm
MarinaGryphon: `if(X) ... return else Y` is equivalent to `if(X) ... return; Y`

code/_helpers/mobs.dm
MarinaGryphon:
```suggestion
			if(current_culture)
				return current_culture.get_random_name(null, gender)
```

code/_onclick/hud/screen_objects.dm
MarinaGryphon: Unless you meant to put this on the internals check section.

PsiOmegaDelta:
Didn't spot any more major issues like https://github.com/NebulaSS13/Nebula/pull/934#discussion_r491666453 but because it's so easy to overlook I'm not sure I'm willing to trust all these changes either.

code/game/machinery/igniter.dm
MarinaGryphon:
```suggestion
```

code/game/machinery/flasher.dm
MarinaGryphon:
```suggestion
```

code/datums/extensions/eye/_eye.dm
```diff
@@ -90,8 +90,7 @@
 			return ..(eye_extension.extension_eye)
 		if(EXTENSION_TARGET)
 			return ..(eye_extension)
-		else
```
MistakeNot4892, line 93: Not sure if this is ideal, in the case of switches the `else` section is the `default` from other languages, it's not quite the same as an unnecessary else after a return.

code/_onclick/hud/screen_objects.dm
MistakeNot4892: This entire section looks fine to me, the logic has only been tabbed back.

code/datums/mil_ranks.dm
```diff
@@ -85,8 +84,7 @@ var/datum/mil_branches/mil_branches = new()
 	var/datum/mil_branch/branch = get_branch(branch_name)
 	if(branch && (rank_name in branch.spawn_ranks(S)))
```
MistakeNot4892, line 85: Could just be `return (rank_name in branch?.spawn_ranks(S))`.

code/game/atoms.dm
```diff
@@ -393,17 +392,15 @@ its easier to just keep the beam vertical.
...
 
 /atom/proc/checkpass(passflag)
 	return pass_flags&passflag
 
 /atom/proc/isinspace()
 	if(istype(get_turf(src), /turf/space))
 		return 1
-	else
-		return 0
+	return 0
```
MistakeNot4892, line 403: All the procs modified in this file could just return their conditional instead of explicitly returning true or false.

code/game/machinery/_machines_base/machinery_power.dm
```diff
@@ -53,8 +53,7 @@ This is /obj/machinery level code to properly manage power usage from the area.
 			return idle_power_usage
 		if(POWER_USE_ACTIVE)
 			return active_power_usage
-		else
-			return 0
+	return 0
```
MistakeNot4892, line 56: As with the previous switch I'm not sure this is a good change.

code/game/machinery/flasher.dm
MistakeNot4892: `return anchored && flash()`?

code/game/machinery/kitchen/icecream.dm
```diff
@@ -42,8 +42,7 @@
 			return list(/decl/material/liquid/nutriment/flour, /decl/material/liquid/nutriment/sugar)
 		if(CONE_CHOC)
 			return list(/decl/material/liquid/nutriment/flour, /decl/material/liquid/nutriment/sugar, /decl/material/liquid/nutriment/coco)
-		else
-			return list(/decl/material/liquid/drink/milk, /decl/material/solid/ice)
+	return list(/decl/material/liquid/drink/milk, /decl/material/solid/ice)
```
MistakeNot4892, line 45: As with other switches.

code/game/machinery/kitchen/icecream.dm
```diff
@@ -61,8 +60,7 @@
 			return "waffle"
 		if(CONE_CHOC)
 			return "chocolate"
-		else
-			return "vanilla"
+	return "vanilla"
```
MistakeNot4892, line 63: As with other switches.

code/modules/mechs/mech_damage.dm
```diff
@@ -73,8 +73,7 @@
 			return arms
 		if(BP_L_LEG , BP_R_LEG)
 			return legs
-		else
-			return body
+	return body
```
MistakeNot4892, line 76: Switch.

code/modules/mob/living/carbon/human/inventory.dm
```diff
@@ -75,8 +75,7 @@ This saves us from having to call add_fingerprint() any time something is put in
 			return 1
 		if(slot_tie_str)
 			return 1
-		else
-			return has_organ(slot)
+	return has_organ(slot)
```
MistakeNot4892, line 78: Switch.

code/datums/extensions/eye/_eye.dm
MarinaGryphon: I checked that in every case I removed an else in a switch, that EVERY case returned. That means that the only time the code after the switch is reached is when all conditions are false, i.e. the default case.

code/game/atoms.dm
MarinaGryphon: See, I was told not to change that. I can fix that right now though!

Lohikar:
Per discord, I don't think this PR (at least in its current state) is a good idea. 
There's so many changes in this PR that it's difficult to properly review and I'm not confident that the logic is all 100% equivalent. 
I don't think what little benefits this minor formatting change offers are worth it.

I'll leave this up to Psi/Loaf though.

code/datums/extensions/eye/_eye.dm
MistakeNot4892: That is my point. There's a distinction to be drawn between `else` as the default case of a switch and `else` as the fallthrough of a conditional and your changes are removing that visual distinction for no tangible benefit.

MistakeNot4892:
Kind of inclined to agree with Lohi especially after hearing this was done with a regex. It has no concrete benefits that would make up for regressions.

MistakeNot4892:
Discussed this on Discord, the general attitude is not positive. It's a very large PR that may introduce strange behaviors, muddies the diff pretty badly, and has some questionable style issues. I am going to close it, apologies.

PR #935, opened 2020-09-20
minimum reimplementation of 28016

quardbreak:
Disables administrative topic inputs when a password is not set.

### DIFF
--- a/code/game/world.dm
+++ b/code/game/world.dm
@@ -104,6 +104,11 @@ var/world_topic_spam_protect_time = world.timeofday
 /world/Topic(T, addr, master, key)
 	diary << "TOPIC: \"[T]\", from:[addr], master:[master], key:[key][log_end]"
 
+	/* * * * * * * *
+	* Public Topic Calls
+	* The following topic calls are available without a comms secret.
+	* * * * * * * */
+
 	if (T == "ping")
 		var/x = 1
 		for (var/client/C)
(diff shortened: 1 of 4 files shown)


code/modules/paperwork/photocopier.dm
```diff
@@ -146,7 +146,7 @@
 		c.info = "<font color = #101010>"
 	else			//no toner? shitty copies for you!
 		c.info = "<font color = #808080>"
-	var/copied = html_decode(copy.info)
```
PsiOmegaDelta, line 149: Isn't the appropriate solution to re-encode it on line 152?

PsiOmegaDelta: Actually, didn't read what happened to c.info just above

code/game/world.dm
```diff
@@ -189,6 +194,51 @@ var/world_topic_spam_protect_time = world.timeofday
...
+		ban_unban_log_save("[input["id"]] has permabanned [C.ckey]. - Reason: [input["reason"]] - This is a ban until appeal.")
+		notes_add(target,"[input["id"]] has permabanned [C.ckey]. - Reason: [input["reason"]] - This is a ban until appeal.",input["id"])
+		qdel(C)
+
+	/* * * * * * * *
+	* Secure Topic Calls
+	* The following topic calls are only available if a comms secret has been defined, supplied, and is correct.
+	* * * * * * * */
+
+	if (!config.comms_password)
```
PsiOmegaDelta, line 239: Noting that this covers both empty string and null, so the change of the default value doesn't do much

PsiOmegaDelta (approved):

PR #942, opened 2020-10-05
Backports mech heat fixes from bay

CrimsonShrike:
Fixes solar storms annihilating them and them being deleted by fires compared to other mobs.

### DIFF
--- a/code/modules/events/solar_storm.dm
+++ b/code/modules/events/solar_storm.dm
@@ -31,6 +31,8 @@
 /datum/event/solar_storm/proc/radiate()
 	// Note: Too complicated to be worth trying to use the radiation system for this.  Its only in space anyway, so we make an exception in this case.
 	for(var/mob/living/L in GLOB.living_mob_list_)
+		if(L.loc?.atom_flags & ATOM_FLAG_SHIELD_CONTENTS)
+			continue
 		var/turf/T = get_turf(L)
 		if(!T || !(T.z in GLOB.using_map.player_levels))
 			continue
@@ -39,10 +41,10 @@
 			continue
 
 		//Apply some heat or burn damage from the sun.
-		if(istype(L, /mob/living/carbon/human))
-			L.bodytemperature += temp_incr
-		else
-			L.adjustFireLoss(fire_loss)
+		if(L.increaseBodyTemp(temp_incr))
+			continue
+			
+		L.adjustFireLoss(fire_loss)
 
 
 /datum/event/solar_storm/end()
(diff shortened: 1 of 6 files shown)


code/modules/mechs/mech_life.dm
```diff
@@ -67,13 +67,18 @@
 /mob/living/exosuit/handle_environment(var/datum/gas_mixture/environment)
 	if(!environment) return
 	//Mechs and vehicles in general can be assumed to just tend to whatever ambient temperature
-	if(abs(environment.temperature - bodytemperature) > 10 )
-		bodytemperature += ((environment.temperature - bodytemperature) / 3)
-
-	if(environment.temperature > material.melting_point * 1.25 ) //A bit higher because I like to assume there's a difference between a mech and a wall
```
CrimsonShrike, line 73: Main issue was that bodytemp was ignored, also damage ticks could be in the many hundreds.

code/modules/events/solar_storm.dm
```diff
@@ -39,7 +41,7 @@
 			continue
 
 		//Apply some heat or burn damage from the sun.
-		if(istype(L, /mob/living/carbon/human))
+		if(istype(L, /mob/living/carbon/human) || istype(L, /mob/living/exosuit))
```
MistakeNot4892: This should be some `AdjustBodyTemp()` proc that humans and exosuits can override.

code/modules/events/solar_storm.dm
```diff
@@ -31,6 +31,8 @@
 /datum/event/solar_storm/proc/radiate()
 	// Note: Too complicated to be worth trying to use the radiation system for this.  Its only in space anyway, so we make an exception in this case.
 	for(var/mob/living/L in GLOB.living_mob_list_)
+		if(istype(L.loc, /mob/living/exosuit)) //Todo, generalize this further like I did for rads if other things need to block or mitigate heat - CrimsonShrike
```
MistakeNot4892: This is the kind of pattern that helped make old mech code absolute shit to deal with. An `ATOM_FLAG_SHIELD_CONTENTS` set on mechs would make more sense.

MistakeNot4892 (changes requested):

code/modules/events/solar_storm.dm
CrimsonShrike: That naming was used by a proc that I dont quite understand, but done.

MistakeNot4892 (approved):

PR #945, opened 2020-10-10
Adds docking beacons

NataKilar:
Adds docking beacons and refactors free landing restrictions a bit.

* Sectors can now have a portion of the interior of their z-level(s) set to restricted landing, forbidden free landing within this area rather than having a blanket ban/permit on free landing per sector.
* Added magnetic docking beacons, which permit landing within their designated area. They are controllable via the docking program, and access can be restricted to by shuttle permission, or docking by code. Landing within a docking beacon designated area is the only way for a shuttle to land "securely", and not be forced back to their (ship) z-level if the sector they are landing in moves away from them.
* Removes the visitor landmarks generated by landable ships on initialization. Ship-Ship encounters can now occur using the landing system, which largely makes these extraneous. The only strange interation that may occur is if a shuttle which does not have an associated ship lands in the other ships z-level, which would cause them to remain even when the other shuttle moves away. If this is a major concern, then the visitor landmarks can be readded.


comma (changes requested):
As per discord discussion pls bring back the visitor landmarks, since aside from the edge case you mentioned it also locks non-experts from visiting other ships

MistakeNot4892:
Poking it to reset actions.

NataKilar:
Visitor Landmarks have been readded, and access problems we discussed on discord have been addressed as well.

comma (approved):

PR #958, opened 2020-11-21
Makes snacks use standard_feed_mob

comma:
Fulfills an ancient TODO.
Behaves same as it did player-facing, only new side effect is half eaten food actually getting trace DNA

### DIFF
--- a/code/modules/reagents/reagent_containers.dm
+++ b/code/modules/reagents/reagent_containers.dm
@@ -153,13 +153,7 @@
 
 
 		else
-			var/mob/living/carbon/H = target
-			if(!H.check_has_mouth())
-				to_chat(user, "Where do you intend to put \the [src]? \The [H] doesn't have a mouth!")
-				return
-			var/obj/item/blocked = H.check_mouth_coverage()
-			if(blocked)
-				to_chat(user, "<span class='warning'>\The [blocked] is in the way!</span>")
+			if(!user.can_force_feed(target, src))
 				return
 
 			other_feed_message_start(user, target)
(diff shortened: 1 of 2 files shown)


code/modules/reagents/reagent_containers.dm
```diff
@@ -153,6 +153,8 @@
 
 
 		else
+			if(!user.can_force_feed(target, src))
+				return
 			var/mob/living/carbon/H = target
 			if(!H.check_has_mouth())
 				to_chat(user, "Where do you intend to put \the [src]? \The [H] doesn't have a mouth!")
```
MistakeNot4892: This is redundant with the above check.

code/modules/reagents/reagent_containers/food/snacks.dm
```diff
@@ -57,58 +82,12 @@
...
-			if(eat_sound)
-				playsound(M, pick(eat_sound), rand(10, 50), 1)
-			if(reagents.total_volume)
-				if(reagents.total_volume > bitesize)
-					reagents.trans_to_mob(M, bitesize, CHEM_INGEST)
-				else
-					reagents.trans_to_mob(M, reagents.total_volume, CHEM_INGEST)
-				bitecount++
-				On_Consume(M)
+		if (fullness > 550)
```
MistakeNot4892, line 85: Could this be moved to `can_force_feed`?

comma: It covers both self-propelled eating and force feeding tho

code/modules/reagents/reagent_containers.dm
comma: axed

MistakeNot4892 (approved):

PR #959, opened 2020-11-21
Antagonist selection for mixed modes is more likely to be successful.

mikomyazaki:
_Originally PR'd for Aurora [here](https://github.com/Aurorastation/Aurora.3/pull/9583) plus some further bugfix/adjustment PRs._

Adjusts antagonist selection for mixed modes, performing some sorting so that the mixed mode is more likely to be successfully started. This has proven important on Aurora where we have many popular mixed modes, and it was common for mode selection to be restarted despite having appropriate candidates.

* Candidates with fewer eligible modes are selected preferentially for that mode, leaving more eligible players for the other modes.
* Candidates are assigned to the mode that requires fewest players, first.
* Selection is still made as random as possible within these limits.

### DIFF
--- a/code/_helpers/lists.dm
+++ b/code/_helpers/lists.dm
@@ -216,12 +216,6 @@ Checks if a list has the same entries and values as an element of big.
 	for(var/e in L)
 		. += L[e]
 
-// Return a list of the keys in an assoc list
-/proc/list_keys(var/list/L)
-	. = list()
-	for(var/e in L)
-		. += e
-
 //Mergesort: divides up the list into halves to begin the sort
 /proc/sortKey(var/list/client/L, var/order = 1)
 	if(isnull(L) || L.len < 2)
--- a/code/_helpers/lists.dm
+++ b/code/_helpers/lists.dm
@@ -216,6 +216,12 @@ Checks if a list has the same entries and values as an element of big.
 	for(var/e in L)
 		. += L[e]
 
+// Return a list of the keys in an assoc list
+/proc/list_keys(var/list/L)
+	. = list()
+	for(var/e in L)
+		. += e
+
 //Mergesort: divides up the list into halves to begin the sort
 /proc/sortKey(var/list/client/L, var/order = 1)
 	if(isnull(L) || L.len < 2)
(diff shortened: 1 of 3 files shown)


code/game/gamemodes/game_mode.dm
```diff
@@ -201,9 +201,37 @@ var/global/list/additional_antag_types = list()
...
 		antag.build_candidate_list(src) //compile a list of all eligible candidates
 
+	if(length(antag_templates) > 1) // If we have multiple templates to satisfy, we must pick candidates who satisfy fewer templates first, and fill the template with fewest candidates first
+		var/list/template_candidates = list()
+		var/list/all_candidates = list() // All candidates for every template, may contain duplicates
+		var/list/antag_templates_by_initial_spawn_req = list()
+
+		for(var/datum/antagonist/antag in antag_templates)
+			template_candidates[antag.id] = length(antag.candidates)
+			all_candidates += antag.candidates
```
MistakeNot4892, line 211: Should probably be `|=`, assuming that other antag templates can grab the same people as candidates.

code/_helpers/lists.dm
```diff
@@ -216,6 +216,12 @@ Checks if a list has the same entries and values as an element of big.
 	for(var/e in L)
 		. += L[e]
 
+// Return a list of the keys in an assoc list
+/proc/list_keys(var/list/L)
```
MistakeNot4892: As mentioned on Discord I think this should probably have a more specific name, as `list_keys` could be misleading about what it actually does.

MistakeNot4892 (dismissed):

code/game/gamemodes/game_mode.dm
mikomyazaki: `all_candidates` needs to be able to contain duplicates, as it is used to generate `valid_templates_per_candidate`. If a candidate appears twice in `all_candidates`, then it means they are valid choices for two templates.

There's a comment on line 206 that mentions this, maybe the list could be named better? Nothing springs to mind.

code/_helpers/lists.dm
mikomyazaki: Removed `list_keys()`.

code/game/gamemodes/game_mode.dm
MistakeNot4892: Ah my bad.

MistakeNot4892 (approved):

PR #965, opened 2020-11-25
Makes lighting bi-directional

Andrew-Fall:
Crawls both up and down z_levels instead of just down


Lohikar:
Per comments on discord, I don't really like this approach. My main issue is with the density of the macros in this - adding anything else to the occlusion loop (like directional lights) would probably be unpleasant.

While I would prefer Z-bleed over this, I don't think there's a huge difference in terms of behavior between this and Z-bleed.

Andrew-Fall:
Intention is to make lights bi-directional using a similar system to what we have now, and in the process remove the last GOTO in favor of a macro. Porting z-bleed is outside the scope of this PR.

PR #966, opened 2020-11-26
Generalized blood/oil coating

comma:
Items now have 'coating' reagent container for that.
Hand/foot blood is stored on appropriate organ objects now.


nebula.dme
```diff
@@ -3234,6 +3234,7 @@
 #include "maps\antag_spawn\wizard\wizard.dm"
 #include "maps\away\away_sites.dm"
 #include "maps\away_sites_testing\away_sites_testing_define.dm"
+#include "maps\example\example.dm"
```
MistakeNot4892: Oi.

comma: it's a superior map smdh

MistakeNot4892 (approved):

PR #970, opened 2020-11-28
Makes roundstart, credits and server reboot sounds map preferred

quardbreak:
Need to figure out why server restart sound don't work. Other sound features work fine.


afterthought2:
Possibly need to delay the reboot for the sound to actually play, which would be undesirable in most uses of rebooting. Could try adding it in one of the end-round steps in the ticker instead.

MistakeNot4892 (approved):
Fine with this as an interim merge prior to getting the reboot sound to actually play.

PR #1011, opened 2020-12-09
Added an exterior variant of open space, refactored open space type checks to use isopenspace() and isspaceturf().

MistakeNot4892:
Working towards multiz exoplanets.
This appears to work fine, I have tested by replacing the tradeship open spaces with the new version and behavior seems to be as expected.


code/game/objects/items/weapons/RCD.dm
```diff
@@ -99,7 +99,7 @@
 	if(!proximity) return
 	if(disabled && !isrobot(user))
 		return 0
-	if(istype(get_area(A),/area/shuttle)||istype(get_area(A),/turf/space/transit))
+	if(istype(get_area(A),/area/shuttle) || istype(get_turf(A), /turf/space/transit))
```
MistakeNot4892, line 102: This seems to be an old mistake.

code/modules/multiz/turf.dm
```diff
@@ -31,9 +22,17 @@
 	icon_state = ""
 	density = 0
 	pathweight = 100000 //Seriously, don't try and path over this one numbnuts
-
 	z_flags = ZM_MIMIC_DEFAULTS | ZM_MIMIC_OVERWRITE | ZM_MIMIC_NO_AO | ZM_ALLOW_ATMOS
 
+/turf/simulated/open/CanZPass(atom/A, direction)
```
Lohikar, line 27: Not in scope for this PR, but I feel like this behavior should be moved down to `/turf` and gated by a ZM flag or the like at some point.

Lohikar (approved):

PR #1013, opened 2020-12-10
Misc tweaks and changes from map tinkering session.

MistakeNot4892:
See commit log for details - more or less just runtime/typo fixes and one repath.


afterthought2 (approved):
urgh, gun add

PR #1022, opened 2020-12-12
Grayscaled Alarm and Status Display Icons

Mazianni:
This PR changes fire alarms and status displays to using greyscaled icons instead of static icons. Included are a handful I whipped up, so this should allow for an easier variety of code displays.


code/datums/security_state.dm
```diff
@@ -148,6 +148,13 @@
 	var/up_description
 	var/down_description
 
+	var/datum/alarm_appearance/alarm_appearance
+	var/alarm_appearance_type
+
+/decl/security_level/New()
+	. = ..()
+	alarm_appearance = new alarm_appearance_type
```
MistakeNot4892: Need to typecheck this in case it's null, since the default value is null.

MistakeNot4892 (changes requested):
Please rebase and squash this, then I will approve it for merge.

MistakeNot4892 (approved):
If I was a cruel man I'd go :eyes: at the fix commit.

Mazianni:
🔪

PR #1023, opened 2020-12-12
Fixes old russian keyboard layout for radio

quardbreak:


afterthought2:
Need confirmation that 513 parses utf8 correctly in the source, including outside of literal strings. This change also means we no longer support compiling on 512, and cannot be gated with defines (or at least not safely/usefully) so I'm not happy about it. Perhaps this is best left to downstreams to mess with.

Lohikar:
This can be gated off with defines just by using includes.

PR #1025, opened 2020-12-12
/datum/species => /decl/species

MistakeNot4892:
Getting the repath out of the way - the only logic change here is redoing how species lists are generated to use the decls repository instead of directly creating new species instances.


afterthought2 (dismissed):
hope it works

afterthought2 (approved):

PR #733, opened 2020-07-26
Telecomms radio => network radio. [MDB IGNORE] [IDB IGNORE]

MistakeNot4892:
Hoo boy this is another stupidly big one. Opening as a draft for input because I am a bit bogged down in it now.

- Deleted telecomms. It was great, would recommend.
- Refactored radios (handheld, intercom, headsets) to be network devices.
- Added a telecomms hub network device that populates an internal table of channels, which headsets/intercoms sync to. I think I got almost all of the globals that were used for this previously.
- Minor thing, but message mode now parses using the radio channel prefix preference instead of always looking for `:` or `.`.
- Deleted a ton of intercom and radio presets that aren't currently mapped or in use.
- Refactored encryption keys and channel access to be handled via access constants like everything else in the game.

**Done so far:**
- Telecomms hub is mapped in on the example map and successfully generates the usual Exodus set of channels on init.
- Radios will connect to the network and sync with the hub, then send their messages to the hub to be broadcast.
- Tested that cycling airlocks and signallers still work - shuttle docks/undocks properly and a signaller on a bolt wire toggles the bolts.
- Telecomms hub can be used to configure and modify the channel list.
- The message monitor, pager, and request console are network devices now.

**Todo:**
- [ ] Readd the headsets to lockers god dammit democracy
- [ ] Update specialty intercomms on antag maps.
- [ ] Test long range relays.
- [ ] Test comms masers/antennae.
- [x] Work out why the tcomms relay stops working when power cycled.

PR #1027, opened 2020-12-12
Pagers, request consoles and the message monitor are now network devices.

MistakeNot4892:
Splitting these changes off #733, DNM, still need to be tested/finalized.

### DIFF
--- a/code/game/machinery/computer/message.dm
+++ b/code/game/machinery/computer/message.dm
@@ -6,21 +6,37 @@
 	icon_screen = "comm_logs"
 	light_color = "#00b000"
 	var/hack_icon = "error"
-	//Server linked to.
-	var/obj/machinery/message_server/linkedServer = null
-	//Sparks effect - For emag
 	var/datum/effect/effect/system/spark_spread/spark_system = new /datum/effect/effect/system/spark_spread
-	//Messages - Saves me time if I want to change something.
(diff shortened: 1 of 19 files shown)


afterthought2 (dismissed):

MistakeNot4892:
I think(?) this is all working. Hard to test on our maps because none of them actually have pagers and request consoles mapped out. Message server still does some silly magic horseshit in send_to_departments but otherwise this seems functional.

code/game/machinery/computer/message.dm
```diff
@@ -271,23 +292,26 @@
 			src.screen = 2
 			update_icon()
 			//Time it takes to bruteforce is dependant on the password length.
-			spawn(100*length(src.linkedServer.decryptkey))
-				if(src && src.linkedServer && usr)
-					BruteForce(usr)
+			var/decrypting = linked_server
+			spawn(100*length(linked_server.decryptkey))
+				if(src)
```
afterthought2: Timer please (`if(src)` is too much pain for me).

afterthought2 (commented):
Current code is an atrocity, so may as well just go for it. If these consoles are broken, it's much more likely to be my fault anyway when I messed with their mapping stuff.

afterthought2 (approved):

PR #1033, opened 2020-12-13
Converted faulty augments into augment faults.

MistakeNot4892:
Adds some simple health ailments that give you a message every 2-5 minutes, cured with simple and fairly common meds/items.
Also converts the augment faults into prosthetic ailments that can be added by EMP damage.

**TODO**
- [x] Add meat organ ailments.
- [x] Add reagent dosage cures for ailments.
- [x] Add item-contact cures for ailments.


code/modules/admin/view_variables/topic.dm
```diff
@@ -316,6 +316,53 @@
...
 
+	else if(href_list["addailment"])
+
+		var/mob/living/carbon/human/H = locate(href_list["addailment"])
+		if(!istype(H))
+			to_chat(usr, "This can only be done to instances of type /mob/living/carbon/human")
+			return
+		var/obj/item/organ/O = input("Select a limb to add the ailment to.", "Add Ailment") as null|anything in (H.organs|H.internal_organs)
+		if(!O)
+			return
```
afterthought2, line 327: Need to recheck that it's a valid limb; list might have changed.

code/modules/admin/view_variables/topic.dm
```diff
@@ -316,6 +316,53 @@
...
+		var/list/possible_ailments = list()
+		for(var/atype in subtypesof(/datum/ailment))
+			var/datum/ailment/ailment = get_ailment_reference(atype)
+			if(ailment && ailment.category != ailment.type && ailment.can_apply_to(O))
+				possible_ailments |= ailment
+
+		var/datum/ailment/ailment = input("Select an ailment type to add.", "Add Ailment") as null|anything in possible_ailments
+		if(!istype(ailment))
+			return
+		if(!QDELETED(O) && O.add_ailment(ailment))
```
afterthought2: Need to do another recheck that it's a valid attached limb, and also maybe the previous valid ailment check if that's contingent on limb status.

code/modules/events/ailments.dm
```diff
@@ -0,0 +1,19 @@
...
+	var/list/candidates = list()
+	for(var/mob/living/carbon/human/H in GLOB.living_mob_list_)
+		if(H.client && !length(H.stasis_sources))
+			candidates += H
+	var/create_ailments = min(length(candidates), rand(1,3))
+	while(length(candidates) && create_ailments)
+		var/mob/living/carbon/human/H = pick_n_take(candidates)
+		var/list/organs = (H.organs|H.internal_organs)
+		while(length(organs))
+			var/obj/item/organ/O = pick_n_take(organs)
```
afterthought2: It's a lot faster to shuffle this once and then loop over it instead of pick_n_taking repeatedly.

code/modules/organs/ailments/_ailment.dm
```diff
@@ -0,0 +1,96 @@
...
+	var/max_time = 5 MINUTES
+	var/category = /datum/ailment
+	var/obj/item/organ/organ
+
+	// Requirements before applying to a target.
+	var/list/applies_to_organ
+	var/specific_organ_subtype = /obj/item/organ/external
+	var/affects_robotics = FALSE
+
+	var/treated_by_item_type
```
afterthought2: Add comment about what this var takes (type? list of types?), as this is probably what anyone touching this system will want to touch.

code/modules/organs/ailments/_ailment.dm
```diff
@@ -0,0 +1,96 @@
...
+			to_chat(organ.owner, SPAN_WARNING(capitalize(replacetext(initial_ailment_message, "$ORGAN$", organ.name))))
+			begin_ailment_event()
+
+/datum/ailment/proc/can_apply_to(var/obj/item/organ/_organ)
+	if(specific_organ_subtype && !istype(_organ, specific_organ_subtype))
+		return FALSE
+	if(affects_robotics && !BP_IS_PROSTHETIC(organ))
+		return FALSE
+	if(!affects_robotics && BP_IS_PROSTHETIC(organ))
+		return FALSE
```
afterthought2, line 39: I think `(!!foo != !!bar)` is more readable for xor, but this works.

code/modules/organs/ailments/_ailment.dm
```diff
@@ -0,0 +1,96 @@
...
+	return TRUE
+
+/datum/ailment/Destroy(force)
+	. = ..()
+	if(organ)
+		LAZYREMOVE(organ.ailments, src)
+		organ = null
+	deltimer(timer_id)
+	timer_id = null
```
afterthought2, line 49: Aren't these cleaned automatically? Even if not, all this should be above the `..()`.

MistakeNot4892: I don't know, this code is copied over from the original implementation.

Lohikar: They're not cleaned per-se, but nulling it out is more or less pointless - this is just a number that shouldn't be reused within a reasonable timespan.

afterthought2 (approved):

PR #1035, opened 2020-12-13
Robot gripper fixes

quardbreak:
Ports https://github.com/Baystation12/Baystation12/pull/29824

### DIFF
--- a/code/modules/mob/living/silicon/robot/drone/drone_items.dm
+++ b/code/modules/mob/living/silicon/robot/drone/drone_items.dm
@@ -28,6 +28,14 @@
 
 	var/obj/item/wrapped = null // Item currently being held.
 
+// Ensures any removed items are actually removed from the gripper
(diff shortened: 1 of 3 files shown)


code/modules/mob/living/silicon/robot/drone/drone_items.dm
```diff
@@ -28,6 +28,14 @@
 
 	var/obj/item/wrapped = null // Item currently being held.
 
+// Ensures any removed items are actually removed from the gripper
+/obj/item/forceMove(atom/dest)
```
MistakeNot4892, line 32: The rest of this PR is fine but this forceMove override specifically is not good. A better solution would be for the gripper to register to the moved/destroyed events of the wrapped item when it is set as the active item, and clear wrapped when the gripped item moves.

quardbreak:
Closing because it's not ready

PR #1038, opened 2020-12-14
Converting grippers to a hand slot for robots.

MistakeNot4892:
**TODO**
- [ ] Work out why put_in_hands isn't working.
- [ ] Make robots drop items.
- [ ] Test.


MistakeNot4892:
Going to revisit this at another point and convert all the robot inventory bullshit into held item slots.

PR #1041, opened 2020-12-14
Patches on voidsuits can now be removed

quardbreak:
Ports https://github.com/Baystation12/Baystation12/pull/29857


MistakeNot4892 (approved):
From discussion with Sierra this is actually a bit of a misnamed PR as the original changed approach midway through. Removing the `!damage` check allows breaches to be repaired properly even when the suit has been patched with duct tape.

PR #1042, opened 2020-12-14
Rag tweaks

SierraKomodo:
Port of https://github.com/Baystation12/Baystation12/pull/29880

Also added a global `has_danger_grab()` proc to mobs that mimics the check handcuffs used.

### DIFF
--- a/code/modules/mob/mob_helpers.dm
+++ b/code/modules/mob/mob_helpers.dm
@@ -35,6 +35,19 @@
 /mob/living/carbon/human/isMonkey()
 	return istype(species, /datum/species/monkey)
 
+
+/**
+ * Checks if the target has a grab from the user
+ */
+/mob/proc/has_danger_grab(mob/user)
+	if (user == src || istype(user, /mob/living/silicon/robot) || istype(user, /mob/living/bot))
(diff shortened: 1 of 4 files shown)


code/modules/mob/mob_helpers.dm
```diff
@@ -35,6 +35,19 @@
 /mob/living/carbon/human/isMonkey()
 	return istype(species, /datum/species/monkey)
 
+
+/**
+ * Checks if the target has a grab from the user
+ */
+/mob/proc/has_grab(mob/user)
```
MistakeNot4892: `has_upgraded_grab` would be more indicative I think.

Andrew-Fall: has_grab(user, grab_level)?

MistakeNot4892: That might be a better approach yeah, some flag-based system.

SierraKomodo: Updated the proc name to `has_danger_grab()` since the chain specifically checks the danger var.

MistakeNot4892 (approved):

PR #1047, opened 2020-12-16
Adds paperwork department goal archetype and a Tradeship example goal.

MistakeNot4892:
image


code/controllers/subsystems/goals.dm
```diff
@@ -16,7 +16,16 @@ SUBSYSTEM_DEF(goals)
...
-	var/list/ambitions =   list()
+	var/list/ambitions = list()
+	var/list/pending_goals = list()
+
+/datum/controller/subsystem/goals/fire(resumed)
+	for(var/datum/goal/goal in pending_goals)
+		if(goal.try_initialize())
+			pending_goals -= goal
+	if(!length(pending_goals))
+		flags |= SS_NO_FIRE
```
comma: this is never set again after this, wouldn't SS only fire until initial queue is cleared?

afterthought2: Correct method is one of `disable()` or `suspend()`; see how others do it.

code/controllers/subsystems/goals.dm
```diff
@@ -16,7 +16,16 @@ SUBSYSTEM_DEF(goals)
...
 		/datum/goal/money
 	)
-	var/list/ambitions =   list()
+	var/list/ambitions = list()
+	var/list/pending_goals = list()
+
+/datum/controller/subsystem/goals/fire(resumed)
+	for(var/datum/goal/goal in pending_goals)
+		if(goal.try_initialize())
+			pending_goals -= goal
```
afterthought2, line 35: Check tick, return, etc.

afterthought2 (approved):

Issue #1037, opened 2020-12-13
Intranet Relay Chat Client is broken


quardbreak:
#### Description of issue

Intranet Relay Chat Client is broken.
If you try open it, every user on same network will act like one user (like there's only one client).

#### Difference between expected and actual behavior

You must have your own client session.

#### Steps to reproduce

Take two PDA, download Intranet Relay Chat Client on them and open it.
Create channel and try perform the chat process, first on the first device, then on another device.

#### Specific information for locating

Dunno, it's Intranet Relay Chat Client

#### Length of time in which bug has been known to occur

Dunno.

#### Client version, Server revision & Game ID

513, nebula downstream with almost same code like on upstream.

#### Issue bingo

- [ ] Issue could be reproduced at least once
- [X] Issue could be reproduced by different players
- [ ] Issue could be reproduced in multiple rounds
- [X] Issue happened in a recent (less than 7 days ago) round
- [X] [Couldn't find an existing issue about this](https://github.com/NebulaSS13/Nebula/issues)

afterthought2:
Should be fixed in mentioned PR.

PR #1053, opened 2020-12-19
Fixes network bugs and adjusts program access

NataKilar:
Fixes a few bugs with networks and their related machinery, and adds an option for grant based program access
* Fixes mainframes going invisible after opening their panel once
* Fixes Network Locks not having their UI properly updated.
* Changes how the transfer files backend works. Now it by default transfers the file with an option for cloning the file instead. Beforehand it used store_file to add the datum to the file list of whatever was being copied to, without removing itself from the former location, which resulted in every program downloaded from a mainframe being one instance. This lead to issues like #1037, which is now fixed.
* Fixes a bug where running a program without proper access would break the computer until it was restarted.
* Added an option for program access to be determined by grants through the access controller. If/when full replacement of the access system with grants occurs, this should probably be reworked., but it works for now.

### DIFF
--- a/code/modules/modular_computers/file_system/program.dm
+++ b/code/modules/modular_computers/file_system/program.dm
@@ -91,7 +91,7 @@
 		if(network)
 			var/datum/extension/network_device/acl/access_controller = network.access_controller
 			if(access_controller)
-				accesses_to_check = access_controller.get_program_access(src.type)
+				accesses_to_check = access_controller.get_program_access(filename)
 				if(!length(accesses_to_check))
 					accesses_to_check = list(required_access)
 				else if(accesses_to_check[1] == "NONE")
--- a/code/modules/modular_computers/file_system/program.dm
+++ b/code/modules/modular_computers/file_system/program.dm
@@ -83,32 +83,43 @@
 // Check if the user can run program. Only humans can operate computer. Automatically called in run_program()
 // User has to wear their ID or have it inhand for ID Scan to work.
 // Can also be called manually, with optional parameter being access_to_check to scan the user's ID
-/datum/computer_file/program/proc/can_run(var/mob/living/user, var/loud = 0, var/access_to_check)
+/datum/computer_file/program/proc/can_run(var/mob/living/user, var/loud = 0, var/list/accesses_to_check, var/datum/computer_network/network)
 	if(!requires_access_to_run)
-		return 1
-	// Defaults to required_access
-	if(!access_to_check)
-		access_to_check = required_access
-	if(!access_to_check) // No required_access, allow it.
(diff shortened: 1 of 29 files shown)


code/modules/modular_computers/file_system/program.dm
```diff
@@ -83,32 +83,43 @@
...
+		return TRUE
+	// Defaults to network accecss, and then required_access
+	if(!length(accesses_to_check))
+		if(network)
+			var/datum/extension/network_device/acl/access_controller = network.access_controller
+			if(access_controller)
+				accesses_to_check = access_controller.get_program_access(filename)
+				if(!length(accesses_to_check))
+					accesses_to_check = list(required_access)
+				else if(accesses_to_check[1] == "NONE")
```
afterthought2: Really not a fan of dummy NONE access like this. Empty list should signify this. If you need to inherit default access some but not all of the time, add a bool and/or proc to the program itself and check/call that to get extra default accesses with a different inheritance structure.

NataKilar:
Adjusted program access/grants to use the normal empty list paradigm. Also went back and adjusted programs to use lists for their required_access var, and cleaned up a couple oddities (Command and Communications and Deck Management) programs who used unique access checks that could easily be replaced.

code/modules/modular_computers/file_system/programs/generic/deck_management.dm
```diff
@@ -7,6 +7,7 @@
 	filename = "deckmngr"
 	filedesc = "Deck Management"
 	nanomodule_path = /datum/nano_module/deck_management
+	required_access = list(access_cargo, access_bridge)
```
afterthought2: Note that this is expected to be AND; use list(list()) for OR (like the original).

code/modules/modular_computers/file_system/program.dm
```diff
@@ -83,32 +83,41 @@
...
-	if(access_to_check in I.access)
-		return 1
-	else if(loud)
-		to_chat(user, "<span class='notice'>\The [computer] flashes an \"Access Denied\" warning.</span>")
+			to_chat(user, SPAN_WARNING("The OS flashes an \"RFID Error - Unable to scan ID\" warning."))
+		return FALSE
+
+	for(var/access in I.access)
+		if(access in accesses_to_check)
+			return TRUE
```
afterthought2: Use the existing helper (`has_access`); this will correctly handle AND/OR usage in a way consistent with other access stuff.

afterthought2 (commented):
Note to self: the report access stuff should really be converted to the main system.

afterthought2 (approved):

PR #1057, opened 2020-12-22
Makes lighting bi-directional

Andrew-Fall:
Accidentally deleted branch, recreated
Makes lights bi-directional, removing last goto in process.
Uses an ugly macro due to this being very hot code.


Lohikar (dismissed):
As I've said before, I don't like this approach: the macro is ugly and unnecessary.

This can be done in about [13 lines](https://github.com/Aurorastation/Aurora.3/pull/3482/files#diff-ee2211ffbc46de88620fd0a108ed14c24c062826406b421ca1c67030dbcc072eR393) without macros.

afterthought2 (approved):
Holding on merge for possible alternative.

afterthought2:
Can be refactored or reverted as part of general lighting overhaul when that's ready, but feel it's fine for now.

PR #1058, opened 2020-12-23
teleportation modpack

ghost:
WIP maybe?... i dont feel like completing it though
adds a separate z-level for so-called subspace, too lazy to create any machinery so i think thats all
also adds a bunch of helpers and new /decl/teleport which will use this subspace for realistic (or not) teleportation
can contain shitcode and low-effort comments


mods/teleport/realm.dm
```diff
@@ -0,0 +1,160 @@
...
+		animate(src,alpha = lastalpha,time = pick(5,10))
+
+/mob/living/var/subspace_defense = 0 //time until subspace defense ends
+
+/turf/unsimulated/floor/subspace/proc/defense_check(var/atom/movable/O)
+	. = 0
+	if(isliving(O))
+		var/mob/living/L = O
+		var/mob/living/carbon/human/H = L
+		if(world.time < L.subspace_defense)               . = 1
```
MistakeNot4892, line 72:
````
. = world.time < L.subspace_defense || issilicon(L) || (istype(H) && H.get_pressure_weakness(0) < 0.5)
````

File: mods/teleport/realm.dm [hunk 2]
```diff
@@ -0,0 +1,160 @@
...
+	subspace_shift(A)
+	addtimer(CALLBACK(GLOBAL_PROC, .proc/subspace_shift, A), time)
+
+//Just throws things from one loc to another
+
+/proc/subspace_channel(var/turf/start, var/turf/end, var/range = 3)
+	var/list/success = list()
+	for(var/atom/movable/A in range(start,range))
+		A.throw_at(end,get_dist(start,end),80)
+	for(var/atom/movable/A in range(end,  range))
```
MistakeNot4892, line 126: Throwing is async so I'm not sure if this would actually work or not.

PR #1060, opened 2020-12-26
Mob traits

PsiOmegaDelta:
Main goal: Simplify having species in different mobs packs yet have, for example, metabolism work as intended without having to override effects individually.

Why traits are decls instead of strings:
Simplifies actually getting/setting all possible traits and valid levels - Useful if, for example, one want to make an UI to more easily allow admins to tweak individual mobs.

Main trait added: Metabolically Inert
* Currently only checked in `/affect_ingest()`
* Prevents stat adjustments, i.e. dizziness, drowsiness, etc., whether good or bad
* Does not prevent temperature adjustment
* Does not affect how much nourishment is derived from a material/reagent
* Does not affect how other materials are adjusted, i.e. capsaicin still removes frostoil

Granted to: Golems

Misc. traits:
* Malus - Animal Protein
  * Causes tox loss when animal protein is ingested, and reduces nutrition gained
* Malus - Ethanol
  * Increases the effective dose of ingested ethanol

**F.A.Q.**
Q: Do traits have to be used only with /decl/material?
A: No, traits can be used in many places where one would check for specific species


code/modules/reagents/chems/chems_drinks.dm
```diff
@@ -20,20 +20,29 @@
...
+	M.sleeping = max(0, M.sleeping + adj_sleepy)
+
 // Juices
 /decl/material/chem/drink/juice
 	fruit_descriptor = "sweet"
 
 /decl/material/liquid/drink/juice/affect_ingest(var/mob/living/carbon/human/M, var/alien, var/removed, var/datum/reagents/holder)
 	..()
+
+	if(M.HasTrait(/decl/trait/metabolically_inert))
```
PsiOmegaDelta, line 42: Unsure if metabolic inertness should affect immunity (which I assume is technically derived from simply having the juice in one's system)

code/modules/reagents/chems/chems_ethanol.dm
```diff
@@ -42,13 +42,25 @@
...
-
 	..()
+
 	M.adjust_nutrition(nutriment_factor * removed)
 	M.adjust_hydration(hydration_factor * removed)
+
+	if(adj_temp > 0 && M.bodytemperature < targ_temp) // 310 is the normal bodytemp. 310.055
+		M.bodytemperature = min(targ_temp, M.bodytemperature + (adj_temp * TEMPERATURE_DAMAGE_COEFFICIENT))
+	if(adj_temp < 0 && M.bodytemperature > targ_temp)
+		M.bodytemperature = min(targ_temp, M.bodytemperature - (adj_temp * TEMPERATURE_DAMAGE_COEFFICIENT))
```
PsiOmegaDelta: The temperature shifting from ethanol should perhaps come after the metabolic check, as I assume the temperature shift is caused by expanding blood vessels?
What about creatures without blood (vessels)? Should perhaps even be moved to `affect_blood()` instead

code/modules/reagents/chems/chems_ethanol.dm
```diff
@@ -157,11 +167,16 @@
 
 /decl/material/liquid/ethanol/coffee/affect_ingest(var/mob/living/carbon/M, var/alien, var/removed, var/datum/reagents/holder)
 	..()
+
+	if(M.bodytemperature > 310)
+		M.bodytemperature = max(310, M.bodytemperature - (5 * TEMPERATURE_DAMAGE_COEFFICIENT))
```
PsiOmegaDelta: Same as for pure ethanol I guess

code/modules/reagents/chems/chems_drinks.dm
comma: nah should be affected, it's a juice not antiseptic

code/modules/reagents/chems/chems_ethanol.dm
MistakeNot4892: This should go after the metabolism check, I agree. Also the nutriment/hydration adjustment should be blocked by not being metabolized.

PsiOmegaDelta:
Metabolic inertness is now checked before both nutritional and temperature adjustments
Reagents negating/reacting with each other is still handled before this check

code/datums/traits/traits.dm
```diff
@@ -0,0 +1,86 @@
...
+	var/list/traits
+
+/mob/living/proc/HasTrait(trait_type)
+	SHOULD_NOT_OVERRIDE(TRUE)
+	SHOULD_NOT_SLEEP(TRUE)
+	return (trait_type in GetTraits())
+
+/mob/living/proc/GetTraitLevel(trait_type)
+	SHOULD_NOT_SLEEP(TRUE)
+	var traits = GetTraits()
```
MistakeNot4892: :eye:

MistakeNot4892 (approved):

PR #1061, opened 2020-12-29
Ports tajaran and tritonians as modpacks

quardbreak:
Adds two bay species as modpacks.


mods/species/booster/_booster.dme
```diff
@@ -0,0 +1,15 @@
+#ifndef MODPACK_BOOSTER
+#define MODPACK_BOOSTER
+// BEGIN_INCLUDE
+#include "_booster.dm"
+#include "accessory.dm"
+#include "species.dm"
+// END_INCLUDE
+#endif
+// BEGIN_INTERNALS
```
MistakeNot4892: Need to prune this from here and the other .dmes.

mods/species/tajaran/clothing.dm
```diff
@@ -0,0 +1,145 @@
+#define TAJ_OBJ 'mods/species/tajaran/icons/clothing/obj.dmi'
+#define TAJ_MOB 'mods/species/tajaran/icons/clothing/onmob.dmi'
+
+//Clothing!//
+
+//Veils
+
+/obj/item/clothing/glasses/tajblind
```
MistakeNot4892: I'm not sold on porting all the visors, they're going to need to be converted to the Nebula single icon system and frankly I always hated them visually and with regards to how they were shoehorned into the lore to make Taj more distinctive. Possibly something for maintainer discussion.

mods/species/tajaran/culture.dm
```diff
@@ -0,0 +1,48 @@
+//Culture
+
+/decl/cultural_info/culture/tajaran
+	name = CULTURE_TAJARA
+	description = "The Tajaran are a species of furred mammalian bipeds hailing from the chilly planet of Ahdomai \
```
MistakeNot4892: Would like to generalize this if possible, down to just 'they're a species of cat dudes from a frozen world'.

mods/species/tritonian/species.dm
```diff
@@ -0,0 +1,28 @@
+/decl/species/human/tritonian
+	name = SPECIES_TRITONIAN
+	name_plural = "Tritonians"
+	description = "Of all human gene-adapts, the Tritonian genotype is probably the most wildly divergent from \
```
MistakeNot4892: As with Taj culture above, would like this to be generalized to 'A human-derived genotype designed for colonizing aquatic worlds' or something like that.

MistakeNot4892 (approved):

PR #1063, opened 2021-01-03
Add randomized light tones

SierraKomodo:
Ported from my PRs on Bay

https://github.com/Baystation12/Baystation12/pull/29924
https://github.com/Baystation12/Baystation12/pull/29942
https://github.com/Baystation12/Baystation12/pull/29958

dreamseeker_7unGsw2FwE

### DIFF
--- a/code/modules/power/lighting.dm
+++ b/code/modules/power/lighting.dm
@@ -479,6 +479,18 @@
 	var/b_colour = "#fffee0"
 	var/list/lighting_modes = list()
 	var/sound_on
+	var/random_tone = TRUE
+	var/static/list/random_tone_options = list(
+		"#fffee0",
+		"#e0fefe",
+		"#fefefe",
+	)
+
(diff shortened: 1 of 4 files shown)


code/modules/power/lighting.dm
```diff
@@ -479,6 +479,18 @@
 	var/b_colour = "#fffee0"
 	var/list/lighting_modes = list()
 	var/sound_on
+	var/random_tone = TRUE
+	var/list/random_tone_options = list(
```
afterthought2: Seems like it should be `/static`.

SierraKomodo: Eh, sure. There are no variants to this list right now.

afterthought2 (approved):

MistakeNot4892:
Not a huge fan of these as discussed but it's easy enough to override I suppose.

PR #1082, opened 2021-01-10
Ports Aurora's animated areas alarms

quardbreak:
Also returns a atmosalarm alarm


quardbreak:
@Lohikar You did these, would you like to review this?

code/game/area/areas.dm
```diff
@@ -28,6 +28,10 @@
 		power_environ = 0
 	power_change()		// all machines set to current power level, also updates lighting icon
 
+	icon = 'icons/turf/areas.dmi'
+	icon_state = "white"
+	blend_mode = BLEND_MULTIPLY
```
afterthought2, line 33: Why is this set here, not in the `/area` definition? Why does it make icon updates outside of `update_icon`?

afterthought2 (commented):
Needs playtest to see if it's tolerable.

code/game/area/areas.dm
Lohikar: This is where they should be, per the reasons discussed on discord.

MistakeNot4892 (approved):

PR #1085, opened 2021-01-10
Converting under items to single clothing icons, purging some states.

MistakeNot4892:
Killed off any items with missing/inconsistent states, also had to infer some inhand states for a bunch of uniforms. Someone can go and make new ones as needed if it's an issue.

**TODO**
- [x] Convert uniforms to single icons.
- [x] Convert rolldown and sleeve behavior to work with single icons.
- [x] ~~Readd 'female' states if discussion indicates they're wanted.~~
- [x] Reimplement removed uniforms (paramedic) if needed.
- [x] Reimplement deleted corp and ascent modpack uniforms.
- [x] Retest everything.


MistakeNot4892:
Ready for review.
````
ok I'm making the unilateral decision to keep them gone, but the code does support gendered states so if someone wants to go through and add them back they can feel free and we can have a discussion about it then
````

File: mods/content/government/away_sites/icarus/icarus.dm [hunk 1]
```diff
@@ -171,7 +171,6 @@
 	name = "expeditionary uniform"
 	desc = "An older model of the utility uniform of the SCG Expeditionary Corps. It has a patch on the left sleeve signifying the wearer served on the SEV Icarus."
 	icon_state = "blackutility_crew"
```
comma, line 173: is that supposed to be still set?

comma (approved):
that aside

mods/content/government/away_sites/icarus/icarus.dm
MistakeNot4892: No, but I have it fixed in another commit (as discussed) and would prefer not to mess with these commits at this point.
Once they're all merged I'll use the debug verb added in a more recent PR to do a fix pass.

deathride58 (approved):
Giving this a quick skim-over, the code that's actually been changed in this PR is pretty solid. Should be safe to merge

PR #1088, opened 2021-01-10
Adds a per-atom overlay cap to SSoverlays.

MistakeNot4892:
@Lohikar @afterthought2


Lohikar (commented):
Not a fan of doing it this way, personally.

Better would be to have add_overlays() check if the added overlay count would (or has) gone over the limit and `crash_with()`ing to get a stack trace to the cause.

PR #1101, opened 2021-01-12
Adds surgery tool selector from OnyxBay

quardbreak:
Everything seems work fine, but it can't perform surgery actions for some reason.


quardbreak:
Closing after discord discussion.

MistakeNot4892:
For reference discussion was:
- Making all the surgical items into one item means you can't preload three slots and do surgery that way.
- Code itself is reimplementing gripper attackby which is a bad pattern we need to get rid of, so generally isn't desirable itself.

PR #1102, opened 2021-01-12
Make windows and airlock windows paintable

SierraKomodo:
Ported from BS12

https://github.com/Baystation12/Baystation12/pull/29925
https://github.com/Baystation12/Baystation12/pull/29998


code/game/objects/structures/window.dm
```diff
@@ -245,7 +250,7 @@
 			playsound(src, 'sound/items/Welder.ogg', 80, 1)
 			construction_state = 0
 			set_anchored(0)
-	else
+	else if (!istype(W, /obj/item/device/paint_sprayer))
```
quardbreak: `else if (!istype(W, /obj/item/paint_sprayer))`
We don't have `/device` type now.

MistakeNot4892 (approved):

PR #1106, opened 2021-01-13
Added tesla links to several turbine-related machines.

### LINKED ISSUE #1097: TEG circulators have no teslalinks
#### Description of issue
They are created without power source (so theres no animations)

#### Issue bingo

MistakeNot4892:
Should fix #1097.


afterthought2 (approved):
It seems likely it's more correct to make them not check power in `update_icon`, but doesn't matter that much.

PR #1114, opened 2021-01-13
Rearranges planetary atmos generation

comma:
Now we always know temperature /before/ we pick gasses.
Makes sure that nothing's going to condense etc.
Moves some things into overridable procs - temperature generation, list of mandatory gases, desired pressure.
Another unrelated change - chlorine planets are now cold rather than boiling hot. Was weird how chlorine puddles were a thing waaay above boiling point.

### DIFF
--- a/code/modules/overmap/exoplanets/exoplanet_atmosphere.dm
+++ b/code/modules/overmap/exoplanets/exoplanet_atmosphere.dm
@@ -1,50 +1,94 @@
-
-
 /obj/effect/overmap/visitable/sector/exoplanet/proc/generate_atmosphere()
 	atmosphere = new
+	var/target_temp = get_target_temperature()
+
+	//Make sure temperature can't damage people on casual planets
+	if(habitability_class <= HABITABILITY_OKAY)
+		var/decl/species/S = get_species_by_key(GLOB.using_map.default_species)
+		target_temp = Clamp(target_temp, S.cold_level_1 + rand(1,5), S.heat_level_1 - rand(1,5))
+
+	atmosphere.temperature = target_temp
+
+	//Skip fun gas gen for perfect terran worlds
 	if(habitability_class == HABITABILITY_IDEAL)
-		atmosphere.adjust_gas(/decl/material/gas/oxygen, MOLES_O2STANDARD, 0)
+		atmosphere.adjust_gas(/decl/material/gas/oxygen, MOLES_O2STANDARD)
 		atmosphere.adjust_gas(/decl/material/gas/nitrogen, MOLES_N2STANDARD)
-	else //let the fuckery commence
-		var/list/newgases = subtypesof(/decl/material/gas)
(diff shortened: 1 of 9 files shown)


code/modules/overmap/exoplanets/exoplanet_atmosphere.dm
```diff
@@ -1,50 +1,94 @@
...
+	var/list/newgases = subtypesof(/decl/material/gas)
+	if(prob(50)) //alium gas should be slightly less common than mundane shit
+		newgases -= /decl/material/gas/alien
+
+	// Prune gases that won't stay gaseous
+	for(var/g in newgases)
+		var/decl/material/mat = decls_repository.get_decl(g)
+		if(mat.gas_flags & badflag)
+			newgases -= g
+		if(mat.gas_condensation_point && mat.gas_condensation_point < atmosphere.temperature)
```
MistakeNot4892: Nit: here and below should be `<=` and `>=`.

code/modules/overmap/exoplanets/exoplanet_atmosphere.dm
```diff
@@ -1,50 +1,94 @@
...
+		if(badflag)
+			for(var/g in newgases)
+				var/decl/material/mat = decls_repository.get_decl(g)
+				if(mat.gas_flags & badflag)
+					newgases -= g
+		var/ng = pick_n_take(newgases)	//pick a gas
+
+		// Make sure atmosphere is not flammable
+		var/decl/material/mat = decls_repository.get_decl(ng)
+		if(mat.gas_flags & XGM_GAS_OXIDIZER)
```
MistakeNot4892, line 61: Isn't it still possible to get an oxidizer first, then have it pick a fuel gas?

code/modules/overmap/exoplanets/exoplanet_flora.dm
```diff
@@ -44,7 +44,7 @@
 	S.set_trait(TRAIT_HIGHKPA_TOLERANCE,   atmosphere.return_pressure() + rand(5,50),500,110)
 	if(S.exude_gasses)
 		S.exude_gasses -= badgas
-	if(atmosphere)
+	if(atmosphere && length(atmosphere.gas))
```
MistakeNot4892, line 47: Nit: `length(atmosphere.gas)` would also null check it.

comma: wouldn't it runtime because null.gas?

MistakeNot4892: Oh I'm dumb and misread

code/modules/overmap/exoplanets/exoplanet_atmosphere.dm
comma: Fixed

MistakeNot4892 (approved):

PR #1116, opened 2021-01-13
Adds tooltips and descriptions for action buttons

quardbreak:
image

### DIFF
--- a/code/modules/mob/living/living_defense.dm
+++ b/code/modules/mob/living/living_defense.dm
@@ -312,6 +312,7 @@
 			if(!I.action)
 				I.action = new I.default_action_type
 			I.action.name = I.action_button_name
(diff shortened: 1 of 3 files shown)


MistakeNot4892 (dismissed):

code/modules/mob/living/living_defense.dm
```diff
@@ -353,6 +354,7 @@
 		B.UpdateIcon()
 
 		B.SetName(A.UpdateName())
+		B.SetName(A.UpdateDesc())
```
comma: why's it setting name to desc immediately after setting name?

quardbreak: Woops

MistakeNot4892 (approved):

comma (approved):

PR #1119, opened 2021-01-14
Fixes for the previous clothing conversion PRs.

MistakeNot4892:
This brings the non-ignored issues reported from the debug verb to zero. Just accessories and rigs after this.


comma (dismissed):

code/modules/clothing/under/_under.dm
```diff
@@ -1,5 +1,6 @@
 /obj/item/clothing/under
 	name = "under"
+	icon = 'icons/clothing/under/jumpsuits/jumpsuit.dmi'
```
comma, line 3: i'd suggest making it some black-pink checkerboard jumpsuit so you can see things that don't have a REAL icon set.

MistakeNot4892: As mentioned on Discord, I will do this for `/obj/item/clothing` when the conversions are finished, that way I I can put a state for every single slot in there.

comma (approved):

PR #1126, opened 2021-01-15
Added neocorvid modpack.

MistakeNot4892:
https://cdn.discordapp.com/attachments/492799900145418270/789288699597815838/unknown.png

Neocorvids are a playable species who are smaller than humans, are quite fragile, and can't wear voidsuits.
Adding them as a less problematic alternative to Resomi and as an example of a small species that overrides some bodytype restrictions.


nebula.dme
```diff
@@ -3272,5 +3272,6 @@
 #include "maps\~mapsystem\maps_unit_testing.dm"
 #include "maps\~unit_tests\unit_testing.dm"
 #include "mods\_modpack.dm"
+#include "mods\species\neocorvids\clothing.dm"
```
quardbreak: 👀

MistakeNot4892: dammit VSC

afterthought2 (approved):

PR #1132, opened 2021-01-15
Exterior turf icon tweaks

comma:
Added 'still water' icon. It's same but has less movement, which looked odd on puddles like chlorine or tar.
Glassed sand is now sand-colored rather than ice.
Changed icon for scorched ground to a darker more ash-like one.
Also adds corners to grass tiles


MistakeNot4892 (approved):
Smh this is just low-energy water.

PR #1133, opened 2021-01-15
Adds water smoothing and coastal turfs to exoplanets

comma:
Adds a second pass on generated maps to eliminate 1 tile water puddles and 1 tile islands inside water.
You can also set type of turf to be placed around bodies of water, currently used by grass planets.
image


MistakeNot4892:
Man that jungle grass turf needs edge states pretty bad.

MistakeNot4892 (approved):
I feel like a case could be made for generalizing this so maps remove -all- single-turfs without a neighbor with the same value, but that's a bit out of scope.

PR #1137, opened 2021-01-15
Deletes the Ascent away site.

MistakeNot4892:
I'll revisit an away site for the Ascent when I get around to updating hunter mode, for the meantime this map doesn't work and perpetuates the problem with submaps anyway.


comma (approved):

Lohikar:
rip

PR #1142, opened 2021-01-16
Improve graphs

PsiOmegaDelta:
Can now handle multiple connect/disconnect calls on the same node per tick
Can now handle moving nodes


MistakeNot4892 (approved):
To be honest a lot of the repercussions of this system are beyond me, but as it isn't currently in active use I don't see a problem with merging this in.

PsiOmegaDelta:
Well, if nothing else there are at least unit tests.

PR #1148, opened 2021-01-16
Make initialize utilize RETURN_TYPE

PsiOmegaDelta:
Compiles locally, all the testing we need really.

### DIFF
--- a/code/__defines/subsystems.dm
+++ b/code/__defines/subsystems.dm
@@ -3,9 +3,9 @@
 #define INITIALIZATION_INNEW_MAPLOAD  2	//New should call Initialize(TRUE)
 #define INITIALIZATION_INNEW_REGULAR  3	//New should call Initialize(FALSE)
 
-#define INITIALIZE_HINT_NORMAL   0  //Nothing happens
-#define INITIALIZE_HINT_LATELOAD 1  //Call LateInitialize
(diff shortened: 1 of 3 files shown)


Lohikar:
I really don't like this solution. Having something as low-level as atom init hints depend on hooks and the decl system feels pointlessly fragile.

code/__defines/subsystems.dm
```diff
@@ -3,9 +3,9 @@
 #define INITIALIZATION_INNEW_MAPLOAD  2	//New should call Initialize(TRUE)
 #define INITIALIZATION_INNEW_REGULAR  3	//New should call Initialize(FALSE)
 
-#define INITIALIZE_HINT_NORMAL   0  //Nothing happens
-#define INITIALIZE_HINT_LATELOAD 1  //Call LateInitialize
-#define INITIALIZE_HINT_QDEL     2  //Call qdel on the atom
+var/INITIALIZE_HINT_NORMAL   = new/datum/init_hint/normal   //Nothing happens
```
MistakeNot4892: Nitpick, but `new /foo` (with a space) makes it easier to search for uses of `new`.

PsiOmegaDelta:
It would appear that `RETURN_TYPE()` is mainly for the linter to do know when you do something like `for(var/x in some_proc())`, which then let it determine if you're trying to loop over a list or not, rather than enforce that the proc and overrides actually returns something of that type.

PR #1151, opened 2021-01-16
Fixes #611

### LINKED ISSUE #611: Sketchy internet medical merchant sells sketchy unspecific items
#### Description of issue
Namely concerned about blood bags sold by the merchant you can hail on a cargo computer and receive goods by teleporter, though lots of the goods sold are very non-specific.  Medical merchant sells blood bags, and, fortunately empty blood bags are obvious because they're cheap, but the more expensive blood bags, you never know what blood type you'll get.  Then there's blister packs which always seem to have painkillers but I've only bought it a couple times, but it'd be good to know what you're getting if there's alternate medications in blister packs.

expected: merchants tell you what you're buying before you have it.
actual: merchants sell surprise blood

#### Steps to reproduce
buy blood from online merchants.  Wonder why you're buying important medical supplies from shady dealers.

the merchant is available by 

/obj/machinery/computer/modular/preset/merchant
--
and money can be input by

/obj/machinery/merchant_pad
--

a couple weeks

Client Version: 512
Server Revision: 49687b99e2f79a4dd8a2030538ec9e5dff91a902 - dev - 2020-06-09
Game ID: b7z-dHQj

#### Issue bingo

MistakeNot4892:


comma (approved):

code/controllers/subsystems/trade.dm
```diff
@@ -2,16 +2,21 @@ SUBSYSTEM_DEF(trade)
...
 	var/max_traders = 10
 
 /datum/controller/subsystem/trade/Initialize()
 	. = ..()
-	for(var/i in 1 to rand(1,3))
-		generate_trader(1)
+
+	for(var/trader_type in GLOB.using_map.get_initial_traders())
+		traders += new trader_type
+	var/total_traders = rand(1,3) - length(traders)
```
comma, line 16: only unsure on this limitation, because wouldn't this make trader list basically static for tradeship (cause it starts with 4 traders)

MistakeNot4892: This is just the starting traders, my understanding is that they cycle.

afterthought2: They do come and go. On bay it's calibrated so that they stay a while, and there are usually close to the maximum number always present; can recalibrate if you want.

PR #1152, opened 2021-01-17
Porting obfuscated medication from Europa.

MistakeNot4892:
image

🆑 
add: When the bigpharma pack is ticked, pre-packaged medication (bottles, pills, syringes) that is not inside a first aid kit or emergency pouch will generate a random brand name and only tell you the actual reagent if you have a level or two in Chemistry.
/🆑


MistakeNot4892:
Guess we need to have a conversation about Unicode again.
@Lohikar 
@afterthought2
@eckff

Lohikar:
Issue with Unicode was resolved by another PR (per discord discussion), rebase and this should be good.

comma:
skill threshold seems a bit steep, especially considering it's chemistry only.
I'd suggest keeping it at Basic, and maybe letting either Chemistry or Medicine be checked.

MistakeNot4892:
Done.

comma (approved):

PR #1155, opened 2021-01-17
Fix compile on 512

Lohikar:


MistakeNot4892 (approved):
Clearly a critical fix

PR #1159, opened 2021-01-17
Generalizes one tile smoothing from exoplanets to all noise maps

comma:
If parameter is set, single tiles without buddies in orthogonal directions are going to get squashed.
I went for orthogonal because allowing diagonals leaves odd configurations that don't look nice.

### DIFF
--- a/code/modules/random_map/noise/noise.dm
+++ b/code/modules/random_map/noise/noise.dm
@@ -10,6 +10,7 @@
 	var/cell_base                   // Set in New()
 	var/initial_cell_range          // Set in New()
 	var/smoothing_iterations = 0
+	var/smooth_single_tiles			// Single turfs of different value are not allowed
 
 /datum/random_map/noise/New()
 	initial_cell_range = cell_range/5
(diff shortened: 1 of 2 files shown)


code/modules/random_map/noise/noise.dm
```diff
@@ -167,4 +168,33 @@
...
+				for(var/cell in neighbors)
+					if(noise2value(map[cell]) == noise2value(map[mapcell]))
+						buddies |= cell
+				if(!length(buddies))
+					map[mapcell] = map[pick(neighbors)]
+
+/datum/random_map/noise/proc/get_neighbors(x, y, include_diagonals)
+	. = list()
+	if(!include_diagonals)
+		for(var/list/offset in list(list(-1, 0), list(1, 0), list(0, 1), list(0,-1)))
```
Lohikar: How hot is this proc? Might be good to break this into a static local var just so you don't need to init 5 lists every time.

code/modules/random_map/noise/noise.dm
```diff
@@ -167,4 +168,33 @@
...
+			var/tmp_cell = get_map_cell(x+offset[1],y+offset[2])
+			if(tmp_cell)
+				. += tmp_cell
+	else
+		var/list/offsets = list(-1,0,1)
```
Lohikar: This should be static (or even just a for-in loop), I don't see it being modified anywhere.

comma: Pushed

code/modules/random_map/noise/noise.dm
comma: Done

MistakeNot4892 (dismissed):

MistakeNot4892:
Conflicts.

MistakeNot4892 (approved):

PR #1165, opened 2021-01-17
Fixes special role prefs.

MistakeNot4892:


afterthought2 (approved):

Lohikar:
Will this affect existing antag bans?

MistakeNot4892:
We don't have any existing antag bans.
Even so, no, the jobban proc converts from /decl/special_role to role.name anyway.

PR #1171, opened 2021-01-18
Reworks persistent book Destroy().

MistakeNot4892:
:cl:
tweak: Hurling a book into space will destroy it.
tweak: Persistent books will respawn in a random bookcase when destroyed.
/:cl:

This isn't a perfect solution to the previous issues, as you can still have a situation where all the bookshelves got destroyed resulting in the persistent books all getting wiped, but at least now it actually works to an extent.

### DIFF
--- a/code/game/objects/items/books/_book.dm
+++ b/code/game/objects/items/books/_book.dm
@@ -4,7 +4,7 @@
 	icon_state = "book"
 	throw_speed = 1
 	throw_range = 5
-	w_class = ITEM_SIZE_NORMAL		 //upped to three because books are, y'know, pretty big. (and you could hide them inside eachother recursively forever)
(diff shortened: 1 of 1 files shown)


code/game/objects/items/books/_book.dm
```diff
@@ -27,24 +27,23 @@
...
-		backup_book.dat =                dat
-		backup_book.author =             author
-		backup_book.title =              title
-		backup_book.last_modified_ckey = last_modified_ckey
-		backup_book.unique =             TRUE
-		backup_book.forceMove(null)
-		backup_book.SetName(backup_book.title)
+/obj/item/book/proc/OnMove()
+	set waitfor = 0
+	if(istype(loc, /turf/space))
```
quardbreak: maybe `isspaceturf(loc)`?

PR #1175, opened 2021-01-18
Makes APC visuals actually care about broken components

comma:
Not drawing screen overlay / light if screen is broken
Not  reporting power from cell if cell component is broken
Properly reporting no terminal connection in UI if component is broken

Turned out the actual meat of interactions already worked fine, just all UI / visuals didn't care.
 

## Changelog
:cl:
tweak: APCs now don't draw screen overlay / glow if screen component is destroyed
tweak: APCs now don't report cell power in UI etc if battery backup is destroyed
tweak: APCs now show that external power is not coming in if terminal component is destroyed
/:cl:

### DIFF
--- a/code/game/machinery/_machines_base/machinery.dm
+++ b/code/game/machinery/_machines_base/machinery.dm
@@ -429,9 +429,9 @@ Class Procs:
 	if(. && !CanFluidPass())
 		fluid_update()
 
-/obj/machinery/get_cell()
+/obj/machinery/get_cell(var/functional_only = TRUE)
 	var/obj/item/stock_parts/power/battery/battery = get_component_of_type(/obj/item/stock_parts/power/battery)
-	if(battery)
+	if(battery && (!functional_only || battery.is_functional()))
 		return battery.get_cell()
 
 /obj/machinery/building_cost()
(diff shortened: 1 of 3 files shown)


code/game/machinery/_machines_base/machinery.dm
```diff
@@ -429,9 +429,9 @@ Class Procs:
 	if(. && !CanFluidPass())
 		fluid_update()
 
-/obj/machinery/get_cell()
+/obj/machinery/get_cell(var/functional_only)
 	var/obj/item/stock_parts/power/battery/battery = get_component_of_type(/obj/item/stock_parts/power/battery)
-	if(battery)
+	if(battery && (!functional_only | battery.is_functional()))
```
MistakeNot4892: `||`

code/game/machinery/_machines_base/machinery.dm
```diff
@@ -429,9 +429,9 @@ Class Procs:
 	if(. && !CanFluidPass())
 		fluid_update()
 
-/obj/machinery/get_cell()
+/obj/machinery/get_cell(var/functional_only)
```
afterthought2: Should consider making `TRUE` the default here. Also, please make this and `terminal` use the same arg name.

comma: Did both

code/game/machinery/_machines_base/machinery.dm
comma: Fixed

afterthought2 (approved):

PR #1182, opened 2021-01-19
Allow to override target turf and floor type for mountains and exopla…

quardbreak:
…nets noise

Found out what I can't fill my map with unsimulated mask to perform planet generation. So... I did this to able override.
This removes hard type set and allow to set custom target. @comma, you probably want to discuss this with me.

### DIFF
--- a/code/modules/overmap/exoplanets/random_map.dm
+++ b/code/modules/overmap/exoplanets/random_map.dm
@@ -2,6 +2,7 @@
 /datum/random_map/noise/exoplanet
 	descriptor = "exoplanet"
 	smoothing_iterations = 1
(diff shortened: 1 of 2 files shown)


code/modules/overmap/exoplanets/random_map.dm
```diff
@@ -2,6 +2,7 @@
 /datum/random_map/noise/exoplanet
 	descriptor = "exoplanet"
 	smoothing_iterations = 1
+	target_turf_type = null
```
comma, line 5: it's alreayd null on base type isn't it

comma (approved):

PR #1189, opened 2021-01-21
Adding an intermediary trade hub structure for localized traders.

MistakeNot4892:
Trade hubs can be global (by default, each map has one global trade hub) or based on overmap objects under the ship's current location.

image

Nothing uses the overmap traders yet, I want to give them a proper name gen and tie them to sites or resources on the overmap in a later PR.


MistakeNot4892:
In a fit of madness I decided to tidy the old station name lists a bit and use them for tradepost namegen.

image

code/datums/trading/_trade_hub.dm
```diff
@@ -0,0 +1,75 @@
...
+	traders += new trader_type(src)
+
+/datum/trade_hub/Process(var/resumed)
+	if(!resumed)
+		processing_traders = traders.Copy()
+		t_ind = 1
+	while(t_ind <= processing_traders.len)
+		var/datum/trader/trader = processing_traders[t_ind++]
+		trader.tick()
+		// tick check
```
comma: i don't think this tick check is working

MistakeNot4892: But it says tick check right there!

comma (dismissed):

comma (approved):

PR #1210, opened 2021-01-25
Clothing Commits 2 Electric Boogalo

thestalker12:
## Changelog
rscadd: Clothing to the textile machine 
/:cl:

 Starting with Hard Hats, not ready for merge

### DIFF
--- a/code/modules/clothing/head/_head.dm
+++ b/code/modules/clothing/head/_head.dm
@@ -12,7 +12,9 @@
 	var/light_applied
 	var/brightness_on
 	var/on = 0
-
+	origin_tech = "{'materials':1,'engineering':1}"
+	matter = /decl/material/solid/cloth
+	
 /obj/item/clothing/head/equipped(var/mob/user, var/slot)
 	light_overlay_image = null
 	..(user, slot)
--- a/code/modules/clothing/head/_head.dm
+++ b/code/modules/clothing/head/_head.dm
@@ -13,7 +13,7 @@
 	var/brightness_on
 	var/on = 0
 	origin_tech = "{'materials':1,'engineering':1}"
-	matter = /decl/material/solid/cloth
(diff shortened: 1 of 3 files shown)


code/modules/clothing/head/hardhat.dm
```diff
@@ -21,6 +21,9 @@
 	heat_protection = SLOT_HEAD
 	max_heat_protection_temperature = FIRE_HELMET_MAX_HEAT_PROTECTION_TEMPERATURE
 	max_pressure_protection = FIRESUIT_MAX_PRESSURE
+	material = /decl/material/solid/metal/steel
+	matter = list(/decl/material/solid/metal/plasteel = MATTER_AMOUNT_TRACE)
```
comma: plasteel seems a bit too costly for something so wide-spread, and it doesn't even have that good armor, steel-tier values.
heck I'd even go for plastic with steel reinforcment.

code/modules/clothing/head/_head.dm
```diff
@@ -12,7 +12,9 @@
 	var/light_applied
 	var/brightness_on
 	var/on = 0
-
+	origin_tech = "{'materials':1,'engineering':1}"
+	material = /decl/material/solid/cloth
```
MistakeNot4892: `origin_tech` and `material` are already set to these values on `/obj/item/clothing`, they don't need to be set again.

code/modules/clothing/head/hardhat.dm
thestalker12: Ill fuck with this later, Truthfully I don't have the energy right now given I have classes starting in like 4 days

code/modules/clothing/head/_head.dm
thestalker12: Ill changed this out later

MistakeNot4892 (approved):

PR #1211, opened 2021-01-26
Converted debug verbs for clothing into unit tests.

MistakeNot4892:
Here you go @PsiOmegaDelta.


comma (approved):

code/unit_tests/clothing.dm
```diff
@@ -0,0 +1,72 @@
...
+		var/res = ""
+		var/obj/item/rig/rig = new rigtype
+		if(initial(rig.gloves) == null)
+			res = "[res] - null initial gloves"
+		if(initial(rig.chest) == null)
+			res = "[res] - null initial chest"
+		if(initial(rig.helmet) == null)
+			res = "[res] - null initial helmet"
+		if(initial(rig.boots) == null)
+			res = "[res] - null initial boots"
```
PsiOmegaDelta, line 16: A minor issue as such but if there are multiple null components then you'll only get the last one (so would have to run the test multiple times to catch them all)

PsiOmegaDelta: Nevermind me, the `res` is "recursive"

MistakeNot4892: Yeah it's a bit unclear.

PR #1223, opened 2021-01-28
Replaces blindness overlays with a render target based overlay.

MistakeNot4892:
Blindness now uses a black backdrop and an alpha filter that clips a circle out of it. Should be able to scale indefinitely now to meet any screen size/view. In theory this could be used for a vision cone down the track.

Many thanks to @Lohikar and @deathride58 for their input, would have been flailing in the dark for another six hours without your help.

**TODO**
- [x] Basic blindness system.
- [ ] Work out why glasses tint is busted.
- [ ] Convert eye damage and nearsightedness to scaling values for vision radius and quality.
- [ ] Try to fix the broken alpha/transform animations.


MistakeNot4892:
Alpha filters are apparently broken so this might need to wait for 514.

MistakeNot4892:
Will reopen this when I get back to it.

PR #1146, opened 2021-01-16
Pixel projectiles rebase.

MistakeNot4892:
:cl:
tweak: Firearms now use pixel-precise projectile iteration, ported from Aurora.
/:cl:

Taking over #509 as @Andromeda-K22 is beset by the vagracies of life. 
**TODO:**
- [x] Fix linter complaints.
- [x] Fix ballistics muzzle flashes.
- [x] Clean up port.
- [x] Add lights to the full span of laser projectiles.

PR #1230, opened 2021-01-28
True autofire mechanic added.

MistakeNot4892:
Pew pew pew.

https://user-images.githubusercontent.com/2468979/106209867-e7f17e80-6219-11eb-8aa0-210277b97405.mp4

### DIFF
--- a/code/_onclick/MouseDrag.dm
+++ b/code/_onclick/MouseDrag.dm
@@ -1,16 +1,23 @@
 //If we intercept it return true else return false
-/atom/proc/RelayMouseDrag(src_object, over_object, src_location, over_location, src_control, over_control, params, var/mob/user)
+/atom/proc/RelayMouseDrag(atom/src_object, atom/over_object, src_location, over_location, src_control, over_control, params, var/mob/user)
 	return FALSE
 
-/mob/proc/OnMouseDrag(src_object, over_object, src_location, over_location, src_control, over_control, params)
+/mob/proc/OnMouseDrag(atom/src_object, atom/over_object, src_location, over_location, src_control, over_control, params)
 	if(istype(loc, /atom))
 		var/atom/A = loc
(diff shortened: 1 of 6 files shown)


Lohikar:
#1146 should go in before this one does.

Lohikar:
#1146 is merged.

MistakeNot4892:
Bit stumped by why it's firing so slowly, about five times a second instead of ten. System works generally though.

MistakeNot4892:
Not sure what this is waiting on, code is ready to go.

MistakeNot4892:
Pls review.

code/_onclick/MouseDrag.dm
```diff
@@ -1,16 +1,23 @@
...
+	var/obj/item/gun/gun = get_active_hand()
+	if(a_intent == I_HURT && istype(over_object) && (isturf(over_object) || isturf(over_object.loc)) && !incapacitated() && istype(gun))
+		gun.set_autofire(over_object, src)
+
+/mob/proc/OnMouseDown(atom/object, location, control, params)
+	var/obj/item/gun/gun = get_active_hand()
+	if(a_intent == I_HURT && istype(object) && (isturf(object) || isturf(object.loc)) && !incapacitated() && istype(gun))
+		gun.set_autofire(object, src)
+
+/mob/proc/OnMouseUp(atom/object, location, control, params)
```
afterthought2, line 20: I'm not technically competent enough to offer meaningful feedback here (which is why I was avoiding this PR) but I believe you need to do similar stuff in `client/Destroy`, in the mob client transfer helper, and possibly check stuff in `Life` (as client disconnection doesn't reliably trigger anything). If you already do this, or you feel it's not necessary, then can discuss.

MistakeNot4892: As discussed on Discord, the autofire process checks for the mob being destroyed, the client being null, or the weapon not being in the firer's contents, and breaks out early if so.

Lohikar (approved):

PR #1231, opened 2021-01-28
Removes gun's empty click message output

quardbreak:


MistakeNot4892:
Redundant with the other fix in.

PR #1238, opened 2021-01-29
Requires mob to be on same tile as rollerbed to buckle

comma:
Mostly to prevent annoying self-buckle when trying to fold it. How /do/ you even fold it again?..
Idea stolen from bay PR with same goal, but this version still lets you buckle self, just have to be on top of the roller.

## Changelog
:cl:
tweak: Need to be on same tile as rollerbed to buckle to it.
/:cl:


SierraKomodo:
As a note, the bay method did still allow buckling yourself using the normal drag self to bed method, even when adjacent instead of on the same tile - Just like all other bed interactions.

MistakeNot4892 (approved):

PR #1240, opened 2021-01-30
mob_can_equip() cleanup pass. Fixes issues with gloves/magboots.

MistakeNot4892:
Gloves and magboots should stop complaining about not being equippable to your inhands now.
This boilerplate really needs to be generalized to /clothing level or something, or ideally replaced with some kind of inventory slot stacking mechanic.

### DIFF
--- a/code/modules/clothing/shoes/magboots.dm
+++ b/code/modules/clothing/shoes/magboots.dm
@@ -6,19 +6,18 @@
 	icon = 'icons/clothing/feet/magboots.dmi'
 	bodytype_restricted = null
 	force = 3
-	overshoes = 1
+	can_fit_under_magboots = FALSE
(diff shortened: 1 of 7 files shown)


code/modules/clothing/shoes/magboots.dm
```diff
@@ -61,49 +60,49 @@
...
 	..()
-	if(!wearer)
-		return
-
-	var/mob/living/carbon/human/H = wearer
-	if(shoes && istype(H))
-		if(!H.equip_to_slot_if_possible(shoes, slot_shoes_str))
-			shoes.dropInto(loc)
-		src.shoes = null
-	wearer.update_floating()
```
comma, line 105: not gonna update floating in dropped()?

comma (approved):

PR #1241, opened 2021-01-30
Optimizing mech drills a little bit.

### LINKED ISSUE #1200: Mining and lags
#### Description of issue
Mining with mech cause a big server lag

#### Steps to reproduce
1)take a mech, powerloader for example
2)take an ore box with hidraulic clamp
3)start mining
Without orebox mimng didnt cause lags

#### Issue bingo

MistakeNot4892:
Appears to fix #1200, or at least I couldn't see any lag when testing.

### DIFF
--- a/code/modules/mechs/equipment/utility.dm
+++ b/code/modules/mechs/equipment/utility.dm
@@ -411,57 +411,50 @@
 		//Better materials = faster drill!
 		var/delay = max(5, 20 - drill_head.material.brute_armor)
 		owner.setClickCooldown(delay) //Don't spamclick!
-		if(do_after(owner, delay, target) && drill_head)
-			if(src == owner.selected_system)
-				if(drill_head.durability <= 0)
-					drill_head.shatter()
-					drill_head = null
-					return
 
-				if(istype(target, /turf/exterior/wall))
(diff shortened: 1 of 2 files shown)


code/modules/mechs/equipment/utility.dm
```diff
@@ -411,57 +411,50 @@
...
+
+		if(!length(owner.hardpoints) || !length(ore_products))
+			return TRUE
+		for(var/hardpoint in owner.hardpoints)
+			//clamps work, but anythin that contains an ore crate internally is valid
+			var/obj/structure/ore_box/ore_box = locate() in owner.hardpoints[hardpoint]
+			if(ore_box)
+				for(var/obj/item/ore/ore in ore_products)
+					if(get_dir(owner,ore) & owner.dir)
+						ore.forceMove(ore_box)
```
afterthought2, line 456: What happens if there's no ore box? Do they just get destroyed?

MistakeNot4892: No, they get left on the ground to be picked up manually.

afterthought2 (dismissed):
Oh, missed they were spawned on the turf.

comma (approved):

PR #1243, opened 2021-01-30
Slime rewrite.

MistakeNot4892:
Converts most slime stuff to use /datum/ai and a series of /decls. Also repaths them to /mob/living/slime as they don't actually use 99% of the carbon logic.

**Stuff to look into after this PR:**
- Slimes will attack a mob that another slime is feeding on. Very petty.
- Slime HP seems to be too high and a single monke is enough to get an adult and one of its children all the way through a lifecycle.


code/modules/reagents/reagent_containers/syringes.dm
```diff
@@ -127,7 +127,7 @@
 			to_chat(user, "<span class='notice'>There is already a blood sample in this syringe.</span>")
 			return
 		if(istype(target, /mob/living/carbon))
-			if(istype(target, /mob/living/carbon/slime))
+			if(istype(target, /mob/living/slime))
```
comma: this will never be reached cause carbon check above

code/modules/mob/living/carbon/xenobiological/colours/colour_metal.dm
```diff
@@ -0,0 +1,19 @@
...
+		/decl/slime_colour/gold,
+		/decl/slime_colour/gold
+	)
+	baby_icon =    'icons/mob/slimes/slime_baby_metal.dmi'
+	adult_icon =   'icons/mob/slimes/slime_adult_metal.dmi'
+	extract_icon = 'icons/mob/slimes/slime_extract_metal.dmi'
+	reaction_strings = list(/decl/material/solid/metal/uranium = "Synthesises steel and plasteel.")
+
+/decl/slime_colour/metal/handle_uranium_reaction(var/datum/reagents/holder)
+	var/obj/item/stack/material/steel/M = new (get_turf(holder.my_atom))
```
comma: can do new(loc,amount) for stacks instead of manually changing var, will handle icon updating etc that way

code/modules/mob/living/carbon/xenobiological/colours/colour_quantum.dm
```diff
@@ -0,0 +1,17 @@
...
+	extract_icon =   'icons/mob/slimes/slime_extract_quantum.dmi'
+	reaction_sound = 'sound/effects/teleport.ogg'
+	reaction_strings = list(/decl/material/solid/metal/uranium = "Randomly teleports everything around the core.")
+
+/decl/slime_colour/quantum/handle_uranium_reaction(var/datum/reagents/holder)
+	var/list/turfs = list()
+	for(var/turf/T in orange(holder.my_atom,6))
+		turfs += T
+	for(var/atom/movable/a in viewers(holder.my_atom,2))
+		if(!a.simulated)
```
comma: maybe anchored check too, sounds like disaster waiting to happen (teleporting APCs etc)

code/__defines/damage_organs.dm
```diff
@@ -65,6 +65,10 @@
 #define DROPLIMB_BLUNT 1
 #define DROPLIMB_BURN 2
 
+// compile stub, remove on merge
+#define DISMEMBER_METHOD_ACID DROPLIMB_BURN
```
comma: merged, remov

MistakeNot4892:
I will need to integrate #1299 when it goes in, and I want to run some more testing with the scavs before I call this merge-ready, but it's ready for review.

code/modules/mob/living/slime/feeding.dm
```diff
@@ -0,0 +1,119 @@
...
+
+/mob/living/slime/proc/set_feeding_on(var/mob/living/victim)
+	if(feeding_on)
+		var/mob/feed_mob = feeding_on.resolve()
+		GLOB.moved_event.unregister(src, src)
+		GLOB.moved_event.unregister(feed_mob, src)
+		GLOB.destroyed_event.unregister(feed_mob, src)
+	feeding_on = null
+	if(victim)
+		if(feeding_on == weakref(victim))
```
comma: weird flow here
first we unregister and null feeding on, and THEN we check if we're getting same target as one we had.
I think this check should be at the top before unregistering so you don't jiggle events back and forth needlesly.

MistakeNot4892:
This has some minor issues remaining but I'm calling it done - can revisit the balancing aspects when it's merged and live on Scav.

code/modules/mob/living/living.dm
```diff
@@ -885,6 +875,29 @@ default behaviour is:
...
+	if(stat == DEAD || getCloneLoss() >= maxHealth)
+		eaten_by_slime()
+
+/mob/living/proc/eaten_by_slime()
+	if(QDELETED(src))
+		return
+	new /obj/effect/decal/cleanable/mucus(get_turf(src))
+	dump_contents()
+	qdel(src)
+	. = rand(2,3)
```
afterthought2: What does this mean?

code/modules/mob/living/slime/_slime.dm
```diff
@@ -0,0 +1,333 @@
...
+			if(is_adult)
+				stat(null, "You can reproduce!")
+			else
+				stat(null, "You can evolve!")
+
+		stat(null,"Power Level: [powerlevel]")
+
+/mob/living/slime/adjustFireLoss(amount)
+	..(-abs(amount)) // Heals them
+	return
```
afterthought2: Extraneous return.

code/modules/mob/living/slime/colours/colour_red.dm
```diff
@@ -0,0 +1,20 @@
...
+		/decl/slime_colour/oil,
+		/decl/slime_colour/oil
+	)
+	baby_icon =    'icons/mob/slimes/slime_baby_red.dmi'
+	adult_icon =   'icons/mob/slimes/slime_adult_red.dmi'
+	extract_icon = 'icons/mob/slimes/slime_extract_red.dmi'
+	reaction_strings = list(/decl/material/liquid/blood = "Causes all nearby slimes to enter a berserk rage.")
+
+/decl/slime_colour/red/handle_blood_reaction(var/datum/reagents/holder)
+	for(var/mob/living/slime/slime in viewers(get_turf(holder.my_atom), null))
```
afterthought2: The explicit null feels confusing; this uses world.view?

code/modules/mob/living/slime/_slime.dm
```diff
@@ -0,0 +1,333 @@
...
+	..()
+
+/mob/living/slime/explosion_act(severity)
+	..()
+	switch(severity)
+		if(1)
+			qdel(src)
+		if(2)
+			adjustBruteLoss(60)
+			adjustFireLoss(60)
```
afterthought2: Is it intentional that this heals them?

code/modules/mob/living/slime/items_extract_enhancer.dm
```diff
@@ -0,0 +1,22 @@
...
+	if(extract.enhanced == TRUE)
+		to_chat(user, SPAN_WARNING("\The [extract] has already been enhanced!"))
+		return TRUE
+	if(extract.Uses <= 0)
+		to_chat(user, SPAN_WARNING("You can't enhance a used extract!"))
+		return TRUE
+	to_chat(user, SPAN_NOTICE("You apply \the [src] to \the [extract]. It now has triple the amount of uses."))
+	extract.Uses = 3
+	extract.enhanced = TRUE
+	user.drop_from_inventory(src)
```
afterthought2: This should be redundant; `qdel` handles cleanup.

code/modules/mob/living/slime/items_potion.dm
```diff
@@ -0,0 +1,35 @@
...
+		if(M.stat)
+			to_chat(user, SPAN_WARNING("\The [M] is dead!"))
+			return TRUE
+		if(M.client)
+			to_chat(user, SPAN_WARNING("\The [M] resists!"))
+			return TRUE
+		var/mob/living/simple_animal/slime/pet = new slime_type(M.loc, M.slime_type)
+		to_chat(user, SPAN_NOTICE("You feed \the [pet] the potion, removing its powers and calming it."))
+		pet.prompt_rename(user)
+		user.drop_from_inventory(src)
```
afterthought2: Same here.

code/modules/mob/living/slime/items_steroid.dm
```diff
@@ -0,0 +1,25 @@
...
+	if(M.stat == DEAD)
+		to_chat(user, SPAN_WARNING("\The [M] is dead!"))
+		return TRUE
+	if(M.cores >= grant_cores)
+		to_chat(user, SPAN_WARNING("\The [M] already has the maximum amount of cores!"))
+		return TRUE
+
+	to_chat(user, SPAN_NOTICE("You feed \the [src] to \the [M] and it suddenly grows two extra cores."))
+	M.cores = grant_cores
+	user.drop_from_inventory(src)
```
afterthought2: And here.

code/modules/mob/living/slime/say.dm
```diff
@@ -0,0 +1,29 @@
+/mob/living/slime/say(var/message)
+	message = sanitize(message)
```
afterthought2: Is this really correct? Seems like it might double-encode.

Not entirely clear on the point of this override; shouldn't base mob handling deal with emotes properly?

code/modules/mob/living/slime/slime_AI.dm
```diff
@@ -0,0 +1,254 @@
...
+		newmood = "mischevous"
+
+	if (!newmood)
+		if (prob(1))
+			newmood = pick("sad", ":3")
+
+	if ((mood == "sad" || mood == ":3") && !newmood)
+		if (prob(75)) newmood = mood
+
+*/
```
afterthought2: ?

afterthought2 (commented):
Looks mostly good to me, though I'm no slime expert.

code/modules/mob/living/living.dm
MistakeNot4892: Feed value returned to the proc to give the slime nutrition.

code/modules/mob/living/slime/colours/colour_red.dm
MistakeNot4892: Not sure, this is copied code.

code/modules/mob/living/living.dm
afterthought2: Worth a comment, as this is the proc definition and that wouldn't be one of my first 3 guesses (it's also just ignored a few lines above, I believe).

MistakeNot4892: Added a comment above this proc in the new file I moved them to.

code/modules/mob/living/slime/slime_AI.dm
MistakeNot4892: Ah copied this in from source for reference when I was fixing some issues with it and forgot to remove it afterward.

code/modules/mob/living/slime/colours/colour_red.dm
MistakeNot4892: Switched null out for a numeric range.

code/modules/mob/living/slime/_slime.dm
MistakeNot4892: Going to revisit this chunk of code, might rip it out entirely.

code/modules/projectiles/projectile/change.dm
```diff
@@ -9,91 +9,56 @@
...
+		var/mob/living/silicon/robot/R = new(get_turf(M))
+		R.gender = M.gender
+		R.job = "Robot"
+		R.mmi = new /obj/item/mmi(R)
+		R.mmi.transfer_identity(M)
+		return R
+
+	if(get_species_by_key(choice))
+		var/mob/living/carbon/human/H = M
+		if(!istype(H))
```
comma, line 34: shouldn't this also transfer player or something? Looks like it just creates a human on target's turf and calls it a day.

code/modules/mob/living/carbon/carbon_defense.dm
```diff
@@ -27,3 +27,13 @@
 			I.has_embedded()
 
 	return 1
+
+var/list/slime_pain_messages = list(
```
comma: shouldn't these be moved to modpack too

code/modules/projectiles/projectile/change.dm
MistakeNot4892: They get mind transferred after apply_transformation is called.

code/modules/mob/living/carbon/carbon_defense.dm
MistakeNot4892: yes

comma (dismissed):

comma (approved):

PR #1248, opened 2021-01-31
Tradeship xenobio fixes.

### LINKED ISSUE #895: Tradeship mapping: xenobio slime containment doors should be secure
#### Description of issue
slimes wander out of containment all willy nilly.

Containment ideally contains, whereas this does not

#### Steps to reproduce
give a slime a few hours, it will almost without fail eventually find its way out of containment.  Likely due to boredom.   

Tradeship-0.dmm

Since xenobio was added to tradeship, probably

Client Version: 513
Server Revision: fa5f5a3a9c8e0f7f9e8f145bc0c70b46c6bf41ba - dev - 2020-09-12
Game ID: b88-akEb

#### Issue bingo

### LINKED ISSUE #904: Tradeship Xenobio lacking some equiptment
Xenobio on the tradeship needs a fire extinguisher, or equivalent, as well as a small supply of activator for the slimes. (whichever activates grey cores into slimes again, I believe it's Radium or Uranium now?) two fairly important parts of the job. A grinder to turn the uranium into the inject able liquid would also be nice, but of low priority, as the large machine in chems could suffice.
(heck, could probably source activator from there too, but a fire extinguisher, or a water sprayer of some kind is a must. materials could easily be replaced by a codex listing for slimes describing the basics, aka: feed DNA, grow, split, mutate, kill, extract cores, and activate cores with blood and/or 'X', but until then, there's no real way of knowing what material is to be used anymore.
(That in and of itself might be intentional these days?))

MistakeNot4892:
Adds uranium, syringes and a grinder to Tradeship xenobio, and fixes the access on the doors. Doesn't implement slime codex, working on that in #1243.

- Fixes #904.

### DIFF
--- a/maps/tradeship/tradeship-0.dmm
+++ b/maps/tradeship/tradeship-0.dmm
@@ -1627,6 +1627,11 @@
 /obj/machinery/light{
 	dir = 4
 	},
+/obj/machinery/alarm{
+	dir = 8;
+	icon_state = "alarm0";
+	pixel_x = 24
+	},
 /turf/simulated/floor/tiled/white,
 /area/ship/trade/livestock)
 "xc" = (
@@ -1658,6 +1663,11 @@
 	d2 = 2;
 	icon_state = "1-2";
 	pixel_y = 0
+	},
+/obj/structure/hygiene/sink/kitchen{
+	dir = 8;
+	icon_state = "sink_alt";
+	pixel_x = 22
 	},
 /turf/simulated/floor/tiled/white,
 /area/ship/trade/livestock)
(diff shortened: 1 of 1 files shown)


maps/tradeship/tradeship-0.dmm
```diff
@@ -658,8 +658,14 @@
 /turf/simulated/floor,
 /area/ship/trade/aft_port_underside_maint)
 "bu" = (
-/obj/machinery/door/window/brigdoor/northright,
-/obj/machinery/door/window/brigdoor/southright,
+/obj/machinery/door/window/brigdoor/northright{
+	req_access = list("ACCESS_XENOBIO");
```
comma: :disgust: 
Why not set access on the area? Area access inheritance is /door level.

MistakeNot4892: Access is set on the area, it just won't inherit it and I couldn't work out why.

MistakeNot4892: #1253

afterthought2 (approved):

PR #1250, opened 2021-01-31
FTL Shunt

Mazianni:
## The FTL Shunt Drive

This is a concept I've had for some time, but essentially would allow (limited, costly) instant movement on the overmap. Useful for getting out of a pinch.

The superluminal drive allows for Point-To-Point movement. Uses fuel depending on distance and vessel mass, split evenly between all of the FTL shunt's fuel ports. Sudden FTL movement is not easy on biological beings nor mechanical objects, however, with the severity growing as the jump distance increases. 

Shunts use fuel as a function of distance, and require a certain amount of charge time as a function of vessel mass.

If sabotaged, bad things can and *will* happen.

TODO:

- [x] Actually finish the shunt drive.
- [ ] Acquire sprites that aren't actual trash.
- [x] Discuss what mechanics are / are wanted with maintainers, narrow scope of project.
- [x] Discuss effects / maladies from jumping too far.
- [x] Discuss if phasing is really something that should be supported or not.


Mazianni:
TODO 2:

- [x] Finish interactions for FTL core and fuel ports
- [x] Write UI code for FTL computer.
- [x] Tweak values
- [x] Add sabotage methods
- [ ] Potentially integrate into starship jump controller.

code/modules/overmap/ftl_shunt/computer.dm
```diff
@@ -0,0 +1,112 @@
...
+					to_chat(user, SPAN_WARNING("Shunt drive inoperable: cooldown interlocks engaged."))
+				if(FTL_START_FAILURE_OTHER)
+					to_chat(user, SPAN_WARNING("Shunt drive inoperable: unknown error."))
+				if(FTL_START_CONFIRMED)
+					to_chat(user, SPAN_WARNING("Shunt drive operational: spooling up."))
+			return TOPIC_REFRESH
+
+	if(href_list["cancel_shunt"])
+		if(linked_core)
+			var/warning = alert(user, "Cancel current shunt?","Jump Safety System", "Yes", "No")
```
MistakeNot4892, line 142: Need to validate usr etc. post-alert.

code/modules/overmap/ftl_shunt/computer.dm
```diff
@@ -0,0 +1,112 @@
...
+				if(FTL_START_FAILURE_POWER)
+					to_chat(user, SPAN_WARNING("Shunt drive inoperable: verify primary power."))
+				if(FTL_START_FAILURE_BROKEN)
+					to_chat(user, SPAN_WARNING("Shunt drive inoperable: safety interlocks engaged."))
+				if(FTL_START_FAILURE_COOLDOWN)
+					to_chat(user, SPAN_WARNING("Shunt drive inoperable: cooldown interlocks engaged."))
+				if(FTL_START_FAILURE_OTHER)
+					to_chat(user, SPAN_WARNING("Shunt drive inoperable: unknown error."))
+				if(FTL_START_CONFIRMED)
+					to_chat(user, SPAN_WARNING("Shunt drive operational: spooling up."))
```
MistakeNot4892: SPAN_NOTICE

code/modules/overmap/ftl_shunt/computer.dm
```diff
@@ -0,0 +1,112 @@
...
+
+/obj/machinery/computer/ship/ftl/Initialize()
+	. = ..()
+	find_core()
+
+/obj/machinery/computer/ship/ftl/proc/find_core()
+	if(!linked)
+		return
+
+	for(var/obj/machinery/ftl_shunt/core/C in world)
```
MistakeNot4892: Ssmachines.machinery

code/game/objects/items/weapons/circuitboards/computer/computer.dm
```diff
@@ -68,6 +68,10 @@
 	name = "circuitboard (helm control console)"
 	build_path = /obj/machinery/computer/ship/helm
 
+/obj/item/stock_parts/circuitboard/ftl
+	name = "circuitboard (ftl control console)"
```
MistakeNot4892: FTL

code/modules/overmap/ftl_shunt/computer.dm
```diff
@@ -0,0 +1,112 @@
...
+			return TOPIC_REFRESH
+
+	if(href_list["start_shunt"])
+		if(linked_core)
+			if(linked_core.get_status() != FTL_STATUS_GOOD)
+				to_chat(user, SPAN_WARNING("Shunt drive inoperable. Please try again later."))
+				return TOPIC_REFRESH
+
+			var/dist = get_dist(locate(linked_core.shunt_x, linked_core.shunt_y, GLOB.using_map.overmap_z), get_turf(linked))
+			if(dist > 3) //We are above the safe jump distance, give them a warning.
```
MistakeNot4892, line 114: Maybe a define or proc? Better machines from upgrades for longer safe dist?

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,344 @@
...
+/obj/machinery/ftl_shunt/core/Destroy()
+	. = ..()
+	for(var/obj/machinery/ftl_shunt/fuel_port/FP in fuel_ports)
+		FP.master = null
+		fuel_ports -= FP
+	ftl_computer.linked_core = null
+	ftl_computer = null
+
+/obj/machinery/ftl_shunt/core/proc/find_ports()
+	for(var/obj/machinery/ftl_shunt/fuel_port/FP in range(7))
```
MistakeNot4892: Consider using a local machine network extension like the RUST

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,344 @@
...
+		if(!FP.master)
+			FP.master = src
+			fuel_ports += FP
+
+/obj/machinery/ftl_shunt/core/proc/start_shunt()
+	var/shunt_distance
+	var/vessel_mass = ftl_computer.linked.get_vessel_mass()
+	var/shunt_turf = locate(shunt_x, shunt_y, GLOB.using_map.overmap_z)
+
+	if(stat & (BROKEN|NOPOWER))
```
MistakeNot4892: No point checking when you check both right after.

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,344 @@
...
+	ftl_announcement.Announce(shunt_cancel_text, "Shunt Drive Management System")
+	update_use_power(POWER_USE_IDLE)
+
+/obj/machinery/ftl_shunt/core/proc/execute_shunt()
+	use_power = POWER_USE_IDLE
+	var/destination = locate(shunt_x, shunt_y, GLOB.using_map.overmap_z)
+	var/jumpdist = get_dist(get_turf(ftl_computer.linked), destination)
+	addtimer(CALLBACK(src, .proc/do_shunt, shunt_x, shunt_y, jumpdist, destination), 6 SECONDS)
+	for(var/mob/living/carbon/M in world)
+		if(!M.z in ftl_computer.linked.map_z)
```
MistakeNot4892: `in` has very low operator precedence, use !(x in y)

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,344 @@
...
+	switch(distance)
+		if(1 to 3)
+			shunt_sev = SHUNT_SEVERITY_MINOR
+		if(4 to 5)
+			shunt_sev = SHUNT_SEVERITY_MAJOR
+		if(6 to INFINITY)
+			shunt_sev = SHUNT_SEVERITY_CRITICAL
+
+	for(var/mob/living/carbon/human/H in world) //Affect mobs, skip synthetics.
+		sound_to(H, 'sound/machines/hyperspace_end.ogg')
```
MistakeNot4892, line 292: As above

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,344 @@
...
+
+			if(SHUNT_SEVERITY_CRITICAL)
+				to_chat(H, SPAN_DANGER("You feel an overwhelming sense of nausea and vertigo wash over you, your instincts screaming that something is wrong!"))
+				if(prob(50))
+					H.hallucination(100, 100)
+				if(prob(45))
+					H.vomit()
+				shake_camera(H, 5, 4)
+
+	for(var/obj/machinery/light/L in world) //Fuck with and or break lights.
```
MistakeNot4892: SSmachines.machinery

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,344 @@
...
+		if(!L.z in ftl_computer.linked.map_z)
```
MistakeNot4892: Operator precedence

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,344 @@
...
+			continue
+		switch(shunt_sev)
+			if(SHUNT_SEVERITY_MINOR)
+				if(prob(15))
+					L.flicker()
+			if(SHUNT_SEVERITY_MAJOR)
+				if(prob(35))
+					L.flicker()
+
+	for(var/obj/machinery/power/apc/A in world)
```
MistakeNot4892: Pls

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,344 @@
...
+			ports_used++
+
+	if(ports_used == fueled_ports.len)
+		return TRUE
+
+/obj/machinery/ftl_shunt/core/proc/get_charge_time()
+	return (ftl_computer.linked.vessel_mass * CHARGE_TIME_PER_TON) * time_multiplier
+
+/obj/machinery/ftl_shunt/core/Process()
+	if(stat & (BROKEN|NOPOWER) && charging)
```
MistakeNot4892: (X & y)

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,344 @@
...
+		cancel_shunt()
+
+	if(!fuel_ports.len)
```
MistakeNot4892: Minor note but lots of .len, consider a lazylist

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,344 @@
...
+
+	if(!charging)
+		update_use_power(POWER_USE_IDLE)
+
+/obj/machinery/ftl_shunt/fuel_port
+	name = "FTL shunt fuel port"
+	desc = "A fuel port for an FTL shunt device."
+	icon_state = "empty"
+
+	var/list/fuels = list(/decl/material/gas/hydrogen/tritium = 25000, /decl/material/gas/hydrogen/deuterium = 25000, /decl/material/gas/hydrogen = 25000, /decl/material/solid/exotic_matter = 50000)
```
MistakeNot4892: Static or global unless this is mutated at some point

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,344 @@
...
+	var/obj/item/fuel_assembly/fuel
+	var/obj/machinery/ftl_shunt/core/master
+	var/max_fuel = 0
+
+/obj/machinery/ftl_shunt/fuel_port/on_update_icon()
+	if(fuel)
```
MistakeNot4892, line 491: Reset icon state if no fuel

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,344 @@
...
+		icon_state = "full"
+
+/obj/machinery/ftl_shunt/fuel_port/Destroy()
+	if(master)
+		master.fuel_ports -= src
+	master = null
+	QDEL_NULL(fuel)
+
+/obj/machinery/ftl_shunt/fuel_port/attackby(var/obj/item/O, var/mob/user)
```
MistakeNot4892, line 510: Return TRUE and . = ..()

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,344 @@
...
+		if(!fuel)
+			if(!do_after(user, 20, src))
+				return
+			if(user && !user.unEquip(O))
+				return
+			O.forceMove(src)
+			fuel = O
+			max_fuel = get_max_fuel()
+			update_icon()
+/obj/machinery/ftl_shunt/fuel_port/attack_hand(var/mob/user)
```
MistakeNot4892: Return TRUE and . = ..()

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,344 @@
...
+		if(!do_after(user, 20, src))
+			return
+		fuel.forceMove(get_turf(user))
+		user.put_in_hands(fuel)
+		fuel = null
+		max_fuel = 0
+
+
+
+/obj/machinery/ftl_shunt/fuel_port/proc/has_fuel()
```
MistakeNot4892: Return !!fuel

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,344 @@
...
+
+	var/total_joules_amount
+
+	for(var/G in fuel.rod_quantities)
+		if(G in fuels)
+			var/joules_per_unit = fuels[G]
+			var/total_joules = 10000 * joules_per_unit
+			total_joules_amount += total_joules
+
+	return total_joules_amount
```
MistakeNot4892: Could just use `.`

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,344 @@
...
+
+/obj/machinery/ftl_shunt/fuel_port/proc/get_avaliable_fuel_joules()
```
MistakeNot4892: Seems very similar to above proc

code/modules/overmap/ftl_shunt/core.dm
Mazianni: Outside of my technical understanding.

code/modules/overmap/ftl_shunt/core.dm
Mazianni: ?

code/modules/overmap/ftl_shunt/core.dm
Mazianni: One gets maximum fuel, one gets current fuel.

code/modules/overmap/ftl_shunt/core.dm
MistakeNot4892: `if((stat & (BROKEN|NOPOWER)) && charging)`
Similar to the issue with 'in', sometimes it can be ambiguous if that will resolve to `(x & y) && z` or `x & (y && z)` .

code/modules/overmap/ftl_shunt/core.dm
MistakeNot4892: You could collapse both procs into a single proc with a bool.
````
/obj/machinery/ftl_shunt/fuel_port/proc/get_fuel_joules(var/get_fuel_maximum)
	if(fuel)
		for(var/G in fuel.rod_quantities)
			if(G in fuels)
				. += (get_fuel_maximum ? fuel.rod_quantities[G] : 10000) * fuels[G]
````

File: code/modules/overmap/ftl_shunt/computer.dm (hunk 1 above)
Comment (Mazianni) [reply code/modules/overmap/ftl_shunt/computer.dm:142]: ?

File: code/modules/overmap/ftl_shunt/computer.dm (hunk 5 above)
Comment (Mazianni) [reply code/modules/overmap/ftl_shunt/computer.dm:114]: Distance is a balance concern at the moment, I'm not inclined to make it longer. *Maybe* longer safe dist, but i'm not sure atm.

File: code/modules/overmap/ftl_shunt/computer.dm [hunk 22]
```diff
@@ -0,0 +1,143 @@
...
+	var/obj/machinery/ftl_shunt/core/linked_core
+	var/cost = 0
+
+/obj/machinery/computer/ship/ftl/Initialize()
+	. = ..()
+	find_core()
+
+/obj/machinery/computer/ship/ftl/Destroy()
+	. = ..()
+	linked_core = null
```
afterthought2: This fails to clean up possible refs to us stored on the core.

code/modules/overmap/ftl_shunt/computer.dm
```diff
@@ -0,0 +1,143 @@
...
+	cost = jump_cost
+
+/obj/machinery/computer/ship/ftl/proc/find_core()
+	if(!linked)
+		return
+
+	for(var/obj/machinery/ftl_shunt/core/C in SSmachines.machinery)
+		if(C.z in linked.map_z)
+			linked_core = C
+			C.ftl_computer = src
```
afterthought2, line 37: Prior to doing this, you need to properly unset an already set core (meaning to wipe its refs to us, not just our refs to it).

code/modules/overmap/ftl_shunt/computer.dm
```diff
@@ -0,0 +1,143 @@
...
+	ui = SSnano.try_update_ui(user, src, ui_key, ui, data, force_open)
+	if (!ui)
+		ui = new(user, src, ui_key, "ftl_computer.tmpl", "[linked.name] Shunt Drive Control", 420, 530, src)
+		ui.set_initial_data(data)
+		ui.open()
+		ui.set_auto_update(1)
+
+/obj/machinery/computer/ship/ftl/OnTopic(var/mob/user, var/list/href_list, state)
+	if(..())
+		return TOPIC_HANDLED
```
afterthought2: You should either not call parent at all, or return `..()` here.

code/modules/overmap/ftl_shunt/computer.dm
```diff
@@ -0,0 +1,143 @@
...
+		return TOPIC_NOACTION
+
+	if(href_list["set_shunt_x"])
+		if(linked_core)
+			var/fumble = 0
+			if(!user.skill_check(SKILL_PILOT, SKILL_ADEPT))
+				fumble = rand(-2, 2)
+			var/input_x = input(user, "Enter Destination X Coordinates", "FTL Computer", linked_core.shunt_x) as num|null
+			if(input_x >= GLOB.using_map.overmap_size)
+				to_chat(user, SPAN_WARNING("Input invalid. Must be between 0 and [GLOB.using_map.overmap_size - 1]."))
```
afterthought2: Seems more user-friendly to just do min(input, overmap_size) instead (especially given the randomization just above. Also, why check for max but not for min, or for it not being an integer?

code/modules/overmap/ftl_shunt/computer.dm
```diff
@@ -0,0 +1,143 @@
...
+			if(input_y >= GLOB.using_map.overmap_size)
+				to_chat(user, SPAN_WARNING("Input invalid. Must be between 0 and [GLOB.using_map.overmap_size - 1]."))
+			else
+				if(isnull(input_y))
+					input_y = linked_core.shunt_y
+				linked_core.shunt_y = input_y + fumble
+				recalc_cost()
+				if(fumble)
+					to_chat(user, SPAN_WARNING("You fumble the input because of your inexperience!"))
+			return TOPIC_REFRESH
```
afterthought2: I detect copypaste.

code/modules/overmap/ftl_shunt/computer.dm
```diff
@@ -0,0 +1,143 @@
...
+				to_chat(user, SPAN_NOTICE("Superluminal jump charge in progress. Please wait for completion of jump!"))
+				return TOPIC_REFRESH
+			if(linked_core.get_status() != FTL_STATUS_GOOD)
+				to_chat(user, SPAN_WARNING("Superluminal shunt inoperable. Please try again later."))
+				return TOPIC_REFRESH
+
+			var/dist = get_dist(locate(linked_core.shunt_x, linked_core.shunt_y, GLOB.using_map.overmap_z), get_turf(linked))
+			if(dist > 3) //We are above the safe jump distance, give them a warning.
+				var/warning = alert(user, "Current shunt distance is above safe distance! Do you wish to continue?","Jump Safety System", "Yes", "No")
+				if(warning == "No" && !user.incapacitated() && user.Adjacent(src) && linked_core)
```
afterthought2: The best check to make here is `CanInteract(user, state)`.

code/modules/overmap/ftl_shunt/computer.dm
```diff
@@ -0,0 +1,143 @@
...
+				if(fumble)
+					to_chat(user, SPAN_WARNING("You fumble the input because of your inexperience!"))
+			return TOPIC_REFRESH
+
+	if(href_list["set_shunt_y"])
+		if(linked_core)
+			var/fumble = 0
+			if(!user.skill_check(SKILL_PILOT, SKILL_ADEPT))
+				fumble = rand(-2, 2)
+			var/input_y = input(user, "Enter Destination Y Coordinates", "FTL Computer", linked_core.shunt_y) as num|null
```
afterthought2: Need to recheck CanInteract or similar.

code/modules/overmap/ftl_shunt/computer.dm
```diff
@@ -0,0 +1,143 @@
...
+				if(FTL_START_FAILURE_OTHER)
+					to_chat(user, SPAN_WARNING("Superluminal shunt inoperable: unknown error."))
+				if(FTL_START_CONFIRMED)
+					to_chat(user, SPAN_NOTICE("Superluminal shunt operational: spooling up."))
+			return TOPIC_REFRESH
+
+	if(href_list["cancel_shunt"])
+		if(linked_core)
+			var/warning = alert(user, "Cancel current shunt?","Jump Safety System", "Yes", "No")
+			if(warning == "Yes" && !user.incapacitated() && user.Adjacent(src) && linked_core)
```
afterthought2: Same here, `CanInteract`.

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,486 @@
+var/list/global/fuels = list(/decl/material/gas/hydrogen/tritium = 25000, /decl/material/gas/hydrogen/deuterium = 25000, /decl/material/gas/hydrogen = 25000, /decl/material/solid/exotic_matter = 50000)
```
afterthought2: Why are these not vars on the decls?

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,486 @@
...
+				sabotaged = SHUNT_SABOTAGE_MAJOR
+			if(50 to INFINITY)
+				sabotaged = SHUNT_SABOTAGE_CRITICAL
+
+		sabotaged_amt = tc_input
+		TC.use(tc_input)
+		to_chat(user, SPAN_DANGER("You successfully sabotage [src] by inserting the crystals!"))
+		return TRUE
+
+/obj/machinery/ftl_shunt/core/attack_hand(var/mob/user)
```
afterthought2: Do not override this proc. Use `physical_attack_hand` instead.

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,486 @@
...
+/obj/machinery/ftl_shunt/core/proc/cancel_shunt(var/silent = FALSE)
+	if(!charging) //Not preparing for a jump.
+		return
+	charging = FALSE
+	charge_time = null
+	charge_started = null
+	cooldown = null
+	required_fuel_joules = null
+	if(silent)
+		ftl_announcement.Announce(shunt_cancel_text, "FTL Shunt Management System", new_sound = sound('sound/misc/notice2.ogg'))
```
afterthought2, line 243: Confusing usage of this arg; do you mean `!silent`?

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,486 @@
...
+	update_use_power(POWER_USE_IDLE)
+
+//Starts the shunt, and then hands off to do_shunt to finish it.
+/obj/machinery/ftl_shunt/core/proc/execute_shunt()
+	if(sabotaged)
+		cancel_shunt(TRUE)
+		do_sabotage()
+		return
+
+	use_power = POWER_USE_IDLE
```
afterthought2: Do not set this var directly.

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,486 @@
...
+	var/destination = locate(shunt_x, shunt_y, GLOB.using_map.overmap_z)
+	var/jumpdist = get_dist(get_turf(ftl_computer.linked), destination)
+	var/obj/effect/portal/wormhole/W = new(destination) //Generate a wormhole effect on overmap to give some indication that something is about to happen.
+	QDEL_IN(W, 6 SECONDS)
+	addtimer(CALLBACK(src, .proc/do_shunt, shunt_x, shunt_y, jumpdist, destination), 6 SECONDS)
+	for(var/mob/living/carbon/M in world)
```
afterthought2: There is a list of living mobs that can be iterated for this.

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,486 @@
...
+	var/shunt_sev
+	switch(distance)
+		if(1 to 3)
+			shunt_sev = SHUNT_SEVERITY_MINOR
+		if(4 to 5)
+			shunt_sev = SHUNT_SEVERITY_MAJOR
+		if(6 to INFINITY)
+			shunt_sev = SHUNT_SEVERITY_CRITICAL
+
+	for(var/mob/living/carbon/human/H in world) //Affect mobs, skip synthetics.
```
afterthought2: There is a list of living mobs that can be iterated for this.

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,486 @@
...
+
+/obj/machinery/ftl_shunt/core/proc/get_charge_time()
+	return (ftl_computer.linked.vessel_mass * CHARGE_TIME_PER_TON) * time_multiplier
+
+/obj/machinery/ftl_shunt/core/Process()
+	if((stat & (BROKEN|NOPOWER)) & charging)
+		cancel_shunt()
+
+	if(!length(fuel_ports))
+		find_ports()
```
afterthought2: Are we really running this every tick? Why? This iterates a giant list.

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,486 @@
...
+
+	if(charging)
+
+		if(sabotaged && !sabotaged_warn)
+			for(var/mob/living/carbon/human/H in world) //Give engineers a hint that something might be very, very wrong.
+				if(!(H.z in ftl_computer.linked.map_z))
+					continue
+				if(H.skill_check(SKILL_ENGINES, SKILL_EXPERT))
+					to_chat(H, SPAN_DANGER("The deck vibrates with a harmonic that sets your teeth on edge and fills you with dread."))
```
afterthought2: This seems awful; you are spamming this on a 2 second repeat.

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,486 @@
...
+			sabotaged_warn = TRUE
+
+		if(use_power != POWER_USE_ACTIVE)
+			update_use_power(POWER_USE_ACTIVE)
```
afterthought2: Why do this every tick? This entire block feels like it should be done on charge start, not in Process.

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,486 @@
...
+		if(world.time >= charge_time) //We've probably finished charging up.
+			charging = FALSE
+			if(use_fuel(required_fuel_joules))
+				execute_shunt()
+			else
+				cancel_shunt() //Not enough fuel for whatever reason. Cancel.
+		SSradiation.radiate(src, (active_power_usage / 1000))
+
+	if(!charging)
+		update_use_power(POWER_USE_IDLE)
```
afterthought2: Again, just have it do this whenever `charging` is toggled off, not check it every tick in `Process`.

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,486 @@
...
+		local_network.set_tag(null, initial_id_tag)
+	. = ..()
+
+/obj/machinery/ftl_shunt/fuel_port/Destroy()
+	. = ..()
+	if(master)
+		master.fuel_ports -= src
+	master = null
+	QDEL_NULL(fuel)
+	return TRUE
```
afterthought2: Do not do this.

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,486 @@
...
+			if(user && !user.unEquip(O))
+				return
+			O.forceMove(src)
+			fuel = O
+			max_fuel = get_fuel_joules(TRUE)
+			update_icon()
+			return TRUE
+
+	. = ..()
+	return FALSE
```
afterthought2: Why?

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,486 @@
...
+
+/obj/machinery/ftl_shunt/fuel_port/attack_hand(var/mob/user)
```
afterthought2: Do not override this proc; use `physical_attack_hand`.

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,486 @@
...
+			return FALSE
+		fuel.forceMove(get_turf(user))
+		user.put_in_hands(fuel)
+		fuel = null
+		max_fuel = 0
+		to_chat(user, SPAN_NOTICE("You remove the fuel assembly!"))
+		return TRUE
+
+	. = ..()
+	return FALSE
```
afterthought2: Why?

code/modules/overmap/ftl_shunt/computer.dm
Mazianni: It didn't occur to me to do those things, and I'm not sure how to check integers.

code/modules/overmap/ftl_shunt/core.dm
Mazianni: No, silent is FALSE by default.

code/modules/overmap/ftl_shunt/core.dm
Mazianni: This message is only ever sent once, as sabotage_warn is set to TRUE afterwards, and it is unreachable after the var is set to TRUE.

Mazianni: moved to start_shunt anyway

code/modules/overmap/ftl_shunt/computer.dm
```diff
@@ -0,0 +1,148 @@
...
+
+	if(href_list["set_shunt_x"])
+		if(linked_core)
+			var/fumble = 0
+			if(!user.skill_check(SKILL_PILOT, SKILL_ADEPT))
+				fumble = rand(-2, 2)
+			var/input_x = input(user, "Enter Destination X Coordinates", "FTL Computer", linked_core.shunt_x) as num|null
+			if(input_x >= GLOB.using_map.overmap_size)
+				input_x = min(input_x, GLOB.using_map.overmap_size - 1)
+			else if(CanInteract(user, state))
```
MistakeNot4892: This `else` means the above two lines are pointless, `input_x` will just be discarded.

code/modules/modular_computers/file_system/programs/command/comm.dm
```diff
@@ -112,6 +113,20 @@
 		return program.can_run(user, program.computer.get_network())
 	return 1
 
+/datum/nano_module/program/comm/proc/get_shunt()
+	var/obj/comp = program.computer.get_physical_host()
+	var/obj/effect/overmap/visitable/ship/sector = map_sectors["[comp.z]"]
```
MistakeNot4892: Would it be worth nullchecking `comp` and `program.computer`? Not sure how robust modular programs are.

code/modules/overmap/ftl_shunt/computer.dm
```diff
@@ -0,0 +1,150 @@
...
+	find_core()
+
+/obj/machinery/computer/ship/ftl/Destroy()
+	. = ..()
+	if(linked_core)
+		linked_core.ftl_computer = null
+		linked_core = null
+
+/obj/machinery/computer/ship/ftl/proc/recalc_cost()
+	var/jump_dist = get_dist(linked, locate(linked_core.shunt_x, linked_core.shunt_y, GLOB.using_map.overmap_z))
```
MistakeNot4892, line 22: Maybe return `INFINITY` or something if `linked` or `linked_core` are null?

code/modules/overmap/ftl_shunt/computer.dm
```diff
@@ -0,0 +1,150 @@
...
+		return
+
+	if(linked_core)
+		linked_core.ftl_computer = null
+		linked_core = null
+
+	for(var/obj/machinery/ftl_shunt/core/C in SSmachines.machinery)
+		if(C.z in linked.map_z)
+			linked_core = C
+			C.ftl_computer = src
```
MistakeNot4892, line 37: `break` or `return` after finding the core or you'll iterate the rest of the machinery list.

code/modules/overmap/ftl_shunt/computer.dm
```diff
@@ -0,0 +1,150 @@
...
+		return
+
+	var/data[0]
+
+	if(!linked_core)
+		find_core()
+		if(!linked_core)
+			to_chat(user, SPAN_WARNING("Unable to establish connection to superluminal shunt."))
+			return
+	recalc_cost()
```
MistakeNot4892, line 52: Seems a bit strange to calc this during the UI update - you might be doing multiple updates per second, etc. Could the cost be calculated when you find your core maybe?

code/modules/overmap/ftl_shunt/computer.dm
```diff
@@ -0,0 +1,150 @@
...
+		ui.set_auto_update(1)
+
+/obj/machinery/computer/ship/ftl/OnTopic(var/mob/user, var/list/href_list, state)
+	if(..())
+		return ..()
+
+	if (!linked)
+		return TOPIC_NOACTION
+
+	if(href_list["set_shunt_x"])
```
MistakeNot4892: Lots of boilerplate. Maybe something like:
````
if(href_list["set_shunt_x"] || href_list["set_shunt_y"])

	if(!linked_core)
		return TOPIC_NOACTION

	var/input_x = linked_core.shunt_x
	var/input_y = linked_core.shunt_y
	var/fumble = user.skill_check(SKILL_PILOT, SKILL_ADEPT) ? 0 : rand(-2, 2)

	if(href_list["set_shunt_x"])
		input_x = Clamp((input(user, "Enter Destination X Coordinates", "FTL Computer", input_x) as num|null) + fumble, 1, GLOB.using_map.overmap_size - 1)

	if(href_list["set_shunt_y"])
		input_y = Clamp((input(user, "Enter Destination Y Coordinates", "FTL Computer", input_y) as num|null) + fumble, 1, GLOB.using_map.overmap_size - 1)

	if(!CanInteract(user, state))
		return TOPIC_NOACTION
	if(fumble)
		to_chat(user, SPAN_WARNING("You fumble the input because of your inexperience!"))
	linked_core.shunt_x = input_x
	linked_core.shunt_y = input_y
	return TOPIC_REFRESH
````
Haven't even tried to compile this, mind.

File: code/modules/overmap/ftl_shunt/computer.dm [hunk 50]
```diff
@@ -0,0 +1,150 @@
...
+				to_chat(user, SPAN_NOTICE("Superluminal jump charge in progress. Please wait for completion of jump!"))
+				return TOPIC_REFRESH
+			if(linked_core.get_status() != FTL_STATUS_GOOD)
+				to_chat(user, SPAN_WARNING("Superluminal shunt inoperable. Please try again later."))
+				return TOPIC_REFRESH
+
+			var/dist = get_dist(locate(linked_core.shunt_x, linked_core.shunt_y, GLOB.using_map.overmap_z), get_turf(linked))
+			if(dist > 3) //We are above the safe jump distance, give them a warning.
+				var/warning = alert(user, "Current shunt distance is above safe distance! Do you wish to continue?","Jump Safety System", "Yes", "No")
+				if(warning == "No" && CanInteract(user, state))
```
MistakeNot4892: Shouldn't this be `|| !CanInteract(user, state))`?

code/modules/overmap/ftl_shunt/computer.dm
```diff
@@ -0,0 +1,150 @@
...
+					return TOPIC_REFRESH
+
+			if(dist >= linked_core.max_jump_distance)
+				to_chat(user, SPAN_WARNING("Shunt aborted: Selected jump coordinates beyond maximum range."))
+				return TOPIC_REFRESH
+
+
+			var/feedback = linked_core.start_shunt()
```
MistakeNot4892, line 124: You could make `start_shunt()` return the failure message rather than an int using your defines and wipe out this switch entirely. Not a requested change, just a thought.

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,551 @@
+/obj/machinery/ftl_shunt
+	anchored = 1
+	icon = 'icons/obj/shunt_drive.dmi'
+	var/initial_id_tag = "ftl"
+
+/obj/machinery/ftl_shunt/core
+	name = "superluminal shunt core"
+	desc = "An immensely powerful particle accelerator capable of forming a micro-wormhole and shunting an entire ship through it in a nanosecond."
```
MistakeNot4892: I know this is just technobabble but particle accelerator so bland, why not some ridiculous shit like `an n-dimensional non-baryonic brane deformation system`

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,551 @@
...
+	var/sabotaged_amt = 0 //amount of crystals used to sabotage us.
+
+	var/static/datum/announcement/priority/ftl_announcement = new(do_log = 0, do_newscast = 1, new_sound = sound('sound/misc/notice2.ogg'))
+
+	var/shunt_start_text = "Attention! All hands brace for faster-than-light transition! ETA: %%TIME%%"
+	var/shunt_cancel_text = "Attention! Faster-than-light transition cancelled."
+	var/shunt_complete_text = "Attention! Faster-than-light transition completed."
+	var/shunt_spooling_text = "Attention! Superluminal shunt charge complete, spooling up."
+
+	var/shunt_sabotage_text_minor = "Warning! Electromagnetic flux beyond safety limits - aborting shunt!"
```
MistakeNot4892: These strings should probably be static (or defines) rather than instance variables.

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,551 @@
...
+	var/shunt_y = 1
+	var/chargepercent = 0
+	var/last_percent_tick = 0
+	var/obj/machinery/computer/ship/ftl/ftl_computer
+	var/required_fuel_joules
+	var/cooldown_delay = 5 MINUTES
+	var/cooldown
+	var/time_multiplier = 5 //The multiplier on how long it takes the shunt drive to charge.
+	var/max_jump_distance = 8 //How many overmap tiles can this move the ship?
+	var/sabotaged = null
```
MistakeNot4892: You don't need to do `= null` when you're declaring a null var.

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,551 @@
...
+/obj/machinery/ftl_shunt/core/Destroy()
+	. = ..()
+	for(var/obj/machinery/ftl_shunt/fuel_port/FP in fuel_ports)
+		FP.master = null
+		fuel_ports -= FP
+	ftl_computer.linked_core = null
+	ftl_computer = null
+
+/obj/machinery/ftl_shunt/core/on_update_icon()
+	overlays.Cut()
```
MistakeNot4892: Please use `cut_overlays()` in conjunction with `add_overlays()` as I'll comment below.

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,551 @@
...
+
+	if(charging)
+		var/image/I = image('icons/obj/shunt_drive.dmi', "activating")
+		var/matrix/M = new()
+		I.transform = M
+		overlays += I
```
MistakeNot4892: `add_overlay(I)`

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,551 @@
...
+
+	if(jumping)
+		overlays += image('icons/obj/shunt_drive.dmi', "activated")
+		var/image/S = image('icons/obj/objects.dmi', "bhole3")
+		var/matrix/M = new()
+		M.Scale(0.75)
+		S.transform = M
+		S.alpha = 0
+		animate(S, alpha = 255, time = 5.9 SECONDS)
+		overlays += S
```
MistakeNot4892: `add_overlay(S)`

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,551 @@
...
+	overlays.Cut()
+
+	if(charging)
+		var/image/I = image('icons/obj/shunt_drive.dmi', "activating")
+		var/matrix/M = new()
+		I.transform = M
+		overlays += I
+
+	if(jumping)
+		overlays += image('icons/obj/shunt_drive.dmi', "activated")
```
MistakeNot4892: `add_overlay(image('icons/obj/shunt_drive.dmi', "activated"))`

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,551 @@
...
+		if(user.skill_check(SKILL_ENGINES, SKILL_ADEPT))
+			switch(sabotaged)
+				if(SHUNT_SABOTAGE_MINOR)
+					to_chat(user, SPAN_WARNING("It looks like it's been tampered with in some way, and the accelerator vanes seem out of place."))
+				if(SHUNT_SABOTAGE_MAJOR)
+					to_chat(user, SPAN_WARNING("Light behaves oddly around the core of [src], and it looks to have been tampered with! The vanes are definitely out of place."))
+				if(SHUNT_SABOTAGE_CRITICAL)
+					to_chat(user, SPAN_DANGER("Light bends around the core of [src] in a manner that eerily reminds you of a singularity... the vanes look completely misaligned!"))
+		else
+			to_chat(user, SPAN_WARNING("It looks like it's been tampered with, but you're not sure about to what extent."))
```
MistakeNot4892: `not sure to what extent` would scan better imo.

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,551 @@
...
+
+/obj/machinery/ftl_shunt/core/attackby(var/obj/item/O, var/mob/user)
+	if(istype(O, /obj/item/stack/telecrystal))
+		var/obj/item/stack/telecrystal/TC = O
+		var/tc_input = input(user, "How many telecrystals do you want to put in?", "TC Input", 0) as num|null
+
+		if(!tc_input)
```
MistakeNot4892, line 114: Should check `user.incapacitated()`, `user.Adjacent(src)` and `(TC in user.get_held_items())` post-input.

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,551 @@
...
+			return FALSE
+
+		if(TC.amount < tc_input)
+			to_chat(user, SPAN_WARNING("You don't have enough telecrystals for that."))
+			return FALSE
+
+		if(tc_input < 10)
+			to_chat(user, SPAN_WARNING("This really isn't a useful amount for sabotage."))
```
MistakeNot4892: Not a fan of this being a gotcha, should just give them a fail message at the start if they have less than 10, and clamp it to 10 as a minimum.

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,551 @@
...
+			return FALSE
+
+		to_chat(user, SPAN_WARNING("You begin to insert the crystals into [src]..."))
```
MistakeNot4892: SPAN_NOTICE or SPAN_DANGER, SPAN_WARNING is for warnings.

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,551 @@
...
+	if(sabotaged)
+		var/mob/living/carbon/human/h_user = null
+		if (!istype(user, /mob/living/carbon/human))
+			return
+		h_user = user //this is used specifically in the do_after because electrocute_act() is not a moblevel proc.
+		var/skill_delay = user.skill_delay_mult(SKILL_ENGINES, 0.3)
+		if(!user.skill_check(SKILL_ENGINES, SKILL_BASIC))
+			to_chat(user, SPAN_DANGER("You are nowhere near experienced enough to stick your hand into that thing."))
+			return FALSE
+		to_chat(user, SPAN_NOTICE("You reach your hand inside of [src] and slowly begin to remove re-align the accelerator vanes..."))
```
MistakeNot4892: `remove re-align`

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,551 @@
...
+		if(!do_after(user, (4 SECOND * skill_delay), src))
+			to_chat(user, SPAN_WARNING("You try to pull your hand away from the vanes, but you touch a conductor!"))
```
MistakeNot4892: Should use SPAN_DANGER for this imo, as above SPAN_WARNING is for warnings.

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,551 @@
...
+			h_user.electrocute_act(rand(150,250), src, def_zone = user.get_active_held_item_slot())
+			return FALSE
+		var/obj/item/stack/telecrystal/TC = new
+		TC.amount = sabotaged_amt
+		TC.forceMove(get_turf(user))
+		user.put_in_hands(TC)
+		to_chat(user, SPAN_NOTICE("You realign the accelerator vanes by removing some kind of crystal, preventing what could have been a catastrophe."))
```
MistakeNot4892: Nit: I feel like `You remove \the [TC] from \the [src] and realign the accelerator vanes, preventing what could have been a catastrophe.` would flow a bit better. It would also not presume that the person doesn't know what a TC is IC.

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,551 @@
...
+				FP.master = src
+				fuel_ports += FP
+
+//Starts the teleport process, returns 1-6, with 6 being the all-clear.
+/obj/machinery/ftl_shunt/core/proc/start_shunt()
+	var/shunt_distance
+	var/vessel_mass = ftl_computer.linked.get_vessel_mass()
+	var/shunt_turf = locate(shunt_x, shunt_y, GLOB.using_map.overmap_z)
+
+	if(stat & (BROKEN))
```
MistakeNot4892: Why the parens around the bitflags?

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,551 @@
...
+		var/list/ports = lan.get_devices(/obj/machinery/ftl_shunt/fuel_port)
+		for(var/obj/machinery/ftl_shunt/fuel_port/FP in ports)
+			if(!FP.master)
+				FP.master = src
+				fuel_ports += FP
+
+//Starts the teleport process, returns 1-6, with 6 being the all-clear.
+/obj/machinery/ftl_shunt/core/proc/start_shunt()
+	var/shunt_distance
+	var/vessel_mass = ftl_computer.linked.get_vessel_mass()
```
MistakeNot4892, line 187: Probably should nullcheck ftl_computer and linked here?

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,551 @@
...
+		return FTL_START_FAILURE_BROKEN
+	if(stat & (NOPOWER))
+		return FTL_START_FAILURE_POWER
+
+	if(!length(fuel_ports)) //no fuel ports
+		find_ports()
+		if(!length(fuel_ports))
+			return FTL_START_FAILURE_OTHER
+
+	if(world.time <= cooldown)
```
MistakeNot4892: Probably slightly better for performance to do this simple comparison before the find_ports() checks above.

code/modules/modular_computers/file_system/programs/command/comm.dm
Mazianni: I'm not really certain myself.

code/modules/overmap/ftl_shunt/computer.dm
Mazianni: This proc should never run if either are linked, but I'll add a guard for it.

code/modules/overmap/ftl_shunt/computer.dm
Mazianni: Cost is calculated during UI update because doing it whenever the destination x/y changed was a little clunky. Jump cost is a factor of mass * distance so it is dynamic depending on how far you're jumping, so it requires updating.

code/modules/overmap/ftl_shunt/computer.dm
Mazianni: No, this returns early if you say you don't want to proceed with the jump. The interaction check presumably returns TRUE/FALSE, so.

code/modules/overmap/ftl_shunt/computer.dm
Mazianni: Implemented.

code/modules/overmap/ftl_shunt/computer.dm
Mazianni: Spans present a bit of a problem.

code/modules/overmap/ftl_shunt/core.dm
Mazianni: This is so that it finds the fuel ports, and if there's still no ports, then returns a failure.

code/modules/overmap/ftl_shunt/computer.dm
MistakeNot4892: Fair enough.

code/modules/overmap/ftl_shunt/computer.dm
MistakeNot4892: Fair enough.

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,551 @@
...
+			max_fuel = get_fuel_joules(TRUE)
+			update_icon()
+			return TRUE
+
+	. = ..()
+
+/obj/machinery/ftl_shunt/fuel_port/physical_attack_hand(var/mob/user)
+	if(fuel)
+		to_chat(user, SPAN_NOTICE("You begin to remove the fuel assembly from [src]..."))
+		if(!do_after(user, 20, src))
```
MistakeNot4892: `|| !fuel || fuel.loc != src`

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,551 @@
...
+			return FALSE
+		fuel.forceMove(get_turf(user))
```
MistakeNot4892: `fuel.dropInto(loc)`

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,551 @@
...
+	. = ..()
+	if(master)
+		master.fuel_ports -= src
+	master = null
+	QDEL_NULL(fuel)
+
+/obj/machinery/ftl_shunt/fuel_port/attackby(var/obj/item/O, var/mob/user)
+	if(istype(O, /obj/item/fuel_assembly))
+		if(!fuel)
+			if(!do_after(user, 20, src))
```
MistakeNot4892: `|| fuel`

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,551 @@
...
+				return
+			if(user && !user.unEquip(O))
```
MistakeNot4892: You can supply a second argument to unEquip() to skip needing the O.forceMove() below (`unEquip(O, src)`).

code/modules/overmap/ftl_shunt/core.dm
MistakeNot4892: This is a nitpick, but `2 SECONDS` reads nicer - there's a bunch of other times in your code that could also be turned into `SECONDS`.

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,551 @@
...
+			else
+				cancel_shunt() //Not enough fuel for whatever reason. Cancel.
+		SSradiation.radiate(src, (active_power_usage / 1000))
+
+/obj/machinery/ftl_shunt/fuel_port
+	name = "superluminal shunt fuel port"
+	desc = "A fuel port for an FTL shunt."
+	icon_state = "empty"
+
+	var/list/global/fuels = list(/decl/material/gas/hydrogen/tritium = 25000, /decl/material/gas/hydrogen/deuterium = 25000, /decl/material/gas/hydrogen = 25000, /decl/material/solid/exotic_matter = 50000)
```
MistakeNot4892: Formatting nitpick, but this would be easier to read as
````
fuels = list(
  blah,
  blah,
  blah
)
````

File: code/modules/overmap/ftl_shunt/core.dm [hunk 74]
```diff
@@ -0,0 +1,551 @@
...
+
+/obj/machinery/ftl_shunt/core/proc/get_charge_time()
+	return (ftl_computer.linked.vessel_mass * CHARGE_TIME_PER_TON) * time_multiplier
+
+/obj/machinery/ftl_shunt/core/Process()
+	if((stat & (BROKEN|NOPOWER)) & charging)
+		cancel_shunt()
+
+	if(charging)
+		chargepercent = round(100.0*(world.time - charge_started)/charge_interval, 0.1)
```
MistakeNot4892: `100.0` won't do anything in DM, every number is floating point unless you round it.

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,551 @@
...
+			ports_used++
+
+	if(ports_used == length(fueled_ports))
+		return TRUE
+
+/obj/machinery/ftl_shunt/core/proc/get_charge_time()
+	return (ftl_computer.linked.vessel_mass * CHARGE_TIME_PER_TON) * time_multiplier
+
+/obj/machinery/ftl_shunt/core/Process()
+	if((stat & (BROKEN|NOPOWER)) & charging)
```
MistakeNot4892: `&& charging` unless you're doing some weird bitwise logic here.

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,551 @@
...
+	return round(100.0*get_fuel(fuel_ports)/fuel_max, 0.1)
+
+
+/obj/machinery/ftl_shunt/core/proc/use_fuel(var/joules_req)
+	var/joules_per_port
+	var/avail_fuel = get_fuel(fuel_ports)
+	var/list/fueled_ports = list()
+	var/ports_used
+
+	if(joules_req > avail_fuel) //Not enough fuel in the system.
```
MistakeNot4892, line 438: You would avoid several var allocations if you reordered this as:
````
	var/avail_fuel = get_fuel(fuel_ports)
	if(joules_req > avail_fuel) //Not enough fuel in the system.
		return FALSE

	var/joules_per_port
	var/list/fueled_ports = list()
	var/ports_used
````

File: code/modules/overmap/ftl_shunt/core.dm [hunk 77]
```diff
@@ -0,0 +1,551 @@
...
+/obj/machinery/ftl_shunt/core/proc/get_fuel(var/list/input)
+	var/total_fuel_joules
+
+	for(var/obj/machinery/ftl_shunt/fuel_port/F in input)
+		total_fuel_joules += F.get_fuel_joules(FALSE)
+
+	return total_fuel_joules
+
+/obj/machinery/ftl_shunt/core/proc/get_fuel_maximum(var/list/input)
+	var/max_fuel_joules = 0
```
MistakeNot4892: Not blocking, but `. = 0` and just using `.` in the loop would remove the need for an explicit return and avoid a var allocation.

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,551 @@
...
+/obj/machinery/ftl_shunt/core/proc/get_status()
+	if(stat & (BROKEN|NOPOWER))
+		return FTL_STATUS_OFFLINE
+	if(cooldown)
+		return FTL_STATUS_COOLDOWN
+	else
+		return FTL_STATUS_GOOD
+
+/obj/machinery/ftl_shunt/core/proc/get_fuel(var/list/input)
+	var/total_fuel_joules
```
MistakeNot4892: As below re: `.` nitpick.

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,551 @@
...
+				A.energy_fail(rand(100,120))
+
+			for(var/mob/living/carbon/human/H in view(7)) //scary text if you're in view, because you're fucked now boy.
+				to_chat(H, SPAN_DANGER("The light around [src] warps before it emits a flash of incredibly bright, searing light!"))
+				H.flash_eyes(FLASH_PROTECTION_NONE)
+
+
+	ftl_announcement.Announce(announcetxt, "FTL Shunt Management System", new_sound = sound('sound/misc/ftlsiren.ogg'))
+
+	if(uhoh)
```
MistakeNot4892: This is so casual, I kinda love it.

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,551 @@
...
+
+			for(var/obj/machinery/power/apc/A in SSmachines.machinery) //Effect Three: shut down power across the ship.
+				if(!(A.z in ftl_computer.linked.map_z))
+					continue
+				A.energy_fail(rand(60,80))
+
+		if(SHUNT_SABOTAGE_CRITICAL)
+			announcetxt = shunt_sabotage_text_critical
+
+			uhoh = TRUE //let the fuckening begin!
```
MistakeNot4892: You don't set `uhoh` anywhere else, why not just spawn the singularity at the end of this block?

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,551 @@
...
+/obj/machinery/ftl_shunt/core/proc/handle_spacefloat(var/mob/living/carbon/human/H)
+	if(!H.check_space_footing())
+		 //Flip a coin ...
+		to_chat(H, SPAN_WARNING("Being untethered from a ship entering FTL is a bad idea, but you roll the dice..."))
+		if(prob(50))
+			to_chat(H, SPAN_NOTICE("and win, surviving the energy dancing over your body. Not unharmed, however."))
+			H.apply_damage(300, IRRADIATE, damage_flags = DAM_DISPERSED)
+			return
+		else
+			to_chat(H, SPAN_DANGER("and lose, being ripped apart in a nanosecond by energies beyond comprehension."))
```
MistakeNot4892, line 362: Could just `gib()` them to be more flexible. Also gibbing humans in space is like fireworks, the limbs go flying away in a pretty flower.

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,551 @@
...
+					L.flicker()
+			if(SHUNT_SEVERITY_MAJOR)
+				if(prob(35))
+					L.flicker()
+
+	for(var/obj/machinery/power/apc/A in SSmachines.machinery)
+		if(!(A.z in ftl_computer.linked.map_z))
+			continue
+		switch(shunt_sev)
+			if(SHUNT_SEVERITY_MAJOR)
```
MistakeNot4892, line 341: This is not really in scope I guess, but I feel like you could datumize/declize the failure states to avoid these big chunky switch statements.

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,551 @@
...
+
+	update_use_power(POWER_USE_IDLE)
+	var/destination = locate(shunt_x, shunt_y, GLOB.using_map.overmap_z)
+	var/jumpdist = get_dist(get_turf(ftl_computer.linked), destination)
+	var/obj/effect/portal/wormhole/W = new(destination) //Generate a wormhole effect on overmap to give some indication that something is about to happen.
+	QDEL_IN(W, 6 SECONDS)
+	addtimer(CALLBACK(src, .proc/do_shunt, shunt_x, shunt_y, jumpdist, destination), 6 SECONDS)
+	jumping = TRUE
+	update_icon()
+	for(var/mob/living/carbon/M in world)
```
MistakeNot4892: You marked afterthought's comment as resolved without changing this or responding.

code/modules/modular_computers/file_system/programs/command/comm.dm
```diff
@@ -116,6 +116,13 @@
 /datum/nano_module/program/comm/proc/get_shunt()
 	var/obj/comp = program.computer.get_physical_host()
 	var/obj/effect/overmap/visitable/ship/sector = map_sectors["[comp.z]"]
+
+	if(isnull(program.computer))
```
MistakeNot4892: These need to be above line 117. Also `program?.computer`.

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -475,7 +476,13 @@ obj/machinery/ftl_shunt/core/proc/fuelpercentage()
 	desc = "A fuel port for an FTL shunt."
 	icon_state = "empty"
 
-	var/list/global/fuels = list(/decl/material/gas/hydrogen/tritium = 25000, /decl/material/gas/hydrogen/deuterium = 25000, /decl/material/gas/hydrogen = 25000, /decl/material/solid/exotic_matter = 50000)
+	var/list/global/fuels = list(
+				/decl/material/gas/hydrogen/tritium = 25000,
```
MistakeNot4892: You monkey's pawed me with your malicious compliance >:(
The `pretty` list format is one tab per entry.

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -503,11 +510,10 @@ obj/machinery/ftl_shunt/core/proc/fuelpercentage()
 /obj/machinery/ftl_shunt/fuel_port/attackby(var/obj/item/O, var/mob/user)
 	if(istype(O, /obj/item/fuel_assembly))
 		if(!fuel)
-			if(!do_after(user, 20, src))
+			if(!do_after(user, 2 SECONDS, src) || fuel)
 				return
-			if(user && !user.unEquip(O))
+			if(user && !user.unEquip(O, src))
```
MistakeNot4892: Should this be `!user || !user.unEquip`?

code/modules/overmap/ftl_shunt/computer.dm
```diff
@@ -0,0 +1,147 @@
...
+
+	ui = SSnano.try_update_ui(user, src, ui_key, ui, data, force_open)
+	if (!ui)
+		ui = new(user, src, ui_key, "ftl_computer.tmpl", "[linked.name] Superluminal Shunt Control", 420, 530, src)
+		ui.set_initial_data(data)
+		ui.open()
+		ui.set_auto_update(1)
+
+/obj/machinery/computer/ship/ftl/OnTopic(var/mob/user, var/list/href_list, state)
+	if(..())
```
MistakeNot4892: You're calling `..()` twice.
```
. = ..()
if(.)
    return
```

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,556 @@
...
+		local_network.set_tag(null, initial_id_tag)
+	find_ports()
+	set_light(1, 1, 2)
+
+/obj/machinery/ftl_shunt/core/Destroy()
+	. = ..()
+	for(var/obj/machinery/ftl_shunt/fuel_port/FP in fuel_ports)
+		FP.master = null
+		fuel_ports -= FP
+	ftl_computer.linked_core = null
```
MistakeNot4892: `if(ftl_computer)`

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,556 @@
...
+		sabotaged_amt = tc_input
+		TC.use(tc_input)
+		to_chat(user, SPAN_DANGER("You successfully sabotage [src] by inserting the crystals!"))
+		return TRUE
+	. = ..()
+
+/obj/machinery/ftl_shunt/core/physical_attack_hand(var/mob/user)
+	if(sabotaged)
+		var/mob/living/carbon/human/h_user = null
+		if (!istype(user, /mob/living/carbon/human))
```
MistakeNot4892: `ishuman(user)`

MistakeNot4892: Or
````
var/mob/living/carbon/human/h_user = user
if(!istype(h_user))
    return 
````

File: code/modules/overmap/ftl_shunt/core.dm [hunk 90]
```diff
@@ -0,0 +1,556 @@
...
+			return
```
MistakeNot4892: `return TRUE`

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,556 @@
...
+		h_user = user //this is used specifically in the do_after because electrocute_act() is not a moblevel proc.
+		var/skill_delay = user.skill_delay_mult(SKILL_ENGINES, 0.3)
+		if(!user.skill_check(SKILL_ENGINES, SKILL_BASIC))
+			to_chat(user, SPAN_DANGER("You are nowhere near experienced enough to stick your hand into that thing."))
+			return FALSE
+		to_chat(user, SPAN_NOTICE("You reach your hand inside of [src] and slowly begin to re-align the accelerator vanes..."))
+		if(!do_after(user, (4 SECOND * skill_delay), src))
+			to_chat(user, SPAN_DANGER("You try to pull your hand away from the vanes, but you touch a conductor!"))
+			h_user.electrocute_act(rand(150,250), src, def_zone = user.get_active_held_item_slot())
+			return FALSE
```
MistakeNot4892: Still want to return TRUE to this method since it indicates that the action was handled, not the success or failure of the action.

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -0,0 +1,556 @@
...
+//Finds fuel ports.
+/obj/machinery/ftl_shunt/core/proc/find_ports()
+	var/datum/extension/local_network_member/network = get_extension(src, /datum/extension/local_network_member)
+	var/datum/local_network/lan = network.get_local_network()
+	if(lan)
+		var/list/ports = lan.get_devices(/obj/machinery/ftl_shunt/fuel_port)
+		for(var/obj/machinery/ftl_shunt/fuel_port/FP in ports)
+			if(!FP.master)
+				FP.master = src
+				fuel_ports += FP
```
MistakeNot4892, line 175: You should probably cut or null `fuel_ports` before this loop.

code/modules/modular_computers/file_system/programs/command/comm.dm
MistakeNot4892: Still needs a `?` in `program.computer`

MistakeNot4892 (dismissed):
Nothing is springing out at me as wrong, willing to give it a go.

code/modules/modular_computers/file_system/programs/command/comm.dm
```diff
@@ -112,6 +113,28 @@
...
+
+	if(isnull(comp))
+		return FALSE
+
+	var/obj/effect/overmap/visitable/ship/sector = map_sectors["[comp.z]"]
+
+	if(!sector)
+		return FALSE
+
+	for(var/obj/machinery/ftl_shunt/core/C in SSmachines.machinery)
```
comma, line 130: should probably be tracked by overmap object, steal how shieldgens do it

MistakeNot4892: Probably worth generalizing the method the shields use to a general proc and sharing it around.

code/modules/overmap/ftl_shunt/computer.dm
```diff
@@ -0,0 +1,148 @@
...
+/obj/machinery/computer/ship/ftl/proc/find_core()
+	if(!linked)
+		return
+
+	if(linked_core)
+		linked_core.ftl_computer = null
+		linked_core = null
+
+	for(var/obj/machinery/ftl_shunt/core/C in SSmachines.machinery)
+		if(C.z in linked.map_z)
```
comma, line 35: check out shield gens for this too because ownership can be tricky when docked overmap capable shuttles enter the scene.

afterthought2 (approved):

MistakeNot4892:
Going to clean up a few remaining issues in another PR.

PR #1254, opened 2021-02-01
Adds a verb to work around the save init issue.

MistakeNot4892:
Thanks @eckff and @afterthought2.


afterthought2 (approved):
On the "Is this code safe, does it do something understandable, is it a reasonable solution to a given problem" scale this is 0/10. But I guess you can test it out if you need it urgently.

PR #1263, opened 2021-02-02
Fixes moved event for zoom items

quardbreak:
## Changelog
:cl:
bugfix: Fixed approximation using the binoculars/scopes. Now it turns off if the source is dropped/put in a backpack/etc.
/:cl:

### DIFF
--- a/code/game/objects/item.dm
+++ b/code/game/objects/item.dm
@@ -757,7 +757,7 @@ modules/mob/living/carbon/human/life.dm if you die, you will be zoomed out.
 	user.visible_message("\The [user] peers through [zoomdevicename ? "the [zoomdevicename] of [src]" : "[src]"].")
 
 	GLOB.destroyed_event.register(src, src, /obj/item/proc/unzoom)
-	GLOB.moved_event.register(src, src, /obj/item/proc/unzoom)
+	GLOB.moved_event.register(src, src, /obj/item/proc/zoom_move)
 	GLOB.dir_set_event.register(src, src, /obj/item/proc/unzoom)
 	GLOB.item_unequipped_event.register(src, src, /obj/item/proc/zoom_drop)
 	GLOB.stat_set_event.register(user, src, /obj/item/proc/unzoom)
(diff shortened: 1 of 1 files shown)


code/game/objects/item.dm
```diff
@@ -757,21 +757,26 @@ modules/mob/living/carbon/human/life.dm if you die, you will be zoomed out.
 	user.visible_message("\The [user] peers through [zoomdevicename ? "the [zoomdevicename] of [src]" : "[src]"].")
 
 	GLOB.destroyed_event.register(src, src, /obj/item/proc/unzoom)
-	GLOB.moved_event.register(src, src, /obj/item/proc/unzoom)
+	GLOB.moved_event.register(src, src, /obj/item/proc/zoom_move)
 	GLOB.dir_set_event.register(src, src, /obj/item/proc/unzoom)
 	GLOB.item_unequipped_event.register(src, src, /obj/item/proc/zoom_drop)
```
afterthought2, line 762: I think all of these are just wrong; it should be registering for the mob's events, not its own.

PsiOmegaDelta: While it is an extra layer moved/dir events are "recursive".

If something has registered listeners and is inside another movable atom then it'll listen to that's objects corresponding events in order to raise its own events.

afterthought2: That's true for Move, but is it actually true for dir set? I don't think it changes dir recursively when its loc's dir is changed. The destroyed one is particularly concerning.

In any case, based on how it's treating the arg values passed to the procs, my point stands.

PsiOmegaDelta: I want to recall that I made the dir event recursive due assemblies and IR-based (line-of-sight) movement detectors but I digress.
It should indeed be fine to register to the mob in both the movement and dir case, as there won't be a zoom anyway otherwise.

As for the destroyed event, I don't see why it shouldn't listen to its own event. The cause might be different than the mob being deleted in which case the mob would get stuck in zoom mode.

MistakeNot4892: The current code subscribes to the holder, and works, so it seems fine to leave it alone for this PR imo.

MistakeNot4892 (dismissed):

afterthought2 (changes requested):
But it's clearly wrong; the dir set event and the destroyed event have the exact same issue that this PR is fixing in the moved event, and it can be fixed by changing the `src` to the mob in the first arg of the register/unregister calls instead of by adding an extra and unnecessary proc.

PR #1265, opened 2021-02-03
/datum/department => /decl/department

### LINKED ISSUE #90: Command is under other departments in the manifest.
#### Description of issue

Command should be in the top of the manifest

Assuming this is an oversight and not intended behavior.

Command should be on the top of the manifest, instead of on the middle.

#### Steps to reproduce

Open the crew manifest program

Since the first round.

rev

#### Issue bingo

MistakeNot4892:
- Fixes https://github.com/EscapePodNo5/EPN5-Neb/issues/90
Should resolve some runtimes to do with not properly nullchecking department references as well.


code/modules/departments/department.dm
```diff
@@ -1,15 +1,24 @@
-/datum/department
-	var/title = "Undefined"         // Player facing. Can be changed freely without breaking code or updating refrences in jobs.
-	var/reference = "undefined"     // Code facing. Jobs reference their department by this.
+/proc/cmp_departments_dsc(var/decl/department/a, var/decl/department/b)
+	. = b.display_priority - a.display_priority
+
+/decl/department
+	var/name                        // Player facing. Can be changed freely without breaking code or updating refrences in jobs.
+	var/reference                   // Code facing. Jobs reference their department by this.
```
comma: I think this is redundant, jobs should just refere to their dept with type, would be able to just get department via decl procs rather than job controller having their own version.

MistakeNot4892:
Tested jobban/unjobban and manifest, all seems to be working.

code/game/machinery/pager.dm
```diff
@@ -59,16 +59,16 @@
 		MS.send_to_department(department,"Page to <b>[location]</b> was acknowledged.", "*ack*")
 
 /obj/machinery/network/pager/medical
-	department = DEPT_MEDICAL
+	department = "Medical"
```
comma: this gotta be type not name

code/controllers/subsystems/jobs.dm
```diff
@@ -105,8 +124,10 @@ SUBSYSTEM_DEF(jobs)
 				must_fill_titles += job.title
 			if(job.department_refs)
 				for(var/dept_ref in job.department_refs)
-					if(dept_ref in SSdepartments.departments)
-						LAZYDISTINCTADD(positions_by_department[dept_ref], job.title)
+					var/decl/department/dept = SSjobs.get_department_by_reference(dept_ref)
+					if(dept)
+						LAZYDISTINCTADD(positions_by_department[dept.type], job.title)
+						LAZYDISTINCTADD(positions_by_department[dept.name], job.title)
```
comma: Doesn't need to be a thing anymore. Seemed to be mostly used by jobbans, and those use department types now.
Only place left doing this is `event_dynamic.dm` and they should be converted to using types too

code/game/antagonist/station/revolutionary.dm
```diff
@@ -27,14 +27,14 @@
 	faction = "revolutionary"
 
 	blacklisted_jobs = list(/datum/job/ai, /datum/job/cyborg)
-
+	var/command_department_id = "Command"
```
comma: Should be using a type instead

code/game/antagonist/station/loyalist.dm
```diff
@@ -23,8 +23,8 @@
 	faction_invisible = 1
 	blacklisted_jobs = list(/datum/job/ai, /datum/job/cyborg, /datum/job/submap)
 	skill_setter = /datum/antag_skill_setter/station
-
 	faction = "loyalist"
+	var/command_department_id = "Command"
```
comma: Type here too

comma (changes requested):

comma (approved):

PR #1267, opened 2021-02-03
Restore loadout icon state test

PsiOmegaDelta:
Adds a new unit test proc for doing slow, paranoid but hopefully always correct string comparisons.


comma (approved):

comma:
hopefully won't be needed with single dmi system but not all items use it

PsiOmegaDelta:
If nothing else, should the need come up the new proc can be used for other similar tests.

PR #1268, opened 2021-02-03
Ports panic bunker function from Hestia

quardbreak:
Adds the panic bunker from Hestia. It lack of SSbabysit which I personally don't like, so I ported only PB since it will be useful for hi-pop servers.

## Changelog
:cl:
rscadd: Added the panic bunker support from Hestia repository. By default, it turned off and can be toggled on in config or manually for a round.
admin: Admins can control the panic bunker state and add/revoke bypass in Server section with R_SERVER flag.
/:cl:

### DIFF
--- a/code/controllers/configuration.dm
+++ b/code/controllers/configuration.dm
@@ -236,6 +236,9 @@ var/list/gamemode_cache = list()
 	var/max_acts_per_interval = 140 //Number of actions per interval permitted for spam protection.
 	var/act_interval = 0.1 SECONDS //Interval for spam prevention.
 
+	var/panic_bunker = FALSE //is the panic bunker enabled?
+	var/panic_bunker_message = "Sorry! The panic bunker is enabled. Please head to our Discord or forum to get yourself added to the panic bunker bypass."
+
 	var/lock_client_view_x
 	var/lock_client_view_y
 	var/max_client_view_x
@@ -773,6 +776,11 @@ var/list/gamemode_cache = list()
 					config.max_acts_per_interval = text2num(value)
 				if ("act_interval")
 					config.act_interval = text2num(value) SECONDS
+
+				if("panic_bunker")
+					config.panic_bunker = TRUE
+				if("panic_bunker_message")
+					config.panic_bunker_message = value
 
 				else
 					log_misc("Unknown setting in configuration: '[name]'")
--- a/code/modules/admin/panicbunker.dm
+++ b/code/modules/admin/panicbunker.dm
@@ -1,0 +1,37 @@
+var/list/panic_bunker_bypass = list()
+
+/datum/admins/proc/panicbunker()
+	set category = "Server"
+	set name = "Toggle Panic Bunker"
+
+	if(!check_rights(R_SERVER))
+		return
+
+	config.panic_bunker = !config.panic_bunker
+
+	log_and_message_admins("[key_name(usr)] has toggled the Panic Bunker, it is now [(config.panic_bunker ? "on" : "off")]")
+	if (config.panic_bunker && (!dbcon || !dbcon.IsConnected()))
+		message_admins("The SQL database is not connected, player age cannot be checked and the panic bunker will not function until the database connection is restored.")
+
+/datum/admins/proc/addbunkerbypass(ckeytobypass as text)
+	set category = "Server"
+	set name = "Add Panic Bunker Bypass"
+	set desc = "Allows a given ckey to connect despite the panic bunker for a given round."
+
+	if(!check_rights(R_SERVER))
+		return
+
+	global.panic_bunker_bypass |= ckey(ckeytobypass)
+
+	log_admin("[key_name(usr)] has added [ckeytobypass] to the current round's bunker bypass list.")
+	message_admins("[key_name_admin(usr)] has added [ckeytobypass] to the current round's bunker bypass list.")
+
+/datum/admins/proc/revokebunkerbypass(ckeytobypass in global.panic_bunker_bypass)
+	set category = "Server"
+	set name = "Revoke Panic Bunker Bypass"
+	set desc = "Revoke's a ckey's permission to bypass the panic bunker for a given round."
+
+	global.panic_bunker_bypass -= ckey(ckeytobypass)
+
+	log_admin("[key_name(usr)] has removed [ckeytobypass] from the current round's bunker bypass list.")
+	message_admins("[key_name_admin(usr)] has removed [ckeytobypass] from the current round's bunker bypass list.")
--- a/code/modules/client/client_procs.dm
+++ b/code/modules/client/client_procs.dm
@@ -325,6 +325,13 @@
 	var/sql_computerid = sql_sanitize_text(src.computer_id)
 	var/sql_admin_rank = sql_sanitize_text(admin_rank)
 
+	if ((player_age <= 0) && !(ckey in global.panic_bunker_bypass)) //first connection
+		if (config.panic_bunker && !holder && !deadmin_holder)
+			log_adminwarn("Failed Login: [key] - New account attempting to connect during panic bunker")
+			message_admins("<span class='adminnotice'>Failed Login: [key] - New account attempting to connect during panic bunker</span>")
+			to_chat(src, config.panic_bunker_message)
+			qdel(src)
+			return 0
 
 	if(sql_id)
 		//Player already identified previously, we need to just update the 'lastseen', 'ip' and 'computer_id' variables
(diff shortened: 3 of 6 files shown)


code/modules/admin/panicbunker.dm
```diff
@@ -0,0 +1,48 @@
...
+	log_admin("[key_name(usr)] has added [ckeytobypass] to the current round's bunker bypass list.")
+	message_admins("[key_name_admin(usr)] has added [ckeytobypass] to the current round's bunker bypass list.")
+
+/datum/admins/proc/revokebunkerbypass(ckeytobypass as text)
+	set category = "Server"
+	set name = "Revoke Panic Bunker Bypass"
+	set desc = "Revoke's a ckey's permission to bypass the panic bunker for a given round."
+	if(!dbcon.IsConnected())
+		to_chat(usr, "<span class='adminnotice'>The Database is not enabled or not working!</span>")
+		return
```
afterthought2: Why make this check when we don't actually store this to the db?

code/modules/client/client_procs.dm
```diff
@@ -325,6 +325,14 @@
...
 	var/sql_admin_rank = sql_sanitize_text(admin_rank)
 
+	if ((player_age == -1) && !(ckey in global.panic_bunker_bypass)) //first connection
+		if (config.panic_bunker && !holder && !deadmin_holder)
+			log_adminwarn("Failed Login: [key] - New account attempting to connect during panic bunker")
+			message_admins("<span class='adminnotice'>Failed Login: [key] - New account attempting to connect during panic bunker</span>")
+			to_chat(src, config.panic_bunker_message)
+			qdel(src)
+			return 0
+		player_age = 0 //math requires this to not be -1.
```
afterthought2: Why are we kicking clients in the middle of a logging helper?

code/modules/client/client_procs.dm
```diff
@@ -325,6 +325,14 @@
 	var/sql_computerid = sql_sanitize_text(src.computer_id)
 	var/sql_admin_rank = sql_sanitize_text(admin_rank)
 
+	if ((player_age == -1) && !(ckey in global.panic_bunker_bypass)) //first connection
```
afterthought2: It seems this branch will never be traversed, as `player_age` is set to `0` on line 285 (unless the db returns an explicit `-1`, which also seems impossible).

config/example/config.txt
```diff
@@ -426,6 +426,9 @@ RADIATION_LOWER_LIMIT 0.15
 ## Uncomment this to modify the number of actions permitted per interval before being kicked for spam.
 #MAX_ACTS_PER_INTERVAL 140
 
+## Is the panic bunker currently on?
+#PANIC_BUNKER
```
afterthought2, line 430: Should be annotated as "on by default" or similar; it's a config setting, not a state var.

code/modules/admin/panicbunker.dm
```diff
@@ -0,0 +1,48 @@
...
+	set category = "Server"
+	set name = "Add Panic Bunker Bypass"
+	set desc = "Allows a given ckey to connect despite the panic bunker for a given round."
+
+	if(!check_rights(R_SERVER))
+		return
+
+	if(!dbcon.IsConnected())
+		to_chat(usr, "<span class='adminnotice'>The Database is not enabled or not working!</span>")
+		return
```
afterthought2: Again, why?

code/controllers/configuration.dm
```diff
@@ -236,6 +236,9 @@ var/list/gamemode_cache = list()
 	var/max_acts_per_interval = 140 //Number of actions per interval permitted for spam protection.
 	var/act_interval = 0.1 SECONDS //Interval for spam prevention.
 
+	var/panic_bunker = FALSE //is the panic bunker enabled?
+	var/panic_bunker_message = "Sorry! The panic bunker is enabled. Please head to our discord or forum to get yourself added to the panic bunker bypass."
```
afterthought2: Why is this a var on the config controller, but cannot actually be set in the config?

afterthought2 (commented):
Not a big fan of the lack of automation here.

code/controllers/configuration.dm
quardbreak: oversight, will make it configruable later.

MistakeNot4892: `Discord` needs a capital.

code/modules/admin/panicbunker.dm
```diff
@@ -0,0 +1,37 @@
...
+
+	if(!check_rights(R_SERVER))
+		return
+
+	global.panic_bunker_bypass |= ckey(ckeytobypass)
+
+	log_admin("[key_name(usr)] has added [ckeytobypass] to the current round's bunker bypass list.")
+	message_admins("[key_name_admin(usr)] has added [ckeytobypass] to the current round's bunker bypass list.")
+
+/datum/admins/proc/revokebunkerbypass(ckeytobypass as text)
```
MistakeNot4892: `ckeytobypass in global.panic_bunker_bypass` would let them just pick one to remove I believe.

code/modules/admin/panicbunker.dm
```diff
@@ -0,0 +1,37 @@
...
+	set category = "Server"
+	set name = "Toggle Panic Bunker"
+
+	if(!check_rights(R_SERVER))
+		return
+
+	config.panic_bunker = (!config.panic_bunker)
+
+	log_and_message_admins("[key_name(usr)] has toggled the Panic Bunker, it is now [(config.panic_bunker ? "on" : "off")]")
+	if (config.panic_bunker && (!dbcon || !dbcon.IsConnected()))
```
MistakeNot4892, line 13: Checking for db here again.

code/modules/admin/panicbunker.dm
```diff
@@ -0,0 +1,37 @@
+var/list/panic_bunker_bypass = list()
+
+/datum/admins/proc/panicbunker()
+	set category = "Server"
+	set name = "Toggle Panic Bunker"
+
+	if(!check_rights(R_SERVER))
+		return
+
+	config.panic_bunker = (!config.panic_bunker)
```
MistakeNot4892: Why `()`?

code/modules/admin/panicbunker.dm
MistakeNot4892: Admin note got updated after discussion on Discord, my brain is smooth

MistakeNot4892 (approved):

code/modules/admin/panicbunker.dm
```diff
@@ -0,0 +1,37 @@
...
+/datum/admins/proc/panicbunker()
+	set category = "Server"
+	set name = "Toggle Panic Bunker"
+
+	if(!check_rights(R_SERVER))
+		return
+
+	config.panic_bunker = !config.panic_bunker
+
+	log_and_message_admins("[key_name(usr)] has toggled the Panic Bunker, it is now [(config.panic_bunker ? "on" : "off")]")
```
PsiOmegaDelta, line 12: `log_and_message_admins("has toggled the Panic Bunker, it is now [(config.panic_bunker ? "on" : "off")]"`
`usr` is automatically prepended by `log_and_message_admins()`

code/modules/admin/panicbunker.dm
```diff
@@ -0,0 +1,37 @@
...
+	set name = "Add Panic Bunker Bypass"
+	set desc = "Allows a given ckey to connect despite the panic bunker for a given round."
+
+	if(!check_rights(R_SERVER))
+		return
+
+	global.panic_bunker_bypass |= ckey(ckeytobypass)
+
+	log_admin("[key_name(usr)] has added [ckeytobypass] to the current round's bunker bypass list.")
+	message_admins("[key_name_admin(usr)] has added [ckeytobypass] to the current round's bunker bypass list.")
```
PsiOmegaDelta, line 27: Replace both log lines above with
`log_and_message_admins("has added [ckeytobypass] to the current round's bunker bypass list.")`

code/modules/admin/panicbunker.dm
```diff
@@ -0,0 +1,37 @@
...
+
+/datum/admins/proc/revokebunkerbypass(ckeytobypass in global.panic_bunker_bypass)
+	set category = "Server"
+	set name = "Revoke Panic Bunker Bypass"
+	set desc = "Revoke's a ckey's permission to bypass the panic bunker for a given round."
+
+	global.panic_bunker_bypass -= ckey(ckeytobypass)
+
+	log_admin("[key_name(usr)] has removed [ckeytobypass] from the current round's bunker bypass list.")
+	message_admins("[key_name_admin(usr)] has removed [ckeytobypass] from the current round's bunker bypass list.")
```
PsiOmegaDelta, line 37: As above, replace both lines with
`log_and_message_admins("has removed [ckeytobypass] from the current round's bunker bypass list.")`

PR #1269, opened 2021-02-03
Shuffles around random pistol icon generation

comma:
Now all parts are in separate dmis for less icon_state manipulation, all parts now just use base icon state of the item.
Removed some useless icon_state overrides on pistols and projecile guns.

## Changelog
Nothing visble


code/modules/projectiles/guns/projectile/random_pistol.dm
```diff
@@ -58,12 +44,24 @@
 		ICON_STATE_WORLD = list(17, 15),
 		ICON_STATE_INV = list(17, 17)
 	)
+	var/list/handle_icons = list(
```
MistakeNot4892, line 47: static/global?

comma: it's a decl anyway, and in theory can be overriden for different looks sets (e.g. handle and front bits melded)

MistakeNot4892 (approved):

PR #1270, opened 2021-02-03
Changed gun safety overlays to a little padlock.

MistakeNot4892:
image
Previously they were codersprites.


comma (approved):
still codersprites

PR #1275, opened 2021-02-04
Basic functionality of breath-holding

Alceris:
Adds the basic functionality of breath-holding. Currently, a human can hold their breath until they pass out in about 4 minutes and cause significant brain damage to themself.

### DIFF
--- a/code/modules/mob/living/carbon/breathe.dm
+++ b/code/modules/mob/living/carbon/breathe.dm
@@ -4,14 +4,20 @@
 
 //Start of a breath chain, calls breathe()
 /mob/living/carbon/handle_breathing()
-	if((life_tick % MOB_BREATH_DELAY) == 0 || failed_last_breath || is_asystole()) //First, resolve location and get a breath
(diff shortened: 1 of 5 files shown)


Mazianni:
This should depend on the Athletics skill (SKILL_HAULING).

MistakeNot4892:
@comma and myself are against this implementation (other devs are neutral) being merged for a few reasons that have been discussed previously on Discord, primarily that the logic is too high up in the breathing flow, and should be integrated into the lungs rather than doing several situational value changes throughout the mob breathing logic (ex. https://github.com/MistakeNot4892/Nebula/commit/b7bc080d743bc8385047a221b7052143b53e9e6c).

I am also very leery of the magic numbers used for `holding_breath` and generally feel like this is making some already labyrinthine and brittle code even more fragile for a pretty tenuous benefit.

As the PR has been left sitting for ages at this point without resolution, and as we've discussed this before on Discord, I'm going to bite the bullet and close the PR. Please feel free to open a new PR if the concerns above are addressed.

PR #1282, opened 2021-02-06
Refactoring reagent metabolism to be more compatible with /mob/living.

MistakeNot4892:
Moves some reagent handling procs down to /mob/living and swaps out usage of member variables to use them. Tested with humans and slimes, works fine.

### DIFF
--- a/code/modules/mob/living/life.dm
+++ b/code/modules/mob/living/life.dm
@@ -48,6 +48,46 @@
 	SHOULD_CALL_PARENT(TRUE)
 	chem_effects = null
 
+	// TODO: handle isSynthetic() properly via Psi's metabolism modifiers for contact reagents like acid.
+	if((status_flags & GODMODE) || isSynthetic())
+		return FALSE
+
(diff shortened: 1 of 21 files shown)


code/modules/mob/living/life.dm
```diff
@@ -47,6 +47,25 @@
 /mob/living/proc/handle_chemicals_in_body()
 	SHOULD_CALL_PARENT(TRUE)
 	chem_effects = null
+	if(status_flags & GODMODE)
+		return FALSE
+	if(isSynthetic())
+		return FALSE
+	var/datum/reagents/metabolism/ingested_reagents = get_ingested_reagents()
```
comma: idk if there's point in getting it here if you check in metabolize_ingested_reagents() anyway

MistakeNot4892: I ended up changing that block to retrieve the datums via `metabolize_x_reagent()` because it then checks them for `chem_dose` updates.

comma (dismissed):

comma (approved):

PR #1296, opened 2021-02-08
Upstream merger

AlexMorgan3817:
Now you can update fork just running this workflow

### DIFF
--- a/.github/workflows/upstream_merge.yml
+++ b/.github/workflows/upstream_merge.yml
@@ -8,7 +8,7 @@ on:
         required: true
         default: 'NebulaSS13/Nebula'
       Branch:
-        description: 'Branch of upstrem to pull'
+        description: 'Branch of upstream to pull'
         required: true
         default: dev
 
--- a/.github/workflows/upstream_merge.yml
+++ b/.github/workflows/upstream_merge.yml
@@ -2,11 +2,20 @@ name: Merge
 
 on:
   workflow_dispatch:
+    inputs:
+      Upstream:
+        description: 'Upstream to pull'
+        required: true
+        default: 'NebulaSS13/Nebula'
+      Branch:
+        description: 'Branch of upstrem to pull'
+        required: true
+        default: dev
 
 jobs:
   Merge:
     runs-on: ubuntu-latest
-    if: github.repository != 'NebulaSS13/Nebula'
+    if: github.repository != ${{ github.event.inputs.Upstream }}
     steps:
       - name: Checkout
         uses: actions/checkout@v1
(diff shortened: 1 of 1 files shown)


AlexMorgan3817:
Ready for reviews

.github/workflows/upstream_merge.yml
```diff
@@ -0,0 +1,38 @@
...
+
+on:
+  workflow_dispatch:
+    inputs:
+      Upstream:
+        description: 'Upstream to pull'
+        required: true
+        default: 'NebulaSS13/Nebula'
+      Branch:
+        description: 'Branch of upstrem to pull'
```
MistakeNot4892: `upstream`

MistakeNot4892:
Does this work on your fork? We had to do all kinds of silly shit to get the changelog workflow working properly due to Github repository write permissions.

.github/workflows/upstream_merge.yml
```diff
@@ -0,0 +1,38 @@
...
+        run: |
+          git remote add upstream https://github.com/${{ github.event.inputs.Upstream }}
+          git fetch --all
+      - name: Merge
+        run: |
+          git config --local user.email "action@github.com"
+          git config --local user.name "Upstream Merger"
+          git merge upstream/${{ github.event.inputs.Branch }}
+      - name: Push
+        uses: ad-m/github-push-action@master
```
PsiOmegaDelta, line 34: Isn't the convention to use a specified release version, and then rely on Dependabot or the like to handle (most of) the update process?

.github/workflows/upstream_merge.yml
AlexMorgan3817: upsteam

AlexMorgan3817:
> Does this work on your fork? We had to do all kinds of silly shit to get the changelog workflow working properly due to Github repository write permissions.

Yes, but have one problem, you need to manually merge if upstream changed any of .github/workflows/*

.github/workflows/upstream_merge.yml
AlexMorgan3817: What?

AlexMorgan3817: Sooo?

PsiOmegaDelta: One head over to https://github.com/ad-m/github-push-action
One find the releases page https://github.com/ad-m/github-push-action/releases
Pick the latest relevant release and use that. In this case: ad-m/github-push-action@v0.6.0

This avoids issues with hostile and/or breaking changes when/if the workflow is updated

Then one optionally setup a Dependabot workflow.
This bot automatically opens PRs when new releases of referenced workflows are released.

.github/workflows/upstream_merge.yml
```diff
@@ -0,0 +1,38 @@
+name: Merge
```
MistakeNot4892, line 1: As discussed, please leave a comment here to tell downstream that this workflow can't be used with a strict merge policy (like Neb main uses). Something like: `This workflow cannot run if your fork uses review-based or CI-based branch protection (such as requiring a review before a merge), as the Github bot will not be able to bypass the protections to push directly to master.`

afterthought2:
Feel like including this commits us to maintaining it and providing support, which is not a commitment I'm happy making. Perhaps best left downstream.

MistakeNot4892:
After some maint discussion, the general consensus is that this isn't really needed on main Neb. The workflow is essentially an alternative way of opening a merge request, but can't run on the main repository or any repository using protected branches, so it will never be used on Neb main, and has limited use otherwise.
Thanks for opening the PR regardless.

PR #1297, opened 2021-02-08
Door machinery interactions tweak

### LINKED ISSUE #1274: Doors hacking is too easy now
#### Description of issue
You don't actually have to hack them, you just unscrew panel, take out power components, and crowbar.

Actual behavior:
Can bypass doors with simple construction steps
Expected behavior:
Bypassing door without proper access requires hacking or breaking it down

#### Steps to reproduce
1. Open hatch with screwdriver
2. Remove tesla part with wrench
3. Crowbar door open

Any airlock

Probably since doors were fully machinery part'd

#### Issue bingo

comma:
Fixes #1274

Doors now use special states a-la APCs.
Screwdriver on closed panel exposes wires for hacking etc.
Crowbar on welded airlock opens the panel for component access / deconstruction.

Panel opening is prevented by airlock circuit's access. Can swipe ID at the door to toggle the hatch locking (doesn't grant door passing access).
Also made circuit inherit access from door when door gets it from area.

Opening as draft cause  I have little idea what am I doing with machinery states.
Also the UX and interactions could be probably improved since currently chain is weird (e.g. you open panel with crowbar, but close with screwdriver, and requirement for welding)

## Changelog
:cl:
tweak: Airlock tool interactions changed.
tweak: Screwdriver on airlock will open wires panel, letting you access wires for hacking.
tweak: Crowbarring a secured (welded/bolted/braced) airlock will open the hatch letting you access components for repairs and deconstruction.
tweak: Hatch can only be opened if you have appropriate access to the door. Use ID on the door to toggle the lock.
/:cl:

### DIFF
--- a/code/game/machinery/doors/_door.dm
+++ b/code/game/machinery/doors/_door.dm
@@ -100,6 +100,9 @@
 	update_nearby_tiles(need_rebuild=1)
 	if(autoset_access) // Delayed because apparently the dir is not set by mapping and we need to wait for nearby walls to init and turn us.
 		inherit_access_from_area()
+		var/obj/item/stock_parts/circuitboard/airlock_electronics/circuit = get_component_of_type(/obj/item/stock_parts/circuitboard/airlock_electronics)
+		if(circuit)
+			circuit.req_access = req_access?.Copy()
 
 /obj/machinery/door/Destroy()
 	set_density(0)
@@ -536,6 +539,11 @@
 	if((!requiresID() || allowed(null)) && can_open_manually)
 		toggle()
 	return TRUE
+
+/obj/machinery/door/components_are_accessible(var/path)
+	if(ispath(path, /obj/item/stock_parts/circuitboard))
+		return TRUE
+	return ..()
 
 // Public access
 
--- a/code/game/machinery/doors/airlock_electronics.dm
+++ b/code/game/machinery/doors/airlock_electronics.dm
@@ -103,6 +103,16 @@
 			conf_access -= access
 		return TOPIC_REFRESH
 
+/obj/item/stock_parts/circuitboard/airlock_electronics/attackby(obj/item/W, mob/user)
+	if(istype(W.GetIdCard(), /obj/item/card/id))
+		if(!locked || check_access(W))
+			locked = !locked
+			to_chat(user, SPAN_NOTICE("[src] beeps as its hatch access is [locked ? null : "un"]locked."))
+		else if (locked)
+			to_chat(user, SPAN_WARNING("[src] buzzes, denying access to the circuitry."))
+		return TRUE
+	. = ..()
+
 /obj/item/stock_parts/circuitboard/airlock_electronics/secure
 	name = "secure airlock electronics"
 	desc = "designed to be somewhat more resistant to hacking than standard electronics."
--- a/code/datums/wires/airlock.dm
+++ b/code/datums/wires/airlock.dm
@@ -45,7 +45,7 @@ var/const/AIRLOCK_WIRE_SPEAKER = 4096
 		if(A.isElectrified())
 			if(A.shock(L, 100))
 				return 0
-	if(A.panel_open)
+	if(istype(A.construct_state, /decl/machine_construction/default/panel_closed/door/hacking))
 		return 1
 	return 0
 
--- a/code/game/machinery/_machines_base/machine_construction/airlock.dm
+++ b/code/game/machinery/_machines_base/machine_construction/airlock.dm
@@ -1,0 +1,50 @@
+/decl/machine_construction/default/panel_closed/door
+	needs_board = "door"
+	down_state = /decl/machine_construction/default/panel_open/door
+	var/hacking_state = /decl/machine_construction/default/panel_closed/door/hacking
+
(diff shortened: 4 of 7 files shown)


afterthought2:
As discussed on discord: this is maybe fine, but seems better to use the part intended for this, `/obj/item/stock_parts/access_lock`.

comma:
If possible I'd prefer to do that in separate PR cause it's effort and it's a fairly nasty issue in actual play with antags etc since it makes break-ins completely trivial.

code/game/machinery/doors/_door.dm
```diff
@@ -100,6 +100,9 @@
 	update_nearby_tiles(need_rebuild=1)
 	if(autoset_access) // Delayed because apparently the dir is not set by mapping and we need to wait for nearby walls to init and turn us.
 		inherit_access_from_area()
+		var/obj/item/stock_parts/circuitboard/airlock_electronics/circuit = get_component_of_type(/obj/item/stock_parts/circuitboard/airlock_electronics)
+		if(circuit)
+			circuit.req_access = req_access.Copy()
```
MistakeNot4892: `req_access?.Copy()`

MistakeNot4892 (dismissed):
As an emergency measure I think this is fine, it's at least closer to the old behavior.
For reference we have some folks willing to do playtesting of the codebase next week so having this merged would help them out.

MistakeNot4892 (dismissed):

code/game/machinery/_machines_base/machinery_components.dm
```diff
@@ -219,6 +219,8 @@ GLOBAL_LIST_INIT(machine_path_to_circuit_type, cache_circuits_by_build_path())
 /obj/machinery/proc/components_are_accessible(var/path)
 	if(ispath(path, /obj/item/stock_parts/access_lock))
 		return TRUE
+	if(ispath(path, /obj/item/stock_parts/circuitboard/airlock_electronics))
+		return TRUE
```
afterthought2: Please not on the base `/machinery`. Also, what does this even do? They don't have forwarded interactions with panel closed.

EDIT: I see why now, but am not convinced you should be doing this. I'd just check access in `cannot_transition_to` or within the attackby code.

code/game/machinery/doors/airlock_electronics.dm
```diff
@@ -103,6 +103,15 @@
 			conf_access -= access
 		return TOPIC_REFRESH
 
+/obj/item/stock_parts/circuitboard/airlock_electronics/attackby(obj/item/W, mob/user)
+	if(istype(W.GetIdCard(), /obj/item/card/id/))
```
afterthought2: Omit terminal `/`.

code/game/machinery/doors/airlock_electronics.dm
```diff
@@ -103,6 +103,15 @@
...
+		if(!locked || check_access(W))
+			locked = !locked
+			to_chat(user, SPAN_NOTICE("[src] beeps as its hatch access is [locked ? null : "un"]locked."))
+		else if (locked)
+			to_chat(user, SPAN_WARNING("[src] buzzes, denying access to the circuitry."))
+		return TRUE
+	. = ..()
 /obj/item/stock_parts/circuitboard/airlock_electronics/secure
```
afterthought2, line 116: Missing whitespace.

code/game/machinery/doors/airlock_electronics.dm
```diff
@@ -103,6 +103,15 @@
 			conf_access -= access
 		return TOPIC_REFRESH
 
+/obj/item/stock_parts/circuitboard/airlock_electronics/attackby(obj/item/W, mob/user)
```
afterthought2, line 106: If this is meant as a passthrough interaction for doors, isn't it just confusing as it conflicts with access checking to open the door? Why not just check access when opening the panel and leave it at that?

Also letting it reset to locked with _any_ ID, including ones that lack access, seems not user friendly.

afterthought2 (commented):
One other note: if I understand, the interaction that closes the panel (down_state -> up_state) is still screwdriver, which may (or may not) be confusing.

Doors have about 50 million interactions in their attackby, so I'm assuming nothing here conflicts but can't be sure.

afterthought2:
After discord discussion am fine with current implementation, other than technical comments. Could use some rethinking at higher level at some point.

code/game/machinery/_machines_base/machinery_components.dm
comma: Moved to door/ one

code/game/machinery/doors/airlock_electronics.dm
comma: Removed

code/game/machinery/doors/airlock_electronics.dm
comma: Added

code/game/machinery/doors/airlock_electronics.dm
comma: ftr resetting to locked with any ID I did so engineers can finish repairs without grabbing someone with access.

afterthought2 (approved):

PR #1300, opened 2021-02-08
Updates AM rifles/ammo caliber and spelling.

MistakeNot4892:
15mmR seems to be the caliber used for hunting with long rifles, not putting holes in tanks. Also the consistent misspellings bugged me.


Techhead0:
A 15 mm rifle round is absolutely *not* a hunting round, IDK where you found that. A slug round for 20 gauge shotguns (which are 15.6 mm) might be used for hunting, but a rifle round of that size is definitely anti-materiel.

And a .50 BMG for comparison is 12.7mm×99mm

It doesn't really matter in the long run, and .50 BMG is a recognizable modern round, but I figured I'd chip in my 2c

MistakeNot4892:
We had a discussion on Discord about this, tldr I think it's better to name it after an immediately identifiable round rather than assuming the person reading will be able to identify the caliber (because I definitely couldn't, and the person who brought it up initially couldn't either).

MistakeNot4892:
After further discussion, using metric calibers that include length and diameter for rifle rounds to keep them visually distinct without using a RL brand of ammo.

comma (approved):

PR #1307, opened 2021-02-10
Stop space carp (and trees?) instantly winning all melee combat.

Alceris:


afterthought2:
Either delete the code or leave it be; don't comment it out.

MistakeNot4892 (approved):

PR #1309, opened 2021-02-11
Added a mouseover highlight preference.

MistakeNot4892:
Allows you to mouseover items and machines adjacent to your mob to get an over-darkness outlined version. Preferences can be set to always, only when holding shift, and never. Good for visibility, blindness RP, and generally making the game feel slightly less opaque about what you're interacting with.

~~However it seems that client images ignore pixel offsets, so stuff like air alarms and APCs look strange. Same with items placed on a table. Not sure how to fix this - it seems to be a limitation of DM.~~ Many thanks to Crazah on the BYOND Discord for suggesting using a dummy object and vis_contents.

https://user-images.githubusercontent.com/2468979/107604981-e007f480-6c85-11eb-9080-bf071474a47a.mp4

:cl:
rscadd: There is now a Mouseover Highlight preference that can be set to Show, Hide or Show While Shift Held. This will highlight the object you are currently mousing over. You can set colour and alpha for the highlight in your UI preferences.
/:cl:

### DIFF
--- a/code/modules/client/preference_setup/global/01_ui.dm
+++ b/code/modules/client/preference_setup/global/01_ui.dm
@@ -5,8 +5,10 @@ var/list/valid_icon_sizes = list(32, 48, 64, 96, 128)
 	var/ooccolor = "#010000" //Whatever this is set to acts as 'reset' color and is thus unusable as an actual custom color
 	var/icon_size = 48
 	var/UI_style = "Midnight"
-	var/UI_style_color = "#ffffff"
-	var/UI_style_alpha = 255
+	var/UI_style_alpha =     255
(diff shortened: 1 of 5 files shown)


MistakeNot4892:
Finally got this as good as I think I will get it, ready to go.

comma (dismissed):
IMO the on-shift setting should be the default but otherwise works.

MistakeNot4892:
Shift is now the default.

comma (dismissed):

afterthought2 (dismissed):
So I'm in favor of this feature and the code looks OK, but I haven't worked with these client events so can't comment much on the code.

comma (approved):

PR #1315, opened 2021-02-11
Connecting locally grants admin rights.

MistakeNot4892:
Idea stolen from /tg/. Not sure if it should be enabled by default or not - on the one hand, enabling it by default means you don't need to mess with config at all. But on the other it's potentially a security hole.


comma (dismissed):

code/modules/client/client_procs.dm
```diff
@@ -1,3 +1,8 @@
+var/list/localhost_addresses = list(
+	"127.0.0.1" = TRUE,
+	"::1" = TRUE
```
Lohikar, line 3: BYOND doesn't support IPv6, so this is kinda pointless.

config/example/config.txt
```diff
@@ -396,6 +396,9 @@ STARLIGHT 0
 ## Uncomment to enable auto-stealthing staff who are AFK for more than specified minutes
 # AUTOSTEALTH 30
 
+## Set to 0/1 to disable/enable automatic admin rights for users connecting from the host the server is running on.
+AUTO_LOCAL_ADMIN 0
```
Lohikar, line 400: Having this disabled by default kinda makes the feature pointless, no?

code/modules/client/client_procs.dm
MistakeNot4892: It's also harmless though. Call it future proofing.

config/example/config.txt
MistakeNot4892: As discussed on Discord I've tweaked things a bit to make it less pointless.

MistakeNot4892:
Ok so:
- It will default to enabled in code, but disabled in example config. So if they take the 5 seconds to set up config, as you would on a prod server, it will disable local connections and prevent any security issues with it. If you're setting up config presumably you also have the capacity to add yourself to admins.txt or the admin DB.
- It will also skip guest checks and player cap checks, meaning no more forgetting to start BYOND before trying to test something and getting told guests are banned.

comma (approved):

Lohikar (approved):

PR #1318, opened 2021-02-12
Sets the DB encoding to utf8mb4 during connection

quardbreak:
We unable to make full Unicode DB support because during connection BYOND or DB thinks what we use different encoding.
Credits to bear1ake, who did this solution for Infinity Station bay fork.

This fixes this issue, which can appear during DB query insert:
изображение

### DIFF
--- a/code/game/world.dm
+++ b/code/game/world.dm
@@ -645,6 +645,12 @@ proc/setup_database_connection()
 	dbcon.Connect("dbi:mysql:[db]:[address]:[port]","[user]","[pass]")
 	. = dbcon.IsConnected()
 	if(.)
+		// Setting encoding and comparison (4-byte UTF-8) for the DB server ~bear1ake
+		var/DBQuery/unicode_query = dbcon.NewQuery("SET NAMES utf8mb4 COLLATE utf8mb4_general_ci")
+		if(!unicode_query.Execute())
+			global.failed_db_connections++
+			to_world_log(unicode_query.ErrorMsg())
+			return
 		global.failed_db_connections = 0	//If this connection succeeded, reset the failed connections counter.
 	else
 		global.failed_db_connections++		//If it failed, increase the failed connections counter.


PsiOmegaDelta:
I'll note that:
* All individual tables/columns should also be updated to use the appropriate character set
* Existing data is (typically) not updated either, and needs to be converted manually
* Can cause issues with non-UTF input - Nebula (and downstreams) will have to use 513 to avoid issues and there should probably be a check added to refuse to run on 512 or earlier (such a check should already exist somewhere already, for an older version)

Lohikar:
I've run a pre-513 SS13 server with utf8mb4 for a while, about the only issues in practice _for latin1_ are around \proper or \improper -- values containing them are rejected by the DB.

Spookerton:
you may find it useful to explore `utf8mb4_unicode_520_ci` as a more up to date collation that uses unicode 5.2 character weighting instead of 4.0 (or non-spec with sorting oddness in _general), available since mysql 5.6 and mariadb 10.2

mysql has also supported improved performance, 9.0 weighting, and accent insensitivity in queries through `utf8mb4_0900_ai_ci` since initial 8.0, but it's sadly [still not available](https://jira.mariadb.org/browse/MDEV-11256) in current (10.5.8) mariadb

ed: There's good discussion of what and why in the description of [this issue ](https://issues.ibexa.co/browse/EZP-29810) from elsewhere.

afterthought2 (approved):
Tentatively approving; @PsiOmegaDelta, @Lohikar, request changes or advise on alternatives you'd find preferable. Do we need to support some conversion setup? Do we want to use a different encoding as suggested in the comments?

Andrew-Fall:
Honestly there's no server's currently using the databases as far as I'm aware, it might just be better to change the migrate files

Lohikar:
There is a server or two running MySQL, but changing charset is pretty easy - I think just changing the existing migration files would be fine.

One of said servers is already using utf8mb4, so won't be impacted anyways.

PR #1320, opened 2021-02-12
adds good faith license decl

Mazianni:
we tried really, really hard to contact you about your car's extended warranty.

https://www.youtube.com/watch?v=0wtOpG54rEU

### DIFF
--- a/code/datums/licences/license.dm
+++ b/code/datums/licences/license.dm
@@ -31,3 +31,7 @@
 	name = "Grandfathered In"
 	url = "https://en.wikipedia.org/wiki/Grandfather_clause"
 	attribution_mandatory = FALSE
+
+/decl/license/good_faith
+	name = "Good Faith Usage"
+	url = "https://en.wikipedia.org/wiki/Good_faith_(law)"


MistakeNot4892 (dismissed):

code/datums/licences/license.dm
```diff
@@ -31,3 +31,9 @@
 	name = "Grandfathered In"
 	url = "https://en.wikipedia.org/wiki/Grandfather_clause"
 	attribution_mandatory = FALSE
+
+/decl/license/good_faith
+	name = "Good Faith Usage"
+	url = "https://en.wikipedia.org/wiki/Good_faith_(law)"
+	attribution_mandatory = FALSE
```
PsiOmegaDelta, line 38: I'd say that we do want attribution to be mandatory, especially if we're going with good faith

PsiOmegaDelta: I'd say that we do want attribution to be mandatory, especially if we're going with the good faith angle

PsiOmegaDelta (dismissed):

MistakeNot4892 (dismissed):

PsiOmegaDelta (approved):

PR #1326, opened 2021-02-13
Refactoring status values into a unified system.

MistakeNot4892:
- /mob/living tracks status condition counters in a list.
- Status conditions are indexed by a decl that also holds the behavior that should be applied when they change or process.
- Status conditions are set or adjusted with SET_STATUS_MAX and ADJ_STATUS macros.

This doesn't currently change anything about the timing or effects of status conditions. It does change where SSD is applied and fiddles with how some eye damage recovery happens.

:cl:
tweak: The status condition backend (paralyzed, asleep, etc) has been completely rewritten; please report any inconsistencies or bugs.
/:cl:


MistakeNot4892:
As far as I can tell this is all working. I'm going to call it ready for review and worry about reworking status checks and Life() after merge.

afterthought2 (dismissed):

afterthought2 (approved):

PR #1328, opened 2021-02-13
fixes projectile angles

silicons:
yeah uh idk why it was using get_angle but that isn't it, we don't care about pixel values for projectiles, only true movement positiosn.

## Changelog
:cl:
bugfix: projectiles use a better get angle function now 
/:cl:


Mazianni:
Need to set obj/item/projectile's randpixel var to FALSE.

MistakeNot4892 (approved):
@eckff's PR solves the other issue that needed looking at so this is good. Thanks.

PR #1337, opened 2021-02-14
Adds enhanced error handler/viewer, thanks very much Lohikar.

MistakeNot4892:
Primary benefit of this is to show proper runtimes before the error handler is initialized.


comma:
`code/unit_tests/~unit_test_subsystems.dm:135:error: GLOB.total_runtimes: undefined var`

comma (approved):

PR #1338, opened 2021-02-14
Fixes some antag spawn templates, again

quardbreak:
Wonder why they're not unit-testing in CI


MistakeNot4892:
Commit was merged in #1341

PR #1342, opened 2021-02-15
Added glide smoothing.

MistakeNot4892:
Adds adaptive glide_size that smooths out your tile movements. By default it includes a one tick delay at the end of your glide size, which causes a pause between each move (nearly identical to current movement behavior). If configured to 0, it will completely smooth your movements between tiles.

### DIFF
--- a/code/datums/movement/mob.dm
+++ b/code/datums/movement/mob.dm
@@ -131,9 +131,10 @@
 	var/next_move
 
 /datum/movement_handler/mob/delay/DoMove(var/direction, var/mover, var/is_external)
-	if(is_external)
-		return
-	next_move = world.time + max(1, mob.movement_delay())
+	if(!is_external)
+		var/delay = max(1, mob.movement_delay())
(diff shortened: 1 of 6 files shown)


comma (approved):

code/datums/movement/mob.dm
```diff
@@ -131,9 +131,10 @@
 	var/next_move
 
 /datum/movement_handler/mob/delay/DoMove(var/direction, var/mover, var/is_external)
-	if(is_external)
-		return
-	next_move = world.time + max(1, mob.movement_delay())
+	if(!is_external)
+		var/delay = max(1, mob.movement_delay())
+		next_move = world.time + delay
+		mob.glide_size = ADJUSTED_GLIDE_SIZE(delay)
```
PsiOmegaDelta, line 137: The previous early exit reduced the general indentation level

MistakeNot4892: But it added a line!

PR #1348, opened 2021-02-17
Removing of using second var in 'decls_repository.get_decl()'

AlexMorgan3817:
Why we need to assign _dot_, if futher in code we create another **var** to access it's proc?
I don't know, so now we not assign _dot_ and use this **var** instead

### DIFF
--- a/code/datums/repositories/decls.dm
+++ b/code/datums/repositories/decls.dm
@@ -29,14 +29,14 @@ var/repository/decls/decls_repository = new()
 
 /repository/decls/proc/get_decl(var/decl_type)
 	ASSERT(ispath(decl_type))
-	. = fetched_decls[decl_type]
-	if(!.)
-		. = new decl_type()
-		fetched_decls[decl_type] = .
+	var/decl/decl = fetched_decls[decl_type]
+	if(!decl)
+		decl = new decl_type()
+		fetched_decls[decl_type] = decl
 
-		var/decl/decl = .
 		if(istype(decl))
 			decl.Initialize()
+	return decl
 
 /repository/decls/proc/get_decls(var/list/decl_types)
 	. = list()


code/datums/repositories/decls.dm
```diff
@@ -29,14 +29,14 @@ var/repository/decls/decls_repository = new()
 
 /repository/decls/proc/get_decl(var/decl_type)
 	ASSERT(ispath(decl_type))
-	. = fetched_decls[decl_type]
-	if(!.)
-		. = new decl_type()
-		fetched_decls[decl_type] = .
+	var/decl/decl = fetched_decls[decl_type]
+	if(!decl)
```
AlexMorgan3817, line 33:
```suggestion
	if(!istype(decl))
```

MistakeNot4892:
This is adding an unnecessary var declaration. `.` doesn't need to be declared, it already exists.
The istype() is also unnecessary due to the `ASSERT()` on the first line. You could safely remove the second istype().

EDIT: For clarity, using `.` and then checking it means that the decl var is only declared the first time something tries to retrieve the decl. Your code makes it declared every time.

comma:
Yeah, this seems pretty pointless, currently it only declares a var as needed if decl is not inited already, with this change it does it every time.
Either way it's... really not hot code, so microoptimizations like that are not going to help in any way, even if it was working.

AlexMorgan3817:
> This is adding an unnecessary var declaration. `.` doesn't need to be declared, it already exists.

But not assigned, i don't talk about declarations
> The istype() is also unnecessary due to the `ASSERT()` on the first line. You could safely remove the second istype().

Where is second...? Loaf, github allow to comment code, not only PR

Lohikar:
Yeah, this change is pretty much completely pointless. This proc's not hot enough for microoptimizations to matter anyways.

code/datums/repositories/decls.dm
```diff
@@ -29,14 +29,14 @@ var/repository/decls/decls_repository = new()
 
 /repository/decls/proc/get_decl(var/decl_type)
 	ASSERT(ispath(decl_type))
-	. = fetched_decls[decl_type]
-	if(!.)
-		. = new decl_type()
-		fetched_decls[decl_type] = .
+	var/decl/decl = fetched_decls[decl_type]
```
AlexMorgan3817, line 32:
```suggestion
	. = fetched_decls[decl_type]
```

code/datums/repositories/decls.dm
AlexMorgan3817, line 33:
```suggestion
	if(!.)
```

code/datums/repositories/decls.dm
```diff
@@ -29,14 +29,14 @@ var/repository/decls/decls_repository = new()
...
+	if(!decl)
+		decl = new decl_type()
```
AlexMorgan3817, line 34:
```suggestion
		var/decl/decl = new decl_type()
```

code/datums/repositories/decls.dm
```diff
@@ -29,14 +29,14 @@ var/repository/decls/decls_repository = new()
...
+		fetched_decls[decl_type] = decl
 
-		var/decl/decl = .
 		if(istype(decl))
 			decl.Initialize()
+	return decl
```
AlexMorgan3817, line 39:
```suggestion
```

AlexMorgan3817:
> Yeah, this change is pretty much completely pointless. This proc's not hot enough for microoptimizations to matter anyways.

No matter how hot is thing, bad things must die :0

AlexMorgan3817:
Blin, can't apply suggestions from phone

AlexMorgan3817:
Wait? Loaf...

AlexMorgan3817:
Do not hurry  Lohikar

MistakeNot4892:
Your code is slower than the existing code, it is not a fix.

PR #1353, opened 2021-02-17
Micro-optimizes get_decl() for materials.

MistakeNot4892:
`/repository/decls` will pre-fetch the entire `/decl/material` tree in `New()`, which allows materials to be fetched with a simple macro. Given how often get_decl() is used for materials I think this should be worthwhile even if it's a bit more brittle.


Andrew-Fall:
Why not GET_DECL_FAST? Easier to say

MistakeNot4892:
I didn't close this PR. Github why.

Techhead0:
Github autocloses PRs w/ 0 line changes, I guess you may have force-pushed the wrong commit

MistakeNot4892:
Yeah I reset to origin on autopilot I think.

MistakeNot4892:
For reference, months later, the `GET_DECL` macro removed the benefit/need for this optimization, hence it was never reopened.

PR #1354, opened 2021-02-17
Fixed leech spawns, fixed overmap shield using SSmachinery list.

MistakeNot4892:
Leech fix doesn't work, dunno how to test shield fix. Going to bed.

### DIFF
--- a/code/modules/shield_generators/shield_generator.dm
+++ b/code/modules/shield_generators/shield_generator.dm
@@ -9,6 +9,7 @@
 	wires = /datum/wires/shield_generator
 	uncreated_component_parts = null
 	stat_immune = 0
+
 	var/list/field_segments = list()	// List of all shield segments owned by this generator.
 	var/list/damaged_segments = list()	// List of shield segments that have failed and are currently regenerating.
 	var/shield_modes = 0				// Enabled shield mode flags
@@ -40,6 +41,8 @@
 	var/spinup_delay      = 20
 	var/spinup_counter    = 0
 
+	var/obj/effect/overmap/visitable/last_linked_overmap_object
+
 /obj/machinery/power/shield_generator/on_update_icon()
 	if(running)
 		icon_state = "generator1"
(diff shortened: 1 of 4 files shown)


code/modules/shield_generators/shield_generator.dm
```diff
@@ -48,22 +51,36 @@
...
+	. = ..()
+	update_overmap_shield_list()
 
 /obj/machinery/power/shield_generator/Destroy()
 	shutdown_field()
 	field_segments = null
 	damaged_segments = null
 	mode_list = null
+	GLOB.moved_event.unregister(src, src)
 	. = ..()
```
comma, line 73: should probably also remove self from overmap object's list

MistakeNot4892: Clears it in the update proc.

MistakeNot4892:
Leeches work, not sure how to test the overmap shield thing but the lists on the object update properly on new, move and delete.

code/modules/shield_generators/shield_generator.dm
```diff
@@ -48,22 +51,36 @@
...
-
+	update_overmap_shield_list()
+
+/obj/machinery/power/shield_generator/proc/update_overmap_shield_list()
+	var/obj/effect/overmap/visitable/current_overmap_object = map_sectors["[z]"]
```
afterthought2: So this is actually not an accurate way to check which obj we should be throwing a shield over: if we are on a shuttle with a z with another ship, e.g., this will potentially return that ship rather than our shuttle.

There is a helper for this called `check_ownership`  on overmap objects, which returns true if you properly "belong" to it. For an example of how you are meant to use it see `attempt_hook_up_recursive` on ship computers; it may be time to generalize that.

MistakeNot4892:
Pushed a more thorough method of finding sector and tested, all working.

afterthought2 (approved):

PR #1363, opened 2021-02-18
Removes view reset default in life; fixes zooming.

### LINKED ISSUE #1359: Binoculars Issues
#### Description of issue
As reported: *"binoculars works like crap"*
Probably was encountered by other players (there was :thumbsup: emoji under report message)
Needs testing, as it may be a "subjective" kind of thing

N/A

#### Steps to reproduce
Acquire a binoculars
Use it
???

/obj/item/binoculars
code/game/objects/items/devices/binoculars.dm

Recently

https://github.com/infinitystation/Nebula/tree/50439ae5d5d062a21cc4b61ac64d4c32e8839dc4

#### Issue bingo

afterthought2:
Closes #1359.
Closes #1263.

Changes the default view reset behavior in Life() from reset to don't reset; machines and eyemobs still reset if conditions aren't met, so this shouldn't affect cameras or AI unless they were extremely broken to begin with. It will cause issues with sloppy/broken code elsewhere if there is any.

### DIFF
--- a/code/game/objects/item.dm
+++ b/code/game/objects/item.dm
@@ -756,9 +756,9 @@ modules/mob/living/carbon/human/life.dm if you die, you will be zoomed out.
 
 	user.visible_message("\The [user] peers through [zoomdevicename ? "the [zoomdevicename] of [src]" : "[src]"].")
 
-	GLOB.destroyed_event.register(src, src, /obj/item/proc/unzoom)
-	GLOB.moved_event.register(src, src, /obj/item/proc/unzoom)
-	GLOB.dir_set_event.register(src, src, /obj/item/proc/unzoom)
+	GLOB.destroyed_event.register(user, src, /obj/item/proc/unzoom)
+	GLOB.moved_event.register(user, src, /obj/item/proc/unzoom)
+	GLOB.dir_set_event.register(user, src, /obj/item/proc/unzoom)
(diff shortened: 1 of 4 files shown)


MistakeNot4892 (approved):
Obviates #1263.
EDIT: didn't realize you can close PRs from a PR, cool.

code/game/objects/item.dm
```diff
@@ -756,9 +756,9 @@ modules/mob/living/carbon/human/life.dm if you die, you will be zoomed out.
 
 	user.visible_message("\The [user] peers through [zoomdevicename ? "the [zoomdevicename] of [src]" : "[src]"].")
 
-	GLOB.destroyed_event.register(src, src, /obj/item/proc/unzoom)
-	GLOB.moved_event.register(src, src, /obj/item/proc/unzoom)
-	GLOB.dir_set_event.register(src, src, /obj/item/proc/unzoom)
+	GLOB.destroyed_event.register(user, src, /obj/item/proc/unzoom)
```
PsiOmegaDelta, line 759: Items still need to listen to their own destroyed-even in case they are deleted for other reasons than the host-mob itself being deleted, in order to undo the zoom-mode

PsiOmegaDelta: Is potentially handled by the unequipped event below, has this been tested?

PR #1365, opened 2021-02-18
Standardizing mousedrop functions.

MistakeNot4892:
Standardizes `MouseDrop` and `MouseDrop_T` overrides to use shared base procs for checking adjacency/incap/validity of drops.

### DIFF
--- a/code/_onclick/drag_drop.dm
+++ b/code/_onclick/drag_drop.dm
@@ -1,28 +1,39 @@
 /*
-	MouseDrop:
-
-	Called on the atom you're dragging.  In a lot of circumstances we want to use the
+	The below procs are called on the atom you're dragging.  In a lot of circumstances we want to use the
 	recieving object instead, so that's the default action.  This allows you to drag
 	almost anything into a trash can.
 */
 
-/atom/proc/CanMouseDrop(atom/over, var/mob/user = usr, var/incapacitation_flags)
-	if(!user || !over)
+/atom/MouseDrop(atom/over)
+	SHOULD_CALL_PARENT(TRUE)
+	if(!can_mouse_drop(over, usr) || !handle_mouse_drop(over, usr))
+		. = ..()
+
+/atom/proc/handle_mouse_drop(var/atom/over, var/mob/user)
+	. = receive_mouse_drop(src, user)
+
+// This proc checks if an atom can be mousedropped onto the target by the user.
+/atom/proc/can_mouse_drop(var/atom/over, var/mob/user = usr, var/incapacitation_flags = INCAPACITATION_DEFAULT)
+	SHOULD_CALL_PARENT(TRUE)
+	if(!istype(user) || !istype(over) ||QDELETED(user) || QDELETED(over) || QDELETED(src))
 		return FALSE
-	if(user.incapacitated(incapacitation_flags))
+	if(user.incapacitated(incapacitation_flags) || !CanPhysicallyInteract(user) || !user.check_dexterity(DEXTERITY_GRIP))
 		return FALSE
-	if(!src.Adjacent(user) || !over.Adjacent(user))
-		return FALSE // should stop you from dragging through windows
+	if(!check_mousedrop_adjacency(over, user))
+		return FALSE
 	return TRUE
 
-/atom/MouseDrop(atom/over)
-	if(!usr || !over) return
-	if(!Adjacent(usr) || !over.Adjacent(usr)) return // should stop you from dragging through windows
+/atom/proc/check_mousedrop_adjacency(var/atom/over, var/mob/user)
+	. = (Adjacent(user) && over.Adjacent(user))
 
-	spawn(0)
-		over.MouseDrop_T(src,usr)
-	return
+/mob/can_mouse_drop(var/atom/over, var/mob/user = usr, var/incapacitation_flags = INCAPACITATION_DEFAULT)
+	. = !anchored && ..()
 
-// Receive a mouse drop
-/atom/proc/MouseDrop_T(atom/dropping, mob/user)
-	return
+// Receive a mouse drop.
+// Returns false if the atom is valid for dropping further up the chain, true if the drop has been handled.
+/atom/proc/receive_mouse_drop(var/atom/dropping, var/mob/user)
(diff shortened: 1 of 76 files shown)


code/_onclick/drag_drop.dm
```diff
@@ -1,28 +1,42 @@
...
 
-/atom/proc/CanMouseDrop(atom/over, var/mob/user = usr, var/incapacitation_flags)
-	if(!user || !over)
+/atom/MouseDrop(atom/over)
+	SHOULD_CALL_PARENT(TRUE)
+	if(!can_mouse_drop(over, usr) || !handle_mouse_drop(over, usr))
+		. = ..()
+
+/atom/proc/handle_mouse_drop(var/atom/over, var/mob/user)
+	addtimer(CALLBACK(over, /atom/proc/receive_mouse_drop, src, user), 0)
```
afterthought2: I think axe the spawn; if it breaks horribly or fails unit tests we'll learn something and you can blame me.

code/_onclick/drag_drop.dm
```diff
@@ -1,28 +1,42 @@
...
 	recieving object instead, so that's the default action.  This allows you to drag
 	almost anything into a trash can.
 */
 
-/atom/proc/CanMouseDrop(atom/over, var/mob/user = usr, var/incapacitation_flags)
-	if(!user || !over)
+/atom/MouseDrop(atom/over)
+	SHOULD_CALL_PARENT(TRUE)
+	if(!can_mouse_drop(over, usr) || !handle_mouse_drop(over, usr))
+		. = ..()
```
afterthought2, line 10: What does this parent call actually do?

code/_onclick/drag_drop.dm
```diff
@@ -1,28 +1,42 @@
...
+/mob/can_mouse_drop(var/atom/over, var/mob/user = usr, var/incapacitation_flags = INCAPACITATION_DEFAULT)
+	. = !anchored && ..()
 
-// Receive a mouse drop
-/atom/proc/MouseDrop_T(atom/dropping, mob/user)
-	return
+// Receive a mouse drop.
+// Returns true if the atom is valid for dropping further up the chain, false if the drop has been handled.
+/atom/proc/receive_mouse_drop(var/atom/dropping, var/mob/user)
+	SHOULD_CALL_PARENT(TRUE)
```
afterthought2: I'm actually a bit hesitant about this. I realize you can unset it, but I'm not sure this should be the expected/default behavior. I'll trust your reading of the actual usage on it, though.

code/_onclick/drag_drop.dm
```diff
@@ -1,28 +1,42 @@
...
+	. = (dropping?.can_mouse_drop(src, user) && istype(dropping) && !QDELETED(dropping))
```
afterthought2: Is this here just because we are spawning? I don't feel like we should be doing extra checks in this proc that the wrapping proc hasn't already done.

code/game/objects/items/weapons/defib.dm
```diff
@@ -67,16 +67,15 @@
...
-		M.put_in_hands(src)
-
+// what is this proc doing?
+/obj/item/defibrillator/handle_mouse_drop(var/atom/over, var/mob/user)
+	if(ismob(loc))
+		var/mob/M = loc
+		if(M.unEquip(src))
+			add_fingerprint(usr)
+			M.put_in_hands(src)
+			return TRUE
```
afterthought2, line 77: Just noting that it's pretty confusing that truthy returns in this one mean "handled", but in the other one they mean "not handled."

code/modules/mechs/interface/screen_objects.dm
```diff
@@ -41,9 +41,10 @@
 	maptext_y = 3
 	maptext_width = 72
 
-/obj/screen/exosuit/hardpoint/MouseDrop()
-	..()
-	if(holding) holding.screen_loc = screen_loc
+/obj/screen/exosuit/hardpoint/handle_mouse_drop(atom/over, mob/user)
+	. = ..()
+	if(. && holding)
+		holding.screen_loc = screen_loc
```
afterthought2, line 46: Shouldn't we return false here?

afterthought2 (commented):
Lots of code here; glanced over it. I think it's fine in principle.

code/_onclick/drag_drop.dm
MistakeNot4892: Parent call adds the `receive_mouse_drop` timer.

code/_onclick/drag_drop.dm
MistakeNot4892: Yes, my thought was that this was necessary due to the async call.

code/game/objects/items/weapons/defib.dm
MistakeNot4892: I could swap the result and checks I guess. Seems a bit more confusing to me, but it would be more consistent.

code/_onclick/drag_drop.dm
MistakeNot4892: Alrighty.

code/_onclick/drag_drop.dm
MistakeNot4892: Nvm wrong block. Nothing as far as I know, but I figured better to call down in case of future override.

code/_onclick/drag_drop.dm
MistakeNot4892: It was set because the root receive_mouse_drop was doing a bunch of checking as commented below. Now that the checks and async are removed I can remove it.

code/_onclick/drag_drop.dm
MistakeNot4892: Removed now.

code/game/objects/items/weapons/defib.dm
MistakeNot4892: Swapped.

code/modules/mechs/interface/screen_objects.dm
MistakeNot4892: FIxed.

afterthought2 (approved):

PR #1369, opened 2021-02-19
Airlock components & damage tweaks

### LINKED ISSUE #1208: Unable to repair emagged airlock
#### Description of issue

See title.

Able to repair airlock after using emag on it

#### Steps to reproduce

1. Emag airlock
2. Try to repair it

airlock

Dunno, heard from player today

513, nebula fork

#### Issue bingo

comma:
Making doors use lock components instead of circuitboard for access.
Moves UI from circuit to lock.
Airlocks now use machinery damage system properly, passing damtypes where needed. It doesn't kick in until airlock's busted normally.
Can now transition construction states on busted airlocks, for disassembly and such.
Also fixes #1208
Emagging now sets machinery-wide emagged var rather than hijacking 'operating' var, which was expected to be bool in most places.
Depowered emagged airlocks can now be closed like normal, it just prevents non-forced opening/closing.

## Changelog
:cl:
bugfix: Emagged airlocks can now be crowbarred close, after they're depowered.
wip: implemented wip things
tweak: Airlocks now use access lock components to decide their access, instead of circuitboards. Locks now use same UI as airlock electronics did for access setup. Mapped airlocks should spawn these as needed to work same.
tweak: Doors can now have their internal components damaged too like the rest of machinery. It won't happen until they're bashed down first though.
tweak: Access locks with auto-set option will automatically set their access to the area's on installation into machine (like doors do)
imageadd: Access lock components (and network lock) now have their own icons instead of reusing scanning module one.
/:cl:


code/game/machinery/_machines_base/stock_parts/access_lock.dm
```diff
@@ -15,6 +18,11 @@
 				return SPAN_WARNING("\The [lock] flashes red! You lack the access to unlock this.")
 	return ..()
 
+/obj/item/stock_parts/access_lock/get_req_access()
+	if(locked)
+		return ..()
+	return list()
```
afterthought2: I think this isn't needed, but if you do include it you should change the proc just above.

code/game/machinery/_machines_base/stock_parts/access_lock.dm
```diff
@@ -25,6 +33,9 @@
 
 /obj/item/stock_parts/access_lock/on_install(obj/machinery/machine)
 	. = ..()
+	if(autoset && istype(machine, /obj/machinery/door))
+		var/obj/machinery/door/D = machine
+		req_access = D.get_auto_access()
```
afterthought2: Maybe worth generalizing this to non-doors, where the base implementation is `get_area(machine)?.req_access?.Copy()`.

code/game/machinery/_machines_base/stock_parts/access_lock.dm
```diff
@@ -46,35 +57,101 @@
...
-		if(selected_desc == "No Access")
-			req_access.Cut()
-			return
-		if(!(choices[selected_desc] in user.GetAccess()))
-			return
-		req_access = list(choices[selected_desc])
-		return
+/obj/item/stock_parts/access_lock/attack_self(mob/user)
+	if (!ishuman(user) && !istype(user,/mob/living/silicon/robot))
+		return ..(user)
```
afterthought2: Please no. Check fine tool user or whatever if you really think it's needed.

code/game/machinery/_machines_base/stock_parts/access_lock.dm
```diff
@@ -46,35 +57,101 @@
...
+		data["oneAccess"] = one_access
+	data["locked"] = locked
+	data["lockable"] = !emagged
+	data["autoset"] = autoset
+	return data
+
+/obj/item/stock_parts/access_lock/OnTopic(mob/user, list/href_list, state)
+	if(!emagged)
+		if(href_list["unlock"])
+			if(!req_access || istype(user, /mob/living/silicon))
```
afterthought2: Why?

comma: ancestors must've really hated synths (this and comment above are on code i just staright up copyported from airlock electronics to lock)

code/game/machinery/_machines_base/stock_parts/access_lock.dm
comma: Wasn't sure if access handling handles getting nulls instead of accesses

code/game/machinery/_machines_base/stock_parts/access_lock.dm
comma: Axed

code/game/machinery/_machines_base/stock_parts/access_lock.dm
comma: Changed the check to check if we're in the machine instead, and returns null here.

code/game/machinery/_machines_base/stock_parts/access_lock.dm
comma: Moved up to machinery level

afterthought2 (dismissed):

afterthought2 (approved):

PR #1373, opened 2021-02-20
Made REVIVAL_BRAIN_LIFE -1 (which is the default) infinite

Alceris:
So now that part of the comment in the config file isn't a lie.


MistakeNot4892 (approved):
Still think just setting the var to `INFINITY` and editing the example config is cleaner, but this does work.

PR #1391, opened 2021-02-24
Various remote engineering improvements

Mazianni:
## About this PR:

-Adds a shutoff valve monitoring and control program
-Tweaks the power monitoring program to be nicer looking and also allow for remote control of APCs.
~~-TODO: remote shield control and touchups on the shield monitoring program.~~


code/modules/modular_computers/file_system/programs/engineering/power_monitor.dm
```diff
@@ -84,32 +84,94 @@
...
 		refresh_sensors()
 		. = 1
+
+	if(href_list["toggle_breaker"])
+		var/obj/machinery/power/apc/A = locate(href_list["toggle_breaker"])
+
+		if(!CanInteract(user, state) || QDELETED(A))
+			return 0
+
+		A.operating = !A.operating
```
afterthought2: There's a proc for this. It's pretty bad just setting vars on APCs, easy to break stuff.

code/modules/modular_computers/file_system/programs/engineering/power_monitor.dm
```diff
@@ -84,32 +84,94 @@
...
+		if(!CanInteract(user, state) || QDELETED(A))
+			return 0
+
+		switch(input)
+			if("ON")
+				power_setting = 3
+			if("OFF")
+				power_setting = 0
+			if("AUTO")
+				power_setting = 4
```
afterthought2: There's a proc that lets you avoid some of the magic numbers. It may be better to make the whole thing into a proc.

code/modules/modular_computers/file_system/programs/engineering/shutoff_valve.dm
```diff
@@ -0,0 +1,61 @@
...
+			return 0
+
+		S.close_on_leaks = !S.close_on_leaks
+
+	if(href_list["toggle_open"])
+		var/obj/machinery/atmospherics/valve/shutoff/S = locate(href_list["toggle_open"])
+
+		// Invalid ref
+		if(!istype(S) || QDELETED(S))
+			return 0
```
afterthought2: It's actually best to return truthy stuff in this kind of situation: you've handled the interaction.

code/modules/modular_computers/file_system/programs/engineering/shutoff_valve.dm
```diff
@@ -0,0 +1,61 @@
...
+
+		if(S.open)
+			S.close()
+		else
+			S.open()
```
afterthought2: This doesn't actually do anything; they will reopen or close anyway next tick. These valves have two states: enabled (close if leaking, open otherwise) or disabled (always open).

code/modules/modular_computers/file_system/programs/engineering/shutoff_valve.dm
```diff
@@ -0,0 +1,61 @@
...
+	category = PROG_ENG
+	size = 10
+
+/datum/nano_module/program/shutoff_monitor
+	name = "Shutoff valve monitor"
+
+/datum/nano_module/program/shutoff_monitor/ui_interact(mob/user, ui_key = "main", var/datum/nanoui/ui = null, var/force_open = 1, var/datum/topic_state/state = GLOB.default_state)
+	var/list/data = host.initial_data()
+	data["valves"] = list()
+	for(var/obj/machinery/atmospherics/valve/shutoff/S in shutoff_valves)
```
afterthought2, line 24: Should check z level or something.

code/modules/modular_computers/file_system/programs/engineering/shutoff_valve.dm
```diff
@@ -0,0 +1,61 @@
...
+		data["valves"][++data["valves"].len] = list("name" = S.name, "enable" = S.close_on_leaks, "open" = S.open, "location" = get_area(S), "ref" = "\ref[S]")
```
afterthought2, line 30: There are sometimes cosmetic issues with using area rather than area.name in html; unsure if relevant here.

code/modules/modular_computers/file_system/programs/engineering/power_monitor.dm
Mazianni: Power channels are promptly cleaned up by the update proc, shouldn't be breaking anything.

Mazianni:
as per maintvote:
image
I guess I'll add a UI toggle.

code/modules/modular_computers/file_system/programs/engineering/power_monitor.dm
afterthought2: The update proc need not be called every tick. There's a proc for this; use it.

MistakeNot4892 (approved):

PR #1392, opened 2021-02-25
Collapsing ChangeTurf into one proc instead of half a dozen overrides.

MistakeNot4892:
Also causes turf changes to queue icon updates instead of hard calling them for every turf change.
Also also deletes unnecessary flood overlay updating being done in ChangeTurf when it's then done in update_icon afterwards anyway.
Some limited testing with tradeship set to 20 exoplanets took it from 90s init to 75s init on my machine, but I'm not sure how much of that is tied to the random element in exoplanet init.

### DIFF
--- a/code/game/turfs/turf_changing.dm
+++ b/code/game/turfs/turf_changing.dm
@@ -22,6 +22,7 @@
 
 //Creates a new turf
 /turf/proc/ChangeTurf(var/turf/N, var/tell_universe = TRUE, var/force_lighting_update = FALSE, var/keep_air = FALSE)
+
 	if (!N)
 		return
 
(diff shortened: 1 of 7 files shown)


code/game/turfs/turf_changing.dm
```diff
@@ -72,37 +70,49 @@
...
-			reconsider_lights()
-		if(dynamic_lighting != old_dynamic_lighting)
-			if(dynamic_lighting)
-				lighting_build_overlay()
-			else
-				lighting_clear_overlay()
-
-	for(var/turf/T in RANGE_TURFS(src, 1))
-		T.update_icon()
+	if(.)
```
Lohikar: This will always be true.

code/game/turfs/turf_changing.dm
```diff
@@ -72,37 +70,49 @@
...
+				W.reconsider_lights()
+			else if(W.opacity != old_opacity)
+				W.reconsider_lights()
+
+		for(var/turf/T in RANGE_TURFS(W, 1))
+			T.queue_icon_update()
+
+		GLOB.turf_changed_event.raise_event(W, old_density, W.density, old_opacity, W.opacity)
+		updateVisibility(W, FALSE)
+		if(isturf(W) && !W.flooded && W.flood_object)
```
Lohikar: `isturf(W)` is always true.

code/game/turfs/turf_changing.dm
```diff
@@ -39,30 +40,27 @@
...
 
 	overlays.Cut()
 	underlays.Cut()
 
-	// Run the Destroy() chain.
 	qdel(src)
 
 	var/old_opaque_counter = opaque_counter
-	var/turf/simulated/W = new N(src)
+	var/turf/W = new N(src)
```
Lohikar, line 59: If you're changing the type of this var, why keep it at all?
As far as I can tell, it was just used to get access to procs on `/simulated`.

MistakeNot4892: Imo it's clearer than src manipulation.
EDIT discussed on Discord but for clarity, I think the way src is overritten when the new turf is created is unclear to people who aren't already familiar with this particular quirk of DM, declaring W to reference isn't expensive and makes the following code unambiguous.

afterthought2 (approved):

PR #1396, opened 2021-02-26
Airlock's SetBounds changes

### LINKED ISSUE #1395: Multitile bounds problem

Multitile airlocks should properly call SetBounds or set_dir during object initialization

#### Steps to reproduce

1. spawn multi tile airlock
2. try go through it 

`/obj/machinery/door/airlock/multi_tile`

Since last multi tile airlock code touch 

#### Issue bingo

quardbreak:
Fixes #1395 
Open for review, currently have some bugs.

Todo:
- [x] ~~Check set_bounds correctly set bounds after set_dir~~
- [x] ~~Fix weird inverted bound (?) in double assembly code~~

* Repaths `multi_tile` to `double`
* Double airlock bounds set is now a door type function
* Double glass airlock is now have `opacity = FALSE`


code/game/machinery/doors/_door.dm
```diff
@@ -129,11 +122,24 @@
 		return 0
 	return 1
 
+/obj/machinery/door/proc/set_bounds()
+	if (dir == NORTH || dir == SOUTH)
```
MistakeNot4892, line 126: You should prune diagonal dirs put somewhere so this only sets on cardinal values.

code/game/machinery/doors/_door.dm
```diff
@@ -446,17 +452,8 @@
...
 	. = ..()
-	if(width > 1)
-		if(dir in list(EAST, WEST))
-			bound_width = width * world.icon_size
-			bound_height = world.icon_size
-		else
-			bound_width = world.icon_size
-			bound_height = width * world.icon_size
-
+	update_nearby_tiles()
```
MistakeNot4892, line 455: Won't this fail to update all tiles at the original position if you do it post ..()?

code/game/machinery/doors/double.dm
```diff
@@ -1,5 +1,5 @@
 //Terribly sorry for the code doubling, but things go derpy otherwise.
-/obj/machinery/door/airlock/multi_tile
+/obj/machinery/door/airlock/double
 	airlock_type = "double"
 	name = "\improper Airlock"
```
MistakeNot4892: Small note, but this doesn't need to be capitalized.

code/game/objects/structures/door_assembly.dm
```diff
@@ -19,12 +19,26 @@
...
 	var/symbol_color = "none"
+	var/width = 1 // For multi-tile doors
 
 /obj/structure/door_assembly/Initialize(mapload, d) // would be desirable to improve material handling here
 	. = ..(mapload)
 	set_dir(d)
 	update_icon()
 
+/obj/structure/door_assembly/set_dir()
+	. = ..()
```
MistakeNot4892: As with the airlock, should prune diagonals.

MistakeNot4892 (commented):
Mostly looks fine to me.

code/game/machinery/doors/_door.dm
afterthought2: I think it should come before `..()` on further inspection, yes. The fact that it actually moves it is weird/bad, but whatever.

MistakeNot4892:
Meant to type this earlier but was on my phone - you should be able to just do `if(newdir in GLOB.diagonals) newdir &= ~(NORTH|EAST)` or something to prune diagonals.

code/game/machinery/doors/_door.dm
```diff
@@ -129,11 +122,33 @@
...
+		bound_height = width * world.icon_size
+
 /obj/machinery/door/set_density(new_density)
 	. = ..()
 	if(.)
 		explosion_resistance = density ? initial(explosion_resistance) : 0
 
+/obj/machinery/door/set_dir(new_dir)
+	if(new_dir in GLOB.cornerdirs) 
+		new_dir &= ~(NORTH|EAST)
```
afterthought2: I think these two lines can be omitted, as the logic below handles this case already.

code/game/objects/structures/door_assembly.dm
```diff
@@ -19,12 +19,35 @@
...
-/obj/structure/door_assembly/Initialize(mapload, d) // would be desirable to improve material handling here
+/obj/structure/door_assembly/Initialize(mapload, _mat, _reinf_mat, direction)
 	. = ..(mapload)
-	set_dir(d)
+	set_dir(direction)
 	update_icon()
 
+/obj/structure/door_assembly/set_dir(new_dir)
+	if(new_dir in GLOB.cornerdirs) 
+		new_dir &= ~(NORTH|EAST)
```
afterthought2: Same here.

afterthought2 (approved):

PR #1397, opened 2021-02-26
Fix turret and cremation staff logs

SierraKomodo:
Fixes the admin logs for turret controllers and attack logs for cremation.

```
  - admin: "Admin logs for turret setting changes now display the correct user instead of *INVALID* and a loc link."
  - admin: "Attack logs for cremation no longer has the attacker and victim reversed."
```


MistakeNot4892 (approved):
Generally changelogs aren't needed for bugfixes.

PR #477, opened 2020-05-07
Turret Rewrite

SierraKomodo:
Refactoring turrets by rewriting them entirely.

Planned updates include:
 - New effects for emagged, EMP'd, and malf upgraded turrets (Including a 'haywire' mode where the turrets go crazy and start firing off randomly)
 - Turrets use the same projectile as the gun used to create them (Ion turrets anyone?)
 - Turrets have the same firemode options as the gun used to create them
 - Turrets have a new control panel UI
 - Turrets have a wiring panel that can be manually hacked
 - Turrets can be constructed, deconstructed, and repaired
 - Turret frames can be built from differing materials that affect turret health

PR #1424, opened 2021-03-03
Turret Rewrite

Neerti:
Rewrites turrets so that they will hopefully both be not pushovers and also more interesting to fight.

Putting it up as a draft so my work can be looked at in-case the foundation is superbad or something. DNM etc etc.

Stuff needing to be done:
- [ ] Look at #477 and #999 for ideas.
- [x] Make new turret `Process()` not be dumb and be ready for production verses just testing.
- [x] Try to find the sweetspot between efficient ticker usage verses code complexity/brittleness. Ideally the turret would only be on SSfastprocess when it's turning/shooting, would use a slower ticker (SSmachinery?) if idle and battery needs charging, and wouldn't tick at all if idle and with a full cell, and could be 'woken up' by its prox trigger event, but that might be too clever of a solution.
- [x] Finish gun tweak so turrets can use guns directly.
- [x] Make turret actually shoot.
- [ ] Make it more obvious how wide a turret's traverse is, visually ideally. One idea might be to change the 'base' of the turret sprite to represent it, e.g. semicircle for 180 degree traverse vs full circle for 360 degrees.
- [x] Find solution to problem where if traverse is greater than 180 degrees, the turret might turn in the wrong direction due to sloppy code marked with `// This probably won't work if traverse is greater than 180.`
- [ ] Remove TODO comments in my code.
- [ ] Consider making the turrets use machine components for power/armor/etc instead of directly handling power.
- [x] NanoUI interface.
- [ ] See if networking can be integrated it as per #999.
- [ ] Make turrets buildable in-round.
- [ ] Implement hostility datum (or decl?) so we can avoid giant blocks of if statements and possibly allow for future code reuse.
- [ ] Delete old turrets from code.
- [ ] Do the Great Rebase at the end and probably have turrets in one commit, FSM in another, and gun changes in third.

## Changelog
:cl:
bugfix: fixed a few things
wip: implemented wip things
tweak: tweaked a few things
soundadd: added a new sound thingy
sounddel: removed an old sound thingy
rscdel: removed old things
rscadd: added new things
imageadd: added some icons and images
imagedel: deleted some icons and images
maptweak: edited some maps
spellcheck: fixed a few typos
experiment: implemented experiment things
balance: rebalanced something
admin: messed with admin stuff
/:cl:


code/game/machinery/turrets/turret_fsm.dm
```diff
@@ -0,0 +1,79 @@
...
+
+/decl/state_transition/turret/yes_power
+	target = /decl/state/turret/idle 
+
+
+/decl/state_transition/turret/lost_power
+	target = /decl/state/turret/no_power
+
+/decl/state_transition/turret/lost_power/is_open(obj/machinery/turret/turret)
+	return (turret.stat & NOPOWER) && !turret.internal_cell.check_charge(turret.idle_power_usage * CELLRATE) // Should this be made a proc on turrets?
```
MistakeNot4892: Probably should make this some variant of `is_powered()` on the turrets so you could potentially override them down the track for wired or immortal (PoI etc) turrets.

code/game/machinery/turrets/turrets.dm
```diff
@@ -0,0 +1,225 @@
+// Base type for rewritten turrets, designed to hopefully be cleaner and more fun to use and fight against.
+// Use the subtypes for 'real' turrets.
+
+/obj/machinery/turret
+	name = "don't use me"
```
MistakeNot4892: Something like `abstract turret` would probably be slightly clearer. You can also do something like `base_type = /obj/machinery/turret` (which is required for constructable machines anyway) and then in `Initialize()` do `if(base_type == type) PRINT_STACK_TRACE("Base abstract turret type instantiated, use a subtype instead!")`.

code/game/machinery/turrets/turrets.dm
```diff
@@ -0,0 +1,225 @@
...
+	icon = 'icons/obj/turrets.dmi'
+	icon_state = "turret_barrel"
+	anchored = TRUE
+	idle_power_usage = 500
+	active_power_usage = 60000 // This determines how fast the internal cell gets charged.
```
MistakeNot4892: `60 KILOWATTS` is the slightly cleaner standard.

code/game/machinery/turrets/turrets.dm
```diff
@@ -0,0 +1,225 @@
...
+
+	transform_animate_time = 0.2 SECONDS
+
+	// Visuals.
+	var/image/turret_stand = null
+	var/image/turret_ray = null
+	var/ray_color = "#FFFFFFFF" // Color of the ray, changed by the FSM when switching states.
```
MistakeNot4892: This will fail unit tests I believe, internally BYOND uses all lowercase for colours so the code quality checks fail on uppercase letters in colour strings.

code/game/machinery/turrets/turrets.dm
```diff
@@ -0,0 +1,225 @@
...
+
+	// Sounds
+	var/turn_on_sound = null // Played when turret goes from off to on.
+	var/turn_off_sound = null // The above, in reverse.
+
+	// Shooting
+	var/obj/item/gun/installed_gun = null // Instance of the gun inside the turret.
```
MistakeNot4892: A pattern I've found useful has been something like `installed_gun = /obj/item/gun` and then in init `if(ispath(installed_gun)) installed_gun = new installed_gun(src)`. Saves an extra largely unused var declaration.

code/game/machinery/turrets/turrets.dm
```diff
@@ -0,0 +1,225 @@
...
+	var/gun_type = /obj/item/gun/energy/laser/practice // The type of gun to put in the above var on init.
+	var/gun_looting_prob = 25 // If the turret dies and then is disassembled, this is the odds of getting the gun.
+
+	// Power
+	var/obj/item/cell/internal_cell = null // The battery that is used to fire the gun and for the turret to work semi-independantly of the powernet.
+	var/cell_type = /obj/item/cell/high // The type this turret will have on initialization.
```
MistakeNot4892: As above with the path comment.

code/game/machinery/turrets/turrets.dm
```diff
@@ -0,0 +1,225 @@
...
+	var/turning_rate = 90 // Degrees per second.
+
+	// Detection.
+	var/datum/proximity_trigger/square/proximity = null
+	var/vision_range = 7	// How many tiles away the turret can see. Values higher than 7 will let the turret shoot offscreen, which might be unsporting.
+							// Higher values may also have a performance cost.
+	
+	// Logic
+	var/datum/state_machine/turret/state_machine = null
+	var/atom/target = null
```
MistakeNot4892: Would probably be worth setting this as a weakref instead of a hard reference - one of the big contributors to qdel failures is this kind of var.

MistakeNot4892:
````
var/weakref/target = weakref(some_atom)
var/atom/some_atom = target?.resolve()
````

File: code/game/machinery/turrets/turrets.dm [hunk 8]
```diff
@@ -0,0 +1,225 @@
...
+	calculate_traverse()
+	update_icon()
+
+	START_PROCESSING(SSfastprocess, src) // TODO
+	return ..()
+
+/obj/machinery/turret/Destroy()
+	QDEL_NULL(state_machine)
+	QDEL_NULL(proximity)
+	if(istype(internal_cell))
```
MistakeNot4892: QDEL_NULL is fine with null vars.

code/game/machinery/turrets/turrets.dm
```diff
@@ -0,0 +1,225 @@
...
+		QDEL_NULL(internal_cell)
+	if(istype(installed_gun))
+		QDEL_NULL(installed_gun)
+	STOP_PROCESSING(SSfastprocess, src) // TODO
+	return ..()
+
+/obj/machinery/turret/Process()
+	process_power()
```
MistakeNot4892, line 105: Any particular reason not to call `..()`? I think there's some power usage and state stuff down there that may be important.

code/game/machinery/turrets/turrets.dm
```diff
@@ -0,0 +1,225 @@
...
+		process_turning()
+	else if(turret_state.type == /decl/state/turret/engaging)
+		process_shooting()
+
+/obj/machinery/turret/proc/process_power()
+	// If the turret is dead or turned off, don't draw any power.
+	if((stat & BROKEN) || !internal_cell)
+		update_use_power(POWER_USE_OFF)
+		// TODO Consider making turret stop processing later.
+		return
```
MistakeNot4892, line 119: Couple of `elses` here and below that are right after `returns` and aren't strictly needed.

code/game/machinery/turrets/turrets.dm
```diff
@@ -0,0 +1,225 @@
...
+		QDEL_NULL(installed_gun)
+	STOP_PROCESSING(SSfastprocess, src) // TODO
+	return ..()
+
+/obj/machinery/turret/Process()
+	process_power()
+
+	state_machine.evaluate()
+	var/decl/state/turret/turret_state = state_machine.current_state
+	if(turret_state.type == /decl/state/turret/turning) // TODO remove after test.
```
MistakeNot4892: Not sure how I feel about the direct type comparisons. The modular pattern would be calling a generic proc (`state_machine.current_state.process_state(src)` or something maybe?) rather than the specific instance procs.

code/game/machinery/turrets/turrets.dm
```diff
@@ -0,0 +1,225 @@
...
+		adjustment = adjustment * -1
+
+	set_bearing(current_bearing + adjustment)
+
+/obj/machinery/turret/proc/calculate_turn_rate_per_process()
+	return turning_rate / (1 SECOND / SSfastprocess.wait)
+
+
+/obj/machinery/turret/proc/process_shooting()
+	if(installed_gun)
```
MistakeNot4892: Worth checking `target` here as well?

code/game/machinery/turrets/turrets.dm
```diff
@@ -0,0 +1,225 @@
...
+		installed_gun.Fire(target, src)
+
+
+/obj/machinery/turret/update_icon()
+	if(!turret_stand)
+		turret_stand = image(icon, "turretCover")
+		turret_stand.layer = src.layer - 0.01
+		turret_stand.appearance_flags = KEEP_APART|RESET_TRANSFORM|TILE_BOUND|PIXEL_SCALE 
+		overlays += turret_stand
```
MistakeNot4892: Please use `add_overlay()` rather than directly manipulating overlays.

code/game/machinery/turrets/turrets.dm
```diff
@@ -0,0 +1,225 @@
...
+	return turning_rate / (1 SECOND / SSfastprocess.wait)
+
+
+/obj/machinery/turret/proc/process_shooting()
+	if(installed_gun)
+		installed_gun.Fire(target, src)
+
+
+/obj/machinery/turret/update_icon()
+	if(!turret_stand)
```
MistakeNot4892, line 190: Probably should `cut_overlays()` before you add any new ones.

MistakeNot4892: Ah nvm I see what you're doing now.

code/game/machinery/turrets/turrets.dm
```diff
@@ -0,0 +1,225 @@
...
+
+	// Changes the ray color based on state.
+	overlays -= turret_ray
+	turret_ray.color = ray_color
+	overlays += turret_ray
+
+	..()
+
+/obj/machinery/turret/proc/calculate_traverse()
+	if(traverse >= 360)
```
MistakeNot4892, line 217: If you just want to clamp it to a sane angle, you can do `traverse = traverse % 360` I believe.

MistakeNot4892: Nvm this is literally what SIMPLIFY_DEGREES does.

code/modules/projectiles/gun.dm
```diff
@@ -249,52 +249,65 @@
...
+				handle_click_empty(user)
+				return
+
+		last_safety_check = world.time
+	
 	if(set_click_cooldown)
 		var/shoot_time = (burst - 1) * burst_delay
-		user.setClickCooldown(shoot_time) //no clicking on things while shooting
+		if(isliving(firer))
+			var/mob/living/user = firer
```
MistakeNot4892: Might be worth declaring `user` up the top somewhere, then just doing `istype(user)` in the spots where you're doing `isliving(firer)`.

code/modules/projectiles/gun.dm
```diff
@@ -331,83 +345,81 @@
 #undef FIREARM_MESSAGE_SPAM_TIME
 
 //called if there was no projectile to shoot
-/obj/item/gun/proc/handle_click_empty(mob/user)
+/obj/item/gun/proc/handle_click_empty(atom/movable/firer)
 	if(check_fire_message_spam("click"))
-		if(user)
-			user.visible_message("*click click*", "<span class='danger'>*click*</span>")
+		if(isliving(firer))
```
MistakeNot4892: Does this need to be living specifically? I noticed during the video you posted turrets click empty and make the noise, so the message seems appropriate.

code/modules/projectiles/gun.dm
```diff
@@ -331,83 +345,81 @@
...
-			for(var/obj/item/rig_module/stealth_field/S in R.installed_modules)
-				S.deactivate()
-
-	if(space_recoil)
-		if(!user.check_space_footing())
-			var/old_dir = user.dir
-			user.inertia_ignore = projectile
-			step(user,get_dir(target,user))
-			user.set_dir(old_dir)
+	if(isliving(firer))
```
MistakeNot4892, line 377: It might be worth passing this entire chunk of user-specific code into a new proc for the sake of readability.

code/modules/projectiles/guns/launcher/pneumatic.dm
```diff
@@ -76,11 +76,12 @@
 /obj/item/gun/launcher/pneumatic/attack_self(mob/user)
 	eject_tank(user)
 
-/obj/item/gun/launcher/pneumatic/consume_next_projectile(mob/user=null)
+/obj/item/gun/launcher/pneumatic/consume_next_projectile(atom/movable/firer)
 	if(!item_storage.contents.len)
 		return null
 	if (!tank)
-		to_chat(user, "There is no gas tank in [src]!")
+		if(isliving(firer))
```
MistakeNot4892: As far as I know it's harmless to pass strings to non-mobs, it just doesn't do anything.

code/modules/projectiles/guns/projectile.dm
```diff
@@ -82,13 +82,15 @@
...
+/obj/item/gun/projectile/process_point_blank(obj/projectile, atom/movable/firer, atom/target)
 	..()
 	if(chambered && ishuman(target))
 		var/mob/living/carbon/human/H = target
 		var/zone = BP_CHEST
-		if(user && user.zone_sel)
-			zone = user.zone_sel.selecting
+		if(isliving(firer))
+			var/mob/living/user = firer
+			if(user && user.zone_sel)
```
MistakeNot4892: You don't need to nullcheck user, isliving() would fail if firer was null.

code/modules/projectiles/guns/projectile/shotgun.dm
```diff
@@ -78,7 +78,7 @@
 	..(user, allow_dump=1)
 
 //this is largely hacky and bad :(	-Pete
-/obj/item/gun/projectile/shotgun/doublebarrel/attackby(var/obj/item/A, mob/user)
+/obj/item/gun/projectile/shotgun/doublebarrel/attackby(var/obj/item/A, mob/living/user)
```
MistakeNot4892: Please avoid changing the types of args in proc headers, particularly in these highly overridden ones, it can make changes to mob vars unpredictable (status PR for example was a bitch due to people overriding the arg type). It would be preferrable to typecheck user and cast it to living if needed.

code/modules/projectiles/projectile.dm
```diff
@@ -13,7 +13,7 @@
 	randpixel = 0	
 	var/bumped = 0		//Prevents it from hitting more than one guy at once
 	var/def_zone = ""	//Aiming at
-	var/mob/firer = null//Who shot it
+	var/atom/movable/firer = null//Who shot it
```
MistakeNot4892, line 16: If you feel like it, and if you don't find weakrefs horrible, changing this to a weakref would be good for the same reasons as `target` earlier.

code/modules/projectiles/targeting/targeting_triggers.dm
```diff
@@ -27,5 +27,5 @@
 	owner.visible_message("<span class='danger'>\The [owner] pulls the trigger reflexively!</span>")
 	var/obj/item/gun/G = aiming_with
 	if(istype(G))
-		G.Fire(aiming_at, owner)
+		G.Fire(aiming_at, owner, target_zone = owner.zone_sel?.selecting)
```
MistakeNot4892: Offhand I don't think `zone_sel` is ever going to be null outside of exceptional circumstances.

MistakeNot4892:
Overall this is looking pretty awesome so far. The state machine in particular looks pretty extensible.

code/game/machinery/turrets/turrets.dm
Neerti: `/obj/machinery/Process()` returns `PROCESS_KILL`

code/game/machinery/turrets/turrets.dm
Neerti: My original code had `add_overlay()` but for some reason I thought Neb didn't have that, yet now it shows up in VSC, so I'll put them back.

code/game/machinery/turrets/turrets.dm
MistakeNot4892: Was going to say 'you don't need to return the value' but it also does literally nothing else so point taken.

MistakeNot4892:
Closing this for the time being, pls do not delete branch, would like to recover it someday.

PR #1427, opened 2021-03-04
Tag modification on mapload; (proper) support for loading duplicate templates.

afterthought2:
Needs lots of testing; opening for discussion of how to implement this (if at all) and to force myself to do this.

On loading a map, dmm suite now sets a unique map id and modifies all vars that atoms opt in to having modified by appending that id to the var. This occurs before Initialize is run.

:cl:
tweak: Additional support has been instated for templates which are to be loaded multiple times. The main user-facing change is that when adding shuttle-restricted waypoints to an overmap object's initial_restricted_waypoints list, use the shuttle's type path, not its name. Templates that load ferry-type shuttles with waypoints not on the template will break with this system, but you can opt out by setting modify_tag_vars on the template datum to FALSE.
/:cl:


afterthought2:
I made this only run on templates with the `ALLOW_DUPLICATES` flag for now. That lets us retain feature parity with legacy maps. It's possible that there should be separate logic for what is going on here and actually _spawning_ duplicate maps during random away site generation; unsure about that.

afterthought2:
Located 3 mapped `fabricator` subtypes that were setting either `id_tag` or `initial_id_tag`. The latter is unused entirely; the former is used by e.g. radio components but they don't seem to have any. Corrected these to `initial_network_id` and removed `initial_id_tag`, but may be missing something here.

afterthought2:
Ready for review.

MistakeNot4892:
Fab changes seems fine, they are probably left over from previous systems.

MistakeNot4892 (dismissed):
Have to admit I barely understand the systems this PR is reworking, but the logic seems sound and none of the changes spring out to me as wrong/broken.

afterthought2:
I probably need to look at the template flag/duplicate opt-out setup before this is merged.

MistakeNot4892 (approved):
Let's give it a go.

PR #1431, opened 2021-03-04
Added a config option for visible death messages in humans.

MistakeNot4892:
Implements #980.

### DIFF
--- a/config/example/game_options.txt
+++ b/config/example/game_options.txt
@@ -9,6 +9,9 @@ HEALTH_THRESHOLD_CRIT -50
 ## level of health at which a mob becomes dead
 HEALTH_THRESHOLD_DEAD -100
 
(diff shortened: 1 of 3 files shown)


config/example/game_options.txt
```diff
@@ -9,6 +9,9 @@ HEALTH_THRESHOLD_CRIT -50
 ## level of health at which a mob becomes dead
 HEALTH_THRESHOLD_DEAD -100
 
+## Comment this line to prevent humans showing a visible message upon death ('X seizes up then falls limp, eyes dead and lifeless').
+SHOW_HUMAN_DEATH_MESSAGE
```
comma: IMO should start commented out, to keep current behavior

MistakeNot4892: Fair call.

comma (approved):

PR #1432, opened 2021-03-04
Added config option to force webhooks to send as plaintext.

MistakeNot4892:
Implements #985.


MistakeNot4892:
Pls review.

afterthought2 (approved):

afterthought2:
Only thing to really consider is whether disabling embeds is better in the json file per webhook than the master config.

PR #1441, opened 2021-03-06
Refactoring grab movement handling.

### LINKED ISSUE #1435: Dragging + doors don't mix
#### Description of issue
Approaching a door while dragging an object too large to swap places with prevents the door from automatically opening, players need to manually use an empty hand to get the door to open.  POSSIBLY related, players in wheelchairs seem to need to do the same.

Automatic doors should be, uh.  automatic.

#### Steps to reproduce
drag something.  typically seen with gas canisters or patient-laden roller beds - this does not happen when dragging a mob or something small you can walk through.   Approach a door and walk into it.  Feel outrage as the door waits for you to empty a hand so you can access it.

any automatic door while dragging things.

A while now, I thought it was a feature.

Client Version: 513
Server Revision: 57b4162b6d60ac230dc76caf62315f602f8cb342 - ministationmockup - 2021-03-02
Game ID: caU-a2kj
Current map: Ministation Zebra

#### Issue bingo

MistakeNot4892:
- Prevents grabbed atoms from moving if the holder does not move, which is what was resulting in doors failing to open.
- Replaced `step(AM, get_dir(src, old_turf))` with `AM.DoMove(foo)` and an adjacency check.
- Added some basic move handling for atoms with no movement handlers defined. Includes a workaround for an annoying issue with dragging non-mobs.
- Grab slowdown has been significantly decreased across the board and scales to the size of the atom being dragged.
- Lactate now takes effect slower.
- Fixes #1435.

The workaround is not an ideal solution, but step() breaks grabs on diagonal moves.

### DIFF
--- a/code/datums/movement/mob.dm
+++ b/code/datums/movement/mob.dm
@@ -188,8 +188,8 @@
 			to_chat(mob, SPAN_WARNING("You're pinned down by \a [mob.pinned[1]]!"))
 		return MOVEMENT_STOP
 
-	for(var/obj/item/grab/G in mob.grabbed_by)
-		if(G.assailant != mob && (mob.restrained() || G.stop_move()))
+	for(var/obj/item/grab/G as anything in mob.grabbed_by)
+		if(G.assailant != mob && G.assailant != mover && (mob.restrained() || G.stop_move()))
 			if(mover == mob)
 				to_chat(mob, SPAN_WARNING("You're restrained and cannot move!"))
 			mob.ProcessGrabs()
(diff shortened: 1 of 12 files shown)


code/datums/movement/mob.dm
```diff
@@ -188,7 +188,7 @@
 			to_chat(mob, SPAN_WARNING("You're pinned down by \a [mob.pinned[1]]!"))
 		return MOVEMENT_STOP
 
-	for(var/obj/item/grab/G in mob.grabbed_by)
+	for(var/obj/item/grab/G as anything in mob.grabbed_by)
 		if(G.assailant != mob && (mob.restrained() || G.stop_move()))
```
PsiOmegaDelta: In testing things seem to work much better when this line is changed to `if(G.assailant != mover && (mob.restrained() || G.stop_move()))` (mob -> mover)

PsiOmegaDelta: If it is the case that a mob can grab itself (to hold wounds perhaps?) then it should likely instead be `if(G.assailant != mob && G.assailant != mover && (mob.restrained() || G.stop_move()))`

comma (dismissed):

comma (approved):

afterthought2 (approved):

PR #1444, opened 2021-03-07
Fixes lighting_overlays breaking SSlighting.overlay_queue

Andrew-Fall:
Deleting a lighting_overlay would remove it from the SSlighting.overlay_queue. This would cause the current index to potentially skip over an item, and rarely cause a runtime.

### DIFF
--- a/code/modules/lighting/lighting_overlay.dm
+++ b/code/modules/lighting/lighting_overlay.dm
@@ -127,8 +127,10 @@
 	return
 
 /atom/movable/lighting_overlay/Destroy()
+	// Do not remove this from SSlighting.overlay_queue
+	// or it will mess with the queue. It will fall out 
+	// safely during processing
 	total_lighting_overlays--
-	SSlighting.overlay_queue -= src
 
 	var/turf/T = loc
 	if(istype(T))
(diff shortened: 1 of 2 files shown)


code/modules/lighting/lighting_overlay.dm
```diff
@@ -127,8 +127,10 @@
 	return
 
 /atom/movable/lighting_overlay/Destroy()
+	// Do not remove this from SSlighting.overlay_queue
```
MistakeNot4892, line 130: What if lighting is suspended and it hard deletes after failing to GC?

Andrew-Fall: As per discord convo: SSlighting is never suspended in code, such an event would be rare/out-of-scope. The chance of the lighting overlay being hard deleted before SSlighting could process it is rare. Finally, the chance of a lighting_overlay being deleted which only occurs during a dynamic light change is also rare. As such it's rare^3 * out-of-scope and is fine.

MistakeNot4892 (approved):

code/modules/lighting/lighting_overlay.dm
Andrew-Fall: Actually even if it hard del the list reference would just be set to null, and SSlighting would be fine with that. QDELETED is a null check as well

PR #1446, opened 2021-03-08
Expands modular computer terminals

NataKilar:
Greatly expands modular computer terminals in preparation for hacking, and for fun shenanigans in the interim

* Adds terminal commands for file manipulation: Listing, moving, copying, renaming, and deletion.
* Adds network device commands which expose public variables and methods on machinery to terminal commands. These can be added through the network device UI, or through terminal commands themselves. The player facing terminology is "alias" for the name of the command, and "reference" for the variable or method that is modified/called by the network command.
* Adds network receivers, to expose non-network machinery to the network. Network locks have been made a subtype of network receivers.
* Tweaked a few old terminal commands to make them more expandable.

Network device commands are analogous to radio signals, and the code in general is similar. The sanitization of arguments for commands in particular needs another set of eyes on it.

## Changelog
:cl:
feature: Added terminal commands for file manipulation.
feature: Added network commands, which allow for manipulation of certain public vars and methods on network devices via terminal commands
feature: Added network receiver stock parts for exposing non-network machinery to network commands.
/:cl:


code/game/machinery/_machines_base/stock_parts/network_receiver.dm
```diff
@@ -0,0 +1,41 @@
...
+	var/datum/extension/network_device/D = get_extension(src, /datum/extension/network_device)
+	D.sanitize_commands()
+
+/obj/item/stock_parts/network_receiver/attack_self(mob/user)
+	var/datum/extension/network_device/D = get_extension(src, /datum/extension/network_device)
+	D.ui_interact(user)
+
+/obj/item/stock_parts/network_receiver/attackby(obj/item/I, mob/user)
+	. = ..()
+	if(isMultitool(I))
```
afterthought2: Should really be a multitool extension. You may be able to generalize the radio component one, as it handles target acquisition to some extent.

code/game/machinery/_machines_base/stock_parts/network_receiver.dm
```diff
@@ -0,0 +1,41 @@
...
+		for(var/obj/machinery/new_machine in view(2, user))
+			candidates += new_machine
+		if(!length(candidates))
+			to_chat(user, SPAN_WARNING("You fail to import configuration settings from anything. Try standing near a machine."))
+			return
+		var/obj/machinery/input = input(user, "Which machine would you like to configure \the [src] for?", "Machine Selection") as null|anything in candidates
+		if(!istype(input) || !CanPhysicallyInteractWith(usr, src))
+			return
+		var/datum/extension/network_device/stock_part/D = get_extension(src, /datum/extension/network_device/)
+		D.machine = weakref(input)
```
afterthought2: Why set it here and not just automatically on install? It's not like there's further configuration that can be done here as far as I can tell.

code/modules/atmospherics/components/binary_devices/pump.dm
```diff
@@ -154,6 +154,10 @@ Thus, the two variables affect pump operation are set in New():
 	ui_interact(user)
 	return TRUE
 
+/obj/machinery/atmospherics/binary/pump/update_use_power(new_use_power)
+	. = ..()
+	queue_icon_update()
```
afterthought2: This should be called automatically if it matters.

code/modules/modular_computers/networking/device_types/_network_device.dm
```diff
@@ -20,6 +24,10 @@
 	if(autojoin)
 		SSnetworking.queue_connection(src)
 	
+	if(has_commands)
+		LAZYINITLIST(command_and_call)
+		LAZYINITLIST(command_and_write)
```
afterthought2: Why?

code/modules/modular_computers/networking/device_types/_network_device.dm
```diff
@@ -196,6 +226,181 @@
...
+		if(IC_FORMAT_DIR)
+			if(istext(command_args))
+				return text2dir(command_args)
+		if(IC_FORMAT_BOOLEAN)
+			if(istext(command_args))
+				switch(uppertext(command_args))
+					if("TRUE")
+						return TRUE
+					if("FALSE")
+						return FALSE
```
afterthought2, line 336: So the write setters should really be doing this already, at least if you use the safe call.

afterthought2 (commented):
I think you should auto-add all available vars and methods with e.g. "name" as alias on install.

This is pretty different from radio, in that for radio the tags -> vars/methods mapping are triggers for radio signals, whereas here the alias is just a convenience for the user. If I were to implement it I would have just had fixed aliases as names, and you can listcom with args (or similar) to obtain a man entry with e.g the name, the desc, the format flag, and the read-only status. Setting alias seems more like a power user feature.

Should also allow you to read vars from the command line; haven't read far enough to see if that's the case but doesn't seem to be.

code/game/machinery/_machines_base/stock_parts/network_receiver.dm
NataKilar: Cut out the multicool configuration entirely per our discussion on discord.

code/modules/modular_computers/networking/device_types/_network_device.dm
NataKilar: Left in for now, I'll go through with a fine-tooth comb and make sure all the public vars are sanitizing as necessary and delete this.

NataKilar:
Adjusted some things re: discussion on discord. Command configuration is solely through the command line now, and listcom can now print out a description of commands fed into it a la man. Unskilled users who attempt to call commands will now call a random method command on a cooldown. This isn't exactly engaging, but it works.

Also added a network scanner for getting network information from nearby network devices.

code/game/machinery/doors/airlock.dm
```diff
@@ -988,11 +988,6 @@ About the new airlock wires panel:
 /obj/machinery/door/airlock/proc/toggle_lock(var/forced = 0)
 	return locked ? unlock() : lock()
 
-/obj/machinery/door/airlock/allowed(mob/M)
```
MistakeNot4892: What does this removal relate to?

code/modules/atmospherics/components/binary_devices/pump.dm
MistakeNot4892: Unsure from author reply if this has been discussed/addressed.

code/modules/modular_computers/terminal/terminal.dm
```diff
@@ -33,6 +54,24 @@
 	return TRUE
 
 /datum/terminal/Process()
+	if(current_move)
+		var/result = current_move.update_progress()
+		var/error
```
MistakeNot4892: You don't need to declare/set this, just call `append_to_history()` in the if/else.

code/modules/modular_computers/terminal/terminal.dm
```diff
@@ -4,20 +4,41 @@
...
 	START_PROCESSING(SSprocessing, src)
 
 /datum/terminal/Destroy()
 	STOP_PROCESSING(SSprocessing, src)
 	if(computer && computer.terminals)
 		computer.terminals -= src
 	computer = null
+	current_disk = null
+	for(var/D in disks)
+		var/datum/file_storage/FS = disks[D]
```
MistakeNot4892, line 37: Couldn't you use QDEL_NULL_LIST() for this?

code/modules/modular_computers/terminal/terminal.dm
```diff
@@ -4,20 +4,41 @@
...
+		disks[D] = new D(computer)
 	START_PROCESSING(SSprocessing, src)
 
 /datum/terminal/Destroy()
 	STOP_PROCESSING(SSprocessing, src)
 	if(computer && computer.terminals)
 		computer.terminals -= src
 	computer = null
+	current_disk = null
+	for(var/D in disks)
```
MistakeNot4892, line 36: Couldn't this just be:
````
	for(var/D in disks)
		qdel(disks[D])
	disks = null
	````

**Comment by MistakeNot4892** [comment]
Could the available commands on terminals be unit tested to make sure they're valid/returning correctly/not hitting a protected variable somehow? Makes me a tiny bit uneasy.

File: code/game/machinery/doors/airlock.dm (hunk 6 above)
Comment (NataKilar) [reply code/game/machinery/doors/airlock.dm]: Because commands check access, this bit of code meant if you bolted a door via a command you could not then unbolt it, since the access check would always return false. As far as I can tell there's no other reason it should return false when bolted regardless, so I removed it.

File: code/modules/atmospherics/components/binary_devices/pump.dm (hunk 3 above)
Comment (NataKilar) [reply code/modules/atmospherics/components/binary_devices/pump.dm]: Without this, it doesn't automatically update the icon on power change. You can see the pump manually calls update_icon() in its Topic proc below, too. Not sure if there's an automatic system in place for this already that's just not getting called.

Comment (afterthought2) [reply code/modules/atmospherics/components/binary_devices/pump.dm]: Oh yeah, this was not the proc I thought it was. This one you have to do manually.

File: code/game/machinery/doors/airlock.dm (hunk 6 above)
Comment (afterthought2) [reply code/game/machinery/doors/airlock.dm]: It's probably better to have network device datums use `check_access`, which is the more modern and saner version of `allowed`. They do the same thing except on some specific objs (like doors, here) where `allowed` carries legacy connotations I'm not familiar with.

File: code/modules/modular_computers/networking/device_types/_network_device.dm [hunk 10]
```diff
@@ -196,6 +221,156 @@
...
+	if(command_and_call[command])
+		var/decl/public_access/public_method/method = command_and_call[command]
+		method.perform(command_target, command_args)
+		return "Successfully called method '[command]' on [network_tag]."
+	if(command_and_write[command])
+		var/decl/public_access/public_variable/variable = command_and_write[command]
+		if(command_args) // Write to a var.
+			command_args = sanitize_command_args(command_args, variable.var_type)
+			if(isnull(command_args)) // Command could not be converted into a form appropriate for the variable.
+				return "Improper argument type; correct type is [variable.var_type]"
```
afterthought2, line 296: It's possible a null literal is a valid argument in some cases, though am not sure. Just be careful if you move this off to vars, as e.g. radio stuff may be passing nulls into num or text stuff.

code/modules/modular_computers/networking/device_types/_network_device.dm
```diff
@@ -196,6 +221,156 @@
...
+
+// Calls a random method for skill failure etc.
+/datum/extension/network_device/proc/random_method(var/user)
+	if(world.time < last_rand_time + 5 SECONDS)
+		return "Reinstancing command system, please try again in a few moments."
+	if(user && !has_access(user))
+		return "Access denied"
+	var/rand_alias = pick(command_and_call)
+	var/decl/public_access/public_method/rand_method = command_and_call[rand_alias]
+	rand_method.perform(get_command_target(), list())
```
afterthought2: Kind of weird to pass an empty list, no?

code/modules/modular_computers/networking/device_types/stock_part.dm
```diff
@@ -0,0 +1,20 @@
...
+// Network receivers check the access of the parent machine, and do not check for network administrator access.
+/datum/extension/network_device/stock_part/has_access(mob/user)
+	var/datum/computer_network/network = get_network()
+	if(!network)
+		return TRUE // If not on network, always TRUE for access, as there isn't anything to access.
+	if(!user)
+		return FALSE
+	var/atom/H = holder
+	var/obj/M = H.loc
+	return M.allowed(user)
```
afterthought2: Yeah, here I recommend `check_access` (and on the base version of this proc).

code/modules/modular_computers/networking/device_types/_network_device.dm
NataKilar: Looking through the writable vars, it doesn't look like any take in null as a possible value. The actual public variable var-type checking I added should still permit it, though, if the var-type is IC_FORMAT_ANYTHING

maps/tradeship/backup/tradeship-2.dmm
```diff
@@ -0,0 +1,788 @@
+"aa" = (/turf/space,/area/space)
```
comma: probably shouldn't commit backup maps

code/modules/power/fission/core.dm
```diff
@@ -0,0 +1,182 @@
+#define ACTIVE_THRESHOLD             200
```
comma: is all the fission stuff intended to be in this PR? I see some related bits where controller is, but that's a lot of fissionstuff so unsure

code/modules/modular_computers/terminal/terminal_commands.dm
```diff
@@ -223,3 +241,479 @@ Subtypes
...
+	pattern = @"^rename"
+
+/datum/terminal_command/rename/proper_input_entered(text, mob/user, datum/terminal/terminal)
+	if(!terminal.current_disk)
+		return "rename: No disk selected."
+
+	var/list/rename_args = get_arguments(text)
+
+	if(length(rename_args) < 2)
+		return "rename: Improper syntax, use move \[file name\] \[new name\]."
```
comma: >use move 

rename?

code/modules/power/fission/core.dm
NataKilar: Nope, my mistake. Accidentally staged the files when I was rebasing. Should be fixed now.

MistakeNot4892 (approved):

PR #1447, opened 2021-03-08
Wide-scale MMI and brainmob refactor. [MDB IGNORE]

MistakeNot4892:
- Repaths brains from carbon to living.
- Deletes a bunch of brain environmental processing. Brains are an organ, and organs die on their own regardless; might as well let the mob take state from the brain instead of handling it separately.
- Reworks ~~some~~ all MMI and brainmob code. Brainmobs are retrieved with a helper proc on /organ/internal to avoid having to subtype /brain repeatedly.

**Issues to resolve:**
- ~~Blindness on the brain is inconsistent/fades in and out.~~
- ~~Bunch of organ and client runtimes alongside testing.~~
- ~~Brains die in MMIs.~~
- ~~Brains have no languages.~~
- ~~MMIs do not transfer key to the mob on reattachment and remain in the implants list.~~


MistakeNot4892:
This is pretty close to done, just need to fix an issue with the MMI deleting itself on removal.

MistakeNot4892:
Will reopen when this is mergeable.

MistakeNot4892:
This appears to be functionally working so ready for review, just has some small stuff to fix.

MistakeNot4892:
Note to self to update/retest.

MistakeNot4892:
Not going to get this properly tested and working.

MistakeNot4892:
Ended up completely reworking this, oops.

code/modules/mob/living/silicon/robot/component.dm
```diff
@@ -1,4 +1,4 @@
-// TODO: remove the robot.mmi and robot.cell variables and completely rely on the robot component system
+// TODO: remove the robot.brain and robot.cell variables and completely rely on the robot component system
```
out-of-phaze: isn't that var named `central_processor` now?

code/modules/organs/internal/brain.dm
```diff
@@ -15,50 +15,55 @@
...
+		_brainmob = new(src)
+	return _brainmob
+
+/obj/item/organ/internal/brain/Initialize()
+	. = ..()
+	if(species)
+		set_max_damage(species.total_health)
+	else
+		set_max_damage(200)
+	initialize_brain()
```
out-of-phaze: Should this be here? Before, brainmobs were only made in `transfer_identity`; it looks like this is why tests are failing, and besides, `create_if_missing` is used in the new equivalent of the `transfer_identity` proc. Why is it necessary to forcibly create a brainmob here, which has to register to its parent + creates a virtual mob + may cause all other sorts of mayhem?

out-of-phaze (changes requested):
Preliminary review to try to get this to pass tests.

code/modules/organs/internal/brain.dm
MistakeNot4892: Renaming `initialize_brain()` to `initialize_brainmob()` and calling it instead of `_brainmob = new` in `get_brainmob()`

code/modules/mob/living/silicon/robot/component.dm
MistakeNot4892: For reference I did a mass rename of `mmi` to `brain` then `brain` to `central_processor` so there'll be places I missed yeah.

code/modules/organs/internal/brain.dm
```diff
@@ -15,50 +14,58 @@
...
-	var/can_use_mmi = TRUE
-	var/mob/living/carbon/brain/brainmob = null
-	var/const/damage_threshold_count = 10
-	var/damage_threshold_value
-	var/healed_threshold = 1
+	transfer_brainmob_with_organ = TRUE
+	var/can_use_brain_interface = TRUE
+	var/should_announce_brain_damage = TRUE
 	var/oxygen_reserve = 6
+	var/VAR_PRIVATE/mob/living/_brainmob = /mob/living/brain
```
out-of-phaze:
```suggestion
	VAR_PRIVATE/mob/living/_brainmob = /mob/living/brain
```

out-of-phaze (dismissed):
Repeating the issue from the Discord so it can't get forgotten/lost:

Brainmob transfer isn't happening with decapitation because it relies on `organ/proc/on_remove_effects`, which is called in `mob/living/proc/on_lost_organ`, which is called in `mob/living/proc/remove_organ`... but not in `organ/proc/do_uninstall`.

When a body part is dismembered, it has `remove_organ` called, which calls `on_lost_organ` on the dismembered limb, but then calls `do_uninstall` to remove its child organs, so they never have `on_lost_organ` or `on_remove_effects` called. The way to solve this is to, well... fix that. Reorder the call chain to be consistent with removal effects.

code/modules/organs/external/head.dm
```diff
@@ -25,13 +25,13 @@
 	var/graffiti_style
 
 /obj/item/organ/external/head/proc/get_eye_overlay()
-	if(glowing_eyes || owner?.has_chemical_effect(CE_GLOWINGEYES, 1))
+	if(owner && (glowing_eyes || owner.has_chemical_effect(CE_GLOWINGEYES, 1)))
```
PsyCommando: This means the glowing eyes effect won't show on decapitated heads, no?

MistakeNot4892:
Issue with detachment turned out to be gating the transfer procs behind `get_brainmob()` after making it lazy, appears to be working now.

code/modules/organs/external/head.dm
MistakeNot4892: Fixed.

code/modules/organs/internal/brain.dm
MistakeNot4892: Fixed.

code/controllers/subsystems/initialization/robots.dm
```diff
@@ -10,18 +10,15 @@ SUBSYSTEM_DEF(robots)
 	var/list/robot_alt_titles            = list()
 
 	var/list/mob_types_by_title = list(
```
MistakeNot4892, line 12: This system needs declification to be modpack compatible, leaving a note for myself. May move shackle reimplementation into that system too.

MistakeNot4892:
Depends on #2811 #2812 #2815 and #2794.

MistakeNot4892:
Summary for future me when I've slept:
- Prosthetics do not set detached on removal for reasons that entirely escape me. This prevents set_detached() from being used as the brainmob transfer point.
- Heads move their internal organs into implants/detach them on removal and do not move them back on reattachment.

MistakeNot4892:
Noting the brain appeared to not have owner set when reinstalled.

out-of-phaze:
Also seems to rename "positronic brain" to "computer intelligence core"?

MistakeNot4892:
Right, yes. I remember now.
- Rename was to remove the positronics lore inherited from Bay along with the BP_POSIBRAIN and /mob/silicon/posimob or whatever it was. Lore-agnostic codebase etc. It should be straightforward for servers that want to make them positronics or readd a distinction between robots and positronics to make a subtype or change the lore strings. The icons are all still the same posibrains iirc.
- `/obj/item/organ/internal/brain/robotic` is separate from `brain_interface` because it doesn't need any of the additional logic that comes with the brain interface. The interface object that existed previously for posibrains/MMIs was solely because the MMI/posibrain wasn't an organ and had to have a wrapper.

out-of-phaze (dismissed):

out-of-phaze (dismissed):

out-of-phaze (dismissed):

out-of-phaze (dismissed):

out-of-phaze (dismissed):

out-of-phaze (approved):

PR #1450, opened 2021-03-09
Coulple small fixes for natural weapons

comma:
Makes them no longer embed
Changes one check which would fail if weapon wasn't instancized before it

Yoinked them from Ilysen's cult PR over on bay

### DIFF
--- a/code/_onclick/other_mobs.dm
+++ b/code/_onclick/other_mobs.dm
@@ -82,7 +82,7 @@
 		return
 	setClickCooldown(DEFAULT_ATTACK_COOLDOWN)
 	if(istype(A,/mob/living))
-		if(!istype(natural_weapon) || a_intent == I_HELP)
+		if(a_intent == I_HELP || !get_natural_weapon())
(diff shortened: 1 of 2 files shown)


code/_onclick/other_mobs.dm
```diff
@@ -82,7 +82,7 @@
 		return
 	setClickCooldown(DEFAULT_ATTACK_COOLDOWN)
 	if(istype(A,/mob/living))
-		if(!istype(natural_weapon) || a_intent == I_HELP)
+		if(!get_natural_weapon() || a_intent == I_HELP)
```
MistakeNot4892: Maybe move this after the intent comparison? Tiny microoptimization.

MistakeNot4892 (dismissed):

code/_onclick/other_mobs.dm
comma: sure why not

MistakeNot4892 (approved):

PR #1451, opened 2021-03-09
Generalizes unit test map

PsiOmegaDelta:
Moves the safe and space landmarks to the unit test map.
Makes it so that main maps (Torch, Exodus, Example,etc.) do not have to map these landmarks themselves.

### DIFF
--- a/code/unit_tests/unit_test.dm
+++ b/code/unit_tests/unit_test.dm
@@ -82,8 +82,9 @@ var/ascii_reset = "[ascii_esc]\[0m"
 	reported = 1
 	log_unit_test("[ascii_yellow]--- SKIPPED --- \[[name]\]: [message][ascii_reset]")
 
+// Executed before the test runs - Primarily intended for shared setup (generally in templates)
 /datum/unit_test/proc/setup_test()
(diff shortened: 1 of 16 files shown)


code/unit_tests/unit_test.dm
```diff
@@ -92,8 +93,32 @@ var/ascii_reset = "[ascii_esc]\[0m"
...
 
+// Executed after the test has run - Primarily intended for shared cleanup (generally in templates)
 /datum/unit_test/proc/teardown_test()
-	// Executed after the test has run - Primarily intended for shared cleanup (generally in templates)
+	SHOULD_CALL_PARENT(TRUE)
+	var/failed = FALSE
+
+#ifdef UNIT_TEST
+	if(!async) // Async tests run at the same time, so cleaning up after any one completes risks breaking following tests
+		var/ignored_types = list(/atom/movable/lighting_overlay, /obj/effect/landmark)
```
MistakeNot4892, line 103: May also want to add `/obj/effect/flood` here.

MistakeNot4892 (approved):

PR #1452, opened 2021-03-09
Optimizes queue_icon_update, reducing load times drastically

Andrew-Fall:
Changing from an associative list to two indexed lists cuts load times on large maps down drastically.
Also repaths SSicon_update from a processing subsystem, as it doesn't use the default /fire
## Changelog
:cl:
tweak: Drastically increases large map loading speed
/:cl:


code/controllers/subsystems/icon_updates.dm
```diff
@@ -0,0 +1,71 @@
+SUBSYSTEM_DEF(icon_update)
+	name = "Icon Updates"
+	wait = 1	// ticks
+	flags = SS_TICKER
+	priority = SS_PRIORITY_ICON_UPDATE
+	init_order = SS_INIT_ICON_UPDATE
+
+	// Linked lists, queue_refs[x] should have TRUE or args stored in queue_args[x]
+	var/list/queue_refs = list()	// Atoms
+	var/list/queue_args = list()	// TRUE or args
```
Lohikar: This contains `null` if no args, not TRUE.

code/controllers/subsystems/icon_updates.dm
```diff
@@ -0,0 +1,71 @@
...
+
+/datum/controller/subsystem/icon_update/stat_entry()
+	..("Queue Refs: [queue_refs.len], Queue Args: [queue_args.len]")
```
Lohikar: These two lists should always be the same length so just something like `Q:[queue_refs.len]` would be enough.

code/controllers/subsystems/icon_updates.dm
```diff
@@ -0,0 +1,71 @@
...
+	
+		if(Master.map_loading)
+			return
+
+		// Pops the atom and it's args, while making sure
+		// to keep them in sync in case runtimes/badmin
+		var/atom/A = queue_refs[queue_refs.len]
+		var/myArgs = queue_args[queue_args.len]
+
+		queue_args.len = --queue_refs.len
```
Lohikar: Ew, pls no:
```
queue_args.len -=1
queue_refs.len -= 1
```

code/controllers/subsystems/icon_updates.dm
```diff
@@ -0,0 +1,71 @@
...
+			return
+
+/atom
+	var/icon_update_queued = FALSE
+
+/atom/proc/queue_icon_update(...)
+	// Skips if this is already queued
+	if(!icon_update_queued)
+
+		icon_update_queued = TRUE
```
Lohikar, line 56: This is never unset.

code/controllers/subsystems/icon_updates.dm
```diff
@@ -0,0 +1,71 @@
...
+		SSicon_update.queue_refs.Add(src)
+
+		// Makes sure these are in sync, in case runtimes/badmin
```
Lohikar, line 59: If they're not the same length, the list is invalid anyways.

Andrew-Fall: Lowers the risk, and doesn't really change functionality, as queue_args len needs to be increased anyways. Looks clean too

MistakeNot4892 (dismissed):

Lohikar (approved):

code/controllers/subsystems/icon_updates.dm
```diff
@@ -0,0 +1,73 @@
+SUBSYSTEM_DEF(icon_update)
+	name = "Icon Updates"
+	wait = 1	// ticks
+	flags = SS_TICKER
```
Lohikar: This shouldn't be a ticker. Remove this flag please.
(You'll need to also move the define in subsystem-priority.dm)

Lohikar (changes requested):

MistakeNot4892 (dismissed):

Lohikar (approved):
Looks good, though this should be squash-merged.

PR #1458, opened 2021-03-13
Makes shuttles rebuild pipenets on move

afterthought2:
From some limiting testing, appears to work and is not perceptibly adding to cost. Obviously is worse with very large shuttles.

### DIFF
--- a/code/modules/shuttles/shuttle.dm
+++ b/code/modules/shuttles/shuttle.dm
@@ -224,7 +224,6 @@
 					bug.gib()
 				else
 					qdel(AM) //it just gets atomized I guess? TODO throw it into space somewhere, prevents people from using shuttles as an atom-smasher
-	var/list/powernets = list()
 	for(var/area/A in shuttle_area)
 		// if there was a zlevel above our origin, erase our ceiling now we're leaving
 		if(HasAbove(current_location.z))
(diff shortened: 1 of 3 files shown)


comma (dismissed):

code/modules/shuttles/shuttle.dm
```diff
@@ -243,8 +240,37 @@
...
+// Remove all powernets and pipenets that were affected, and rebuild them.
+/datum/shuttle/proc/handle_pipes_and_power_on_move(var/list/new_turfs)
+	var/list/powernets = list()	
 	var/list/cables = list()
+	var/list/pipenets = list()
+	var/list/pipelines = list()
+	var/list/pipes = list()
+
+	for(var/turf/T in new_turfs)
+		for(var/obj/structure/cable/cable in T)
```
MistakeNot4892, line 271: Would this be faster as `for(var/thing in T)` and then typechecking?

code/modules/shuttles/shuttle.dm
```diff
@@ -253,14 +279,17 @@
...
 			propagate_network(C,C.powernet)
-
-	if(mothershuttle)
-		var/datum/shuttle/mothership = SSshuttle.shuttles[mothershuttle]
-		if(mothership)
-			if(current_location.landmark_tag == motherdock)
-				mothership.shuttle_area |= shuttle_area
-			else
-				mothership.shuttle_area -= shuttle_area
+	for(var/datum/pipe_network/pipenet in pipenets)
```
MistakeNot4892: Can you rely on `powernet` and `return_network()` to return correct instances? If so you could `as anything` these loops.

MistakeNot4892:
Other than optimization notes on the loops this looks fine, happy to approve for merge if you don't want to do the loop stuff.

afterthought2:
I don't know enough about powernets to know if it's safe to do `as anything` there, but I did fix up the pipe ones. The power I left the way it was.

MistakeNot4892 (dismissed):

comma (approved):

PR #1472, opened 2021-03-19
Minor overmap fixes.

MistakeNot4892:
- Fixes dark overmap.
- Fixes ghost runtimes when looking through nav.


afterthought2 (approved):
Well, OK, but it's supposed to be doing if(area.not_dynamic || turf.not_dynamic), not &&, pretty sure.

MistakeNot4892:
Re: the above comment, turns out it's not. Lohi has clarified that the idea is apparently that either the area or the turf can opt out of dynamic lighting.
Still not sure what's going on with the overmap lighting rebuilds so will need further investigation.

PR #1483, opened 2021-03-23
Ports Hud Implants from baystation

quardbreak:
Ports https://github.com/Baystation12/Baystation12/pull/30153

### DIFF
--- a/code/modules/augment/active/hudimplant.dm
+++ b/code/modules/augment/active/hudimplant.dm
@@ -1,0 +1,60 @@
+/obj/item/organ/internal/augment/active/hud
+	name = "integrated HUD"
+	desc = "A small implantable heads-up display."
+	icon_state = "booster"
+
+	action_button_name = "Toggle integrated HUD"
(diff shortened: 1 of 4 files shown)


code/modules/augment/active/hudimplant.dm
```diff
@@ -0,0 +1,60 @@
...
+
+	origin_tech = "{'magnets':2}"
+
+	var/active = FALSE
+
+/obj/item/organ/internal/augment/active/hud/Process()
+	..()
+	if (owner && active)
+		if (allowed(owner))
+			callback_hud_to_process(owner, 1)
```
afterthought2, line 22: Not a huge fan of this implementation: it's fragile, can have ordering problems, and doesn't update the mob's hud vars properly. I feel a better way to do this is to have mobs look through their organs for hud flags and add them before updating huds in Life. This can then set whether it has a hud flag (though I don't think the `allowed` check makes any sense).

quardbreak:
Will revisit later, for now - closed

PR #1485, opened 2021-03-23
Synced RUST field processing to core processing.

MistakeNot4892:
Thank you to @Atermonera for pointing out what was causing the wild power fluctuation issue.


afterthought2 (approved):
Wow, an actually sane implementation of this (take a look in atmos code for some worst practices examples).

PR #1486, opened 2021-03-24
Custom icon and refit refactors.

MistakeNot4892:
- Removes `icon_override`, it is outdated with the current system.
- Converts the custom item system to load single icons from the filesystem rather than having them compiled into a shared icon.
- Rewrites suit cycler decls to populate their info from suit instances instead of hardcode.

Tested, works.


code/controllers/subsystems/initialization/customitems.dm
```diff
@@ -79,45 +73,46 @@ SUBSYSTEM_DEF(customitems)
...
+	apply_to_target_type = data["apply_to_target_type"]
+
+	try
+		character_ckey =       ckey(character_ckey)
+		character_name =       lowertext(character_name)
+		item_path =            item_path && text2path(item_path)
+		apply_to_target_type = apply_to_target_type && text2path(apply_to_target_type)
+		item_icon =            item_icon && file(item_icon)
+	catch(var/exception/E)
+		to_world_log("Exception while loading custom item: [E]")
```
afterthought2: This is sort of confusing, as it doesn't actually throw the error (meaning it's logged differently from a runtime). Simpler handling may be to just let it runtime.

code/modules/games/cards.dm
```diff
@@ -36,10 +36,17 @@
...
+
+				try
+					if(!isnull(card_decl["use_custom_front"]))
+						var/card_front = card_decl["use_custom_front"]
+						P.use_custom_front = fexists(card_front) && file(card_front)
+					if(!isnull(card_decl["use_custom_back"]))
+						var/card_back = card_decl["use_custom_back"]
+						P.use_custom_back = fexists(card_back) && file(card_back)
+				catch(var/exception/E)
+					to_world_log("Exception while loading card icons: [E]")
```
afterthought2: Same here; can pull it out into a separate proc to make sure it doesn't halt execution.

config/custom_items/README.md
```diff
@@ -1 +1,75 @@
...
+
+| Key                    | Expected Value   | Function                                                                                                                |
+|------------------------|------------------|-------------------------------------------------------------------------------------------------------------------------|
+| ckey                   | string           | Your ckey. This is not quite the same as your BYOND key, ask an admin or check the BYOND docs if you are unsure.        |
+| character_name         | string           | The name of the character the item should spawn with.                                                                   |
+| item_name              | string           | The name of your custom item ingame. For kits, the name of the kit product.                                             |
+| item_desc              | string           | The description of your custom item ingame. For kits, the descriptor for the kit product.                               |
+| item_icon              | file path        | The location of the icon file to load for your item (ex. icons/custom_icon_file.dmi)                                    |
```
afterthought2, line 14: May be worth having a standardized relative path for this (possibly as a config option/define?). This is always used by people with marginal understanding of project structure, and even I find navigating the file tree confusing when you have to include multiple `..`, etc.

afterthought2 (commented):
Huge fan of the concept. The explanation I always got when I asked why it didn't work this way already is that since the icon files are not directly referenced in the compiled code here, they aren't sent to the rsc. Is this not a problem/avoided/irrelevant with this implementation?

config/custom_items/README.md
MistakeNot4892: I could make it use a config value but if anything that's adding a step to the confusion as the server could set it to anything.

MistakeNot4892:
Wrt: the RSC, yes, the files have to be physically present in a dir next to the binary. Places like Bay would need to adjust their workflow significantly and copy the icons over along with the binary, etc.

code/modules/games/cards.dm
MistakeNot4892: Done.

config/custom_items/README.md
MistakeNot4892: Done, commented out by default.

code/controllers/subsystems/initialization/customitems.dm
MistakeNot4892: Done.

afterthought2 (approved):

MistakeNot4892:
Accidental selfmerge :(

PR #1490, opened 2021-03-25
Adds auxtools support for debugging

quardbreak:
Implements [auxtools](https://github.com/willox/auxtools) support which will be helpful for debugging with SpaceDMM.
Moves debugging GetConfig() into separate folder where it calls via hooks.

### DIFF
--- a/SpacemanDMM.toml
+++ b/SpacemanDMM.toml
@@ -4,3 +4,6 @@ dreamchecker = true
 [code_standards]
 disallow_relative_type_definitions = true
 disallow_relative_proc_definitions = true
+
+[debugger]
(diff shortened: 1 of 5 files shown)


PsiOmegaDelta:
How does, or doesn't, this affect debugging using Visual Studio Code?

quardbreak:
You use debugging function (F5) in Visual Studio Coder with SpacemanDMM and it attach debugger into game, allowing to do some debugging functions. It doesn't work without spacemandmm so it won't work on live game server.

MistakeNot4892 (approved):
Seems fine as it won't run on a live server as discussed on Discord.

PR #1503, opened 2021-03-26
Ports some presets from Starlight

quardbreak:
* Adds variable for fabricators, which will set max materials during Initialize()
* Adds some network mainframe presets for mapping

### DIFF
--- a/code/modules/fabrication/_fabricator.dm
+++ b/code/modules/fabrication/_fabricator.dm
@@ -57,6 +57,9 @@
 
 	var/species_variation = /decl/species/human // If this fabricator is a variant for a specific species, this will be checked to unlock species-specific designs.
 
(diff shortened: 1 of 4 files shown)


code/modules/fabrication/_fabricator.dm
```diff
@@ -97,6 +99,10 @@
 				var/decl/material/reg = mat
 				stored_substances_to_names[mat] = lowertext(initial(reg.name))
 
+	if(prefilled)
```
MistakeNot4892, line 103: This would be better as an override of Initialize on a mapped subtype imo.

afterthought2: To be clear, while maybe I have minor quibbles on implementation, I agree that this should be done using subtypes.

MistakeNot4892 (approved):

PR #1506, opened 2021-03-27
Display OOC notices for admin paralysis

SierraKomodo:
This should save people the step of having to constantly inform passers-by to not touch someone while they're being bonked.

## Changelog
:cl:
rscadd: Admin paralyzed players now have notices to inform other players they're paralyzed and being handled by staff, and to not interact until staff are done.
/:cl:

### DIFF
--- a/code/modules/admin/admin.dm
+++ b/code/modules/admin/admin.dm
@@ -122,7 +122,7 @@ var/global/floorIsLava = 0
 		if(extra_body)
 			body += "<br><br>"
 			body += extra_body
-			
+
 	if (M.client)
 		if(!istype(M, /mob/new_player))
 			body += "<br><br>"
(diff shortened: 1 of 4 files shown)


code/modules/admin/admin.dm
```diff
@@ -1447,10 +1447,18 @@ var/global/floorIsLava = 0
...
+				SPAN_OCCULT("OOC: \The [M] has been paralyzed by a staff member. Please hold all interactions with them until staff have finished with them."),
+				SPAN_OCCULT("OOC: You have been paralyzed by a staff member. Please refer to your currently open ahelp ticket or, if you don't have one, ahelp for assistance.")
+			)
 			M.set_status(STAT_PARA, 8000)
+			M.admin_paralyzed = TRUE
 		else
 			M.set_status(STAT_PARA, 0)
+			M.admin_paralyzed = FALSE
+			M.visible_message(SPAN_OCCULT("OOC: \The [M] has been released from paralysis by staff. You may resume interactions with them."))
+			to_chat(M, SPAN_OCCULT("OOC: You have been released from paralysis by staff and can return to your game."))
```
SierraKomodo, line 1454: The mob doesn't seem to immediately wakeup, and thereby can't see the visible message, so this uses a to_chat for the mob instead.

code/modules/admin/admin.dm
```diff
@@ -1447,10 +1447,18 @@ var/global/floorIsLava = 0
 		return
 
 	if(check_rights(R_INVESTIGATE))
-		if(!HAS_STATUS(M, STAT_PARA))
+		if(!M.admin_paralyzed)
+			M.visible_message(
+				SPAN_OCCULT("OOC: \The [M] has been paralyzed by a staff member. Please hold all interactions with them until staff have finished with them."),
+				SPAN_OCCULT("OOC: You have been paralyzed by a staff member. Please refer to your currently open ahelp ticket or, if you don't have one, ahelp for assistance.")
```
MistakeNot4892: Would suggest using `Adminhelp` rather than `ahelp` just because the verb isn't called ahelp.

SierraKomodo: Updated

MistakeNot4892 (approved):

PR #1511, opened 2021-03-28
Added overhead indicators for some status effects.

MistakeNot4892:
Stolen shamelessly from @Neerti's Polaris feature, though the code and icons are entirely new.
Can be disabled in prefs for people who find them jarring.

- If the mob is unconscious, no status indicators are shown other than sleeping.
- If the mob is dead, no status indicators are shown at all.
- Conscious mobs will be visibly marked if stunned, confused or weakened.
- You will always be able to see your own status effects, both the three above as well as mute, blind, deaf, paralyzed or drugged.

Video of the animation in action (bit outdated compared to above):
https://user-images.githubusercontent.com/2468979/112756694-2779f280-9032-11eb-99d4-ff969d839c74.mp4

:cl:
rscadd: Some status effects (sleeping, stunned, confusion, weakened) are visible in the form of markers over the victim's head. This can be disabled in preferences if you'd rather not see them.
/:cl:


MistakeNot4892:
Still working on blindness/deafness but otherwise should be good for review.

MistakeNot4892:
Ready to go.

Koollan:
Is there a way to set these to pref'd off as default on a downstream, but still give the ability to turn them on as a player?

MistakeNot4892:
You'd have to reorder the preference options to change the default, which would need to be done in the code.

code/modules/mob/login.dm
```diff
@@ -115,6 +115,16 @@
 				var/spell/S = screen.spell
 				mind.learned_spells |= S
 
+	var/show_status_markers = get_preference_value(/datum/client_preference/show_status_markers) == GLOB.PREF_SHOW
+	for(var/obj/status_marker_holder/marker as anything in global.status_marker_holders)
+		client.images -= marker.mob_image
+		client.images -= marker.mob_image_personal
```
afterthought2: Why do we do this if the preference is don't show? They should have been removed on pref change or otherwise never added.

afterthought2 (commented):
Don't really follow the `vis_contents` manipulations going on here, but if it works it works.

code/modules/mob/login.dm
MistakeNot4892: I figured it was harmless and the extra check wasn't really necessary. I can add it back in if you like.

afterthought2: Well, it's in Login so on the one hand doesn't matter; on the other hand having the whole loop not be run only if the pref is to show makes it strange. Don't care either way.

MistakeNot4892: Right, misunderstood, the purpose of this loop is just to refresh client images. I pruned out the `-=` and tidied the block a bit.

afterthought2 (approved):

PR #1512, opened 2021-03-28
Z-Mimic: Actually Working edition

Lohikar:
PR title not guaranteed to be accurate.

changes:
- Layering should actually work now, even for stacked non-OVERWRITE turfs - some edge cases with bigturfs and space remain.
- Pointless non-head updates of z-turfs are just skipped now.
- Z-Copy has been told to let go and not attempt to update de-queued zturfs.
- Openturf analyzer now shows absolute z-stack ordering, rather than throwing a pile of numbers at you and letting you figure it out.
- Renamed some Z-M types to be less misleading.
- ~~Some mouse opacity changes that do things I forget, but presumably they're important~~ Mimics of non-OVERWRITE turfs are now not clickable so they won't spam the right click menu.
- Z-Stack messages have been made a bit clearer & now show more information.
- Z-Mimic will now print logs if it finds a Z-Stack taller than what Z-Mimic can deal with without loss of layering information.
- Lighting mimics now respect SSoverlays, not that they're actually used right now.

This port has not been tested at all beyond 'it compiles' (the upstream O7 version *has* been tested, though).

### DIFF
--- a/code/controllers/subsystems/zcopy.dm
+++ b/code/controllers/subsystems/zcopy.dm
@@ -1,6 +1,7 @@
 #define OPENTURF_MAX_PLANE -70
 #define OPENTURF_MAX_DEPTH 10		// The maxiumum number of planes deep we'll go before we just dump everything on the same plane.
 #define SHADOWER_DARKENING_FACTOR 0.6	// The multiplication factor for openturf shadower darkness. Lighting will be multiplied by this.
+#define SHADOWER_DARKENING_COLOR "#999999"	// The above, but as an RGB string for lighting-less turfs.
 
 SUBSYSTEM_DEF(zcopy)
 	name = "Z-Copy"
@@ -19,7 +20,7 @@ SUBSYSTEM_DEF(zcopy)
 
 	// Highest Z level in a given Z-group for absolute layering used by FIX_BIGTURF.
 	// zstm[zlev] = group_max
-	var/list/zstack_maximums = list()
+	var/list/zlev_maximums = list()
 
 // for admin proc-call
 /datum/controller/subsystem/zcopy/proc/update_all()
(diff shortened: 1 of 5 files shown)


Lohikar:
I'm pretty sure that this'll work; the same code was tested as working on another codebase, so it's unlikely to break.

code/controllers/subsystems/zcopy.dm
```diff
@@ -88,15 +89,19 @@ SUBSYSTEM_DEF(zcopy)
...
-	zstack_maximums = new(world.maxz)
+	zlev_maximums = new(world.maxz)
 	var/start_zlev = 1
 	for (var/z in 1 to world.maxz)
 		if (!HasAbove(z))
 			for (var/member_zlev in start_zlev to z)
-				zstack_maximums[member_zlev] = z
+				zlev_maximums[member_zlev] = z
+			if (z - start_zlev > OPENTURF_MAX_DEPTH)
+				log_ss("zcopy", "WARNING: Z-levels [start_zlev] through [z] exceed maximum depth of [OPENTURF_MAX_DEPTH]; layering may behave strangely in this Z-stack.")
```
afterthought2, line 99: Do they actually behave strangely? This is in principle a supported case if your map has enough z levels, right?

afterthought2 (approved):

code/controllers/subsystems/zcopy.dm
Lohikar: After 10 levels, Z-Mimic *should* just start dumping everything on the same layer. I'm not sure if this behavior still works though, it's not tested.

Lohikar: If you actually encounter this error in practice, you can also just increase MAX_DEPTH. It's more or less arbitrary (can't see further than like 5 turfs down anyways), only real limit is bumping into planes below z-mimic's set.

PR #1513, opened 2021-03-29
Removed force_ending variable, replaced with direct calls

quardbreak:
* ~~Adds option for restart button allowing restart without MC shutdown.~~
* Removed `force_ending` variable, replaced with direct calls

### DIFF
--- a/code/game/world.dm
+++ b/code/game/world.dm
@@ -450,6 +450,8 @@ var/world_topic_last = world.timeofday
 		to_world("<span class=danger>World reboot waiting for external scripts. Please be patient.</span>")
 		return
 
+	game_log("World rebooted at [time_stamp()]")
+
 	..(reason)
 
 /world/Del()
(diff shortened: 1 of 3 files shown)


code/game/world.dm
```diff
@@ -435,11 +435,12 @@ var/world_topic_last = world.timeofday
 
 #undef SET_THROTTLE
 
-/world/Reboot(var/reason)
+/world/Reboot(reason, hard_reboot = FALSE)
 	if(GLOB.using_map.reboot_sound)
 		sound_to(world, sound(pick(GLOB.using_map.reboot_sound)))// random end sounds!! - LastyBatsy
 
-	Master.Shutdown()
+	if(!hard_reboot)
```
Lohikar: I don't think this should be exposed to admins.

There shouldn't be any reason for them to need it, and the code assumes that Shutdown() will run if the round ends without daemon crashing.

MistakeNot4892: I agree, if the server is entering a state where it can't be rebooted, that's an issue that should be solved, not something to paper around with something like this.

MistakeNot4892 (approved):

PR #1516, opened 2021-03-30
Refactoring areas and area path dependencies.

MistakeNot4892:
- Areas now use flags instead of typechecks for the majority of area-dependent logic.
- Prison break area lists have been cleared and will need to be set up per-map instead of in main code.
- Outdated area definitions have been deleted.


code/_helpers/areas.dm
```diff
@@ -93,7 +92,7 @@
 	. = !istype(A,/area/space)
 
 /proc/is_not_shuttle_area(var/area/A)
-	. = !istype(A,/area/shuttle)
+	. = !(istype(A) && (A.area_flags & AREA_FLAG_SHUTTLE))
```
comma: is `!(istype(A)` right here? Wouldn't this always fail on areas, flag or not (since areas are usually of type area)

code/_helpers/areas.dm
```diff
@@ -102,7 +101,7 @@
 	. = !is_area_with_turf(A)
 
 /proc/is_maint_area(var/area/A)
-	. = istype(A,/area/maintenance)
+	. = !(istype(A) && (A.area_flags & AREA_FLAG_MAINTENANCE))
```
comma: ditto

code/_helpers/areas.dm
MistakeNot4892: Discussed on Discord, rewrote the logic to be clearer.

code/_helpers/areas.dm
MistakeNot4892: As above.

comma (approved):

PR #1519, opened 2021-04-01
fix skill panel losing focus on button presses

SierraKomodo:
Ported from https://github.com/Baystation12/Baystation12/pull/30455

## Changelog
:cl: MuckerMayhem (Ported by SierraKomodo)
bugfix: Fixed the skill panel losing focus when editing skills.
/:cl:

### DIFF
--- a/code/modules/client/preference_setup/occupation/occupation.dm
+++ b/code/modules/client/preference_setup/occupation/occupation.dm
@@ -329,7 +329,7 @@
 				set_to = JOB_LEVEL_NEVER
 			else if(set_to > JOB_LEVEL_NEVER)
 				set_to = JOB_LEVEL_HIGH
-		if(SetJob(user, set_job, set_to)) 
+		if(SetJob(user, set_job, set_to))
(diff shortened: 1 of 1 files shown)


MistakeNot4892 (dismissed):

afterthought2:
The problem is that if you don't refresh at all (i.e. don't return `TOPIC_REFRESH` here), you'll break anything in the main window which changes on skill value change (notably the colors in the job window and stuff in loadout).

SierraKomodo:
Yep, re-tested and confirmed that's an issue. I'll figure out a fix if Mucker hasn't by the time I'm done with work today.

SierraKomodo:
Updated with an alternative fix Mucker came up with that updates the parent window while keeping focus on the skill panel.

afterthought2 (approved):
For what it's worth, I've tested this on bay before and at the time this was worse for about 2/3 of the people on live and better for 1/3. Local testing is not particularly meaningful with this stuff.

PR #1524, opened 2021-04-07
O7 Lighting

Lohikar:
Pulls in the version of lighting from O7, which brings:
- More vibrant colors.
- Support for directional lighting.
    - Supports up to 180° cones, minimal performance impact.
- Removal of darksight.
    - Can be readded, but out of scope for this PR.
- Synchronous updates if server load is low enough.
    - This is mostly to improve responsiveness of lighting.
- Bidirectional Z-lights via corner z-bleed.
    - This tends to be more accurate + makes the occlusion loop simpler.
- Simpler lighting opacity calcs.
    - Now binary, rather than a counter.
    - This is a bit more robust, plus a bit faster.
- Lighting now uses Destroy() for cleanup rather than its own thing.
- Single-light atoms allocate 1 less list.
- Support for turf-offset lights.
    - This allows falloff to be calculated from where the light appears to be, rather than where BYOND thinks it is.
    - Mostly useful for light fixtures or wall-mounts.

Other changes:
- Only turf types that actually need an update_icon on init now update icon on init.
    - Only `/turf/exterior` (excl. `/turf/exterior/wall`) and `/turf/simulated/floor` seem to actually need this.
    - This also removes the two SKIP turf flags, as they're now redundant.
- atom/Initialize has been copied into turf/Initialize to remove some overhead.
- The opacity counter test has been removed as there's no longer an opacity counter.
- Light tubes no longer have random tints, instead defaulting to a cold shade of white.
- The color of ready/emergency lights was made a bit less intense (should be more colorful overall still due to changes in color mixing.)

To-do:
- [x] Update all set_light() calls and lighting vars to the new ones
- [x] Remove dummy vars / define
- [x] Set lighting offset vars on lighting fixtures?
- [ ] Testing


code/unit_tests/observation_tests.dm
```diff
@@ -362,111 +362,3 @@
...
-	T.set_opacity(TRUE)
-
-/datum/unit_test/observation/opacity_simple/unsetting_opacity_of_opaque_turf_shall_only_trigger_once
-	name = "Unsetting opacity of opaque turf shall only trigger once"
-	expected_opacity_unsets = 1
-	expected_opaque_counter = 0
-	turf_landmark_type = /obj/effect/landmark/proximity_wall
-
-/datum/unit_test/observation/opacity_simple/unsetting_opacity_of_opaque_turf_shall_only_trigger_once/ChangeOpacity(turf/T)
-	T.set_opacity(FALSE)
```
PsiOmegaDelta, line 472: I'll note that these tests are fairly important to ensure that proximity listener code works both correctly and decently-ish efficiently (while we do have dedicated proximity tests, they don't keep track of the number of actual event triggers)

Lohikar: I don't think it's really possible to port them to binary opacity without adding a counter specific to this test though.

MistakeNot4892 (approved):
To be completely honest I have no idea how most of this works and reviewing 1500 lines of what might as well be ancient Sumerian glyphs won't mean much. Willing to trust Lohi and go ahead with this merge if nobody else objects/has changes to request.

Lohikar:
Whoever merges this, pls _don't_ squash - want to preserve commit messages + quard's commits.

PR #1525, opened 2021-04-07
adds a standard glass floor type

Mazianni:
🆑 
:rscadd: adds a new glass floor type
🆑


code/game/turfs/flooring/flooring_premade.dm
```diff
@@ -367,3 +367,12 @@
 	icon = 'icons/turf/flooring/crystal.dmi'
 	icon_state = ""
 	initial_flooring = /decl/flooring/crystal
+
+/turf/simulated/floor/glass
+	icon = 'icons/turf/flooring/glassfloor.dmi'
+	icon_state = "glassfloor"
+	z_flags = ZM_MIMIC_DEFAULTS
```
MistakeNot4892: You should set these flags in `set_flooring` based on the flooring decl.

MistakeNot4892 (approved):

PR #1526, opened 2021-04-08
Send all nanoui templates in one file.

afterthought2:
Proof of concept aimed at people who have trouble with nanoui assets. Seems that asset send rate is (partially) limited by number of files (it sends a couple a second at most), and there are a lot of templates. This batches them all into one big file and sends only that.

Current file size is 376 KB. This is about 1.5x the size of the next largest asset, and 3x the size of each of two js libraries sent. It does store a copy in data; unsure if that's necessary or not.

Disadvantages: client needs to parse the file on UI open (seems fast enough to not be noticeable). Sending a big file may produce lag (unable to test). Code is bad (sorry).

Advantages: should speed up how early in the game you see nanoui assets, which is relevant in dev (less so in play). May or may not work better for people with nanoui troubles.

### DIFF
--- a/code/modules/client/asset_cache.dm
+++ b/code/modules/client/asset_cache.dm
@@ -174,6 +174,8 @@ You can set verify to TRUE if you want send() to sleep until the client has the 
 
 
 //DEFINITIONS FOR ASSET DATUMS START HERE.
+var/template_file_name = "all_templates.json"
+
 /datum/asset/nanoui
 	var/list/common = list()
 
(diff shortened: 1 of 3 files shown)


afterthought2:
I think it's fine to try this on a live server if/when there is interest. The performance hit from opening the local json is negligible as far as I can tell: client-side js processing which does not involve render tasks is probably the single fastest thing you can do in dm. The only concern is actually sending the file, which needs testing.

In a purely local setup (i.e. for dev) this implementation is just better than current, as the network issue is irrelevant.

I'm unsure of the best practices for dm-side file manipulation, so am happy to act on any advice regarding that.

In any case, removing `do not merge` tag: feel free to review, merge, give feedback, etc.

MistakeNot4892 (approved):
We can't really test this with no live server, but on the other hand it's not going to cause issues if it breaks something.

PR #1527, opened 2021-04-10
fix ice taste and add toolbelts for engineers

SolatK:


MistakeNot4892 (approved):
Usually you'd want this to be in two different commits but eh.

SolatK:
> 
> 
> Usually you'd want this to be in two different commits but eh.

I was thinking about doing two different pull requests, but meh. Too small changes.

PR #1528, opened 2021-04-11
adds supermatter hum

Mazianni:
todo: figure out how to swap the base sound around so the hum changes when it's delaminating.


code/modules/supermatter/supermatter.dm
```diff
@@ -159,6 +187,10 @@ var/list/supermatter_final_thoughts = list(
 	var/aw_delam = FALSE
 	var/aw_EPR = FALSE
 
+	var/datum/sound_token/sound_token
+	var/sound_id
+	var/working_sound
```
MistakeNot4892, line 192: Not clear on why you track `working_sound` at all since it only seems to be passed to the callback as an arg.

code/modules/supermatter/supermatter.dm
```diff
@@ -574,6 +607,30 @@ var/list/supermatter_final_thoughts = list(
 /obj/machinery/power/supermatter/get_artifact_scan_data()
 	return "Superdense crystalline structure - appears to have been shaped or hewn, lattice is approximately 20 times denser than should be possible."
 
+/obj/machinery/power/supermatter/proc/update_sound()
+	if(!sound_id)
+		sound_id = "[type]_[sequential_id(/obj/machinery/power/supermatter)]"
```
MistakeNot4892, line 612: Couldn't this just be `sound_id = "[sequential_id(type)]"`? 
Also any particular reason it isn't set in `Initialize()` rather than here?

code/modules/supermatter/supermatter.dm
```diff
@@ -574,6 +607,30 @@ var/list/supermatter_final_thoughts = list(
...
+
+	switch(damage)
```
MistakeNot4892, line 614: Switch could be after `if(power > 0)` to avoid unneeded var setting unless I'm misreading.

code/modules/supermatter/supermatter.dm
```diff
@@ -24,6 +24,34 @@
 
 #define WARNING_DELAY 20			//seconds between warnings.
 
+//This is really gross and I hate doing it like this.
```
MistakeNot4892, line 27: I am still bewildered by why you think an orderly list of sounds is gross.
It's fine.
This comment is gross though.

code/modules/supermatter/supermatter.dm
```diff
@@ -574,6 +607,30 @@ var/list/supermatter_final_thoughts = list(
...
+		if(!sound_token)
+			sound_token = GLOB.sound_player.PlayLoopingSound(src, sound_id, working_sound, volume = volume)
+		sound_token.SetVolume(volume)
+		addtimer(CALLBACK(src, .proc/play_accent_sound), 6 SECONDS)
+	else if(sound_token)
+		QDEL_NULL(sound_token)
+
+/obj/machinery/power/supermatter/proc/play_accent_sound() //This is a proc purely so i can call it via timer.
+	if(damage > warning_point)
+		playsound(src, pick(accent_sounds_delam), 50, 1)
```
MistakeNot4892, line 630: Style is `global.foo` for global lists.

code/modules/supermatter/supermatter.dm
```diff
@@ -574,6 +607,30 @@ var/list/supermatter_final_thoughts = list(
...
+		sound_id = "[type]_[sequential_id(/obj/machinery/power/supermatter)]"
+
+	switch(damage)
+		if(0 to warning_point)
+			working_sound = 'sound/machines/supermatter/sm_calm.ogg'
+		if(warning_point to INFINITY)
+			working_sound = 'sound/machines/supermatter/sm_delam.ogg'
+	if(power > 0)
+		var/volume = 10 + max(100, (15*power))
+		if(!sound_token)
```
MistakeNot4892, line 621: Bit confused here. You only seem to null out the sound token if `power` is <= 0, rather than if it's <= `warning_point`. Doesn't that mean it will never progress to the delamination background sound unless it goes from 0 to delaminating between Process() ticks?

Mazianni:
parking this for now

PR #1534, opened 2021-04-12
Reworking material stacks and material spawning.

MistakeNot4892:
In an effort to make material sheet behavior less type-dependent I've removed all non-generic material sheets and swapped out direct spawning of sheets to use `material.place_sheet()`. I've also moved sheet single/plural names onto the sheets themselves and replaced stacktype on the material with a `default_solid_form` var.

:cl:
tweak: Some material stack items and messages around them may have changed due to a backend refactor of materials.
/:cl:

PR #1535, opened 2021-04-13
Allows /obj/item and /obj/structure to melt/vaporize/burn up in high temperatures.

MistakeNot4892:
Splitting this out of #1534 since it's more or less standalone.
- Removes ATOM_FLAG_NO_TEMP_CHANGE as temperature stuff is tickchecked and doesn't seem to blip either init times or MC overhead, even during map-spanning fires.
- Adds some more granular value checking to heat damage on materials for walls/grilles.
- Changes items and structures to check their material value heat tolerances and change state accordingly when temperature changes.
- Gives reagents their own temperature var and temperature setting proc.

:cl:
experimental: Items and structures can now melt at hot temperatures.
/:cl:


MistakeNot4892:
Putting this on the backburner for now.

PR #1537, opened 2021-04-14
Adds fiberglass and copper to general use.

MistakeNot4892:
:cl:
tweak: Many objects previously made of glass are now made of fiberglass, which is stronger but less heat resistant.
tweak: Cables are now made of copper rather than steel.
tweak: Slag can now be processed via grinding and acid to recover some useful trace elements.
rscadd: Fiberglass sheets can be made from glass and plastic at an autolathe or textiles fab, or by grinding glass and plastic then heating over 100C.
/:cl:


MistakeNot4892:
Depends on #1538

afterthought2:
Fine by me apart from conflicts. Is there a reason the material is not a subtype of `/glass` like borosilicate, or do we avoid that kind of inheritance for materials?

afterthought2 (approved):

PR #1540, opened 2021-04-15
Refactoring matter into a datum.

MistakeNot4892:
DNM, opening for diff.


MistakeNot4892:
Putting this on the backburner for now.

PR #1543, opened 2021-04-16
Standardize health

SierraKomodo:
First pass on standardizing health and damage processing. Currently, the health changes are only applied to cult rune walls and to blob. These shouldn't massively affect a round if something breaks, but should be present and available enough to allow for general in-round testing. Rune walls use the simple health system, while blob is using the extension to allow for damage resistance calculation.

Also replaces all the damage defines with DAMAGE_* and adds some additional damage types. New types are currently unused, pending implementation in future updates to bring everything in to the updated system.

Further tweaks and optimizations to come based on review and feedback from live play.

## Changelog
:cl: SierraKomodo
refactor: First pass on health and damage standardization has been completed. Cult rune walls and blobs now use the new health system. There shouldn't be any player-facing changes. Please report any and all bugs with health and damage to the Nebula GitHub and tag `@SierraKomodo`.
/:cl:


code/datums/extensions/health/_defines.dm
```diff
@@ -0,0 +1,51 @@
+// Flags for health adjustment when changing max health
+/// No adjustment to `current_health`, except to not exceed the new `max_health`
+var/const/MAXHEALTH_ADJUST_NONE   = "none"
+/// Apply the exact same difference to `current_health` as was applied to `max_health`
+var/const/MAXHEALTH_ADJUST_DIRECT = "direct"
+/// Adjusts `current_health` by the same percentage difference as was applied to `max_health`
+var/const/MAXHEALTH_ADJUST_MULT   = "mult"
+/// Adjusts `current_health` to the new `max_health`
+var/const/MAXHEALTH_ADJUST_MAX    = "max"
```
SierraKomodo: Self-reminder to purge these, they're not actually being used.

code/datums/extensions/health/atom.dm
```diff
@@ -0,0 +1,411 @@
...
+	var/use_health_handler = FALSE
+
+	/**
+	 * Initialization config for the health handler.
+	 *
+	 * `max_health` - Default maximum health.
+	 *
+	 * `resist_*` - Resistance multiplier for various damage types, where `*` is replaced by the damage type's string value.
+	 */
+	var/list/health_handler_config = list(
```
Lohikar: No `= list()` vars on root types.

Lohikar (changes requested):

code/datums/extensions/health/atom.dm
```diff
@@ -0,0 +1,411 @@
...
+	 */
+	var/simple_health
+
+	/**
+	 * Maximum health for simple health processing
+	 */
+	var/simple_max_health
+
+
+/atom/Initialize(mapload, ...)
```
Lohikar: Don't override Initialize on `/atom` itself, this is the level where that proc is defined.

code/datums/extensions/health/atom.dm
SierraKomodo: Replaced with an overridable proc that returns a list.

code/datums/extensions/health/atom.dm
SierraKomodo: Moved.

SierraKomodo:
Going to tag this ready for review. The only changes that will be made to this branch from here on will be any review changes requested and fixing of any bugs that might be caught during review. I also updated the original post to reflect the current state of the PR and the intent.

code/datums/extensions/armor/ablative.dm
```diff
@@ -11,7 +11,7 @@
 		armor_degradation_coef = _armor_degradation_speed
 
 /datum/extension/armor/ablative/on_blocking(damage, damage_type, damage_flags, armor_pen, blocked)
-	if(!(damage_type == BRUTE || damage_type == BURN))
+	if(!(damage_type in list(DAMAGE_BRUTE, DAMAGE_BURN)))
```
MistakeNot4892: Why change this to create a list every time it blocks?

MistakeNot4892 (changes requested):
Hitting 'go' on the single comment I made like two weeks ago, will add a proper review following.

code/datums/extensions/health/_defines.dm
```diff
@@ -0,0 +1,41 @@
...
+var/global/const/DAMAGE_TOXIN     = "toxin"
+/// Genetic damage
+var/global/const/DAMAGE_GENETIC   = "genetic"
+/// Oxyloss damage
+var/global/const/DAMAGE_OXY       = "oxy"
+/// Brain damage
+var/global/const/DAMAGE_BRAIN     = "brain"
+
+/// All damage flags
+#define DAMAGE_ALL list(DAMAGE_BRUTE, DAMAGE_BURN, DAMAGE_STUN, DAMAGE_SHOCK, DAMAGE_EMP, DAMAGE_EXPLODE, DAMAGE_FIRE, DAMAGE_RADIATION, DAMAGE_BIO, DAMAGE_PAIN, DAMAGE_TOXIN, DAMAGE_GENETIC, DAMAGE_OXY, DAMAGE_BRAIN)
```
MistakeNot4892: Why make the above consts but then keep this as a define? This is more expensive than all of the above; every time you use the define you're creating a large list.

code/datums/extensions/health/_health.dm
```diff
@@ -0,0 +1,175 @@
...
+
+
+/**
+ * Retrieves the extension's `max_health`.
+ * Provided as a proc for override use.
+ */
+/datum/extension/health/proc/max_health(real = FALSE)
+	var/atom/A = holder
+	return real ? max_health : A.get_max_health()
+
```
MistakeNot4892, line 34: This file has a ton of unnecessary whitespace. Not necessarily a blocker but it looks untidy.

code/datums/extensions/health/atom.dm
```diff
@@ -0,0 +1,397 @@
...
+	var/simple_health
+
+	/**
+	 * Maximum health for simple health processing
+	 */
+	var/simple_max_health
+
+
+/// Returns a list containing the atom's requested health_handler configuration.
+/atom/proc/health_handler_config()
```
MistakeNot4892: Could probably use a more informative name, ex. `get_initial_health_handler_config()`

MistakeNot4892: Not sure if I like it returning a new list instance but if I'm understanding your code correctly it's only called during Init, so not really an issue.

code/datums/extensions/health/_health.dm
```diff
@@ -0,0 +1,175 @@
...
+	LAZYREMOVE(damage_resistances, damage_type)
+
+
+/**
+ * Retrieves the damage resistance multiplier for the given damage type.
+ */
+/datum/extension/health/proc/get_resistance(damage_type)
+	SHOULD_CALL_PARENT(TRUE)
+	var/resistance_value = LAZYACCESS(damage_resistances, damage_type)
+	return resistance_value == null ? 1 : resistance_value
```
MistakeNot4892: `isnull(resistance_value)`

code/datums/extensions/health/_health.dm
```diff
@@ -0,0 +1,175 @@
...
+		return set_death_state(TRUE)
+	else
+		return set_death_state(FALSE)
+
+
+/**
+ * Modifies `current_health` by the given modifier.
+ * Returns `TRUE` if `death_state` was changed, `FALSE` otherwise.
+ */
+/datum/extension/health/proc/mod_health(health_mod, damage_type = null)
```
MistakeNot4892: `mod` has a pretty specific meaning in mathematics (modulo) and another meaning in Neb's context (modpack) - `adjust_health` would line up with the other damage procs and avoid ambiguity.

code/datums/extensions/health/_health.dm
```diff
@@ -0,0 +1,175 @@
...
+	A.handle_death_change(new_death_state)
+	return TRUE
+
+
+/**
+ * Checks the current condition of health and sets the death state accordingly.
+ * Returns `TRUE` if `death_state` was changed, `FALSE` otherwise.
+ */
+/datum/extension/health/proc/update_death_state()
+	if (current_health() == 0)
```
MistakeNot4892: `<=` might be safer here.
Also `return set_death_state(current_health() <= 0)`

code/datums/extensions/health/atom.dm
```diff
@@ -0,0 +1,397 @@
...
+		if ("resist_[damage_type]" in health_handler_config)
+			set_damage_resistance(damage_type, health_handler_config["resist_[damage_type]"])
+
+	on_initialize_health(health_handler_config)
+
+
+/**
+ * To allow for additional init processing.
+ */
+/atom/proc/on_initialize_health(list/health_handler_config)
```
MistakeNot4892: I feel like you could just make `initialize_health()` return true/false and let people override it for additional logic.

code/datums/extensions/health/atom.dm
```diff
@@ -0,0 +1,397 @@
...
+
+/**
+ * Whether or not the atom has initialized health.
+ * Can be overridden for additional checks.
+ * Returns `TRUE` or `FALSE`.
+ */
+/atom/proc/has_health()
+	SHOULD_CALL_PARENT(TRUE)
+	if (using_simple_health())
+		if (!simple_max_health)
```
MistakeNot4892: This is redundant when you're returning it the line under.

code/datums/extensions/health/atom.dm
```diff
@@ -0,0 +1,397 @@
...
+
+/**
+ * Checks if the atom has received damage.
+ * Essentially an alias of `get_damage_value()` that converts the value to a boolean.
+ * If using the `damage_sources` extension subtype, setting `damage_type` will only show the damage of that given type.
+ */
+/atom/proc/health_damaged(damage_type = null)
+	if (!has_health())
+		return
+	return !!get_damage_value(damage_type)
```
MistakeNot4892: `return has_health() && get_damage_value(damage_type)`

code/modules/xenoarcheaology/artifacts/triggers/energy.dm
```diff
@@ -10,7 +10,7 @@
 	. = ..()
 	if(istype(O, /obj/item/projectile))
 		var/obj/item/projectile/P = O
-		. = (P.damage_type == BURN) || (P.damage_type == ELECTROCUTE)
+		. = (P.damage_type in list(DAMAGE_BURN, DAMAGE_SHOCK))
```
MistakeNot4892: See above regarding replacing checks with creating a list.

code/modules/tables/interactions.dm
```diff
@@ -43,7 +43,7 @@
 	return 1
 
 /obj/structure/table/bullet_act(obj/item/projectile/P)
-	if(!(P.damage_type == BRUTE || P.damage_type == BURN))
+	if(!(P.damage_type in list(DAMAGE_BRUTE, DAMAGE_BURN)))
```
MistakeNot4892: Replacing straight checks with list creation.

code/modules/shieldgen/emergency_shield.dm
```diff
@@ -47,7 +47,7 @@
 
 	//Calculate damage
 	var/aforce = W.force
-	if(W.damtype == BRUTE || W.damtype == BURN)
+	if(W.damtype in list(DAMAGE_BRUTE, DAMAGE_BURN))
```
MistakeNot4892: Replacing straight checks with list creation.

code/modules/projectiles/projectile.dm
```diff
@@ -141,12 +141,12 @@
...
 /obj/item/projectile/can_embed()
 	//embed must be enabled and damage type must be brute
-	if(!embed || damage_type != BRUTE)
+	if(!embed || damage_type != DAMAGE_BRUTE)
 		return FALSE
 	return TRUE
 
 /obj/item/projectile/proc/get_structure_damage()
-	if(damage_type == BRUTE || damage_type == BURN)
+	if(damage_type in list(DAMAGE_BRUTE, DAMAGE_BURN))
```
MistakeNot4892: Replacing straight checks with list creation.

code/modules/mob/living/carbon/human/human_defense.dm
```diff
@@ -389,7 +389,7 @@ meteor_act
 /mob/living/carbon/human/proc/handle_suit_punctures(var/damtype, var/damage, var/def_zone)
 
 	// Tox and oxy don't matter to suits.
-	if(damtype != BURN && damtype != BRUTE) return
+	if(!(damtype in list(DAMAGE_BRUTE, DAMAGE_BURN))) return
```
MistakeNot4892: Replacing straight checks with list creation.

code/modules/mob/living/carbon/human/human_defense.dm
```diff
@@ -191,7 +191,7 @@ meteor_act
 		forcesay(global.hit_appends)	//forcesay checks stat already
 		radio_interrupt_cooldown = world.time + (RADIO_INTERRUPT_DEFAULT * 0.8) //getting beat on can briefly prevent radio use
 
-	if((I.damtype == BRUTE || I.damtype == PAIN) && prob(25 + (effective_force * 2)))
+	if((I.damtype in list(DAMAGE_BRUTE, DAMAGE_PAIN)) && prob(25 + (effective_force * 2)))
```
MistakeNot4892: Replacing straight checks with list creation.

code/modules/mob/living/carbon/human/human_damage.dm
```diff
@@ -375,7 +375,7 @@ This function restores all organs.
 			organ = get_organ(check_zone(def_zone, src))
 
 	//Handle other types of damage
-	if(!(damagetype in list(BRUTE, BURN, PAIN, CLONE)))
+	if(!(damagetype in list(DAMAGE_BRUTE, DAMAGE_BURN, DAMAGE_PAIN, DAMAGE_GENETIC)))
```
MistakeNot4892: Making this list a global would be good.

code/modules/clothing/masks/smokable.dm
```diff
@@ -145,7 +145,6 @@
 
 /obj/item/clothing/mask/smokable/proc/extinguish(var/mob/user, var/no_message)
 	lit = 0
-	damtype = "brute"
```
MistakeNot4892: This should be set back to brute when the cig goes out.

code/game/objects/structures/window.dm
```diff
@@ -247,7 +247,7 @@
 			set_anchored(0)
 	else if (!istype(W, /obj/item/paint_sprayer))
 		user.setClickCooldown(DEFAULT_ATTACK_COOLDOWN)
-		if(W.damtype == BRUTE || W.damtype == BURN)
+		if(W.damtype in list(DAMAGE_BRUTE, DAMAGE_BURN))
```
MistakeNot4892: List etc.

code/game/objects/structures/railing.dm
```diff
@@ -274,7 +274,7 @@
 			update_icon()
 		return
 
-	if(W.force && (W.damtype == "fire" || W.damtype == "brute"))
+	if(W.force && (W.damtype in list(DAMAGE_BRUTE, DAMAGE_BURN)))
```
MistakeNot4892: List etc.

code/game/objects/structures/inflatable.dm
```diff
@@ -133,7 +133,7 @@
 
 /obj/structure/inflatable/attackby(obj/item/W, mob/user)
 
-	if((W.damtype == BRUTE || W.damtype == BURN) && (W.can_puncture() || W.force > 10))
+	if((W.damtype in list(DAMAGE_BRUTE, DAMAGE_BURN)) && (W.can_puncture() || W.force > 10))
```
MistakeNot4892: List etc.

code/game/objects/items/weapons/special_attacks/backstab.dm
```diff
@@ -28,10 +28,10 @@ location_check: bool. allows facestabs if set to false, skipping the check for b
 Proc returns a boolean if successful.
 */
 
-/obj/item/proc/backstab(var/mob/living/target, mob/user, var/damage = 30, var/damage_type = BRUTE, var/damage_flags, var/target_zone = BP_CHEST, var/location_check = TRUE)
+/obj/item/proc/backstab(var/mob/living/target, mob/user, var/damage = 30, var/damage_type = DAMAGE_BRUTE, var/damage_flags, var/target_zone = BP_CHEST, var/location_check = TRUE)
 
 	//Runtime prevention.
-	if( !( damage_type in list( BRUTE, BURN, TOX, OXY, CLONE, PAIN ) ) ) //End the proc with a false return if we're not doing a valid damage type.
+	if( !( damage_type in list(DAMAGE_BRUTE, DAMAGE_BURN, DAMAGE_TOXIN, DAMAGE_OXY, DAMAGE_GENETIC, DAMAGE_PAIN) ) ) //End the proc with a false return if we're not doing a valid damage type.
```
MistakeNot4892: As below re: moving this to a global.

code/game/machinery/camera/camera.dm
```diff
@@ -211,7 +211,7 @@
 			else to_chat(O, "<b><a href='byond://?src=\ref[O];track2=\ref[O];track=\ref[U];trackname=[U.name]'>[U]</a></b> holds \a [itemname] up to one of your cameras ...")
 			show_browser(O, text("<HTML><HEAD><TITLE>[]</TITLE></HEAD><BODY><TT>[]</TT></BODY></HTML>", itemname, info), text("window=[]", itemname))
 
-	else if(W.damtype == BRUTE || W.damtype == BURN) //bashing cameras
+	else if(W.damtype in list(DAMAGE_BRUTE, DAMAGE_BURN)) //bashing cameras
```
MistakeNot4892: List etc.

code/game/machinery/atmoalter/canister.dm
```diff
@@ -253,7 +253,7 @@ update_flag
 	return 0
 
 /obj/machinery/portable_atmospherics/canister/bullet_act(var/obj/item/projectile/Proj)
-	if(!(Proj.damage_type == BRUTE || Proj.damage_type == BURN))
+	if(!(Proj.damage_type in list(DAMAGE_BRUTE, DAMAGE_BURN)))
```
MistakeNot4892: List etc.

code/game/machinery/_machines_base/stock_parts/shielding.dm
```diff
@@ -9,21 +9,21 @@
...
-	protection_types = list(BRUTE)
+	protection_types = list(DAMAGE_BRUTE)
 	material = /decl/material/solid/metal/steel
 
 /obj/item/stock_parts/shielding/heat
 	name = "heatsink"
 	icon_state = "heatsink"
 	desc = "Active cooling system protecting machinery against the high temperatures."
-	protection_types = list(BURN)
+	protection_types = list(DAMAGE_BURN)
```
MistakeNot4892: Looking at the code that uses this var, this doesn't actually need to be a list.
More of a nitpick than an actual change request.

code/game/machinery/_machines_base/machinery_damage.dm
```diff
@@ -1,14 +1,14 @@
-/obj/machinery/proc/take_damage(amount, damtype = BRUTE, silent = FALSE)
+/obj/machinery/proc/take_damage(amount, damtype = DAMAGE_BRUTE, silent = FALSE)
 	//Let's not bother initializing all the components for nothing
 	if(amount <= 0)
 		return
-	if(damtype != BRUTE && damtype != BURN && damtype != ELECTROCUTE)
+	if(!(damtype in list(DAMAGE_BRUTE, DAMAGE_BURN, DAMAGE_SHOCK)))
```
MistakeNot4892: List etc.

code/datums/extensions/health/atom.dm
```diff
@@ -0,0 +1,397 @@
...
+	if (!has_health())
+		return
+	if (!is_alive())
+		to_chat(user, SPAN_DANGER("It looks broken."))
+		return
+
+	var/health_percentage = get_damage_percentage()
+	switch (health_percentage)
+		if (1)
+			to_chat(user, SPAN_NOTICE("It looks fully intact."))
```
MistakeNot4892, line 268:
````
var/decl/pronouns/G = get_pronouns()
...
to_chat(user, SPAN_NOTICE("[G.He] looks fully intact."))
````
Might be worth adding a `s` to the pronouns decl so you could do `[G.He] look[G.s]`

File: code/datums/extensions/health/_health.dm [hunk 31]
```diff
@@ -0,0 +1,175 @@
...
+	if (health_mod == 0)
+		return FALSE // No health change, no further processing needed
+	current_health = Clamp(current_health + health_mod, 0, max_health())
+
+	var/atom/A = holder
+	if (health_mod > 0)
+		A.post_restore_health(health_mod, damage_type)
+	else if (health_mod < 0)
+		A.post_damage_health(abs(health_mod), damage_type)
+	A.post_health_change(health_mod, damage_type)
```
MistakeNot4892, line 86: This block highlights a concern I have about the current implementation, in that it's adding tons of proc calls for stuff that isn't actually implemented. You could just override this proc and handle post-health logic there, or on the extension, rather than making every atom call at least two additional procs for no reason.

MistakeNot4892:
Alright largely I like what you're doing, but I have some concerns about the implementation. If you split the define renaming off into another PR it might make this one easier to review, also.

Some thoughts in no particular order:
- As mentioned, I feel like this is adding lengthy chains of proc calls to what was previously simple mathematical operations. Streamlining the basic implementation as much as possible would be desirable considering you're implementing a lot of this on `/atom`.
- There's a lot of boilerplate in the `/atom` helpers. Not sure how much of it could be shunted to the extension but it feels unwieldly.
- Rather than future proofing by providing hooks to override when nothing overrides them, I would suggest returning true/false where possible and letting people handle it themselves when they need to.
- There are a few procs like `copy_health` that don't appear to be used, but are pretty chunky. Is this future proofing or for debugging?

code/datums/extensions/health/_health.dm
```diff
@@ -0,0 +1,175 @@
...
+		return FALSE
+	max_health = max(round(new_max_health, 1), 1)
+	current_health = max_health()
+	return update_death_state()
+
+
+/**
+ * Clones health data from another health extension.
+ */
+/datum/extension/health/proc/clone(datum/extension/health/source)
```
Lohikar: This function works differently from what you'd expect a `clone()` function to -- `copy_from()` would be a better name.

code/datums/extensions/health/atom.dm
```diff
@@ -0,0 +1,397 @@
...
+
+	/**
+	 * Current health for simple health processing
+	 */
+	var/simple_health
+
+	/**
+	 * Maximum health for simple health processing
+	 */
+	var/simple_max_health
```
Lohikar, line 9: Please don't use multiline comments for vars.

code/datums/extensions/health/atom.dm
```diff
@@ -0,0 +1,397 @@
...
+		simple_max_health = null
+		return
+	remove_extension(src, use_health_handler)
+	health_handler = null
+
+
+/**
+ * Whether or not the health handler is set to simple health. Used to bypass extension calls.
+ */
+/atom/proc/using_simple_health()
```
Lohikar: Does this need to be a proc?

code/datums/extensions/health/damage_sources.dm
```diff
@@ -0,0 +1,51 @@
...
+ * Variant of the health extension that tracks individual damage types.
+ * Separated due to list usage.
+ */
+/datum/extension/health/damage_sources
+	/// List of damage divided into each damage type.
+	var/list/damage_sources = list()
+
+
+/datum/extension/health/damage_sources/New(datum/holder)
+	. = ..()
```
Lohikar: Do not return a value in New().

code/datums/extensions/health/damage_sources.dm
```diff
@@ -0,0 +1,51 @@
...
+
+/datum/extension/health/damage_sources/revive()
+	. = ..()
+	for (var/damage_type in damage_sources)
+		damage_sources[damage_type] = 0
+
+
+/datum/extension/health/damage_sources/set_max_health(new_max_health)
+	. = ..()
+	return recalculate_health()
```
Lohikar, line 28: Why are you setting `.` to parent if you're explicitly returning something else?

SierraKomodo:
Removed the mass-replacement of damage type defines due to constant conflict resolution shenanigans. I'll do that in a separate PR.

code/datums/extensions/health/_defines.dm
SierraKomodo: I wasn't able to get constant lists to work compile. I don't recall what the exact error was - Irrelevant now either way since I've removed the damage constants for now.

code/datums/extensions/health/_health.dm
SierraKomodo: Old habits of having two blank lines between functions/classes/etc. I can reduce it down to 1 line break per.

code/datums/extensions/health/_defines.dm
SierraKomodo: Actually still relevant as I do still need this list at least for the resistance stuff to work. I'll poke on discord about the constant error.

code/datums/extensions/health/atom.dm
SierraKomodo: There was some discussion we had, I've forgotten the details at this point, regarding the best method for an initialization configuration list that can be overridden by children while also being able to use and modify the parent's version (I.e. if an atom's child only needs to modify one of the entries and keep the rest identical) - This was decided as a good option for that use case.

code/datums/extensions/health/_health.dm
SierraKomodo: `post_health_change()` was added because an atom may want to do additional things after the extensions modify its health, without needing to create additional subtypes of the health extension.

The other two `post_*_health()` procs were mostly just a convenience thing but I can remove those.

code/datums/extensions/health/_health.dm
SierraKomodo: I renamed it to `copy_health()` to match the atom proc that uses it.

code/datums/extensions/health/atom.dm
SierraKomodo: I'll add this in a separate PR once I'm done going over the reviews on this one.

code/datums/extensions/health/atom.dm
SierraKomodo: In hindsight probably not.

code/datums/extensions/health/damage_sources.dm
SierraKomodo: VSC's autocomplete does that and I tend to forget to update it.

code/datums/extensions/health/_health.dm
MistakeNot4892: Additional subtypes of the health extension would be preferable to future-proofing that causes proc overhead, imo.

code/datums/extensions/health/_defines.dm
MistakeNot4892: Pinged you on Discord, but `var/global/list/all_damage_types` would do the same job without complaining about lists being nonconstant.

MistakeNot4892:
I don't think this is very likely to be merged due to the overhead involved in atom-level extensions for something as fundamental as health. I do think extensions for complex health is feasible at a higher level in the object hierarchy but as it stands I can't see this PR being workable. :-1:

MistakeNot4892:
Closing as discussed on Discord.

PR #1544, opened 2021-04-17
Added ice forms for the common gases used in engineering/atmos

PsyCommando:
So in order to make it possible to obtain everything "in the wild" on outreach13 I added ice form for some of the more commonly used gas types. And I was told to push this upstream instead. So here you go!

Hydrogen, nitrogen, oxygen, carbon dioxide, methane, ammonia, argon, neon, xenon, chlorine, and sulfur dioxide all have a mineable ice material type declared now.

Helium and tritium are fusion products, so I didn't add ices for those. I also didn't add ices for some of the compounds used in chemistry and medical since I'm not sure if we want those as ices? I can add them if needed however.

Also, I didn't make the ices turn to liquid first since there's no plumbing mechanic or easier ways to collect liquid messes. Melting temps might need to be increased by a lot as well, since its just the real-world values.

Also, I'm waiting on feedback first before adding those to the actual mineral spawn lists, and tweaking occurrence rates and ore values.

:cl:
rscadd: added ices for most of the base gases
\:cl:


MistakeNot4892:
You should probably do this just by adding the sparse/rich ore values to the gas datums directly, since they're not actually distinct materials so much as different states of the same material (ice is only separate because it's used in mixed drinks and heat capacity/phase transitions aren't a thing yet). Would also suggest adding this to `_mat_gas.dm` to avoid having to set the name a bunch:
````
/decl/material/gas/New()
	if(!liquid_name)
		liquid_name = "liquid [name]"
	if(!solid_name)
		solid_name = "[name] ice" // or "frozen [name]"
	..()
````

**Comment by PsyCommando** [comment]
So, why are the materials divided by phases and have duplicates in other phases? I gotta admit I was kinda confused coming from older baycode.

Also, the only reason I added the defines in the ices file is because there seemed to be some mineable ices materials already in there. Like `/decl/material/solid/ice/rubenium` which is actually referred in the ore noise generator:

    /datum/random_map/noise/ore/ice
	  surface_metals = list(
		  /decl/material/solid/ice/aspium =              list(RESOURCE_MID_MIN,  RESOURCE_MID_MAX),
		  /decl/material/solid/ice/lukrite =             list(RESOURCE_MID_MIN,  RESOURCE_MID_MAX),
		  /decl/material/solid/ice/rubenium =	           list(RESOURCE_MID_MIN,  RESOURCE_MID_MAX),
		  /decl/material/solid/ice/trigarite =           list(RESOURCE_MID_MIN,  RESOURCE_MID_MAX),
		  /decl/material/solid/ice/ediroite =	           list(RESOURCE_MID_MIN,  RESOURCE_MID_MAX),
		  /decl/material/solid/metal/iron =              list(RESOURCE_MID_MIN,  RESOURCE_MID_MAX),
		  /decl/material/solid/metal/aluminium =         list(RESOURCE_LOW_MIN,  RESOURCE_LOW_MIN),
		  /decl/material/solid/metal/gold =              list(RESOURCE_LOW_MIN,  RESOURCE_LOW_MAX),
		  /decl/material/solid/metal/silver =            list(RESOURCE_LOW_MIN,  RESOURCE_LOW_MAX),
		  /decl/material/solid/metal/uranium =           list(RESOURCE_LOW_MIN,  RESOURCE_LOW_MAX)
	  )

Its a bit hard to tell between mineable ores which may contain more than one material VS pure material phases imo. But, then again, I'm not even sure I understand how its meant to work together.

But, anyways, I'll try to implement the changes.

**Comment by MistakeNot4892** [comment]
As far as I am aware the only material that has two definitions because of phases is `liquid/water` and `solid/ice`. The distinction between the existing ices and your new ices is that the existing ones are compounds of several substances (water, ammonia and mercury etc) rather than just pure single substances like these new ones. If your ices were compounds I wouldn't be requesting the change as they would need to be new material defs.

**Comment by MistakeNot4892** [comment]
To be clear, the `gas` `solid` and `liquid` types only define some inherited values, they don't dictate the actual phase of the material. Generally they are just shorthand for 'this material is X phase at room temperature'.

**Comment by PsyCommando** [comment]
Well, technically those ice defines could have more than one substance. I just kept it a only a single material for now, since there's the whole question of balance behind it, and I don't know if those ices will need some toning down. That's one of the dilemma I'm facing. 
Those ice defines I made are really supposed to be exclusively for mineable ores, and probably should be separate from the actual ice form of the gas. Just so they can be tweaked without being mistakenly used in the future as the pure ice phase of each gas. 

Now I kinda wonder if ores items themselves shouldn't actually contains the details on one or more materials they contain instead of only having a single materials defines per ore?

**Comment by MistakeNot4892** [comment]
Well, the idea of having the lists nested on the materials is that you can have a production chain of processes to refine a given material, possibly several times, with different properties for each form. Having it specific to the ore item would mean you couldn't (for example) make something out of a composite material like asperite ice by compressing it.

A longer discussion on the topic might be better on Discord. For the purposes of this PR, I would be happy to approve either your new ices being composite materials like the existing ices, or the ore info being put onto the gas datums directly.

**Comment by PsyCommando** [comment]
Sorry for the late reply, but alright. I'll get on the nebula discord later then.
And I'll make the ices composites for now as well!

**Review by MistakeNot4892 — DISMISSED** [review]
The macro seems like overkill but there's nothing wrong with the code that I can see.

**Comment by comma** [comment]
```
code/modules/materials/definitions/solids/materials_solid_ice.dm, line 140, column 1:
error: undefined var 'heating_products' on type ''
```

PsyCommando:
Well, the nice thing about the macro is that it avoids a ton of repetition for nothing, or creating new useless vars to init hydrates with minimal repetition, and it has no runtime footprint. Besides, since I don't know how things will change when phases of matter are implemented that might make maintaining those easier. Or maybe not XD
Also fixed the dumb warning spaceman dm is throwing by shoving everything in the constructor inside the macro.

PsyCommando:
Just to double check, are those ore occurrence values ok? And I made fuels more expensive than the others, by 0.1, but I'm not sure if that's relevant/enough? Also, do you need me to add those to the ore spawn list?

comma (approved):

PR #1549, opened 2021-04-19
Expanded the map marker landmark to allow for non-template maploading.

MistakeNot4892:
- Map marker templates now have a subtype that loads during init instead of after template placement.
- Submap archetypes actually use the decl repository now.

### DIFF
--- a/code/controllers/subsystems/mapping.dm
+++ b/code/controllers/subsystems/mapping.dm
@@ -3,18 +3,21 @@ SUBSYSTEM_DEF(mapping)
 	init_order = SS_INIT_MAPPING
 	flags = SS_NO_FIRE
 
-	var/list/map_templates = list()
-	var/list/space_ruins_templates = list()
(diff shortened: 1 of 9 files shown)


code/controllers/subsystems/mapping.dm
```diff
@@ -3,18 +3,21 @@ SUBSYSTEM_DEF(mapping)
...
+	var/list/compile_time_map_markers =  list()
 
 /datum/controller/subsystem/mapping/Initialize(timeofday)
+
 	// Load templates and build away sites.
+	for(var/obj/effect/landmark/map_load_mark/marker as anything in compile_time_map_markers)
+		compile_time_map_markers -= marker
+		marker.load_template()
+
 	preloadTemplates()
```
afterthought2, line 20: I'm not sure this is the best order to do these two things in, but OK.

afterthought2 (approved):

PR #1551, opened 2021-04-19
Optimizes known sector checking.

MistakeNot4892:


code/modules/overmap/sectors.dm
```diff
@@ -58,6 +61,13 @@
 	if(. == INITIALIZE_HINT_QDEL)
 		return
 
+	if(sector_flags & OVERMAP_SECTOR_KNOWN)
+		LAZYADD(global.known_overmap_sectors, src)
+		layer = ABOVE_LIGHTING_LAYER
```
comma, line 66: these are probably meaningless since sensor stuff isn't using lighting anymore, layer and plane i mean

maps/away/liberia/liberia.dm
```diff
@@ -25,7 +25,7 @@
 	vessel_mass = 3000
 	fore_dir = WEST
 	max_speed = 1/(1 SECOND)
-	known = 1
+	sector_flags = OVERMAP_SECTOR_KNOWN
```
comma, line 28: pretty useless to set mobile ship ot known since it'll only gives its spawn position to helm consoles

MistakeNot4892: I think that's the idea, so they can head towards it? Not too sure of the logic, just wanted to make it 1:1.

comma (approved):

PR #1552, opened 2021-04-19
Radius of supermatter light depends on number of shots fired at it

quardbreak:


code/modules/supermatter/supermatter.dm
```diff
@@ -87,6 +89,7 @@ var/list/supermatter_final_thoughts = list(
 	density = 1
 	anchored = 0
 	light_range = 4
+	light_power = 1
```
Lohikar: This is the default value for this var.

MistakeNot4892 (approved):

PR #1556, opened 2021-04-20
medbot stuff from virgo

Mazianni:
Medbots can now be tipped over.
They will eventually panic and right themselves after a little while, and make a bit of noise in the process of panicking.

Also re-adds voice lines to them. Can be toggled by changing vocal to FALSE.


code/modules/mob/living/bot/medibot.dm
```diff
@@ -246,3 +324,74 @@
...
+		if(MEDBOT_PANIC_MED)
+			messagevoice = list("Please put me back." = 'sound/voice/medbot/please_put_me_back.ogg')
+		if(MEDBOT_PANIC_HIGH)
+			messagevoice = list("Please, I am scared!" = 'sound/voice/medbot/please_im_scared.ogg')
+		if(MEDBOT_PANIC_FUCK)
+			messagevoice = list("I don't like this, I need help!" = 'sound/voice/medbot/dont_like.ogg', "This hurts, my pain is real!" = 'sound/voice/medbot/pain_is_real.ogg')
+		if(MEDBOT_PANIC_ENDING)
+			messagevoice = list("Is this the end?" = 'sound/voice/medbot/is_this_the_end.ogg', "Nooo!" = 'sound/voice/medbot/nooo.ogg')
+		if(MEDBOT_PANIC_END)
+			GLOB.global_announcer.autosay("PSYCH ALERT: Crewmember [tipper_name] recorded displaying antisocial tendencies torturing bots in [get_area(src)]. Please schedule psych evaluation.", "[src]", "Medical")
```
Lohikar: I don't think this should be able to send an announcement, a medbot being tipped over is pretty minor.

Techhead0: I'd probably step it down from medical radios to just medihuds like its other broadcasts, but IMO it is a pretty funny way of reporting it

code/modules/mob/living/bot/medibot.dm
```diff
@@ -246,3 +324,74 @@
...
+	else
+		visible_message(SPAN_NOTICE("[src] manages to [pick("writhe", "wriggle", "wiggle")] enough to right itself."))
+		messagevoice = list("Fuck you." = 'sound/voice/medbot/fuck_you.ogg', "Your behavior has been reported, have a nice day." = 'sound/voice/medbot/reported.ogg')
+
+	tipper_name = null
+	if(world.time > last_tipping_action_voice + 15 SECONDS)
+		last_tipping_action_voice = world.time
+		var/message = pick(messagevoice)
+		say(message)
+		playsound(src, messagevoice[message], 70)
```
Techhead0: These messages (and other new ones) should also respect `voice`

Mazianni: Addressed.

code/modules/mob/living/bot/medibot.dm
Mazianni: Moved to medhuds.

code/modules/mob/living/bot/medibot.dm
```diff
@@ -23,19 +31,57 @@
...
+			to_chat(user, SPAN_WARNING("They are tipped over and appear visibly distressed.")) // now we humanize the medbot as a they, not an it
+		if(MEDBOT_PANIC_HIGH to MEDBOT_PANIC_FUCK)
+			to_chat(user, SPAN_WARNING("They are tipped over and visibly panicking!"))
+		if(MEDBOT_PANIC_FUCK to INFINITY)
+			to_chat(user, SPAN_DANGER("They are freaking out from being tipped over!"))
+
 /mob/living/bot/medbot/handleIdle()
 	if(vocal && prob(1))
-		var/message = pick("Radar, put a mask on!", "There's always a catch, and it's the best there is.", "I knew it, I should've been a plastic surgeon.", "What kind of infirmary is this? Everyone's dropping like dead flies.", "Delicious!")
+		var/message_options = list(
```
comma: should probably be /static/

comma: same with other message lists

Mazianni: some can't be statics because they have variables in them, but i've converted the ones that can be (i think)

MistakeNot4892: Probably clearer to move the static var declarations onto the type instead of doing it at local scope.

comma (approved):

PR #1559, opened 2021-04-21
Fixes kinetic harvester.

MistakeNot4892:
This is a leftover bug from the material/reagent rewrite stuff.

### DIFF
--- a/code/modules/power/fusion/fusion_circuits.dm
+++ b/code/modules/power/fusion/fusion_circuits.dm
@@ -8,12 +8,15 @@
 	build_path = /obj/machinery/kinetic_harvester
 	board_type = "machine"
 	origin_tech = "{'programming':4,'engineering':4,'materials':4}"
+	additional_spawn_components = list(
+		/obj/item/stock_parts/console_screen = 1,
+		/obj/item/stock_parts/keyboard = 1
+	)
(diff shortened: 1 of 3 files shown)


code/modules/power/fusion/fusion_circuits.dm
```diff
@@ -9,23 +9,25 @@
 	board_type = "machine"
 	origin_tech = "{'programming':4,'engineering':4,'materials':4}"
 	req_components = list(
+		/obj/item/stock_parts/keyboard = 1,
```
comma: IIRC keyobard/screen components should be in additional_spawn_components rather than in req ones, or entire machine will stop working if screen is broken

MistakeNot4892: Seems inconsistent in code, base circuitboards for example have them in `req_components`.

MistakeNot4892: `[10:07 AM] afterthought: they should be in req_components if you want the machine to not function without them, iirc`

MistakeNot4892: I'm ok with the machine breaking if they're broken since it has no other use.

comma (approved):

PR #1560, opened 2021-04-21
Rewrote theft objective to allow for non-type-dependant material sheet targets.

MistakeNot4892:
It seems like objectives don't actually check for success or failure anymore, not sure if that got removed on Neb or not but all this code is basically unused currently. Cleaned it up anyway for the material stack PR.

### DIFF
--- a/maps/~mapsystem/maps_antagonism.dm
+++ b/maps/~mapsystem/maps_antagonism.dm
@@ -22,13 +22,13 @@
 	)
 
 	var/list/potential_special_theft_targets = list(
-		"nuclear gun"             = /obj/item/gun/energy/gun/nuclear,
-		"diamond drill"           = /obj/item/pickaxe/diamonddrill,
-		"bag of holding"          = /obj/item/storage/backpack/holding,
-		"hyper-capacity cell"     = /obj/item/cell/hyper,
-		"10 diamonds"             = /obj/item/stack/material/diamond,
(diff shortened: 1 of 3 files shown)


maps/~mapsystem/maps_antagonism.dm
```diff
@@ -22,13 +22,13 @@
...
-		"bag of holding"          = /obj/item/storage/backpack/holding,
-		"hyper-capacity cell"     = /obj/item/cell/hyper,
-		"10 diamonds"             = /obj/item/stack/material/diamond,
-		"50 gold bars"            = /obj/item/stack/material/gold,
-		"25 refined uranium bars" = /obj/item/stack/material/uranium,
+		"nuclear gun"         = /obj/item/gun/energy/gun/nuclear,
+		"diamond drill"       = /obj/item/pickaxe/diamonddrill,
+		"bag of holding"      = /obj/item/storage/backpack/holding,
+		"hyper-capacity cell" = /obj/item/cell/hyper,
+		"10 diamonds"         = list(/obj/item/stack/material/*gemstone*/, 10, /decl/material/solid/gemstone/diamond),
```
comma, line 29: wtf with *s? in here?

MistakeNot4892: Inline comment because the paths don't exist yet. Makes my life easier when updating the sheets PR.

afterthought2 (approved):

PR #1561, opened 2021-04-21
Added ore lists to strata.

MistakeNot4892:
- Exoplanets now select a strata based on their ambient temperature. This logic can stand to be made more interesting in the future.
- Strata now contain an ore list and autogenerate weights based on the material decls in the list at compile time.


comma (dismissed):

code/controllers/subsystems/initialization/materials.dm
```diff
@@ -143,3 +145,27 @@ SUBSYSTEM_DEF(materials)
...
+		return
+	var/obj/effect/overmap/visitable/sector/exoplanet/planet = map_sectors["[location.z]"]
+	if(istype(planet))
+		return planet.get_strata(location)
+	var/s_key = "[location.z]"
+	if(!global.default_strata_type_by_z[s_key])
+		global.default_strata_type_by_z[s_key] = pick(subtypesof(/decl/strata))
+	return global.default_strata_type_by_z[s_key]
+
+var/warned_two
```
afterthought2: Debug code?

afterthought2 (changes requested):

afterthought2 (approved):

PR #1564, opened 2021-04-21
Electrical Storm Additions.

Mazianni:
Electrical storms have been somewhat tweaked.

-APCs already overloaded will no longer be overloaded repeatedly.
-APCs now spark when they are overloaded by an electrical storm, and play a suitably rumbly thunder sound at their loc.
-Anyone in the area of an APC that is overloaded by a storm has a small chance of getting their screen flashed like they were flashed (no stun)
-Some cosmetic text messages.
-Flying into electrical storms will now accelerate you *very* quickly, and has a not-significant chance of yeeting your ship off in a random direction.
-Minor screenshake and cosmetic message when a ship is knocked off course.
-Shields prevent APCs from being disabled (previous behavior), and shields will also prevent course disruption or velocity increase.

Inertial dampeners are advised to be set to the highest possible setting before entering a storm.


code/modules/events/electrical_storm.dm
```diff
@@ -42,7 +58,7 @@
 		var/obj/machinery/power/shield_generator/shield_gen = pick(shields)
 		//Minor breaches aren't enough to let through frying amounts of power
 		if(shield_gen.take_shield_damage(30 * severity, SHIELD_DAMTYPE_EM) <= SHIELD_BREACHED_MINOR)
-			return
+			shielded = TRUE
```
MistakeNot4892, line 82: Shouldn't it `break` here?

code/modules/events/electrical_storm.dm
```diff
@@ -51,27 +67,79 @@
 
 	for(var/obj/machinery/power/apc/T in picked_apcs)
```
MistakeNot4892: While you're in the area, this could be `T as anything in picked_apcs`.

code/modules/events/electrical_storm.dm
```diff
@@ -5,6 +5,21 @@
 	has_skybox_image = TRUE
 	var/list/valid_apcs		// Shuffled list of valid APCs.
 	var/global/lightning_color
+	var/last_bounce //track the last time we bounced around whatever overmap effect we're affecting.
+	var/bounce_delay = 10 SECONDS // this is divided by severity, so a intense storm means you're going *flying*.
+	var/static/lightning_sounds = list(
```
MistakeNot4892, line 10: Lohi got grumpy about me suggesting this and said not to move it. :(

MistakeNot4892:
It might be worth adding `!apc.is_critical` to the initial APC selection so the ion storm doesn't kill the engine bay or the bridge. That would go a long way towards making ion storms less horrible.

code/modules/events/electrical_storm.dm
Mazianni: Lack of break is intentional, just so rumbly storm sounds happen and such.

code/modules/events/electrical_storm.dm
Mazianni: This is the way I did it in the first place actually, his comments were on the medibot PR.

code/modules/events/electrical_storm.dm
MistakeNot4892: Oh, thanks Github, it looked like this was in the for loop so it was hitting every shield.

code/modules/events/electrical_storm.dm
MistakeNot4892: Oh derp.

MistakeNot4892 (approved):

PR #1568, opened 2021-04-22
Makes possible to add sound list to airlock's open/close vars

quardbreak:


afterthought2:
Is there a reason to do this if you are not adding any doors which use it?

quardbreak:
This is change from my downstream. I made a modpack which use list of sounds but using list in playsound without the `pick()` will make game crash.

MistakeNot4892 (approved):

PR #1569, opened 2021-04-22
Fixes from stable.

MistakeNot4892:


afterthought2 (approved):

PsiOmegaDelta:
Did there happen to be a merge conflict between stable/dev?
Pondering the reason for the merge commit to already be there

MistakeNot4892:
I pulled from dev into stable, I could rebase it instead.

PR #1570, opened 2021-04-23
Tool System

Mazianni:
-Tool Quality system, like Eris. Tools can now be used as various other tools.
-Greater than 1 is faster, less than is significantly slower.
-Tools now have a generic interaction proc that can replace the do_after/playsound combo used in many different places.
WIP as I comb through the code and insert the multiplier where appropriate. Agony.


Mazianni:
handing this over to loaf.

MistakeNot4892:
Going to close this in favour of #1587 if you don't mind. Lemme know if I should reopen.

PR #1571, opened 2021-04-24
Screenshake refactor

ghostsheet:
🆑
bugfix: Camera shake has been refactored.
/🆑

Not much here, just same old camera shake without sleeps and waitfor

### DIFF
--- a/code/modules/mob/mob_helpers.dm
+++ b/code/modules/mob/mob_helpers.dm
@@ -326,22 +326,17 @@ It's fairly easy to fix if dealing with single letters but not so much with comp
 #define PIXELS_PER_STRENGTH_VAL 16
 
 /proc/shake_camera(mob/M, duration, strength=1)
-	set waitfor = 0
-	if(!M || !M.client || M.shakecamera || M.stat || isEye(M) || isAI(M))
+	var/current_time = world.time
+	if(!M || !M.client || (M.shakecamera > current_time)|| M.stat || isEye(M) || isAI(M))
 		return
-	M.shakecamera = TRUE
+	M.shakecamera = current_time + max(TICKS_PER_RECOIL_ANIM, duration)
 	strength = abs(strength)*PIXELS_PER_STRENGTH_VAL
 	var/steps = min(1, Floor(duration/TICKS_PER_RECOIL_ANIM))-1
-	animate(M.client, pixel_x = rand(-(strength), strength), pixel_y = rand(-(strength), strength), time = TICKS_PER_RECOIL_ANIM)
-	sleep(TICKS_PER_RECOIL_ANIM)
+	animate(M.client, pixel_x = rand(-(strength), strength), pixel_y = rand(-(strength), strength), time = TICKS_PER_RECOIL_ANIM, easing = JUMP_EASING|EASE_IN)
 	if(steps)
 		for(var/i = 1 to steps)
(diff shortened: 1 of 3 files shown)


MistakeNot4892:
Does this definitively work? I wrote it with the sleeps initially because the animation queuing was unreliable and tended to break.

ghostsheet:
I've tried in on a local, it works? Idk if that's enough evidence

code/modules/mob/mob_helpers.dm
```diff
@@ -326,22 +326,17 @@ It's fairly easy to fix if dealing with single letters but not so much with comp
 #define PIXELS_PER_STRENGTH_VAL 16
 
 /proc/shake_camera(mob/M, duration, strength=1)
-	set waitfor = 0
-	if(!M || !M.client || M.shakecamera || M.stat || isEye(M) || isAI(M))
+	var/current_time = world.time
```
MistakeNot4892, line 329: Why do you store this?

code/modules/mob/mob_helpers.dm
```diff
@@ -326,22 +326,17 @@ It's fairly easy to fix if dealing with single letters but not so much with comp
...
+	if(!M || !M.client || (M.shakecamera > current_time)|| M.stat || isEye(M) || isAI(M))
 		return
-	M.shakecamera = TRUE
+	M.shakecamera = current_time + min(TICKS_PER_RECOIL_ANIM, duration)
```
MistakeNot4892: This should be `(steps + 2) * TICKS_PER_RECOIL_ANIM` I think, otherwise you're only delaying until the end of the first animation.

code/modules/mob/mob_helpers.dm
ghostsheet: I was just thinking it would be faster than getting it twice... is it faster? who knows?

code/modules/mob/mob_helpers.dm
ghostsheet: Oh right nevermind I was meant to clamp it

ghostsheet: there we go should be fixed. the steps are based off the duration so it should be correct

code/modules/mob/mob_helpers.dm
```diff
@@ -326,22 +326,17 @@ It's fairly easy to fix if dealing with single letters but not so much with comp
 #define PIXELS_PER_STRENGTH_VAL 16
 
 /proc/shake_camera(mob/M, duration, strength=1)
-	set waitfor = 0
-	if(!M || !M.client || M.shakecamera || M.stat || isEye(M) || isAI(M))
+	var/current_time = world.time
+	if(!M || !M.client || (M.shakecamera > current_time)|| M.stat || isEye(M) || isAI(M))
 		return
-	M.shakecamera = TRUE
+	M.shakecamera = current_time + max(TICKS_PER_RECOIL_ANIM, duration)
```
MistakeNot4892, line 332: The math I pasted earlier was correct. It delays twice for `TICKS_PER_RECOIL_ANIM` before and after the main loop, and delays `steps` times during the loop. Your current math is not accurate to the actual delay that the animation will take.

ghostsheet: Main loop has a -1 at the end. It only runs if the duration is atleast 2x TICKS_PER_RECOIL_ANIM. So the duration account for the first/initial animate and the loop.

The last animate (the return to screen centre) can be safely overridded by the next screen shake (if there are any), it looks better when overridded (you don't want recoil to return perfectly to the centre before another screenshake)

MistakeNot4892: Alright fair enough.

MistakeNot4892 (approved):

PR #1574, opened 2021-04-26
Ports Paradise progress bar rewrite

quardbreak:
## Description of changes
Rewrites progress bar code to one of closed paradise pull-requests.

Progress bars now support themes and use less icons:

изображение

Previously it used 40 sprites: 
* 21 to animate progress from empty to full.
* 19 to animate progress failure.

Now it use per one theme:
1. static backdrop sprite
2. mask overlay
3. fill overlay

Plus we can specify .dmi file for theme, which useful for overriding and creating themes inside modpacks.

Some icon operations use gradient - a function, which introduced in 514 and allow to create transition from one color to another.
With animate, we can animate color transition from one side to another.

* Less sprites.
* Ability to create other shape.
* Grayscaled overlays to use different colors by code.
* Smooth color transition with different segments.
* Flexibility to creating different bar themes, like hazard.
* Modpack-friendly.

## Authorship
@dearmochi: https://github.com/ParadiseSS13/Paradise/pull/15695

## Changelog
:cl:
rscadd: Progress bars use gradient function for transit animation now.
rscdel: Failure animation for progress bars has been removed.
/:cl:


code/datums/progress_bar.dm
```diff
@@ -0,0 +1,190 @@
...
+/**
+  * Creates a new progress bar datum with the given theme.
+  *
+  * Arguments:
+  * * user - The mob seeing the progress bar.
+  * * goal - The goal number.
+  * * target - The atom to display the bar on.
+  * * theme - The progress bar's theme.
+  */
+/proc/create_progress_bar(mob/user, goal, atom/target, theme = "default")
```
MistakeNot4892: Why not just define themes as /decl and pass the type instead of a string?

MistakeNot4892: Or just straight up pass the type of the progress bar to create.

MistakeNot4892:
Closing this for the time being, please reopen when 514 goes stable.

afterthought2:
Reopened by author's request.

code/datums/progress_bar.dm
```diff
@@ -0,0 +1,182 @@
...
+	var/client/client
+	/// The bar's index number within the owning mob's active progress bar list.
+	var/list_index
+	/// Whether the bar is stopping.
+	var/stopping = FALSE
+
+/datum/progress_bar/New(mob/user, goal, atom/target)
+	. = ..()
+	if(!istype(target))
+		EXCEPTION("Invalid target given")
```
afterthought2: I don't think this even throws the exception. Probably best to use `CRASH`, unless you have further handling for invalid targets below.

code/datums/progress_bar.dm
```diff
@@ -0,0 +1,182 @@
...
+	src.goal = goal
+
+	if(user)
+		client = user.client
+		client?.images |= holder
+
+		LAZYINITLIST(user.progressbars)
+		LAZYINITLIST(user.progressbars[target])
+		var/list/bars = user.progressbars[target]
+		bars += src
```
afterthought2: We add ourselves here, but we do not remove ourselves in `Destroy`. This may need to be handled via events.

code/datums/progress_bar.dm
```diff
@@ -0,0 +1,182 @@
...
+		list_index = length(bars)
+		animate(holder, pixel_z = world.icon_size + (height * (list_index - 1)), alpha = 255, time = 0.5 SECONDS, easing = SINE_EASING | EASE_OUT)
+	else
+		animate(holder, alpha = 255, time = 0.5 SECONDS, easing = SINE_EASING | EASE_OUT)
+
+/datum/progress_bar/Destroy()
+	client?.images -= holder
+	QDEL_NULL(holder)
+	QDEL_NULL(backdrop)
+	QDEL_NULL(fill)
```
afterthought2, line 71: Do not qdel images.

code/datums/progress_bar.dm
```diff
@@ -0,0 +1,182 @@
...
+	if(stopping)
+		return
+	if(user?.progressbars[target])
+		var/list/bars = user.progressbars[target]
+		for(var/I in (bars - src))
+			var/datum/progress_bar/P = I
+			if(P.list_index > list_index)
+				P.shift_down()
+
+		bars -= src
```
afterthought2: Suggest doing this before the loop just above for generally cleaner code.

code/datums/progress_bar.dm
```diff
@@ -0,0 +1,182 @@
...
+		if(!length(bars))
+			LAZYREMOVE(user.progressbars, target)
+
+	stopping = TRUE
+	animate(holder, alpha = 0, time = 0.5 SECONDS)
+	QDEL_IN(src, 0.5 SECONDS)
```
afterthought2: May be cleaner to just sleep here (but if you do, check if it needs to be waitfor-false).

code/datums/progress_bar.dm
```diff
@@ -0,0 +1,182 @@
...
+		animate(holder, pixel_z = world.icon_size + (height * (list_index - 1)), alpha = 255, time = 0.5 SECONDS, easing = SINE_EASING | EASE_OUT)
+	else
+		animate(holder, alpha = 255, time = 0.5 SECONDS, easing = SINE_EASING | EASE_OUT)
+
+/datum/progress_bar/Destroy()
+	client?.images -= holder
+	QDEL_NULL(holder)
+	QDEL_NULL(backdrop)
+	QDEL_NULL(fill)
+	return ..()
```
afterthought2, line 71: Need to clear target and user here, based on the patterning of the rest of the GC code.

code/datums/progress_bar.dm
```diff
@@ -0,0 +1,182 @@
...
+  * Arguments:
+  * * progress - The new value, divided with the given goal on creation to calculate the fraction.
+  */
+/datum/progress_bar/proc/update(progress)
+	if(QDELETED(target))
+		qdel(src)
+		return
+	if(user?.client != client)
+		client?.images -= holder
+		if(user.client)
```
afterthought2: Need to check for user existing and not being qdeleted here, unless that's handled elsewhere [EDIT: it's not as far as I can tell].

code/datums/progress_bar.dm
```diff
@@ -0,0 +1,182 @@
...
+		holder.overlays += fill
+
+/**
+  * Updates the fill. Do your visual changes here.
+  *
+  * Arguments:
+  * * fraction - The new (value / goal) fraction.
+  */
+/datum/progress_bar/proc/update_fill(fraction)
+	fill?.filters[1]?:x = -fill_width * (1 - fraction)
```
afterthought2: Use `?.`, or break out the code if need be.

code/datums/progress_bar.dm
```diff
@@ -0,0 +1,182 @@
...
+
+/**
+  * Shifts the progress bar down.
+  */
+/datum/progress_bar/proc/shift_down()
+	--list_index
+	var/shift_height = holder.pixel_z - height
+	animate(holder, pixel_z = shift_height, time = 0.5 SECONDS, easing = SINE_EASING | EASE_OUT)
```
afterthought2: Won't this break the timing of the rest of the animation? You may need to be setting some flag here, or just not animating this.

quardbreak:
I can't make `shift_down()` work again. Probably something broken in `/datum/progress_bar/proc/stop()`. Need help.
Edit: Work fine. Keeping new version.

MistakeNot4892:
Please address @afterthought2's comments and respond on each so I know if they're resolved or not.

quardbreak:
I can't do much now due:
1. `/datum/progress_bar/proc/update_fill` works as is because messing with filters is painful, and we dealing with datum which don't support filter helpers. Changing the look will produce more lines or even make dreamchecker scream.
2. Animations work fine if I see correctly.

code/datums/progress_bar.dm
quardbreak: /datum/progress_bar/proc/update_fill works as is because messing with filters is painful, and we dealing with datum which don't support filter helpers. Changing the look will produce more lines or even make dreamchecker scream.

code/datums/progress_bar.dm
quardbreak: Change applied, comment outdated.

code/datums/progress_bar.dm
quardbreak: Target and user clearing added, comment outdated.

code/datums/progress_bar.dm
quardbreak: Added stop call in Destroy function which should cleanup self from list. Comment outdated.

code/datums/progress_bar.dm
afterthought2: I'm fine with more lines. If there are helpers to avoid this that don't work, improve the helpers. The only excuse for doing it like this is that it's literally impossible to write functionally equivalent code in a different way.

code/datums/progress_bar.dm
quardbreak: Animation flags are added, comment outdated.

code/datums/progress_bar.dm
quardbreak: Replaced with `?.`, comment outdated.

code/datums/progress_bar.dm
quardbreak: `!QDELETED(user)` added, comment outdated.

code/datums/progress_bar.dm
MistakeNot4892: `bars` does not appear to remove src from bars or cut bars?

MistakeNot4892: Didn't realize it was a local case of user.progress_bars.
Only seems to be cleared if the list is empty, rather than properly removed?

code/datums/progress_bar.dm
```diff
@@ -0,0 +1,213 @@
...
+
+		bars -= src
+		if(!length(bars))
+			LAZYREMOVE(user.progress_bars, target)
+
+	if(!QDELING(src))
+		stopping = TRUE
+		update(last_progress)
+		animate(holder, alpha = 0, time = animation_time_fade, flags = ANIMATION_PARALLEL)
+		QDEL_IN(src, animation_time_fade)
```
MistakeNot4892: Think after's previous comment about sleeping rather than using qdel_in might still apply?

quardbreak: I don't know a benefits of sleep to make the corresponding change.

code/datums/progress_bar.dm
afterthought2: The bar itself removes itself in `stop`, which is called concurrently (?) with `Destroy`.

code/datums/progress_bar.dm
```diff
@@ -0,0 +1,213 @@
...
+		LAZYINITLIST(user.progress_bars[target])
+		var/list/bars = user.progress_bars[target]
+		bars += src
+		list_index = length(bars)
+		animate(holder, pixel_z = world.icon_size + (height * (list_index - 1)), alpha = 255, time = animation_time_appear, easing = SINE_EASING | EASE_OUT, flags = ANIMATION_PARALLEL)
+	else
+		animate(holder, alpha = 255, time = animation_time_appear, easing = SINE_EASING | EASE_OUT, flags = ANIMATION_PARALLEL)
+
+/datum/progress_bar/Destroy()
+	stop()
```
afterthought2: OK, so I read through this whole thing again, and this just doesn't pass muster with me. `stop` here has branching logic based on whether we are deleting or not; if you call stop (rather than qdel) it actually gets called twice, first to do cleanup and start the animation, and then again from here to do ???. On the other hand, if you call `qdel`stop is called once, from here, and the animation stuff is triggered but happens in nullspace.

This feels like a needlessly confusing way to structure the execution flow. Do the cleanup in one proc, do the animation in another. Have the call chain be (animation proc) -> (delay and call qdel) -> (Destroy calls the cleanup proc or just does the cleanup there). The cleanup probably shouldn't happen until the animation completes anyway. If I'm way off the mark with that, well, call the cleanup proc at the top of the animation proc too. You can then call the animation proc for a fancy deletion or qdel for a quick one.

code/datums/progress_bar.dm
afterthought2: I think `sleep` is simpler and naturally prefer it (note that you'd need to make this proc waitfor-FALSE for that to work). I don't have strong reasoning beyond that, and would not consider it blocking.

code/datums/progress_bar.dm
```diff
@@ -0,0 +1,213 @@
...
+		animate(holder, alpha = 255, time = animation_time_appear, easing = SINE_EASING | EASE_OUT, flags = ANIMATION_PARALLEL)
+
+/datum/progress_bar/Destroy()
+	client?.images -= holder
+
+	QDEL_NULL(holder)
+	QDEL_NULL(backdrop)
+	QDEL_NULL(fill)
+
+	LAZYASSOCLISTREMOVE(user.progress_bars, target, src)
```
MistakeNot4892: Need a QDELETED check on `target` and `user`.

code/datums/progress_bar.dm
```diff
@@ -0,0 +1,213 @@
...
+/datum/progress_bar/proc/stop()
+	set waitfor = FALSE
+
+	shift_down_all()
+	update(last_progress)
+	animate(holder, alpha = 0, time = animation_time_fade, flags = ANIMATION_PARALLEL)
+	LAZYASSOCLISTREMOVE(user.progress_bars, target, src)
+
+	sleep(animation_time_fade)
+	qdel(src)
```
MistakeNot4892, line 87: `QDEL_IN(src, animation_time_fade)`

code/datums/progress_bar.dm
```diff
@@ -0,0 +1,213 @@
...
+/**
+  * Stops the progress bar and destoy it after a bit.
+  */
+/datum/progress_bar/proc/stop()
+	set waitfor = FALSE
+
+	shift_down_all()
+	update(last_progress)
+	animate(holder, alpha = 0, time = animation_time_fade, flags = ANIMATION_PARALLEL)
+	LAZYASSOCLISTREMOVE(user.progress_bars, target, src)
```
MistakeNot4892: Need to check `QDELETED(user)` and `QDELETED(target)`

code/datums/progress_bar.dm
```diff
@@ -0,0 +1,213 @@
...
+
+	sleep(animation_time_fade)
+	qdel(src)
+
+/**
+  * Calls the shift animation on all attached bars.
+  */
+/datum/progress_bar/proc/shift_down_all()
+	if(user?.progress_bars[target])
```
MistakeNot4892: `progress_bars` is a lazylist, this may runtime.

code/datums/progress_bar.dm
```diff
@@ -0,0 +1,213 @@
...
+	holder.alpha = 0
+	holder.plane = HUD_PLANE
+	holder.appearance_flags = KEEP_TOGETHER | APPEARANCE_UI_IGNORE_ALPHA
+
+	backdrop = image(icon, target, "backdrop")
+
+	fill = image(icon, target, "fill")
+	fill.filters = filter(type = "alpha", icon = icon(icon, "mask"), x = -fill_width)
+
+	holder.overlays += backdrop
```
MistakeNot4892: `holder.overlays += list(backdrop, fill)` to avoid churn

code/datums/progress_bar.dm
```diff
@@ -0,0 +1,213 @@
...
+  *
+  * Arguments:
+  * * progress - The new value, divided with the given goal on creation to calculate the fraction.
+  */
+/datum/progress_bar/proc/update(progress)
+	if(QDELETED(target))
+		qdel(src)
+		return
+
+	if(user?.client != client)
```
MistakeNot4892: This block is going to runtime or do a bunch of `null -= foo` when `client` is null, would be better to explicitly check it here and remove all the `?`.

code/datums/progress_bar.dm
```diff
@@ -0,0 +1,213 @@
...
+		client?.images -= holder
+		if(!QDELETED(user) && user?.client)
+			client = user.client
+			client.images += holder
+
+	. = clamp(progress / goal, 0, 1)
+	last_progress = progress
+	if(fill)
+		holder.overlays -= fill
+		update_fill(.)
```
MistakeNot4892, line 140: Why is this its own proc?

code/datums/progress_bar.dm
```diff
@@ -0,0 +1,213 @@
...
+  * Arguments:
+  * * fraction - The new (value / goal) fraction.
+  */
+/datum/progress_bar/proc/update_fill(fraction)
+	UNLINT(fill?.filters[1]?.x = -fill_width * (1 - fraction))
+
+/**
+  * Shifts the progress bar down.
+  */
+/datum/progress_bar/proc/shift()
```
MistakeNot4892, line 155: Why is this a separate proc?

code/datums/progress_bar.dm
```diff
@@ -0,0 +1,213 @@
...
+/**
+  * Creates a new progress bar datum with the given theme.
+  *
+  * Arguments:
+  * * user - The mob seeing the progress bar.
+  * * goal - The goal number.
+  * * target - The atom to display the bar on.
+  * * theme - The progress bar's theme (type).
+  */
+/proc/create_progress_bar(mob/user, goal, atom/target, theme = /datum/progress_bar/default)
```
MistakeNot4892: What is the purpose of making this a global proc instead of just doing `new` like the older code?

code/datums/progress_bar.dm
quardbreak: For overriding.

code/datums/progress_bar.dm
quardbreak: For overriding.

code/datums/progress_bar.dm
quardbreak: After told me to use sleeps. Why I should go back to qdel_in now?

code/datums/progress_bar.dm
quardbreak: Creating by type is a bit hardcode. I want to left a way to pass different themes by preference choice.

code/datums/progress_bar.dm
MistakeNot4892: In what circumstance would you override it?

quardbreak: If you want to make a new bar theme.

MistakeNot4892 (approved):

PR #1578, opened 2021-04-27
Adding duplicate (non-repository) decl checking to unit tests.

MistakeNot4892:
Adds unit test checking for duplicate or non-repository decls.
Depends on #1577, #1576 and #1575.


PsiOmegaDelta:
It isn't as localized but we do have https://github.com/NebulaSS13/Nebula/blob/dev/code/unit_tests/~unit_test_overrides.dm for unit test related overrides.

comma (approved):

PR #1587, opened 2021-04-30
Extension-based tool backend.

MistakeNot4892:
Tools can now have variable quality and are type agnostic, being stored on an extension.
This PR does not implement usage of the interaction proc, only the tool backend, extension setting and surgery aspects.

:cl:
experimental: A large backend refactor of tools means that some interactions, timings and sounds may have changed. Please report bugs to MistakeNot on Discord.
tweak: Examining an object will show you if it is usable as a tool.
tweak: Ghetto surgery steps have been adjusted to use a shared tool flag system, check the codex for updated tools for surgery steps.
tweak: The incision manager no longer has a special surgery type, but counts as several different surgical tools at once.
/:cl:


code/_onclick/item_attack.dm
```diff
@@ -21,7 +21,8 @@ avoid code duplication. This includes items that may sometimes act as a standard
 
 // Called when the item is in the active hand, and clicked; alternately, there is an 'activate held object' verb or you can hit pagedown.
 /obj/item/proc/attack_self(mob/user)
-	return
+	var/datum/extension/tool/tool = get_extension(src, /datum/extension/tool)
+	return (tool?.handle_physical_manipulation(user)) || FALSE
```
afterthought2, line 25: Noting that this proc is notoriously bad about calling parent and checking return values. I believe some instance chains even invert the meaning of the standard returns for it.

code/modules/tools/tool_archetype.dm
```diff
@@ -0,0 +1,45 @@
...
+/decl/tool_archetype/proc/handle_interaction(var/mob/user, var/atom/tool, var/atom/target, var/expend_fuel = 0)
+	. = istype(user) && istype(target) && can_use_tool(tool, expend_fuel)
+
+/decl/tool_archetype/proc/get_tool_quality_descriptor(var/value)
+
+	if(value < TOOL_QUALITY_WORST)
+		return
+
+	if(value < TOOL_QUALITY_BAD)
+		. = pick("crude", "awful", "dysfunctional")
```
afterthought2: Feel like once you pick, maybe you should attach it to the extension so it uses the same one each time?

code/modules/tools/tool_archetype.dm
```diff
@@ -0,0 +1,45 @@
...
+	if(value < TOOL_QUALITY_DECENT)
+		. = "adequate"
+	if(value < TOOL_QUALITY_GOOD)
+		. = "decent"
+	else
+		. = pick("excellent", "exquisite", "great")
+			
+	if(.)
+		if(article)
+			. = "[(lowertext(.[1]) in global.vowels_no_u) ? "an" : "a"] [.]"
```
afterthought2: The \a macro doesn't work here? Also, I don't think string[n] works (at least on 513); I believe you have to use copytext.

code/modules/tools/tool_archetype_definitions_welder.dm
```diff
@@ -0,0 +1,28 @@
+/decl/tool_archetype/welder
+	name = "welder"
+	use_sound = list('sound/items/Welder.ogg','sound/items/Welder2.ogg')
+	codex_key = TOOL_CODEX_WELDER
+
+/decl/tool_archetype/welder/handle_pre_interaction(var/mob/user, var/atom/tool, var/expend_fuel = 0)
```
afterthought2: I feel like you also have to do the can_use check you do in the main interaction proc here, unless you do it in whatever calls this.

code/modules/tools/tool_atom.dm
```diff
@@ -0,0 +1,39 @@
...
+	if(start_message)
+		to_chat(user, start_message)
+
+	var/datum/extension/tool/tool = get_extension(src, /datum/extension/tool)
+	. = tool.do_tool_interaction(archetype, user, target, delay, fuel_expenditure)
+
+	if(!QDELETED(user))
+		if(.)
+			if(success_message)
+				to_chat(user, success_message)
```
afterthought2: Feel like you should supply a default success message, either from the tool extension or even from the return value of `.`

code/modules/tools/tool_atom.dm
```diff
@@ -0,0 +1,39 @@
...
+		else
+			if(failure_message)
+				to_chat(user, failure_message)
```
afterthought2: This, on the other hand, seems like it will be double-messaging if supplied, as you already give some error messages in the action proc. Maybe either pass the `success_message` and `fail_message` parameters into the `tool.do_tool_interaction` step for more consistent handling, or double-check that what I am saying is false.

code/modules/tools/tool_atom.dm
```diff
@@ -0,0 +1,39 @@
...
+
+/atom/examine(mob/user, distance, infix, suffix)
```
afterthought2: Does this really need `/atom`-level handling? The extension even specifies it only goes on items.

code/modules/tools/tool_extension.dm
```diff
@@ -0,0 +1,39 @@
+/datum/extension/tool
+	expected_type = /obj/item
+	base_type = /datum/extension/tool
+	var/list/tool_values     // Delay multipliers/general tool quality indicators, lower is better but 0 or lower means it isn't valid as a tool.
+	var/list/tool_use_sounds // Associative list of tool to sound/list of sounds used to override the archetype config sounds.
+
+/datum/extension/tool/New(datum/holder, list/_tool_values, list/_tool_use_sounds)
+	..()
+	tool_values = _tool_values
+	tool_use_sounds = _tool_use_sounds
```
afterthought2: I think you should just set these on the extension after creating it (get_extension(); extension.sounds = blah) for better clarity. The one place this is used there are two lists of sounds and it's really confusing to read.

code/_onclick/item_attack.dm
MistakeNot4892: I'm fine with that, it's only at this level so that it can be inherited as needed - only a handful of items even use this functionality.

code/modules/tools/tool_archetype.dm
MistakeNot4892: Why? It's purely decorative.

code/modules/tools/tool_archetype.dm
MistakeNot4892: Oh, this is because the code previously had 'ugly' as an adjective and `\a ugly` produces `a ugly`. I'll remove it.

code/modules/tools/tool_archetype.dm
MistakeNot4892: Removed the pick() in any case, it's more consistent with single values.

code/modules/tools/tool_extension.dm
MistakeNot4892: I'm not sure I agree that it's confusing to read, it's only a couple of extra lines. Setting them after the fact would require retrieving the extension, setting the vars, and then having the extension rebuild the sound lists as it currently does in Initialize - it wouldn't be any more readable than a couple of extra lines in a constructor.

code/modules/tools/tool_atom.dm
MistakeNot4892: Moved it all onto item.

code/modules/tools/tool_atom.dm
MistakeNot4892: I don't understand what you mean. This proc is intended to replace the existing tool interactions which all have distinct starting/ending messages.

code/modules/tools/tool_archetype_definitions_welder.dm
MistakeNot4892: Yep, I got confused a bit and never actually called can_use_tool in main code.

code/modules/tools/tool_atom.dm
MistakeNot4892: Added generic messages generated if no message is supplied.

code/modules/tools/tool_atom.dm
MistakeNot4892: Fixed.

code/_onclick/item_attack.dm
MistakeNot4892: Specifically:
````
/obj/item/clustertool
/obj/item/hydraulic_cutter
/obj/item/power_drill
/obj/item/psychic_power/tinker
````

**Review by afterthought2 — DISMISSED** [review]

**Review by afterthought2 — DISMISSED** [review]

File: code/modules/tools/tool_extension_variable.dm [hunk 9]
```diff
@@ -0,0 +1,37 @@
+/datum/extension/tool/variable
+	var/current_tool
+	var/list/tool_config_sounds // Associative list of tool to sound/list of sounds used to override the archetype use sounds.
+
+/datum/extension/tool/variable/New(datum/holder, list/_tool_values, list/_tool_use_sounds, list/_tool_config_sounds)
+	. = ..()
```
PsiOmegaDelta: I recall there was a post by Lummux, stating that `New/()` should never return anything (at by extension have its `.` value adjusted)

code/modules/tools/tool_extension.dm
MistakeNot4892: With the discovery of `get_or_create_extension` I went ahead and refactored those procs to accept shorter arguments supplied post-extension-setting.

afterthought2 (approved):

PR #1594, opened 2021-05-02
Ices processor

PsyCommando:
Here's the other part of the code for ices I was working on. 
Its a machine that can be setup right before a smelter for extracting gases and liquids from ices ores themselves, and just skipping over anything else.

## Changelog
:cl:
rscadd: Added Ice extractor.
imageadd: Added a placeholder icon for the extractor.
/:cl:

Its currently a work in progress, will be ready to merge soon-ish once I can test this out more and get some feedback.


code/modules/mining/machinery/material_extractor.dm
```diff
@@ -0,0 +1,248 @@
...
+	if(get_dir(loc, O.loc) == get_input_dir())
+		//try processing
+		if(process_ore(O))
+			time_last_input = world.time
+		else
+			//Reject stuff that we cannot process
+			O.dropInto(get_output_loc())
+			if(world.time > (time_last_bark + 4 SECONDS))
+				time_last_bark = world.time
+				state("\The [O.name] was rejected for processing!")
```
Andromeda-K22, line 69: This should probably be silent or have a delay, otherwise it'll get kinda spammy if you accidentally put something like ore into it.

Andromeda-K22: Oh, nevermind.

code/modules/materials/_materials.dm
```diff
@@ -306,6 +306,21 @@ INITIALIZE_IMMEDIATE(/obj/effect/gas_overlay)
...
 
+// Returns the phase of the matterial at the given temperature and pressure
+// #FIXME: pressure is unused currently
+/decl/material/proc/phase_at_temperature(var/temperature, var/pressure = ONE_ATMOSPHERE)
+	//#TODO: implement plasma temperature and do pressure checks
+	if(temperature >= boiling_point)
+		return MAT_PHASE_GAS
+	else if(temperature >= heating_point)
+		return MAT_PHASE_LIQUID
+	else
```
MistakeNot4892: Extraneous `else`.

code/modules/materials/geology/strata_permafrost.dm
```diff
@@ -6,5 +6,14 @@
 		/decl/material/solid/ice/lukrite,
 		/decl/material/solid/ice/rubenium,
 		/decl/material/solid/ice/trigarite,
-		/decl/material/solid/ice/ediroite
+		/decl/material/solid/ice/ediroite,
```
MistakeNot4892, line 9: Might be worth splitting some of these changes off into separate PRs to make this one more manageable for you.

code/modules/mining/machinery/material_extractor.dm
```diff
@@ -0,0 +1,392 @@
...
+			if(MAT_PHASE_PLASMA)
+				dump_gas(M, available_volume)
+
+/obj/machinery/atmospherics/unary/material/extractor/proc/dump_solid(var/decl/material/M, var/available_volume)
+	//Check if we got enough to dump out a single sheet at least
+	if(available_volume < REAGENT_UNITS_PER_MATERIAL_SHEET)
+		return
+	var/expected_sheets = round(available_volume / REAGENT_UNITS_PER_MATERIAL_SHEET)
+	var/removed_volume = round(expected_sheets * REAGENT_UNITS_PER_MATERIAL_SHEET, GAS_EXTRACTOR_MIN_REAGENT_AMOUNT)
+	M.place_sheet(get_output_loc(), expected_sheets)
```
MistakeNot4892: This got changed to `create_object()`, same arg order though.

code/modules/materials/geology/strata_permafrost.dm
PsyCommando: Not too sure if it would really make things more manageable really? This and the ices are pretty much intertwined together.

code/modules/materials/_materials.dm
PsyCommando: Don't think it really changes anything if its there or not? The line will exec properly regardless?

MistakeNot4892: Yes, but it's an unnecessary line and indent as you return the immediate line above.

PsyCommando:
btw, I'm a bit concerned how you managed to start a review on a draft pr XD

PsyCommando:
Anyways, I think this is as good as it'll get for now. Without a way to control reactions accurately there's gonna be issues with a small amount of reagents making it through without reacting.. Using 2 different reagents holder in the code massively reduced the chances of it happening though.

Also, icons are all placeholder debug stuff. Idk if there's a procedure to do to get some new icons done?

code/modules/mining/machinery/material_extractor.dm
PsyCommando: Yeah, fixed that

code/modules/materials/_materials.dm
PsyCommando: done

PsyCommando:
@MistakeNot4892 Can you take a look please? It should be ready for review now.

MistakeNot4892:
Sorry, trying to find time/motivation to go over this one but work has been sapping my ability to brain. It's on my radar.

code/modules/materials/_materials.dm
```diff
@@ -306,6 +306,20 @@ INITIALIZE_IMMEDIATE(/obj/effect/gas_overlay)
 /decl/material/proc/products_need_process()
 	return (radioactivity>0) //todo
 
+// Returns the phase of the matterial at the given temperature and pressure
+// #FIXME: pressure is unused currently
+/decl/material/proc/phase_at_temperature(var/temperature, var/pressure = ONE_ATMOSPHERE)
+	//#TODO: implement plasma temperature and do pressure checks
+	if(temperature >= boiling_point)
```
MistakeNot4892, line 313: `if(!isnull(boiling_point) && temperature >= boiling_point)`

code/modules/materials/_materials.dm
```diff
@@ -306,6 +306,20 @@ INITIALIZE_IMMEDIATE(/obj/effect/gas_overlay)
...
+		return MAT_PHASE_GAS
+	else if(temperature >= heating_point)
```
MistakeNot4892, line 315: `melting_point` is the value you want, `heating_point` is for separating a material into its constituent elements.
Also needs a `!isnull()`.

code/modules/mining/machinery/material_extractor.dm
```diff
@@ -0,0 +1,408 @@
...
+/obj/machinery/atmospherics/unary/material/extractor/Initialize(mapload, d = 0, populate_parts = TRUE)
+	. = ..()
+	if(populate_parts)
+		output_container = new/obj/item/chems/glass/bucket(src)
+	air_contents.volume = GAS_EXTRACTOR_GAS_TANK
+	if(!reagents)
+		create_reagents(GAS_EXTRACTOR_REAGENTS_TANK)
+	if(!input_buffer)
+		input_buffer = new(src)
+		input_buffer.create_reagents(GAS_EXTRACTOR_REAGENTS_INPUT_TANK) //Did this here because reimplementing that in the new() proc failed a test for some reasons
```
MistakeNot4892, line 79: For reference, atom-level New() overrides are verboten on Neb, gotta use Initialize.

code/modules/mining/machinery/material_extractor.dm
```diff
@@ -0,0 +1,408 @@
...
+	verbs |= /obj/machinery/atmospherics/unary/material/extractor/verb/FlushReagents
+	verbs |= /obj/machinery/atmospherics/unary/material/extractor/verb/FlushGas
```
MistakeNot4892: These will already be in the verb list for the machine.

code/modules/mining/machinery/material_extractor.dm
```diff
@@ -0,0 +1,408 @@
...
+	QUEUE_TEMPERATURE_ATOMS(src)
+
+/obj/machinery/atmospherics/unary/material/extractor/Destroy()
+	if(istype(input_buffer))
```
MistakeNot4892: Unneeded check, also you need to null input_buffer to allow GC. `QDEL_NULL(input_buffer)` would handle that.

code/modules/mining/machinery/material_extractor.dm
```diff
@@ -0,0 +1,408 @@
...
+/obj/machinery/atmospherics/unary/material/extractor/examine(var/mob/user)
+	. = ..()
+	//Only display info if the screen is there
+	if(get_component_of_type(/obj/item/stock_parts/console_screen))
+		to_chat(user, SPAN_NOTICE("The processing tank gauge reads [round(input_buffer.reagents.total_volume)]/[input_buffer.reagents.maximum_volume] units of liquid"))
+		to_chat(user, SPAN_NOTICE("The internal storage tank gauge reads [round(reagents.total_volume)]/[reagents.maximum_volume] units of liquid"))
+		to_chat(user, SPAN_NOTICE("The internal gas tank pressure gauge reads [air_contents.return_pressure()] kPa"))
+
+		if(is_output_container_full() || is_internal_tank_full())
+			to_chat(user, SPAN_WARNING("Currently idling because one or more of its liquid tanks are full.."))
```
MistakeNot4892: Double full stop, also could probably use 'It is currently' as an opener.

code/modules/mining/machinery/material_extractor.dm
```diff
@@ -0,0 +1,408 @@
...
+		else
+			to_chat(user, SPAN_NOTICE("Everything is working correctly."))
+
+	if(output_container)
+		var/output_desc = SPAN_NOTICE("It has a [output_container.name] in place to receive reagents.")
```
MistakeNot4892: `\a [output_container]`

code/modules/mining/machinery/material_extractor.dm
```diff
@@ -0,0 +1,408 @@
...
+		if(is_output_container_full())
+			output_desc = "[output_desc] [SPAN_WARNING("Its full!")]"
```
MistakeNot4892: `It's`

code/modules/mining/machinery/material_extractor.dm
```diff
@@ -0,0 +1,408 @@
...
+		else
+			to_chat(user, SPAN_NOTICE("Everything is working correctly."))
+
+	if(output_container)
+		var/output_desc = SPAN_NOTICE("It has a [output_container.name] in place to receive reagents.")
+		if(is_output_container_full())
+			output_desc = "[output_desc] [SPAN_WARNING("Its full!")]" 
+		to_chat(user, output_desc)
+	else 
+		to_chat(user, SPAN_NOTICE("It has nothing to pour reagents into."))
```
MistakeNot4892, line 131: You may want SPAN_WARNING here, SPAN_NOTICE is meant for neutral or positive things I think.

code/modules/mining/machinery/material_extractor.dm
```diff
@@ -0,0 +1,408 @@
...
+
+/obj/machinery/atmospherics/unary/material/extractor/attackby(var/obj/item/I, var/mob/user)
+	if(istype(I, /obj/item/chems/glass))
+		add_fingerprint(user)
+		if(!output_container)
+			if(!user.unEquip(I, src))
+				return
+			output_container = I
+			user.visible_message(SPAN_NOTICE("\The [user] place \a [I] in \the [src]."), SPAN_NOTICE("You place \a [I] in \the [src]."))
```
MistakeNot4892: `places`

code/modules/mining/machinery/material_extractor.dm
```diff
@@ -0,0 +1,408 @@
...
+	if(input_buffer.reagents.total_volume >= GAS_EXTRACTOR_MIN_REAGENT_AMOUNT)
+		return
+	input_buffer.reagents.clear_reagents()
+
+//Calculate the amount of reagents we can get from this. Returns a list with each materials and the amount of expected reagents
+/obj/machinery/atmospherics/unary/material/extractor/proc/gather_resulting_reagents_vol(var/obj/O)
+	var/list/processable = list()
+	for(var/k in O.matter)
+		if(can_process_material_name(k))
+			processable[k] = MATERIAL_UNITS_TO_REAGENTS_UNITS(O.matter[k])
```
MistakeNot4892: Would suggest `LAZYSET(processable, k, MATERIAL_UNITS_TO_REAGENTS_UNITS(O.matter[k])` as you can then remove the initial list alloc and it won't waste memory on atoms that have no matter.

code/modules/mining/machinery/material_extractor.dm
```diff
@@ -0,0 +1,408 @@
...
+			O.matter -= k
+	return processable
+
+/obj/machinery/atmospherics/unary/material/extractor/proc/calc_resulting_reagents_total_vol(var/obj/O)
+	var/total = 0
```
MistakeNot4892, line 209: Replacing `total` with `.` (ex. `. = 0`) and omitting the final return would save you a var alloc.

code/modules/mining/machinery/material_extractor.dm
```diff
@@ -0,0 +1,408 @@
...
+	return output_container_free_volume() <= 0
+
+/obj/machinery/atmospherics/unary/material/extractor/proc/is_input_tank_empty()
+	return round(input_buffer.reagents.total_volume, GAS_EXTRACTOR_MIN_REAGENT_AMOUNT) == 0
+	
+/obj/machinery/atmospherics/unary/material/extractor/proc/is_internal_tank_full()
+	return internal_tank_free_volume() <= 0
+
+/obj/machinery/atmospherics/unary/material/extractor/proc/can_process_object(var/obj/O)
+	if(istype(O) && is_type_in_list(O, global.material_extractor_items_whitelist))
```
MistakeNot4892: Could check `length(O.matter)` prior to looking up the type list to speed this proc up.

code/modules/mining/machinery/material_extractor.dm
```diff
@@ -0,0 +1,408 @@
+//Machine meant to extra gases from hydrates
+// Input is clockwise east, output clockwise west, gas output clockwise south
+// The general concept is that it heats up things up to 20c and grabs anything gaseous or liquid at that temp range, with a debuff on liquids, so it doesn't replace more specialized machines.
+#define GAS_EXTRACTOR_GAS_TANK 1000
+#define GAS_EXTRACTOR_REAGENTS_TANK 500
+#define GAS_EXTRACTOR_REAGENTS_INPUT_TANK 500
+#define GAS_EXTRACTOR_OPERATING_TEMP T20C + 5 //Temperature the machine heat stuff to.. Has to be 20c + 5 because someone decided ice melted at 20c
```
MistakeNot4892, line 7: This is solely due to ice being used in drinks :(

code/modules/mining/machinery/material_extractor.dm
```diff
@@ -0,0 +1,408 @@
...
+// Since reagents_holder can't exist on its own for some reasons
+////////////////////////////////////////////////////
+/obj/input_holder
+	atom_flags = ATOM_FLAG_NO_TEMP_CHANGE | ATOM_FLAG_NO_REACT //We wanna disable those to trigger reactions on our own terms
+
+////////////////////////////////////////////////////
+// Actual machine
+////////////////////////////////////////////////////
+/obj/machinery/atmospherics/unary/material/extractor
+	name = "Gas extractor"
```
MistakeNot4892: Lowercase `G` please.

code/modules/mining/machinery/material_extractor.dm
```diff
@@ -0,0 +1,408 @@
...
+		/decl/public_access/public_method/material_extractor/flush_reagents,
+	)
+	var/obj/item/chems/glass/output_container //#TODO: change this when plumbing is a thing
+	var/obj/input_holder/input_buffer //Since reagent_holder needs a parent object to exist on creation we gotta do this horrible hack
+
+/obj/machinery/atmospherics/unary/material/extractor/proc/get_output_loc()
+	return get_step(loc, get_output_dir())
+/obj/machinery/atmospherics/unary/material/extractor/proc/get_output_dir()
+	return turn(dir, 90)
+/obj/machinery/atmospherics/unary/material/extractor/proc/get_input_dir()
```
MistakeNot4892, line 67: Noticing quite a lot of procs in this file that could be macros to save on proc overhead, particularly these dir procs and some of the value checking procs in the bottom of the file. Not a blocker, but maybe worth thinking about.

code/modules/mining/machinery/material_extractor.dm
MistakeNot4892: Also need to handle the output atom as well.

code/modules/mining/machinery/material_extractor.dm
```diff
@@ -0,0 +1,408 @@
...
+		input_buffer.temperature = GAS_EXTRACTOR_OPERATING_TEMP
+		temperature = GAS_EXTRACTOR_OPERATING_TEMP
+		return TRUE
+	return ..()
+
+/obj/machinery/atmospherics/unary/material/extractor/examine(var/mob/user)
+	. = ..()
+	//Only display info if the screen is there
+	if(get_component_of_type(/obj/item/stock_parts/console_screen))
+		to_chat(user, SPAN_NOTICE("The processing tank gauge reads [round(input_buffer.reagents.total_volume)]/[input_buffer.reagents.maximum_volume] units of liquid"))
```
MistakeNot4892: This value looks like a percentage calc but has no *100 or %.

code/modules/mining/machinery/material_extractor.dm
```diff
@@ -0,0 +1,408 @@
...
+		to_chat(user, SPAN_NOTICE("The internal storage tank gauge reads [round(reagents.total_volume)]/[reagents.maximum_volume] units of liquid"))
```
MistakeNot4892: As above.

code/modules/materials/_materials.dm
PsyCommando: Alright, but what about heating products?

code/modules/mining/machinery/material_extractor.dm
PsyCommando: Alright.

code/modules/mining/machinery/material_extractor.dm
PsyCommando: I thought so, but when I tested they weren't.

code/modules/mining/machinery/material_extractor.dm
PsyCommando: Not sure why I did that check there... But yeah, that's true about QDEL_NULL I forgot about it. I'll fix that.

Output atom is just a ref to an object that is stored inside the machine's contents,  It would be a bit redundant to do it again. But I can still qdel it if you really want to.

code/modules/mining/machinery/material_extractor.dm
PsyCommando: Not sure what you mean here? Double full stop? And will do for the flavor text fix.

code/modules/mining/machinery/material_extractor.dm
PsyCommando: Oh, good catch!

code/modules/mining/machinery/material_extractor.dm
PsyCommando: gotcha

code/modules/mining/machinery/material_extractor.dm
PsyCommando: Yeah, its just not, say, an error condition? The machine will work regardless, and fill its internal tank instead. Its more like an extra cue for telling if there's a bucket being filled in there. I didn't want people to feel like they absolutely had to have a bucket in there.

code/modules/mining/machinery/material_extractor.dm
PsyCommando: One issue with that is that this proc also strip out matter entries that can be processed from the item. Because if there's anything left that can't be processed in the ore, it gets spat out the back so the smelter can process it. And we don't want the ore to essentially possibly give double the yield that way.

code/modules/mining/machinery/material_extractor.dm
PsyCommando: True, its just I tend to focus on readability above marginal perf gains. Especially since that's just a primitive.

code/modules/mining/machinery/material_extractor.dm
PsyCommando: Good point!

code/modules/mining/machinery/material_extractor.dm
PsyCommando: Yeah I know. I was just poking fun :P 
Maybe having to take into account the transmission of heat from the container to the reagent content over time could fix that though? So the reagent would have to first heat up to the container's temp before it actually melts. Would probably need to take into account thermal conductivity of reagent containers though. Sounds like a bunch of work though ^^;

code/modules/mining/machinery/material_extractor.dm
PsyCommando: Gotcha

code/modules/mining/machinery/material_extractor.dm
PsyCommando: I guess macros could work. I'll try to cut that down a bit tomorrow,

code/modules/mining/machinery/material_extractor.dm
PsyCommando: Its not a percentage. Its just decoration really. So basically it displays the current volume, a slash, and then the maximum volume.  EX: 120/200

PsyCommando:
Okay pushed a few of the fixes. Still gotta turn some of those extra procs into macros.

code/modules/mining/machinery/material_extractor.dm
MistakeNot4892: Keeping the ref will cause the output atom to fail to GC if it is deleted via an explosion etc. separately to being removed from the machine. You don't necessarily need to qdel it but you need to at least null the ref.

code/modules/mining/machinery/material_extractor.dm
MistakeNot4892: You have two full stops at the end of the string is what I meant.

code/modules/mining/machinery/material_extractor.dm
MistakeNot4892: Fair enough.

code/modules/mining/machinery/material_extractor.dm
MistakeNot4892: Heat capacity is a big hurdle in general, there's a lot of things that would benefit but Physics Hard

code/modules/mining/machinery/material_extractor.dm
MistakeNot4892: Oh I'm a fool, I read that as a single expression not two.

code/modules/mining/machinery/material_extractor.dm
MistakeNot4892: As as above, I'm dumb and misread.

code/modules/materials/_materials.dm
MistakeNot4892: This is related to the Discord mention referring to structural stuff - heating products are kind of hacky and any transition to a proper matter phase system is going to need to fiddle with them for them to make sense.
I need to check your code again, are heating products relevant to the phase of the material?

code/modules/mining/machinery/material_extractor.dm
MistakeNot4892: Alright makes sense.

code/modules/mining/machinery/material_extractor.dm
MistakeNot4892: Behaviorally it should be identical to your current code, just without setting `processable` to a list for objects that have nothing in their matter list. It's a memory optimization rather than a logic change.

code/modules/mining/machinery/material_extractor.dm
MistakeNot4892: Not a big deal, just wary of proc overhead as it is ludicrously expensive in DM. Probably not an important optimization here, happy for it to be left.

code/modules/mining/machinery/material_extractor.dm
MistakeNot4892: That's weird, are you completely certain? I wonder if atmos machines are doing some weird verb stripping at some point for pipes or something.

code/modules/mining/machinery/material_extractor.dm
PsyCommando: I think you don't understand here. 
`#define LAZYSET(L, A, I) if(!L) { L = list(); } L[A] = I;`
Lazyset doesn't remove the element from the source list, which I need to do with line 205 that's missing from that snippet:
`O.matter -= k` 
I fixed it by adding a check if the matter list is empty before instancing a new list instead.

code/modules/mining/machinery/material_extractor.dm
PsyCommando: Oh I see it now! Sorry.

code/modules/mining/machinery/material_extractor.dm
PsyCommando: okay gonna do that then.

code/modules/mining/machinery/material_extractor.dm
PsyCommando: I'll run another test to see if it was just a one time thing...

code/modules/materials/_materials.dm
PsyCommando: The problem is if you look at all ices, even those that were there before, they all use their heating_products list for "melting"?

Also, consider that a lot of the defined materials are compounds and not individual elements, and would probably not phase change cleanly as a single material? Say, you take saltwater, if you heat it to 100c and let it boil away, you'll end up with only the salt left? So its not gonna be salt water vapor, but water vapor and salt?

code/modules/mining/machinery/material_extractor.dm
MistakeNot4892: Ah gotcha, my bad.

code/modules/mining/machinery/material_extractor.dm
PsyCommando: Yeah it works now. I had already removed the verbs from the init proc, so should be fine.

code/modules/materials/_materials.dm
MistakeNot4892: Going to roll with your code since the system is still inconsistent and nothing else uses state changes - will revisit when materials are updated to handle them in a less hacky way.

MistakeNot4892:
I have some quibbles/concerns with some of the specific implementation here, but a lot of it is down to personal taste + the materials system being underdeveloped. This feature is self-contained enough that I'm happy to approve it as-is and we can revisit specific issues down the line when broader updates to the material system are made.

MistakeNot4892 (approved):

PR #1595, opened 2021-05-04
Added update/reaction deferral to chemistry procs, optimized fluids.

MistakeNot4892:
- Reagent addition and removal procs now have an argument that allows general updates to be deferred until later.
- Fluid spreading has been drastically simplified/optimized and should now support turf depth a la Bay (hi @CrimsonShrike).
- Fluid interactions are currently not hooked in to code as they turned out to be the primary reason fluids were killing the master controller. I'm going to look into a permanent solution to them but for now people will need to splash acid with a beaker instead of a vat.

Post-optimizing (Tradeship galley, `Spawn-Fluid`, 10, 5000, `/decl/material/liquid/water`):
image

Pre-optimizing (as above, with fluid_act disabled in code for a fair comparison):
image

:cl:
tweak: Fluid interactions with atoms have been disabled until they can be optimized to be less of a server killer. Mobs will still drown, you just won't melt in acid. On the plus side, fluids are more performant now.
/:cl:

### DIFF
--- a/code/game/atoms.dm
+++ b/code/game/atoms.dm
@@ -307,7 +307,9 @@ its easier to just keep the beam vertical.
 	return
 
 /atom/proc/get_contained_external_atoms()
-	. = contents
+	for(var/atom/movable/AM in contents)
+		if(!QDELETED(AM) && AM.simulated)
+			LAZYADD(., AM)
 
 /atom/proc/dump_contents()
 	for(var/thing in get_contained_external_atoms())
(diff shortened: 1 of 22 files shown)


code/game/atoms.dm
```diff
@@ -307,7 +307,9 @@ its easier to just keep the beam vertical.
 	return
 
 /atom/proc/get_contained_external_atoms()
-	. = contents
+	for(var/atom/movable/AM in contents)
```
comma, line 310: maybe do this fancy as anything thing here? Looks like this would be called a lot

MistakeNot4892: The typecheck is intentional here since it might have weirdness or nulls or unfinished atoms in it. Saves typechecking in loops that call the proc.

afterthought2 (approved):

PR #1596, opened 2021-05-04
Adds Generic FSM System

Neerti:
Separates the generic FSM objects from #1424 into a standalone PR, since @MistakeNot4892 has an interest in them, and the turret PR is gonna probably be awhile.

I imagine it will be useful for certain things, ~~such as for rewritten turrets~~, a future AI system, or something more humble.

An example of how to use can be seen here https://github.com/NebulaSS13/Nebula/pull/1424/files#diff-9f1a6b5f561a3da3fbc72dc5340c344b9a4d64b690ca99e216be60eafc9ab637R1 , but I can try to provide a trivial use-case for it in the code if I or someone else can think of one.


code/datums/state_machine/state.dm
```diff
@@ -0,0 +1,36 @@
...
+		if(!istype(T))
+			CRASH("[type] expected a state transition object but was not given one.")
+		T.set_up(src)
+		transitions[i] = T // Replace the path with a ref to the instance.
+
+// Returns a list of transitions that a FSM could switch to.
+// Note that `holder` is NOT the FSM, but instead the thing the FSM is attached to.
+/decl/state/proc/get_open_transitions(datum/holder)
+	. = list()
+	for(var/thing in transitions)
```
MistakeNot4892: You can do `for(var/decl/state_transition/T as anything in transitions)` for the performance increase of an untyped loop without having to cast the value afterwards.

code/datums/state_machine/state.dm
```diff
@@ -0,0 +1,36 @@
+// An individual state, defined as a `/decl` to save memory.
+// On a directed graph, these would be the nodes themselves, connected to each other by unidirectional arrows.
+/decl/state
+	// Transition decl types, which get turned into refs to those types.
+	// Note that the order DOES matter, as decls earlier in the list have higher priority
+	// if more than one becomes 'open'.
+	var/list/transitions = null
+
+// Called by `SSstate_machines` to have the state decls build the directed graph themselves.
+/decl/state/proc/set_up()
```
MistakeNot4892: As discussed on Discord, it may be worth looking into making this an Initialize() override to avoid having to have an init subsystem.

Neerti:
Gonna draft the PR since I'm gonna make it use the Initialize() override and the FSM be an extension instead of a normal datum.
Also adding a hopefully easy use-case demo for the system.

code/datums/extensions/state_machine.dm
```diff
@@ -0,0 +1,39 @@
+// This contains the current state of the FSM and should be held by whatever the FSM is controlling.
+// Unlike the individual states and their transitions, the state machine objects are not singletons, and hence aren't `/decl`s.
+/datum/extension/state_machine
+	expected_type = /datum
+	base_type = /datum/extension/state_machine
+	var/decl/state/current_state = null // Acts both as a ref to the current state and holds which state it will default to on init.
+
+/datum/extension/state_machine/New()
+	set_state(current_state)
+	return ..()
```
MistakeNot4892: New() shouldn't return a value I believe. Would also probably be better to call `..()` before any further logic since extensions have a bit of setup to do.

code/datums/extensions/state_machine.dm
```diff
@@ -0,0 +1,39 @@
...
+		var/decl/state_transition/choice = choose_transition(options)
+		current_state.exited_state(holder)
+		current_state = choice.target
+		current_state.entered_state(holder)
+		return current_state
+
+// Decides which transition to walk into, to the next state.
+// By default it chooses the first one on the list.
+/datum/extension/state_machine/proc/choose_transition(list/valid_transitions)
+	return LAZYACCESS(valid_transitions, 1)
```
MistakeNot4892: You're already validating that `valid_transitions` has a length in `evaluate()`, might as well just do `valid_transitions[1]` surely?

code/datums/extensions/state_machine.dm
```diff
@@ -0,0 +1,39 @@
...
+
+// Forces the FSM to switch to a specific state, no matter what.
+// Use responsibly.
+/datum/extension/state_machine/proc/set_state(new_state_type)
+	if(istype(current_state))
+		current_state.exited_state(holder)
+	current_state = GET_DECL(new_state_type)
+	current_state.entered_state(holder)
```
MistakeNot4892: Worth typechecking/nullchecking `current_state` here?

code/game/objects/items/paper_fortune_teller.dm
```diff
@@ -0,0 +1,75 @@
...
+*/
+#define MAX_FORTUNES 8
+
+/obj/item/paper_fortune_teller
+	name = "paper fortune teller"
+	desc = "Origami, for children."
+	var/busy = FALSE
+	var/list/fortunes = list(null, null, null, null, null, null, null, null) //why
+	var/color_choices = list(
+		"Red" = "#FF8888",
```
MistakeNot4892: Hex colours will fail CI unless you lowercase the letter components.

code/datums/extensions/state_machine.dm
Neerti: I think I had `Initialize()` on the brain when I wrote that so I'll shove the `..()` above.

code/datums/state_machine/state.dm
```diff
@@ -0,0 +1,32 @@
...
+	. = ..()
+	for(var/i in 1 to LAZYLEN(transitions))
+		var/decl/state_transition/T = GET_DECL(transitions[i])
+		T.from += src
+		transitions[i] = T
+
+// Returns a list of transitions that a FSM could switch to.
+// Note that `holder` is NOT the FSM, but instead the thing the FSM is attached to.
+/decl/state/proc/get_open_transitions(datum/holder)
+	. = list()
```
MistakeNot4892: You LAZYLEN() the result of this proc, but it doesn't seem to be a lazylist. Changing `. += T` to `LAZYADD(., T)` and removing the initial list setting would resolve the inconsistency.

code/game/objects/items/paper_fortune_teller.dm
```diff
@@ -0,0 +1,75 @@
...
+*-----*
+
+*/
+#define MAX_FORTUNES 8
+
+/obj/item/paper_fortune_teller
+	name = "paper fortune teller"
+	desc = "Origami, for children."
+	var/busy = FALSE
+	var/list/fortunes = list(null, null, null, null, null, null, null, null) //why
```
MistakeNot4892: Alternatives:
````
var/list/fortunes[8]
var/list/fortunes = new(8)
````
Lohi prefers the second one.

File: code/game/objects/items/paper_fortune_teller.dm [hunk 9]
```diff
@@ -0,0 +1,75 @@
...
+		"Please help me, I'm trapped in a fortune cookie factory!",
+		"Improve, don't remove.",
+		"Run.",
+		"Please wake up."
+	)
+
+/obj/item/paper_fortune_teller/pre_filled/Initialize(ml, material_key)
+	var/list/fortune_options = possible_fortunes.Copy()
+	for(var/i = 1 to MAX_FORTUNES)
+		var/fortune = pick(fortune_options)
```
MistakeNot4892: `pick_n_take(fortune_options)` will handle removal for you.

MistakeNot4892: ie. `fortunes[i] = pick_n_take(fortune_options)`

code/game/objects/items/paper_fortune_teller.dm
```diff
@@ -0,0 +1,75 @@
...
+		fortune_options -= fortune
+		fortunes[i] = fortune
+	return ..()
+
+
+/obj/item/paper_fortune_teller/proc/alternate(amount)
+	busy = TRUE
+	for(var/i = 1 to amount)
+		advance()
+		sleep(0.5 SECONDS)
```
MistakeNot4892, line 78: Probably should check QDELETED here.

MistakeNot4892:
Did some work on this PR but can't PR it to your branch: https://github.com/MistakeNot4892/Nebula/tree/fsm_stuff

Neerti:
I'm still alive.
~~Was it decided if it should stay as an extension or should I undo the extension-ification and go back to standalone datum?~~ Apparently that has already been done so I guess this is done now.
Tho a rebase might be a good idea.

code/datums/extensions/state_machine.dm
```diff
@@ -0,0 +1,99 @@
+// List and procs for caching state machine instances.
+var/global/list/state_machines = list()
+
+/proc/get_state_machine(var/datum/holder, var/base_type)
+	if(istype(holder) && base_type)
+		var/list/machines = LAZYACCESS(global.state_machines, "\ref[holder]")
```
afterthought2: Why LAZYACCESS? It's not lazy.

code/datums/extensions/state_machine.dm
```diff
@@ -0,0 +1,99 @@
...
+		return islist(machines) && machines[base_type]
+
+/proc/add_state_machine(var/datum/holder, var/base_type, var/fsm_type)
+	if(istype(holder) && base_type)
+		var/holder_ref = "\ref[holder]"
+		var/list/machines = LAZYACCESS(global.state_machines, holder_ref)
+		if(!islist(machines))
+			machines = list()
+			global.state_machines[holder_ref] = machines
+		if(!LAZYACCESS(machines, base_type))
```
afterthought2: Why LAZY? It's just a list now.

code/datums/extensions/state_machine.dm
```diff
@@ -0,0 +1,99 @@
...
+			if(!fsm_type)
+				fsm_type = base_type
+			var/datum/state_machine/machine = new fsm_type(holder)
+			LAZYSET(machines, base_type, machine)
```
afterthought2: Again, it's a list here.

code/datums/state_machine/paper_fortune_fsm.dm
```diff
@@ -0,0 +1,73 @@
+/decl/state/paper_fortune
+	var/state_name
+	var/list/fortune_choices
+	var/icon_state = "closed"
+
+/decl/state/paper_fortune/entered_state(obj/item/paper_fortune_teller/P)
+	if(istype(P))
```
afterthought2: Not sure you want to typecheck every time with these things. Maybe just runtiming is better (with this setup there's no clean unit test, though could test initial states I guess).

code/datums/state_machine/paper_fortune_fsm.dm
```diff
@@ -0,0 +1,73 @@
...
+/decl/state/paper_fortune/closed/entered_state(obj/item/paper_fortune_teller/P)
+	..()
+	if(istype(P))
+		P.choice_counter = 0
+
+// Alternates between this and the below state.
+/decl/state/paper_fortune/open_vertical
+	transitions = list(
+		/decl/state_transition/paper_fortune/vertical_to_horizontal,
+		/decl/state_transition/paper_fortune/open_to_closed
```
afterthought2, line 30: OK, actually I guess you can unit test: for every FSM, start with initial state, then poputate a list of all states accessible from there via `transitions`. For each state, the FSM's expected type shall be a subtype of the expected type of the state.

Would need to add expected type to each state, of course.

afterthought2 (commented):
I think you should make the FSM datum a subtype of `/extension`. It will do exactly what your FSM attachment code is doing for you. Set the non-lazy-init flag to bypass that aspect.

afterthought2:
Upon further discussion, consider my comment re: extensions as a possible thing to consider at best. Rest is worth checking; overall the system seems fine.

afterthought2 (approved):

PR #1597, opened 2021-05-04
Runtime's power grows.

MistakeNot4892:
Every error, every stack trace, every type mismatch merely feeds her unending appetite for chaos.


MistakeNot4892:
Discussed and not desirable due to fiddling with core error code.

PR #1602, opened 2021-05-05
Ports some baystation changes

quardbreak:


code/game/objects/structures/railing.dm
```diff
@@ -290,3 +299,10 @@
 	if(.)
 		if(!anchored || material.is_brittle())
 			take_damage(maxhealth) // Fatboy
+
+	var/start_layer = user.layer
+	user.layer = VEHICLE_LOAD_LAYER
+	addtimer(CALLBACK(src, .proc/reset_layer, start_layer, user), 2)
```
afterthought2: Not a big fan of this implementation; do not mess with a mob's layering externally. If you want to do this, add logic on mobs that adjusts their layers.

code/modules/mob/living/maneuvers/maneuver_leap.dm
```diff
@@ -12,13 +12,19 @@
...
 		animate(user, pixel_z = 16, time = 3, easing = SINE_EASING | EASE_IN)
 		animate(pixel_z = user.default_pixel_z, time = 3, easing = SINE_EASING | EASE_OUT)
 		user.throw_at(get_turf(target), strength, 1, user, FALSE, CALLBACK(src, /decl/maneuver/leap/proc/end_leap, user, target, old_pass_flags))
+		addtimer(CALLBACK(src, .proc/reset_layer, start_layer, user), 4)
 
 /decl/maneuver/leap/proc/end_leap(var/mob/living/user, var/atom/target, var/pass_flag)
 	user.pass_flags = pass_flag
 
+/decl/maneuver/leap/proc/reset_layer(start_layer, mob/living/user)
+	user.layer = start_layer
```
afterthought2: This is closer to being reasonable, but you should make the mob recompute its layer, not use the cached layer.

code/modules/goals/goal_mind.dm
```diff
@@ -39,7 +41,6 @@
 		var/min_goals = 1
 		var/max_goals = 3
 		if(job && LAZYLEN(job.possible_goals))
-			available_goals |= job.possible_goals
```
afterthought2, line 42: Wait, why move this?

code/game/objects/structures/railing.dm
Domic2: Uh, not sure what you mean, something like a mob proc that gets called when layers need changing?

code/modules/goals/goal_mind.dm
CSCMe: Because before it was only called if the add_amount was null.
This was only the case when the goals were generated at spawn, meaning that job-goals would only generate at spawn.
The move makes it so job-goals generate when clicking "reroll goal" or "generate new goal" as well

code/game/objects/structures/railing.dm
afterthought2: The current model mobs have to handle layering is: `reset_layer` sets the layer to whatever it should be, depending on the mob's state. This is called every `Life` tick, and can be called more frequently if need be. Setting the layer on the mob directly conflicts with this: it will be unset randomly the next time `Life` is called (so in 0-2 seconds) and at other times.

So, good implementation: set a state var on mobs (I think there even already are some, having to do with climbing), have the mob's `reset_layer` proc set layer to whichever one you use here if that state var is true, then call `reset_layer` on the mob here. In the callback, unset the state var and call `reset_layer` again.

Tolerable implementation comparable to mediocre code elsewhere: change the layer here, call `reset_layer` on the mob in the callback, do not cache the old layer.

afterthought2: I'll also note that vehicles are not a good example: they do not actually change a mob's layer, despite how the code there looks like.

Domic2: Would it be okay to override UpdateLyingBuckledAndVerbStatus() for the duration of the jump and handle layer changing by two mob/living procs (second one for callback)?
Nothing UpdateLyingBuckledAndVerbStatus() does seems useful or possible for a leaping/vaulting mob, but do correct me if I'm wrong

afterthought2: Yes, just make sure multiple sources of layering changes are handled reasonably (first wins, last wins, whatever, just so long as it's consistent at dealing with the callbacks). You probably should block `reset_layer` and/or `reset_plane_and_layer`, not `UpdateLyingBuckledAndVerbStatus` (which, notably, handles pixel shifting, which is similarly brittle, has identical issues, and comes up a lot more).

Domic2: Alright, all done, thanks for the help. I guess we just wait for eckff now.
Probably a good idea to port stuff after they get merged, though

afterthought2 (approved):

PR #1612, opened 2021-05-07
Radiation medpouch in radiation closet

ghost:
reported by players: "why *radiation* closet contains *toxin* medpouch"
it occured to me that it's faster to make it in browser than write an issue

## Changelog
:cl:
tweak: Radiation closet now contains radiation medpouch instead of toxin one
/:cl:


comma (approved):
probably wasn't updated when rad pouches were added

PR #1614, opened 2021-05-07
job: moves default job type to mapsystem

quardbreak:


maps/~mapsystem/maps_jobs.dm
```diff
@@ -3,7 +3,11 @@
 	var/species_to_job_blacklist = list()
 	var/job_to_species_whitelist = list()
 	var/job_to_species_blacklist = list()
-	var/default_assistant_title = "Assistant"
+
+	/// Default job title which will be set in some cases.
+	var/default_job_title = "Assistant"
```
Techhead0: Why have a separate var for this instead of just pulling the type's title?

MistakeNot4892: Caching the title seems fine since it's used in a bunch of places, but I do think it would be good to set this in /datum/map/New() based on the default job type. ex
````
/datum/map/New()
    ...
    if(ispath(default_job_type, /datum/job))
        var/datum/job/J = default_job_type
        default_job_title = initial(J.title)
````

MistakeNot4892 (approved):

PR #1617, opened 2021-05-08
Adds constructable shuttles

NataKilar:
## Changelog
:cl:
rscadd: You can now build landable ships by toggling docking beacons to construction mode.
/:cl:

Landable ships are now constructable using docking beacons. Much of this is taken directly from Outreach, although I cleaned up the code a little bit. In addition, you can now resync ship engines after construction by hitting them with a multitool. In the future engines should use Loaf's sector registrations system, but because the engines list on ships has extensions rather than the machines itself, this is probably out of scope for this PR


code/modules/overmap/ships/created.dm
```diff
@@ -0,0 +1,7 @@
+/obj/effect/overmap/visitable/ship/landable/created/Initialize(mapload, ship_name, ship_color)
+	name = ship_name
+	shuttle = ship_name
+
+	if(ship_color) color = ship_color
```
MistakeNot4892: Nit: should really be on a newline.

code/modules/overmap/ships/machines/gas_thruster.dm
```diff
@@ -27,6 +27,16 @@
 		if(use_power)
 			add_overlay(emissive_overlay(icon, "nozzle_idle"))
 
+/obj/machinery/atmospherics/unary/engine/attackby(obj/item/I, mob/user)
+	. = ..()
```
MistakeNot4892: Probably want to `return TRUE` to the multitool logic and then call this after that, otherwise you're going to get multiple messages back.

code/modules/overmap/ships/machines/ion_thruster.dm
```diff
@@ -42,6 +42,16 @@
...
 
+/obj/machinery/ion_thruster/attackby(obj/item/I, mob/user)
+	. = ..()
+	if(isMultitool(I) && !panel_open)
+		var/datum/extension/ship_engine/engine = get_extension(src, /datum/extension/ship_engine)
+		if(engine.sync_to_ship())
+			to_chat(user, SPAN_NOTICE("\The [src] emits a ping as it syncs its controls to a nearby ship."))
+		else
+			to_chat(user, SPAN_WARNING("\The [src] flashes an error!"))
+		return
```
MistakeNot4892: As above.

code/modules/shuttles/docking_beacon.dm
```diff
@@ -145,19 +158,70 @@
...
+	if(href_list["check_validity"])
+		if(!construction_mode)
+			return TOPIC_HANDLED
+		check_ship_validity(get_areas())
+		return TOPIC_REFRESH
+	
+	if(href_list["finalize"])
+		if(!construction_mode)
+			return TOPIC_HANDLED
+		var/confirm = alert(user, "This will permanently finalize the ship, are you sure?", "Ship finalization", "Yes", "No")
```
MistakeNot4892: Standard pattern for these is to default to `No` so people don't accidentally confirm.

code/modules/shuttles/docking_beacon.dm
```diff
@@ -145,19 +158,70 @@
...
+		if(!CanInteract(usr,state))
+			return TOPIC_NOACTION
+		if(confirm == "No")
+			return TOPIC_HANDLED
+		else
```
MistakeNot4892: Unnecessary `else`/indentation given the return a line above.

code/modules/shuttles/docking_beacon.dm
```diff
@@ -21,6 +23,11 @@
 	var/docking_height = 10
 	var/projecting = FALSE
 
+	var/construction_mode = FALSE		 // Whether or not the docking beacon is constructing a ship.
+	var/ship_name = ""
+	var/ship_color = COLOR_WHITE
+	var/list/errors = list()
```
MistakeNot4892: Probably should lazylist this.

code/modules/shuttles/docking_beacon.dm
```diff
@@ -145,19 +158,70 @@
...
+		errors.Cut()
+		return TOPIC_REFRESH
+
+	if(href_list["change_color"])
+		var/new_color = input(user, "Choose a color.", "\the [src]", ship_color) as color|null
+		if(!CanInteract(usr,state))
+			return TOPIC_NOACTION
+		if(new_color && new_color != ship_color)
+			ship_color = new_color
+			to_chat(user, SPAN_NOTICE("You set \the [src] to create a ship with <font color='[ship_color]'>this color</font>"))
```
MistakeNot4892: Needs a full stop.

MistakeNot4892 (changes requested):

MistakeNot4892:
I'll probably do an engine refactor in the next few days just because the way that is handled bugs me a bit.

code/modules/shuttles/docking_beacon.dm
NataKilar: Might be fixed now? Not exactly sure what you mean since there is no default argument for alert

code/modules/overmap/ships/machines/gas_thruster.dm
NataKilar: Not sure what you're referring to as the "multitool logic" here.

MistakeNot4892: The line immediately under this comment.

code/modules/shuttles/docking_beacon.dm
MistakeNot4892: When the alert pops up, the leftmost value is highlighted, and hitting space or enter will select it. I am requesting that you change `"Yes", "No"` to `"No", "Yes"` to avoid people accidentally confirming.

code/modules/overmap/ships/machines/gas_thruster.dm
MistakeNot4892: To elaborate: You should handle logic specific to this type before calling `..()`, and you should return TRUE before `..()` if you have already handled an interaction. Otherwise you end up with multiple interactions being completed.

MistakeNot4892 (dismissed):

code/modules/shuttles/shuttle_created.dm
```diff
@@ -0,0 +1,9 @@
+/datum/shuttle/autodock/overmap/created
+	defer_initialisation = TRUE
+
+/datum/shuttle/autodock/overmap/created/New(map_hash, obj/effect/shuttle_landmark/initial_location, list/initial_areas, tag)
+	name = tag
+	. = ..()
+	// Because these are created at runtime, we manually add the shuttle areas to the subsystem.
+	shuttle_area = initial_areas
+	SSshuttle.shuttle_areas += shuttle_area
```
afterthought2: I don't like how you are not calling `SSshuttle.initialize_shuttle` on this thing. It means the shuttle added event won't fire, the SS now acts inconsistently, you have to duplicate misc code here, etc.

code/modules/shuttles/docking_beacon.dm
```diff
@@ -168,4 +230,114 @@
...
+		return FALSE
+	var/obj/effect/shuttle_landmark/temporary/construction/landmark = new(get_turf(src), base_area, get_base_turf(z))
+	new /datum/shuttle/autodock/overmap/created(null, landmark, shuttle_areas.Copy(), ship_name)
+
+	new /obj/effect/overmap/visitable/ship/landable/created(get_turf(src), ship_name, ship_color, global.reverse_dir[dir])
+	permitted_shuttles |= ship_name
+	return TRUE
+
+/obj/effect/shuttle_landmark/temporary/construction
+	flags = null
```
afterthought2: `= 0`

afterthought2 (approved):

PR #1619, opened 2021-05-09
ATM fix

### LINKED ISSUE #1094: Third security mode on ATM
#### Description of issue
It does not work. Doesnt accepts credentials (card is inside)

Login to your account / get blocked ATM

#### Steps to reproduce
Walk to ATM, log in
Select maximum security
Log out
Try to login

#### Issue bingo

ghost:
there was a known problem when you couldnt log into an ATM with enabled third-level security. this happens because of linked account lack on user's ID. (no mind in human at the moment of account number linkage)

also i changed ATM a bit, but it's really useless/pointless (it worked before, the problem isn't in the atm)
it seems to me that it's better to have a `var/valid_card` than a `var/security_level_passed`, because these two could only be binary (in the latter we NEVER could get a zero so it'd be 2 or 3 - no point in it)

this PR closes #1094

## Changelog
:cl:
bugfix: Identification cards now properly get their account number; ATM third-level security will work.
/:cl:

### DIFF
--- a/code/modules/economy/cael/Accounts.dm
+++ b/code/modules/economy/cael/Accounts.dm
@@ -98,9 +98,9 @@
 	return M
 
 //this returns the first account datum that matches the supplied accnum/pin combination, it returns null if the combination did not match any account
-/proc/attempt_account_access(var/attempt_account_number, var/attempt_pin_number, var/security_level_passed = 0)
+/proc/attempt_account_access(var/attempt_account_number, var/attempt_pin_number, var/valid_card)
 	var/datum/money_account/D = get_account(attempt_account_number)
-	if(D && D.security_level <= security_level_passed && (!D.security_level || D.remote_access_pin == attempt_pin_number) )
+	if(D && (D.security_level != 2 || valid_card) && (!D.security_level || D.remote_access_pin == attempt_pin_number) )
 		return D
 
 /proc/get_account(var/account_number)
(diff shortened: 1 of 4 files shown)


code/modules/economy/cael/Accounts.dm
```diff
@@ -98,9 +98,9 @@
 	return M
 
 //this returns the first account datum that matches the supplied accnum/pin combination, it returns null if the combination did not match any account
-/proc/attempt_account_access(var/attempt_account_number, var/attempt_pin_number, var/security_level_passed = 0)
+/proc/attempt_account_access(var/attempt_account_number, var/attempt_pin_number, var/valid_card)
 	var/datum/money_account/D = get_account(attempt_account_number)
-	if(D && D.security_level <= security_level_passed && (!D.security_level || D.remote_access_pin == attempt_pin_number) )
+	if(D && (D.security_level != 2 || valid_card) && (!D.security_level || D.remote_access_pin == attempt_pin_number) )
```
MistakeNot4892, line 103: What is `2`?

MistakeNot4892 (approved):
I think you could probably stand to make that magic number 2 into a define somewhere but otherwise this looks fine.

ghost:
welp it's account security level.. *level*
and there's no define made for it right now 
i can make if it's necessary

PR #1625, opened 2021-05-10
bone setter quality for wrench

ghost:
I tried to do some surgery, was setting a broken bone. Wrench always was working as a bone setter and it is even said in the codex. So why not to return it?

### DIFF
--- a/code/game/objects/items/weapons/tools/wrench.dm
+++ b/code/game/objects/items/weapons/tools/wrench.dm
@@ -17,7 +17,7 @@
 
 /obj/item/wrench/Initialize()
 	. = ..()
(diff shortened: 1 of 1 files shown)


Techhead0:
Many surgical steps define tools to work for them, you can see wrench works for bone setting here:
https://github.com/NebulaSS13/Nebula/blob/dev/code/modules/surgery/bones.dm#L56-L62

75% means it's got a fair chance of working 2 or 3 times without a fail

MistakeNot4892:
If it's definitely not working (rather than just failing) then that's a bug and should be reported as an issue.

MistakeNot4892:
Obviated by #1630.

PR #1626, opened 2021-05-11
Allow const use

Lohikar:
Stops `var/const/whatever` vars from failing the code quality checks.


MistakeNot4892:
`var/global/const` doesn't fail. This was deliberate.

afterthought2:
I think reasonable alternatives are either:

- status quo
- test _against_ `var/global/const/foo` in favor of `var/const/foo`, everywhere

I don't really care which of the above we go with.

PR #1647, opened 2021-05-16
Fire alarms now use a sound token and a looping sound.

### LINKED ISSUE #1640: Fire alarm sounds seem to queue over the top of each other until they are unbearably loud and clashing.
#### Description of issue
Moving in an area with a fire alarm going causes the sound to compound until it's extremely loud and grating.

Fire alarm sound should play once and loop.

#### Steps to reproduce
Start a fire.
Walk into the area while the alarm is going.

Unsure.

Client Version: 513
Server Revision: 38ce123cb4878492f2e6bf6e7e3d395d50eea402 - staging - 2021-05-15
Game ID: cb4-aLLW

#### Issue bingo

- [ ] [Couldn't find an existing issue about this](https://github.com/NebulaSS13/Nebula/issues)

MistakeNot4892:
Fixes #1640.

### DIFF
--- a/code/game/machinery/alarm.dm
+++ b/code/game/machinery/alarm.dm
@@ -53,13 +53,19 @@
 	initial_access = list(list(access_atmospherics, access_engine_equip))
 	clicksound = "button"
 	clickvol = 30
-
 	layer = ABOVE_WINDOW_LAYER
+
+	base_type = /obj/machinery/alarm
+	frame_type = /obj/item/frame/air_alarm
+	stat_immune = 0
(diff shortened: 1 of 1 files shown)


afterthought2 (commented):
Is this actually a looping sound? I thought the point was it should be played once and that's it.

MistakeNot4892:
It's looping. It's eight seconds long, so previously the alarm would play a new instance every two seconds and end up with a nightmarish mess.

afterthought2 (approved):
all right

PR #1649, opened 2021-05-16
multiz throwing

quardbreak:

### DIFF
--- a/code/modules/mob/living/carbon/carbon.dm
+++ b/code/modules/mob/living/carbon/carbon.dm
@@ -226,7 +226,7 @@
 	src.throw_mode_off()
 	if(src.stat || !target)
 		return
-	if(target.type == /obj/screen) 
+	if(target.type == /obj/screen)
 		return
 
 	if(!item)
@@ -257,9 +257,7 @@
 		var/obj/item/I = item
 		itemsize = I.w_class
 
-	if(!unEquip(item))
-		return
-	if(!item || !isturf(item.loc))
+	if(!item)
 		return
 
 	var/message = "\The [src] has thrown \the [item]!"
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/carbon.dm
```diff
@@ -274,17 +272,45 @@
...
 		src.lastarea = get_area(src.loc)
+
 	if(!check_space_footing())
 		if(prob((itemsize * itemsize * 10) * MOB_SIZE_MEDIUM/src.mob_size))
 			var/direction = get_dir(target, src)
 			step(src,direction)
 			space_drift(direction)
 
+	if(multiz_throwing)
+		item.obj_flags |= OBJ_FLAG_NOFALL
```
Techhead0, line 312: Why are you messing with `OBJ_FLAG_NOFALL`? This is the flag for objs to keep mobs from falling through open space (eg a lattice or catwalk)

code/modules/mob/living/carbon/carbon.dm
```diff
@@ -274,17 +272,45 @@
...
+
+		if(!(item.w_class < ITEM_SIZE_GARGANTUAN))
+			to_chat(src, SPAN_WARNING("\The [item] is too huge to throw!"))
+			return
+
+		if(!do_after(src, 2 SECONDS))
+			to_chat(src, SPAN_WARNING("You were interrupted!"))
+			return
+		else
+			multiz_throwing = TRUE
```
Techhead0, line 289: I'd put in a `throw_range -= 1` if to make things be thrown slightly less far

code/modules/mob/living/carbon/carbon.dm
quardbreak: This is only workaround I know to make object not fall after forceMove()

afterthought2: The problem is that it's not being used properly here. This prevents _other_ objects (actually movables, comment is wrong) from falling. It preventing _itself_ from falling is accidental at best, and it will also prevent other stuff from falling for the duration.

The real way to solve this is probably to override `/obj/item/can_fall` to return `FALSE` if currently being thrown (I believe there's a var for that).

bigpep: Did something similar to this, there's a throwing var you can set.

quardbreak: on neb throwing var is a datum, which you can't just TRUE/FALSE

MistakeNot4892: You can check if it's null or not to achieve the same goal.

afterthought2: The issue, I believe, is that this won't work because the var isn't set at this point: this is all happening _prior_ to the throw proper (which happens at the very end of the proc).

Which begs the question of why it's not in the throw handler itself.

MistakeNot4892:
Please feel free to reopen this when it has been updated.

PR #1650, opened 2021-05-16
Removes 513 stubs due to 512 support end

quardbreak:


afterthought2:
Is there a reason to? Are we even supporting 512 at this point?

quardbreak:
dunno, we can just remove those stubs

afterthought2:
Our current 512 support level turns out to be `not supported`, so I think can just delete this stuff (unless you have specific 512-based deployments in mind).

code/__defines/_compile_options.dm
```diff
@@ -1,3 +1,9 @@
 // The default value for all uses of set background. Set background can cause gradual lag and is recommended you only turn this on if necessary.
 // 1 will enable set background. 0 will disable set background.
 #define BACKGROUND_ENABLED 0
+
+#define REQUIRED_DM_VERSION 513
+
+#if DM_VERSION < REQUIRED_DM_VERSION
+#warn Nebula is not tested on BYOND versions older than 513. The code may not compile, and you get to keep both halves if it breaks.
```
MistakeNot4892: Should really be 'and if it does compile it may have severe problems' at the end there, current phrasing is a bit confusing.

code/game/world.dm
```diff
@@ -81,8 +80,8 @@ var/global/game_id = null
 		// dumb and hardcoded but I don't care~
 		config.server_name += " #[(world.port % 1000) / 100]"
 
-	if(byond_version < RECOMMENDED_VERSION)
-		to_world_log("Your server's byond version does not meet the recommended requirements for this server. Please update BYOND")
+	if(byond_version < REQUIRED_DM_VERSION)
+		to_world_log("Your server's byond version does not meet the recommended requirements for this server. Please update BYOND.")
```
MistakeNot4892: Swap `recommended requirements` for `minimum DM version` imo,

MistakeNot4892: Also BYOND should be uppercase.

MistakeNot4892 (approved):

PR #1653, opened 2021-05-17
Testing continue-on-error in CI, ignore pls

quardbreak:


quardbreak:
This thing seems don't makes continuous tests and marks them finished even with errors.

PR #1667, opened 2021-05-20
Spawn logic stub

Mazianni:
Implements a stub for special job spawning logic like the AI, so it can be modified easier by mods.


MistakeNot4892 (approved):

MarinaGryphon (approved):
>no newline at end of file

also pedantry but if it has a default implementation it's not a stub

PR #1668, opened 2021-05-20
Fixes area entered event

quardbreak:
Due of override, it unable to perform entered event trigger.


PsiOmegaDelta (dismissed):

afterthought2:
The omitted parent call is deliberate here; I would not describe this as a fix.

This _will_ have secondary consequences for things which rely on moved events.

At the very least, you need to change the event triggers in ChangeArea to go with this, and then possibly fix resulting unit test failures. I would suggest playtesting machinery power as well.

MistakeNot4892:
Please feel free to reopen this when it has been updated.

PR #1671, opened 2021-05-21
Happy 10th birthday to /datum/recipe.

MistakeNot4892:
- Repathed recipes.
- Removed SScuisine.
- Cleaned up some recipe code.


code/datums/recipe.dm
```diff
@@ -1,102 +1,87 @@
...
+
+/decl/recipe/proc/check_items(var/obj/container)
+	if(!length(items))
+		return TRUE
+	var/container_contents = container?.get_contained_external_atoms()
+	if(length(container_contents) < length(items))
+		return FALSE
+	var/list/needed_items = items.Copy()
+	for(var/itype in needed_items)
+		for(var/atom/movable/AM as anything in container_contents)
```
Lohikar: Why use a typed var at all if you're just doing an istype?

MistakeNot4892: Avoiding a cast.

MistakeNot4892: Nvm, just realized type is irrelevant there.

Lohikar (approved):
I haven't looked at the actual logic being used here, but the code looks good.

PR #1672, opened 2021-05-21
Fire and fluids fixes.

MistakeNot4892:
Pointing this at `staging` despite being a fairly complex rewrite because fire and fluids are completely broken there.

- SSair no longer handles fires, except igniting them if the zone is above the fuel flashpoint.
- The fire object now handles checking for and consuming fuel.

Still tweaking the numbers and chasing down bugs with this, DNM.

### DIFF
--- a/code/game/objects/effects/fluids.dm
+++ b/code/game/objects/effects/fluids.dm
@@ -51,28 +51,8 @@
 	REMOVE_ACTIVE_FLUID(src)
 	SSfluids.pending_flows -= src
 	. = ..()
-	if(istype(T))
-		if(T.zone)
-			T.zone.fuel_objs -= src
-		if(reagents?.total_volume > 0)
-			T.wet_floor()
-
-/obj/effect/fluid/proc/remove_fuel(var/amt)
-	for(var/rtype in reagents.reagent_volumes)
-		var/decl/material/liquid/fuel = GET_DECL(rtype)
-		if(fuel.fuel_value)
-			var/removing = min(amt, reagents.reagent_volumes[rtype])
-			reagents.remove_reagent(rtype, removing)
-			amt -= removing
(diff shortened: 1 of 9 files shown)


code/game/objects/effects/fluids.dm
```diff
@@ -116,6 +96,39 @@
...
+	if(!reagents || !loc)
+		return 0
+
+	var/update_air = FALSE
+	var/update_reagents = FALSE
+	var/datum/gas_mixture/air_contents = loc.return_air()
+	for(var/rtype in reagents?.reagent_volumes)
+		var/decl/material/mat = GET_DECL(rtype)
+		var/fuel_amount = reagents.reagent_volumes[rtype] * FIRE_LIQUD_MIN_BURNRATE
+		if(fuel_amount)
```
comma, line 118: wouldn't this make non-burnabe reagents disappear too, since you only check fuel_value inside of this block for return value?

code/modules/ZAS/Fire.dm
```diff
@@ -181,34 +133,46 @@ If it gains pressure too slowly, it may leak or just rupture instead of explodin
 			else
 				enemy_tile.adjacent_fire_act(loc, air_contents, air_contents.temperature, air_contents.volume)
 
+	if(firelevel > 6)
+		set_light(7, 3, no_update = TRUE)
```
comma, line 137: should probably be in update icon too, I tihnk that's where we put light changes too

code/game/objects/effects/fluids.dm
MistakeNot4892: Oh yeah.

MistakeNot4892:
Moved the fixes part of this to #1689, will reopen this to dev when fire rewrite is ready to go.

PR #1673, opened 2021-05-21
Something part of lying combat

quardbreak:
* Disables resting button for synths.
* Rest button work immediately now. Pressing button again to get up have do_after timer now.
* ~~You can pickup and hold items while resting now. If you pickup something larger than normal sized item, it will drop as before.~~ Not in this PR

### DIFF
--- a/code/modules/mob/living/silicon/robot/robot.dm
+++ b/code/modules/mob/living/silicon/robot/robot.dm
@@ -143,6 +143,9 @@
 
 	add_robot_verbs()
 
+	// Disables lay down verb for robots due they're can't lay down and it cause some movement, vision issues.
+	verbs -= /mob/living/verb/lay_down
+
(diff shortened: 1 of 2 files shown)


MistakeNot4892:
I would like the inability to throw items or use guns while prone to be part of any buffs to lying down item use.

code/modules/mob/living/silicon/robot/robot.dm
```diff
@@ -1124,3 +1124,8 @@
 
 /mob/living/silicon/robot/handle_pre_transformation()
 	QDEL_NULL(mmi)
+
+// Disables lay down verb for robots due they're can't lay and it cause some movement/resting issues.
+/mob/living/silicon/robot/lay_down()
+	set category = null
+	return
```
afterthought2: Should really be removing the verb, not doing this.

afterthought2 (commented):
Not into letting you use items while downed. Downed is the premier disable in the game; I feel it needs to stay.

MistakeNot4892 (approved):

PR #1675, opened 2021-05-21
Adds chance to kick lying ammo cases on walk

quardbreak:


code/modules/projectiles/ammunition.dm
```diff
@@ -41,6 +41,29 @@
...
+	if(isliving(AM))
+		var/mob/living/L = AM
+
+		if(L.buckled)
+			return
+
+		if(!MOVING_DELIBERATELY(L) && prob(10))
+			on_casing_cross(L)
+
+/obj/item/ammo_casing/proc/on_casing_cross(atom/AM)
```
MistakeNot4892: Why make this a separate proc? Just adds proc overhead and a redundant check on `AM`.

MistakeNot4892 (approved):

PR #1676, opened 2021-05-22
fixes ai hologram choosing

Mazianni:
Fixes a bug with AI hologram choosing. Is_traitor always returns false for non-traitors, meaning you couldn't pick an AI icon.

### DIFF
--- a/code/datums/ai/ai_holo.dm
+++ b/code/datums/ai/ai_holo.dm
@@ -11,7 +11,12 @@
 
 
 /decl/ai_holo/proc/may_be_used_by_ai(var/mob/living/silicon/ai/AI)
-	return AI.is_traitor()
+	if(requires_malf)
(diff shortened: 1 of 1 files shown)


code/datums/ai/ai_holo.dm
```diff
@@ -11,7 +11,12 @@
 
 
 /decl/ai_holo/proc/may_be_used_by_ai(var/mob/living/silicon/ai/AI)
-	return AI.is_traitor()
+	if(requires_malf)
```
MistakeNot4892: `return !requires_malf || AI.is_traitor()`? Current code will lock malf AIs to the malf icons only.

MistakeNot4892 (approved):

MistakeNot4892 (approved):

PR #1677, opened 2021-05-22
implements do_spawn_special into round-start spawning

Mazianni:
rips out the old AI spawn logic and jams the actual special job spawning proc in.

### DIFF
--- a/code/controllers/subsystems/ticker.dm
+++ b/code/controllers/subsystems/ticker.dm
@@ -275,14 +275,16 @@ Helpers
 /datum/controller/subsystem/ticker/proc/create_characters()
 	for(var/mob/new_player/player in global.player_list)
 		if(player && player.ready && player.mind)
-			if(player.mind.assigned_role=="AI")
-				player.close_spawn_windows()
-				player.AIize()
-			else if(!player.mind.assigned_role)
+			if(!player.mind.assigned_role)
+				continue
+			var/mob/living/newplayer = player.create_character()
+			if(isnull(newplayer)) //This means create_character has failed, so, pass them over and continue.
+				continue
+			if(newplayer.mind.assigned_job.do_spawn_special(newplayer, player, FALSE))
+				qdel(player)
 				continue
 			else
-				if(player.create_character())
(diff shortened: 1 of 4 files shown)


code/controllers/subsystems/ticker.dm
```diff
@@ -275,14 +275,16 @@ Helpers
 /datum/controller/subsystem/ticker/proc/create_characters()
 	for(var/mob/new_player/player in global.player_list)
 		if(player && player.ready && player.mind)
```
Lohikar: The typed loop above should make it impossible for `player` to be null here: removing the two player null checks would be a bit cleaner.

code/controllers/subsystems/ticker.dm
```diff
@@ -274,15 +274,15 @@ Helpers
...
+			if(!player.mind.assigned_role)
+				continue
+			var/mob/living/newplayer = player.create_character()
+			if(newplayer.mind.assigned_job.do_spawn_special(newplayer, player, FALSE))
+				qdel(player)
 				continue
 			else
-				if(player.create_character())
-					qdel(player)
+				qdel(player)
```
afterthought2: I may be misreading the diff here, but as the clause following this is an `else if`, this seems equivalent to
```
newplayer.mind.assigned_job.do_spawn_special(newplayer, player, FALSE)
qdel(player)
```
(from line 281-285).

code/game/jobs/job/silicon.dm
```diff
@@ -39,7 +39,8 @@
 	var/mob/living/silicon/ai/A = character
 	A.on_mob_init()
 
-	new_player_mob.AnnounceCyborg(character, title, "has been downloaded to the empty core in \the [character.loc.loc]")
+	if(latejoin)
+		new_player_mob.AnnounceCyborg(character, title, "has been downloaded to the empty core in \the [character.loc.loc]")
```
afterthought2: `get_area` is the 21st century way of doing `.loc.loc`.

afterthought2 (approved):

PR #1681, opened 2021-05-23
Update helm skill and operator handling

SierraKomodo:
Port of https://github.com/Baystation12/Baystation12/pull/30643

- Fixes ambiguity issues with how the helm decides who's skills are being used to calculate maneuvering and meteor dodging.
- Adds visual messages indicating when someone takes or releases control of a helm.
- Adds debug log messages for the above to allow for easier debugging if this breaks.
- Restricts helm control and overmap view access to a single person (The same person whose skills are being used).

## Changelog
:cl:
bugfix: Ghosts will no longer break skill checks for helm controls.
tweak: Ship helm consoles now only allow a single person to use manual control at a time, including viewing the overmap through the helm.
rscadd: Messages are now displayed whenever another mob takes control of a helm control, to make it outwardly obvious who's skills are used to calculate flight.
admin: Debug log messages were added for helm control changes, for tracking of potential issues on live.
/:cl:

### DIFF
--- a/code/modules/overmap/ships/computers/helm.dm
+++ b/code/modules/overmap/ships/computers/helm.dm
@@ -23,6 +23,8 @@ var/global/list/overmap_helm_computers
 	var/dy		//coordinates
 	var/speedlimit = 1/(20 SECONDS) //top speed for autopilot, 5
 	var/accellimit = 1 //manual limiter for acceleration
+	/// The mob currently operating the helm - The last one to click one of the movement buttons and be on the overmap screen. Set to `null` for autopilot or when the mob isn't in range.
+	var/mob/current_operator
 
 /obj/machinery/computer/ship/helm/Initialize()
 	. = ..()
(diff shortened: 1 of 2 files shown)


code/modules/overmap/ships/computers/helm.dm
```diff
@@ -23,6 +23,8 @@ var/global/list/overmap_helm_computers
 	var/dy		//coordinates
 	var/speedlimit = 1/(20 SECONDS) //top speed for autopilot, 5
 	var/accellimit = 1 //manual limiter for acceleration
+	/// The mob currently operating the helm - The last one to click one of the movement buttons and be on the overmap screen. Set to `null` for autopilot or when the mob isn't in range.
+	var/mob/current_operator
```
MistakeNot4892: Would recommend using a weakref for this if you're adding new references.

code/modules/overmap/ships/computers/helm.dm
```diff
@@ -43,6 +45,14 @@ var/global/list/overmap_helm_computers
 
 /obj/machinery/computer/ship/helm/Process()
 	..()
+
+	if (current_operator)
+		if (!linked)
+			to_chat(current_operator, SPAN_DANGER("\The [src]'s controls lock up with an error flashing across the screen: Connection to vessel lost!"))
+			set_operator(null, TRUE)
+		if (!Adjacent(current_operator) || CanUseTopic(current_operator) != STATUS_INTERACTIVE || !viewing_overmap(current_operator))
```
MistakeNot4892: Was this meant to be an `else if` or is there a missing return earlier?

SierraKomodo: That should in fact be an `else if`

code/modules/overmap/ships/computers/helm.dm
SierraKomodo: Pushed as a separate commit for review purposes, will squash once I know that's the route we're going.

MistakeNot4892 (approved):

PR #1682, opened 2021-05-24
Fixes fields tag exploit in word processor

quardbreak:
Thanks OnyxBay for this solution.

### DIFF
--- a/code/modules/modular_computers/file_system/programs/generic/wordprocessor.dm
+++ b/code/modules/modular_computers/file_system/programs/generic/wordprocessor.dm
@@ -7,13 +7,15 @@
 	size = 4
 	available_on_network = 1
 	nanomodule_path = /datum/nano_module/program/computer_wordprocessor/
+
+	usage_flags = PROGRAM_ALL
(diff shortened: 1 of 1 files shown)


code/modules/modular_computers/file_system/programs/generic/wordprocessor.dm
```diff
@@ -108,6 +112,21 @@
 		var/newtext = sanitize(replacetext(input(usr, "Editing file '[open_file]'. You may use most tags used in paper formatting:", "Text Editor", oldtext) as message|null, "\n", "\[br\]"), MAX_TEXTFILE_LENGTH)
 		if(!newtext)
 			return
+
+		//Count the fields
+		var/laststart = 1
+		var/fields
+		while(TRUE)
```
comma: infinite loop with no sanity doesn't look that good
maybe use regex instead?
something like

```
	var/fields = 0
	var/regex/re = regex(@"\[field\]","g")
	while(re.Find(text))
		fields++
```

comma (approved):

PR #1686, opened 2021-05-25
Implements gauss in projectile click point dispersion

quardbreak:
Instead using rand, it uses gauss proc for click dispersion.
Gaussian proc from tgstation. Idea by bear1ake and this is port from Infinity Station.

### DIFF
--- a/code/_helpers/gauss.dm
+++ b/code/_helpers/gauss.dm
@@ -1,0 +1,33 @@
+#define ACCURACY 10000
+
+/**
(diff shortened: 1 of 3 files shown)


MistakeNot4892:
I still don't understand what the purpose of this change is.

MistakeNot4892:
ty Jade
````
[10:15 PM] essentially done: basically it makes the dispersion less extreme
[10:15 PM] essentially done: it tends towards the mean far more often
````

MistakeNot4892 (approved):

PR #1621, opened 2021-05-09
Holotablets

ghost:
Adds another type of modular computer: holotablets.
They're meant to be used as a replacement for paper (reports, clipboards, this stuff. also, e-reading) - no network and additional hardware. Available only in games vending for now.. might be worth adding them into fabricators?

It's just fancy and useless, that's all. Of course you can read books and fill out reports on your PDA.. But hey, it glows and has big screen!

image
^ all of the types for now (they also have powered down state, and the "screen" is animated with scanline)
image
^ light power

PR #1691, opened 2021-05-26
Holotablets

quardbreak:
closes #1621
I cleanup some things, fixed light radius for normal tablets.


code/modules/modular_computers/computers/subtypes/dev_tablet.dm
```diff
@@ -5,7 +5,7 @@
 	icon_state = "tablet"
 
 	w_class = ITEM_SIZE_SMALL
-	light_strength = 5 // same as PDAs
+	light_strength = 2 // same as PDAs
```
Techhead0, line 8: IDK if it's the original or the new one, but one of these two comments is false

EDIT: The original comment is wrong, so harmony is restored

MistakeNot4892 (approved):

PR #1692, opened 2021-05-27
Various unrelated changes.

MistakeNot4892:
- Deletes medical side effects. Ailments make them redundant, and the code is spectacularly bad (instantiating every type of medical effect in human Life() then qdeling them).
- Repathed `/datum/persistent` to `/decl/persistence_handler` as they were just doing singleton stuff manually anyway.
- Null checked species on lung breath setting because it was showing up as the load_failed message for saves. No idea why.
- Replaced some `get_decl()` with `GET_DECL()` because SHOUTING.


afterthought2 (approved):
So I looked into that breath runtime myself a while back, and can't understand how species is null there.

MistakeNot4892:
Neither can I, but this seems to have stopped it happening at least.

PR #1695, opened 2021-05-28
optional rust-g support

quardbreak:
Opening for diff and CI, don't replacing legacy logging because I still didn't figured out how to make this part optional.


MistakeNot4892:
Looks like this generally doesn't touch live code unless you specifically opt to tell it to override builtins. No objections to a merge on my end.

@afterthought2 looks like the logging issue isn't relevant as it doesn't replace logging in this PR.

@quardbreak please fix the conflicts and this can likely be approved.

afterthought2:
May want to have @Lohikar look it over.

PR #1702, opened 2021-05-29
fixes a couple of sort errors that were being spammed.

afterthought2:


MistakeNot4892:
Grabbed this commit for staging.

PR #1708, opened 2021-05-29
Supply packs now autogenerate their prices.

MistakeNot4892:
Depends on #1704 #1705 #1706 #1707. Prices may be insane even after deps are merged, waiting to see.


MistakeNot4892:
Also depends on #1716 #1717

MistakeNot4892:
Cigarette vendor is being a pain, I think it's because individual cig papers are 174cr.

MistakeNot4892:
Cigs are now the only supply pack that doesn't autogenerate worth.
Current prices: https://cdn.discordapp.com/attachments/678826323728400394/849618086519635978/message.txt
Personally I am happy with how things stand currently, weirdness can be resolved as it crops up.

afterthought2 (approved):

PR #1710, opened 2021-05-30
H/E Pipes glow

Mazianni:
Radiator pipes now emit light and have a faint proper glow to them depending on temperature.
https://cdn.discordapp.com/attachments/678826323728400394/848542092765823056/unknown.png


code/modules/atmospherics/he_pipes.dm
```diff
@@ -100,6 +103,10 @@
 				for (var/thing in animate_targets)
 					var/atom/movable/AM = thing
 					animate(AM, color = rgb(h_r, h_g, h_b), time = 20, easing = SINE_EASING)
+				filters[1].color = rgb(h_r, h_g, h_b)
```
MistakeNot4892: Not sure I like this. Can you store the `filter()` return somewhere instead?

Lohikar: This also isn't updating the color of mimics.

Mazianni: The mimics don't have any filters - the filters are on the base H/E pipe itself, so it doesn't update the mimic filters as intended.

MistakeNot4892 (approved):

code/_helpers/visual_filters.dm
```diff
@@ -35,4 +35,23 @@
 	if(thing)
 		LAZYREMOVE(filter_data, filter_name)
 		filters -= thing
-		update_filters()
\ No newline at end of file
+		update_filters()
+
+/// Animate a given filter on this atom. All params after the first are passed to animate().
```
Lohikar, line 40: This comment didn't get updated to match the new proc version.

PR #1711, opened 2021-05-30
ports visual filter helpers from polaris

Mazianni:
dependency for #1710 also generally useful to use filters in more places for pretty things ™️

### DIFF
--- a/code/_helpers/visual_filters.dm
+++ b/code/_helpers/visual_filters.dm
@@ -23,6 +23,7 @@
 		var/list/arguments = data.Copy()
 		arguments -= "priority"
 		filters += filter(arglist(arguments))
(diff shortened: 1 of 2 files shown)


MistakeNot4892 (dismissed):

Lohikar:
I feel like this should hook into Z-Mimic as well, an `UPDATE_OO_IF_PRESENT` in `update_filters()` would probably work.

Lohikar (changes requested):
This should hook into Z-Mimic as well, an UPDATE_OO_IF_PRESENT in update_filters() would probably be all that you need to do.

Lohikar:
Though handling animating filters for mimics might be more involved, and is something that #1710 probably will need.

Lohikar (approved):

PR #1722, opened 2021-06-02
nexus station

ghost:
bare minimum intended to be used for "multiple shuttles plus central station" idea
features 6 docking ports
no exterior airlocks i was too lazy
equipped with a rust
has bar 
image

no actual jobs/access set onto the areas
the checks will obv fail since human error

@afterthought2


ghost:
all set ready for merge
not a playable map right now it needs to be integrated with this wonderful shuttle modpack

quardbreak:
I have only two things which I don't personally like:
1. Map name, which is don't really describe map itself.
2. The lower z-map file called as "mapname-`0`", which refers to tradeship's disused level. Would be better to name maps from 1, because first z level is one.

quardbreak:
The entire areas file have spaces instead tabs. Replace them.

afterthought2:
Yeah, `core` is confusing, call it something less generic. Whether this should be a tested map, in a modpack with the shuttles, or in a separate repo is worth maintainer discussion.

MistakeNot4892:
Maybe `nexus` or `locus` as replacements for `core`?

ghost:
(ready for review)

MistakeNot4892 (approved):
Fine with this being merged, it's tiny and won't hold up CI.

PR #1732, opened 2021-06-05
Fixes air passing on inflatables.

MistakeNot4892:
Thanks @Lohikar


Lohikar (approved):
I swear I knew what I was doing in that Europa PR.

PR #1735, opened 2021-06-05
railing material appearance inherency

ghost:
sounds like a bug so im putting this to stable

### DIFF
--- a/code/game/objects/structures/railing.dm
+++ b/code/game/objects/structures/railing.dm
@@ -18,12 +18,7 @@
 	var/neighbor_status = 0
 
 /obj/structure/railing/mapped
-	color = COLOR_GUNMETAL
 	anchored = TRUE
-
-/obj/structure/railing/mapped/Initialize()
-	. = ..()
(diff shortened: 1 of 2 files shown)


afterthought2:
Don't think that's the fix. Add code to `on_update_icon` or `update_material`. There's also a hard override of the color on the mapped subtype for some reason.

ghost:
moved to railing file,

image
^^^
everything is fine (color has been inherited, thats aluminium color)
image
steel & glass

code/game/objects/structures/railing.dm
```diff
@@ -57,6 +52,11 @@
 	else
 		..()
 
+/obj/structure/railing/update_materials(keep_health)
+	. = ..()
+	color = material.color
+	alpha = 50 + (material.opacity * 255)
```
MistakeNot4892: `Clamp((50 + material.opacity * 255), 0, 255)`

MistakeNot4892 (approved):
There should be a material flag you can slap on structures to do this automatically but this will work as a temp fix.

PR #1736, opened 2021-06-06
fusion engines changes

ghost:
changes current fusion thrusters from magical temperature increase to actual fusion reacting based on RUST's reaction datums
also: i removed that e_type from engine extension initialization because it did no sense at all *(engines extension had a engine_type variable used which is just being overrided to engine_type = e_type so this variable did not matter at all, i removed the init override for it for gas thruster why am i even writing this idk)*

new version does some processing inside the engine allowing the hydrogen to be used 
it is way more efficient than simple gas or ions and it's construction/assembly should cost more $ (thats why i pumped a lot of components on it)

working fuel mixes would be.. only `hydrogen/hydrogen` or `hydrogen/deut` *<- most profitable perhaps* or `deut/trit`, `deut/deut` will not work since there is not /decl/material/neutron - not quite realistic but still can be used for supplementary reaction
uses almost no power compared to other versions

will violently explode like those consoles on star trek if emped

upd:
it does not work like rust (keeping constant nuclear fusion), it works differently - for a moment of thrust it pumps the gas inside the chamber and contains it, starting the fusion and directing it into the thrust direction

### DIFF
--- a/code/modules/overmap/ships/circuits.dm
+++ b/code/modules/overmap/ships/circuits.dm
@@ -17,13 +17,15 @@
 	name = "circuitboard (fusion thruster)"
 	icon = 'icons/obj/modules/module_controller.dmi'
 	build_path = /obj/machinery/atmospherics/unary/engine/fusion
-	origin_tech = "{'powerstorage':1,'engineering':2}"
+	origin_tech = "{'wormholes':1,'magnets':3,'powerstorage':3,'engineering':2}"
 	req_components = list(
-		/obj/item/stack/cable_coil = 30,
-		/obj/item/pipe = 2
+		/obj/item/stack/cable_coil = 90,
+		/obj/item/pipe = 6,
+		/obj/item/stock_parts/capacitor/adv = 3,
+		/obj/item/stock_parts/matter_bin = 4,
+		/obj/item/stock_parts/subspace/crystal = 1,
+		/obj/item/stock_parts/scanning_module = 1,
+		/obj/item/stock_parts/micro_laser/ultra = 2,
+		/obj/item/stock_parts/manipulator = 1
 		)
-	additional_spawn_components = list(
-		/obj/item/stock_parts/matter_bin = 1,
-		/obj/item/stock_parts/capacitor = 2,
-		/obj/item/stock_parts/power/apc/buildable = 1,
-		)
+	additional_spawn_components = list(/obj/item/stock_parts/power/apc/buildable = 1)
--- a/code/modules/overmap/ships/device_types/_engine.dm
+++ b/code/modules/overmap/ships/device_types/_engine.dm
@@ -14,9 +14,8 @@ var/global/list/ship_engines = list()
 	var/volume_per_burn = 40			// Amount in litres of propellant to use per burn.
 	var/charge_per_burn = 36000 		//10Wh for default 2 capacitor, chews through that battery power! Makes a trade off of fuel efficient vs energy efficient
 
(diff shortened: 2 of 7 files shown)


ghost:
`[21:00:05] Runtime in timer.dm,155: Invalid timer: Timer: Timer: -1 ([0x2107bc97]), TTR: 2313.76, wait:2.13 Flags: , callBack: [0x2107bc9a], callBack.object: the basic radio transmitter[0x20148dc](/obj/item/stock_parts/radio/transmitter/basic/buildable), callBack.delegate:/obj/item/stock_parts/radio/transmitter/proc/transmit(), source: code/game/machinery/_machines_base/stock_parts/radio/transmitter.dm:14Prev: NULL, Next: NULL, SPENT(2322.99), QDELETED, NO CALLBACK world.time: 2322.99, head_offset: 598.4, practical_offset: 736
  proc name: fire (/datum/controller/subsystem/timer/fire)
  src: Timer (/datum/controller/subsystem/timer)
  call stack:
  Timer (/datum/controller/subsystem/timer): fire(0)
  Timer (/datum/controller/subsystem/timer): ignite(0)
## WARNING: Invalid timer: Timer: Timer: -1 ([0x2107bc97]), TTR: 2313.76, wait:2.13 Flags: , callBack: [0x2107bc9a], callBack.object: the basic radio transmitter[0x20148dc](/obj/item/stock_parts/radio/transmitter/basic/buildable), callBack.delegate:/obj/item/stock_parts/radio/transmitter/proc/transmit(), source: code/game/machinery/_machines_base/stock_parts/radio/transmitter.dm:14Prev: NULL, Next: NULL, SPENT(2322.99), QDELETED, NO CALLBACK world.time: 2323.7, head_offset: 2313.76, practical_offset: 1 in code/controllers/subsystems/timer.dm at line 282 src: Timer usr: .`

the heck is this it ruined checks

Mazianni:
random invalid timer error, suggest pushing something to make CI run again and it should clear up.

MistakeNot4892:
I don't quite understand the purpose of this, and I'm not sure that I like the amount of duplicated code and additional gas handling and whatnot required - what is the benefit of this system compared to using plasma temperature from a central RUST? Even 'realistically' it doesn't make a huge amount of sense, igniting fusion is the expensive part so doing it every time you want thrust would be incredibly wasteful.

Not blocking this outright, but would like some more details on intent/goals.

ghost:
well i dont know how to implement this other way than that it currently is, so-o.
the goal is to:
1) change the current version magic (welp, i just dont like them, though it can be fun to add some plasma conduits to collect plasma from tokamak directly into these)
2) high-tier version of engines existing 

this engine is very space-requirement efficient in the means of `exhaust velocity` per `turf wasted on thruster installation`
ions > gas > fusion
it will allow for construction of big shuttles/ships which are still capable of fast space travel at the same time using low amount of space on thruster installations
@MistakeNot4892

afterthought2:
I feel like if ship speed is actually a problem, it can be resolved by revising the constants in the formula. Feels weird how it's entirely independent of the RUST proper.

MistakeNot4892:
I think I am going to downthumb this. I don't think it makes conceptual sense to reimplement so much fusion logic in each individual engine when it's perfectly possible to just use the RUST as a heat source for a chemical thruster, and codewise duplicating it all makes it harder to maintain and easier to have bugs/inconsistencies creep in. 

:-1:

ghost:
probably closing this since no necessity right now and hydrogen fusion got removed

PR #1737, opened 2021-06-10
Trays now use vis_contents

MarinaGryphon:
- Trays now use vis_contents, so you can interact with (pick up, eat, add reagents to) tray objects by clicking on them.
  - Trays no longer drop their contents when set on a table.
  - Storage UIs now ignore pixel_x and pixel_y when displaying items.

### DIFF
--- a/code/game/objects/items/weapons/storage/trays.dm
+++ b/code/game/objects/items/weapons/storage/trays.dm
@@ -18,6 +18,7 @@
 	use_sound = null
 
 	var/cooldown = 0	//Cooldown for banging the tray with a rolling pin. based on world.time. very silly
+	var/no_drop = FALSE
 
 	material = /decl/material/solid/cardboard
 	applies_material_colour = TRUE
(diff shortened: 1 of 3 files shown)


code/game/objects/items/weapons/storage/trays.dm
```diff
@@ -64,14 +58,44 @@
...
+		W.pixel_x = rand(-W.randpixel, W.randpixel)
+		W.pixel_y = rand(-W.randpixel, W.randpixel)
+		W.pixel_z = 0
+		return
+
+	if (!click_params)
+		return
+
+	var/list/click_data = params2list(click_params)
+	if (!click_data["icon-x"] || !click_data["icon-y"])
```
MistakeNot4892, line 84: Probably want isnull here as afaik 0 is a valid icon position.

MistakeNot4892: Nvm it's a string, disregard.

MistakeNot4892 (approved):

PR #1739, opened 2021-06-14
Window polarization tweaks

SierraKomodo:
:cl: SierraKomodo
tweak: There is now a user feedback message for wiring a window.
rscadd: You can now de-polarize a window by using wirecutters to remove the wiring.
tweak: The set id prompt when using multitools on polarized windows now defaults to the current id instead of null.
rscadd: Using a multitool on an anchored (fully installed) window now toggles the tint instead of changing the ID. To change the ID, unanchor the window first.
tweak: Examining windows now tells you if they are anchored and/or polarized, and the construction state (screwdriver/crowbar steps) of reinforced windows.
/:cl:

### DIFF
--- a/code/game/objects/structures/window.dm
+++ b/code/game/objects/structures/window.dm
@@ -219,20 +219,40 @@
 			playsound(src.loc, 'sound/items/Ratchet.ogg', 75, 1)
 			visible_message(SPAN_NOTICE("[user] dismantles \the [src]."))
 			dismantle()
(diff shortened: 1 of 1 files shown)


comma (dismissed):

SierraKomodo:
Added construction state feedback on examine for reinforced windows, as requested on Bay's version of the PR.

comma (approved):

PR #1742, opened 2021-06-15
FTL Shunt Tweaks

Mazianni:
This has become a moderately comprehensive overhaul:
-The FTL shunt now accumlates charge over time. Target level is customizable by user, and so is the charge rate (capped at 5MW - may make it upgradable or add seperate capacitor machines.)
-Rewrote the fuel consumption proc ENTIRELY so fuel consumption is equal between all fuel ports.
-Reorganized the FTL computer's nanoui a fair amount.
-FTL drive / dangerous jump distance is now configurable.
-adds a 'jump plotting' mechanic to the FTL computer, time depending jump distance.
-improves fluff around the fuel consumption stuff.


code/modules/overmap/ftl_shunt/computer.dm
```diff
@@ -78,45 +146,49 @@
 	if(href_list["set_shunt_x"] || href_list["set_shunt_y"])
 		if(!linked_core)
 			return TOPIC_NOACTION
+		if(plotting_jump)
+			to_chat(user, SPAN_WARNING("Unable to alter programmed coordinates: Plotting in progress."))
+			return TOPIC_NOACTION
```
afterthought2: Usually refresh in cases when the UI seems to have been wrong.

code/modules/overmap/ftl_shunt/computer.dm
```diff
@@ -23,6 +28,57 @@
...
+	var/delay
+	switch(jump_dist)
+		if(1 to linked_core.safe_jump_distance)
+			plot_delay_mult = 1
+		if(linked_core.safe_jump_distance to linked_core.moderate_jump_distance)
+			plot_delay_mult = 1.5
+		if(linked_core.moderate_jump_distance to INFINITY)
+			plot_delay_mult = 2
+
+	delay = (jump_dist * BASE_PLOT_TIME_PER_TILE) * plot_delay_mult
```
afterthought2: It's possible this is 0 or something else due to how `get_dist` actually works. Should probably clamp either jump_dist or this.

code/modules/overmap/ftl_shunt/computer.dm
```diff
@@ -146,3 +218,46 @@
...
+	if(href_list["adjust_power"])
+		var/input_power
+		input_power = input(user, "Enter allowed power input (in kilowatts)", "FTL Computer", (linked_core.allowed_power_usage / 1000)) as num|null
+		if(!input_power)
+			return
+
+		if(!CanInteract(user, state))
+			return TOPIC_NOACTION
+
+		input_power = input_power * 1000
```
afterthought2: `input_power *= 1000`. There is also a KILOWATTS define.

code/modules/overmap/ftl_shunt/computer.dm
```diff
@@ -146,3 +218,46 @@
...
+	if(href_list["adjust_target"])
+		var/target_joules
+		target_joules = input(user, "Enter desired capacitor charge level (in megajoules)", "FTL Computer", (linked_core.target_charge / 1000)) as num|null
+		if(!target_joules)
+			return
+
+		if(!CanInteract(user, state))
+			return TOPIC_NOACTION
+
+		target_joules = target_joules * 1000
```
afterthought2: Same here.

code/modules/overmap/ftl_shunt/computer.dm
```diff
@@ -146,3 +218,46 @@
...
+		return TOPIC_REFRESH
+
+	if(href_list["toggle_charge"])
+		linked_core.charging = !linked_core.charging
+
+	if(href_list["plot_jump"])
+		if(jump_plot_timer)
+			return // already plotting a jump.
+		var/jd = plot_jump(to_plot_x, to_plot_y)
+		ping("Jump plotting initiated, ETA [jd/20] seconds.")
```
afterthought2: 20?

afterthought2 (commented):
So there's a simple test for whether a machine should use the `/power` subtype: "does it supply power to a powernet?" If the answer is no, you are usually better off using the terminal power component instead.

code/modules/overmap/ftl_shunt/computer.dm
Mazianni: magic numbered until it seemed accurate enough

Mazianni:
random nullspace wire. OK.

code/modules/overmap/ftl_shunt/computer.dm
afterthought2: But isn't `jd` literally the delay in deciseconds until the timer fires?

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -54,6 +63,10 @@
 		local_network.set_tag(null, initial_id_tag)
 	find_ports()
 	set_light(2)
+	target_charge = max_charge * 0.25 //Target charge set to a quarter of our maximum charge, just for weirdness prevention
+	var/obj/item/stock_parts/power/terminal/term = get_component_of_type(/obj/item/stock_parts/power/terminal)
+	if(!term.terminal)
+		term.make_terminal(src)
```
afterthought2: This needs to be gated behind if `populate_parts` is true.

code/modules/overmap/ftl_shunt/core.dm
```diff
@@ -440,38 +463,89 @@
...
+	if(accumulated_charge >= required_charge)
+		return TRUE
+
+/obj/machinery/ftl_shunt/core/proc/draw_charge(var/input)
+	if(stat & NOPOWER)
+		return FALSE
+
+	var/obj/item/stock_parts/power/terminal/term = get_component_of_type(/obj/item/stock_parts/power/terminal)
+
+	var/drawn_charge = term.terminal.draw_power(input)
```
afterthought2: use `use_power_oneoff`.

code/modules/overmap/ftl_shunt/computer.dm
```diff
@@ -146,3 +218,46 @@
...
+		return TOPIC_REFRESH
+
+	if(href_list["toggle_charge"])
+		linked_core.charging = !linked_core.charging
+
+	if(href_list["plot_jump"])
+		if(jump_plot_timer)
+			return // already plotting a jump.
+		var/jd = plot_jump(to_plot_x, to_plot_y)
+		ping("Jump plotting initiated, ETA [jd/10] seconds.")
```
afterthought2: `round(jd / (1 SECOND))` or similar

afterthought2 (approved):

Issue #1358, opened 2021-02-17
Accessory Icons


ghost:
#### Description of issue
On recent playtest, there was multiple reports of clothing accessory icon issues
![image](https://user-images.githubusercontent.com/30006750/108210503-ae3ad600-713c-11eb-8671-bf20f061df37.png)
*^ Accessory not displaying on uniform*
![image](https://user-images.githubusercontent.com/30006750/108210564-c01c7900-713c-11eb-940a-2d876434dfa6.png)
*^ Shoulder pads using inventory icon (why would they?)*
Also, one report of engineering shoulder stripe not displaying on *inventory* uniform icon (no screenshot)

Not sure about all these, needs to be re-checked

#### Difference between expected and actual behavior
Proper icons for accessories / Not proper

#### Steps to reproduce
Create different clothing and accessory
Try to combine, wear them in different ways (especially described ones)

#### Specific information for locating

Clothing and accessories

#### Length of time in which bug has been known to occur

Very Recently

#### Client version, Server revision & Game ID

https://github.com/infinitystation/Nebula/tree/50439ae5d5d062a21cc4b61ac64d4c32e8839dc4

#### Issue bingo

- [x] Issue could be reproduced at least once
- [x] Issue could be reproduced by different players
- [ ] Issue could be reproduced in multiple rounds
- [x] Issue happened in a recent (less than 7 days ago) round
- [x] [Couldn't find an existing issue about this](https://github.com/NebulaSS13/Nebula/issues)

MistakeNot4892:
Will look into this over the weekend.

MistakeNot4892:
Should be partially addressed by https://github.com/NebulaSS13/Nebula/pull/1488

MistakeNot4892:
Accessory offsets are still busted for non-humans currently.
EDIT: Fixed in the above PR now.

MistakeNot4892:
Ok so going forward, as of #1772, the pattern will be that items do not render an inventory/world overlay on the item unless a `world-attached` or `inventory-attached` is defined in the accessory's icon file. Not rendering on the holding item is going to be the default behavior, rather than a bug.

PR #1750, opened 2021-06-23
Upstream bug hunt

Naxodile:
Cherry-picked bug fixes from downstream playtesting. 
- Mainframes had the wrong icon state names.
- Igniters would start processing from machinery's init regardless of their own state
- Metal tables had stone table sprites
- Repair toolbox had an invalid onmob state
- It was possible that an object could have enough momentum to pin, but not enough to pass the embed check
- Inflatables didn't block airflow
- Light armor plate didn't have an icon, generic armguards didn't have an onmob (#1358)
- No density rails didn't have the right sprites for internal corners
- Default mode loading had a null separator, causing single letter gamemodes
- Suit cyclers had a few bugs with people hiding / dying inside them
- Minor typos


MistakeNot4892:
Bless you for this PR.
Inflatables specifically are fixed on `stable` - bit of an ask, but if possible could this be pointed at stable rather than dev?

MistakeNot4892 (approved):

PR #1761, opened 2021-06-24
Ports and updates aspect system from Europa.

### LINKED ISSUE #1728: Saving a character slot causes the preview to go blank.
#### Description of issue
If you click 'Save slot' in chargen, the character preview on the right side will become blank.

The preview should not become blank.

#### Steps to reproduce
Click 'save slot'.

Staging branch during test today.

Unclear.

Client Version: 513
Server Revision: 4a00fd7bbc4709ae91fb12b265e0de8fb60dec69 - staging - 2021-06-05
Game ID: cco-awGq
Current map: NSS Exodus

#### Issue bingo

MistakeNot4892:
:cl:
rscadd: Ports the aspect system from Europa, specifically the backend, not the aspects themselves. This is a menu in character preferences that allows you to pick a series of traits for your character, as in some RPGs.
tweak: Moves prosthetic limbs, prosthetic organs, amputated limbs and nearsightedness onto aspects instead of the Physical tab.
tweak: Moves preview options from the Physical tab to the header of character preferences.
/:cl:

Fixes #1728.


MistakeNot4892:
image

MistakeNot4892:
This is all working except for the preview icon, which is not reliably updating.

MistakeNot4892:
Ready for review. I'd like to implement the aspects for utility frames etc in a separate PR.

MistakeNot4892:
Colour blindness now obfuscates wire colours as well:
image

MistakeNot4892:
Thanks Github.

code/__defines/aspects.dm
```diff
@@ -0,0 +1,70 @@
...
+#define ASPECTS_PHYSICAL  BITFLAG(0)
+#define ASPECTS_MENTAL    BITFLAG(1)
+
+#define DEFINE_ROBOLIMB_MODEL_ASPECTS(MODEL_PATH, MODEL_ID, MODEL_NAME, COST) \
+/decl/aspect/prosthetic_limb/left_hand/##MODEL_ID {                           \
+	name = MODEL_NAME + " Left Hand";                                         \
+	model = MODEL_PATH;                                                       \
+	parent = /decl/aspect/prosthetic_limb/left_hand;                          \
+	aspect_cost = COST;                                                       \
+	base_type = /decl/aspect/prosthetic_limb/left_hand;                       \
```
afterthought2, line 20: Can't this and parent be inherited?

code/modules/aspects/aspects.dm
```diff
@@ -0,0 +1,177 @@
...
+
+	if(category)
+		var/datum/aspect_category/AC = global.aspect_categories[category]
+		if(!istype(AC))
+			AC = new(category)
+			global.aspect_categories[category] = AC
+		AC.aspects += src
+		if(AC.hide_from_chargen && available_at_chargen)
+			AC.hide_from_chargen = FALSE
+		//global.aspect_icons += use_icon
```
afterthought2: commented code

code/modules/aspects/aspects.dm
```diff
@@ -0,0 +1,177 @@
...
+
+	if(parent)
+		if(ispath(parent))
+			parent = GET_DECL(parent)
```
afterthought2: I believe if you do this you need testing to make sure the parent relation is tree-like (no loops, no self-parenting). May also be worth considering decl/hierarchy, which may do some of this for you.

code/modules/client/preference_setup/general/03_aspects.dm
```diff
@@ -0,0 +1,178 @@
...
+	pref.aspects = R.read("aspects")
+	var/list/all_aspects = decls_repository.get_decls_of_subtype(/decl/aspect)
+	for(var/aspect in pref.aspects)
+		if(ispath(aspect, /decl/aspect))
+			continue
+		pref.aspects -= aspect
+		for(var/aspect_type in all_aspects)
+			var/decl/aspect/aspect_decl = all_aspects[aspect_type]
+			if(aspect_decl.name == aspect)
+				pref.aspects |= aspect_type
```
afterthought2, line 34: May be worth storing them by type; otherwise there may be name clashes and you need to test for that.

code/modules/client/preference_setup/general/03_aspects.dm
```diff
@@ -0,0 +1,178 @@
...
+						A = aspects_to_remove[1]
+						aspects_to_remove -= A
+						if(!(A.name in pref.aspects))
+							continue
+						total_aspect_cost += A.aspect_cost
+						remove_aspects += A.name
+						if(A.children)
+							aspects_to_remove |= A.children
+					if(get_aspect_total() - total_aspect_cost <= config.max_character_aspects)
+						pref.aspects -= remove_aspects
```
afterthought2: Possibly worth pruning again here; I guess above code makes sure "sub"aspects are removed, but all the rest of your sanitization seems completely independent of any structural assumptions on which combinations of aspects are valid.

afterthought2 (commented):
Need to sanitize aspects on species change in prefs (unless I missed that), as well as any other prefs change that might affect aspects. Possibly also need to update aspects in `set_species` somehow if not already doing that.

code/modules/client/preference_setup/general/03_aspects.dm
MistakeNot4892: I can't store them by type unless I want to start using text2path(). I'll add a unit test for uniqueness.

code/modules/aspects/aspects.dm
MistakeNot4892: I'm a bit sad about the commented code, the little icons looked nice but caused crippling multi second lag. :(

code/__defines/aspects.dm
MistakeNot4892: They can, but I wanted to be explicit about it here in case it's ever changed lower down the tree.

code/modules/aspects/aspects.dm
MistakeNot4892: Added a unit test.

code/modules/client/preference_setup/general/03_aspects.dm
MistakeNot4892: Added a unit test.

code/modules/client/preference_setup/general/03_aspects.dm
MistakeNot4892: Removed the odd check here and added a prune call afterwards.

code/modules/aspects/aspects.dm
MistakeNot4892: Removed.

code/modules/client/preference_setup/general/03_aspects.dm
afterthought2: You can do for(thing in decls) if(saved_key == "[decl.type]") as a text2path alternative, potentially, but this also works.

afterthought2 (approved):

PR #1768, opened 2021-06-25
implements a decl-based method for adding onto human/examine()

Mazianni:
implements /decl/human_examination - which is more or less just a stub proc so modpacks can add onto human/examine() without too much fuss.

### DIFF
--- a/code/modules/mob/living/carbon/human/examine.dm
+++ b/code/modules/mob/living/carbon/human/examine.dm
@@ -351,6 +351,14 @@
 	var/show_descs = show_descriptors_to(user)
 	if(show_descs)
 		msg += "<span class='notice'>[jointext(show_descs, "<br>")]</span>"
+
+	var/list/human_examines = decls_repository.get_decls_of_subtype(/decl/human_examination)
+	for(var/exam in human_examines)
+		var/decl/human_examination/HE = human_examines[exam]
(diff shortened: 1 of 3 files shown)


code/modules/mob/living/carbon/human/examine.dm
```diff
@@ -351,6 +351,14 @@
 	var/show_descs = show_descriptors_to(user)
 	if(show_descs)
 		msg += "<span class='notice'>[jointext(show_descs, "<br>")]</span>"
+
+	var/list/human_examines = decls_repository.get_decls_of_subtype(/decl/human_examination)
+	for(var/exam in human_examines)
+		var/decl/human_examination/HE = human_examines[exam]
+		var/adding_text = HE.do_examine(user, distance, src)
+		if(adding_text)
+			msg += adding_text
```
Techhead0, line 360: If you wanna optimize, you could compact this down to:
```
msg += HE.do_examine(user, distance, src) || null
```
But it doesn't really matter. (And the `|| null` is only there to keep it from type-mismatching if a 0 is returned)

MistakeNot4892 (dismissed):

Lohikar (dismissed):
Squash.

MistakeNot4892 (approved):

PR #1769, opened 2021-06-25
Fixes incorrect secret project disks examine, causes stack trace print

quardbreak:


MistakeNot4892:
Please point fixes at stable.

PR #1771, opened 2021-06-26
Adds pear cider

quardbreak:
Just a copypasta of apple cider but with different taste and mix process.


PlayerDeer:
Wouldn't this be called perry, the alcoholic drink?
https://en.wikipedia.org/wiki/Perry

quardbreak:
I thinked about this but decided to use `pear cider` to match with apple variant.

MistakeNot4892 (approved):
It's called pear cider at my local bottle-o so I'm ok with this.

PR #1777, opened 2021-06-26
Various map changes to allow for more control by the map datum.

MistakeNot4892:
Trying to make a variant map and running into various things that are hardcoded or included by default, so I'm moving them to the relevant map files.

DNM, may add more.


code/game/antagonist/station/cultist.dm
```diff
@@ -22,7 +22,7 @@
 /decl/special_role/cultist
 	name = "Cultist"
 	name_plural = "Cultists"
-	blacklisted_jobs = list(/datum/job/ai, /datum/job/cyborg, /datum/job/submap)
```
comma, line 25: maybe whole no-synth thing should be a flag rather than having to add it to every list like this.

MistakeNot4892: Made it use event categories instead.

afterthought2 (approved):

PR #1783, opened 2021-06-27
Ports SSprofiler from Paradise Station

quardbreak:
Ports SS220 (Actually, Paradise one) SSprofiler, provided by @Vallat.
This thing allows to collect profiler's data in JSON format as log entry.

~~Currently don't respect config option because I don't know how to integrate the check in subsystem. This is main reason why I putting it on draft.~~

### DIFF
--- a/code/__defines/_compile_options.dm
+++ b/code/__defines/_compile_options.dm
@@ -7,3 +7,8 @@
 #if DM_VERSION < REQUIRED_DM_VERSION
 #warn Nebula is not tested on BYOND versions older than 513. The code may not compile, and if it does compile it may have severe problems.
 #endif
+
+// Log the full sendmaps profile on 514.1556+, any earlier and we get bugs or it not existing.
+#if DM_VERSION >= 514 && DM_BUILD >= 1556
+#define SENDMAPS_PROFILE
+#endif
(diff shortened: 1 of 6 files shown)


Lohikar:
I don't really see what the utility of this is.

afterthought2:
Might be better for downstream use if performance or debugging is a particular downstream concern. It's completely modular, so will be easy to add there. On the other hand, we probably don't have much use for it upstream, and we definitely don't want to recommend/encourage downstreams to use it in general (it has some performance impact, requires expertise to use the collected data, produces a lot of log volume, etc.)

If others feel like this is useful upstream, then that changes things. @MistakeNot4892, I guess, as you do most of the testing. Can also check in with Outreach.

quardbreak:
I don't think it's modular. It requires make defines and update it every time when someone adds new subsystem define. Also, it uses config, which is not modular.

MistakeNot4892:
I would not personally get much use out of this for testing. The inbuilt profiler is fine for what I've needed to profile in the past. The sendmaps stuff might be useful but I'm ambivalent about the rest.

MistakeNot4892:
Maint vote was not in favour and the argument for adding it wasn't compelling/relevant to our general usage.

PR #1790, opened 2021-06-29
Adds modpack list verb

quardbreak:
изображение

### DIFF
--- a/mods/_modpack.dm
+++ b/mods/_modpack.dm
@@ -1,16 +1,21 @@
 /decl/modpack
-	var/name                         // A string name for the modpack. Used for looking up other modpacks in init.
+	/// A string name for the modpack. Used for looking up other modpacks in init.
+	var/name
+	/// A string desc for the modpack. Can be used for modpack verb list as description.
+	var/desc
+	/// A string with authors of this modpack.
+	var/author
 
-	var/list/dreams                  // A list of strings to be added to the random dream proc.
+	var/list/dreams //! A list of strings to be added to the random dream proc.
 
-	var/list/credits_other           // A list of strings that are used by the end of round credits roll.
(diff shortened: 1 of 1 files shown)


mods/_modpack.dm
```diff
@@ -1,16 +1,30 @@
...
+	var/list/credits_crew_names
+	/// A list of strings that are used by the end of round credits roll.
+	var/list/credits_holidays
+	/// A list of strings that are used by the end of round credits roll.
+	var/list/credits_adjectives
+	/// A list of strings that are used by the end of round credits roll.
+	var/list/credits_crew_outcomes
+	/// A list of strings that are used by the end of round credits roll.
+	var/list/credits_topics
+	/// A list of strings that are used by the end of round credits roll.
```
MistakeNot4892: Why make the comments worse? :(

mods/_modpack.dm
```diff
@@ -41,3 +55,33 @@
...
+						. += "<br>Description: [M.desc]"
+					if(M.author)
+						. += "<br><i>Author: [M.author]</i>"
+				. += "</div><br>"
+
+		var/datum/browser/popup = new(mob, "modpacks_list", "Modpacks List", 480, 580)
+		popup.set_content(.)
+		popup.open()
+	else
+		to_chat(src, SPAN_WARNING("This server doesn't have any available modpacks."))
```
MistakeNot4892: 'Does not include any modpacks' would read better imo.

mods/_modpack.dm
quardbreak: I did it to respect dmdoc :^(

MistakeNot4892: As Lohi said you can use `// !` instead of making it untidy.

MistakeNot4892 (approved):

PR #1799, opened 2021-07-01
Ports composite sounds from Aurora.

MarinaGryphon:
Composite sounds are sounds formed from various components:
- A start sound
- Middle sounds (potentially chosen via pickweight), repeated until the set time is elapsed
- An ending sound
This is useful for things like fryer noises, microwave noises, a supermatter hum, shower noises, etc.

### DIFF
--- a/code/datums/composite_sounds/_composite_sound.dm
+++ b/code/datums/composite_sounds/_composite_sound.dm
@@ -1,0 +1,90 @@
+/*
+	output_atoms	(list of atoms)			The destination(s) for the sounds
+	mid_sounds		(list or soundfile)		Since this can be either a list or a single soundfile you can have random sounds. May contain further lists but must contain a soundfile at the end.
+	mid_length		(num)					The length to wait between playing mid_sounds
+	start_sound		(soundfile)				Played before starting the mid_sounds loop
+	start_length	(num)					How long to wait before starting the main loop after playing start_sound
+	end_sound		(soundfile)				The sound played after the main loop has concluded
+	chance			(num)					Chance per loop to play a mid_sound
+	volume			(num)					Sound output volume
+	max_loops		(num)					The max amount of loops to run for.
+	direct			(bool)					If true plays directly to provided atoms instead of from them
+*/
+
+/datum/composite_sound
+	var/list/atom/output_atoms
+	var/mid_sounds
+	var/mid_length
+	var/start_sound
+	var/start_length
+	var/end_sound
+	var/chance
+	var/volume = 100
+	var/max_loops
+	var/direct
+
+	var/timerid
+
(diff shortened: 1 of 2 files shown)


code/datums/composite_sounds/_composite_sound.dm
```diff
@@ -0,0 +1,90 @@
...
+	if(max_loops && (world.time >= starttime + mid_length * max_loops))
+		stop()
+		return
+	if(!chance || prob(chance))
+		play(get_sound(starttime))
+	if(!timerid)
+		timerid = addtimer(CALLBACK(src, .proc/sound_loop, world.time), mid_length, TIMER_CLIENT_TIME | TIMER_STOPPABLE | TIMER_LOOP)
+
+/datum/composite_sound/proc/play(soundfile)
+	var/list/atoms_cache = output_atoms
```
MistakeNot4892: What's the purpose of having a second reference to output_atoms? Afaict this doesn't do anything in this context.

code/datums/composite_sounds/_composite_sound.dm
```diff
@@ -0,0 +1,90 @@
...
+		start()
+
+/datum/composite_sound/Destroy()
+	stop()
+	output_atoms = null
+	return ..()
+
+/datum/composite_sound/proc/start(atom/add_thing)
+	if(add_thing)
+		output_atoms |= add_thing
```
MistakeNot4892: LAZYDISTINCTADD possibly, this will runtime if the sound was created with no output atoms.

code/datums/composite_sounds/_composite_sound.dm
```diff
@@ -0,0 +1,90 @@
...
+	if(timerid)
+		return
+	on_start()
+
+/datum/composite_sound/proc/stop(atom/remove_thing)
+	if(remove_thing)
+		output_atoms -= remove_thing
```
MistakeNot4892: Likewise, LAZYREMOVE

code/datums/composite_sounds/_composite_sound.dm
```diff
@@ -0,0 +1,90 @@
...
+	if(!timerid)
+		return
+	on_stop()
+	deltimer(timerid)
+	timerid = null
```
MistakeNot4892, line 58: I don't believe it's necessary to null timerid, it's not a reference.

MistakeNot4892: Nvm I see it's used for state checking.

code/datums/composite_sounds/_composite_sound.dm
MarinaGryphon: I think it locally caches it for performance reasons? But it's a bit silly here, it's not particularly hot code.

MistakeNot4892 (approved):

PR #1804, opened 2021-07-01
Facial hair now checks for a bodytype, rather than pronouns

quardbreak:


code/modules/client/preference_setup/general/01_basic.dm
```diff
@@ -105,6 +103,8 @@
 			pref.bodytype = new_body.name
 			if(new_body.associated_gender) // Set to default for male/female to avoid confusing people
 				pref.gender = new_body.associated_gender
+			if(!(pref.f_style in S.get_facial_hair_styles(new_body.associated_gender)))
```
MistakeNot4892, line 106: You need to update every place where get_facial_hair_styles() and get_hair_styles() are called so they use bodytype gender. Doing it piecemeal just means there'll be issues elsewhere.

Koollan: alternatively, we could just have facial hair not locked behind gender regardless, seems a lot simpler that way and more customizable

MistakeNot4892: Yep, but if people later want to have markings restricted by bodytype (sexual dimorphism or variant silhouettes) the system will still be broken.

quardbreak:
Ready for review.

MistakeNot4892 (approved):

PR #1778, opened 2021-06-26
Aurora/Hestia-Inspired Cooking System

MarinaGryphon:
A port of Hestia's cooking system with some backported features from Aurora. No new foods or recipes have been added, but the current recipes have been redone.

TODO before completion:
- [ ] Add new items to kitchen vendor
- [ ] Port ladles and chopsticks
- [ ] Test beating people up with skillets
- [ ] Large test with friends
- [ ] Get around to implementing world/inventory/worn icons for saucepans

## Changelog
:cl: Jade2562
tweak: Completely overhauled recipe system and recipes
soundadd: Random sound loops, used by the fryer and microwave
rscadd: New microwave UI with an animated progress bar
tweak: Completely overhauled the oven and fryer
rscadd: Added the stove, which also heats reagents and has dynamic overlays displaying its contents! Nifty.
rscadd: Recipes that can be done in more than one appliance *without* being duplicated
rscadd: Cooking containers, used in certain appliances to cook things
rscadd: Appliances that can take more than one type of cooking container
rscadd: The ability to beat people with frying pans!
rscadd: The ability to wear saucepans
rscdel: Removed the grill, to be re-implemented at a later date.
imageadd: Added some nice new sprites for appliances and the microwave
imagedel: deleted some icons and images
maptweak: Replaced grills on the map with stoves.
/:cl:

PR #1806, opened 2021-07-01
Microwave Refactor and Recipe Improvements

MarinaGryphon:
Splits out some features from #1778, with improvements.
* NanoUI microwave UI
* Ejecting individual items/reagents
* Bulk recipe creation
* More nuanced reagent handling in recipe creation
* Makes the `check_x` procs on recipes pure
* Removes some global lists
* Caches complexity for recipe decls
* Incorporates number of unique items into recipe complexity calculation
* Composite sounds for microwaves


code/game/machinery/kitchen/microwave.dm
```diff
@@ -395,25 +416,53 @@
...
+	..()
+	var/bin_rating = 0 // 2
+	var/cap_rating = 0 // 3
+	var/las_rating = 0 // 1
+
+	bin_rating = total_component_rating_of_type(/obj/item/stock_parts/matter_bin)
+	cap_rating = total_component_rating_of_type(/obj/item/stock_parts/capacitor)
+	las_rating = total_component_rating_of_type(/obj/item/stock_parts/micro_laser)
+
+	active_power_usage = initial(active_power_usage) - (cap_rating * 25)
```
afterthought2: You are not allowed to modify this var except through the setter.

code/game/machinery/kitchen/microwave.dm
```diff
@@ -153,237 +161,250 @@
...
 
-// Behold: the worst proc name in the codebase.
-/obj/machinery/microwave/proc/wzhzhzh(var/seconds)
-	for (var/i=1 to seconds)
-		if (stat & (NOPOWER|BROKEN))
-			return 0
-		use_power_oneoff(500)
-		sleep(10)
-	return 1
+	use_power_oneoff(active_power_usage)
```
afterthought2: The point of this var is that if you set the power to active, it will _automatically_ do this every tick (in a more performant manner).

code/game/machinery/kitchen/microwave.dm
```diff
@@ -153,237 +161,250 @@
...
+		failed = FALSE
+		cook_time = update_cook_time(round(recipe.time * 2))
+
+	start()
+
+/obj/machinery/microwave/proc/update_cook_time(var/ct = 200)
+	return (ct / cooking_power)
+
+/obj/machinery/microwave/proc/finish_cooking()
+	var/decl/recipe/recipe = src.select_recipe()
```
afterthought2: Why are there `src` everywhere on this proc? Suppressing comments on this elsewhere.

code/game/machinery/kitchen/microwave.dm
```diff
@@ -153,237 +161,250 @@
...
+	if(!recipe)
+		return
+	var/result = recipe.result
+	var/list/cooked_items = list()
+	while(recipe)
+		cooked_items += recipe.make_food(src)
+		recipe = src.select_recipe()
+		if (!recipe || (recipe.result != result))
+			break
```
afterthought2, line 261: Don't really follow: is the point that you continue applying various recipes as long as the result is the same? Why that particular condition (seems arbitrary)?

code/game/machinery/kitchen/microwave.dm
```diff
@@ -153,237 +161,250 @@
...
+
+	//Any leftover reagents are divided amongst the foods
+	var/total = reagents.total_volume
+	for (var/obj/item/chems/food/snacks/S in cooked_items)
+		reagents.trans_to_holder(S.reagents, total/cooked_items.len)
+		S.dropInto(loc) // since eject only ejects ingredients!
```
afterthought2: Is it guaranteed that all things in cooked_items are /snacks? If not, probably still need to move them out.

code/modules/food/recipes_microwave.dm
```diff
@@ -195,6 +180,36 @@ I said no!
 	)
 	result = /obj/item/chems/food/snacks/donkpocket //SPECIAL
 
+/decl/recipe/donkpocket/warm/check_items(obj/container) // NO FUCKING INFINITE LOOPS
```
afterthought2: Seems vastly overcomplicated: can't you just check if there's a non-warm donk packet? You don't gain anything by making it more generic because as you say it's never going to be generic.

code/datums/recipe.dm
```diff
@@ -109,19 +117,104 @@ var/global/list/microwave_accepts_items =    list(
...
+			container.reagents.trans_type_to_holder(buffer, r, reagents[r])
+
+	/*
+	Now we've removed all the ingredients that were used and we have the buffer containing the total of
+	all their reagents.
+	If we have multiple results, holder will be used as a buffer to hold reagents for the result objects.
+	If, as in the most common case, there is only a single result, then it will just be a reference to
+	the single-result's reagents
+	*/
+	var/datum/reagents/holder = new/datum/reagents(1e12, global.temp_reagents_holder)
```
afterthought2: 1e12?

code/game/machinery/kitchen/microwave.dm
MarinaGryphon: So that I can add /proc/select_recipe on a downstream. I mean, I could rename it, but... ehh

code/game/machinery/kitchen/microwave.dm
MarinaGryphon: Because it makes sense to double the recipe to make twice the amount, but trying to cook eggs and bake bread in the same container doesn't make sense.

code/datums/recipe.dm
MarinaGryphon: Large amount.

code/modules/food/recipes_microwave.dm
MarinaGryphon: That would stop you from cooking more than one donk pocket at a time unless I make it eject the results when they're made rather than after.

code/game/machinery/kitchen/microwave.dm
MarinaGryphon: (specifically, the idea is that microwave/select_recipe() will call /proc/select_recipe(MICROWAVE, src) downstream)
Wasn't sure if it would default to the instance or global proc if it's unmarked. If it's the instance, src can be removed.

MistakeNot4892: It defaults to the nearest scope, ie. local in this case.

afterthought2: Not interested in weird hacks you plan to use downstream. The thing you describe works fine with using differently named local/global procs (and is like 10x easier to read), but that's up to you. Upstream, style is to not use src. unless there's an overwhelming functional reason to.

code/modules/food/recipes_microwave.dm
afterthought2: Why? A warm donk packet doesn't prevent cooking, it's just not a valid ingredient. I don't get it.

code/game/machinery/kitchen/microwave.dm
afterthought2: OK.

code/modules/food/recipes_microwave.dm
MarinaGryphon: ... I'm not sure what you're suggesting?

code/datums/recipe.dm
MistakeNot4892: `INFINITY` would be clearer.

MarinaGryphon: Can you actually make a container with infinite reagents? I mean, I guess you *can...*

MistakeNot4892: It's as valid as making one with `1e12` reagents for all practical purposes.

code/game/machinery/kitchen/microwave.dm
```diff
@@ -27,114 +38,111 @@
 /obj/machinery/microwave/Initialize()
 	. = ..()
 	create_reagents(100)
+	soundloop = new(list(src), FALSE)
```
afterthought2, line 41: Not seeing where this reference is cleared in `Destroy`.

MarinaGryphon: d'oh

MarinaGryphon:
```suggestion
	soundloop = new(list(src), FALSE)

/obj/machinery/microwave/Destroy()
	dispose()
	QDEL_NULL(soundloop)
	return ..()
```
This look good to you?

afterthought2 (approved):

PR #1812, opened 2021-07-02
Visual-novel style message box

ghost:
No portraits yet! 
Adds a proc for simple on-screen queued message box (this means if the box already exists, the message will be queued just like in real VN), also tweaked maptext so it dont cover third part of window, reserved for portraits.

Currently used by show_location_blurb() and in hear_say().
hear_say() will open this window every time player *actually* heard something (well it describes itself) if the player turned the window option on in preferences (disabled by default)
*note: pref currently affects only hear_say*

Delays for the window can be easily tweaked on the `/obj/screen/text/message`
also: the point of *deleting box after usage* is in possible support for multiple message box windows in future
currently if you supply proc with the box type which already exists in player screen, it will delete old box upon message delivery and instantly open the new one, so no worries here (you can test this by saying literally anything after roundstart: show_location_blurb() does not uses default box type)

also i hope `QDEL_LIST(whatever_list_of_callbacks)` is proper for deleting callbacks?

Preview:
pp11sujYqa

questionable:
if there is no `speaker_name` supplied on the default box the text will start drawing out of box (since <br> has nothing before it, idk how to fix it) - but right now it's impossible since this messagebox type is used only when humans talk (hear_say, and it'll always put Unknown where if there is no name)
image


MistakeNot4892:
I would like to have a play with this locally and see how it feels/works, how it could be integrated into things, etc. this weekend, before considering a merge, so I am going to slap DNM on it for now.

Kurgis:
Is `text_delay` alterable for specific messages? Lets say, I want the `text_delay` in one message from hear_say() to be 0.5, but others to still use 1.

ghost:
> Is `text_delay` alterable for specific messages? Lets say, I want the `text_delay` in one message from hear_say() to be 0.5, but others to still use 1.
@Kurgis 

currently no, but i dont think that would be necessary
it can be added as an argument for the proc itself
if you need to change it you need to define another /obj/screen/text/message
sadly window changing is handled very crappy right now

MistakeNot4892:
I backed this up in a branch: https://github.com/MistakeNot4892/Nebula/tree/vnchat
Hope you're doing alright Coltrane.

PR #1822, opened 2021-07-03
SSinput

quardbreak:
~~Ports SS220 keybindings. They're can not match with current keybindings since SS220 uses Paradise's layout.
This is initial port towards configurable keybindings. This PR doesn't have it yet, but it would be simpler to mess with them if we got input system.~~

## SSinput
- Added new subsystem which will control pressed keyloop and macro grant.

## Keybinding system
- tgstation keybinding system allow to make bind and configure it in setup window.

## Diagonal movement
- Added support for diagonal movement with better diagonal gliding. Off by default.

## ~~Breaks fps settings into client and world config option.~~
not in this PR anymore


code/__defines/misc.dm
```diff
@@ -271,4 +271,14 @@
 #define AS_ANYTHING as()
 #else
 #define AS_ANYTHING as anything
-#endif
\ No newline at end of file
+#endif
+
+//intent defines
+#define INTENT_HELP "help"
```
MistakeNot4892: We have these already, they're under I_HELP etc.

code/modules/mob/living/carbon/human/inventory.dm
```diff
@@ -2,10 +2,12 @@
 Add fingerprints to items when we put them in our hands.
 This saves us from having to call add_fingerprint() any time something is put in a human's hands programmatically.
 */
-/mob/living/carbon/human/verb/quick_equip()
+/mob/verb/quick_equip()
 	set name = "quick-equip"
 	set hidden = 1
+	return
```
MistakeNot4892: What is this verb for?

code/modules/mob/mob.dm
```diff
@@ -215,6 +216,7 @@
 	return log(2, mob_size / MOB_SIZE_MEDIUM)
 
 /mob/proc/Life()
+	set waitfor = FALSE
```
MistakeNot4892: This seems like a very strange addition considering the next line should stop it sleeping.

code/modules/keybindings/bindings_mob.dm
```diff
@@ -0,0 +1,77 @@
+// Technically the client argument is unncessary here since that SHOULD be src.client but let's not assume things
+// All it takes is one badmin setting their focus to someone else's client to mess things up
+// Or we can have NPC's send actual keypresses and detect that by seeing no client
+
+/mob/key_down(_key, client/user)
```
MistakeNot4892: Noting here that we should convert all of this to a decl system after this PR, I know you're splitting them up but it's worth mentioning briefly.

code/datums/movement/movement.dm
```diff
@@ -139,6 +139,7 @@ if(LAZYLEN(movement_handlers) && ispath(movement_handlers[1])) { \
 	. = ..()
 
 /datum/movement_handler/proc/DoMove(var/direction, var/mob/mover, var/is_external)
+	set waitfor = FALSE
```
MistakeNot4892: As mentioned in Discord I'm curious what the reason for this is.

code/controllers/subsystems/input.dm
```diff
@@ -0,0 +1,123 @@
...
+	var/static/list/azerty_movement_keys = list(
+		"Z" = NORTH, "Q" = WEST, "S" = SOUTH, "D" = EAST,				// WASD
+		"North" = NORTH, "West" = WEST, "South" = SOUTH, "East" = EAST,	// Arrow keys & Numpad
+	)
+	movement_keys = default_movement_keys.Copy()
+	alt_movement_keys = azerty_movement_keys.Copy()
+
+// Badmins just wanna have fun ♪
+/datum/controller/subsystem/input/proc/refresh_client_macro_sets()
+	var/list/clients = global.clients
```
MistakeNot4892: Seems like this loop and below could be replaced by `for(var/client/C)` with none of the supporting code.

code/modules/keybindings/setup.dm
```diff
@@ -0,0 +1,61 @@
...
+			var/list/split_name = splittext(macro_set[k], ".")
+			var/macro_name = "[split_name[1]].[split_name[2]]" // [3] is "command"
+			erase_output = "[erase_output];[macro_name].parent=null"
+	winset(src, null, erase_output)
+
+/client/proc/set_macros()
+	set waitfor = FALSE
+
+	//Reset and populate the rolling buffer
+	keys_held.Cut()
```
MistakeNot4892: `keys_held = new(HELD_KEY_BUFFER_LENGTH)`

code/modules/keybindings/bindings_atom.dm
```diff
@@ -0,0 +1,37 @@
...
+			movement_dir &= ~(NORTH|SOUTH)
+		if((movement_dir & EAST) && (movement_dir & WEST))
+			movement_dir &= ~(EAST|WEST)
+		if(!config.allow_diagonal_movement)
+			movement_dir = (movement_dir & NORTH) || (movement_dir & SOUTH) || (movement_dir & WEST) || (movement_dir & EAST)
+		if(!movement_dir)
+			return
+		user.Move(get_step(loc, movement_dir), movement_dir)
+
+/atom/movable/key_down(key, client/user)
```
afterthought2: As I mentioned elsewhere, this override does not make sense when used alongside the above looping proc. You're going to try moving twice, and not in the correct direction.

code/modules/keybindings/bindings_client.dm
```diff
@@ -0,0 +1,112 @@
+// Clients aren't datums so we have to define these procs indpendently.
+// These verbs are called for all key press and release events
+/client/verb/keyDown(_key as text)
+	set instant = TRUE
```
afterthought2: Not a huge fan of how this makes non-instant commands instant.

code/modules/keybindings/bindings_client.dm
```diff
@@ -0,0 +1,112 @@
...
+	//don't want to kick people if a lag spike results in a huge flood of commands being sent
+	if(cache >= MAX_KEYPRESS_AUTOKICK)
+		if(!keysend_tripped)
+			keysend_tripped = TRUE
+			next_keysend_trip_reset = world.time + (2 SECONDS)
+		else
+			log_admin("Client [ckey] was just autokicked for flooding keysends; likely abuse but potentially lagspike.")
+			message_admins("Client [ckey] was just autokicked for flooding keysends; likely abuse but potentially lagspike.")
+			qdel(src)
+			return
```
afterthought2: We have systems for this; either replace or leave alone.

code/modules/keybindings/bindings_client.dm
```diff
@@ -0,0 +1,112 @@
...
+		mob.focus.key_down(_key, src)
+
+/client/verb/keyUp(_key as text)
+	set instant = TRUE
+	set hidden = TRUE
+	//Can't just do a remove because it would alter the length of the rolling buffer, instead search for the key then null it out if it exists
+	for(var/i in 1 to HELD_KEY_BUFFER_LENGTH)
+		if(keys_held[i] == _key)
+			keys_held[i] = null
+			break
```
afterthought2: This needs to handle multiple instances of the same key in the ring buffer, as the KeyDown code does not. I get that that's an "error," but I'm almost certain you can produce it by manipulating client connections.

code/modules/keybindings/bindings_atom.dm
```diff
@@ -0,0 +1,37 @@
...
+/atom/movable/keyLoop(client/user)
+	if(!user.keys_held["Ctrl"] && isturf(loc))
+		var/movement_dir = null
+		var/list/movement = SSinput.movement_keys
+		for(var/_key in user.keys_held)
+			movement_dir |= movement[_key]
+		if(user.next_move_dir_add)
+			movement_dir |= user.next_move_dir_add
+		if(user.next_move_dir_sub)
+			movement_dir &= ~user.next_move_dir_sub
```
afterthought2: I don't get what this does. Why is using held keys not sufficient (note that these are tracking recently pressed held keys)?

code/modules/keybindings/bindings_client.dm
```diff
@@ -0,0 +1,112 @@
...
+	var/movement = SSinput.movement_keys[_key]
+	if(!(next_move_dir_add & movement))
+		next_move_dir_sub |= movement
```
afterthought2: Again, don't get what this is doing.

code/modules/keybindings/focus.dm
```diff
@@ -0,0 +1,9 @@
+/mob
+	/// What receives our keyboard inputs. `src` by default.
+	var/datum/focus
+
+/mob/proc/set_focus(datum/new_focus)
+	if(focus == new_focus)
+		return
+	focus = new_focus
+	reset_view(focus) //Maybe this should be done manually? You figure it out, reader
```
afterthought2: I think this (and the whole focus setting thing) is nonsensical on neb.

code/modules/keybindings/readme.dm
```diff
@@ -0,0 +1,42 @@
...
+and `Southwest()`.
+
+The basic premise of this system is to not screw with .dmf macro setup at all and handle
+pressing those keys in the code instead. We have every key call `client.keyDown()`
+or `client.keyUp()` with the pressed key as an argument. Certain keys get processed
+directly by the client because they should be doable at any time, then we call
+`keyDown()` or `keyUp()` on the client's holder and the client's mob's focus.
+By default `mob.focus` is the mob itself, but you can set it to any datum to give control of a
+client's keypresses to another object. This would be a good way to handle a menu or driving
+a mech. You can also set it to null to disregard input from a certain user.
```
afterthought2: I think you should (A) not do this at all or (B) use existing systems (relaymove, etc) or (C) track it on client, not mob (roughly in order of preference).

code/modules/keybindings/readme.dm
```diff
@@ -0,0 +1,42 @@
...
+
+Movement is handled by having each client call `client.keyLoop()` every game tick.
+As above, this calls holder and `focus.keyLoop()`. `atom/movable/keyLoop()` handles movement
+Try to keep the calculations in this proc light. It runs every tick for every client after all!
```
afterthought2: As I comment elsewhere, this seems inaccurate: you are also moving directly on KeyDown.

code/modules/mob/mob_movement.dm
```diff
@@ -148,7 +148,10 @@
 		return
 	if(!mob)
 		return // Moved here to avoid nullrefs below
-	return mob.SelfMove(direction)
+	. = mob.SelfMove(direction)
+	if(.)
+		next_move_dir_add = 0
+		next_move_dir_sub = 0
```
afterthought2, line 158: I don't think this does anything.

code/modules/mob/new_player/login.dm
```diff
@@ -17,6 +17,9 @@
 
 	show_lobby_menu()
 
+	if(client && SSinput.initialized)
+		client.set_macros()
```
afterthought2: Why isn't this in client/New? What happens if the SS isn't initialized? Feels like it should just go in the prefs thing (if anywhere).

code/datums/movement/atom_movable.dm
```diff
@@ -8,6 +8,7 @@
 
 // Movement relay
 /datum/movement_handler/move_relay/DoMove(var/direction, var/mover)
+	set waitfor = FALSE
```
afterthought2: This will break various movement-related stuff, like falling. Please don't do this.

code/game/verbs/ooc.dm
```diff
@@ -1,12 +1,16 @@
-/client/verb/ooc(message as text)
+/client/verb/ooc(message = "" as text)
 	set name = "OOC"
 	set category = "OOC"
 
-	sanitize_and_communicate(/decl/communication_channel/ooc, src, message)
+	if(!message)
+		message = input(src.mob, "", "ooc \"text\"") as text|null
```
afterthought2: Maybe I'm misreading the diff but it seems this verb does nothing now.

code/modules/client/client_procs.dm
```diff
@@ -236,6 +236,9 @@ var/global/list/localhost_addresses = list(
 	if(prefs && !istype(mob, world.mob))
 		prefs.apply_post_login_preferences()
 
+	if(SSinput.initialized)
+		set_macros()
```
afterthought2, line 240: OK, yeah, didn't see it here before. It shouldn't be in all the places you put it in; use the char prefs thing to deal with it.

code/modules/keybindings/bindings_admin.dm
```diff
@@ -0,0 +1,29 @@
...
+		if("F8")
+			user.cmd_admin_pm_panel()
+			return
+		if("F9")
+			user.invisimin()
+			return
+		if("F10")
+			user.get_dead_say()
+			return
+	..()
```
afterthought2: Other places do `. = ..()`; pick one and stick with it.

afterthought2 (commented):
As I said on discord, I'm not a huge fan. I think the movement stuff is not very robust. I think that tracking held keys on client will be nightmarish going forward (would rather track stuff like ctrl, tab, and shift held via separate macro bindings), using the buffer thing only for diagonal movement).

That being said, if others think this is needed, then OK.

code/game/verbs/ooc.dm
quardbreak: woops

code/modules/mob/new_player/login.dm
quardbreak: I don't think our pref system support this yet.

code/_helpers/lists.dm
```diff
@@ -181,6 +181,11 @@ Checks if a list has the same entries and values as an element of big.
 		else
 			.[key] = call(merge_method)(.[key], b_value)
 
+
+//any value in a list
+/proc/sortList(list/L, cmp = /proc/cmp_text_asc)
+	return sortTim(L.Copy(), cmp)
```
afterthought2: Feel like this helper is pointless.

code/_helpers/lists.dm
```diff
@@ -718,3 +723,16 @@ var/global/list/json_cache = list()
 	catch(var/exception/e)
 		log_error("Exception during JSON decoding ([json_to_decode]): [e]")
 		return list()
+
+/proc/compare_list(list/l, list/d)
```
afterthought2: There's a softcode helper that does this (same_as_list or similar) and possibly a native one too.

code/_helpers/sanitize_values.dm
```diff
@@ -24,13 +24,17 @@
 		return text
 	return default
 
+/proc/sanitize_islist(value, default)
+	if(islist(value) && length(value))
+		return value
+	if(default)
+		return default
```
afterthought2, line 31: Don't think this needs a helper.

code/modules/keybindings/bindings_client.dm
quardbreak: According to DM reference, this is done for more flexible control:

> Normally a player can only call one verb per tick, but they can call any number of "instant" verbs in the same tick. This setting is useful for commands called by the game's interface, or for more responsive controls like for instance the use of "combos" in fighting games.

code/modules/client/client_procs.dm
quardbreak: Reduced places where it will call.

MistakeNot4892:
FWIW I'm in favour of this functionality, having rebindable keys and a panel that lists them instead of a single help verb is a major improvement. Movement loop and the capacity for diagonal movement is also nice

code/modules/keybindings/_keybindings.dm
```diff
@@ -0,0 +1,22 @@
...
+	var/category = CATEGORY_MISC
+	var/weight = WEIGHT_LOWEST
+
+/datum/keybinding/New()
+	// Default keys to the master "hotkey_keys"
+	if(LAZYLEN(hotkey_keys) && !LAZYLEN(classic_keys))
+		classic_keys = hotkey_keys.Copy()
+
+/datum/keybinding/proc/down(client/user)
+	return FALSE
```
afterthought2, line 15: What do these return values mean?

code/modules/keybindings/_keybindings.dm
```diff
@@ -0,0 +1,22 @@
+/datum/keybinding
+	var/list/hotkey_keys
+	var/list/classic_keys
+	var/name
+	var/full_name
+	var/description = ""
+	var/category = CATEGORY_MISC
+	var/weight = WEIGHT_LOWEST
```
afterthought2: Comment on what all these vars do.

code/modules/keybindings/bindings_client.dm
afterthought2: This doesn't resolve my concern. In the past, movement keys were instant but nothing else was. Certain keys where holding them matters may need to trigger instant verbs, but most do not.

code/modules/keybindings/readme.md
```diff
@@ -0,0 +1,40 @@
+# In-code keypress handling system
```
afterthought2, line 1: This stuff seems inaccurate.

code/modules/mob/mob_movement.dm
afterthought2: Please don't mark my comments as resolved unless you have addressed them. In fact, don't mark them as resolved at all, ever: reply to them instead.

code/modules/client/preference_setup/controls/01_keybindings.dm
```diff
@@ -0,0 +1,254 @@
...
+	var/list/user_binds = list()
+	for (var/key in pref.key_bindings)
+		for(var/kb_name in pref.key_bindings[key])
+			user_binds[kb_name] += list(key)
+
+	var/list/kb_categories = list()
+	// Group keybinds by category
+	for (var/name in global.keybindings_by_name)
+		var/datum/keybinding/kb = global.keybindings_by_name[name]
+		kb_categories[kb.category] += list(kb)
```
afterthought2, line 77: Why are we doing this stuff each time the UI is refreshed?

afterthought2 (commented):
A (large) number of my previous review comments are unaddressed, even if the diff has changed.

code/_helpers/lists.dm
MistakeNot4892: Didn't I remove this helper in the sortTim PR?

code/modules/client/preference_setup/controls/01_keybindings.dm
MistakeNot4892: Responding here because I can't respond elsewhere, but ironically a bunch of your comments can't be replied to.

code/modules/keybindings/bindings_client.dm
MistakeNot4892: Can you elaborate on why this is a concern? The rationale based on the doc seems valid and I don't think I'm following the practical aspect of your objection.

code/_helpers/global_lists.dm
```diff
@@ -93,6 +93,17 @@ var/global/list/string_slot_flags = list(
 		var/datum/sprite_accessory/marking/M = new path()
 		global.body_marking_styles_list[M.name] = M
 
+	// Keybindings
+	for(var/KB in subtypesof(/datum/keybinding))
```
MistakeNot4892, line 97: If you never actually need duplicate keybdings, it would be preferable to make a `/decl/keybinding`.

code/modules/client/preference_setup/controls/01_keybindings.dm
```diff
@@ -0,0 +1,254 @@
...
+					LAZYADD(key_bindings[classickeytobind], kb.name)
+					addedbind = TRUE
+		if(!addedbind)
+			notadded += kb
+
+	if(length(notadded))
+		addtimer(CALLBACK(src, .proc/announce_conflict, notadded), 5 SECONDS)
+
+/datum/preferences/proc/announce_conflict(list/notadded)
+	to_chat(client, "<span class='userdanger'>KEYBINDING CONFLICT!!!\n\
```
MistakeNot4892: `!` is sufficient, multiple exclamation marks just looks silly.
Do we have a `userdanger` class?

code/modules/keybindings/admin.dm
```diff
@@ -0,0 +1,93 @@
+/datum/keybinding/admin
+	category = CATEGORY_ADMIN
+
+/datum/keybinding/admin/can_use(client/user)
+	return user.holder ? TRUE : FALSE
```
MistakeNot4892: `return !!user.holder`

code/modules/keybindings/admin.dm
```diff
@@ -0,0 +1,93 @@
...
+	description = "Allows you to send a message to dead chat"
+
+/datum/keybinding/admin/dead_say/down(client/user)
+	user.get_dead_say()
+	return TRUE
+
+/datum/keybinding/admin/deadmin
+	hotkey_keys = list("None")
+	name = "deadmin"
+	full_name = "Deadmin"
```
MistakeNot4892: `De-Admin` would be clearer.

code/modules/keybindings/bindings_client.dm
MistakeNot4892: Specifically `code/modules/admin/spam_prevention.dm`.

code/modules/keybindings/readme.md
quardbreak: The entire readme is inaccurate. It copied from tgstation or TauCeti, and I have no language knowledge to rewrite it if needed.

code/modules/client/preference_setup/controls/01_keybindings.dm
quardbreak: I didn't come with UI part very well. It in state of Just WorksTM.

code/_helpers/global_lists.dm
quardbreak: I still don't know about decl for keybindings. If main philosophy of decl is to make static data holder, keybindings which contain edited data will probably a mistake.

code/modules/client/preference_setup/controls/01_keybindings.dm
quardbreak: Probably no. I will replace it with better span.

code/modules/keybindings/_keybindings.dm
quardbreak: I didn't see any related code which can work with it. Probably, it was designed to "confirm" what down button is successfully pressed.

code/_helpers/lists.dm
quardbreak: I removed this helper again so no issue with this.

code/modules/keybindings/_keybindings.dm
quardbreak: It needed because keypress verb uses check, and check for true return.

code/_helpers/lists.dm
quardbreak: Removed.

code/modules/keybindings/_keybindings.dm
quardbreak: Removed, due of lack of weight.

code/modules/keybindings/bindings_client.dm
afterthought2: Instant verbs increase server/networking load, and are a security concern. I think it's fine to say, hey, that's out of scope here, but it's also possible that instant isn't actually needed here and someone was trying to make this more "responsive." Due to how byond actually handles instant (and how this system handles movement), if it's the latter then I doubt instant is achieving anything. On the other hand, it's possible instant _is_ needed for the held down keys to register properly if you keep spamming shift or whatever, in which case it's something which is good to know and comment on.

I'm not really asking that `instant` be removed (unless it's not necessary) or the system changed to separate instant/noninstant keybinds, just that we are precise about _why_ and _where_ it's needed here.

code/modules/client/client_procs.dm
afterthought2: We discussed this and apparently this needs to trigger each time a client's mob changes (?). In any case, that addresses this and related concerns.

code/modules/client/client_defines.dm
```diff
@@ -53,3 +53,31 @@
...
+	/// A buffer for combinations such of modifiers + keys (ex: CtrlD, AltE, ShiftT). Format: `"key"` -> `"combo"` (ex: `"D"` -> `"CtrlD"`)
+	var/list/key_combos_held = list()
+	/*
+	** These next two vars are to apply movement for keypresses and releases made while move delayed.
+	** Because discarding that input makes the game less responsive.
+	*/
+	/// On next move, add this dir to the move that would otherwise be done
+	var/next_move_dir_add
+	/// On next move, subtract this dir from the move that would otherwise be done
+	var/next_move_dir_sub
```
afterthought2, line 73: To summarize several previous comments: I just don't understand what this does which `keys_held` isn't already doing. They don't seem to hold any additional information.

code/modules/keybindings/bindings_client.dm
afterthought2: After a tedious discussion on discord, we concluded that this is necessary to send multiple keystrokes on the same tick, which is required by this system. As discussed above, the latency issue varies between negligible and subtle (due to movement queuing).

afterthought2:
Still needs spam prevention addressed, and then probably some testing. Also some stray comments on the UI and helpers, possibly.

quardbreak:
Did quick testing with two people and seems no issues (at least with non-diagonal movement).
Probably can also test on neb test server but since nobody coming in it will be pointless.

code/controllers/subsystems/input.dm
```diff
@@ -0,0 +1,37 @@
...
+	flags = SS_TICKER
+	priority = SS_PRIORITY_INPUT
+	runlevels = RUNLEVELS_DEFAULT | RUNLEVEL_LOBBY
+
+	var/list/macro_set
+
+/datum/controller/subsystem/input/Initialize()
+	setup_default_macro_sets()
+
+	initialized = TRUE
```
afterthought2: I don't think this is allowed.

code/modules/client/preference_setup/controls/01_keybindings.dm
```diff
@@ -0,0 +1,254 @@
...
+			if(clear_key)
+				if(pref.key_bindings[old_key])
+					pref.key_bindings[old_key] -= kb_name
+					if(!(kb_name in pref.key_bindings["None"]))
+						LAZYADD(pref.key_bindings["None"], kb_name)
+					if(!length(pref.key_bindings[old_key]))
+						pref.key_bindings -= old_key
+				show_browser(user, null, "window=capturekeypress")
+				user.client.set_macros()
+				SScharacter_setup.queue_preferences_save(pref)
```
afterthought2: You probably shouldn't save here. They can save with the save button like for every other pref.

code/modules/client/preference_setup/controls/01_keybindings.dm
```diff
@@ -0,0 +1,254 @@
...
+			if(pref.key_bindings[old_key])
+				pref.key_bindings[old_key] -= kb_name
+				if(!length(pref.key_bindings[old_key]))
+					pref.key_bindings -= old_key
+			pref.key_bindings[full_key] += list(kb_name)
+			pref.key_bindings[full_key] = sortTim(pref.key_bindings[full_key], /proc/cmp_text_asc)
+
+			show_browser(user, null, "window=capturekeypress")
+			user.client.set_macros()
+			SScharacter_setup.queue_preferences_save(pref)
```
afterthought2: Again, not good to save without the user actually hitting save. Not going to comment in other places.

code/controllers/subsystems/input.dm
```diff
@@ -0,0 +1,37 @@
...
+	)
+
+// Badmins just wanna have fun ♪
+/datum/controller/subsystem/input/proc/refresh_client_macro_sets()
+	for(var/client/C in global.clients)
+		C.set_macros()
+
+/datum/controller/subsystem/input/fire()
+	for(var/client/C in global.clients)
+		C.keyLoop()
```
afterthought2, line 33: You should check tick here, probably. This is possibly quite slow.

code/__defines/client.dm
```diff
@@ -1,2 +1,2 @@
-#define CLIENT_MIN_FPS 0
+#define CLIENT_MIN_FPS -1
```
afterthought2: I don't know what this is doing. Please move all the client FPS stuff to a separate PR unless the system does not work without it.

code/modules/mob/mob_defines.dm
```diff
@@ -3,11 +3,14 @@
 	plane = DEFAULT_PLANE
 	layer = MOB_LAYER
 
+	appearance_flags = PIXEL_SCALE | LONG_GLIDE
 	animate_movement = 2
 	movable_flags = MOVABLE_FLAG_PROXMOVE
 
 	virtual_mob = /mob/observer/virtual/mob
 
+	var/shift_to_open_context_menu = FALSE
```
afterthought2: Please move this and everything related to it into a separate PR.

code/modules/mob/mob_movement.dm
```diff
@@ -146,8 +146,20 @@
...
+
+	var/datum/movement_handler/H = mob.GetMovementHandler(/datum/movement_handler/mob/delay)
+	if(H && H.MayMove() != MOVEMENT_PROCEED)
+		return
+	else
+		next_move_dir_add = 0
+		next_move_dir_sub = 0
+
+	if(!n || !direction)
+		return FALSE
```
afterthought2: I don't see the reason for this.

html/browser/common.css
```diff
@@ -173,35 +271,47 @@ h4
...
+.bgviolet{background-color:#552dab!important}
+.bgpurple{background-color:#8b2baa!important}
+.bgpink{background-color:#cf2082!important}
+.bgbrown {background-color:#8c5836!important}
+.bggrey{background-color:#646464!important}
+.bggray{background-color:#646464!important}
+.bggood{background-color:#4d9121!important}
+.bgaverage {background-color:#cd7a0d!important}
+.bgbad {background-color:#bd2020!important}
+.bglabel {background-color:#657a94!important}
```
afterthought2: Please move any CSS changes not critical for the PR into a separate pull request.

afterthought2 (approved):

PR #1824, opened 2021-07-03
Rewrites crematorium and morgue

quardbreak:


code/game/objects/structures/crematorium.dm
```diff
@@ -0,0 +1,262 @@
...
+
+	if(!connected_tray)
+		return
+
+	playsound(src, 'sound/items/Deconstruct.ogg', 50, 1)
+	var/turf/T = get_step(src, dir)
+	connected_tray.forceMove(T)
+	connected_tray.set_dir(dir)
+	for(var/atom/movable/A in src)
+		if(!ismob(A) && !isobj(A))
```
Lohikar: This should be checking `simulated` and/or `anchored`, not relying on the object not being a mob or obj.

code/game/objects/structures/crematorium.dm
```diff
@@ -0,0 +1,262 @@
...
+
+/obj/structure/crematorium/proc/close()
+	if(!open)
+		return
+	
+	if(!connected_tray)
+		return
+
+	for(var/atom/movable/A in get_turf(connected_tray))
+		if(!ismob(A) && !isobj(A))
```
Lohikar: I don't really see the point of these checks.

code/game/objects/structures/crematorium.dm
```diff
@@ -0,0 +1,262 @@
...
+			continue
+		if(!A.anchored && A.simulated && !(A == connected_tray))
```
Lohikar: `A != connected_tray`

afterthought2 (approved):

PR #1825, opened 2021-07-04
Fixes runtimes for get_facial_hair_styles

quardbreak:
If it wiil return null, it will pick random hair

### DIFF
--- a/code/modules/client/preference_setup/general/02_body.dm
+++ b/code/modules/client/preference_setup/general/02_body.dm
@@ -446,12 +446,12 @@ var/global/list/valid_bloodtypes = list("A+", "A-", "B+", "B-", "AB+", "AB-", "O
 
 	else if(href_list["facial_style"])
 		var/decl/bodytype/B = mob_species.get_bodytype_by_name(pref.bodytype)
-		var/list/valid_facialhairstyles = mob_species.get_facial_hair_styles(B.associated_gender)
+		var/list/valid_facialhairstyles = mob_species.get_facial_hair_styles(B?.associated_gender)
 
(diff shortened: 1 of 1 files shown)


code/modules/client/preference_setup/general/02_body.dm
```diff
@@ -658,7 +658,7 @@ var/global/list/valid_bloodtypes = list("A+", "A-", "B+", "B-", "AB+", "AB-", "O
 /datum/category_item/player_setup_item/proc/ResetFacialHair()
 	var/decl/species/mob_species = get_species_by_key(pref.species)
 	var/decl/bodytype/B = mob_species.get_bodytype_by_name(pref.bodytype)
-	var/list/valid_facialhairstyles = mob_species.get_facial_hair_styles(B.associated_gender)
+	var/list/valid_facialhairstyles = mob_species.get_facial_hair_styles(B?.associated_gender)
```
MistakeNot4892, line 661: This doesn't make any sense, if `B` is null then that implies `pref.bodytype` isn't set.

quardbreak: Well, yeah?.. If it null, then we will pick any hair anywaay.

MistakeNot4892: `pref.bodytype` should _never_ be null.

quardbreak: It is null.

изображение

MistakeNot4892: I'm not questioning that it's null, I'm saying that it being null is a fundamentally more significant issue than not getting the right facial hair and should be investigated/fixed rather than bandaiding over it.

PR #1831, opened 2021-07-06
Small fixes for vedning UI

quardbreak:


nano/css/shared.css
```diff
@@ -721,3 +730,23 @@ th.bot {
...
+	border-radius   : 0px 5px;
+	color           : #d1d1dd;
+	text-align      : center;
+	white-space     : pre-line;
+	overflow: wrap;
+}
+.withToolTip:hover .ImTooltip, .withToolTipInTooltip:hover .ImTooltipInTooltip{
+	display			: block;
+	position		: absolute;
+}
```
quardbreak: I tried to toy with tooltips on UI, but haven't figured out how to make it work there.
So I place it here for future use.

code/game/machinery/vending/_vending.dm
```diff
@@ -270,7 +270,7 @@
 		data["mode"] = 1
 		data["product"] = currently_vending.item_name
 		data["price"] = cur.format_value(currently_vending.price)
-		data["message_err"] = 0
+		data["price_num"] = text2num(cur.format_value(currently_vending.price))
```
MistakeNot4892: `text2num` makes no sense here, `format_value` returns a string.

code/game/machinery/vending/_vending.dm
```diff
@@ -287,6 +287,7 @@
 				"key" =    key,
 				"name" =   I.item_name,
 				"price" =  cur.format_value(I.price),
+				"price_num" = text2num(cur.format_value(I.price)),
```
MistakeNot4892: As above.

MistakeNot4892 (approved):

PR #1834, opened 2021-07-06
Browser cleanup

quardbreak:


code/datums/browser.dm
```diff
@@ -109,94 +119,47 @@
 	if(written_text)
 		. = user.handle_reading_literacy(user, .)
 
-/datum/browser/proc/open(var/use_onclose = 1)
```
afterthought2, line 112: You've searched for all instances of `open` and `update` and made sure they aren't using the old named args? People love using named args for these kinds of things.

code/datums/browser.dm
```diff
@@ -109,94 +119,47 @@
...
-// Optionally, specify the "ref" parameter as the controlled atom (usually src)
-// to pass a "close=1" parameter to the atom's Topic() proc for special handling.
-// Otherwise, the user mob's machine var will be reset directly.
-//
-/proc/onclose(mob/user, windowid, var/atom/ref=null)
-	if(!user || !user.client) return
-	var/param = "null"
-	if(ref)
-		param = "\ref[ref]"
+/// Register windowclose to be called, optionally with ref, when the window window_id is closed
```
afterthought2: Why are we removing the documentation here? This is a super non-intuitive feature, I'd rather all this stuff stayed.

code/game/objects/items/weapons/cards_ids.dm
```diff
@@ -222,7 +222,6 @@ var/global/const/NO_EMAG_ACT = -50
 		send_rsc(user, side, "side.png")
 	var/datum/browser/written/popup = new(user, "idcard", name, 600, 250)
 	popup.set_content(dat())
-	popup.set_title_image(usr.browse_rsc_icon(src.icon, src.icon_state))
```
afterthought2, line 225: Why are these being removed?

quardbreak: Legacy code which doesn't even work.

afterthought2: Fair enough, I guess. You should probably remove `proc/set_title_image` in that case.

afterthought2 (approved):

PR #1835, opened 2021-07-06
Constructable trays

Kurgis:
You can now construct hydroponics trays, since they no longer require a class of multiple unobtainable objects.

## Changelog
:cl:
bugfix: Hydroponics trays may now be constructed.
/:cl:


MistakeNot4892:
Please point this at the stable branch rather than dev. It's unlikely to make it to Outreach anytime soon if it goes to dev also.

PR #1836, opened 2021-07-06
Modularize lobby menu

quardbreak:
Allows to override lobby browser to use link buttons (not implemented because of the division of opinions)
Removed some poll system leftovers due it's outdated and nobody use it.


quardbreak:
Forgot to credit original PR which converted browser popup menu to lobbybrowser buttons!
https://github.com/TauCetiStation/TauCetiClassic/pull/6901

afterthought2:
OK, I tested this out. 

- I think you need to play around with the fonts to get one which looks higher-res. 
- The ready checkbox in particular looks really bad in this font.
- The mouse cursor on mouseover works very poorly; maybe try not modifying it at all.
- The bounding boxes on the text are very large and not visible, so it's possible to misclick. Try making them smaller.
- You should probably center the options, or even move them right.
- You should try to conceal the hyperlink readout at the bottom left that shows up on mouseover.
- There's a black border around the lobby image that looks terrible.
- I don't think this should be a prefs option: all or nothing.

This needs to be pretty polished if we are going to go with it.

code/_helpers/unsorted.dm
```diff
@@ -770,6 +770,15 @@ var/global/list/WALLITEMS = list(
 			colour += temp_col
 	return "#[colour]"
 
+/proc/refresh_lobby_browsers()
+	for(var/mob/new_player/player in global.player_list)
+		INVOKE_ASYNC(global.using_map, /datum/map/proc/show_titlescreen, player.client)
+		INVOKE_ASYNC(player, /mob/new_player/proc/show_lobby_menu)
```
afterthought2: Why are these async?

code/modules/admin/admin_verbs.dm
```diff
@@ -896,3 +897,29 @@ var/global/list/admin_verbs_mod = list(
...
+
+/client/proc/change_lobby_screen()
+	set name = "Lobby Screen: Change"
+	set category = "Fun"
+
+	if(!check_rights(R_FUN))
+		return
+
+	log_admin("[key_name(usr)] is trying to change the title screen.")
+	message_admins("[key_name_admin(usr)] is trying to change the title screen.")
```
afterthought2: `log_and_message_admins`

code/modules/client/client_procs.dm
```diff
@@ -532,8 +532,20 @@ var/global/list/localhost_addresses = list(
 	var/aspect_ratio = view_size[1] / view_size[2]
 
 	// Calculate desired pixel width using window size and aspect ratio
-	var/sizes = params2list(winget(src, "mainwindow.split;mapwindow", "size"))
-	var/map_size = splittext(sizes["mapwindow.size"], "x")
+	var/list/sizes = params2list(winget(src, "mainwindow.split;mapwindow", "size"))
```
afterthought2, line 536: If this is why it's async, just `waitfor = FALSE` someplace.

code/modules/mob/new_player/new_player.dm
```diff
@@ -100,28 +84,32 @@
...
 
-	if(href_list["ready"])
-		if(GAME_STATE <= RUNLEVEL_LOBBY) // Make sure we don't ready up after the round has started
-			ready = text2num(href_list["ready"])
-		else
-			ready = 0
+	if(href_list["lobby_ready"])
+		if(ready && SSticker.pregame_timeleft <= 5 SECONDS)
+			to_chat(src, SPAN_WARNING("Locked! The round is about to start."))
+			return
```
afterthought2: Why are we doing this? It's going to screw people over for no reason.

code/modules/mob/new_player/new_player.dm
```diff
@@ -549,8 +482,9 @@
 /mob/new_player/handle_reading_literacy(var/mob/user, var/text_content, var/skip_delays)
 	. = text_content
 
-/mob/new_player/handle_writing_literacy(var/mob/user, var/text_content, var/skip_delays)
-	. = text_content
```
afterthought2: Why did this get removed?

quardbreak: See above its the same proc just copy pasted

code/_helpers/unsorted.dm
quardbreak: First one ported from Tau and I pretty sure it's to not cause problems during updating lobby screen.

quardbreak: Changed according to Lohikar's tips.

code/modules/mob/new_player/new_player.dm
MistakeNot4892: Because it's an override. Put it back please.

quardbreak: I put it back.

code/modules/mob/new_player/new_player.dm
quardbreak: Removed.

code/_helpers/unsorted.dm
```diff
@@ -770,6 +770,15 @@ var/global/list/WALLITEMS = list(
...
 	return "#[colour]"
 
+/proc/refresh_lobby_browsers()
+	for(var/mob/new_player/player in global.player_list)
+		global.using_map.show_titlescreen(player.client)
+		player.show_lobby_menu()
+
+/proc/change_lobbyscreen(new_screen)
+	global.using_map.current_lobby_screen = new_screen || pick(global.using_map.lobby_screens)
+	refresh_lobby_browsers()
```
afterthought2: Wait, why are these in unsorted.dm?

code/modules/admin/admin_verbs.dm
```diff
@@ -896,3 +897,28 @@ var/global/list/admin_verbs_mod = list(
...
+	set name = "Lobby Screen: Change"
+	set category = "Fun"
+
+	if(!check_rights(R_FUN))
+		return
+
+	log_and_message_admins("is trying to change the title screen.")
+	SSstatistics.add_field_details("admin_verb", "LSC")
+
+	switch(alert(usr, "How change Lobby Screen?", "Lobby Screen", "Change", "Reset", "Cancel"))
```
afterthought2: `How change Lobby Screen`?

afterthought2 (commented):
Tested it out, otherwise seems fine.

afterthought2 (approved):

PR #1843, opened 2021-07-07
adds an area description blurb system

Mazianni:
This PR adds a system for displaying a small blurb of text to a player when they enter an area. This only happens **once** per area.

### DIFF
--- a/code/game/area/areas.dm
+++ b/code/game/area/areas.dm
@@ -43,6 +43,7 @@ var/global/list/areas = list()
 	var/list/forced_ambience
 	var/sound_env = STANDARD_STATION
 	var/turf/base_turf //The base turf type of the area, which can be used to override the z-level's base turf
+	var/description //A text-based description of what this area is for.
 
 	var/static/global_uid = 0
 	var/uid
@@ -53,6 +54,7 @@ var/global/list/areas = list()
 	var/list/air_scrub_names = list()
 	var/list/air_vent_info = list()
 	var/list/air_scrub_info = list()
+	var/list/blurbed_stated_to = list() //This list of names is here to make sure we don't state our descriptive blurb to a person more than once.
 
 /area/New()
 	icon_state = ""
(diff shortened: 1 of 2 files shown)


code/game/area/areas.dm
```diff
@@ -307,12 +309,24 @@ var/global/list/mob/living/forced_ambiance_list = new
...
 
+/area/proc/do_area_blurb(var/mob/living/L)
+	if(isnull(description))
+		return
+
+	if(!(L && L.client && L.get_preference_value(/datum/client_preference/area_info_blurb) == PREF_YES))
+		return 
+	
+	if(!(L.client in blurbed_stated_to))
+		blurbed_stated_to += L.client
```
Lohikar: Yeah, don't shove client refs into long lived lists please. Clients that log out and back in will be shown the message again.

Add the ckey instead.

Mazianni: Addressed.

code/game/area/areas.dm
```diff
@@ -307,12 +309,24 @@ var/global/list/mob/living/forced_ambiance_list = new
...
 	return ..()
 
+/area/proc/do_area_blurb(var/mob/living/L)
+	if(isnull(description))
+		return
+
+	if(!(L && L.client && L.get_preference_value(/datum/client_preference/area_info_blurb) == PREF_YES))
+		return 
+	
+	if(!(lowertext(L.ckey) in blurbed_stated_to))
```
MistakeNot4892: ckeys are always lowercase, they're always the same as output from `ckey()`.

code/game/area/areas.dm
```diff
@@ -307,12 +309,24 @@ var/global/list/mob/living/forced_ambiance_list = new
...
 /area/Exited(A)
 	if(isliving(A))
 		clear_ambience(A)
 	return ..()
 
+/area/proc/do_area_blurb(var/mob/living/L)
+	if(isnull(description))
+		return
+
+	if(!(L && L.client && L.get_preference_value(/datum/client_preference/area_info_blurb) == PREF_YES))
```
MistakeNot4892: `if(L?.get_preference_value(foo) != PREF_YES)` would be considerably shorter.

code/modules/client/preference_setup/global/preferences.dm
```diff
@@ -333,3 +333,12 @@ var/global/list/_client_preferences_by_type
...
 	flags = R_ADMIN|R_DEBUG
+
+/********************
+* Area Info Blurb *
+********************/
+
+/datum/client_preference/area_info_blurb
+	description ="Show area information"
+	key = "AREA_INFO"
+	default_value = PREF_YES
```
MistakeNot4892, line 344: Honestly I would prefer this to default to PREF_NO, though I'm not strongly opinionated either way.

Mazianni: I have a feeling this will never be seen if it defaults to No, tbh.

MistakeNot4892 (approved):

PR #1847, opened 2021-07-07
Makes energy shield light into a variable

listerla:
Some credit to @Casper3667 for assistance in this.

Changed the light the shield gives off to a var, and changed the icon_state on_update_icon to not be as hardcoded. For purposes of making a child of it downstream.

### DIFF
--- a/code/game/objects/items/weapons/shields.dm
+++ b/code/game/objects/items/weapons/shields.dm
@@ -150,11 +150,11 @@
 	origin_tech = "{'materials':4,'magnets':3,'esoteric':4}"
 	attack_verb = list("shoved", "bashed")
 	var/active = 0
-	var/shield_light = "#006aff"
+	var/shield_light_color = "#006aff"
 
 /obj/item/shield/energy/Initialize()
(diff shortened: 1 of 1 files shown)


code/game/objects/items/weapons/shields.dm
```diff
@@ -150,6 +150,11 @@
 	origin_tech = "{'materials':4,'magnets':3,'esoteric':4}"
 	attack_verb = list("shoved", "bashed")
 	var/active = 0
+	var/shield_light = "#006aff"
```
Lohikar: It would probably be better to name this `shield_light_color`, at a glance this looks like it would set light range.

code/game/objects/items/weapons/shields.dm
```diff
@@ -150,6 +150,11 @@
...
+
+/obj/item/shield/energy/Initialize()
+    icon_state = "[initial(icon_state)][active]"
```
afterthought2: Call `update_icon` here instead.

afterthought2 (approved):

PR #1848, opened 2021-07-07
Ports talk bubble sprite animation from Baystation

quardbreak:


MistakeNot4892:
Not sure I'm a fan, that rapid stutter looks bugged rather than intentional.

AlexMorgan3817:
Where before.gif?

MistakeNot4892:
Vote was not very positive for this one: 
image

I'd suggest some discussion on Discord about an alternative that is less choppy prior to reopening.

PR #1849, opened 2021-07-08
Firedoors now push people after closing self

quardbreak:
## Description of changes

This makes firedoors push people on side after successfully close.

When you close firedoor on same tile where are you, nobody able to fire at you with a gun (projectiles bump into firedoor).
This change should prevent people do such things.

## Authorship

AndroBetel, https://github.com/TauCetiStation/TauCetiClassic/pull/7548

## Changelog

:cl:
tweak: Firedoors now push people on close.
/:cl:

### DIFF
--- a/code/game/machinery/doors/firedoor.dm
+++ b/code/game/machinery/doors/firedoor.dm
@@ -63,6 +63,7 @@
 	blend_objects = list(/obj/machinery/door/firedoor, /obj/structure/wall_frame, /turf/unsimulated/wall, /obj/structure/window) // Objects which to blend with
 
 	var/allow_multiple_instances_on_same_tile = FALSE
+	var/push_on_close = TRUE
 
 /obj/machinery/door/firedoor/autoset
 	autoset_access = TRUE	//subtype just to make mapping away sites with custom access usage
@@ -317,7 +318,13 @@
 
 /obj/machinery/door/firedoor/close()
 	latetoggle()
-	return ..()
+	..()
+
+	if(push_on_close)
+		for(var/mob/living/L in loc)
+			for(var/dir in global.cardinal)
+				if(L.Move(get_step(loc, dir)))
+					break
 
 /obj/machinery/door/firedoor/open(var/forced = 0)
 	if(panel_open)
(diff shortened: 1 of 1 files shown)


code/game/machinery/doors/firedoor.dm
```diff
@@ -317,7 +318,13 @@
 
 /obj/machinery/door/firedoor/close()
 	latetoggle()
-	return ..()
+	..()
```
MistakeNot4892: `. = ..()`

code/game/machinery/doors/firedoor.dm
```diff
@@ -317,7 +318,13 @@
...
+
+	if(push_on_close)
+		for(var/mob/living/L in loc)
+			for(var/dir in global.cardinal)
+				if(L.Move(get_step(loc, dir)))
+					break
```
MistakeNot4892, line 328: This isn't going to stop firedoors from closing on top of people if there's nowhere to push them.

Mazianni:
This makes impromptu airlocks kind of hard to do in small spaces if I'm reading what the PR does right. Being able to stand on a tile and close a firelock on yourself is somewhat useful.

quardbreak:
No it's not. Closing door with contents which can't pass is literally an abuse.

MistakeNot4892:
Directional firedoors preserve that aspect of full tile firedoor utility without also allowing people to abuse closing a door on their own tile to block gunfire and atmos.

afterthought2:
~~Last I checked directional firedoors were broken and unmaintained.~~ I agree that they would cleanly accomplish the thing this PR attempts. I don't think people use firedoors as makeshift airlocks: it doesn't really work unless you build multiple of them, and in that case you might as well use inflatables.

Mazianni:
You use (1) inflatable and (1) firelock for a ghetto airlock. Not two firelocks.

code/game/machinery/doors/firedoor.dm
afterthought2: Could make it fail to close if it can't push, if we want to double down on this change.

Issue #1845, opened 2021-07-07
Steel table frame construction is glitched


Kurgis:
#### Description of issue

Steel table frames are bugged, when you attempt to create one you create a full steel table instead. When you deconstruct it, it drops a single glitched steel material stack which can't be interact with other material stacks. Additionally, when you do deconstruct it, it actually turns into the table frame that you wanted, but when you deconstruct said frame it just deletes the objects dropping no further material. Since my character is trained in Construction, you shouldn't lose 5 out of 6 material. Should also be noted that steel table frame appears twice in the material construction menu. Further testing shows that the 1 material sheet is what you get from covering a table frame with that material, for some reason the entire initial 6 steel sheets are being deleted, and it's placing a steel table instead of the frame. As mentioned before the frame drops no material, leading me to believe it creates a "ghost frame" with zero material in it.

#### Difference between expected and actual behavior

A steel table frame is constructed, which I can then layer a sheet atop of. Instead I get a steel table with a bugged frame if I deconstruct the top layer off.

#### Steps to reproduce

Play Engineering.

#### Specific information for locating

https://github.com/NebulaSS13/Nebula/blob/dev/code/modules/tables/tables.dm
https://github.com/NebulaSS13/Nebula/blob/dev/code/modules/materials/definitions/solids/materials_solid_metal.dm#L168

#### Length of time in which bug has been known to occur

Might me related with material changes, but I don't really know how long or when.

#### Client version, Server revision & Game ID

Downstream: Outreach
513 - cae48ed581a783fad60ce376bac09a1c521ff06f - ccT-cSuk

#### Issue bingo

- [x] Issue could be reproduced at least once
- [x] Issue could be reproduced by different players
- [ ] Issue could be reproduced in multiple rounds
- [x] Issue happened in a recent (less than 7 days ago) round
- [x] [Couldn't find an existing issue about this](https://github.com/NebulaSS13/Nebula/issues)

Kurgis:
Hasn't been tested if this issue extends to other table frames of different materials.

PR #1852, opened 2021-07-09
Simplifying and rewriting tablecode. [IDB IGNORE] [MDB IGNORE]

MistakeNot4892:
Pointing this at staging because the current table behavior is a mess (see #1845).
:cl:
tweak: Table construction has been simplified. The initial material used to build the frame determines the frame material, then you can use sheets to add a tabletop, rods to add reinforcement, or carpet to add felting. Screwdriver will remove each of these in turn and when all have been removed the frame can be dismantled with welding torch.
tweak: All structures should dismantle to more reasonable sheet amounts based on their matter var.
tweak: Racks are no longer a table subtype.
/:cl:


MistakeNot4892:
~~Ended up bundling a structure icon unification in with this as I am also messing with table icon updates.~~ Never mind, rolled that commit back.

MistakeNot4892:
I have no bloody idea why this is now causing cables to report as disconnected - wondering if I fucked something up with border tables? Unsure.

PR #1855, opened 2021-07-09
Bodyfall Sounds

Geevies:
:cl: Geeves
soundadd: Added a sound to falling over, willingly resting does not play the sound, however.
/:cl:

Code ported from: https://github.com/Aurorastation/Aurora.3/pull/8866, sounds acquired from https://freesound.org, further licensing detail is in the attributions file.


code/game/sound.dm
```diff
@@ -163,4 +163,6 @@ var/global/const/FALLOFF_SOUNDS = 0.5
 			if ("button") soundin = pick(global.button_sound)
 			if ("chop") soundin = pick(global.chop_sound)
 			if ("glasscrack") soundin = pick(global.glasscrack_sound)
+			if ("bodyfall") soundin = pick(global.bodyfall_sound)
+			if ("bodyfall_machine") soundin = pick(global.bodyfall_sound_machine)
```
quardbreak: not a fan of this

code/modules/species/species.dm
```diff
@@ -84,6 +84,8 @@
 	var/gibbed_anim =   "gibbed-h"
 	var/dusted_anim =   "dust-h"
 
+	var/bodyfall_sound = "bodyfall"
```
quardbreak: should be a list of sounds

code/modules/mob/living/carbon/human/human_movement.dm
```diff
@@ -161,3 +161,9 @@
 
 /mob/living/carbon/human/can_sprint()
 	return (stamina > 0)
+
+/mob/living/carbon/human/UpdateLyingBuckledAndVerbStatus()
+	var/old_lying = lying
+	. = ..()
+	if(lying && !old_lying && !resting && !buckled) // fell down
+		playsound(loc, species.bodyfall_sound, 50, TRUE, -1)
```
quardbreak: pick(species.bodyfall_sound)

quardbreak (changes requested):

quardbreak:
It would be nice if you move bodyfall sounds into new folder and include credit file there

code/game/sound.dm
MistakeNot4892: What's the issue?

quardbreak: Entire get_sfx is rude and complex to maintain. It works based on strings, not types, it requires to add new sounds into endless list. I'd prefer to move everything to, for an example of new system which can be made `/decl/sound_repository` but this is not the PR goal.

Geevies:
> 
> 
> It would be nice if you move bodyfall sounds into new folder and include credit file there

if the maintainers want it, sure thing

mods/species/utility_frames/species_bodytypes.dm
```diff
@@ -4,6 +4,10 @@
 	bodytype_category = BODYTYPE_HUMANOID
 	limb_blend =        ICON_MULTIPLY
 
+	bodyfall_sounds = list(
```
MistakeNot4892: Would suggest moving the clonk sound onto the base decl as `synthetic_bodyfall_sounds` and checking `isSynthetic()` in `UpdateLyingBuckledAndVerbStatus()` to decide which list to pick from. Otherwise a FBP (if they ever get readded) will make meat sounds when falling.

MistakeNot4892:
Attribution file and sound location looks fine, we should look at adding as much accreditation as we can to the file after this PR.

MistakeNot4892 (approved):

PR #1859, opened 2021-07-09
Updates pull request template to tgstation and TauCeti version

quardbreak:
## Description of changes

Changes Nebula's PR template to tgstation and TauCeti version.

Adds authorship section and section where people *should* explain which changes they do and why.
It allow reviewers and maintainers easier to know with what they deal.

## Authorship

TauCetiStation, https://github.com/TauCetiStation/TauCetiClassic/blob/master/.github/PULL_REQUEST_TEMPLATE.md
tgstation, https://github.com/tgstation/tgstation/blob/master/.github/PULL_REQUEST_TEMPLATE.md

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,35 +1,38 @@
-<!-- 
-Do not forget to add a changelog when you have made admin/player facing changes that can alter gameplay.
-Examples which require a changelog entry include:
-* Adding/removing objects that players may interact with, or the way they function.
-* Adding/removing/altering admin tools.
-* Changing the map.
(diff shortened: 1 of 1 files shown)


MistakeNot4892:
This will need some discussion from maints but I think it's fine personally.

afterthought2:
On the one hand, I think these templates are childish (you should just describe your PR succinctly in the header). On the other hand, I have noticed an uptick in the number of PRs with no header at all or unhelpful stuff in the header, so maybe we need something like this.

Kurgis:
As more beginning contributors are entering the downstreams and introduced here, this PR template will become more useful. I agree it really wasn't needed _before_, but that may change.

MistakeNot4892 (approved):
I don't think it harms anything to have it in place.

PR #1864, opened 2021-07-10
Outreach ported fixes

NataKilar:
Ports a few fixes from Outreach. Most of these are fixing missing designs or broken construction states. The exterior wall change is necessary to fix a bug with overlays updating properly.

### DIFF
--- a/code/game/objects/items/weapons/circuitboards/machinery/chemistry.dm
+++ b/code/game/objects/items/weapons/circuitboards/machinery/chemistry.dm
@@ -47,15 +47,11 @@
 		/obj/item/stock_parts/keyboard = 1,
 		/obj/item/stock_parts/power/apc/buildable = 1
 	)
+	buildtype_select = TRUE
 
-/obj/item/stock_parts/circuitboard/chemical_dispenser/bar_alc
-	name = "circuitboard (chemical dispenser)"
-	build_path = /obj/machinery/chemical_dispenser/bar_alc
-
-/obj/item/stock_parts/circuitboard/chemical_dispenser/bar_soft
-	name = "circuitboard (chemical dispenser)"
-	build_path = /obj/machinery/chemical_dispenser/bar_soft
-
-/obj/item/stock_parts/circuitboard/chemical_dispenser/bar_coffee
-	name = "circuitboard (chemical dispenser)"
-	build_path = /obj/machinery/chemical_dispenser/bar_coffee
+/obj/item/stock_parts/circuitboard/chemical_dispenser/get_buildable_types()
+	. = list()
+	for(var/path in typesof(/obj/machinery/chemical_dispenser))
+		var/obj/machinery/chemical_dispenser/chem_dispenser = path
+		if(initial(chem_dispenser.buildable))
+			. |= path
--- a/code/game/objects/items/weapons/circuitboards/machinery/chemistry.dm
+++ b/code/game/objects/items/weapons/circuitboards/machinery/chemistry.dm
@@ -47,3 +47,15 @@
 		/obj/item/stock_parts/keyboard = 1,
 		/obj/item/stock_parts/power/apc/buildable = 1
 	)
+
+/obj/item/stock_parts/circuitboard/chemical_dispenser/bar_alc
+	name = "circuitboard (chemical dispenser)"
+	build_path = /obj/machinery/chemical_dispenser/bar_alc
+
+/obj/item/stock_parts/circuitboard/chemical_dispenser/bar_soft
+	name = "circuitboard (chemical dispenser)"
+	build_path = /obj/machinery/chemical_dispenser/bar_soft
+
+/obj/item/stock_parts/circuitboard/chemical_dispenser/bar_coffee
+	name = "circuitboard (chemical dispenser)"
+	build_path = /obj/machinery/chemical_dispenser/bar_coffee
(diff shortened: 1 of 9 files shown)


code/game/objects/items/weapons/circuitboards/machinery/chemistry.dm
```diff
@@ -16,3 +16,46 @@
...
+	name = "circuitboard (chemical dispenser)"
+	build_path = /obj/machinery/chemical_dispenser/bar_alc
+
+/obj/item/stock_parts/circuitboard/chemical_dispenser/bar_soft
+	name = "circuitboard (chemical dispenser)"
+	build_path = /obj/machinery/chemical_dispenser/bar_soft
+
+/obj/item/stock_parts/circuitboard/chemical_dispenser/bar_coffee
+	name = "circuitboard (chemical dispenser)"
+	build_path = /obj/machinery/chemical_dispenser/bar_coffee
```
afterthought2: I suggest using the subtypes-via-multitool thing on some of these, to lower circuit type bloat.

code/modules/reagents/dispenser/dispenser2.dm
```diff
@@ -35,6 +39,7 @@
 	to_chat(user, "It has [cartridges.len] cartridges installed, and has space for [DISPENSER_MAX_CARTRIDGES - cartridges.len] more.")
 
 /obj/machinery/chemical_dispenser/proc/add_cartridge(obj/item/chems/chem_disp_cartridge/C, mob/user)
+	. = FALSE
```
afterthought2, line 42: On line 33-35, you aren't allowed to add full cartridges if it's being manually built. Check the `full_populate` var. You may also need to dump cartridges in deconstruct if it doesn't do so already.

code/modules/reagents/dispenser/dispenser2.dm
```diff
@@ -58,13 +63,11 @@
...
-
-	C.forceMove(src)
-	cartridges[C.label] = C
-	cartridges = sortAssoc(cartridges)
-	SSnano.update_uis(src)
+			C.forceMove(src)
+			cartridges[C.label] = C
+			cartridges = sortTim(cartridges, /proc/cmp_text_asc)
+			SSnano.update_uis(src)
+			return TRUE
```
afterthought2: This, on the other hand, doesn't work: it means mapped machines don't get any cartridges on init.

code/modules/reagents/dispenser/dispenser2.dm
```diff
@@ -74,16 +77,18 @@
 /obj/machinery/chemical_dispenser/attackby(obj/item/W, mob/user)
 	if(istype(W, /obj/item/chems/chem_disp_cartridge))
 		add_cartridge(W, user)
+		return TRUE
 
-	else if(isScrewdriver(W))
+	if(isCrowbar(W) && length(cartridges))
```
afterthought2: Should check `panel_open` to avoid interaction clashes. This makes it non-deconstructable.

code/modules/reagents/dispenser/dispenser2.dm
```diff
@@ -103,10 +108,9 @@
 		update_icon()
 		to_chat(user, "<span class='notice'>You set \the [RC] on \the [src].</span>")
 		SSnano.update_uis(src) // update all UIs attached to src
-
-	else
-		..()
-	return
+		return TRUE
```
afterthought2, line 114: Most of the early returns above should be `TRUE` as well.

code/modules/reagents/dispenser/dispenser2.dm
```diff
@@ -6,6 +6,10 @@
 	clicksound = "button"
 	clickvol = 20
 
+	construct_state = /decl/machine_construction/default/panel_closed
+	uncreated_component_parts = null
+	stat_immune = 0
+	base_type = /obj/machinery/chemical_dispenser
```
afterthought2, line 12: This potentially needs to be reset on subtypes (though see my earlier comment): basically, base_type must match what the circuitboard you want can build.

code/game/objects/items/weapons/circuitboards/machinery/chemistry.dm
```diff
@@ -16,3 +16,46 @@
...
+		/obj/item/stock_parts/manipulator = 1
+	)
+	additional_spawn_components = list(
+		/obj/item/stock_parts/console_screen = 1,
+		/obj/item/stock_parts/keyboard = 1,
+		/obj/item/stock_parts/power/apc/buildable = 1
+	)
+
+/obj/item/stock_parts/circuitboard/chemical_dispenser/bar_alc
+	name = "circuitboard (chemical dispenser)"
```
MistakeNot4892: Needs a unique name.

code/modules/reagents/dispenser/dispenser2.dm
NataKilar: Assuming you meant ``populate_parts`` here, but should be resolved. Cartridges seem to be dumped in deconstruct already,

code/modules/reagents/dispenser/dispenser2.dm
NataKilar: The idea here was that it would be deconstructable if there weren't any cartridges left, but because cartridges get dumped anyway, I added the panel_open check.

code/game/objects/items/weapons/circuitboards/machinery/chemistry.dm
NataKilar: Resolved by using the multitool circuit changing

MistakeNot4892 (approved):
All looks good to me, @afterthought2 might have further changes.

PR #1865, opened 2021-07-10
Adds PDA frame to modular computer recipe list

VNLoverMay:
## Description of changes

Adds PDA assemblies to the list of constructable modular computer assemblies.

PDAs aren't constructable like other modular computer subtypes, but they're pretty vital. They're the only portable device that takes a credstick broadcaster, for example. Not being able to construct them, while you can construct tablets and the like, doesn't make much sense.

## Authorship

Me. I literally just added a new recipe, that's it.

## Changelog

:cl:
rscadd: PDAs are now constructable
/:cl:

### DIFF
--- a/code/modules/materials/recipes_items.dm
+++ b/code/modules/materials/recipes_items.dm
@@ -138,6 +138,11 @@
 	result_type = /obj/item/modular_computer/tablet
 	difficulty = 2
 
+/datum/stack_recipe/computer/pda
+	title = "modular PDA frame"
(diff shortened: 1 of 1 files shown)


VNLoverMay:
Whew, had a moment of panic there when I somehow managed to submit the PR before typing it up.

MistakeNot4892 (approved):
Strikes me as weird to have any modular computers as stack recipes but this is consistent with existing code.

PR #1866, opened 2021-07-10
SScharacters queue and save file migration changes

quardbreak:
## Description of changes

Changes how SScharacters queue work a bit, now it moved to separate proc.

Changes migration thing process, now it will convert all characters (even legacy one) to json format.

Backed changes towards preferences loading.

## Authorship

OnyxBay and little bit of me.

### DIFF
--- a/code/controllers/subsystems/initialization/character_setup.dm
+++ b/code/controllers/subsystems/initialization/character_setup.dm
@@ -6,8 +6,8 @@ SUBSYSTEM_DEF(character_setup)
 	wait = 1 SECOND
 	runlevels = RUNLEVEL_LOBBY | RUNLEVELS_DEFAULT
 
-	var/list/prefs_awaiting_setup = list()
 	var/list/preferences_datums = list()
+	var/list/prefs_awaiting_setup = list()
 	var/list/newplayers_requiring_init = list()
 
 	var/list/save_queue = list()
@@ -16,12 +16,21 @@ SUBSYSTEM_DEF(character_setup)
 	while(prefs_awaiting_setup.len)
 		var/datum/preferences/prefs = prefs_awaiting_setup[prefs_awaiting_setup.len]
 		prefs_awaiting_setup.len--
-		prefs.setup()
+		prefs.setup_preferences(TRUE)
+
 	while(newplayers_requiring_init.len)
 		var/mob/new_player/new_player = newplayers_requiring_init[newplayers_requiring_init.len]
 		newplayers_requiring_init.len--
 		new_player.deferred_login()
+
 	. = ..()
+
+/datum/controller/subsystem/character_setup/proc/queue_prefs(datum/preferences/prefs)
+	// Calling this after subsytem's initialization is pointless
+	// and the client's preferences will never be loaded.
+	ASSERT(!initialized)
+
+	prefs_awaiting_setup += prefs
 
 /datum/controller/subsystem/character_setup/fire(resumed = FALSE)
 	while(save_queue.len)
@@ -34,5 +43,5 @@ SUBSYSTEM_DEF(character_setup)
 		if(MC_TICK_CHECK)
 			return
 
-/datum/controller/subsystem/character_setup/proc/queue_preferences_save(var/datum/preferences/prefs)
+/datum/controller/subsystem/character_setup/proc/queue_preferences_save(datum/preferences/prefs)
 	save_queue |= prefs
--- a/code/datums/mil_ranks.dm
+++ b/code/datums/mil_ranks.dm
@@ -16,9 +16,12 @@ var/global/datum/mil_branches/mil_branches = new()
  *  Global object for handling branches
  */
 /datum/mil_branches
-	var/list/branches                   // All branches that exist
-	var/list/spawn_branches_            // Branches that a player can choose for spawning, not including species restrictions.
-	var/list/spawn_branches_by_species_ // Branches that a player can choose for spawning, with species restrictions. Populated on a needed basis
+	/// All branches that exist
+	var/list/branches = list()
+	/// Branches that a player can choose for spawning, not including species restrictions.
+	var/list/spawn_branches_ = list()
+	/// Branches that a player can choose for spawning, with species restrictions. Populated on a needed basis
+	var/list/spawn_branches_by_species_ = list()
 
 /**
  *  Retrieve branch object by branch name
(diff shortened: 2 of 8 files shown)


code/modules/client/preference_setup/global/05_settings.dm
```diff
@@ -80,15 +80,23 @@
...
-/client/proc/get_preference_value(var/preference)
-	if(prefs)
-		var/datum/client_preference/cp = get_client_preference(preference)
-		if(cp)
-			return prefs.preference_values[cp.key]
-		else
-			return null
+/client/proc/get_preference_value(preference)
+	if(!SScharacter_setup.initialized)
+		return // Too early to use any preferences
```
afterthought2: This is incorrect: it must throw an error here. Returning null silently obscures logic errors in anything using prefs.

code/modules/client/preferences.dm
```diff
@@ -58,14 +61,13 @@ var/global/list/time_prefs_fixed = list()
...
 /datum/preferences/New(client/C)
-	if(istype(C))
-		client = C
-		client_ckey = C.ckey
-		SScharacter_setup.preferences_datums[C.ckey] = src
-		if(SScharacter_setup.initialized)
-			setup()
-		else
-			SScharacter_setup.prefs_awaiting_setup += src
+	ASSERT(istype(C))
```
afterthought2: This assertion can fail. It should not throw an error; the failure is expected.

code/modules/client/preferences.dm
```diff
@@ -443,8 +446,11 @@ var/global/list/time_prefs_fixed = list()
 	close_browser(user, "window=saves")
 
 /datum/preferences/proc/apply_post_login_preferences()
-	set waitfor = 0
-	if(!client)
-		return
+	set waitfor = FALSE
+
+	ASSERT(client)
```
afterthought2: Again, any assertion like this should be expected to fail.

code/controllers/subsystems/initialization/character_setup.dm
```diff
@@ -6,22 +6,33 @@ SUBSYSTEM_DEF(character_setup)
 	wait = 1 SECOND
 	runlevels = RUNLEVEL_LOBBY | RUNLEVELS_DEFAULT
 
-	var/list/prefs_awaiting_setup = list()
 	var/list/preferences_datums = list()
+	var/list/clients_awaiting_setup = list()
```
afterthought2: As discussed, I believe this is a strict downgrade.

code/modules/client/preference_setup/global/05_settings.dm
quardbreak: I removed log_error because it rely on preferences, causing a loop.

afterthought2: Use `CRASH`.

quardbreak: CRASH will pause server with enabled clientside debugger every time, because F5 in VSC makes you join in server right before initialization.

quardbreak: This also will flood with errors because people after server restart join before SScharacter_setup init.

afterthought2: This is a blocking comment: it _must_ crash here. Note that currently it does in fact crash here.

In re: your comments, this only happens with improper preference checks prior to the SS initializing. You are not allowed to make a prefs check before the SS is ready, that's how the logic works here. This cannot fail silently, as that's indistinguishable from certain pref values (and everything uses the returned values as bools regardless, even if they are distinguishable).

quardbreak: I decided to remove some new additions.

code/modules/client/preference_setup/global/05_settings.dm
```diff
@@ -80,15 +80,12 @@
...
-			return prefs.preference_values[cp.key]
-		else
-			return null
+/client/proc/get_preference_value(preference)
+	var/datum/client_preference/cp = get_client_preference(preference)
+	if(cp)
+		return prefs.preference_values[cp.key]
 	else
-		log_error("Client is lacking preferences: [log_info_line(src)]")
+		return null
```
MistakeNot4892: This is pointless.

code/datums/mil_ranks.dm
```diff
@@ -28,15 +31,15 @@ var/global/datum/mil_branches/mil_branches = new()
 		var/datum/mil_branch/branch = branch_name
 		branch_name = initial(branch.name)
 	if(branch_name && branch_name != "None")
-		return branches[branch_name]
+		return LAZYACCESS(branches, branch_name)
```
MistakeNot4892: Why change these to LAZY? They're always set to list().

quardbreak: Seems fixed by something, so removing LAZY.

code/modules/client/client_procs.dm
```diff
@@ -233,9 +224,6 @@ var/global/list/localhost_addresses = list(
 	if(holder)
 		src.control_freak = 0 //Devs need 0 for profiler access
 
-	if(prefs && !istype(mob, world.mob))
-		prefs.apply_post_login_preferences()
```
afterthought2, line 237: This will break stuff for non-newplayer mobs.

code/modules/client/preference_setup/global/05_settings.dm
```diff
@@ -85,8 +85,6 @@
 		var/datum/client_preference/cp = get_client_preference(preference)
 		if(cp)
 			return prefs.preference_values[cp.key]
-		else
-			return null
```
afterthought2: I believe this should be unreachable, but if it is reached the error thrown is incorrect. Probably just removing `if(cp)` is better.

afterthought2 (commented):
I still don't see why you want to screw around with the prefs stuff. It's putting the burden on me to essentially verify that it's functionally identical to current code (and I'm only 95% sure it is, as client deletion is finicky).

code/modules/client/client_procs.dm
```diff
@@ -185,9 +185,6 @@ var/global/list/localhost_addresses = list(
 	prefs = SScharacter_setup.preferences_datums[ckey]
 	if(!prefs)
 		prefs = new /datum/preferences(src)
-	prefs.last_ip = address				//these are gonna be used for banning
-	prefs.last_id = computer_id			//these are gonna be used for banning
-	apply_fps(prefs.clientfps)
```
afterthought2, line 190: Where is this stuff called if `prefs` exists (i.e. if line 187 isn't reached)?

quardbreak: Moved to deferred_login()

afterthought2: That doesn't work: the ip logging needs to happen immediately, it should not be deferrred.

quardbreak: Moved back.

afterthought2 (approved):

PR #1876, opened 2021-07-12
adds decl UID system

Mazianni:
## Description of changes

Adds a dynamic UID system to decls.

This adds a UID system to decls so there is always a unique identifer for them for easy fetching.

## Authorship

Mazian / Andromeda-K22
MistakeNot

## Changelog
This is not playerfacing. 

-->

### DIFF
--- a/code/datums/repositories/decls.dm
+++ b/code/datums/repositories/decls.dm
@@ -77,7 +77,7 @@ var/global/repository/decls/decls_repository = new
 	SHOULD_CALL_PARENT(TRUE)
 	SHOULD_NOT_SLEEP(TRUE)
 	if(!UID)
-		UID = "IDENTIFYME_[sequential_id(/decl)]"
+		UID = "AUTOGEN_[sequential_id(/decl)]"
 	decls_repository.decls_by_id[UID] = src
 	return
 
--- a/code/datums/repositories/decls.dm
+++ b/code/datums/repositories/decls.dm
@@ -67,20 +67,20 @@ var/global/repository/decls/decls_repository = new
 	if(!.)
 		for(var/decl in typesof(/decl))
 			var/decl/D = GET_DECL(decl)
-			if(D.UID == id)
+			if(D.uid == id)
 				return D
 
 /decl
-	var/UID //unique identifier. If this is not hardcoded, it is dynamically generated on runtime in Initialize() below.
+	var/uid //unique identifier. If this is not hardcoded, it is dynamically generated on runtime in Initialize() below.
 
 /decl/proc/Initialize()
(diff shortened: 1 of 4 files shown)


code/datums/repositories/decls.dm
```diff
@@ -60,9 +62,23 @@ var/global/repository/decls/decls_repository = new
...
+/repository/decls/proc/get_decl_by_id(var/id)
+	. = decls_by_id[id]
+	if(!.)
+		for(var/decl in typesof(/decl))
+			var/decl/D = GET_DECL(decl)
+			if(D.UID == id)
+				return D
+
+/decl
+	var/UID //unique identifier. If this is not hardcoded, it is dynamically generated on runtime in Initialize() below.
```
MistakeNot4892: Should be lowercase per the style guide as it isn't a const or a define.

code/datums/repositories/decls.dm
```diff
@@ -60,9 +62,25 @@ var/global/repository/decls/decls_repository = new
...
+	var/uid //unique identifier. If this is not hardcoded, it is dynamically generated on runtime in Initialize() below.
+
 /decl/proc/Initialize()
 	SHOULD_CALL_PARENT(TRUE)
 	SHOULD_NOT_SLEEP(TRUE)
+	if(!uid)
+		var/uid_pregen = replacetext("[type]", "/", "_")
+		uid_pregen = replacetext(uid_pregen, "_decl", "")
+		uid = "AUTOGEN[uid_pregen]"
+	decls_repository.decls_by_id[uid] = src
```
MistakeNot4892, line 83: Might be worth adding `if(decls_repository.decls_by_id[uid]) PRINT_STACK_TRACE("Duplicate /decl uid for [type]: [uid].")` before this line, though I suppose the unit test will catch that if you add one.

MistakeNot4892 (dismissed):

code/modules/materials/definitions/gasses/material_gas_mundane.dm
```diff
@@ -86,6 +90,7 @@
 		/decl/material/gas/methyl_bromide = 1
 	)
 	value = 0.25
+	uid = "gas_material_methyl_bromide"
```
MistakeNot4892: You've mixed up `gas_material` and `material_gas` in these names - might be good to just use `gas` and reduce uid length and potential for inconsistency.

MistakeNot4892:
As discussed on Discord briefly, the uid lookup proc is going to potentially instantiate abstract types and cause runtimes.

MistakeNot4892:
Misclick.

MistakeNot4892:
Alternative implementation: #1921

PR #1879, opened 2021-07-12
Fab and turf QOL/fixes.

### LINKED ISSUE #79: Standing on freshly placed tiles inflicted suffocation
Building new tiles, and immediately standing on them, you will take damage as if you where out into space. I had my lungs destroyed real quickly in my haste to assemble floor.
Also triggered an air alarm.

### LINKED ISSUE #1842: [QoL] Fabricator UI renders them nearly unusable at times
#### Description of issue

Autolathe, protolathe, circuit imprinter: Despite the categories, the Find function does not work with their menus due to an unknown issue (multiple people experienced this). This makes items near impossible to find because not only are objects not in alphabetical order, but there is no search function.

I can find what I'm looking for; I cannot.

#### Steps to reproduce

Look for Seed Extractor in a menu where both Seed Storage and Honey Extractor exist, while you have 30+ items in a black + green thin-type font.

Forever.

Downstream: Outreach
513 - cae48ed581a783fad60ce376bac09a1c521ff06f - ccT-cSuk

#### Issue bingo

MistakeNot4892:
- Closes #1842.
- Fixes #79.

### DIFF
--- a/code/modules/fabrication/fabricator_topic.dm
+++ b/code/modules/fabrication/fabricator_topic.dm
@@ -35,8 +35,8 @@
 		. = TOPIC_REFRESH
 
 	if(href_list["set_filter"])
-		var/new_filter_string = sanitize(input(usr, "Enter a new filter string.", "Fabricator Filter") as text|null)
-		if(CanInteract(usr, DefaultTopicState()))
+		var/new_filter_string = sanitize(input(user, "Enter a new filter string.", "Fabricator Filter") as text|null)
+		if(CanInteract(user, DefaultTopicState()))
 			filter_string = new_filter_string
 			. = TOPIC_REFRESH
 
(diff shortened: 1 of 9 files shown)


code/modules/fabrication/fabricator_topic.dm
```diff
@@ -28,7 +32,13 @@
 		if(!choice)
 			return TOPIC_HANDLED
 		selected_color = choice
-		return TOPIC_REFRESH
+		. = TOPIC_REFRESH
+
+	if(href_list["set_filter"])
+		var/new_filter_string = sanitize(input(usr, "Enter a new filter string.", "Fabricator Filter") as text|null)
+		if(CanInteract(usr, DefaultTopicState()))
```
afterthought2: `user`

afterthought2 (approved):

PR #1884, opened 2021-07-12
Changes ambition text output order 

quardbreak:

### DIFF
--- a/code/game/antagonist/antagonist_print.dm
+++ b/code/game/antagonist/antagonist_print.dm
@@ -8,15 +8,16 @@
 	for(var/datum/mind/P in current_antagonists)
 		text += print_player(P)
 		text += get_special_objective_text(P)
-		var/datum/goal/ambition = SSgoals.ambitions[P]
-		if(ambition)
-			text += "<br>Their goals for today were..."
-			text += "<br><span class='notice'>[ambition.summarize()]</span>"
 		if(!global_objectives.len && P.objectives && P.objectives.len)
 			var/num = 1
 			for(var/datum/objective/O in P.objectives)
 				text += print_objective(O, num)
 				num++
+
+		var/datum/goal/ambitions = SSgoals.ambitions[P]
+		if(ambitions)
+			text += "<br>Their goals for today were..."
+			text += "<br><span class='notice'>[ambitions.summarize()]</span>"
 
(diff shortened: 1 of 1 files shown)


code/game/antagonist/antagonist_objectives.dm
```diff
@@ -41,25 +41,24 @@
...
 		return
 
-	var/datum/goal/ambition/goal = SSgoals.ambitions[mind]
-	var/new_goal = sanitize(input(src, "Write a short sentence of what your character hopes to accomplish \
-	today as an antagonist.  Remember that this is purely optional.  It will be shown at the end of the \
-	round for everybody else.", "Antagonist Goal", (goal ? html_decode(goal.description) : "")) as null|message)
+	var/datum/goal/ambition/goals = SSgoals.ambitions[mind]
+	var/new_goal = sanitize(input(src, "Write down what you want to achieve in this round as an antagonist \
+	- your goals. They will be visible to all players after the end of the round. \
+	remember you cannot rewrite them - only add new lines.", "Antagonist Goal") as null|message)
```
afterthought2: It seems terrible that you can't rewrite them. Prepare for an eternity of "Can you modify my ambitions" ahelps.

quardbreak: You can just add new one with info like "I decided to not do X".

afterthought2: But what if you just made an error? I don't see how this is better than the current, more flexible setup.

quardbreak: Editing ambitions partially will ruin diary effect, means you can rewrite your entire history.

quardbreak: Current one is just a static text like "I want to do X" and all. New one allows to track special roles history and know how people change their tastes goals and exactly when.

MistakeNot4892: I'm with Afterthought, not letting them be freely editable is a downgrade from current function. You could achieve the same 'diary' thing by letting them add more ambitions to their goals list and optionally letting them specify a timestamp.

Overall not a fan of this PR as a mandatory thing.

afterthought2 (approved):

PR #1886, opened 2021-07-12
New dust gib animation

quardbreak:
## Description of changes

Changes how mob dust animation work.

It allow to build complete icon of mob and animate it with almost like real dust look effect.

## Authorship

Leshiy-ART, https://github.com/TauCetiStation/TauCetiClassic/pull/4340

## Changelog

:cl:
rscadd: Added new mob dust animation based on alpha mask.
/:cl:


code/_helpers/icons.dm
```diff
@@ -828,6 +828,21 @@ The _flatIcons list is a cache for generated icon files.
 	flat_icon.AddAlphaMask(alpha_mask)//Finally, let's mix in a distortion effect.
 	return flat_icon
 
+/proc/build_disappear_icon(atom/A)
```
MistakeNot4892, line 831: Please cache the result of these icon procs.

quardbreak: Why do we need to cache it if it used only one time and exclusive for signle atom?

quardbreak: And how I can cache it?

MistakeNot4892: You need to cache it because icon procs are one of the costliest things in the game and using them shouldn't be done for disposable stuff.
I'm not sure how you would cache this. In this instance I would say you should use an alpha mask filter or not do it at all. The GetFlatIcon()-based animation here is almost certainly not going to get merged.

PR #1890, opened 2021-07-14
Moves show_location_blurb to client level

quardbreak:
## Description of changes

Moves show_location_blurb() proc from global to client level.

This is a forced measure due to the inability to rewrite the global proc.
On good side, it should allow at least a way to override the proc from downstream, without messing with core file.

## Authorship

Myself.

### DIFF
--- a/code/controllers/subsystems/jobs.dm
+++ b/code/controllers/subsystems/jobs.dm
@@ -564,7 +564,7 @@ SUBSYSTEM_DEF(jobs)
 
 	job.post_equip_rank(H, alt_title || rank)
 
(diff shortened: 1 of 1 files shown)


code/controllers/subsystems/jobs.dm
```diff
@@ -564,7 +564,7 @@ SUBSYSTEM_DEF(jobs)
 
 	job.post_equip_rank(H, alt_title || rank)
 
-	INVOKE_ASYNC(GLOBAL_PROC, .proc/show_location_blurb, H.client, 30)
+	INVOKE_ASYNC(H.client, /client/proc/show_location_blurb, 30)
```
afterthought2: Why do you invoke it async if it's waitfor false?

afterthought2 (approved):

PR #1896, opened 2021-07-14
Blastdoor and shutters assembly fix

PsyCommando:
## Description of changes
Blast doors and shutters cannot be welded so they needed to have their own construct state since airlocks' states expect the door to be welded before being able to open the maintenance panel. 
Also, building blast doors/shutters from an assembly would result in the electronics not being moved to the blast door since the code handling inheriting electronics from assemblies was left completely in the airlock class. That resulted in the blast door/shutter being broken after construction.

This PR attempts to fix those issues, along with some extra minor bugfixes.

The PR moves the inherit_from_assembly() proc to the door class, so both the blast door and airlock class can reuse common code between both. 
It also create 2 new states for blast doors, for the panel_open and panel_closed state, so that it enforces door electronics, and uses the default machine behavior for opening the maintenance panel.
Also fixes the wrong icon being used on shutter assemblies, and the wrong assembly being used when deconstructing a shutter.

## Authorship
Me

## Changelog

:cl:
bugfix: Fixed shutters and blast doors deconstructed and panel_open state being unreachable.
bugfix: Fixed shutters assemblies having the wrong icon.
bugfix: Fixed shutters deconstructing to blast door assemblies.
/:cl:

### DIFF
--- a/code/game/machinery/_machines_base/machine_construction/blast_doors.dm
+++ b/code/game/machinery/_machines_base/machine_construction/blast_doors.dm
@@ -1,0 +1,10 @@
+//Uses default states except with a door board needed.
+// Cannot use airlock states, because blast doors do not weld or act the same way as airlocks
+/decl/machine_construction/default/panel_closed/blast_door
+	needs_board = "door"
+	down_state = /decl/machine_construction/default/panel_open/blast_door
+
+/decl/machine_construction/default/panel_open/blast_door
+	needs_board = "door"
+	up_state = /decl/machine_construction/default/panel_closed/blast_door
+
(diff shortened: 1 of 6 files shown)


afterthought2 (approved):
Haven't tested that all the interactions actually work out, but am going to assume that you have.

PsyCommando:
> 
> 
> Haven't tested that all the interactions actually work out, but am going to assume that you have.

Actually, I'm not 100% sure... I'll try to retest it again soon. Just been juggling between repos a whole bunch, so losing track a bit.. To be fair, I didn't expect the pr to be merged so quickly XD

PsyCommando:
Okay took a look and seems like hacking shutters is gone. I forgot to slip in a hacking state. Going to have to look into it later.

PsyCommando:
@MistakeNot4892 @afterthought2 Just curious, are blast doors supposed to be hackable? Because I don't think they have a wire datum setup??

MistakeNot4892:
I don't think they've ever been hackable as long as I've been playing.

afterthought2:
Don't think they are meant to be hackable, no.

PR #1897, opened 2021-07-14
Removes AS_ANYTHING and replaces with as anything

quardbreak:
## Description of changes

Removes overkill define which used to be a gap for old byond versions where `as anything` had issues.

Main reason why I opened this PR:
изображение

Currently MapDiff rendering is broken on our dev branch, because MapDiffBot trying to parse define instead just `as anything`.
I know what I can just replace it with `as anything` in one place, but then, this means what `AS_ANYTHING` is not what we need to support. In my option, this is crutches. 

## Authorship

Affected found the issue, I removed the define.


quardbreak:
Closed due someone else working at this already.

PR #1903, opened 2021-07-15
Moves AO layer down to avoid mouse opacity issues.

### LINKED ISSUE #1355: Gas Heater/Cooler mouse opacity
#### Description of issue
Can't properly click on it, click passes on turf underneath instead (turf is also displayed on statusbar, maybe it's some overlay problem?)

Able to use it without issues / Unable to click

#### Steps to reproduce
Create unary heater/cooler
Try to use it

/obj/machinery/atmospherics/unary/heater
or
/obj/machinery/atmospherics/unary/freezer
^^^ was not specified in report, freezer/heater. Problem probably may occur only with one since its code are copypasted from each other

iirc was also occuring to me even on bay, year++

https://github.com/infinitystation/Nebula/tree/50439ae5d5d062a21cc4b61ac64d4c32e8839dc4

#### Issue bingo

afterthought2:
## Description of changes

Moves the AO layer up. Makes a bunch of atmos devices I think are "safe" use structure layer by default (instead of exposed pipe layer).

AO overlays inherit the mouse opacity of whatever they are attached to (i.e. turfs). This cannot be overwritten, apparently, and turfs are mouse opaque. This means that if you click on anything layered above turfs but below the AO layer, your click is passed to the AO overlay and so to the turf. This leads to objects seeming (partially) mouse transparent.

The atmos devices touched (and possibly some others, but I'm unsure) should likely be using the default machinery layer, which is STRUCTURE_LAYER. This alone "fixes" the mouse opacity issue for them as well, but the AO layer change should address other (or future) offenders.

This does make AO layer below pipes and wires, but it seems to look OK to me.

Fixes #1355.

## Authorship

me

### DIFF
--- a/code/__defines/_planes+layers.dm
+++ b/code/__defines/_planes+layers.dm
@@ -93,6 +93,7 @@ What is the naming convention for planes or layers?
 	//ABOVE TURF
 	#define DECAL_LAYER                 2.03
 	#define RUNE_LAYER                  2.04
+	#define AO_LAYER                    2.045
 	#define ABOVE_TILE_LAYER            2.05
 	#define EXPOSED_PIPE_LAYER          2.06
 	#define EXPOSED_WIRE_LAYER          2.07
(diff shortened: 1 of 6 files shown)


MistakeNot4892 (approved):
We should consider moving this to viscontents or baking it into the lighting or something so it doesn't make pipes and other atoms on the edge of areas look bad, but this is fine for now.

PR #1913, opened 2021-07-16
Stickyban Support

quardbreak:
## Description of changes

This PR adds stickyban support, which is mainly inspired by tgstation's stickyban system.

It uses byond internal bans which would be hardly to bypass, and mostly should be used for times when user keep bypassing normal bans.

There are many cases when banned players could bypass a normal SQL bans. It possible to bypass it if you try to change your appearance. Many codebases's administration had pain and total control over those who did not stop bypassing the bans.

Automatic byond ban control allows to preserve cases when ban bypassers from different connection. It will sticky them if find out if they trying to bypass.
If sticky will apply for more than one player or etc, it will automatically revert all latest stickybans to preserve total sticky ban wave.

## Authorship

tgstation, Bear1ake and me


quardbreak:
Will reopen after discussion/when it will be ready.

PR #1918, opened 2021-07-18
Fix being unable to anchor/unanchor /obj/structure/closet

PsyCommando:
## Description of changes
~~Ensure the base class' handling for wrenches is reachable when the closet is closed so closets can be anchored like they're supposed to. Also set the tool_interaction_flags var  to allow anchoring.~~
EDIT:
Seems like my life is a lie and closets couldn't be anchored on bay apparently?? Anyways, treat this as a feature PR I guess.

## Changelog

:cl:
tweak: Allow anchoring/unanchoring closets/crates to the ground.
/:cl:

### DIFF
--- a/code/game/objects/structures/crates_lockers/closets/__closet.dm
+++ b/code/game/objects/structures/crates_lockers/closets/__closet.dm
@@ -5,6 +5,7 @@
 	icon_state = "base"
 	density = 1
 	maxhealth = 100
+	tool_interaction_flags = TOOL_INTERACTION_ANCHOR
 
 	var/welded = 0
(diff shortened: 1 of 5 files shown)


quardbreak:
Can you check if you unable to anchor/anchor wall closets? Because you should not do this kind of action on them.

PsyCommando:
> 
> 
> Can you check if you unable to anchor/anchor wall closets? Because you should not do this kind of action on them.

Done! 
I now wish wall mounted closets had their own subclass -_-

MistakeNot4892 (approved):

PR #1921, opened 2021-07-18
Added UIDs to material decls plus a method of retrieving them.

MistakeNot4892:
## Description of changes
- Adds a `uid` var to `/decl` along with some associated validation via unit test and a proc to retrieve based on `uid`.
- Unifies `abstract_type` and `hierarchy_type` on `/decl` and adds checks against it.
- Adds `uid` values to `/decl/material`.
- Removes ghost subtype `/decl/material/solid/mineral`.
- Bans `/decl/New()` and swaps all extant uses to `Initialize()`.
- Removes outfit macro names due to string concat breaking during compilation, again.

Currently there is no way to supply a human-readable (ie. non-path) value when mapping that will correspond to a material decl, specifically relevant to gas filters/gas mixers. This will allow them to be retrieved and also lays the groundwork to unit testing decl uniqueness/abstract type creation (which is useful for things like emotes in the future).

## Authorship
Myself and @Andromeda-K22.

## Changelog
Nothing player-facing.


MistakeNot4892:
String concat changes are being added to this PR as I was getting very weird compiler errors because of them, presumably because I added a bunch of strings to materials? Not sure.

MistakeNot4892:
All the map tests are stalling on this PR and I'm not sure why.

code/datums/repositories/decls.dm
```diff
@@ -17,26 +17,39 @@
...
+			PRINT_STACK_TRACE("Banned abstract /decl type instantiated: [decl_type]")
+		fetched_decls[decl_type] = decl // This needs to be done prior to calling Initialize() to avoid circular get_decl() calls by dependencies/children.
+		// TODO: maybe implement handling for LATELOAD and QDEL init hints?
+		var/init_result = decl.Initialize()
+		switch(init_result)
+			if(INITIALIZE_HINT_NORMAL)
+				. = decl
+			else
+				if(fetched_decls[decl_type] == decl)
+					fetched_decls -= decl_type
```
afterthought2, line 51: Don't really follow why we do this.

code/datums/repositories/decls.dm
```diff
@@ -17,26 +17,39 @@
...
 	. = fetched_decls[decl_type]
 	if(!.)
-		. = new decl_type()
-		fetched_decls[decl_type] = .
-
-		var/decl/decl = .
-		if(istype(decl))
-			decl.Initialize()
+		var/decl/decl = new decl_type()
+		if(decl_type == decl.abstract_type && decl.crash_on_abstract_init)
```
afterthought2: Should this be using the `is_abstract` helper?

code/datums/repositories/decls.dm
MistakeNot4892: Prevents error'd or otherwise invalid decl init from being kept in the list. We remove it rather than not adding it because it needs to be present in the list before Initialize() so that any decls that retrieve other decls in Initialize() don't end up in a loop.

code/datums/repositories/decls.dm
MistakeNot4892: Sure.

afterthought2 (dismissed):

afterthought2 (approved):

Issue #2, opened 2020-02-17
MEDpatch can be seen through z levels


MistakeNot4892:
Ref: https://github.com/ScavStation/ScavStation/issues/212

MistakeNot4892:
Humans need some code added to apply plane mangling to the atom when an emissive overlay is present in mob equipment.

PR #1933, opened 2021-07-24
Z-Mimic: Plane Fixup/Mangling

Lohikar:
This fixes issues with emissive overlays shining through floors.

- `no_z_overlay` is now a more general `z_flags` bitfield on movable, with `ZMM_IGNORE` controlling skip behavior now.
- `ZMM_MANGLE_PLANES` can be set on movables to forcibly correct planes of overlays/underlays during mimic so that they mimic properly.
- Existing Z-Mimic defines' comments are doccomments now.

I haven't added logic for handling fixup/mangling on humantypes for things like glowing eyes or smokables.


MistakeNot4892 (approved):
Fixes #2.

PR #1935, opened 2021-07-24
Adds images and contents list to accessories and closets

quardbreak:
## Description of changes

Ghosts (with antag hud) and admin-ghosts can examine closets to get info about what they contain, with icons!

Probably it is faster than messing with VV so you can just click and see what it have right now.

## Authorship

Polaris for counting english list procs

### DIFF
--- a/code/modules/clothing/_clothing.dm
+++ b/code/modules/clothing/_clothing.dm
@@ -165,13 +165,15 @@
 	if(istype(armor_datum) && LAZYLEN(armor_datum.get_visible_damage()))
 		to_chat(user, SPAN_WARNING("It has some <a href='?src=\ref[src];list_armor_damage=1'>damage</a>."))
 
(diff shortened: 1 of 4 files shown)


code/modules/clothing/_clothing.dm
```diff
@@ -165,13 +165,15 @@
 	if(istype(armor_datum) && LAZYLEN(armor_datum.get_visible_damage()))
 		to_chat(user, SPAN_WARNING("It has some <a href='?src=\ref[src];list_armor_damage=1'>damage</a>."))
 
-	for(var/obj/item/clothing/accessory/A in accessories)
-		to_chat(user, "[html_icon(A)] \A [A] is attached to it.")
+	if(LAZYLEN(accessories))
```
MistakeNot4892, line 168: Not a fan of moving this onto a single line.

MistakeNot4892 (approved):

PR #1939, opened 2021-07-24
Running all the current maps through StrongDMM's key reordering/sanitizing.

MistakeNot4892:
## Description of changes

Literally just opened each map in SDMM with TGM for format and duplicate removal ticked and saved again. Pointing it at stable first as it is just tidying up and I don't want to have to manually resolve conflicts with `staging` and `dev` after this.

Minimize diff noise from people using StrongDMM and generally improves the size/quality of the maps.

## Authorship

Me.

## Changelog

Nothing player facing.


afterthought2 (approved):

MistakeNot4892:
Tests have all passed, map diffs are taking ages for obvious reasons.

PR #1944, opened 2021-07-26
Adds default custom client fps in config

quardbreak:
Nobody reading pr body tldw: Breaks fps settings into client and world config option. By default client pref is 20 (can be changed in config to any hecking number you want).


Lohikar:
Per comments on Discord, I don't really see what the value in having this configurable over just having a hard-coded default is.

quardbreak:
less prs less pain

PR #1948, opened 2021-07-27
Emote sounds declarized

PsyCommando:
## Description of changes
I have been adding back in some old emote sounds in outreach and I ended up making a system using decl to have species have their own sounds for the same emote. I thought I might as well see if you guys have any use for it. Since it could potentially also make our lives easier downstream. It uses json files containing sounds for each emotes for each genders.

Also note that there are a bunch of placeholder decl and json files that are basically empty. I just left those in there to give an idea how it works. But you don't really need to make a new decl if there's no differences with the base /decl/emote_sounds.

Also, let me know if you guys are interested in having some new human emote sounds. Because I could also add those in here.

Makes sounds for emotes overridable by specie and/or mobtypes. Its pretty flexible. Also makes it so changing sounds for emotes can be done without having to make code edits. 

## Changelog

:cl:
tweak: Made emote sounds overridable by species/mob types. Moved emote sounds lists to json config files.
/:cl:


quardbreak:
Not really like these changes. I see these changes pretty useless, because:
1) All sounds in repo which should be used in code should be always included in rsc.
2) ~~Currently you can set per-gender assoc list for emotes, not sure if you know it but I'm not sure what this will be useful with json file.~~
3) You already can override emotes for species by making a subtype and placing new type with same emote key into species emotes.

tl;dr, as a person who uses audible emotes feature, I see these changes making system only complex.

quardbreak:
Also you already can override emotes by making subtype and placing it in species datum.

MistakeNot4892:
Runtime loading of emotes sort of breaks things - you end up needing to ship the entire assets folder along with the RSC and DMB, which isn't needed if you compile your assets directly. You also lose the ability to compile check your assets and have to do it via unit testing etc. which just frontloads the delays onto CI.

Not sure what use case you would have where you want want to change sounds via config either - overriding decls to change the sounds is simpler than working with an entire secondary configuration system as well. I like the infrastructure in place for hotswapping sounds for different species and mobs, but the json config and runtime loading aren't desirable imo. Will raise with other maints to get a consensus but I suspect they'll have similar views.

code/modules/emotes/definitions/audible.dm
```diff
@@ -78,10 +79,6 @@
 	emote_message_3p = "USER snores."
 	conscious = 0
 
-/decl/emote/audible/whimper
```
MistakeNot4892, line 81: Was this removed? Can't see it readded anywhere.

code/modules/emotes/emote_sounds.dm
```diff
@@ -0,0 +1,52 @@
+/*
+* Declare for storing species and gender specific emote sounds and make them overridable between species. 
+* Also the soundlist is loaded from a json array that sorts lists of sounds by emote name and optionally gender.
+*/
+
+//Mainly used for debugging and verifying everything loaded properly
+var/global/list/decl/emote_sounds/all_emote_decl = list()
```
MistakeNot4892, line 7: This is redundant - `get_decls_of_type(/decl/emote_sounds)` returns a list of all emote decls.

code/modules/emotes/emote_sounds.dm
```diff
@@ -0,0 +1,52 @@
...
+
+//merging proc for the merge assoc list proc
+/proc/merge_replace_if_not_null(var/a, var/b)
```
MistakeNot4892, line 10: IMO generic helpers like this should be in the helpers folder rather than mixed in with feature code.

code/modules/emotes/emote_sounds.dm
```diff
@@ -0,0 +1,52 @@
...
+	return b? b : a
+
+//A set of sounds for a set of emotes
+/decl/emote_sounds
+	var/list/soundbank = list() // 2d array, first is emote_name, second level contains sounds list by gender, or just a soundlist if there's no need for gender
+
+/decl/emote_sounds/New()
```
MistakeNot4892, line 17: Please avoid `New()`, `Initialize()` exists for decls and is preferred.

code/modules/emotes/emote_sounds.dm
```diff
@@ -0,0 +1,52 @@
...
+	..()
+	init_soundlist()
+	all_emote_decl += src
+
+/decl/emote_sounds/Destroy()
+	all_emote_decl -= src
+	. = ..()
+
+/decl/emote_sounds/proc/init_soundlist()
+	load_sound_config(sound_config_paths())
```
MistakeNot4892, line 27: This could be done in Initialize(), seems like it's only called in New() currently.

code/modules/emotes/definitions/audible.dm
PsyCommando: It was a duplicate definition for the emote.

PsyCommando:
> 
> 
> Runtime loading of emotes sort of breaks things - you end up needing to ship the entire assets folder along with the RSC and DMB, which isn't needed if you compile your assets directly. You also lose the ability to compile check your assets and have to do it via unit testing etc. which just frontloads the delays onto CI.
> 
> Not sure what use case you would have where you want want to change sounds via config either - overriding decls to change the sounds is simpler than working with an entire secondary configuration system as well. I like the infrastructure in place for hotswapping sounds for different species and mobs, but the json config and runtime loading aren't desirable imo. Will raise with other maints to get a consensus but I suspect they'll have similar views.

The main reason I made it like this is because it was easier to override partially lists. Since there's not really anything already in place to cleanly override list entries. 

And I guess I can see now how resources not being in the rsc could be an issue... I didn't really think about that tbh ^^;

Worst case its not hard to move the data back to a dm file really. Its just the sound override feature might need some extra work to pull off. Might need to turn each emote sound path into into a decl too so it can be overriden. But its doable.

code/modules/emotes/emote_sounds.dm
PsyCommando: Yeah, I'm aware. Its because I couldn't get the global decl list table to display in the global variables while debugging, so I left that here for that reason. Otherwise its nearly impossible to see the state of the list.
I'll remove it once debugging is done.

code/modules/emotes/emote_sounds.dm
PsyCommando: I agree with you on this. I just didn't know where to dump it. I couldn't even find where the other merge procs are supposed to be. So if you got suggestions let me know!

code/modules/emotes/emote_sounds.dm
PsyCommando: Gotcha, will change that.

code/modules/emotes/emote_sounds.dm
PsyCommando: Yeah, the main reason why I'm splitting it into another proc is for overriding. Its nearly impossible to completely override stuff that ends up being in initialize without having to copy paste the code you need to still run. While putting it in a separate proc you can easily completely override it without breaking init or copy pasting crap. 
I deal with that everyday currently when overriding stuff for saving soo, I thought I'd do something about it XD
Also consider that initialize is run exactly once and proc overhead won't be much of an issue here.

PsyCommando:
> 
> 
> Not really like these changes. I see these changes pretty useless, because:
> 
>     1. All sounds in repo which should be used in code should be always included in rsc.
> 
>     2. ~Currently you can set per-gender assoc list for emotes, not sure if you know it but I'm not sure what this will be useful with json file.~
> 
>     3. You already can override emotes for species by making a subtype and placing new type with same emote key into species emotes.
> 
> 
> tl;dr, as a person who uses audible emotes feature, I see these changes making system only complex.

The whole point is to avoid having to make 50 different poorly named useless types for emotes just because you need it to have a different sound for a specie. That's why its detached from the emote itself. 
Though its true it adds more complexity. But seems like most things are headed fro a decl approach on this codebase anyways.

PsyCommando:
After thinking for a bit, I think I'll just close this. It was a fun little experiment for me, but its probably not all that useful seeing the reactions to it.

PR #1953, opened 2021-07-27
Simplifying and rewriting tablecode. [IDB IGNORE] [MDB IGNORE]

MistakeNot4892:
Reopening from #1852 due to branch change.
Pointing this at staging because the current table behavior is a mess (see #1845).
:cl:
tweak: Table construction has been simplified. The initial material used to build the frame determines the frame material, then you can use sheets to add a tabletop, rods to add reinforcement, or carpet to add felting. Screwdriver will remove each of these in turn and when all have been removed the frame can be dismantled with welding torch.
tweak: All structures should dismantle to more reasonable sheet amounts based on their matter var.
tweak: Racks are no longer a table subtype.
/:cl:


MistakeNot4892:
Something this PR is causing mapped wires that aren't straight or corners to turn into nodes: https://cdn.discordapp.com/attachments/678826323728400394/869559425373974588/unknown.png

afterthought2 (approved):

PR #1956, opened 2021-07-28
Moves apply_fps calling from pref to client/New

quardbreak:
Seems it doesn't apply fps during post login.
Moving back to client/New but not sure if this fix this.


MistakeNot4892 (approved):
CI fail unrelated to PR.

PR #1957, opened 2021-07-28
Should stop the linter freaking out about transit turfs sleeping in Initialize().

MistakeNot4892:


MistakeNot4892:
Might also be feasible to just remove the pointless CHECK_TICK in toggle_transit.

MistakeNot4892:
Went with removing CHECK_TICK instead as throwing is already tickchecked by the subsystem.

afterthought2 (approved):

PR #1960, opened 2021-07-30
Power machine rework

NataKilar:
## Description of changes
Attempts to kill the /obj/machinery/power subtype and nearly succeeds. Now all machinery that uses the power subtype are those which require a direct reference to the powernet. Every other feature of the power subtype have been moved to base machinery
* Power generation procs have been moved to the base machinery type. Generating power adds available power to a cable node beneath the machine, and does not use the terminal stock part.
* All machinery which used to draw power directly from a cable node beneath it now use terminals. Machinery which will theoretically always be mapped over a power node (shield generations, FTL shunts) will create a terminal using the terminal_setup stock part preset. Other machinery such as emitters use the terminal_connect stock part preset, which will only connect to a terminal which it has been mapped on top of.

The power subtype is current a mix of objects actually related to power generation and management, and machinery which just draws power from the net directly. Bringing this machinery in line with other machines which use the terminal stock part for  drawing power from the powernet improves clarity.

These changes also allow for more flexibility in power generation. For example, currently atmospheric machinery cannot generate power, which is something I plan to use for Stirling engines in the near future.

## Authorship
Myself

## Changelog
:cl:
tweak: Machinery which previously connected to cables directly for power now require terminals, which can be added by attacking the machine with a stack of cables.
/:cl:

### DIFF
--- a/code/game/machinery/_machines_base/machinery_power.dm
+++ b/code/game/machinery/_machines_base/machinery_power.dm
@@ -162,5 +162,22 @@ This is /obj/machinery level code to properly manage power usage from the area.
 	if(power_init_complete && (use_power_mode == use_power))
 		REPORT_POWER_CONSUMPTION_CHANGE(old_power, new_power_consumption)
 
+// Return the powernet of a cable node underneath the machine.
+/obj/machinery/proc/get_powernet()
+	var/turf/T = loc
+	if(!T || !istype(T))
+		return
+
+	var/obj/structure/cable/C = T.get_cable_node()
+	if(C)
+		return C.powernet
+
+// Adds available power to a connected powernet, if available.
+/obj/machinery/proc/generate_power(var/amount)
+	var/datum/powernet/P = get_powernet()
+	if(!P)
+		return
+	P.newavail += amount
+
 #undef REPORT_POWER_CONSUMPTION_CHANGE
 #undef MACHINE_UPDATES_FROM_AREA_POWER
--- a/code/game/machinery/_machines_base/machinery_power.dm
+++ b/code/game/machinery/_machines_base/machinery_power.dm
@@ -165,7 +165,7 @@ This is /obj/machinery level code to properly manage power usage from the area.
 // Return the powernet of a cable node underneath the machine.
 /obj/machinery/proc/get_powernet()
 	var/turf/T = loc
-	if(!T || !istype(T))
+	if(!istype(T))
 		return
 
 	var/obj/structure/cable/C = T.get_cable_node()
--- a/code/game/machinery/syndicatebeacon.dm
+++ b/code/game/machinery/syndicatebeacon.dm
@@ -14,7 +14,7 @@
 
 	anchored = 1
 	density = 1
-
+	
 	var/temptext = ""
 	var/selfdestructing = 0
 	var/charges = 1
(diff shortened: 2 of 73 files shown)


code/game/machinery/_machines_base/machinery_power.dm
```diff
@@ -162,5 +162,22 @@ This is /obj/machinery level code to properly manage power usage from the area.
 	if(power_init_complete && (use_power_mode == use_power))
 		REPORT_POWER_CONSUMPTION_CHANGE(old_power, new_power_consumption)
 
+// Return the powernet of a cable node underneath the machine.
+/obj/machinery/proc/get_powernet()
+	var/turf/T = loc
+	if(!T || !istype(T))
```
afterthought2: `!T` is redundant.

code/game/machinery/syndicatebeacon.dm
```diff
@@ -81,89 +81,83 @@
...
+	if(ispath(path, /obj/item/stock_parts/power/terminal))
+		return TRUE
+	return ..()
 
-/obj/machinery/power/singularity_beacon/Destroy()
-	if(active)
+/obj/machinery/singularity_beacon/Destroy()
+	if(use_power)
 		Deactivate()
 	..()
```
afterthought2: `. = ..()`

...must be one of the last holdouts.

code/game/machinery/syndicatebeacon.dm
```diff
@@ -81,89 +81,83 @@
...
 
-//stealth direct power usage
-/obj/machinery/power/singularity_beacon/Process()
-	if(!active)
-		return PROCESS_KILL
-	if(draw_power(1500) < 1500)
+/obj/machinery/singularity_beacon/Process()
+	if(use_power && (stat & NOPOWER))
 		Deactivate()
```
afterthought2, line 160: Please change this into a `power_change` override that deactivates on stat getting `NOPOWER` added, and remove the `Process` proc.

afterthought2 (commented):
This seems largely fine. Can you comment on:

(A) Are there now functional differences in the construction/deconstruction of certain machines, and are they well-signaled to players in examine or via construction hints? If not, I would suggest adding explicit lines in examine about it, as e.g. emitter is a relatively common machine with a lot of ingrained player habits. 
(B) What is necessary to convert a map to the new setup? Is regexing `/machinery/power` to `/machinery` going to be enough, or do you need to do something more for certain machine types?

NataKilar:
> (A) Are there now functional differences in the construction/deconstruction of certain machines, and are they well-signaled to players in examine or via construction hints?

Machinery such as emitters, gyrotrons, etc. which previously were powered by direct connection to the powernet now require constructing a terminal first. When examining a machine which is not powered, instructions for powering the machine based off the installed power stock parts is given, as discussed on discord.

> (B) What is necessary to convert a map to the new setup? Is regexing `/machinery/power` to `/machinery` going to be enough, or do you need to do something more for certain machine types?

Regexing should be enough for most conversions. This doesn't totally kill the power subtype, so it'll need to be specific to each type that needs to be replaced. Most static machinery like shield generators which previously required direct connection to the powernet automatically generate terminals on Initialization through the terminal_setup stock part preset. Emitters and gyrotrons will require either manual mapping of terminals, or adding cables to the emitter will simply be another step on top of wrenching/welding.

code/game/machinery/syndicatebeacon.dm
NataKilar: I think I've done this correctly now, but it may be worth a second glance

afterthought2 (approved):

PR #1962, opened 2021-07-30
Adding a lot of material recipes for things that couldn't be built before

PsyCommando:
* Added a ton of new material recipes for things that couldn't be made before.
* Changed how curtains work by datumizing their appearance and feeding it to either the curtain's item form, or its structure form, in order to cut on repetition and make my life easier.
* Added a ton of empty variants to items and structures

## Description of changes
Added a lot of structure, furnitures, some basic clothes to the material recipes.
Modified how curtains works a bit so creating template classes wouldn't repeat code for nothing.
Added empty variants for the recipes.

It makes even more things buildable now. And I was asked to do it, so yeah.

## Changelog

:cl:
tweak: Added a lot of new material recipes.
/:cl:


code/game/objects/items/weapons/storage/boxes.dm
```diff
@@ -273,7 +291,9 @@
 	name = "box of donk-pockets"
 	desc = "<B>Instructions:</B> <I>Heat in microwave. Product will cool if not eaten within seven minutes.</I>"
 	icon_state = "donk_kit"
-	startswith = list(/obj/item/chems/food/donkpocket = 6)
+	startswith = list((/obj/item/chems/food/donkpocket = 6)
```
MistakeNot4892: Extra `(`

code/game/objects/structures/crates_lockers/closets/utility_closets.dm
```diff
@@ -67,39 +73,43 @@
...
-	if(prob(30))
-		new /obj/item/stack/cable_coil/random(src)
-	if(prob(20))
-		new /obj/item/multitool(src)
-	if(prob(5))
-		new /obj/item/clothing/gloves/insulated(src)
-	if(prob(40))
-		new /obj/item/clothing/head/hardhat(src)
-
+	if(spawn_contents)
```
MistakeNot4892: Would be better to move all this into `WillContain()` maybe?

code/modules/materials/material_recipes.dm
```diff
@@ -54,19 +54,23 @@
 	if(integrity >= 50 && hardness >= MAT_VALUE_FLEXIBLE + 10)
 		. += new/datum/stack_recipe/furniture/door(src)
 		. += new/datum/stack_recipe/furniture/barricade(src)
+		. += new/datum/stack_recipe/furniture/barricade/spikes(src)
```
MistakeNot4892: Spikes are added to the barricade after construction - doesn't need a separate recipe imo.

MistakeNot4892:
A very large chunk of these recipes are excessive imo - being able to create a janitor cart or a toilet or a crematorium out of a stack of materials strains credulity in a way that a stool or a chair doesn't to me. It also completely invalidates cargo as a concept, which makes sense for Outreach but I don't think is a good choice for main Nebula. I would suggest maybe having an `extended_crafting_recipes` modpack or such as part of this PR.

code/modules/materials/material_recipes.dm
PsyCommando: Ooooooooh, I actually didn't ever realized this was a thing...

PsyCommando:
> 
> 
> A very large chunk of these recipes are excessive imo - being able to create a janitor cart or a toilet or a crematorium out of a stack of materials strains credulity in a way that a stool or a chair doesn't to me. It also completely invalidates cargo as a concept, which makes sense for Outreach but I don't think is a good choice for main Nebula. I would suggest maybe having an `extended_crafting_recipes` modpack or such as part of this PR.

I mean, you can currently build padded chairs with steel sheets.. And they give you cloth when you take them appart. So I'm not exactly sure what is your limit on these sorts of things. You'll have to give me a list of things you think are too much.
Also, I'm absolutely down with making some things have a more accurate way of building them. But as of now, you can only assemble machines from frames, not structures. And I don't know if I should mess with machine assembly code or just write something similar from scratch for structures.

Also not entirely sure what you mean by it invalidating cargo? Most of the things in here cannot be ordered via cargo. They're just unobtainable.

And about the modpack. I'd agree to that. But there's nothing in place to validate against duplicate definitions for recipes. Which is going to cause massive headaches in the future downstream. Ideally, we should probably rewrite crafting to be unit-testable. And probably also decouple it from individual material types decl. Maybe something like the old table crafting system, but better?

code/game/objects/structures/crates_lockers/closets/utility_closets.dm
PsyCommando: That's actually true, but I wasn't sure how it would behave with probabilities being there. I guess we'll see.

code/game/objects/items/weapons/storage/boxes.dm
PsyCommando: I love how I managed to break the code by fixing a merge conflict.. :D

code/game/objects/structures/crates_lockers/closets/utility_closets.dm
PsyCommando: I used the atom_creator/simple class like something else was. Hopefully that thing works.

code/game/objects/structures/curtains.dm
```diff
@@ -1,3 +1,71 @@
+//
+// Curtain types declaration
+//
+/decl/curtain_kind
```
MistakeNot4892, line 4: Would you mind splitting the curtain code out into a separate PR? It'll review easier and looks fine to go in independently of the other changes in this PR.

PsyCommando: Alright, I'll do that in a bit

PsyCommando: Still haven't gotten around to it. Had more server issues with outreach so..

PsyCommando: Okay its done!

PsyCommando:
Soo, the unit test seems to be a bit out of its mind?
Its kinda funny its not picking up padded chairs since those have had leather as secondary material for ages now?

MistakeNot4892:
Unit tests are actually valid, refer to #1987.

Some notes from looking this over. This is mostly stuff that I think should be kept to Outreach as it doesn't conceptually gel with Nebula in my view. 
- Being able to craft secure closets, which have preset access, strikes me as a questionable idea. You could craft the locker on top of a handful of items, the items would be loaded into the locker, and you then wouldn't be able to open it. It's also just a bit odd that you can hand-shape electrical circuitry out of thin air for it.
- Locker/crate subtypes in general are ehhhh - would prefer there to be some kind of ability to attach an electronic lock to a locker to make it secure, and change the appearance with a paint gun.
- The various box subtypes that are only different visually should probably be handled by using the paint gun on a box, or a pen, rather than separate crafting recipes. This would also mean you wouldn't need to define nine billion empty subtypes.
- When did handrails lose their recipe? That's so odd. I could have sworn you could craft them already.

The above might be worth broader discussion with maints if you're unwilling to shunt it all into extended crafting.

This other stuff is in the extended crafting module so ambivalent about it in general, but for the record (non-blocking):
- Functional atoms like stasis cages, ore boxes, roller beds really shouldn't be craftable with simple stacks imo.
- Textiles stuff should definitely have a crafting mechanic other than sheets in hands, but I appreciate textiles fab is already loaded with stuff.
- Reagent dispensers - kind of ehhhh on these as well, it feels like they're fairly complex. Maybe if they were crafted from a frame and then had tanks painted/added. 
- Sinks - Infinite reagent sources made out of metal, not a fan of this. Maybe when plumbing exists meaningfully.
- Toilets/urinals/etc - not currently set up to support materials, not a fan of crafting them.

PsyCommando:
> 
> 
> Unit tests are actually valid, refer to #1987.
> 
> Some notes from looking this over. This is mostly stuff that I think should be kept to Outreach as it doesn't conceptually gel with Nebula in my view.
> 
>     * Being able to craft secure closets, which have preset access, strikes me as a questionable idea. You could craft the locker on top of a handful of items, the items would be loaded into the locker, and you then wouldn't be able to open it. It's also just a bit odd that you can hand-shape electrical circuitry out of thin air for it.
> 
>     * Locker/crate subtypes in general are ehhhh - would prefer there to be some kind of ability to attach an electronic lock to a locker to make it secure, and change the appearance with a paint gun.
> 
>     * The various box subtypes that are only different visually should probably be handled by using the paint gun on a box, or a pen, rather than separate crafting recipes. This would also mean you wouldn't need to define nine billion empty subtypes.
> 
>     * When did handrails lose their recipe? That's so odd. I could have sworn you could craft them already.
> 
> 
> The above might be worth broader discussion with maints if you're unwilling to shunt it all into extended crafting.
> 
> This other stuff is in the extended crafting module so ambivalent about it in general, but for the record (non-blocking):
> 
>     * Functional atoms like stasis cages, ore boxes, roller beds really shouldn't be craftable with simple stacks imo.
> 
>     * Textiles stuff should definitely have a crafting mechanic other than sheets in hands, but I appreciate textiles fab is already loaded with stuff.
> 
>     * Reagent dispensers - kind of ehhhh on these as well, it feels like they're fairly complex. Maybe if they were crafted from a frame and then had tanks painted/added.
> 
>     * Sinks - Infinite reagent sources made out of metal, not a fan of this. Maybe when plumbing exists meaningfully.
> 
>     * Toilets/urinals/etc - not currently set up to support materials, not a fan of crafting them.

I wish you'd have said that in the first place so I wouldn't have bothered running it through here. Its more of a stopgap measure because on outreach you cannot do much currently without sinks for instance. And you can't get any of those lockers, crates, reagent dispensers, etc... Its fully intended to go away eventually. Paint gun on crate is an interesting idea though. Might be worth revisiting that in another PR. There's just the issue with non-standard containers that don't use the appearance decl.

As for the other options. Multi-step crafting requires extensive amounts of spriting and a lot of repetitive code, so its not a good short-term solution. And I'm still very inexperienced when it comes to spriting. 
Fabs are not working great right now, so I'd rather avoid loading them with more stuff. Plus research is currently broken on outreach, and I don't know when/if nata is gonna fix it. 
Plumbing is something we definitely should implement the sooner the better. But that's beyond the scope of this PR. 

Also about handrails. There's 2 types of them. One's called railling, the other is handrail. The handrail is different in that you can attach yourself to them, and they're meant to be placed alongside walls. Maybe they need an updated description/name?

If there's no objections I'll probably close this PR, since there's not really much point in having it here if you don't need any of it.

PR #1982, opened 2021-08-03
Add pipefail set to linter annotation.

MistakeNot4892:
Should fix the linter not causing CI to fail. Thanks @Lohikar.


MistakeNot4892:
Verified that this fails CI when an error is introduced, should be good to merge.

Lohikar (approved):

PR #1983, opened 2021-08-03
Added fuel vaporization to fluid puddles.

MistakeNot4892:
Adding some very basic fire interaction to fuel puddles, as a stopover between now and a proper fire rewrite. Essentially just causes puddles to vaporize any fuel in their reagents on exposure to sufficient heat, meaning that zone fires will start.
DNM, have not tested locally due to putting it together at work.

### DIFF
--- a/code/modules/ZAS/Fire.dm
+++ b/code/modules/ZAS/Fire.dm
@@ -28,9 +28,14 @@ If it gains pressure too slowly, it may leak or just rupture instead of explodin
 		return 0
 	if(locate(/obj/fire) in src)
 		return 1
+
 	var/datum/gas_mixture/air_contents = return_air()
 	if(!air_contents || exposed_temperature < FLAMMABLE_GAS_MINIMUM_BURN_TEMPERATURE)
 		return 0
+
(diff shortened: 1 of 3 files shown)


MistakeNot4892:
Does not work :(

MistakeNot4892:
Working to an extent - not super reliable and seems to make welding tanks more dangerous than hydrogen tanks, but it does at least make fuel spills dangerous.

afterthought2 (approved):

code/modules/ZAS/Fire.dm
```diff
@@ -125,6 +130,12 @@ If it gains pressure too slowly, it may leak or just rupture instead of explodin
 	for(var/atom/A in loc)
 		A.fire_act(air_contents, air_contents.temperature, air_contents.volume)
 
+	// prioritize nearby fuel overlays first
+	for(var/direction in global.cardinal)
```
MistakeNot4892, line 134: DNM, this is going to stack overflow.

MistakeNot4892: Isn't an issue, got confused.

PR #1984, opened 2021-08-04
Fire rewrite draft.

MistakeNot4892:
Opening this as a draft for collaboration/input/progress tracking.

Current loose plan:
- Fires will be atoms that are attached to another atom, which is considered to be on fire.
- Fires will process independently of ZAS, searching for fuel and oxidizer in their host atom and environment. They will use matter, reagents or local fluids as well as gas.
- ZAS will handle flashpoint fires and zone explosions (replacing the current fire behavior with an immediate violent detonation like a real gas fire).

Considerations:
- Atoms should have some handling in place for having some of their matter consumed by fires.
- There should be some kind of precalculation or caching to prevent every individual fire needing to iterate all turf contents etc while burning.
- Messing with the local air datum (via consming gas or producing products) should be queued or otherwise optimized to avoid ZAS updating 40 times per fire processor tick.

Current progress:
- Fires are largely located in a submodule folder and have been decoupled from ZAS.
- Various extinguish/ignition procs have been merged into each other.


code/modules/materials/definitions/liquids/materials_liquid_water.dm
```diff
@@ -60,13 +60,7 @@
...
 	var/min_temperature = T20C + rand(0, 20) // Room temperature + some variance. An actual diminishing return would be better, but this is *like* that. In a way. . This has the potential for weird behavior, but I says fuck it. Water grenades for everyone.
 
-	var/hotspot = (locate(/obj/fire) in T)
-	if(hotspot && !isspaceturf(T))
-		var/datum/gas_mixture/lowertemp = T.remove_air(T:air:total_moles)
-		lowertemp.temperature = max(min(lowertemp.temperature-2000, lowertemp.temperature / 2), 0)
-		lowertemp.react()
-		T.assume_air(lowertemp)
-		qdel(hotspot)
+	T.extinguish()
```
CrimsonShrike, line 63: Intended that water can no longer cool places or was it bugging or something?

MistakeNot4892: It doesn't make much sense. Can be revisited.

MistakeNot4892:
To the todo pile.

PR #1986, opened 2021-08-05
Turned curtain types into decl

PsyCommando:
## Description of changes

Moved the curtains decl into a new PR branch as requested on #1962.

Basically makes all curtains a single structure and a single item that just take on the properties specified in a decl. 

## Changelog

:cl:
tweak: Make curtains use decl instead of several subtypes.
/:cl:

### DIFF
--- a/code/game/objects/structures/curtains.dm
+++ b/code/game/objects/structures/curtains.dm
@@ -1,3 +1,71 @@
+//
+// Curtain types declaration
+//
+/decl/curtain_kind
+	var/name = "curtain"
+	var/color = "white"
+	var/alpha = 255
+	var/material_key = /decl/material/solid/plastic
+
+/decl/curtain_kind/proc/make_item(var/loc)
+	var/obj/item/curtain/C = new(loc, material_key, src)
+	C.alpha = alpha
+	C.set_color(color)
+	C.SetName("rolled [name]")
+	return C
+
+/decl/curtain_kind/proc/make_structure(var/loc, var/dir, var/opened = FALSE)
+	var/obj/structure/curtain/C = new(loc, dir, material_key, null, src)
+	C.alpha = alpha
+	C.set_color(color)
+	C.SetName(name)
+	C.set_opacity(opened)
+	return C
+
+//Cloth curtains
+/decl/curtain_kind/cloth
+	material_key = /decl/material/solid/cloth
+
+/decl/curtain_kind/cloth/bed
+	name = "bed curtain"
+	color = "#854636"
+
+/decl/curtain_kind/cloth/black
+	name = "black curtain"
+	color = "#222222"
+
+/decl/curtain_kind/cloth/bar
+	name = "bar curtain"
+	color = "#854636"
+
+//Plastic curtains
+/decl/curtain_kind/plastic
+	name = "plastic curtain"
+	color = "#b8f5e3"
+	material_key = /decl/material/solid/plastic
+
+/decl/curtain_kind/plastic/medical
+	alpha = 200
+
+/decl/curtain_kind/plastic/privacy
+	name = "privacy curtain"
+
+/decl/curtain_kind/plastic/shower
+	name = "shower curtain"
+	color = "#acd1e9"
+	alpha = 200
+
+/decl/curtain_kind/plastic/shower/engineering
+	color = "#ffa500"
+
+/decl/curtain_kind/plastic/shower/security
+	color = "#aa0000"
+
+/decl/curtain_kind/plastic/canteen
+	name = "privacy curtain"
+	color = COLOR_BLUE_GRAY
+
+
 /obj/item/curtain
 	name = "rolled curtain"
 	desc = "A rolled curtains."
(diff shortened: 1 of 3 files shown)


code/game/objects/structures/curtains.dm
```diff
@@ -1,3 +1,71 @@
...
+
+/decl/curtain_kind/plastic/shower/security
+	color = "#aa0000"
+
+/decl/curtain_kind/plastic/canteen
+	name = "privacy curtain"
+	color = COLOR_BLUE_GRAY
+
+
 /obj/item/curtain
```
MistakeNot4892, line 1: While you're in the area it might be worth doing something like
````
/obj/item/curtain/Initialize()
  . = ..()
  var/decl/curtain_kind/curtain_decl = GET_DECL(curtain_kind)
  matter = atom_info_repository.get_matter_for(/obj/structure/curtain, curtain_decl.material_key)
````
Otherwise the recipe cost population might not function correctly.

File: code/game/objects/structures/curtains.dm [hunk 2]
```diff
@@ -1,3 +1,71 @@
+//
+// Curtain types declaration
+//
+/decl/curtain_kind
+	var/name = "curtain"
+	var/color = "white"
+	var/alpha = 255
```
MistakeNot4892: I feel like it might be more appropriate to grab alpha from the material decl (think it's `material.opacity` or something like that) since that would remove the need for a handful of these subtypes.

MistakeNot4892 (commented):
Other than comments this looks good. I feel like in future it might be possible to completely move these properties onto a combination of material decls plus paintable structures but that's something that can be left for later.

code/game/objects/structures/curtains.dm
PsyCommando: Well, the thing is you got some plastic curtains that are transparent and some that aren't. Like the privacy curtain and medical ones. So its not really dependent on material. Its like you  can have cloth curtains with gaps in the cloth that makes them see-through. Its more of a factor of the way the material is used than its innate properties. At least I think.

code/game/objects/structures/curtains.dm
PsyCommando: Is that recipe cost population thing new?

MistakeNot4892: It's Nebula specific but it's something like a year old.

MistakeNot4892: Turns out these items aren't used by the recipes and I'm dumb.

MistakeNot4892 (dismissed):

PsyCommando:
Just noticed I made an extra implementation of initialize for nothing... Fixing that.

PsyCommando:
Not sure why it says there's a runtime, because the logs don't show any..

code/game/objects/structures/curtains.dm
```diff
@@ -6,11 +74,45 @@
...
+/obj/item/curtain/Initialize(ml, material_key, var/decl/curtain_kind/kind = null)
+	if(kind)
+		if(istype(kind))
+			curtain_kind_path = kind.type
+		else if(ispath(kind))
+			curtain_kind_path = kind
+	else
+		kind = GET_DECL(curtain_kind_path)
+	src.alpha = kind.alpha
+	src.set_color(kind.color)
```
MistakeNot4892: Alpha/colour set should be done in on_update_icon() after calling ..() otherwise material updates will potentially overwrite them.

code/game/objects/structures/curtains.dm
```diff
@@ -48,16 +149,27 @@
...
+/obj/structure/curtain/Initialize(ml, _mat, _reinf_mat, var/decl/curtain_kind/kind = null)
+	if(kind)
+		if(istype(kind))
+			curtain_kind_path = kind.type
+		else if(ispath(kind))
+			curtain_kind_path = kind
+	else
+		kind = GET_DECL(curtain_kind_path)
+	_mat = kind.material_key
+	matter = atom_info_repository.get_matter_for(/obj/structure/curtain, kind.material_key)
```
MistakeNot4892: This is an infinite loop - needs to be in/obj/item/curtain/Initialize(), in this instance it's causing an infinite loop of Initialize.

code/game/objects/structures/curtains.dm
```diff
@@ -48,16 +149,27 @@
...
 	. = ..()
+	src.SetName(kind.name)
+	src.alpha = kind.alpha
```
MistakeNot4892: As with the above comment this should probably be moved into an icon update or update_materials override.

MistakeNot4892:
https://github.com/PsyCommando/Nebula-1/pull/1

code/game/objects/structures/curtains.dm
PsyCommando: I'm not sure I understand what you mean here? Does the atom_info_repository instantiate atoms?

PsyCommando: Also, if this creates an infinite loop putting it in the item would just create an infinite loop too no?

PsyCommando:
Found an issue with assembling the curtains and fixed it. Basically, the direction parameter in the new on the curtain structure was shifting the other parameters in the Initialized proc. So I just called set_dir instead.
Also, I just made the matter cost a var in the decl instead. The structure fetch the value from the entry of the item version to keep things coherent.

PsyCommando:
Merged in loaf's pr

PsyCommando:
So yeah, should be good to merge now.

MistakeNot4892 (approved):

PR #1988, opened 2021-08-08
adds ATOM_FLAG_WHEELED and handcarts

Mazianni:
## Description of changes

Adds ATOM_FLAG_WHEELED, alters the calculations involved in encumberance to take this into account. Currently takes 50% off the exertion. 
Adds the handcart, a way to easily move heavy and bulky things around.

The concept of wheels has eluded spaceman, up until now. Utilizing basic physics, productivity will skyrocket!

## Authorship

Mazian/Andromeda-K22
Sprites by Azlan

## Changelog

:cl:
rscadd: handcarts are now a thing. control-click a pullable atom, then click the cart with them to load them on. Only works with ITEMS and OBJECTS, not PEOPLE. Items have a minimum size to be loaded onto the cart.
tweak: wheeled objects should now be given ATOM_FLAG_WHEELED and are easier and less exhausting to move.
/:cl:


code/game/objects/structures/hand_cart.dm
```diff
@@ -0,0 +1,60 @@
...
+	icon_state = "cart"
+	anchored = 0
+	density = 1
+	atom_flags = ATOM_FLAG_WHEELED
+	material_alteration = MAT_FLAG_ALTERATION_NAME | MAT_FLAG_ALTERATION_COLOR
+	material = /decl/material/solid/metal/steel
+	w_class = ITEM_SIZE_NO_CONTAINER
+	var/pixel_y_offset = 4
+	var/pixel_x_offset = 4
+	var/atom/movable/C
```
MistakeNot4892: Would probably be better to use mob buckling for this effect, similar to what rollerbeds do.

MistakeNot4892: Buckling only works for mobs currently, nvm.

code/game/objects/structures/hand_cart.dm
```diff
@@ -0,0 +1,60 @@
...
+/obj/structure/hand_cart/grab_attack(var/obj/item/grab/G)
+	if(G.affecting && istype(G.affecting, /obj/))
+		to_chat(G.assailant, SPAN_NOTICE("You start loading \the [G.affecting] onto \the [src]."))
+		if(load_item(G.affecting, G.assailant))
+			qdel(G)
+
+/obj/structure/hand_cart/attack_hand(mob/user)
+	. = ..()
+	if(C)
+		unload_item(user)
```
MistakeNot4892, line 48: imo
````
	if(C)
		unload_item(user)
		return TRUE
	. = ..()
````

File: code/game/objects/structures/hand_cart.dm [hunk 3]
```diff
@@ -0,0 +1,60 @@
+/obj/structure/hand_cart
```
MistakeNot4892, line 1: Need a Destroy() override to drop or remove C. Also a physically_destroyed override possibly.

code/game/objects/structures/hand_cart.dm
```diff
@@ -0,0 +1,60 @@
...
+		unload_item(user)
+
+/obj/structure/hand_cart/proc/load_item(var/obj/A, var/user)
+	if(!A.anchored && (A.w_class > min_object_size))
+		if(do_after(user, 2 SECONDS, src))
+			A.pixel_y = pixel_y_offset
+			A.pixel_x = pixel_x_offset
+			A.plane = plane
+			A.layer = layer + 0.01 //just above STRUCTURE_LAYER
+			add_overlay(A)
```
MistakeNot4892: This block is slightly odd. Feel like generating an image would be clearer than mutating `A` before taking the appearance.

code/game/objects/structures/hand_cart.dm
PsiOmegaDelta: Rather than `C`, the var should have a name that clarifies its intent/purpose
It's one thing when proc-scoped vars are single-letters, type-level ones should be more descriptive

code/game/objects/structures/hand_cart.dm
```diff
@@ -0,0 +1,70 @@
...
+		var/image/CA = image(carrying.icon, carrying.icon_state)
+		CA.pixel_y = pixel_y_offset
+		CA.pixel_x = pixel_x_offset
+		CA.plane = plane
+		CA.layer = layer + 0.01 //just above STRUCTURE_LAYER
+
+/obj/structure/hand_cart/Destroy()
+	. = ..()
+	if(carrying)
+		carrying.forceMove(get_turf(src))
```
MistakeNot4892, line 35: Need to null `carrying`.

code/game/objects/structures/hand_cart.dm
```diff
@@ -0,0 +1,70 @@
...
+
+/obj/structure/hand_cart/grab_attack(var/obj/item/grab/G)
+	if(G.affecting && istype(G.affecting, /obj/))
+		to_chat(G.assailant, SPAN_NOTICE("You start loading \the [G.affecting] onto \the [src]."))
+		if(load_item(G.affecting, G.assailant))
+			qdel(G)
+
```
MistakeNot4892, line 45: `return ..()`

code/game/objects/structures/hand_cart.dm
```diff
@@ -0,0 +1,70 @@
...
+/obj/structure/hand_cart/Destroy()
+	. = ..()
+	if(carrying)
+		carrying.forceMove(get_turf(src))
+
+/obj/structure/hand_cart/grab_attack(var/obj/item/grab/G)
+	if(G.affecting && istype(G.affecting, /obj/))
+		to_chat(G.assailant, SPAN_NOTICE("You start loading \the [G.affecting] onto \the [src]."))
+		if(load_item(G.affecting, G.assailant))
+			qdel(G)
```
MistakeNot4892, line 42: `return TRUE` possibly.

code/game/objects/structures/hand_cart.dm
```diff
@@ -0,0 +1,70 @@
...
+	var/pixel_y_offset = 4
+	var/pixel_x_offset = 4
+	var/atom/movable/carrying
+	var/min_object_size = ITEM_SIZE_NORMAL
+
+/obj/structure/hand_cart/on_update_icon()
+	underlays.Cut()
+	cut_overlays()
+	var/image/W = image(icon, "cart_wheel")
+	underlays += W
```
MistakeNot4892: Pretty sure `underlays += "cart_wheel"` would be equivalent here.

code/game/objects/structures/hand_cart.dm
```diff
@@ -0,0 +1,70 @@
...
+	var/image/I = image(icon, "handcart_layer_north")
+	I.layer = STRUCTURE_LAYER + 0.02
+	I.color = BlendRGB(color, material.color, 0.5)
+	add_overlay(I)
+	if(carrying)
+		var/image/CA = image(carrying.icon, carrying.icon_state)
+		CA.pixel_y = pixel_y_offset
+		CA.pixel_x = pixel_x_offset
+		CA.plane = plane
+		CA.layer = layer + 0.01 //just above STRUCTURE_LAYER
```
MistakeNot4892, line 29: `add_overlay(CA)`

code/game/objects/structures/hand_cart.dm
MistakeNot4892: This still applies, I don't think you need to create an image at all.

code/game/objects/structures/hand_cart.dm
MistakeNot4892: Definitely needs some kind of return to prevent the rest of the grab logic acting on a qdeleted grab.

code/game/objects/structures/hand_cart.dm
```diff
@@ -0,0 +1,70 @@
...
+		carrying.forceMove(get_turf(src))
+		carrying = null
+
+/obj/structure/hand_cart/grab_attack(var/obj/item/grab/G)
+	if(G.affecting && istype(G.affecting, /obj/))
+		to_chat(G.assailant, SPAN_NOTICE("You start loading \the [G.affecting] onto \the [src]."))
+		if(load_item(G.affecting, G.assailant))
+			qdel(G)
+		return TRUE
+	..()
```
MistakeNot4892: `. = ..()`

code/game/objects/structures/hand_cart.dm
```diff
@@ -0,0 +1,70 @@
...
+
+/obj/structure/hand_cart/attack_hand(mob/user)
+	if(carrying)
+		unload_item(user)
+	. = ..()
+
+/obj/structure/hand_cart/proc/load_item(var/obj/A, var/user)
+	if(!A.anchored && (A.w_class > min_object_size))
+		if(do_after(user, 2 SECONDS, src))
+			var/image/I = image(A.icon, A.icon_state)
```
MistakeNot4892, line 54: Not going to block to request this since you might actually straight up die, but using vis_contents would be preferrable to a straight copy of the base icon considering how many structures etc. use composite overlays and emissives.

MistakeNot4892:
Will approve after the `. = ..()` is added.

code/game/objects/structures/hand_cart.dm
MistakeNot4892: As an alternative to vis_contents you could use a mutable appearance and take the appearance of `A`.

MistakeNot4892 (approved):

PR #1993, opened 2021-08-12
Decouples AI icons from the decl repository.

MistakeNot4892:
Fixes a runtime with AI creation and the last extant non-decl usage of the decl repo (\o/).


afterthought2 (approved):
I don't love it but OK.

PR #1995, opened 2021-08-12
Fixes accept_drinking on reagent dispensers

MarinaGryphon:


code/modules/reagents/dispenser/dispenser2.dm
```diff
@@ -91,7 +91,7 @@
 			C.dropInto(loc)
 			return TRUE
 
-	if(istype(W, /obj/item/chems/glass) || istype(W, /obj/item/chems/food))
+	if(istype(W, /obj/item/chems/glass) || istype(W, /obj/item/chems/condiment) || istype(W, /obj/item/chems/drinks)
```
MarinaGryphon:
```suggestion
	if(istype(W, /obj/item/chems/glass) || istype(W, /obj/item/chems/condiment) || istype(W, /obj/item/chems/drinks))
```

MistakeNot4892 (approved):
Should make this a var on /chems instead of using this istype block.

PR #2008, opened 2021-08-16
Keybindings are now decls.

MistakeNot4892:
## Description of changes
Repaths keybindings to decls, removes now unused global list population hook.

decl good

## Authorship
Me.

## Changelog
Nothing player-facing.


MistakeNot4892:
Very broken, pls hold.

MistakeNot4892:
This has turned out to be a huge pain due to the really strange way keybindings are coded, may revisit in the future.

PR #2009, opened 2021-08-17
Fixes movement

MarinaGryphon:
## Description of changes

Fixes movement no longer working in one direction if you stop holding that direction key.
Prioritises the buckling message over the anchoring message in `/datum/movement_handler/mob/physically_restrained/MayMove()`

#2007 introduced a bug where you would no longer be able to move in a direction after raising that direction key due to an oversight in SSInput. This has more robust logic to avoid that.
Also stops chairs from spamming messages at you, and makes it show the right message if you try to move while buckled in a chair.

## Authorship
MoondancerPony for fixes, Afterthought for the suggestion to move the var resets to `keyLoop`

### DIFF
--- a/code/datums/movement/mob.dm
+++ b/code/datums/movement/mob.dm
@@ -173,14 +173,14 @@
 
 // Is anything physically preventing movement?
 /datum/movement_handler/mob/physically_restrained/MayMove(var/mob/mover)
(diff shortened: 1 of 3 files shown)


MistakeNot4892 (approved):
Needs squashmerge.

PR #1912, opened 2021-07-15
Adds kaboom debug command for iterative explosions

quardbreak:
## Description of changes

Adds kaboom debug command for iterative explosions

## Why and what will this PR improve

kaboom

## Authorship

Lohikar

PR #2019, opened 2021-08-21
Removing /simulated typing from iterative explosions.

MistakeNot4892:
This doesn't fix them but it needs to be done as far as I can tell.

Implements an alternative to #1912.

### DIFF
--- a/code/game/objects/explosion.dm
+++ b/code/game/objects/explosion.dm
@@ -102,7 +102,7 @@
 #define SEARCH_DIR(dir) \
 	search_direction = dir;\
 	search_turf = get_step(current_turf, search_direction);\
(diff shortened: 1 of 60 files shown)


code/game/objects/explosion.dm
```diff
@@ -102,7 +102,7 @@
 #define SEARCH_DIR(dir) \
 	search_direction = dir;\
 	search_turf = get_step(current_turf, search_direction);\
-	if (istype(search_turf, /turf/simulated)) {\
+	if (istype(search_turf, /turf)) {\
```
afterthought2: isturf

afterthought2 (approved):

PR #2032, opened 2021-08-27
Add master list of ignored revisions

PsiOmegaDelta:
Works with GitLens in Visual Studio Code, due to the provided settings
Works with `git blame` if one execute the command listed in  `.git-blame-ignore-revs`

May eventually be supported natively on GitHub - Or at least there is a community request to add support

In case anyone is interested, I borrowed the header in `.git-blame-ignore-revs` from https://chromium.googlesource.com/chromium/src.git/+/refs/heads/main/.git-blame-ignore-revs

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,27 @@
+# git blame master ignore list.
+#
+# This file contains a list of git hashes of revisions to be ignored by git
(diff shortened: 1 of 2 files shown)


MistakeNot4892 (approved):
Nice, I didn't know you could do this.

PR #2033, opened 2021-08-29
Camera networkification

NataKilar:
## Description of changes

Moves security cameras over to computer networks for ease of runtime configuration. Moves most camera-like behavior over to a subtype of network device extensions as a replacement to the old method of storing a security camera inside of another objects contents. Moves access off of camera networks and onto individual cameras for parity with other machinery. Renames camera networks to camera channels for clarity with computer networks. Network devices will now only attempt to connect to a network automatically a limited amount of times to prevent clogging up the queue.

For now, much of the cameranet/visualnet functionality completely ignores the computer network the camera is tied to. In the future, AI can be tied more closely to computer networks rather than having magical connections.

The alarm system has been left largely untouched, as tying alarms to computer networks is beyond the scope of this PR and may not even be necessary.

This change enables players to build and replace cameras at runtime and tie them to channels as desired. It will also allow camera access to be customized. In the future, this will also allow for multiple cameranets to exist in a given round.

## Authorship

Myself

## Changelog

:cl:
tweak: Camera networks are now known as channels and are not tied to access.
rscadd: Security cameras are now tied directly to computer networks, and can be managed as with other network devices.
/:cl:
-->


code/modules/modular_computers/file_system/programs/generic/camera.dm
```diff
@@ -1,67 +1,54 @@
...
 	filedesc = "Camera Monitoring"
 	nanomodule_path = /datum/nano_module/program/camera_monitor
 	program_icon_state = "cameras"
 	program_key_state = "generic_key"
 	program_menu_icon = "search"
-	extended_desc = "This program allows remote access to the camera system. Some camera networks may have additional access requirements."
+	extended_desc = "This program allows remote access to the camera system. Some cameras may have additional access requirements."
 	size = 12
 	available_on_network = 1
+	requires_network = 1
```
comma: Do you have to set this? Its behaviour is pretty gross, just killing program whenever network connection is lost, mostly left on programs that can't handle gracefully not getting any data or network.

comma:
CI fails on tradeship with 
```
[MAP: Camera nil c_tag check]: Following camera does not have a c_tag set: The security camera (/obj/machinery/camera) ([0x200761a]) (NULL) (NULL) (NULL)
```

code/game/objects/items/devices/spy_bug.dm
```diff
@@ -95,69 +92,53 @@
...
+	events_repository.unregister(/decl/observ/destroyed, SB, src, .proc/unpair)
+	if(selected_camera == SB)
+		selected_camera = null
+	cameras -= SB
 
 /obj/item/spy_monitor/proc/view_cameras(mob/user)
-	if(!can_use_cam(user))
+	if(!cameras.len)
+		to_chat(user, SPAN_WARNING("No paired cameras detected!"))
+		to_chat(user, SPAN_WARNING("Bring a bug in contact with this device to pair the camera.</span>"))
```
comma: extra closing tag

code/modules/modular_computers/file_system/programs/generic/camera.dm
NataKilar: Resolved. The program itself will still require a network connection to connect to any cameras, even for television cameras, which makes sense to me but this can be changed if desired.

code/__defines/machinery.dm
```diff
@@ -41,23 +41,23 @@ var/global/defer_powernet_rebuild = 0      // True if net rebuild will be called
...
+
+#define CHANNEL_ROBOTS "Robots"
+#define CHANNEL_MINE "Mining"
+#define CHANNEL_SECRET "Secret"
+
+// Non-station channels
+#define CHANNEL_CRESCENT "Crescent"
+#define CHANNEL_ERT "ZeEmergencyResponseTeam"
+#define CHANNEL_MERCENARY "MercurialNet"
+#define CHANNEL_TELEVISION "Television"
```
MistakeNot4892: Maybe CAMERA_CHANNEL to avoid confusion with radio channels?

code/controllers/subsystems/initialization/computer_networks.dm
```diff
@@ -12,6 +14,8 @@ SUBSYSTEM_DEF(networking)
 	var/list/connection_queue = list()
 	var/list/tmp_queue = list()
 
+	var/datum/extension/network_device/last_failed_device
```
MistakeNot4892: Ctrl-F isn't showing if you've added checking/unsetting this in Destroy() for the type but if not please do so.

code/datums/wires/camera.dm
```diff
@@ -37,12 +38,17 @@ var/global/const/CAMERA_WIRE_NOTHING2 = 32
...
-			var/range = (mended ? initial(C.view_range) : C.short_range)
-			C.setViewRange(range)
+			var/datum/extension/network_device/camera/D = get_extension(holder, /datum/extension/network_device)
+			var/new_range = (mended ? C.long_range : C.short_range)
+			D.set_view_range(new_range)
 
 		if(CAMERA_WIRE_POWER)
-			if(C.status && !mended || !C.status && mended)
-				C.deactivate(usr, 1)
+			if(!mended)
```
MistakeNot4892: Could just do:
````
C.cut_power = !mended
C.set_status(mended, usr)
````

code/controllers/subsystems/initialization/computer_networks.dm
NataKilar: This was an artifact from testing, thanks for catching it. I went ahead and removed it.

MistakeNot4892 (dismissed):
Let's do it.

comma (approved):

PR #2034, opened 2021-08-30
Adds struts

Mazianni:
## Description of changes

Adds struts, a material stack intended to reflect a 'unistrut' like material stack type.

## Why and what will this PR improve

It doesn't make a lot of sense to bend metal into girders with your BEAR hands.

## Authorship

@MistakeNot4892 / @Andromeda-K22 

## Changelog

:cl:
rscadd: adds the 'strut' material stack, used for a number of items.
tweak: metal sheets are no longer used for many things, struts are used instead.
/:cl:

PR #2035, opened 2021-08-30
Adjusted structure debris gen to use parts_type for rods.

MistakeNot4892:
Moves rod and catwalk deconstruction into the shared proc instead of doing it bespoke in overrides. Needed for #2034.

### DIFF
--- a/code/game/objects/structures/_structure_materials.dm
+++ b/code/game/objects/structures/_structure_materials.dm
@@ -56,14 +56,22 @@
 
 /obj/structure/proc/create_dismantled_products(var/turf/T)
 	SHOULD_CALL_PARENT(TRUE)
-	if(parts_type)
-		new parts_type(T, (material && material.type), (reinf_material && reinf_material.type))
+	if(parts_type && !ispath(parts_type, /obj/item/stack))
+		for(var/i = 1 to max(parts_amount, 1))
+			new parts_type(T, (material && material.type), (reinf_material && reinf_material.type))
 	else
 		for(var/mat in matter)
 			var/decl/material/M = GET_DECL(mat)
-			var/placing = FLOOR((matter[mat] / SHEET_MATERIAL_AMOUNT) * 0.75)
+			var/placing
+			if(isnull(parts_amount))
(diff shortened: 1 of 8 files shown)


code/game/objects/structures/_structure_materials.dm
```diff
@@ -56,14 +56,22 @@
...
+			if(isnull(parts_amount))
+				placing = (matter[mat] / SHEET_MATERIAL_AMOUNT) * 0.75
+				if(parts_type)
+					placing *= atom_info_repository.get_matter_multiplier_for(parts_type, mat, placing)
+				placing = FLOOR(placing)
+			else
+				placing = parts_amount
 			if(placing > 0)
-				M.place_dismantled_product(T, amount = placing)
+				M.place_dismantled_product(T, FALSE, placing, parts_type)
```
afterthought2, line 74: I feel it would make more sense to feed an absolute amount of matter into this proc, and then have it adjust the stack size produced so that matter is conserved. Alternatively, to wrap this proc in a variant that does what I describe (and what the logic above accomplishes, though perhaps without the caching). Is that radically incompatible with existing systems?

MistakeNot4892: That would also be feasible, but would largely just be shifting some code around. Structures are the only things that call this proc currently. Can do it if you like, not fussed.

afterthought2: My worry is that having to reproduce the code above each time I want to rods is undesirable.

MistakeNot4892: Why would you need to?

MistakeNot4892: Having looked at the code I don't really want to move the stack logic onto the material. Currently `parts_amount` will bypass the matter value checking and caching and I would need to multiply that or do some other math to munge it into an appropriate matter amount. Not really seeing a benefit or a problem with the current setup given that nothing except structures use this proc, and the other uses are not in a context that will easily take a matter value.

afterthought2 (approved):
OK

PR #2031, opened 2021-08-27
Adds mob riding for small mobs and simple_animals.

MistakeNot4892:
Allows corgis and possums to be used as mounts by small mobs.

image

PR #2036, opened 2021-08-30
Minor movement handler refactor.

MistakeNot4892:
Split out of #2031, just some housekeeping. The removed type is half-implemented and was intended for borers before Psi realized they swap clients instead of relaying movement.

### DIFF
--- a/code/datums/movement/mob.dm
+++ b/code/datums/movement/mob.dm
@@ -1,24 +1,3 @@
-// Movement relayed to self handling
-/datum/movement_handler/mob/relayed_movement
-	var/prevent_host_move = FALSE
-	var/list/allowed_movers
-
-/datum/movement_handler/mob/relayed_movement/MayMove(var/mob/mover, var/is_external)
-	if(is_external)
-		return MOVEMENT_PROCEED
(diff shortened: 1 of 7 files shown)


PsiOmegaDelta:
I'll note that the removed handler should prevent mobs that somehow ended up inside other mobs from moving their host, unless allowed.

I'd verify that mobs held by other mobs as well as passengers (not pilots) inside mechs don't have unintentional control.

Someone unwillingly buckled to another mob, which might be a future feature, would also be potentially able to control whoever is carrying them around.

MistakeNot4892:
Mobs held by mobs are wrapped in items - as far as I am aware there is no intentional case where a mob is inside another mob other than mechs, which implement relaymove. There's no capacity to have passengers in a mech as far as I recall, just multiple pilots.

afterthought2 (approved):
If mobs directly inside mobs are added, they can reference this and build on it as needed.

PsiOmegaDelta:
Well, the removed handler already prevent buckled mobs from forcefully moving a mob that they're buckled to (unless added to the list of allowed mobs)

PR #2037, opened 2021-08-30
Unifying mob pixel offsets in a single proc.

MistakeNot4892:
## Description of changes
Split off from #2031. Unifies all extant common pixel offsets for mobs into one giant horrible proc, so that they stop interfering with each other and can be rebuilt consistently. Various other changes also included to make it work, including grab and buckling changes to stuff like the janicart.

Currently buckled offsets, grab offsets and structure offsets compete with each other and the only point of references is `default_pixel_foo`, meaning that you can't stack offsets. This PR resolves that.

## Authorship
Myself.

## Changelog
Nothing player facing.


code/modules/vehicles/cargo_train.dm
```diff
@@ -34,12 +33,16 @@
 	load_item_visible = 1
 	load_offset_x = 0
 	load_offset_y = 4
-	buckle_pixel_shift = @"{'x':0,'y':0,'z':8}"
+
+/obj/vehicle/train/cargo/trolley/Initialize()
+	buckle_pixel_shift = list("x" = 0, "y" = 0, "z" = 8)
```
comma: why do this in init instead of var declaration?

MistakeNot4892: List declarations at compile time create an invisible magic `(init)` proc.

MistakeNot4892: Would be fine, just realized these ones aren't directional so are nominally constant for the compiler's sake.

code/datums/movement/mob.dm
```diff
@@ -113,7 +113,7 @@
 	if(mob.buckled) // Wheelchair driving!
 		if(isspaceturf(mob.loc))
 			return // No wheelchair driving in space
-		if(istype(mob.buckled, /obj/structure/bed/chair/wheelchair))
+		if(istype(mob.buckled, /obj/structure/bed/chair/wheelchair) || istype(mob.buckled, /obj/structure/bed/chair/janicart))
```
afterthought2: I get it's not a major part of this PR but yuk.

code/game/objects/buckling.dm
```diff
@@ -57,19 +58,18 @@
...
-			var/list/pixel_shift = cached_json_decode(buckle_pixel_shift)
-			animate(M, pixel_x = M.default_pixel_x + pixel_shift["x"], pixel_y = M.default_pixel_y + pixel_shift["y"], pixel_z = M.default_pixel_z + pixel_shift["z"], 4, 1, LINEAR_EASING)
-		else
-			animate(M, pixel_x = M.default_pixel_x, pixel_y = M.default_pixel_y, pixel_z = M.default_pixel_z, 4, 1, LINEAR_EASING)
-
-/mob/proc/can_be_buckled(var/mob/user)
-	. = user.Adjacent(src) && !istype(user, /mob/living/silicon/pai)
+	SHOULD_CALL_PARENT(TRUE)
+	reset_plane_and_layer()
+	reset_offsets(5)
```
afterthought2: 5? Wasn't it 4 before?

code/game/objects/structures/__structure.dm
```diff
@@ -145,25 +145,38 @@
...
 
 /obj/structure/Move()
+	var/turf/T = get_turf(src)
 	. = ..()
 	if(. && !CanFluidPass())
 		fluid_update()
+		if(T)
+			for(var/atom/movable/AM in T)
+				AM.reset_offsets()	
+				AM.reset_plane_and_layer()
```
afterthought2, line 179: Is this actually correct? If I delete an unrelated structure on a turf where a movable is offset for some other unrelated reason, will it retain its state? For example, consider maploaded wallshifted objs.

code/modules/mob/animations.dm
```diff
@@ -68,18 +70,13 @@
...
-		default_pixel_x = mob.default_pixel_x
-		default_pixel_y = mob.default_pixel_y
-
 	var/matrix/initial_transform = matrix(transform)
 	var/matrix/rotated_transform = transform.Turn(15 * turn_dir)
 
 	animate(src, pixel_x = pixel_x + pixel_x_diff, pixel_y = pixel_y + pixel_y_diff, transform = rotated_transform, time = 2, easing = BACK_EASING | EASE_IN)
-	animate(pixel_x = default_pixel_x, pixel_y = default_pixel_y, transform = initial_transform, time = 2, easing = SINE_EASING)
+	animate(pixel_x = pixel_x, pixel_y = pixel_y, transform = initial_transform, time = 2, easing = BACK_EASING | EASE_IN)
+	sleep(4)
```
afterthought2, line 78: Is 4 actually right, rather than `time`?

code/modules/mob/grab/grab_object.dm
```diff
@@ -48,7 +48,7 @@
 	LAZYADD(affecting.grabbed_by, src) // This is how we handle affecting being deleted.
 	adjust_position()
 	action_used()
-	assailant.do_attack_animation(affecting)
+	INVOKE_ASYNC(assailant, /atom/movable/proc/do_attack_animation, affecting)
```
afterthought2, line 51: It's been made waitfor-false.

code/modules/mob/mob_layering.dm
```diff
@@ -0,0 +1,102 @@
...
+	if (hiding)
+		return HIDING_MOB_LAYER
+	. = ..()
+
+/mob/living/carbon/human/get_base_layer()
+	if(lying)
+		return LYING_HUMAN_LAYER
+	. = ..()
+
+/mob/reset_offsets(var/anim_time = 2)
```
afterthought2, line 56: If I understand your approach correctly, maybe worth commenting here (and above for layer) that if anything ever touches mob offsets, it must do so here and here alone, and it must call this proc to enact whatever change it desires.

I don't necessarily love that, but I can see how it's much better than stuff setting these ad-hoc.

code/game/objects/structures/__structure.dm
MistakeNot4892: Wallshifted objs should be setting their default pixel offsets as well as their pixel offsets though.

code/modules/mob/animations.dm
MistakeNot4892: It uses a time of 2, twice. Not sure what else I would sleep for.

code/modules/mob/grab/grab_object.dm
MistakeNot4892: Linter doesn't care. :(

code/modules/mob/mob_layering.dm
MistakeNot4892: Alrighty, will add a comment.

code/datums/movement/mob.dm
MistakeNot4892: This is temporary at least, when the other movement tweaks PR goes in I can resolve this to use an override instead.

code/game/objects/structures/__structure.dm
afterthought2: Do any actually do so (particularly mapped ones, but also there are some constructed ones and some that set offset in update_icon).

code/modules/mob/animations.dm
afterthought2: I misunderstood which proc this is.

code/game/objects/structures/__structure.dm
MistakeNot4892: Likely not, will need to go have a look.
Possibly structures/machines should set their default pixel offsets to whatever pixel_x/pixel_y are set to during Initialize()?

code/game/objects/buckling.dm
MistakeNot4892: Yep, my mistake.

code/datums/movement/mob.dm
MistakeNot4892: Rebasing to remove it now.

code/game/objects/structures/__structure.dm
MistakeNot4892: I went ahead and made `/atom/Initialize()` set default pixel offsets if they're not set to anything, and to set pixel offsets otherwise. Between that and updating several of the hardcoded offset atoms like APCs to use default pixel offsets and reset_offsets() I think it should be fine now.

code/modules/mob/mob_layering.dm
MistakeNot4892: Added.

afterthought2 (approved):
All right, just watch for any "when I bombed the medbay X shifted a turf over" bugs.

PR #2042, opened 2021-08-30
Fix runtime deleting mob ai var

PsyCommando:
## Description of changes
In some circumstances mobs would delete their ai var before it was actually replaced by a datum, causing mob/Destroy() to fail on qdeleting a path. Added a check to prevent that, like other type-abused vars do.

## Changelog

:cl:
bugfix: Fix occasional runtime in mob/Destroy() when deleting an ai var that wasn't initialized yet.
/:cl:


MistakeNot4892 (approved):

afterthought2:
I generally object to this pattern; I think its use with skillsets in `Destroy` here is basically erroneous as well, for example. It suggests the entity has been deleted prior to `Initialize` having fully run, which is always incorrect.

PsyCommando:
Its also incorrect to have a destructor throw exceptions. And well type abused vars aren't exactly the cleanest thing ever. I personally like that something even if its broken dies gracefully. 
Plus, its not like it silences the issue in the first place. Deleting non-initialized things already runtimes in qdel.

PR #2043, opened 2021-08-30
Fix APCs sometimes spamming a lot of runtimes when ending up in nullspace

PsyCommando:
## Description of changes

Sometimes when doing admin delete, or some other times when players attempt constructing in space, APCs would end up with a null area, and slow things down spamming runtimes every ticks. Making debugging an absolute pain in the ass. This ensures the APC just stops processing when its not in a valid area instead of hanging machine processing. 
This is just to prevent the runtime spam though, not preventing the APC from being in a null area, which is most likely only an issue on outreach anyways.

## Changelog

:cl:
bugfix: Prevent APC runtime spam when in a null area.
/:cl:

### DIFF
--- a/code/modules/power/apc.dm
+++ b/code/modules/power/apc.dm
@@ -966,10 +966,7 @@ var/global/list/all_apcs = list()
 	//Make sure to resume processing if our area changed to something else than null
 	if(QDELETED(src))
 		return
-	if(area)
-		if(is_processing && !(processing_flags & MACHINERY_PROCESS_SELF))
-			processing_flags |= MACHINERY_PROCESS_SELF //don't use START_PROCESSING_MACHINE here or it'll print a stack trace for being already in the processing list
-		else if(!is_processing)
-			START_PROCESSING_MACHINE(src, MACHINERY_PROCESS_SELF)
+	if(area && !(processing_flags & MACHINERY_PROCESS_SELF))
+		START_PROCESSING_MACHINE(src, MACHINERY_PROCESS_SELF)
 
 #undef APC_UPDATE_ICON_COOLDOWN
--- a/code/modules/power/apc.dm
+++ b/code/modules/power/apc.dm
@@ -576,7 +576,7 @@ var/global/list/all_apcs = list()
 	if (!ui)
 		// the ui does not exist, so we'll create a new() one
 		// for a list of parameters and their descriptions see the code docs in \code\modules\nano\nanoui.dm
-		ui = new(user, src, ui_key, "apc.tmpl", "[area.name] - APC", 520, data["siliconUser"] ? 465 : 440)
+		ui = new(user, src, ui_key, "apc.tmpl", "[area? area.name : "ERROR"] - APC", 520, data["siliconUser"] ? 465 : 440)
 		// when the ui is first opened this is the data it will use
 		ui.set_initial_data(data)
 		// open the new ui window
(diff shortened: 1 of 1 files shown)


MistakeNot4892 (dismissed):

afterthought2:
APCs being in a null area must be valid and not throw errors, so this seems fine. It may be better to intercept area change events, though, so processing resumes if the area is changed.

MistakeNot4892:
Is there a proc for area changes?

MistakeNot4892:
Specifically `/obj/machinery/proc/area_changed(area/old_area, area/new_area)`

PsyCommando:
Isn't there something already in place to detect if machinery is moved, or connected to a power net? I mean in regard to area changes and apcs/power?

afterthought2:
There is (see the base /machinery definition of the proc Loaf mentions) but not in a way which will help you here. Yes, an APC's power components should process without problems as needed when it's moved, but that is unrelated to the main `Process` call here. Machines manage power independently of `Process`.

PsiOmegaDelta:
If an APC is admin-deleted, why would it still be processing?

PsiOmegaDelta:
Alternatively, if APCs somehow regularly end up in nullspace without being deleted, that seems like the actual bug that needs to be fixed.

afterthought2:
All machinery, and potentially any obj, can be spawned in nullspace for the purposes of caching material information, etc. They can be moved to nullspace during shuttle moves. Usually they are then either moved back out or deleted synchronously so no errors are registered, but I don't feel having these be in nullspace should be considered a bug.

Their specific case (asynchronous map load, I guess) may indeed be a bug, but that's for them to judge.

PsyCommando:
> 
> 
> Alternatively, if APCs somehow regularly end up in nullspace without being deleted, that seems like the actual bug that needs to be fixed.

Its not that they end up in literal nullspace. At least most of the time. On admin delete they do, but not when an area gets deleted indirectly. Possibly through some save logic or shuttle/changeturf logic (And yes, you cannot modify an area without removing the apc in it, but that didn't stop people from triggering that problem..)

Its just because there are a few bugs that tend to happen when you allow players to build things from scratch that result in the area being null sometimes. And instead of just giving feedback its in a null area, and should be reinstalled, it just takes down the whole server with it by spamming garbage runtimes every tick. The problem is, there isn't really a cleaner way to do this as it is now? 
We do fix those bugs as we find them, but its just not sustainable to have everything break when they happen.

PsyCommando:
Okay, from my tests this should work unless the area is changed directly. But that's acceptable, since changing it directly is dangerous anyways. 
I could always add a check at the start of Process to grab the area from the turf its on if it finds the area var is null if that's necessary.

code/modules/power/apc.dm
```diff
@@ -961,6 +961,15 @@ var/global/list/all_apcs = list()
...
 
-
+/obj/machinery/power/apc/area_changed()
+	. = ..()
+	//Make sure to resume processing if our area changed to something else than null
+	if(QDELETED(src))
+		return
+	if(area)
+		if(is_processing && !(processing_flags & MACHINERY_PROCESS_SELF))
+			processing_flags |= MACHINERY_PROCESS_SELF //don't use START_PROCESSING_MACHINE here or it'll print a stack trace for being already in the processing list
```
afterthought2: Should only runtime if it's processing on a different SS, which isn't relevant here.

PsyCommando: You're right, I misread the macro -_- Pushing a fix in a bit.

afterthought2 (approved):

PR #2049, opened 2021-09-01
Datumizing the overmap.

MistakeNot4892:
## Description of changes
Converts the overmap from a hardcoded single level to a datum system, allowing for multiple distinct overmaps to exist at once.

This will (in the future!) allow for non-space overmaps; something like landing on a planet from the overmap and then driving a buggy around on a second planetary overmap.

## Authorship
Myself.

## Changelog
Initially, nothing player facing.


code/game/machinery/hologram.dm
```diff
@@ -56,6 +56,7 @@ var/global/const/HOLOPAD_MODE = RANGE_BASED
 	var/base_icon = "holopad-B"
 
 	var/allow_ai = TRUE
+	var/static/list/reachable_overmaps = list(OVERMAP_ID_SPACE)
```
comma, line 59: maybe should just grab whatever overmap is our location is on?
also might need third type for super long range for planet-space comms

MistakeNot4892: Thing is, a single z-level can be in multiple overmaps (exoplanet has an exoplanet loc and presumably a loc on its own overmap) so how do you tell which one to use?

MistakeNot4892: This is no longer relevant since I gave up on multiple overmap entities per site.

MistakeNot4892: I'm going to stick with the list version just so it's predictable what overmap the holopad reaches.

MistakeNot4892:
I think this is now ready to go. Just needs some more testing. I can join, get to the overmap, and pilot the ship into an electrical storm.

comma:
```
## UNIT_TEST ##: [AREA: Areas shall be pure]: Deep Space (/area/overmap) ([0x400001e]) (1,1,5) has an edited name.
## UNIT_TEST ##: !!! FAILURE !!! [AREA: Areas shall be pure]: Found 1 impure area.
```

code/modules/overmap/_overmap.dm
```diff
@@ -0,0 +1,143 @@
...
+	..()
+
+/datum/overmap/proc/generate_overmap()
+	testing("Building overmap [name]...")
+	INCREMENT_WORLD_Z_SIZE
+	assigned_z = world.maxz
+	testing("Putting [name] on [assigned_z].")
+
+	var/area/overmap/A = new map_area_type
+	A.name = name
```
comma: need to add it to area purity test exemptions

MistakeNot4892: Why was this not failing before?

comma: Because it wasn't a thing before, it's new code? Overmap area just used default name

MistakeNot4892: Removed the name setting, overmap datums can just define their own areas properly.

comma (approved):

PR #2059, opened 2021-09-07
Fixes client scaling view problems.

MistakeNot4892:
Noticed client view was locked to 41x41 for reasons, this fixes the issue.

### DIFF
--- a/code/modules/client/client_procs.dm
+++ b/code/modules/client/client_procs.dm
@@ -483,19 +483,31 @@ var/global/list/localhost_addresses = list(
 	to_chat(usr, "xDim: [round(text2num(winsize_string) / divisor)]")
 	to_chat(usr, "yDim: [round(text2num(copytext(winsize_string,findtext(winsize_string,"x")+1,0)) / divisor)]")
 
+var/global/const/MIN_VIEW = 15
+var/global/const/MAX_VIEW = 41
 /client/verb/OnResize()
 	set hidden = 1
 
 	var/divisor = text2num(winget(src, "mapwindow.map", "icon-size")) || world.icon_size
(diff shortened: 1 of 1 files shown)


code/modules/client/client_procs.dm
```diff
@@ -483,19 +483,31 @@ var/global/list/localhost_addresses = list(
...
-	if(!isnull(config.lock_client_view_x) && !isnull(config.lock_client_view_y))
-		last_view_x_dim = config.lock_client_view_x
-		last_view_y_dim = config.lock_client_view_y
-	else
-		var/winsize_string = winget(src, "mapwindow.map", "size")
-		last_view_x_dim = config.lock_client_view_x || Clamp(CEILING(text2num(winsize_string) / divisor), 15, config.max_client_view_x || 41)
-		last_view_y_dim = config.lock_client_view_y || Clamp(CEILING(text2num(copytext(winsize_string,findtext(winsize_string,"x")+1,0)) / divisor), 15, config.max_client_view_y || 41)
-		if(last_view_x_dim % 2 == 0) last_view_x_dim++
-		if(last_view_y_dim % 2 == 0) last_view_y_dim++
+	var/list/view_components = splittext(winget(src, "mapwindow.map", "size"), "x")
```
afterthought2, line 492: I believe repeated winget calls like this can be condensed into single calls querying multiple properties; check the syntax. This can be relevant as winget is extremely slow.

MistakeNot4892: I am reluctant to combine these calls in this PR, I don't really want to have to wrestle with parameter ordering etc.

afterthought2 (approved):

PR #2065, opened 2021-09-07
Reworking skybox scaling to take client scale into account.

MistakeNot4892:
## Description of changes
- Changes skybox scaling to be more reliable and to make use of scaling procs.
- Changes skybox parallax calc to take rescaled skybox into account.

Still working on getting the offsets to actually work, this is WIP and also depends on #2059. I am struggling with the math involved in calculating the scale and offset to center the skybox and allow sufficient wiggle room, so any input is appreciated.

This will fix the skybox looking far too huge or not being large enough to cover the backdrop.

## Authorship
Myself.

## Changelog
Nothing player facing.

### DIFF
--- a/code/_onclick/hud/skybox.dm
+++ b/code/_onclick/hud/skybox.dm
@@ -1,5 +1,4 @@
-#define SKYBOX_MAX_BOUND 736
-
+var/global/const/SKYBOX_DIMENSION = 736 // Largest measurement for icon sides, used for offsets/scaling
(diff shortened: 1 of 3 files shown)


MistakeNot4892:
Can't get this to work and triaging my projects, closing.

quardbreak:
I was asked to test it. Seems work fine, but I having a weird `Cannot modify null.screen_loc,` error.

PR #2071, opened 2021-09-09
Reimplementing geothermal power from Europa.

MistakeNot4892:
## Description of changes

Readds geothermal generators using Skowron's icons from the original Europa project. Geothermal generators build over geothermal vents produce power. Adjacent generators will propagate pressure through each other until it is expended. Basic generators over one vent will produce roughly 30k power.

Just a fun alternative to solars. Nothing implements the extension needed to power the generators currently.

## Authorship
Myself for the code, Skowron for the icons.

## Changelog

:cl:
rscadd: Ports geothermal generators from Europa.
/:cl:


MistakeNot4892:
https://cdn.discordapp.com/attachments/678826323728400394/885436053140418570/pS0O21hPIp.mp4

code/modules/power/geothermal/_geothermal.dm
```diff
@@ -0,0 +1,103 @@
...
+	var/tmp/processing = FALSE
+	var/efficiency = 0.5
+
+/obj/machinery/power/geothermal/RefreshParts()
+	..()
+	efficiency = Clamp(total_component_rating_of_type(/obj/item/stock_parts/capacitor) * GEOTHERMAL_EFFICIENCY_MOD, GEOTHERMAL_EFFICIENCY_MOD, 1)
+
+/obj/machinery/power/geothermal/proc/add_pressure(var/pressure)
+	current_pressure = Clamp(current_pressure + pressure, 0, MAX_GEOTHERMAL_PRESSURE)
+	if(!processing)
```
comma: can just check `is_processing`, it's set to SS when processing and null when not

code/modules/power/geothermal/_geothermal.dm
```diff
@@ -0,0 +1,103 @@
...
+		processing = TRUE
+		START_PROCESSING(SSmachines, src)
+
+/obj/machinery/power/geothermal/Process()
+	if(anchored && !(stat & BROKEN) && loc)
+		var/lost_pressure = current_pressure * 0.5
```
comma: this looks like a lot of calc but basic gist is that we 
1. Drop half of pressure no questions asked
2. Rest is distributed between lost/remaining based on efficiency

right?

Feels weird that we /always/ just toss out half of pressure, why not just have lower number in first place?
then could just have
```
lost_pressure = current_pressure * efficiency
remaining_pressure = current_pressure - lost_pressure 
```

Also this means that efficient turbines do not propogate anything down the chain, is that intended?

code/modules/power/geothermal/_geothermal.dm
```diff
@@ -0,0 +1,103 @@
...
+/obj/machinery/power/geothermal/proc/propagate_pressure(var/remaining_pressure)
+	var/list/neighbors
+	for(var/neighbordir in global.cardinal)
+		var/obj/machinery/power/geothermal/neighbor = (locate() in get_step(loc, neighbordir))
+		if(neighbor?.anchored && !(neighbor.stat & BROKEN))
+			LAZYADD(neighbors, neighbor)
+	if(LAZYLEN(neighbors))
+		if(remaining_pressure)
+			for(var/obj/machinery/power/geothermal/neighbor AS_ANYTHING in neighbors)
+				neighbor.add_pressure(remaining_pressure)
```
comma, line 78: also we're giving full remaining_pressure to each, not spreading between them?
that'd make building more vents around central one plain better, which sounds okay from gameplay perspective, just kinda weird physics wise

MistakeNot4892: Should be divided by length of neighbors, not sure where that code went.

code/modules/power/geothermal/_geothermal.dm
MistakeNot4892: Process is:
- Take some portion of our total pressure.
- Turn some portion of that into power.
- Propagate whatever is left to our neighbors.

Totally efficient turbines eating the entire pressure chunk is fine imo.

Making the amount used each tick depend on efficiency means that efficiency primarily just determines the speed that pressure is used, rather than making it a lossier process, you will still get all the power out of a steam burst, just over a longer period.

code/modules/power/geothermal/_geothermal.dm
```diff
@@ -0,0 +1,131 @@
...
+	var/efficiency = 0.5
+
+/obj/machinery/power/geothermal/RefreshParts()
+	..()
+	efficiency = Clamp(total_component_rating_of_type(/obj/item/stock_parts/capacitor) * GEOTHERMAL_EFFICIENCY_MOD, GEOTHERMAL_EFFICIENCY_MOD, 1)
+
+/obj/machinery/power/geothermal/proc/add_pressure(var/pressure)
+	current_pressure = Clamp(current_pressure + pressure, 0, MAX_GEOTHERMAL_PRESSURE)
+	if(!is_processing)
+		START_PROCESSING(SSmachines, src)
```
afterthought2: Needs to be `START_PROCESSING_MACHINE(src, MACHINERY_PROCESS_SELF)`, otherwise it won't actually do what you think.

code/modules/power/geothermal/_geothermal.dm
```diff
@@ -0,0 +1,131 @@
...
+	if(last_neighbors != neighbors)
+		update_icon()
+
+/obj/machinery/power/geothermal/Initialize()
+	. = ..()
+	refresh_neighbors()
+	for(var/turf/T AS_ANYTHING in RANGE_TURFS(loc, 1))
+		for(var/obj/machinery/power/geothermal/neighbor in T)
+			neighbor.refresh_neighbors()
+	STOP_PROCESSING(SSmachines, src)
```
afterthought2: Here you should do `STOP_PROCESSING_MACHINE(src, MACHINERY_PROCESS_SELF)`.

code/modules/power/geothermal/geothermal_extension.dm
```diff
@@ -0,0 +1,34 @@
...
+	var/tmp/next_steam = 0
+	var/datum/effect/effect/system/steam_spread/steam
+
+/datum/extension/geothermal_vent/New()
+	..()
+	START_PROCESSING(SSobj, src)
+
+/datum/extension/geothermal_vent/Destroy()
+	. = ..()
+	STOP_PROCESSING(SSobj, src)
```
afterthought2: I guess there's nothing wrong with this per se, but traditionally there's a generic processing SS for things which aren't objs. Don't personally care much.

afterthought2:
My review is addressed; I'll let @comma approve as he understands the core logic better.

comma (approved):

PR #2074, opened 2021-09-10
Converts bodytype_restricted to a bitfield.

MistakeNot4892:
## Description of changes
Someone on Bay linked https://github.com/BeeStation/BeeStation-Hornet/pull/5363 and it reminded me to do this. This PR swaps the `bodytype_restricted` list out for a bitfield. Means there can be an effective maximum of 23 restriction types but current code only has 7 so I don't think it'll be an issue unless Virgo tries to make a Neb fork.

Decreased init time and memory footprint for clothing.

## Authorship
Myself.

## Changelog
Nothing player facing.

### DIFF
--- a/code/__defines/mobs.dm
+++ b/code/__defines/mobs.dm
@@ -256,6 +256,10 @@
 #define SPECIES_ALIEN            "Humanoid"
 #define SPECIES_GOLEM            "Golem"
 
+#define BODY_FLAG_EXCLUDE        BITFLAG(0)
+#define BODY_FLAG_HUMANOID       BITFLAG(1)
+#define BODY_FLAG_MONKEY         BITFLAG(2)
+
 #define BODYTYPE_HUMANOID        "humanoid body"
 #define BODYTYPE_OTHER           "alien body"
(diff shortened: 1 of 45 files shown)


afterthought2:
I don't think I'm into it. Yes, I get it's less memory, etc, but it also means you have to maintain separate bitflag lists for bodytypes, and now there are multiple ways to check for bodytype, and bodytype data can come in several "formats", and it's messier overall. This reminds me of those species flags that chems used.

MistakeNot4892:
Bodytype category is a string that is used in icon state selection, it's conceptually distinct from equipment. The bitfield I'm adding can in theory handle granular stuff like NO_GLOVES or NO_SHOES without needing a completely separate value across bodytypes.

I don't think the IS_ALIEN stuff is a good comparison because that's a single constant used for hard-checking in reagent procs, not a generic system like this.

Kapu1178:
I was going to do this in my Bee pr, thanks for the assistance homie. Love your work.

afterthought2 (approved):

PR #2095, opened 2021-09-14
Changes the VV type prompt when setting a var's type to be more useful

PsyCommando:
## Description of changes

Made the prompt actually ask for a type path string instead of freezing the whole server while it builds an horrible listbox with all the atom types in the game. As a bonus you can also use datum types now, which is incredibly handy for debugging and admining.

## Changelog

:cl:
tweak: Makes changing a var type in VV take a type path string instead of having the slow nearly useless listbox with all the atom types listed in.
/:cl:

### DIFF
--- a/code/modules/admin/callproc/callproc.dm
+++ b/code/modules/admin/callproc/callproc.dm
@@ -139,8 +139,9 @@
 				if(isnull(current)) return CANCEL
 
 			if("type")
-				current = input("Select type for [arguments.len+1]\th argument") as null|anything in typesof(/obj, /mob, /area, /turf)
+				current = input("Enter type path for [arguments.len+1]\th argument") as null|text
 				if(isnull(current)) return CANCEL
+				current = text2path(current)
 
 			if("obj reference")
 				current = input("Select object for [arguments.len+1]\th argument") as null|obj in world
--- a/code/modules/admin/callproc/callproc.dm
+++ b/code/modules/admin/callproc/callproc.dm
@@ -142,6 +142,9 @@
 				current = input("Enter type path for [arguments.len+1]\th argument") as null|text
 				if(isnull(current)) return CANCEL
 				current = text2path(current)
+				if(!ispath(current))
+					to_chat("Inputed a bad path: '[current]'")
+					return CANCEL
 
 			if("obj reference")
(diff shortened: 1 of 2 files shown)


code/modules/admin/callproc/callproc.dm
```diff
@@ -139,8 +139,9 @@
 				if(isnull(current)) return CANCEL
 
 			if("type")
-				current = input("Select type for [arguments.len+1]\th argument") as null|anything in typesof(/obj, /mob, /area, /turf)
+				current = input("Enter type path for [arguments.len+1]\th argument") as null|text
 				if(isnull(current)) return CANCEL
+				current = text2path(current)
```
afterthought2: You need to ispath check this (and ispath(current, /atom) possibly, as you could inject other paths), as text2path has weird fail behavior. On older BYOND versions you'd have to do this in a try-catch block, but I believe that's no longer the case.

PsyCommando: The big reason I added that was to be able to pass datums, sooo. Also that's a debugging tool, which expects whoever uses it to know what they're doing with it?

afterthought2: Passing datums is fine, just do `ispath(foo)` before proceeding in case they didn't enter a valid path at all. I agree it doesn't need to be super safe, but at the same time no reason to cause unnecessary runtimes when the user misspells things. Can even to_chat them that the path wasn't found or whatever.

PsyCommando: Alright then. I'll do that!

code/modules/admin/callproc/callproc.dm
```diff
@@ -143,7 +143,7 @@
 				if(isnull(current)) return CANCEL
 				current = text2path(current)
 				if(!ispath(current))
-					to_chat("Inputed a bad path: '[current]'")
+					to_chat(usr, "Inputed a bad path: '[current]'")
```
MistakeNot4892: `current` will be null post text2path if it didn't produce a path I believe, might want to keep the raw entered text to print here, otherwise printing it is never going to show anything.

PsyCommando:
Alright should be good to go now!

MistakeNot4892 (approved):
Needs squashmerge.

PR #2097, opened 2021-09-16
Moving tail icon stuff onto the organ. [IDB IGNORE]

MistakeNot4892:
Mostly just cleans up the tailcode changes from earlier.


MistakeNot4892:
There's a visual bug with tails persisting when changing species, I think it must be related to the bug with body blending mode not updating across bodytype changes. Don't think it relates to this PR so would like this merged without it being fixed. The visual issues don't persist into the actual round at least.

PsiOmegaDelta (approved):

PR #2129, opened 2021-09-30
Standardize health

SierraKomodo:
This time without extensions.

No user-facing changes unless I overlooked something.

### DIFF
--- a/code/game/atoms.dm
+++ b/code/game/atoms.dm
@@ -212,6 +212,8 @@ its easier to just keep the beam vertical.
 
 	to_chat(user, "[html_icon(src)] That's [f_name] [suffix]")
 	to_chat(user, desc)
+	if (health_max)
+		examine_damage_state(user)
 	return TRUE
 
 // called by mobs when e.g. having the atom as their machine, loc (AKA mob being inside the atom) or buckled var set.
--- a/code/game/atoms_health.dm
+++ b/code/game/atoms_health.dm
@@ -1,0 +1,256 @@
+/// Current health for health processing. Use `get_current_health()`, `damage_health()`, or `restore_health()` for general health references.
+/atom/var/health_current
+
+/// Maximum health for simple health processing. Use `get_max_health()` or `set_max_health()` to reference/modify.
+/atom/var/health_max
+
+/**
+ * LAZY List of damage type resistance or weakness multipliers, decimal form. Only applied to health reduction. Use `set_damage_resistance()`, `remove_damage_resistance()`, and `get_damage_resistance()` to reference/modify.
+ *
+ * Index should be one of the `DAMAGE_` flags.
+ * Value should be a multiplier that is applied against damage. Values below 1 are a resistance, above 1 are a weakness.
+ * Value of `0` is considered immunity.
+ */
+/atom/var/list/health_resistances
+
+/// Minimum damage required to actually affect health in `can_damage_health()`.
+/atom/var/health_min_damage = 0
+
+/**
+ * Retrieves the atom's current health, or `null` if not using health
+ */
+/atom/proc/get_current_health()
+	SHOULD_CALL_PARENT(TRUE)
+	if (!health_max)
+		return
+	return health_current
+
+/**
+ * Retrieves the atom's maximum health.
+ */
+/atom/proc/get_max_health()
+	SHOULD_CALL_PARENT(TRUE)
+	return health_max
+
+/**
+ * Retrieves the atom's current damage, or `null` if not using health.
+ * If `use_raw_values` is `TRUE`, uses the raw var values instead of the `get_*` proc results.
+ */
+/atom/proc/get_damage_value(use_raw_values)
+	if (!health_max)
+		return
+	if (use_raw_values)
+		return health_max - health_current
+	else
+		return get_max_health() - get_current_health()
+
+/**
+ * Retrieves the atom's current damage as a percentage where `100%` is `1.00`.
+ * If `use_raw_values` is `TRUE`, uses the raw var values instead of the `get_*` proc results.
+ */
+/atom/proc/get_damage_percentage(use_raw_values)
+	if (!health_max)
+		return
+	var/max_health = use_raw_values ? health_max : get_max_health()
+	return round(get_damage_value(use_raw_values) / max_health, 0.01)
+
+/**
+ * Checks if the atom's health can be restored.
+ * Called by `restore_health()`. Use `mod_health()` directly to skip this check.
+ * Returns `null` if health is not in use.
+ * NOTE: Does not include an `is_alive()` check to allow repairing/healing atoms back to life.
+ */
+/atom/proc/can_restore_health(damage, damage_type = null)
+	SHOULD_CALL_PARENT(TRUE)
+	if (!health_max)
+		return
+	if (!damage)
+		return FALSE
+	if (get_current_health() == get_max_health())
+		return FALSE
+	return TRUE
+
+/**
+ * Checks if the atom's health can be damaged.
+ * Called by `damage_health()`. Use `mod_health()` directly to skip this check.
+ * Returns `null` if health is not in use.
+ */
+/atom/proc/can_damage_health(damage, damage_type = null)
+	SHOULD_CALL_PARENT(TRUE)
+	if (!health_max)
+		return
+	if (!is_alive())
+		return FALSE
+	if (!damage || damage < health_min_damage)
+		return FALSE
+	if (get_damage_resistance(damage_type) == 0)
+		return FALSE
+	return TRUE
+
+/**
+ * Checks if the atom is 'alive' or 'dead'.
+ * Returns `null` if health is not in use.
+ */
+/atom/proc/is_alive()
+	SHOULD_CALL_PARENT(TRUE)
+	if (!health_max)
+		return
+	return get_current_health() > 0
+
+/**
+ * Health modification for the health system. Applies `health_mod` directly to `simple_health` via addition and calls `handle_death_change` as needed.
+ * Has no pre-modification checks, you should be using `damage_health()` or `restore_health()` instead of this unless you explicitly need to bypass those checks.
+ * `skip_death_state_change` will skip calling `handle_death_change()` when applicable. Used for when the originally calling proc needs handle it in a unique way.
+ * Returns `TRUE` if the death state changes, `null` if the atom is not using health, `FALSE` otherwise.
+ */
+/atom/proc/mod_health(health_mod, damage_type = null, skip_death_state_change = FALSE)
+	SHOULD_CALL_PARENT(TRUE)
+	if (!health_max)
+		return
+	var/death_state = is_alive()
+	health_current = Clamp(health_current + health_mod, 0, get_max_health())
+	post_health_change(health_mod, damage_type)
+	var/new_death_state = is_alive()
+	if (death_state == new_death_state)
+		return FALSE
+	if (!skip_death_state_change)
+		handle_death_change(new_death_state)
+	return TRUE
+
+/**
+ * Sets `health_current` to the new value, clamped between `0` and `health_max`.
+ * Has no pre-modification checks.
+ * Returns `TRUE` if the death state changes, `null` if the atom is not using health, `FALSE` otherwise.
+ */
+/atom/proc/set_health(new_health)
+	SHOULD_CALL_PARENT(TRUE)
+	var/health_mod = new_health - health_current
+	return mod_health(health_mod)
+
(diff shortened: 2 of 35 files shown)


code/game/atoms_health.dm
```diff
@@ -0,0 +1,310 @@
+/atom
+	/// Current health for health processing. Use `get_current_health()`, `damage_health()`, or `restore_health()` for general health references.
+	var/health_current
+	/// Maximum health for simple health processing. Use `get_max_health()` or `set_max_health()` to reference/modify.
+	var/health_max
+	/// Whether or not the atom is considered 'dead'. Use `is_alive()` for general checks.
```
MistakeNot4892: Isn't this a bit redundant in view of `stat`?

SierraKomodo: `stat` is only a mob thing isn't it?

MistakeNot4892: Most atoms that need it have a way to track dead status though. Machines have the `BROKEN` flag, mobs have `stat == DEAD`, structures have `physically_destroyed()` etc. This seems like an additional one that isn't really going to be used for those cases.

SierraKomodo: Removed the var and replaced it with `is_alive()` checks before and after modifying health to maintain the `handle_death_change()` proc functionality. Atom overrides for `is_alive()` will need to include checks for stat, BROKEN flags, physically_destroyed(), etc as applicable.

code/game/atoms_health.dm
```diff
@@ -0,0 +1,293 @@
...
+	SHOULD_CALL_PARENT(TRUE)
+	health_max = new_max_health
+	health_current = set_current_health ? get_max_health() : min(health_current, get_max_health())
+
+/**
+ * Sets the atom's resistance/weakness to the given damage type.
+ * Value should be a multiplier that is applied against damage. Values below 1 are a resistance, above 1 are a weakness.
+ * Value of `0` is considered immunity.
+ */
+/atom/proc/set_damage_resistance(damage_type, resistance_value)
```
MistakeNot4892, line 194: As far as I can tell this proc and the associated `health_resistances` var is only ever used by blobs. Putting it at atom level means things like areas are going to inherit it. It's not much overhead as it is LAZY but I'm not sure how ideal it is for it to be at the base object type level when only one thing in the game uses it.

code/game/atoms_health.dm
```diff
@@ -0,0 +1,293 @@
...
+	if (get_current_health() == get_max_health())
+		return FALSE
+	return TRUE
+
+/**
+ * Checks if the atom's health can be damaged.
+ * Should be called before `damage_health()` in most cases.
+ * Returns `null` if health is not in use.
+ */
+/atom/proc/can_damage_health(damage, damage_type = null)
```
MistakeNot4892, line 78: Unless the diff is incomplete this proc is not called anywhere.

code/game/atoms_health.dm
```diff
@@ -0,0 +1,293 @@
...
+	if (new_death_state != is_alive())
+		return FALSE
+	health_current = new_death_state ? 0 : get_max_health()
+	handle_death_change(new_death_state)
+	return TRUE
+
+/**
+ * Ease of use proc for `set_death_state()`. This one kills the atom.
+ */
+/atom/proc/kill_health()
```
MistakeNot4892, line 162: Not called anywhere.

code/game/atoms_health.dm
```diff
@@ -0,0 +1,293 @@
...
+	SHOULD_CALL_PARENT(TRUE)
+	return set_death_state(TRUE)
+
+/**
+ * Ease of use proc for `set_death_state()`. This one revives the atom.
+ */
+/atom/proc/revive_health()
```
MistakeNot4892, line 169: Not called anywhere.

code/game/atoms_health.dm
```diff
@@ -0,0 +1,293 @@
...
+	var/max_health = use_raw_values ? health_max : get_max_health()
+	return round(get_damage_value(use_raw_values) / max_health, 0.01)
+
+/**
+ * Checks if the atom's health can be restored.
+ * Should be called before `restore_health()` in most cases.
+ * Returns `null` if health is not in use.
+ * NOTE: Does not include a check for death state by default, to allow repairing/healing atoms back to life.
+ */
+/atom/proc/can_restore_health(damage, damage_type = null)
```
MistakeNot4892, line 63: Not called anywhere. Is there a file missing from the diff? Bit confused.

code/game/atoms_health.dm
```diff
@@ -0,0 +1,293 @@
...
+
+	/// Minimum damage required to actually affect health in `can_damage_health()`.
+	var/health_min_damage = 0
+
+/**
+ * Whether or not the atom has health. Checks for a value in `health_max`.
+ * Can be overridden for additional checks.
+ * Returns `TRUE` or `FALSE`.
+ */
+/atom/proc/has_health()
```
MistakeNot4892: What additional checks would you expect this to end up needing?

code/game/atoms_health.dm
```diff
@@ -0,0 +1,293 @@
...
+	return health_current
+
+/**
+ * Retrieves the atom's maximum health, or `null` if not using health.
+ */
+/atom/proc/get_max_health()
+	SHOULD_CALL_PARENT(TRUE)
+	if (!has_health())
+		return
+	return health_max
```
MistakeNot4892, line 33: Considering that `health_max` is falsey if the mob has no health this proc chain seems a bit redundant - couldn't you just return max health and let the caller handle it being zero?

code/game/atoms_health.dm
```diff
@@ -0,0 +1,293 @@
...
+		return FALSE
+	handle_death_change(new_death_state)
+	return TRUE
+
+/**
+ * Sets `health_current` to the new value, clamped between `0` and `health_max`.
+ * Has no pre-modification checks.
+ * Returns `TRUE` if the death state changes, `null` if the atom is not using health, `FALSE` otherwise.
+ */
+/atom/proc/set_health(new_health)
```
MistakeNot4892, line 125: Not called anywhere.

code/game/atoms_health.dm
```diff
@@ -0,0 +1,293 @@
...
+ * Has no pre-modification checks, you should be using `damage_health()` or `restore_health()` instead of this.
+ * Returns `TRUE` if the death state changes, `null` if the atom is not using health, `FALSE` otherwise.
+ */
+/atom/proc/mod_health(health_mod, damage_type)
+	SHOULD_CALL_PARENT(TRUE)
+	if (!has_health())
+		return
+	var/death_state = is_alive()
+	health_current = Clamp(health_current + health_mod, 0, get_max_health())
+	post_health_change(health_mod, damage_type)
```
MistakeNot4892, line 112: I think we discussed this previously - can't blobs just override mod_health()? They are the only thing that actually uses this proc.

code/game/atoms_health.dm
SierraKomodo: Plans are for everything that has any form of health (With the exception of human mobs themselves because of complicated organ code) to use the new health stuff, and a number of things across the code base do have some form of armor/resistance based values in their current implementations that'll have to be transferred over. Walls specifically come to mind as an example, as I've already tied those in on Bay code. I believe some subtypes of structures and simple mobs have resistance values as well.

code/game/atoms_health.dm
SierraKomodo: A case of something that was in use and then ended up not being used by the updated blob or rune wall code, but is something that will be used elsewhere when I incorporate this into other parts of the code.

code/game/atoms_health.dm
SierraKomodo: Same as above. At this point I should probably port over the implementations to walls and items now rather than waiting for the base stuff to get reviewed.

code/game/atoms_health.dm
SierraKomodo: As above.

code/game/atoms_health.dm
SierraKomodo: As above

code/game/atoms_health.dm
SierraKomodo: That's a fair point - Will update

code/game/atoms_health.dm
SierraKomodo: In hindsight, nothing. This one's a holdover from when there was the extension.

code/game/atoms_health.dm
SierraKomodo: This is to allow any effects from the health changes to occur before the rest of the mod_health proc runs with the death state changing - I was having issues with atoms, such as blobs, that qdel() themselves on death when I was just straight overriding mod_health itself for things like icon state checks and changes.

code/game/atoms_health.dm
SierraKomodo: As above with the other currently unused procs

code/game/atoms.dm
```diff
@@ -212,6 +212,8 @@ its easier to just keep the beam vertical.
 
 	to_chat(user, "[html_icon(src)] That's [f_name] [suffix]")
 	to_chat(user, desc)
+	if (health_max)
+		examine_damage_state(user)
```
MistakeNot4892, line 216: Structures will need to be converted to use this var and proc down the track.

SierraKomodo: Yes, that's the plan - To eventually have everything converted. Just trying to avoid doing the entire codebase at once because that's a _massive_ overhaul and the last time I tried a 'Update the entire code base at once' I was perpetually locked behind never-ending merge conflicts.

code/game/atoms_health.dm
MistakeNot4892: Is there an expected override for this that makes it require an override rather than just calling and overriding `set_death_state(TRUE)`?

code/game/gamemodes/cult/runes.dm
```diff
@@ -246,15 +246,15 @@
 /obj/effect/rune/wall/cast(var/mob/living/user)
 	var/t
 	if(wall)
-		if(wall.health >= wall.max_health)
+		t = wall.get_damage_value()
+		if(!t)
```
MistakeNot4892, line 250: Won't this prevent repair if the wall is at or below zero? Seems like the proc returns a null in instances where health is irrelevant.

code/game/atoms_health.dm
```diff
@@ -0,0 +1,279 @@
...
+		if (0.01 to 0.32)
+			to_chat(user, SPAN_WARNING("It looks slightly damaged."))
+		if (0.33 to 0.65)
+			to_chat(user, SPAN_WARNING("It looks moderately damaged."))
+		else
+			to_chat(user, SPAN_DANGER("It looks severely damaged."))
+
+/mob/examine_damage_state(mob/user)
+	if (!is_alive())
+		to_chat(user, SPAN_DANGER("They look severely hurt and is not moving or responding to anything around them."))
```
MistakeNot4892: Probably should use pronouns for this message and the other strings above and below.

code/modules/clothing/shoes/_shoes.dm
```diff
@@ -142,15 +142,13 @@
...
-		if (attached_cuffs.health)
-			attached_cuffs.health -= 1
-			if (attached_cuffs.health < 1)
-				visible_message(SPAN_WARNING("\The [attached_cuffs] attached to \the [src] snap and fall away!"), range = 1)
-				verbs -= /obj/item/clothing/shoes/proc/remove_cuffs
-				LAZYINITLIST(slowdown_per_slot[slot_shoes_str])
-				slowdown_per_slot[slot_shoes_str] -= attached_cuffs.elastic ? 10 : 15
-				QDEL_NULL(attached_cuffs)
+	if (attached_cuffs?.health_max && running)
+		if (attached_cuffs.damage_health(1, skip_death_state_change = TRUE))
```
MistakeNot4892: `if` checks could be merged and training return at the end could be removed.

code/game/atoms_health.dm
SierraKomodo: With the code fresh in my mind once more:

`post_health_change()` exists specifically for processing to occur _before_ `handle_death_state()` is called so that whatever interactions need to happen due to being hit or what have you can occur before things process dying.

code/modules/clothing/shoes/_shoes.dm
SierraKomodo: Also decided to remove the initial `attached_cuffs?.health_max` check since that's also checked in the damage health proc chain

PR #2146, opened 2021-10-13
Refactoring grabs to allow some more information to pass through the call chain.

MistakeNot4892:
Split out of #2031. Basically just passes proximity and target atom through the grab chain, to be used in the ridable mob system to direct the mob towards a target.


MistakeNot4892:
- [x] Grab attack doafter is applied before click cooldown, so you can spam them.
- [x] Upgrading/using grabs seems inconsistent.

MistakeNot4892:
This appears to be working.

afterthought2 (approved):

PR #2154, opened 2021-10-15
Killing as many 'in world' loops as feasible.

MistakeNot4892:
Also adds a unit test to prevent people adding more.

Not completely convinced of my own genius in regards to a number of these changes, particularly the debugging marker ones that would not reasonably be seen during general play and probably don't need to sit in memory.

### DIFF
--- a/code/game/objects/structures/flora.dm
+++ b/code/game/objects/structures/flora.dm
@@ -14,7 +14,7 @@
 /obj/structure/flora/tree/pine/Initialize()
 	. = ..()
 	icon_state = "pine_[rand(1, 3)]"
-
+	
 /obj/structure/flora/tree/pine/xmas
 	name = "\improper Christmas tree"
 	desc = "O Christmas tree, O Christmas tree..."
--- a/code/modules/games/cards.dm
+++ b/code/modules/games/cards.dm
@@ -17,10 +17,19 @@
 	else
 		return image(use_custom_front || deck_icon, "[card_icon]")
 
+var/global/list/card_decks = list()
 /obj/item/deck
 	w_class = ITEM_SIZE_SMALL
 	icon = 'icons/obj/items/playing_cards.dmi'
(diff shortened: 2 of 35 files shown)


code/game/objects/structures/flora.dm
```diff
@@ -6,15 +6,21 @@
 	pixel_x = -16
 	layer = ABOVE_HUMAN_LAYER
 
+var/global/list/christmas_trees = list()
```
MistakeNot4892: Unsure if this is worth doing for the sake of one meme holiday proc.

code/modules/ZAS/Fire.dm
```diff
@@ -86,6 +86,7 @@ If it gains pressure too slowly, it may leak or just rupture instead of explodin
 	zone.fire_tiles |= src
 	return 0
 
+var/global/list/hotspots = list()
```
MistakeNot4892: This is primarily only checked during a debugging proc so unsure if it's needed.

code/modules/games/cards.dm
```diff
@@ -17,11 +17,20 @@
 	else
 		return image(use_custom_front || deck_icon, "[card_icon]")
 
+var/global/list/card_decks = list()
```
MistakeNot4892, line 20: This is another possibly unnecessary one, the only use is the missing card item.

code/modules/ZAS/Fire.dm
comma: Yeah I tihnk debugging guy can take lag, better than adding list operation to init of every fire object out there IMO

code/modules/games/cards.dm
MistakeNot4892: Going to keep this one, there aren't many card decks either so it's not a big imposition on memory.

comma (approved):

PR #2156, opened 2021-10-16
Removed some more 'in world's, <<s and >>s.

MistakeNot4892:
- Swaps out some `in world`s for existing global lists.
- Adds a global list for engine setup markers and pAIs.
- Removes an unnecessary seed init thing from SSplants (init order takes care of this now).
- Adds macros: `BITSHIFT_LEFT`, `BITSHIFT_RIGHT`, `direct_output`, `from_savefile`, `to_output` and `to_savefile`.
- Swaps tons of `>>` and `<<` for macros.
- Deletes some commented/unused code.


code/unit_tests/power_tests.dm
```diff
@@ -21,8 +21,7 @@
 	var/failed = 0
 	var/list/found_cables = list()
 
-	//there is a cable list, but for testing purposes we search every cable in the world
-	for(var/obj/structure/cable/C in world)
+	for(var/obj/structure/cable/C in global.cable_list)
```
MistakeNot4892, line 24: This was discussed previously on Discord - iterating `in world` means it will pick up qdel'd cables and report them as not connecting, when they are deleted by away site or submap loading, runtime explosions, etc.

comma (approved):

code/__defines/xenoarcheaology.dm
```diff
@@ -1,6 +1,6 @@
-#define XENOFIND_APPLY_PREFIX 	(2 << 0)
-#define XENOFIND_APPLY_DECOR 	(2 << 1)
-#define XENOFIND_REPLACE_ICON 	(2 << 2)
```
PsiOmegaDelta, line 3: Don't know why these used to be a 2 that's shifted, which would be different from `BITFLAG(x)`, but from the look of things it should have no practical side-effects

code/modules/client/asset_cache.dm
```diff
@@ -53,11 +53,7 @@ You can set verify to TRUE if you want send() to sleep until the client has the
 	client.sending |= asset_name
 	var/job = ++client.last_asset_job
 
-	client << browse({"
-	<script>
-		window.location.href="?asset_cache_confirm_arrival=[job]"
-	</script>
-	"}, "window=asset_cache_browser")
+	direct_output(client, browse("<script>window.location.href='?asset_cache_confirm_arrival=[job]'</script>", "window=asset_cache_browser"))
```
PsiOmegaDelta, line 56: Was pondering if this wasn't a `show_browser()` situation, but at the same time the browser isn't actually shown

code/__defines/xenoarcheaology.dm
MistakeNot4892: Yeah I went to have a look and it's just bitflags.

code/modules/client/asset_cache.dm
MistakeNot4892: Initially I tried to retrofit this into show_browser but ended up going for the safer option.

code/__defines/xenoarcheaology.dm
comma: yeah 2 is just bit shifted 1 anyway

PR #2159, opened 2021-10-16
Brute and burn meds now use max chem effect instead of stacking.

MistakeNot4892:
## Description of changes
Previously you could make things like keloderm mixes that stacked various tiers of drug with the same effect to get cumulative effects. This PR changes brute and burn meds to use the maximum chem value instead of stacking.

This removes one of the more irritating tendencies of chem/med players to push for optimized play. It is also preparation for adding more tiered chems that would be exploitable with the previous behavior.

## Authorship
Myself.

## Changelog

:cl:
tweak: Brute and burn meds no longer stack with regenerative serum.
/:cl:


comma (approved):

SolatK:
Nooo, my keloderm and bicatric

PR #2163, opened 2021-10-16
Fixes a few network camera bugs

NataKilar:
## Description of changes

Fixes a few bugs with network cameras. Adds a hook for post-extension construction since it's needed to properly check valid sources for the camera net.

Bugs bad.

## Authorship

Myself
## Changelog

:cl:
bugfix: Robots now properly work as cameras for the purpose of AI vision
bugfix: TV cameras can now have their network settings adjusted as intended.
/:cl:

### DIFF
--- a/code/datums/extensions/extensions.dm
+++ b/code/datums/extensions/extensions.dm
@@ -9,6 +9,8 @@
 		CRASH("Invalid holder type. Expected [expected_type], was [holder.type]")
 	src.holder = holder
 
+/datum/extension/proc/post_construction()
+
 /datum/extension/Destroy()
 	holder = null
 	. = ..()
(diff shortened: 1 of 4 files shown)


code/datums/extensions/extensions.dm
```diff
@@ -9,6 +9,8 @@
 		CRASH("Invalid holder type. Expected [expected_type], was [holder.type]")
 	src.holder = holder
 
+/datum/extension/proc/post_construction()
```
PsiOmegaDelta, line 12: What does this proc do that overriding `/New()` doesn't?

NataKilar: In New() the extension knows its holder, but the extension is not yet in the holder's extensions list. In this case it's necessary because the camera extension needs to pass its holder to the cameranet to add it as a source, but the cameranet needs to check if the holder has the camera extension. The alternative is to pass the extension itself as a source to the cameranet and have some special handling, but this felt a little wonky to me.

PsiOmegaDelta: Suppose that isn't unreasonable.

MistakeNot4892 (approved):

PR #2170, opened 2021-10-18
Misc tweaks/fixes for buckling and grabbing refactors.

MistakeNot4892:
Splitting them out of the riding PR because they apply more general.

### DIFF
--- a/code/modules/mob/grab/grab_object.dm
+++ b/code/modules/mob/grab/grab_object.dm
@@ -98,14 +98,14 @@
 /obj/item/grab/afterattack(atom/target, mob/user, proximity_flag, click_parameters)
 	if(QDELETED(src) || !current_grab || !assailant || proximity_flag) // Close-range is handled in resolve_attackby().
 		return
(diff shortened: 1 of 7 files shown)


comma (approved):

code/modules/mob/grab/grab_object.dm
```diff
@@ -98,14 +98,14 @@
...
-	if(current_grab.hit_with_grab(src, target, TRUE))
+	if(current_grab.hit_with_grab(src, target, proximity_flag))
 		return
 	. = ..()
 
 /obj/item/grab/resolve_attackby(atom/A, mob/user, var/click_params)
 	if(QDELETED(src) || !current_grab || !assailant)
 		return TRUE
-	if(A.grab_attack(src) || current_grab.hit_with_grab(src, A))
+	if(A.grab_attack(src) || current_grab.hit_with_grab(src, A, get_dist(user, A) <= 1))
```
PsiOmegaDelta, line 108: Took a moment to track down but there is an `Adjacent()`-check before this call so `get_dist()` should therefor be fine

PR #2171, opened 2021-10-19
Manual cherrypicking of several Hearth fix PRs.

MistakeNot4892:
## Description of changes

Manually stolen tweaks and fixes from Hearth PRs:
https://github.com/HearthOfHestia/Nebula/pull/8
https://github.com/HearthOfHestia/Nebula/pull/24
https://github.com/HearthOfHestia/Nebula/pull/81
https://github.com/HearthOfHestia/Nebula/pull/93
https://github.com/HearthOfHestia/Nebula/pull/95
https://github.com/HearthOfHestia/Nebula/pull/99
https://github.com/HearthOfHestia/Nebula/pull/114

Also pulls in commits from #2172, ty @MarinaGryphon.

Fixes various runtimes and oversights.

## Authorship
Largely Mazian and Jade, see original PRs for authorship.

## Changelog
Nothing player facing.


MistakeNot4892:
Also tweaks some redundant returns, changes a null coalescence check, and corrects @Lohikar's spelling of mimicking. That's not from Hearth, that's just me meddling.

Lohikar:
Please don't rename the defines, syncing nebula's version of z-mimic with O7's is already fairly manual.

MistakeNot4892:
It's a one-off spelling fix that can be done with find and replace, the change footprint and effort involved is not enough to leave it in. It bugs me every time I see it.

code/modules/modular_computers/laptop_vendor.dm
```diff
@@ -59,9 +59,10 @@
 /obj/machinery/lapvend/proc/fabricate_and_recalc_price(var/fabricate = 0)
```
PsiOmegaDelta, line 59: This proc and it's reoccurring `if(fabricated)`-checks is a bit of a code smell..

code/modules/clothing/clothing_accessories.dm
```diff
@@ -82,7 +82,7 @@
 	update_clothing_icon()
 
 /obj/item/clothing/proc/remove_accessory(mob/user, obj/item/clothing/accessory/A)
-	if(!(A in accessories))
```
PsiOmegaDelta, line 85: This to me seems to imply that `accessories()` have ended up with a null-entry at some point, typically because a deleted accessory hasn't removed itself properly and gets cleaned up by a hard-del

That is: This is a band-aid fix of a symptom, and doesn't fix the root cause

code/controllers/subsystems/atoms.dm
```diff
@@ -23,12 +23,6 @@ SUBSYSTEM_DEF(atoms)
 /datum/controller/subsystem/atoms/Initialize(timeofday)
 	atom_init_stage = INITIALIZATION_INNEW_MAPLOAD
 	InitializeAtoms()
-
-	// Mannequins refuse to spawn prior to atoms init, so refresh them in case anyone connected before init finished.
```
PsiOmegaDelta, line 27: Are we sure this runtime fix won't result in mannequins not being properly refreshed?

MistakeNot4892: This code was added when character setup initialized before SSatoms, these days it initializes after. There should be no way for it to be an issue.

code/modules/modular_computers/laptop_vendor.dm
MistakeNot4892: The whole vendor is horrible. Also six years old. It needs to be torn out and rewritten imo.

code/modules/clothing/clothing_accessories.dm
MistakeNot4892: It also implies that something is passing a null to `remove_accessory` which is pretty weird,

PsiOmegaDelta: At least nulls are not being added via `attach_accessory()` because that'd result in a runtime on line 77

code/modules/client/preference_setup/loadout/loadout.dm
```diff
@@ -349,6 +351,9 @@ var/global/list/gear_datums = list()
 /decl/loadout_option/Initialize()
 	. = ..()
 
+	if(is_abstract())
+		return .
```
PsiOmegaDelta, line 355: Seems a bit silly to return `.` since it's always the implicit return value

PsiOmegaDelta (approved):

code/modules/client/preference_setup/loadout/loadout.dm
MistakeNot4892: Kinda agree, but it's also more explicit than a naked return and doesn't really hurt anything.

PR #2176, opened 2021-10-20
Adding customizable z-level data markers. [MDB IGNORE]

MistakeNot4892:
## Description of changes
Adds a landmark-y structure used to automatically set z-level information.

Don't need to define it with specific z-levels in /datum/map, allows for laterally connected z-levels.

## Authorship
Myself, idea from Polaris.

## Changelog
Nothing player facing.


code/modules/multiz/level_data.dm
```diff
@@ -0,0 +1,87 @@
+#define ZLEVEL_STATION BITFLAG(0)
```
PsiOmegaDelta, line 1: Over time we probably want to replace/remove the station-level concept. It never felt like something that was multi-map compatible.
Typically events should probably affect a given Z-block (as occasionally connected Z-blocks)

PsiOmegaDelta (approved):

code/modules/multiz/level_data.dm
MistakeNot4892: A lot of the events already do, thankfully. I agree anyway.

MarinaGryphon: Do keep in mind that it's currently used to determine what Z-levels/areas to run certain unit tests on. If they ran everywhere, it would improperly mark areas that are intended to have a run-down or destroyed look as failing unit tests. Maybe "testing_levels" would be a good compromise?

PR #2178, opened 2021-10-24
set_species() now uses default pronoun and bodytype instead of random.

MistakeNot4892:
Fixes inconsistent preview icons.

### DIFF
--- a/code/modules/species/species.dm
+++ b/code/modules/species/species.dm
@@ -203,6 +203,7 @@ var/global/const/DEFAULT_SPECIES_HEALTH = 200
 
 	var/list/override_limb_types // Used for species that only need to change one or two entries in has_limbs.
 
+	var/decl/pronouns/default_pronouns
 	var/list/available_pronouns = list(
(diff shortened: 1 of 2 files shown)


code/modules/species/species.dm
```diff
@@ -346,6 +347,11 @@ var/global/const/DEFAULT_SPECIES_HEALTH = 200
 		available_pronouns -= pronoun
 		available_pronouns += GET_DECL(pronoun)
 
+	if(ispath(default_pronouns))
+		default_pronouns = GET_DECL(default_pronouns)
+	else if(length(available_pronouns) && !default_pronouns)
+		default_pronouns = available_pronouns
```
PsiOmegaDelta: `default_pronouns` is a decl while `available_pronouns` is a list, so I'm not quite sure how this assignment works out

MistakeNot4892: Oop should be [1]

PsiOmegaDelta (approved):

PR #2181, opened 2021-10-26
Fix runtime on airlock init

PsyCommando:
## Description of changes

Airlocks that autoset their access and names would runtime on init if placed outside an area. Added a warning on init without a proper area. (Edge case)

Also made sure external airlocks wouldn't runtime if they tried to get their access req from their areas if there weren't any. Was mostly an issue when building a new airlock from scratch.

## Changelog

:cl:
bugfix: Fixed runtime on autosetting airlocks placed outside an area.
bugfix: Fixed runtime on building an external airlock outside an area.
/:cl:

### DIFF
--- a/code/game/machinery/doors/airlock_autoname.dm
+++ b/code/game/machinery/doors/airlock_autoname.dm
@@ -3,7 +3,10 @@
 /obj/machinery/door/airlock/hatch/autoname/Initialize()
 	. = ..()
 	var/area/A = get_area(src)
-	SetName("hatch ([A.name])")
+	if(A)
+		SetName("hatch ([A.name])")
+	else
+		log_warning("\the [src] at ([x], [y], [z]) couldn't find an area to take its name from!")
 
 /obj/machinery/door/airlock/hatch/autoname/general
 	stripe_color = COLOR_CIVIE_GREEN
--- a/code/game/machinery/doors/airlock_autoname.dm
+++ b/code/game/machinery/doors/airlock_autoname.dm
@@ -5,8 +5,6 @@
 	var/area/A = get_area(src)
 	if(A)
 		SetName("hatch ([A.name])")
-	else
-		log_warning("\the [src] at ([x], [y], [z]) couldn't find an area to take its name from!")
 
 /obj/machinery/door/airlock/hatch/autoname/general
 	stripe_color = COLOR_CIVIE_GREEN
--- a/code/game/machinery/doors/airlock_subtypes.dm
+++ b/code/game/machinery/doors/airlock_subtypes.dm
@@ -147,8 +147,10 @@
 
 /obj/machinery/door/airlock/external/get_auto_access()
 	. = ..()
-	if(is_station_area(get_area(src)))
+	var/area/A = get_area(src)
+	if(A && is_station_area(A))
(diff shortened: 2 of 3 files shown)


MistakeNot4892:
How is an airlock initializing in a spot with no area in the first place?
EDIT: atom info repo might spawn it, this is valid.

code/game/machinery/doors/airlock_autoname.dm
```diff
@@ -3,7 +3,10 @@
 /obj/machinery/door/airlock/hatch/autoname/Initialize()
 	. = ..()
 	var/area/A = get_area(src)
-	SetName("hatch ([A.name])")
+	if(A)
+		SetName("hatch ([A.name])")
+	else
+		log_warning("\the [src] at ([x], [y], [z]) couldn't find an area to take its name from!")
```
MistakeNot4892: If this is going to throw a warning why bother checking it in the first place? Warnings will fail CI as much as a runtime will.
Coords will also be null because the atom is in nullspace, making it doubly pointless.

code/game/machinery/doors/airlock_subtypes.dm
```diff
@@ -147,8 +147,10 @@
 
 /obj/machinery/door/airlock/external/get_auto_access()
 	. = ..()
-	if(is_station_area(get_area(src)))
+	var/area/A = get_area(src)
+	if(A && is_station_area(A))
```
MistakeNot4892, line 151: Should probably just make `is_station_area()` check `istype(A, /area)`.

code/game/machinery/doors/airlock_autoname.dm
PsyCommando: Well, not with a map loaded from save. But yeah, the coordinates are kinda useless. It just came with my standard copy-pasta debug line ^^;
I'll remove the warning then. I assumed log_error() was the one treated as a runtime since it actually runtimes.

code/game/machinery/doors/airlock_subtypes.dm
PsyCommando: done!

code/game/machinery/doors/airlock_autoname.dm
PsyCommando: fixed!

code/game/machinery/doors/airlock_subtypes.dm
```diff
@@ -148,7 +148,7 @@
 /obj/machinery/door/airlock/external/get_auto_access()
 	. = ..()
 	var/area/A = get_area(src)
-	if(A && is_station_area(A))
+	if(istype(A))
```
MistakeNot4892: This isn't what I meant - I meant making `is_station_area(A)`, the proc, do `if(istype(A))` before returning the proc it currently class.

PsyCommando: Oh, so you want me to change is_station_area itself to do a type check?

MistakeNot4892: Geh I swear I commented on this already - yes, that's the ideal fix imo.

MistakeNot4892 (approved):
Needs squash merge.

PR #2182, opened 2021-10-26
Cryopod init fix

PsyCommando:
## Description of changes
Cryopods used without control computers would runtime on init, even though they shouldn't. Since the control computer was implemented to be mostly optional everywhere else in the code. 

## Changelog

:cl:
bugfix: Fix cryopods without control computers from runtiming during init.
/:cl:

### DIFF
--- a/code/game/machinery/cryopod.dm
+++ b/code/game/machinery/cryopod.dm
@@ -268,7 +268,8 @@
 /obj/machinery/cryopod/proc/find_control_computer()
 	if(!control_computer)
 		control_computer = locate(/obj/machinery/computer/cryopod) in src.loc.loc
-		events_repository.register(/decl/observ/destroyed, control_computer, src, .proc/clear_control_computer)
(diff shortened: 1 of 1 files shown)


code/game/machinery/cryopod.dm
```diff
@@ -268,7 +268,8 @@
 /obj/machinery/cryopod/proc/find_control_computer()
 	if(!control_computer)
 		control_computer = locate(/obj/machinery/computer/cryopod) in src.loc.loc
```
MistakeNot4892: Might be worth changing `src.loc.loc` to `get_area(src)` so it won't runtime on a null loc for the cryopod.

PsyCommando: True, pretty sure there's a couple other instances like that in that old piece of code.

MistakeNot4892 (approved):
Needs squash merge

PR #2183, opened 2021-10-26
Fix runtime on update icon for injectors

PsyCommando:
## Description of changes
If an autoinjector had no reagent contents and the reagents var wasn't initialized it would result in runtime spam on update icon. 

## Changelog

:cl:
bugfix: Fix runtime on updating the icon for empty autoinjectors.
/:cl:

### DIFF
--- a/code/modules/reagents/reagent_containers/hypospray.dm
+++ b/code/modules/reagents/reagent_containers/hypospray.dm
@@ -168,7 +168,7 @@
 
 /obj/item/chems/hypospray/autoinjector/on_update_icon()
 	overlays.Cut()
-	if(reagents.total_volume > 0)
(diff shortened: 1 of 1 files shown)


MistakeNot4892 (dismissed):
Bit leery of null guards being put into code that should never be running prior to Initialize() but this is a small one at least.

MistakeNot4892:
Might want to talk with Nata on this one, he is leery of the change as well (not just me blocking it :U)

PsyCommando:
Yeah, its all good, to be honest, I was just getting rid of old PRs that were laying around. In retrospect this one is probably an unrelated issue I mistook for something that should go upstream.. Thanks for the feedback though!

PR #2184, opened 2021-10-27
Cut up and reorganized the huma init code a bit

PsyCommando:
## Description of changes

So, my first idea didn't turn out as planned. I've toned down the scope of the changes I wanted to make to something more manageable. 
Mob factories got messy and turned into a pain in the ass to use and a ton of barely useful subtypes. So after trying them out and noticing they couldn't cover all the use cases for creating human mobs, like initializing from dna or appearance I just went back to something much simpler.

* My main goal is to split init into easily overridable sub-procs so loading mobs from save is a much less messy affair, and doesn't involve re-initing the mob several times and storing copies of a lot of the mob's data and adding redundant version of the existing init code. 
* My secondary goal is to make it possible for species to override init a bit more in-depth if possible.
* My last goal is to get rid of redundant crap (there's so much of that!), and tidy up things only if possible.

This is mostly just a draft right now. I'd like to discuss how to best split the process of initializing human mobs,

## Changelog

:cl:
tweak: Changes to human mob initialization.
/:cl:


code/modules/mob/living/carbon/human/human.dm
```diff
@@ -628,10 +609,11 @@
 	dna.check_integrity(src)
 	return
 
-/mob/living/carbon/human/get_species_name()
-	if(!species)
-		set_species()
-	. = ..()
+//Not cool, getters shouldn't modify the state of the object like that
```
MistakeNot4892: This is an unnecessary safeguard anyway. Please delete the code instead of leaving it commented though.

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -984,135 +823,46 @@
...
+	var/new_species = get_species_by_key(new_species_name)
+	if(species?.name == new_species_name)
+		return
+	if(!new_species)
+		CRASH("set_species on mob '[src]' was passed a bad species name '[new_species_name]'!")
 
+	//Handle old species transition
 	if(species)
-
-		if(species.name && species.name == new_species)
```
MistakeNot4892, line 679: Why are you removing this? The idea was to skip a long and intensive process if the mob is already the target species.

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -984,135 +823,46 @@
...
-	set_bodytype(pick(species.available_bodytypes), TRUE)
-
-	species.create_organs(src)
-	species.handle_post_spawn(src)
-
+	//Update our species
+	species = new_species
+	if(dna)
+		dna.species = new_species_name
+	holder_type = species.holder_type?  species.holder_type : holder_type
```
MistakeNot4892: The ternary here means that you'll always have a holder type if you ever held a species that has one. You probably want `initial(holder_type)` or (imo) just assign species holder_type and don't check it.

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -984,135 +823,46 @@
...
-	reset_offsets()
-
-	appearance_descriptors = null
-	if(LAZYLEN(species.appearance_descriptors))
-		for(var/desctype in species.appearance_descriptors)
-			var/datum/appearance_descriptor/descriptor = species.appearance_descriptors[desctype]
-			LAZYSET(appearance_descriptors, descriptor.name, descriptor.default_value)
+	var/decl/pronouns/new_pronouns = get_pronouns_by_gender(get_sex())
+	if(!istype(new_pronouns) || !(new_pronouns in species.available_pronouns))
+		new_pronouns = pick(species.available_pronouns) //Might wanna wrap this into the species code instead of here?
```
MistakeNot4892: I recently changed this to use `species.default_pronouns` instead, pls keep that.

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -1141,29 +894,42 @@
...
+	force_update_limbs() //updates bodytype
+	default_pixel_x = initial(pixel_x) + bodytype.pixel_offset_x
+	default_pixel_y = initial(pixel_y) + bodytype.pixel_offset_y
+	default_pixel_z = initial(pixel_z) + bodytype.pixel_offset_z
+	
+	reset_offsets()
+
+	// Rebuild the HUD and visual elements.
+	if(client)
+		Login() //#FIXME: It might be a bit dangerous to call the entirety of login just for that?
```
MistakeNot4892: What is the purpose of calling Login() here?

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -1794,6 +1434,7 @@
 		to_chat(src, SPAN_DANGER("You feel a chill and your skin feels lighter..."))
 
 /mob/living/carbon/human/increaseBodyTemp(value)
+	//#TODO: Might wanna put a upper limit to this!
```
MistakeNot4892: Putting in random todos unrelated to the PR is a bit iffy.

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -1828,3 +1469,73 @@
...
+
+	set_species(species_name)
+
+	if(dna)
+		set_real_name(dna.real_name)
+	else
+		try_generate_default_name()
+
+	if(!dna)
+		src.dna.ready_dna(src) //regen dna filler only if we haven't forced the dna already
```
MistakeNot4892: This will runtime - you're checking if dna is falsey and then calling ready_dna on it.

MistakeNot4892:
Generally in favour of the concept, have some quibbles with some of the implementation. Will review properly when it's rebased/ready to go.

code/modules/mob/living/carbon/human/human.dm
PsyCommando: Well, the thing is its not a long and intensive process anymore. Most of the extra stuff is moved elsewhere.
And well, if we load from save some of the things set_species does are still required to be done, which would be skipped if that check was done. I tend to assume people know what they want when they call the proc?

code/modules/mob/living/carbon/human/human.dm
PsyCommando: ~~I never really looked into that tbh, I just copy-pasted the content of the original proc. But I can probably change that.~~
Nevermind I see it now. I think I'll just put back the if..

code/modules/mob/living/carbon/human/human.dm
PsyCommando: What did you replace? species.available_pronouns?

code/modules/mob/living/carbon/human/human.dm
PsyCommando: I have no ideas why they did it. That was in the base code. That's why I put the FIXME next to it.
I think they wanted to call the procs that refresh the HUD and decided calling the enirety of login was a good idea for some reasons.
I did tell you guys that the human init code was hot garbage XD

code/modules/mob/living/carbon/human/human.dm
PsyCommando: Yeah, it was just a personal todo that slipped in

code/modules/mob/living/carbon/human/human.dm
PsyCommando: Nope. Notice the src.dna. dna is the argument to the proc.

PsyCommando:
> 
> 
> Generally in favour of the concept, have some quibbles with some of the implementation. Will review properly when it's rebased/ready to go.

Well, I'd like if you'd expose a bit what you think about the structuring of things? Because this is really just a first draft of things.

code/modules/mob/living/carbon/human/human.dm
PsyCommando: Yeah, there's a lot of those. And I will. I tend to mark things with comments to remember why I did something before publishing my PRs ^^;

code/modules/mob/living/carbon/human/human.dm
PsyCommando: Also actually, I looked at the code and I didn't actually remove the check?

code/modules/mob/living/carbon/human/human.dm
MistakeNot4892: God I hate var shadowing.

code/modules/mob/living/carbon/human/human.dm
MistakeNot4892: `new_pronouns = species.default_pronouns` instead of the pick yeah. Did the same with bodytype.

code/modules/mob/living/carbon/human/human.dm
PsyCommando: Tbh, that goes to show I probably should name the arg something else cause its easy to make a mistake XD

code/modules/mob/living/carbon/human/human.dm
MistakeNot4892: Honestly I didn't notice it was a draft :(

MistakeNot4892:
I think we should look at a more general approach to breaking up Init, since we're in the region - I'm not very good at explaining my thought processes so I might do a quick draft PR of my own as an example of the structure I think we should ultimately go with. Your work will be easy enough to integrate into the structural changes I have in mind anyway.

PsyCommando:
> 
> 
> I think we should look at a more general approach to breaking up Init, since we're in the region - I'm not very good at explaining my thought processes so I might do a quick draft PR of my own as an example of the structure I think we should ultimately go with. Your work will be easy enough to integrate into the structural changes I have in mind anyway.

Sounds good!
I ended up reorganizing a lot more stuff after a lot of testing. Fixed a few problems too. Gonna try to apply the conflict fix and your pronoun changes soon.

code/modules/mob/living/carbon/human/human_verbs.dm
```diff
@@ -97,7 +97,7 @@
 		target.show_message("<span class='notice'>You hear a voice that seems to echo around the room: [say]</span>")
 	usr.show_message("<span class='notice'>You project your mind into [target.real_name]: [say]</span>")
 	log_say("[key_name(usr)] sent a telepathic message to [key_name(target)]: [say]")
-	for(var/mob/observer/ghost/G in world)
+	for(var/mob/observer/ghost/G in global.player_list)
```
MistakeNot4892: I could have sworn I did this already when I added `in world` to the CI checks.

PsyCommando: I did a dev merge to fix the conflicts so that's probably it? XD

PsyCommando:
Alright so, now with this last commit I'm thinking things are looking good for taking this off the draft status. I fixed all the bugs I could find when testing mob transformation, deleting and dismemberment. 

To sum it up, I changed the way human inits to be easier to override and a lot less redundant overall. Then I changed organs init and installation/removal so it involved the owner mob a lot more to make it easier to override and control the flow of things. And reduce duplicate code. 

Also I split doing the actual installation/removal of the organs from effects happening on install/removal so that we can have a single proc for handling install/removal without worrying about it triggering unwanted effects. (For example preventing child limbs from triggering owner death or various other owner deep update on their own.)

I also ended up implementing a different logic for handling organs being considered vital or not since I was tackling things related to that. Its a proc in the specie datum that end up looking at the current mob's state before returning whether an organ is vital or not, which enables things like a specie having multiple hearts and not automatically dying when one is removed for example. It doesn't completely replace the vital var in organs though since some things referring to it confuse the concept of it being a vital organ with it being an important organ that you shouldn't fuck with.

code/_onclick/hud/hud.dm
```diff
@@ -19,10 +19,10 @@
 /datum/hud
 	var/mob/mymob
 
-	var/hud_shown = 1			//Used for the HUD toggle (F12)
-	var/inventory_shown = 1		//the inventory
-	var/show_intent_icons = 0
-	var/hotkey_ui_hidden = 0	//This is to hide the buttons that can be used via hotkeys. (hotkeybuttons list of buttons)
+	var/hud_shown 			= 1			//Used for the HUD toggle (F12)
+	var/inventory_shown 	= TRUE		//the inventory
```
MistakeNot4892: Weird additional spaces?

code/game/dna/dna2.dm
```diff
@@ -358,6 +359,8 @@ var/global/list/datum/dna/gene/dna_genes[0]
 		if(length(unique_enzymes) != 32)
 			unique_enzymes = md5(character.real_name)
 	else
+		if(!species)
+			species = using_map.default_species
```
MistakeNot4892: `global`

code/game/objects/auras/regenerating_aura.dm
```diff
@@ -76,7 +76,7 @@
 			var/obj/item/organ/external/E = H.organs_by_name[limb_type]
 			if(E && E.organ_tag != BP_HEAD && !E.vital && !E.is_usable())	//Skips heads and vital bits...
 				if (H.nutrition > grow_threshold)
-					E.removed()			//...because no one wants their head to explode to make way for a new one.
+					H.remove_organ(E) 		//...because no one wants their head to explode to make way for a new one.
```
MistakeNot4892, line 79: I am opposed to this structural change as discussed on Discord. The organ should handle being removed and then call any mob-specific procs so that the behavior can be easily overridden on organ subtypes.

code/_onclick/hud/ability_screen_objects.dm
```diff
@@ -134,6 +134,14 @@
 			return S
 	return null
 
+/obj/screen/movable/ability_master/proc/synch_spells_to_mind(var/datum/mind/M)
+	if(!M)
+		return
+	LAZYINITLIST(M.learned_spells)
+	for(var/obj/screen/ability/spell/screen in spell_objects)
+		var/spell/S = screen.spell
+		M.learned_spells |= S
```
MistakeNot4892: `LAZYDISTINCTADD` would let you not bother with `LAZYINITLIST` before the loop.

code/modules/admin/view_variables/topic.dm
```diff
@@ -487,9 +487,9 @@
 	else if(href_list["remorgan"])
 		if(!check_rights(R_SPAWN))	return
 
-		var/mob/living/carbon/M = locate(href_list["remorgan"])
+		var/mob/living/carbon/human/M = locate(href_list["remorgan"])
```
MistakeNot4892: What is the purpose of making organ removal a mob proc if you're then going to restrict it to humans specifically?

mods/species/vox/organs_vox.dm
```diff
@@ -200,25 +200,28 @@
...
-		prompting = TRUE
-		var/response = alert(find_dead_player(ownerckey, 1), "Your neural backup has been placed into a new body. Do you wish to return to life as the mind of [backup.name]?", "Resleeving", "Yes", "No")
-		prompting = FALSE
-		if(src && response == "Yes" && owner == current_owner)
-			overwrite()
-	sleep(-1)
-	do_backup()
+	
+	//Need spawn here so that this interactive bit doesn't lock up init
+	spawn(0)
```
MistakeNot4892: Should be using a timer or a `waitfor = FALSE` proc.

mods/species/ascent/items/id_control.dm
```diff
@@ -51,9 +53,6 @@
 	if(ispath(id_card))
 		id_card = new id_card(src)
 	. = ..()
-	robotize()
```
MistakeNot4892, line 54: Why are you removing these lines?

MistakeNot4892: Nvm, I'm reading the changes backwards and missed the other lines.
Not completely sure that setting status like that will work but can see what you're shooting for.

mods/mobs/borers/mob/overrides.dm
```diff
@@ -31,11 +31,10 @@
 		return TRUE
 	return ..()
 
-/obj/item/organ/internal/brain/removed(var/mob/living/user)
+/obj/item/organ/internal/brain/uninstall(in_place, detach, ignore_children)
 	if(istype(owner))
 		var/mob/living/simple_animal/borer/borer = HAS_BRAIN_WORMS(owner)
-		if(borer)
-			borer.detatch()
+		borer?.detatch()
```
MistakeNot4892: Previous pattern is preferred, `if(foo) foo.bar()` is conceptually fine if foo isn't truthy, `foo?.bar()` evaluates to just `null` by itself on a line which is odd.

code/modules/admin/view_variables/topic.dm
PsyCommando: Surgery is coded that way. I can probably turn the check back to being a carbon. But I don't think organs work too well on carbon mobs currently. Plus I remember you mentioning the carbon subtype was possibly going away?

code/_onclick/hud/hud.dm
PsyCommando: That's VSCode being dumb again.. For some reasons tabs come out weird only with VSCode, but with anything else it doesn't do it. I'll take a look.

code/game/dna/dna2.dm
PsyCommando: Good catch!

mods/species/vox/organs_vox.dm
PsyCommando: Yeah, I could probably get a timer done. I was just a bit lazy there.

mods/species/ascent/items/id_control.dm
PsyCommando: Its been working fine so far. 
And I saw a bunch of species do set the prothetics flag already in their definition already anyways.

mods/mobs/borers/mob/overrides.dm
PsyCommando: I guess I can revert that.

code/_onclick/hud/hud.dm
MistakeNot4892: You shouldn't be using tabs for spacing, it's not just VSC that displays them inconsistently.

code/game/objects/auras/regenerating_aura.dm
PsyCommando: Currently the chain goes like this in order:
/mob/living/carbon/human/remove_organ()
/obj/item/organ/uninstall()
/mob/living/carbon/human/on_lost_organ()
/obj/item/organ/on_removal()

I just don't understand why you want the organs themselves to be the first in the chain? 
It feels like you want organs to just have a wall of "if" checks and type checks to determine what they're connecting to and in what state they are? All that when we got a perfectly good tool to handle avoiding coding case by case in the organ itself?
What does the old structure did that the new one doesn't do better?

code/_onclick/hud/ability_screen_objects.dm
PsyCommando: True, but then again, we don't need to check if the list is null on every iteration imo.

mods/mobs/borers/mob/overrides.dm
PsyCommando: done!

mods/species/vox/organs_vox.dm
PsyCommando: done!

code/modules/admin/view_variables/topic.dm
PsyCommando: done!

code/game/dna/dna2.dm
PsyCommando: done!

code/_onclick/hud/hud.dm
PsyCommando: done!

PsyCommando: Also, can't help it, I'm used to not having to care between spaces and tabs since I just usually have them all replaced with spaces in my ide. DM is just messing with me XD

PsyCommando:
Btw, can someone remove the "work in progress"  marker on this.

code/game/objects/auras/regenerating_aura.dm
MistakeNot4892: I linked you my alternative approach on Discord. tldr I don't think there should be a chain like that at all - I think organs, as the atoms that are primarily the point of contact, should handle all that and then pass off to the mob to update as needed. Your code in this PR does stuff like istyping() the organs as they come in as args, which I think is a less than ideal pattern.

PsyCommando:
Still waiting on #2235

PsyCommando:
Well, since loaf gave up on his PR its no longer waiting on #2235
Though depending on whoever reviews this, we may want to integrate some of the changes from that PR. But it'll take a bunch of extra time, and this PR has been due for like 2 months downstream ^^;

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -5,32 +5,23 @@
 	icon = 'icons/mob/human.dmi'
 	icon_state = "body_m_s"
 	mob_sort_value = 6
+	dna = new /datum/dna(null)
```
afterthought2: Should drop the `null` literal, dm has weird behavior that can make it misleading. I know it's oldcode.

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -659,68 +641,44 @@
...
-	species = get_species_by_key(new_species)
-	species.handle_pre_spawn(src)
-
-	skin_colour = (species.base_color && default_colour) ? species.base_color : COLOR_BLACK
-
+	//Update our species
+	species = new_species
+	if(dna)
+		dna.species = new_species_name
+	holder_type = null
```
afterthought2, line 663: This seems to be done twice.

code/modules/mob/living/carbon/human/human_organs.dm
```diff
@@ -241,3 +241,97 @@
...
+			affected = organs_by_name[O.parent_organ]
+			if(!affected)
+				CRASH("mob/living/carbon/human/proc/add_organ(): Tried to add an internal organ to a non-existing parent external organ!")
+		internal_organs |= O
+		internal_organs_by_name[O.organ_tag] = O
+
+	else if(ispath(O.type, /obj/item/organ/external))
+		organs |= O
+		organs_by_name[O.organ_tag] = O
+		refresh_modular_limb_verbs()
```
afterthought2, line 254: The reason Loaf didn't like this structure is this branching here; it seems like this should all be done in an organ proc. The stuff below the install call can presumably be moved to the `on_gained-organ` follow-up, so that doesn't leave much that would naturally belong in this main call.

code/modules/organs/internal/_internal.dm
```diff
@@ -3,96 +3,80 @@
...
-		return 0 //organs don't work very well in the body when they aren't properly attached
+		return
 
+	//#FIXME: This feels like a hack
+		//# Also it probably should be of the species of the parent limb tbh..
 	// robotic organs emulate behavior of the equivalent flesh organ of the species
-	if(BP_IS_PROSTHETIC(src) || !species)
-		species = target.species
+	// if(BP_IS_PROSTHETIC(src) || !species)
+	// 	set_species(target.species)
```
afterthought2: What's the status on this?

code/modules/organs/internal/_internal.dm
```diff
@@ -3,96 +3,80 @@
...
 /obj/item/organ/internal/remove_rejuv()
-	if(owner)
-		owner.internal_organs -= src
-		owner.internal_organs_by_name -= organ_tag
-		while(null in owner.internal_organs)
-			owner.internal_organs -= null
-		var/obj/item/organ/external/E = owner.organs_by_name[parent_organ]
-		if(istype(E))
-			LAZYREMOVE(E.internal_organs, src)
+	uninstall()
```
afterthought2: Is this kind of thing legal with the new structure? Don't you have to call the mob's removal proc to update the lists in the block I commented on above?

code/modules/organs/internal/appendix.dm
```diff
@@ -40,5 +40,5 @@
 				E.sever_artery()
 				E.germ_level = max(INFECTION_LEVEL_TWO, E.germ_level)
 				owner.adjustToxLoss(25)
-				removed()
+				uninstall()
```
afterthought2: Same here, though harder to tell with the qdel on the next line.

mods/species/vox/organs_vox.dm
```diff
@@ -200,25 +200,29 @@
...
-/obj/item/organ/internal/voxstack/replaced()
+/obj/item/organ/internal/voxstack/on_replacement()
 	if(!..()) return 0
 	if(prompting) // Don't spam the player with twenty dialogs because someone doesn't know what they're doing or panicking.
 		return 0
+	
+	//Need spawn here so that this interactive bit doesn't lock up init
 	if(owner && !backup_inviable())
-		var/current_owner = owner
+		addtimer(CALLBACK(src, .proc/prompt_revive_callback, owner), 0)
```
afterthought2: If the goal is just to unblock (i.e. returning the 1 on the next line before calling this proc is not important), invoking directly and setting `waitfor FALSE` on the callback would be preferred. This kind of timer/spawn pattern is often used to hack ordering--and that's usually highly undesirable.

afterthought2 (commented):
I agree with Loaf that the call chain for installing/removing could use a second look, though maybe for slightly different reasons. To me, a very high priority is making sure that the organ lists on mobs remain 100% accurate. In the current setup (if I am reading correctly) this is the responsibility of the top-level mob call, but one is allowed to bypass it in some circumstances. It's possible that the exact usage avoids inaccuracies, but it's really hard to very from examining the system. I think that logic probably needs to be on organ procs, and positioned so that it's always called no matter what questionable stuff someone is doing.

Also worth double-checking that all the organ printers still function (including with blood input, etc.); they were always a sore point with the old system due to how organ spawn with partial information was handled. Possibly also the dead mob spawners, the dollys/dummies, etc.

Some of that should be unit-tested, but perhaps out of scope.

code/modules/mob/living/carbon/human/human.dm
PsyCommando: Good catch! It was just copy-pasta of the old code, so I didn't really look into it much. And there isn't even a reason for the null to be there in the first place weirdly.

code/modules/mob/living/carbon/human/human.dm
PsyCommando: You mean the holder_type? Yeah, I'm not entirely sure why.. I'll see what I can do!

code/modules/mob/living/carbon/human/human_organs.dm
PsyCommando: From what loaf told me, what he didn't like was the type checking. Or at least part of what he didn't like. Which makes sense. I did that to avoid having to mess with the organ lists. But yeah, its not ideal, so I've been applying some of his changes to the organ lists from his PR. Its not yet ready though.

And well, I'm not entirely sure why the chain should be started by the organs themselves? As it is currently with basically anything else, when you add something to something else, you add it via the future owner's procs. It makes sense in terms of hierarchy. 
Plus, the organs themselves are making direct changes and triggering things in the parent mob without being aware of the context. For instance, one of the common issue I've been having with the old way its done is how organs would keep triggering body or dna update at the worst possible time, just cause they feel like it. Feels like the mob initiating the chain makes it easier to deal with things like that, since the mob actually knows about its own state, while the organs, or even the caller, don't.

And there's a reason why on_install and install are separate. Another issue of the old code is that you couldn't selectively avoid/override triggering effects from organ installation/removal. Which cause problems when rejuvinating(Cause it just deleted all old organs in place and caused ui bugs and the like) or when replacing organs at runtime via code and not surgery. I try to keep the bare minimum in install, and then anything like dropping clothing, dying, and etc is done in the on_***** proc. 
Its pretty subtle, but it makes a huge difference in my opinion.

code/modules/organs/internal/_internal.dm
PsyCommando: Oh right! I left that commented because I was trying to see if it was actually needed. Because with the new setup, organs are created with dna info, and get the specie through it already. 
So, changing this post-init could cause unforeseen problems. XD

I'm just a bit unsure on the use cases here. Because, this bit of code would mean if we have prosthetic alien organs they could be slapped into any races without the downside that comes with the real thing.. And organs transplanting and prosthesis are already very buggy as it is. 
Simplest way imo is to just create them for a target specie, so validation can be done at whatever created it.

Honestly, I feel like overall the way prosthetics works might need some rethinking? I'd be down to make some changes if it comes to that.

code/modules/organs/internal/_internal.dm
PsyCommando: Well, the idea is that uninstall/install is self-contained. And since when we rejuv we actually just reset the organs (since I added a check to avoid having to recreate all organs and just carpet delete everthing) we only care that the organ itself is in an uninstalled state. Before "remove_rejuv" was just "removed()" but without uninstallation effects and was just clearing the mob refs so it could get deleted by create_organs properly. But now create_organs doesn't just mass delete everything so.. It was essentially a hack to prevent deletion triggering effects iirc.

Its not necessarily great, but I've been trying to plug in my changes to get things to work, and assuming we could always go back and fix this kind of things once we get around to fixing more stuff.

Rejuvination is still doing some hacky stuff to avoid organs breaking everything. It could always be done in other ways if we can work out rejuvination too.

code/modules/organs/internal/appendix.dm
PsyCommando: For this one, we don't know or care that we're on a mob, or inside a body part. (Even though, that probably shouldn't ever happen when inside a body part..)
And qdel already handles clearing parent refs. I guess, I don't even really need to call uninstall there.

mods/species/vox/organs_vox.dm
PsyCommando: The reason why I'm invoking it like this, is because there's a sleep in "prompt_revive_callback", and a blocking prompt yeah. This code gets called in "life()" sometimes, and dreamchecker doesn't like it, rightfully so. Would calling "prompt_revive_callback()" with its "waitfor FALSE" set work in that case? Cause I don't know if it can even tell.

afterthought2: Yes, setting `waitfor = FALSE` on the callback and invoking it directly here should be fine then, and the linter should pick that up.

code/modules/mob/living/carbon/human/human_organs.dm
afterthought2: I don't really care what starts the chain, and if you believe there's a good reason for the separation of the stuff below from `on_install` that's fine. Loaf and I don't love the istypes right above here.

I do think you should aim for 100% accuracy of the organ lists at all times, even during "temporary" installations or whatever (re: the two comments below too). That means that if the lists are modified in this proc, and you are installing/removing an organ, you need to commit to _only_ using this proc, treating the rest as internal/private.

PsyCommando: Don't worry, I don't like the istype either XD I've already redid it in a better way, actually using polymorphism for it. I'm currently integrating that with loaf's approach to organs lists for my next push. (Just been having issues with all human mobs working completely fine, but not spawning with eyes... I wish loaf had time to test his PR's code before XD)

And, I agree. I'm also aiming for 100% accuracy and consistency of the mob organs. 
Its just organs aren't always inside mobs. And sometimes, they can be inside mobs in a detached state (Basically the surgery code turns them into "implants" when detaching them, and essentially remove them from the organ list). So, some cases get a bit ambiguous on that front, and I want to reuse code as much as possible. I might rename the install/uninstall procs on organs to make their intended usage a bit more clear.. I'm also looking into reducing the amount of args on those procs..

I'm also considering getting rid of the auto-installation code in the organ init code, since it introduce more unknowns that I'm willing to deal with for a really dubious convenience advantage? And anyways, the only time we're creating organs actually inside a mob is in the create_missing_organ proc, and the VV menu...

mods/species/vox/organs_vox.dm
PsyCommando: Gotcha

code/modules/organs/internal/_internal.dm
PsyCommando: Decided to remove that code, since it was a hacky and counter productive way to guarantee something that we're already guaranteeing before we get to this point.

code/modules/organs/internal/appendix.dm
PsyCommando: Removed the call to uninstall here, and let qdel handle it.

code/modules/mob/living/carbon/human/human.dm
PsyCommando: Removed the extra nulling out of the holder type.

PsyCommando:
I'd love to have some feedback on the names I used when renaming the organ procs. install/uninstall being renamed to do_install and do_uninstall? I'm not entirely sure its the most intuitive name still. Maybe I should have went for on_**** but, I'm afraid it might cause some confusion and cause people to put install effects into those, instead of more basic stuff..

afterthought2 (approved):

PR #2186, opened 2021-10-30
Fixing decl UID test.

MistakeNot4892:
Thanks to @MarinaGryphon for pointing out the issues with the test.

### DIFF
--- a/code/datums/outfits/outfit.dm
+++ b/code/datums/outfits/outfit.dm
@@ -21,6 +21,7 @@ var/global/list/outfits_decls_by_type_
 
 /decl/hierarchy/outfit
 	name = "Naked"
+	abstract_type = /decl/hierarchy/outfit
 
 	var/uniform = null
 	var/suit = null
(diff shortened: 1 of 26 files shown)


MarinaGryphon (approved):
IDK if this review is even valid since, you know, I was the one to find some of this stuff, but also Loaf's a better coder than I am and I got it mostly-working, so.

afterthought2 (approved):
Classic error.

code/datums/outfits/outfit.dm
```diff
@@ -21,6 +21,7 @@ var/global/list/outfits_decls_by_type_
 
 /decl/hierarchy/outfit
 	name = "Naked"
+	abstract_type = /decl/hierarchy/outfit
```
PsiOmegaDelta, line 24: This outfit is valid for selection (or at least originally was designed to be a valid choice)

MistakeNot4892: It throws a very strange runtime comparing `Naked` against 0.

PR #2188, opened 2021-10-30
Updates some old code to use modern helpers

MarinaGryphon:
## Description of changes
Updates some old code to use modern helpers like get, get_area, and isEquipped.

Code modernisation.

## Authorship

Me

### DIFF
--- a/code/game/machinery/_machines_base/machinery_power.dm
+++ b/code/game/machinery/_machines_base/machinery_power.dm
@@ -9,7 +9,7 @@ This is /obj/machinery level code to properly manage power usage from the area.
 		if(A) A.power_use_change(old_power, new_power, power_channel)}
 
 // returns true if the area has power on given channel (or doesn't require power), defaults to power_channel.
-// May also optionally specify an area, otherwise defaults to src.loc.loc
+// May also optionally specify an area, otherwise defaults to get_area(src)
 /obj/machinery/proc/powered(var/chan = -1, var/area/check_area = null)
 
 	if(!src.loc)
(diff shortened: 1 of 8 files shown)


code/game/objects/items/weapons/tanks/tanks.dm
```diff
@@ -227,13 +227,7 @@ var/global/list/global/tank_gauge_cache = list()
...
 
 /obj/item/tank/ui_interact(mob/user, ui_key = "main", var/datum/nanoui/ui = null, var/force_open = 1)
-	var/mob/living/carbon/location = null
-
-	if(istype(loc, /obj/item/rig))		// check for tanks in rigs
-		if(istype(loc.loc, /mob/living/carbon))
-			location = loc.loc
-	else if(istype(loc, /mob/living/carbon))
-		location = loc
+	var/mob/living/carbon/location = get(src, /mob/living/carbon)
```
MistakeNot4892, line 230: Amazing proc name.

MistakeNot4892 (approved):

code/game/machinery/_machines_base/machinery_power.dm
```diff
@@ -21,7 +21,7 @@ This is /obj/machinery level code to properly manage power usage from the area.
 	//	return 1
 
 	if(!check_area)
-		check_area = src.loc.loc		// make sure it's in an area
+		check_area = get_area(src)		// make sure it's in an area
```
afterthought2, line 24: For the record, I believe this used to be a deliberate micro-optimization back when the `get_area` implementation was slow and this proc was being called hundreds of millions of times. Nowadays it's not called as much and `get_area` is as fast/faster than `:loc:loc`.

afterthought2 (approved):

PR #2192, opened 2021-10-30
adds graphite bricks to mapping/orders

anconfuzedrock:
## Description of changes
Adds graphite bricks so you can put them in maps and order them, mostly so you can power pacman generators.

pacman generators can actually be fueled consistently now, which is cool.

## Authorship
Rock

## Changelog

:cl:
tweak: graphite is now spawnable and orderable.
/:cl:

### DIFF
--- a/code/datums/supplypacks/materials.dm
+++ b/code/datums/supplypacks/materials.dm
@@ -52,6 +52,10 @@
 	contains = list(/obj/item/stack/material/reinforced/mapped/ocp/fifty)
 	containername = "osmium carbide plasteel sheets crate"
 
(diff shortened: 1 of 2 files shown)


MarinaGryphon:
Do note that I said bricks because that's what the default stack type for solids is. Personally I'd go for lumps or pucks.

anconfuzedrock:
iii'd rather not make things more complicated, but I can if so desired.

MistakeNot4892 (approved):
Lumps etc wouldn't work with the portable generator, it uses the default solid form of the material.

PR #2193, opened 2021-10-30
Adds fail state to gear tweak unit test.

MistakeNot4892:
## Description of changes
Adds a return value to gear tweaks and checks it in the unit test.

Runtimes two procs deep were not being reported as failures.

## Authorship
@MarinaGryphon identified the issue, I wrote the code.

## Changelog
Nothing player facing.

### DIFF
--- a/code/modules/client/preference_setup/loadout/gear_tweaks.dm
+++ b/code/modules/client/preference_setup/loadout/gear_tweaks.dm
@@ -1,3 +1,6 @@
+#define GEAR_TWEAK_SUCCESS 1
+#define GEAR_TWEAK_SKIPPED 2
+
 /datum/gear_tweak/proc/get_contents(var/metadata)
 	return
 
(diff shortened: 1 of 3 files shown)


code/modules/mob/new_player/new_player.dm
```diff
@@ -24,6 +24,8 @@
 /mob/new_player/Initialize()
 	. = ..()
 	verbs += /mob/proc/toggle_antag_pool
+	loc = null
```
afterthought2: I know I left it alone previously, but possibly this should be `forceMove(null)` or even `ASSERT(loc == null)`.

afterthought2 (dismissed):

code/modules/mob/new_player/new_player.dm
MistakeNot4892: I considered that but wasn't sure if it was deliberately skipping Entered() etc being called.

afterthought2: I don't know; they should never be spawned outside of the client/New anyway, but I guess you can bus them in (?).

MistakeNot4892: Removed these changes from this PR in any case.

afterthought2 (approved):

PR #2194, opened 2021-10-30
Removes/renames an unused horrible proc.

MistakeNot4892:
## Description of changes
Unifies `get()` and `get_holder_of_type()` as an `/atom` proc (actually a `/datum` proc but shh).
Calling it on a `/datum` will crash, this is to expose issues that previously went hidden with the other procs.
Fixes a couple of aforementioned issues (being called on extensions instead of extension holders).

Code quality.

## Authorship
Myself.

## Changelog
Nothing player facing.

### DIFF
--- a/code/datums/datum.dm
+++ b/code/datums/datum.dm
@@ -57,3 +57,10 @@
 /datum/proc/Process()
 	SHOULD_NOT_SLEEP(TRUE)
 	return PROCESS_KILL
+
+// This is more or less a helper to avoid needing to cast extension holders to atom.
+// Previously called get() and get_holder_of_type().
+// See /atom/get_recursive_loc_of_type() for actual logic.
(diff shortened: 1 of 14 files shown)


code/datums/datum.dm
```diff
@@ -57,3 +57,9 @@
 /datum/proc/Process()
 	SHOULD_NOT_SLEEP(TRUE)
 	return PROCESS_KILL
+
+// This is more or less a helper to avoid needing to cast extension holders to atom.
+// See /atom/get_recursive_loc_of_type() for actual logic.
+/datum/proc/get_recursive_loc_of_type(var/loc_type)
```
afterthought2, line 64: Can you add the old name(s) to the comment here so people searching for it have a greater chance to find it?

afterthought2 (dismissed):

afterthought2 (approved):

PR #2198, opened 2021-11-01
[MDB IGNORE] Removes or implements unused vars

MarinaGryphon:
## Description of changes
Removes a number of unused vars, implements ones where it made sense.
A number are left untouched/with comments, because I couldn't figure out how to implement them but removing them would be unacceptable.

Fewer unused vars is good, yeah?

## Authorship
Me, for the script and the removals/implementations. The team behind SpacemanDMM for the library/code I used to make the script.


PsiOmegaDelta:
Looks like there are mapped variables in a DMM belonging to `/datum/map_template/ruin/exoplanet/hydrobase` which refer to no a no longer existing variable (`/obj/machinery/power/tracker/var/id`)

code/modules/vehicles/vehicle.dm
```diff
@@ -1,7 +1,6 @@
 //Dummy object for holding items in vehicles.
 //Prevents items from being interacted with.
 /datum/vehicle_dummy_load
-	var/name = "dummy load"
```
MistakeNot4892, line 4: This isn't necessarily unused, defining a name var means datums will use it in modal inputs instead of their type.

code/modules/reagents/reagent_containers/glass/bottle/robot.dm
```diff
@@ -4,7 +4,12 @@
 	possible_transfer_amounts = @"[5,10,15,25,30,50,100]"
 	atom_flags = ATOM_FLAG_OPEN_CONTAINER
 	volume = 60
-	var/reagent = ""
+	var/reagent
+
+/obj/item/chems/glass/bottle/robot/stabilizer/Initialize()
```
MistakeNot4892: `stabilizer` seems unintentional.

code/modules/vehicles/vehicle.dm
MarinaGryphon: I think the issue is that it was also never set, so it would always show up as "dummy load".

code/datums/trading/traders/food.dm
```diff
@@ -102,7 +102,6 @@
 							/obj/item/chems/food/fruit_slice                = TRADER_BLACKLIST,
 							/obj/item/chems/food/slice                      = TRADER_BLACKLIST_ALL,
 							/obj/item/chems/food/grown                      = TRADER_BLACKLIST_ALL,
-							/obj/item/chems/food/human                      = TRADER_BLACKLIST_ALL,
```
MistakeNot4892, line 105: Feels like this should be `/obj/item/chems/food/human/burger` which itself should just be `humanburger`.

code/datums/traits/metabolically_inert.dm
```diff
@@ -1,3 +1,3 @@
 /decl/trait/metabolically_inert
 	name = "Metabolically Inert"
-	description = "For good and ill there is little that this creature metabolizes"
\ No newline at end of file
+	description = "For better or worse, there is little that this creature metabolizes."
```
MistakeNot4892, line 3: `good or ill` would be equally valid here.

code/game/objects/items/devices/spy_bug.dm
```diff
@@ -17,7 +17,6 @@
 	origin_tech = "{'programming':1,'engineering':1,'esoteric':3}"
 
 	var/obj/item/radio/spy/radio
-	var/obj/machinery/camera/spy/camera
```
MistakeNot4892, line 20: While you're in the area it might be worth changing the examine() text to remove the 'must be attached to device' requirement.

code/game/turfs/simulated/floor.dm
```diff
@@ -20,7 +20,6 @@
 	var/flooring_override
 	var/initial_flooring
 	var/decl/flooring/flooring
-	var/mineral = DEFAULT_WALL_MATERIAL
```
MistakeNot4892, line 23: Implementing this is on my todo lol.

MistakeNot4892:
Conflicts.

MistakeNot4892:
Superceded and conflictpilled.

PR #2200, opened 2021-11-02
Ctrl-click grabs are now deferred to an empty hand

MarinaGryphon:
## Description of changes
Grabs with ctrl-click will now go into any available empty hand, if one exists.

Highly requested QoL change from Hearth. I think this is an acceptable compromise between requiring hand-switching (which is sensible on grab intent) and the old pull functionality. Now, ctrl-clicking will put the grab in any hand

## Changelog

:cl:
tweak: Grabs with ctrl-click will now go into any available empty hand, if one exists.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human_grabs.dm
+++ b/code/modules/mob/living/carbon/human/human_grabs.dm
@@ -1,7 +1,10 @@
-/mob/living/carbon/human/add_grab(var/obj/item/grab/grab)
-	. = put_in_active_hand(grab)
+/mob/living/carbon/human/add_grab(var/obj/item/grab/grab, var/defer_hand = FALSE)
+	if(defer_hand)
+		. = put_in_hands(grab)
+	else
+		. = put_in_active_hand(grab)
 
-/mob/living/carbon/human/can_be_grabbed(var/mob/grabber, var/target_zone)
+/mob/living/carbon/human/can_be_grabbed(var/mob/grabber, var/target_zone, var/defer_hand = FALSE)
 	. = ..()
 	if(.)
 		var/obj/item/organ/external/organ = get_organ(target_zone)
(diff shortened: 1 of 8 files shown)


code/modules/mob/living/living_grabs.dm
```diff
@@ -37,16 +41,16 @@
 
 	face_atom(target)
 	var/obj/item/grab/grab
-	if(ispath(grab_tag, /decl/grab) && can_grab(target, zone_sel?.selecting) && target.can_be_grabbed(src, zone_sel?.selecting))
-		grab = new /obj/item/grab(src, target, grab_tag)
+	if(ispath(grab_tag, /decl/grab) && can_grab(target, zone_sel?.selecting, defer_hand = defer_hand) && target.can_be_grabbed(src, zone_sel?.selecting))
+		grab = new /obj/item/grab(src, target, grab_tag, defer_hand = defer_hand)
```
MistakeNot4892: Can't use named keywords for `Initialize` since you're actually calling `New`.

code/modules/mob/living/living_grabs.dm
```diff
@@ -1,8 +1,12 @@
-/mob/living/proc/can_grab(var/atom/movable/target, var/target_zone)
+/mob/living/proc/can_grab(var/atom/movable/target, var/target_zone, var/defer_hand = FALSE)
```
MistakeNot4892, line 1: Need to add the var to the hostile simplemob override.

code/modules/mob/living/carbon/human/human_grabs.dm
```diff
@@ -9,7 +12,7 @@
 			to_chat(grabber, SPAN_WARNING("\The [src] is missing that body part!"))
 			return FALSE
 		if(grabber == src)
-			var/using_slot = get_active_held_item_slot()
+			var/using_slot = get_empty_hand_slot()
```
MistakeNot4892: This is going to introduce inconsistencies since it will no longer fail when your main hand is occupied.

MarinaGryphon: I guess failing to grab yourself with ctrl-click sometimes is okay?

MarinaGryphon: (or I could pass defer_hand to yet another proc)

MistakeNot4892: Would prefer you pass the arg honestly.

MistakeNot4892 (approved):

PR #2204, opened 2021-11-03
Ports compasses and GPS from Polaris.

MistakeNot4892:
## Description of changes
Ports the GPS system and compass overlay system from Polaris. This will need a ton of reviewing/work I suspect.

Adds a visual effect and tracking system to GPS units.

## Authorship
Myself.

## Changelog
:cl:
tweak: GPS units now provide a visual compass to track each other.
/:cl:


MistakeNot4892:
This is done other than visual effects for the scanner beam and radius, and content to be scanned.

MistakeNot4892:
Ended up gutting the marker info concept from this (was going to show current heading and speed for movables) but the overmap and handheld functionality is there.

MistakeNot4892:
image

MistakeNot4892:
Needs `legitimate salvage` label, but is complete and mergeable if someone wants to straight up cherrypick it.
Taking a long break from SS13, may come back to this someday.

MistakeNot4892:
I've snipped out the unfinished overmap compass stuff and refined the base compass stuff a bit, this is ready for review.

MistakeNot4892:
https://cdn.discordapp.com/attachments/678820025355862048/982190334643695626/dreamseeker_DxkhpAgsu8.mp4

MistakeNot4892:
Overmap GPS stuff is back in with the maptext stuff removed, this works alright:
https://cdn.discordapp.com/attachments/492799900145418270/982225100378095656/dreamseeker_xI7m0wTpLT.mp4

MistakeNot4892:
Pls review.

code/_helpers/text.dm
```diff
@@ -641,4 +641,42 @@ var/global/list/fullstop_alternatives = list(".", "!", "?")
...
+		if(247.5)
+			return "West-Southwest"
+		if(270)
+			return "West"
+		if(292.5)
+			return "West-Northwest"
+		if(315)
+			return "Northwest"
+		if(337.5)
+			return "North-Northwest"
```
afterthought2, line 682: Need to include 360 here (as "North").

code/game/objects/compass/compass_holder.dm
```diff
@@ -0,0 +1,134 @@
...
+
+/obj/compass_holder/proc/get_heading_strength()
+	return 1
+
+/obj/compass_holder/proc/get_heading_angle()
+	var/atom/A = loc
+	// Bit of a strange loop - grab the highest atom in our stack that isn't a turf.
+	while(istype(A) && A.loc && !istype(A.loc, /turf))
+		A = A.loc
+	if(istype(A) && !isturf(A))
```
afterthought2, line 79: Why isturf here but istype just above?

code/game/objects/compass/compass_overmap.dm
```diff
@@ -0,0 +1,85 @@
...
+	if(!istype(owner))
+		PRINT_STACK_TRACE("Overmap compass initialized in non-helm atom: [type]")
+		return INITIALIZE_HINT_QDEL
+
+	compass_heading_marker = new /image/compass_marker
+	compass_heading_marker.maptext = STYLE_SMALLFONTS("<center>△</center>", 7, COLOR_WHITE)
+	compass_heading_marker.filters = filter(type="drop_shadow", color = "[owner.linked.color]aa", size = 2, offset = 1,x = 0, y = 0)
+	compass_heading_marker.layer = HUD_BASE_LAYER
+	compass_heading_marker.plane = HUD_PLANE
+	compass_heading_marker.color = owner.linked.color
```
afterthought2: What if linked is null? It's possible it's fine, but I can't tell. Same for other procs below.

code/game/objects/compass/compass_overmap.dm
```diff
@@ -0,0 +1,85 @@
...
+
+	. = ..()
+
+	var/turf/owner_turf = get_turf(owner.linked)
+	var/list/seen_waypoint_ids = list()
+	for(var/key in owner.known_sectors)
+		var/datum/computer_file/data/waypoint/R = owner.known_sectors[key]
+		var/wp_id = "\ref[R]"
+		set_waypoint(wp_id, uppertext(R.fields["name"]), R.fields["x"], R.fields["y"], owner_turf.z, R.fields["color"] || COLOR_SILVER)
```
afterthought2, line 35: What if owner_turf is null?

code/game/objects/items/devices/gps.dm
```diff
@@ -1,28 +1,423 @@
...
+	. = ..()
+	update_holder()
+
+/obj/item/gps/Process()
+	if(!tracking)
+		is_in_processing_list = FALSE
+		return PROCESS_KILL
+	update_holder()
+	if(holder)
+		update_compass(TRUE)
```
afterthought2, line 104: Causing me pain, but maybe it's superior to purely using events. If the waypoints don't move, why not just do this with a moved event?

I will note that if you did do this purely with events, you'd need to subscribe to the mob login (and possibly logout) events as well.

code/game/objects/items/devices/gps.dm
```diff
@@ -1,28 +1,423 @@
...
+
+/obj/item/gps/Destroy()
+	STOP_PROCESSING(SSobj, src)
+	is_in_processing_list = FALSE
+	global.all_gps_units -= src
+	. = ..()
+	update_holder()
```
afterthought2: I think this is wrong, you should call this before `..()`. You also need to unsubscribe from all events manually here.

code/game/objects/items/devices/gps.dm
```diff
@@ -1,28 +1,423 @@
...
+	QDEL_NULL(compass)
+
+/obj/item/gps/proc/can_track(var/obj/item/gps/other, var/reachable_z_levels)
+	if(!other.tracking || other.emped || other.hide_signal)
+		return FALSE
+	reachable_z_levels = reachable_z_levels || list()
```
afterthought2: LAZYINITLIST? ||=?

code/game/objects/items/devices/gps.dm
```diff
@@ -1,28 +1,423 @@
...
+		to_chat(user, SPAN_WARNING("\The [src] is busted!"))
+		return
+	
+	toggle_tracking()
+	if(tracking)
+		to_chat(user, SPAN_NOTICE("\The [src] is now tracking, and visible to other GPS devices."))
+	else
+		to_chat(user, SPAN_NOTICE("\The [src] is no longer tracking, or visible to other GPS devices."))
+
+/obj/item/gps/proc/toggle_tracking()
```
afterthought2: Please don't have two procs named the same thing except one using normal style and one not.

code/game/objects/items/devices/gps.dm
```diff
@@ -1,28 +1,423 @@
...
+			is_in_processing_list = TRUE
+			START_PROCESSING(SSobj, src)
+	else
+		is_in_processing_list = FALSE
+		STOP_PROCESSING(SSobj, src)
+	update_compass()
+	update_holder()
+	update_icon()
+
+/obj/item/gps/emp_act(severity)
```
afterthought2, line 185: Does this actually toggle it off somewhere (don't see it doing so)? If not, I think you should be able to toggle it off if it's emp'd.

code/game/objects/items/devices/gps.dm
```diff
@@ -1,28 +1,423 @@
...
+/obj/item/gps/on_update_icon()
+
+	cut_overlays()
+
+	var/image/I
+	for(var/decal in decals)
+		if(check_state_in_icon(decal, decal_icon))
+			I = image(decal_icon, decal)
+			I.color = decals[decal]
+			I.appearance_flags |= RESET_COLOR
```
afterthought2, line 211: You aren't doing anything with `I` before moving on to the next loop iteration.

code/game/objects/items/devices/gps.dm
```diff
@@ -1,28 +1,423 @@
...
+		if(gps_list.len)
+			dat += "<tr><td colspan = 6><b>Detected signals</b></td></tr>"
+			for(var/gps in gps_data["gps_list"])
+
+				var/weakref/gps_weakref = gps["ref"]
+				if(!istype(gps_weakref))
+					continue
+				var/obj/item/gps/gps_obj = gps_weakref.resolve()
+				if(!istype(gps_obj) || QDELETED(gps_obj))
+					continue
```
afterthought2: I don't get it. We literally just created this weakref in the helper at the head of this proc, and now we unwrap it synchronously here? Why?

code/game/objects/items/devices/gps.dm
```diff
@@ -1,28 +1,423 @@
...
+		I.color = screen_color
+		I.appearance_flags |= RESET_COLOR
+		add_overlay(I)
+
+/obj/item/gps/attack_self(mob/user)
+	display(user)
+
+ // Compiles all the data not available directly from the GPS
+ // Like the positions and directions to all other GPS units
+/obj/item/gps/proc/display_list()
```
afterthought2: Yeah, not really getting the UI model here. Separating this from the template layer maybe makes sense when using nanoUI, but why here? You're bypassing a bunch of safety stuff, potentially.

code/game/objects/items/devices/gps.dm
```diff
@@ -1,28 +1,423 @@
...
+			var/new_colour = input("Enter a new tracking color.", "GPS Waypoint Color") as color|null
+			if(new_colour && istype(gps) && !QDELETED(gps) && holder == user && !user.incapacitated())
+				to_chat(user, SPAN_NOTICE("You adjust the colour \the [src] is using to highlight [gps.gps_tag]."))
+				LAZYSET(tracking_devices, href_list["track_color"], new_colour)
+				update_compass()
+				. = TOPIC_HANDLED
+
+	if(href_list["tag"])
+		var/a = input("Please enter desired tag.", name, gps_tag) as text
+		a = uppertext(copytext(sanitize(a), 1, 11))
```
afterthought2, line 359: Are you sure you want to sanitize here? The main thing this proc will do is html-encode & and similar characters. If you _do_ in fact want that, you may need to decode in the default of `input` above, though as you are also truncating it won't always work.

afterthought2 (commented):
Lots of code, hard to review.

code/_helpers/text.dm
MistakeNot4892: `%360`, a value of 360 will be 0.

code/game/objects/compass/compass_overmap.dm
MistakeNot4892: Null color is valid, not sure what you mean?

code/game/objects/compass/compass_holder.dm
MistakeNot4892: Copypaste artifact I suspect, will check it over.

code/game/objects/compass/compass_overmap.dm
MistakeNot4892: That would imply that the object the GPS overlay was trying to draw was not on the overmap (or inside an atom on the overmap) which I'm not sure is a thing? Won't helm consoles etc all explode if you try to look at the overmap from nullspace?

code/game/objects/items/devices/gps.dm
MistakeNot4892: Waypoints do move, they are generally other GPS units. I thought a periodic update would scale better than having every GPS subscribe to every other GPS.
Why would I need to hook login/logout for the mob? Other than adding the compass to screen.

code/game/objects/items/devices/gps.dm
MistakeNot4892: To be honest all of the display stuff is something I tidied a little on Polaris and then ported over. I may just write a NanoUI template and burn it to the ground.

code/_helpers/text.dm
afterthought2: Yes, but 359 rounds to 360. Maybe you can change the order of rounding and mod instead, idk.

code/game/objects/compass/compass_overmap.dm
afterthought2: if the `linked` var in this line is null, this will runtime. It's unclear to me from code inspection if this is possible, and want to make sure you are certain it's not.

code/game/objects/compass/compass_overmap.dm
afterthought2: I don't know; the main concern is with deletion, e.g. if the console is deleted with the UI open.

code/game/objects/items/devices/gps.dm
afterthought2: When the client logs back in, things are not automatically placed back on the screen. This block here will probably trigger within 2 seconds anyway and do that, so it's not the biggest deal, but I would probably listen for the login event anyway. The logout event is less clear; you'd only need that if you wanted to do cleanup, like e.g. exit out of compass on logout.

code/game/objects/items/devices/gps.dm
afterthought2: I think the UI stuff is weird but wouldn't block over most of it--with this weakref thing being the exception.

code/game/objects/compass/compass_overmap.dm
MistakeNot4892: Added a check above this block.

code/game/objects/items/devices/gps.dm
MistakeNot4892: I'm a bit confused by this comment, I was under the impression it was standard practice to sanitize after an `as text` input. This is meant to be a very short all caps string so I'm not too fussed if people want to put weird stuff into it and get nonsense as their default.

code/game/objects/items/devices/gps.dm
MistakeNot4892: No, this is correct. If the loc is not the holder it unsubscribes and clears relevant vars.

code/game/objects/items/devices/gps.dm
MistakeNot4892: Used LAZYDISTINCTADD down the bottom instead.

code/game/objects/items/devices/gps.dm
MistakeNot4892: Merged.

code/game/objects/items/devices/gps.dm
MistakeNot4892: Added a toggle.

code/game/objects/items/devices/gps.dm
MistakeNot4892: Whoops, fixed.

code/game/objects/items/devices/gps.dm
MistakeNot4892: It's honestly nonsensical. I think I got sick of the revision on Polaris and they didn't review it thoroughly at all.

code/game/objects/items/devices/gps.dm
MistakeNot4892: I think I am going to leave the `process()` loop and not hook the client login event, unless you feel strongly about it. Registering to other GPS in particular is going to be a complicated thing to implement.

code/game/objects/compass/compass_overmap.dm
MistakeNot4892: Gotcha, added a check.

code/_helpers/text.dm
MistakeNot4892: Went with rounding before modulo.

code/game/objects/items/devices/gps.dm
MistakeNot4892: I think weakref was my halfhearted attempt at GC cleanliness without reading the code properly. Have split it out in favour of \ref now since that can then be used by OnTopic().

MistakeNot4892:
Now uses NanoUI:
https://cdn.discordapp.com/attachments/256853514624434177/984420518331547648/unknown.png

code/game/objects/items/devices/gps.dm
MistakeNot4892: Gutted it.

code/game/objects/items/devices/gps.dm
MistakeNot4892: Gutted it.

code/game/objects/items/devices/gps.dm
afterthought2: Yes, it unregisters--after the parent call checked that you failed to unregister for those events, logged an error, and did a really expensive general search for possible event registrations.

You need to unregister events before calling `..()`.

MistakeNot4892: Added a force arg to it and called it before ..()

MistakeNot4892:
Looks like I've pulled in some invalid commits at some point.

code/game/objects/items/devices/gps.dm
```diff
@@ -1,31 +1,388 @@
...
+			LAZYSET(showing_tracked_names, gps_ref, TRUE)
+			to_chat(user, SPAN_NOTICE("\The [src] is now tracking [gps.gps_tag]."))
+			update_compass()
+			. = TOPIC_REFRESH
+
+	if(href_list["track_color"])
+		var/obj/item/gps/gps = locate(href_list["track_color"])
+		if(istype(gps) && !QDELETED(gps))
+			var/new_colour = input("Enter a new tracking color.", "GPS Waypoint Color") as color|null
+			if(new_colour && istype(gps) && !QDELETED(gps) && holder == user && !user.incapacitated())
```
afterthought2: Correct check is `CanInteract` here. Can grab the state from the third arg, even.

code/game/objects/items/devices/gps.dm
```diff
@@ -1,31 +1,388 @@
...
+				to_chat(user, SPAN_NOTICE("You adjust the colour \the [src] is using to highlight [gps.gps_tag]."))
+				LAZYSET(tracking_devices, href_list["track_color"], new_colour)
+				update_compass()
+				. = TOPIC_REFRESH
+
+	if(href_list["tag"])
+		var/a = input("Please enter desired tag.", name, gps_tag) as text
+		a = uppertext(copytext(sanitize(a), 1, 11))
+		if(in_range(src, user))
```
afterthought2: `CanInteract`, probably. If you _actually_ want what this proc is doing (doubtful) use `CanPhysicallyInteract`, I guess.

code/modules/overmap/ships/computers/helm.dm
```diff
@@ -183,24 +200,48 @@ var/global/list/overmap_helm_computers
...
 		if(R)
 			known_sectors.Remove(R.fields["name"])
+			compass.clear_waypoint("\ref[R]")
 			qdel(R)
 
+	if(href_list["toggle_wp_tracking"])
+		var/datum/computer_file/data/waypoint/R = locate(href_list["toggle_wp_tracking"])
+		if(R && (R.fields["name"] in known_sectors))
+			R.fields["tracking"] = !R.fields["tracking"]
+			to_chat(usr, SPAN_NOTICE("[R.fields["name"]] is [R.fields["tracking"] ? "now" : "no longer"] being tracked."))
```
afterthought2: `user`

afterthought2 (approved):

PR #2206, opened 2021-11-05
Apply KEEP_TOGETHER to humantype mobs

MarinaGryphon:
## Description of changes
Applies KEEP_TOGETHER to humantype mobs.

In testing, this fixed the issue of non-32x32 mobs with 32x32 overlays falling apart/dismantling when transforms are applied.

## Authorship
Thanks to Loaf for suggesting this.


MistakeNot4892:
There's some client performance impact from this, but humans are already probably the most expensive mob in the game to render so I don't think it'll matter practically. Worth noting Polaris/Virgo apply this at a much more general level and generally don't experience client performance drops.

MistakeNot4892:
As noted on Discord, human update_transform needs tweaking.

MistakeNot4892 (approved):

PR #2209, opened 2021-11-07
Machines rely more on dex checks than brain damage, dex checks now influenced by brainloss

Mazianni:
## Description of changes

Machinery no longer has the weird brainloss check that makes you completely incapable of using machines.
Instead, human dexterity is influenced by a malus to dexterity over a certain threshold of brain damage.

It's really bad that if you have 55 brain damage, you can't do anything. Even press buttons.

This PR changes that behavior. As of current, over 30 brain damage, for every 10 brainloss you take a -1 to dexterity. Meaning you can take 110 brainloss before you are completely incapable of using objects.

## Authorship

Mazian/Andromeda-K22

## Changelog

:cl:
tweak: machines no longer directly check brainloss, but instead rely fully on dexterity checks.
balance: dexterity is influenced by brain damage, starting at 30 brainloss (configurable). Dexterity loss is on a sliding scale.
/:cl:

### DIFF
--- a/code/controllers/configuration.dm
+++ b/code/controllers/configuration.dm
@@ -255,6 +255,8 @@ var/global/list/gamemode_cache = list()
 
 	var/no_throttle_localhost
 
+	var/dex_malus_brainloss_threshold = 30 //The threshold of when brainloss begins to affect dexterity. 
+
 	var/static/list/protected_vars = list(
 		"comms_password",
 		"ban_comms_password",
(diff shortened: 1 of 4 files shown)


MistakeNot4892 (dismissed):
Looks good. Only suggestions I would have would be making that `!silent` check not repeated, and maybe swapping the magic numbers in the dex_malus calc for macros.

MistakeNot4892 (approved):

PR #2222, opened 2021-11-25
Hearth fix theft.

MistakeNot4892:
Steals a bunch of fixes from downstream, with thanks.


code/game/objects/structures/stool_bed_chair_nest/bed.dm
```diff
@@ -44,14 +44,15 @@
 // Reuse the cache/code from stools, todo maybe unify.
 /obj/structure/bed/on_update_icon()
 	..()
```
MistakeNot4892: Note to self: parent call does `overlays.Cut()`, will not cooperate with these changes.

comma (approved):

PR #2227, opened 2021-11-27
General machinery optimization

Gaxeer:
## Description of changes

General machinery optimization, escpecially when change_power() proc is called

- significantly decrease CPU consumption by change_power() proc (up to 20-30 times)

## Authorship

@Vallat and his PR https://github.com/ss220-space/Baystation12/pull/90

## Changelog

:cl:
tweak: optimize looping through area machine when change_power() is called by creating area machinery list and using it
tweak: optimize update() calling, to not being called when not needed
/:cl:

### DIFF
--- a/code/game/area/area_power.dm
+++ b/code/game/area/area_power.dm
@@ -4,6 +4,9 @@
 #define ENVIRON 3
 */
 
+/area
+	var/list/machinery_list
+
 /area/proc/powered(var/chan)		// return true if the area has power to given channel
 	if(!requires_power)
 		return 1
@@ -23,7 +26,7 @@
 
 // called when power status changes
 /area/proc/power_change()
-	for(var/obj/machinery/M in src)	// for each machine in the area
+	for(var/obj/machinery/M in machinery_list)	// for each machine in the area
 		M.power_change()			// reverify power status (to update icons etc.)
 	if (atmosalm || fire || eject || party)
 		update_icon()
@@ -74,7 +77,7 @@
 	used_equip = 0
 	used_light = 0
 	used_environ = 0
-	for(var/obj/machinery/M in src)
+	for(var/obj/machinery/M in machinery_list)
 		switch(M.power_channel)
 			if(EQUIP)
 				used_equip += M.get_power_usage()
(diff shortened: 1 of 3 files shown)


code/game/machinery/_machines_base/machinery.dm
```diff
@@ -134,12 +134,20 @@ Class Procs:
...
 /obj/machinery/Destroy()
 	if(istype(wires))
 		QDEL_NULL(wires)
 	SSmachines.machinery -= src
 	QDEL_NULL_LIST(component_parts) // Further handling is done via destroyed events.
 	STOP_PROCESSING_MACHINE(src, MACHINERY_PROCESS_ALL)
+
+	var/area/A = get_area(src)
+	if(A)
+		LAZYREMOVE(A.machinery_list, src) // Remove machinery object from area machinery list
```
MistakeNot4892, line 150: This needs to be updated on move.

code/game/machinery/_machines_base/machinery.dm
```diff
@@ -134,12 +134,20 @@ Class Procs:
 	RefreshParts()
 	power_change()
 
+	var/area/A = get_area(src)
+	if(A)
+		LAZYADD(A.machinery_list, src) // Add machinery object to area machinery list
```
MistakeNot4892, line 139: Would suggest using `weakref(src)` and then `foo.resolve()` in the loop over the list, to avoid GC problems if machinery isn't properly cleaned up from the area lists.

code/game/area/area_power.dm
```diff
@@ -23,7 +26,7 @@
 
 // called when power status changes
 /area/proc/power_change()
-	for(var/obj/machinery/M in src)	// for each machine in the area
+	for(var/obj/machinery/M in machinery_list)	// for each machine in the area
```
MistakeNot4892, line 29: If you use the weakref suggestion below, this will need to be changed to var/weakref. Either way should add `AS_ANYTHING` before `in` as the list will only contain strictly typed objects.

MistakeNot4892:
Generally in favour of this. There aren't that many areas in the game so memory impact is minimal (per discussions about memory use being a problem on Discord earlier).

Gaxeer:
This PR is no longer needed because of https://github.com/NebulaSS13/Nebula/pull/2230

PR #2228, opened 2021-11-27
Expand and refactor ailment treatment

MarinaGryphon:
## Description of changes
- Refactors ailment treatment code in general for better quality.
- Some ailments can now be treated with a specific CE_EFFECT of above a certain strength.
- Ailments that required painkillers now require CE_PAINKILLER strength 25.
- Cramps are now treated with sedatives (midazolam on Hearth) instead of painkillers.

- More consistent token-replacement functionality; a few messages still used `capitalize(replacetext(foo, "$ORGAN$", organ.name))` instead of `replace_tokens(foo)`.
- "Treatement" has been bothering me for a while.
- Ailment treatment reagent checking makes more logical sense and uses a previously-unused helper. This means it also accepts subtypes.
- CE_PAINKILLER makes more sense to check than a specific dosage of a certain kind of painkiller.
- It makes more sense to treat cramps with muscle relaxers than with painkillers.

## Authorship
Loaf for the original implementation, the ISEO Endeavour medical staff for bringing these issues to my attention, and myself for the code in this PR.

## Changelog

:cl:
tweak: Cramps are now treated with 1u sedative instead of 1u opiate painkillers.
tweak: Other ailments that were treated with painkillers can now be treated with anything that causes painkilling effects.
tweak: Subtypes of treatment reagents now count for treating ailments.
/:cl:

### DIFF
--- a/code/modules/organs/organ.dm
+++ b/code/modules/organs/organ.dm
@@ -145,22 +145,16 @@
 		handle_germ_effects()
 
 	if(owner && length(ailments))
-		for(var/datum/ailment/ailment in ailments)
-			if(!ailment.treated_by_reagent_type)
-				continue
-			var/treated
-			var/datum/reagents/bloodstr_reagents = owner.get_injected_reagents()
-			if(bloodstr_reagents)
-				if(REAGENT_VOLUME(bloodstr_reagents, ailment.treated_by_reagent_type) >= ailment.treated_by_reagent_dosage)
-					treated = bloodstr_reagents
-				else if(REAGENT_VOLUME(owner.reagents, ailment.treated_by_reagent_type) >= ailment.treated_by_reagent_dosage)
-					treated = owner.reagents
-				else
-					var/datum/reagents/ingested = owner.get_ingested_reagents()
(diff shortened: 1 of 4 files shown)


code/modules/organs/organ.dm
```diff
@@ -145,22 +145,16 @@
...
-					treated = bloodstr_reagents
-				else if(REAGENT_VOLUME(owner.reagents, ailment.treated_by_reagent_type) >= ailment.treated_by_reagent_dosage)
-					treated = owner.reagents
-				else
-					var/datum/reagents/ingested = owner.get_ingested_reagents()
-					if(ingested && REAGENT_VOLUME(ingested, ailment.treated_by_reagent_type) >= ailment.treated_by_reagent_dosage)
-						treated = ingested
-			if(treated)
-				ailment.was_treated_by_medication(treated)
+		ailment_loop:
```
MistakeNot4892: As discussed I think this should be split out into proc calls instead of using this structure. We generally do not permit goto to be used and this is functionally a goto.

comma (approved):

PR #2230, opened 2021-11-27
Changing area power updates to use an event.

MistakeNot4892:
## Description of changes
Areas raise an event on power change instead of iterating machinery in contents.
Alternative to #2227 that hopefully avoids memory and GC issues while also removing atmos pipes from the power change update lists.

Shaves 65k /obj/machinery/power_change calls off the grid check event and appears to drastically reduce the number of atoms iterated over by area power changes.

## Authorship
Gaxer raised the issue and PR'd the first implementation, I coded this one.
## Changelog
Nothing player facing.

### DIFF
--- a/code/datums/observation/area_power_change.dm
+++ b/code/datums/observation/area_power_change.dm
@@ -1,0 +1,11 @@
+//	Observer Pattern Implementation: Destroyed
+//		Registration type: /area
+//
+//		Raised when: An /area's power state changes.
+//
+//		Arguments that the called proc should expect:
+//			/area/power_changed: The instance that had a power change.
+
+/decl/observ/area_power_change
+	name = "Area Power Change"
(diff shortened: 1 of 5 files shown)


comma (dismissed):

afterthought2:
Sure, this is fine by me. The var could probably be replaced with checking something like `(use_power != POWER_USE_OFF)` possibly with `&& !(stat_immune & NOPOWER)`.

I guess I prefer this to the other implementation, but don't care strongly. I didn't do this before because this is an optimization mostly relevant to power outages, whereas all my previous work mostly targeted routine ticks. It's also not 100% clear it's an optimization at all without testing (but it looks like you've tested).

afterthought2 (dismissed):

afterthought2 (approved):

PR #2233, opened 2021-11-28
Add reboot hook to world/Reboot

MarinaGryphon:
## Description of changes
Adds a reboot hook to world/Reboot

Sometimes things need to be called on reboot but prior to world/Del(), and creating a dummy subsystem is overkill especially compared to how simple hooks are.

## Authorship
Me.

### DIFF
--- a/code/controllers/hooks-defs.dm
+++ b/code/controllers/hooks-defs.dm
@@ -29,6 +29,12 @@
 /hook/shutdown
 
 /**
+ * Reboot hook.
+ * Called in world.dm prior to the parent call in world/Reboot.
+ */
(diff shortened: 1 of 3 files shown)


code/controllers/hooks-defs.dm
```diff
@@ -28,6 +28,12 @@
  */
 /hook/shutdown
 
+/**
+ * Reboot hook.
+ * Called in world.dm prior to the parent call in world/Reboot.
```
MistakeNot4892, line 33: This is tripping the `in world` CI regex.

MistakeNot4892 (approved):

PR #2234, opened 2021-11-28
Some material stuff and numbers

SolatK:
## Description of changes
add formula for boiling temp to depends on pressure and some specific numbers for materials
изображение

In the future, it will be easier to add more realistic behavior of substances


MistakeNot4892:
Might be worth changing the codex line to say 'under standard atmospheric conditions it will boil at 1000K' so it's clear it can change.

SolatK:
Like this?

code/modules/materials/_materials.dm
```diff
@@ -333,11 +334,15 @@ INITIALIZE_IMMEDIATE(/obj/effect/gas_overlay)
 /decl/material/proc/products_need_process()
 	return (radioactivity>0) //todo
 
+
+//Clausius–Clapeyron relation
+/decl/material/proc/get_boiling_temp(var/pressure = ONE_ATMOSPHERE)
+	return (1 / (1/boiling_point) - ((R_IDEAL_GAS_EQUATION * log(pressure / ONE_ATMOSPHERE)) / (latent_heat * molar_mass)))
```
MistakeNot4892: `boiling_point` might be 0 or null - should probably return INFINITY if it's not >0.

MistakeNot4892 (approved):

PR #2240, opened 2021-11-30
update_transform: respect human direction for turn_angle

quardbreak:
## Description of changes

Title.

* The direction you are facing (EAST/WEST) now influences the direction you lay down in.
* canface() now checks for INCAPACITATION_UNRESISTING which means you can change facing direction ~~and pixelshift~~ while lying down, but not while dead.

## Authorship

@NotRanged as author of an improvements.

## Changelog

:cl: NotRanged
tweak: Lying down while facing left or right now lays you down left or right.
tweak: You can now change facing direction while lying down.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/update_icons.dm
+++ b/code/modules/mob/living/carbon/human/update_icons.dm
@@ -231,7 +231,14 @@ Please contact me on #coderbus IRC. ~Carn x
 	// Scale/translate/rotate and apply the transform.
 	var/matrix/M = matrix()
 	if(lying)
-		M.Turn(pick(90, -90))
+		var/turn_angle
+		if(dir & WEST)
+			turn_angle = -90
+		else if(dir & EAST)
+			turn_angle = 90
+		else 
+			turn_angle = pick(-90, 90)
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/carbon/human/update_icons.dm
```diff
@@ -231,7 +231,12 @@ Please contact me on #coderbus IRC. ~Carn x
 	// Scale/translate/rotate and apply the transform.
 	var/matrix/M = matrix()
 	if(lying)
-		M.Turn(pick(90, -90))
+		var/turn_angle = pick(-90, 90)
+		if(dir == WEST)
+			turn_angle = -90
+		if(dir == EAST)
+			turn_angle = 90
```
MistakeNot4892, line 238:
````
	var/turn_angle
	if(dir & WEST)
		turn_angle = -90
	else if(dir & EAST)
		turn_angle = 90
	else 
		turn_angle = pick(-90, 90)
````
imo

MistakeNot4892 (approved):

PR #2245, opened 2021-12-02
Integrates world.map_cpu to MC

quardbreak:
## Description of changes

_tick.dm: Replaces some define values in _tick.dm with tgstation numbers.

stoplag(): Replaces sleeps with stoplag() in do_after and do_mob.

map_cpu: Integrates it into tick. Plus there is map cpu indicator in MC panel now, which will... indicate it. Informative.

And colorful numbers.

_tick.dm: I'm not sure if this fine to just update MC/tick defines like that. Most updates which bay got for MC was updates straight from tg, and this is why I think some tg's things compatible with nebula.
But while being skeptical, I think values in this PR might be weird and not performant for our codebase, so I asking @Lohikar to check if values are fine.

stoplag(): Lohikar said what stoplag() is fine and better than sleep() in do_after() and do_mob(). Loaf said, that this doesn't seems fine, so I probably will continue to discuss this with them to find answer what to do with it.

map_cpu: The indicator is useful - it gives info how server loaded while dispatching map information to players. 
Seems tgstation integrated map_cpu into tick, so I think this too should be noted before merging.

## Authorship

tgstation for initial tick define values
Vallat for colored CPU numbers idea

### DIFF
--- a/code/__defines/_tick.dm
+++ b/code/__defines/_tick.dm
@@ -12,11 +12,8 @@
 /// Tick limit while initializing
 #define TICK_LIMIT_MC_INIT_DEFAULT (100 - TICK_BYOND_RESERVE)
 
-
 /// for general usage of tick_usage
 #define TICK_USAGE world.tick_usage
-/// to be used where the result isn't checked
-#define TICK_USAGE_REAL world.tick_usage
 
 /// Returns true if tick_usage is above the limit
 #define TICK_CHECK ( TICK_USAGE > Master.current_ticklimit )
@@ -27,7 +24,7 @@
 //percent_of_tick_used * (ticklag * 100(to convert to ms)) / 100(percent ratio)
 //collapsed to percent_of_tick_used * tick_lag
 #define TICK_DELTA_TO_MS(percent_of_tick_used) ((percent_of_tick_used) * world.tick_lag)
-#define TICK_USAGE_TO_MS(starting_tickusage) (TICK_DELTA_TO_MS(TICK_USAGE_REAL - starting_tickusage))
+#define TICK_USAGE_TO_MS(starting_tickusage) (TICK_DELTA_TO_MS(TICK_USAGE - starting_tickusage))
 
 //time of day but automatically adjusts to the server going into the next day within the same round.
 //for when you need a reliable time number that doesn't depend on byond time.
--- a/code/__defines/_tick.dm
+++ b/code/__defines/_tick.dm
@@ -1,13 +1,27 @@
-#define TICK_LIMIT_RUNNING 80
-#define TICK_LIMIT_TO_RUN 78
+/// Percentage of tick to leave for master controller to run
+#define MAPTICK_MC_MIN_RESERVE 70
+#define MAPTICK_LAST_INTERNAL_TICK_USAGE (world.map_cpu)
+
+/// Tick limit while running normally
+#define TICK_BYOND_RESERVE 2
+#define TICK_LIMIT_RUNNING (max(100 - TICK_BYOND_RESERVE - MAPTICK_LAST_INTERNAL_TICK_USAGE, MAPTICK_MC_MIN_RESERVE))
+/// Tick limit used to resume things in stoplag
+#define TICK_LIMIT_TO_RUN 70
+/// Tick limit for MC while running
 #define TICK_LIMIT_MC 70
-#define TICK_LIMIT_MC_INIT_DEFAULT 98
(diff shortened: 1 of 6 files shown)


code/__defines/_tick.dm
```diff
@@ -1,13 +1,27 @@
...
+/// Tick limit while initializing
+#define TICK_LIMIT_MC_INIT_DEFAULT (100 - TICK_BYOND_RESERVE)
+
 
-#define TICK_USAGE world.tick_usage //for general usage
-#define TICK_USAGE_REAL world.tick_usage    //to be used where the result isn't checked
+/// for general usage of tick_usage
+#define TICK_USAGE world.tick_usage
+/// to be used where the result isn't checked
+#define TICK_USAGE_REAL world.tick_usage
```
MistakeNot4892: What is the purpose of having two defines with identical values?

code/__defines/_tick.dm
```diff
@@ -1,13 +1,27 @@
...
 
+/// Returns true if tick_usage is above the limit
 #define TICK_CHECK ( TICK_USAGE > Master.current_ticklimit )
-#define CHECK_TICK if TICK_CHECK stoplag()
+/// runs stoplag if tick_usage is above the limit
+#define CHECK_TICK ( TICK_CHECK ? stoplag() : 0 )
```
MistakeNot4892, line 21: Why change this to a ternary?

code/_helpers/mobs.dm
```diff
@@ -91,7 +91,7 @@
 	var/starttime = world.time
 	. = 1
 	while (world.time < endtime)
-		sleep(1)
+		stoplag(1)
```
MistakeNot4892, line 94: These changes are unrelated to the PR. Why do it in this PR? Why do it at all?

quardbreak: They're documented in PR description. If Lohi decide that this should be split - I will remove this.

code/__defines/_tick.dm
quardbreak: I don't know. All modern codebases (tg, nebula, bay) have two identical defines. TICK_USAGE_REAL used less and I found only two places where it currently use. TICK_USAGE using more often tho.

code/__defines/_tick.dm
quardbreak: This adds support for if() statements.

code/_helpers/time.dm
```diff
@@ -138,10 +138,11 @@ var/global/rollovercheck_last_timeofday = 0
 //as sleeps aren't cheap and sleeping only to wake up and sleep again is wasteful
 #define DELTA_CALC max(((max(world.tick_usage, world.cpu) / 100) * max(Master.sleep_delta-1,1)), 1)
 
+//returns the number of ticks slept
 /proc/stoplag(initial_delay)
 	// If we're initializing, our tick limit might be over 100 (testing config), but stoplag() penalizes procs that go over.
 	// 	Unfortunately, this penalty slows down init a *lot*. So, we disable it during boot and lobby, when relatively few things should be calling this.
-	if (!Master || Master.current_runlevel < 3)
+	if (!Master || !(Master.current_runlevel & RUNLEVELS_DEFAULT))
```
afterthought2: Doesn't make sense, it's not a bitflag.

code/_helpers/mobs.dm
afterthought2: I don't really get it either; it's one thing to stoplag after tick checks, but just a straight stoplag like this feels expensive and pointless. I don't know enough about the internal scheduling to make fully informed comments, though.

EDIT: from talking to Lohikar on discord, it seems doing this is in fact reasonable, if maybe not very consequential.

code/__defines/_tick.dm
afterthought2: If you want to keep them distinct, clarify the comment as to their expected usage. `/// to be used where the result isn't checked` is nonsensical to me: result of what? It returns a number; obviously you're checking it, not just calling it for the sake of calling it.

quardbreak:
Not sure how address afterthought's request, because:
1) These two defines existed before me and I didn't change their behavior.
2) I don't know how to write a better explanation in which situation they can be picked.

afterthought2:
I think just combine them into one, unless someone has a coherent explanation of which one should be used when (in which case modify the comments). The comment just doesn't make sense as written, to the point of where deleting it improves clarity.

afterthought2 (approved):

PR #2249, opened 2021-12-10
Lighting: Remove goto

Lohikar:
Turns out that DM procs with labels are slower than normal procs because of the utterly cursed way they're implemented.

This implementation does the same thing with a do-while instead, but avoids the indentation increase with a mildly cursed placement of the do-while. Functionally equivalent to the old code.


afterthought2 (approved):
Don't think putting the `do` where you put it makes it easier to read in comparison to indenting (reader will gloss over the do) but don't care enough to fight over it.

PR #2251, opened 2021-12-11
keybind and door fixes

### LINKED ISSUE #2138: Move-upwards doesn't work when bound to keyboard.
#### Description of issue
Move-upwards key, when bound to keyboard via keybinds, does nothing. It works when used directly from the verb.

 I want to be able to move freely in multi-Z.
Instead: Mobilityn't

#### Steps to reproduce

press . , it doesn't do anythink.

keybind, etc etc.

Probably since the keybind system.

Client Version: 514
Server Revision: Revision Unknown
Game ID: cek-apr0
Current map: ISEO Endeavour

However I was able to reproduce this on Nebula.

#### Issue bingo

### LINKED ISSUE #2250: Material doors spawn without correct icon state
#### Description of issue

Material doors spawn with empty string in `icon_state`, causing invisibility.
The effect will drop after door open or close.

изображение

Material doors should spawn with `icon_state`, picking string from `icon_base` (which should take string from material during initialization).

#### Steps to reproduce

1. spawn any material door type.
2. the door is wizard now!

/obj/structure/door

Not sure when it started act like that. Discovered today.

514

#### Issue bingo

quardbreak:
## Description of changes
Keybind: the problem was in `down()` proc override, which used wrong path and didn't execute properly. Now keybind use correct type.

Doors: structure initialization calls `update_materials()`, where the proc calls icon update. It happens before `icon_base` variable get value from current material, and the variable by default have null value, so it will set `icon_state` to null, causing icon disappearance. To make door pick correct icon state on spawn, I added `update_icon()` after value set. 

Additional changes: removed resin door icons, because no materials use them and these icons don't match with current directional doors. Moved door material-based lighting to structure level. MATERIAL_DOOR_SOUND_VOLUME converted to variable with same value to able override it.

This pull-request addresses issues below:
- fixes https://github.com/NebulaSS13/Nebula/issues/2138
- fixes https://github.com/NebulaSS13/Nebula/issues/2250

## Authorship
Myself.


code/game/objects/structures/_structure_materials.dm
```diff
@@ -30,6 +30,8 @@
 			if(!keep_health)
 				health += bonus_health
 		health = keep_health ? min(health, maxhealth) : maxhealth
+	if(material?.luminescence)
+		set_light(material.luminescence, 0.5, material.color)
```
afterthought2: Needs an else clause.

code/game/objects/structures/doors/_door.dm
```diff
@@ -41,26 +36,23 @@
 	. = 10
 
 /obj/structure/door/on_update_icon()
-	..()
-	if(density)
-		icon_state = "[icon_base]"
-	else
-		icon_state = "[icon_base]open"
+	. = ..()
+	icon_state = "[icon_base][!density ? "open" : null]"
```
afterthought2: "", not null. Why `. = ..()`?

quardbreak: Afaik most on_update_icons() use `. = ..()`, so I stick to that pattern.

afterthought2: From reading @MistakeNot4892's structure icon update PR this is not true, and as far as I'm aware there is no meaningful return value in this proc.

MistakeNot4892: This is correct, on_update_icon() does not usefully return.

code/game/objects/structures/doors/_door.dm
```diff
@@ -23,44 +24,38 @@
...
+	if(material?.luminescence)
 		set_light(material.luminescence, 0.5, material.color)
 
-	if(material.opacity < 0.5)
-		alpha = 180
-
 /obj/structure/door/Destroy()
 	update_nearby_tiles()
 	QDEL_NULL(lock)
+	set_light(0)
```
afterthought2: I'm 99% sure this is not needed.

afterthought2 (approved):

PR #2253, opened 2021-12-14
floating-messages: rebases onto tgstation runechat

quardbreak:
## Description of changes
Floating messages are gone. Welcome on aboard, Runechat.

This pull-request removes our current SS-rely floating messages and replaces with the tgstation version.

The good part: 
* Uses his own subsystem.
* Uses slightly different place in saycode for runechat speech call (`say()` -> `hear_say()`).
* Still checks languages as floating messages do.
* Can be used not only on mobs, but on atoms.
* If two persons have a same name (changeling moment), they both should use a same color.
* [ ] ~Emote and radio icon append next to a message.~
* [ ] ~Can output emote messages.~

The bad part: 
* Runechat contains more code than floating messages:
	* Runechat should use his own subsystem for performance reasons.
	* Runechat use his own string colorization method.
	* Obviously, this is a port :(

## Authorship
tgstation: @Bobbahbrown (runechat debut on tgstation) and who maintained Runechat.
onyx: @TobyThorne (draft port from paradise station to bay-like codebase).
nebula: @comma (languages support from floating messages), @quardbreak (hybrid port), @Lohikar (wrote new subsystem to replace tgstation's)

## Changelog

:cl:
rscdel: Removed old BeeStation's floating messages.
rscadd: Floating messages has been replaced with modified tgstation's Runechat.
/:cl:


bobbah:
now this is epic

quardbreak:
@bobbahbrown Thanks! I actually feel a bit awkward to hear what this port is epic. But if I add per-mob font support (for synths and creatures), you really can say it's epic!

quardbreak:
Don't know how to implement radio style icon and emotes with current code. 
Think this works ok and ready for review.

code/__defines/lists.dm
```diff
@@ -29,6 +29,10 @@
 #define LAZYISIN(L, I) (L ? (I in L) : FALSE)
 // Null-safe L.Cut()
 #define LAZYCLEARLIST(L) if(L) { L.Cut(); L = null; }
+// This is used to add onto lazy assoc list when the value you're adding is a /list/. This one has extra safety over lazyaddassoc because the value could be null (and thus cant be used to += objects)
+#define LAZYADDASSOCLIST(L, K, V) if(!L) { L = list(); } L[K] += list(V);
+// Removes value V and key K from associative list L
+#define LAZYREMOVEASSOC(L, K, V) if(L) { if(L[K]) { L[K] -= V; if(!length(L[K])) L -= K; } if(!length(L)) L = null; }
```
afterthought2: Confusing macro names; in general associative lists need not have lists as associated values. I don't mind the macro set per se, as these are indeed common and annoying operations, but they should reference lists with list values somehow more directly.

code/_helpers/turfs.dm
```diff
@@ -1,10 +1,15 @@
...
+ * For example, using this on a disk, which is in a bag, on a mob, will return the mob because it's on the turf.
+ * Optional arg 'type' to stop once it reaches a specific type instead of a turf.
+**/
+/proc/get_atom_on_turf(atom/movable/atom_on_turf, stop_type)
+	var/atom/turf_to_check = atom_on_turf
+	while(turf_to_check?.loc && !isturf(turf_to_check.loc))
+		turf_to_check = turf_to_check.loc
+		if(stop_type && istype(turf_to_check, stop_type))
+			break
+	return turf_to_check
```
afterthought2: There's another helper that does this (I think @MistakeNot4892 renamed it recently); in fact they should likely be integrated. Look for `get_holder_of_type` or similar.

code/datums/runechat.dm
```diff
@@ -0,0 +1,306 @@
...
+	owned_by = owner.client
+
+	// Clip message
+	var/maxlen = RUNECHAT_MESSAGE_MAX_LENGTH
+	if (length_char(text) > maxlen)
+		text = copytext_char(text, 1, maxlen + 1) + "..." // BYOND index moment
+
+	// Remove spans in the message from things like the recorder
+	var/static/regex/span_check = new(@"<\/?span[^>]*>", "gi")
+	text = replacetext(text, span_check, "")
```
afterthought2, line 113: If you're going to strip html, you may as well strip all of it using the appropriate helper.

code/datums/runechat.dm
```diff
@@ -0,0 +1,306 @@
...
+
+	// Calculate target color if not already present
+	if (!target.chat_color || target.chat_color_name != target.name)
+		target.chat_color = colorize_string(target.name)
+		target.chat_color_darkened = colorize_string(target.name, 0.85, 0.85)
+		target.chat_color_name = target.name
+
+	// Get rid of any URL schemes that might cause BYOND to automatically wrap something in an anchor tag
+	var/static/regex/url_scheme = new(@"[A-Za-z][A-Za-z0-9+-\.]*:\/\/", "g")
```
afterthought2: Is this really the byond criteria for this? That seems hard to believe.

code/datums/runechat.dm
```diff
@@ -0,0 +1,306 @@
...
+ * * lum_shift - A value between 0 and 1 that will be multiplied against the luminescence
+ */
+/datum/runechat/proc/colorize_string(name, sat_shift = 1, lum_shift = 1)
+	// seed to help randomness
+	var/static/rseed = rand(1,26)
+
+	// get hsl using the selected 6 characters of the md5 hash
+	var/hash = copytext(md5(name + global.game_id), rseed, rseed + 6)
+	var/h = hex2num(copytext(hash, 1, 3)) * (360 / 255)
+	var/s = (hex2num(copytext(hash, 3, 5)) >> 2) * ((CM_COLOR_SAT_MAX - CM_COLOR_SAT_MIN) / 63) + CM_COLOR_SAT_MIN
```
afterthought2: Not sure what's going on here; `>>`?

code/datums/runechat.dm
```diff
@@ -0,0 +1,306 @@
...
+	var/l = (hex2num(copytext(hash, 5, 7)) >> 2) * ((CM_COLOR_LUM_MAX - CM_COLOR_LUM_MIN) / 63) + CM_COLOR_LUM_MIN
+
+	// adjust for shifts
+	s *= clamp(sat_shift, 0, 1)
+	l *= clamp(lum_shift, 0, 1)
+
+	// convert to rgb
```
afterthought2: There's an unimplemented helper for this; move this stuff there.

code/controllers/subsystems/runechat.dm
```diff
@@ -0,0 +1,232 @@
+/// Controls how many buckets should be kept, each representing a tick. (30 seconds worth)
+#define BUCKET_LEN (world.fps * 1 * 30)
```
afterthought2: Please don't reuse macro names; it's super easy to make mistakes doing so.

code/datums/runechat.dm
```diff
@@ -0,0 +1,306 @@
...
+ */
+/datum/runechat/New(text, atom/target, mob/owner, decl/language/language, list/extra_classes = list(), lifespan = RUNECHAT_MESSAGE_LIFESPAN)
+	. = ..()
+	if (!istype(target))
+		CRASH("Invalid target given for runechat")
+	if(QDELETED(owner) || !istype(owner) || !owner.client)
+		PRINT_STACK_TRACE("runechat datum created with [isnull(owner) ? "null" : "invalid"] mob owner")
+		qdel(src)
+		return
+	INVOKE_ASYNC(src, .proc/generate_image, text, target, owner, language, extra_classes, lifespan)
```
afterthought2: Why? Just make the called thing waitfor false.

afterthought2 (commented):
Really don't love the cloned subsystem. Try thinking of a lighter-weight solution.

code/_helpers/turfs.dm
quardbreak: Do it return root holder, or just atoms loc? @afterthought2

code/datums/runechat.dm
quardbreak: Byond really can make links in maptext, in fact, you can click on them

afterthought2: I know it can make links, I just don't buy that it makes links from exactly the results of your regex here.

code/_helpers/turfs.dm
afterthought2: The correct helper is `get_recursive_loc_of_type`.

code/datums/runechat.dm
quardbreak: anything that goes before link in ex. `http://`example.com

code/__defines/lists.dm
quardbreak: These macro straight from tg, I have nothing to say about why are they named like this. I would not port them if they won't been used here.

code/datums/runechat.dm
quardbreak: Personally I wish to avoid strip html proc because I'm not sure if it will work fine with unicode texts.

code/datums/runechat.dm
quardbreak: What helper?

code/datums/runechat.dm
quardbreak: Idk any alternative here.

quardbreak:
Awaiting @Lohikar to review new subsystem. Currently messages fade out way too early.

code/datums/runechat.dm
afterthought2: At a bare minimum there's a macro wrapper for this you should use. Should also comment on what's going on here (presumably some number is being rescaled to be in some range).

code/datums/runechat.dm
afterthought2: `hsl2rgb`

code/modules/client/preference_setup/global/preferences.dm
```diff
@@ -248,9 +248,9 @@ var/global/list/_client_preferences_by_type
 	key = "EXAMINE_MESSAGES"
 	options = list(PREF_SHOW, PREF_HIDE)
 
-/datum/client_preference/floating_messages
-	description = "Floating chat messages"
-	key = "FLOATING_CHAT"
+/datum/client_preference/runechat
+	description = "Show Runechat (Above-Head-Speech)"
+	key = "RUNECHAT_SHOW"
```
MistakeNot4892: Don't really like changing this preference, it means people's old prefs will be invalidated and the description is silly, there's no reason to call it Runechat in the pref if you then have to explain it anyway. Keep it as 'show floating speech' and keep the key imo.

code/controllers/subsystems/runechat.dm
```diff
@@ -0,0 +1,3 @@
+TIMER_SUBSYSTEM_DEF(runechat)
+	name = "Runechat"
```
MistakeNot4892: I know I'm harping on about this, but I feel like calling it 'floating chat' or something is more immediately obvious what it's for. As mentioned on Discord, calling it runechat just because runescape was the initial inspiration isn't particularly helpful from a maintenance and understanding perspective.

code/datums/runechat.dm
MistakeNot4892: Why would unicode texts be stripped by html? Html stripping usually uses `<>`.

MistakeNot4892 (dismissed):
Willing to give this a shot.

Lohikar (changes requested):
Still a no on the timer subsystem.

code/__defines/MC.dm
```diff
@@ -137,4 +137,12 @@ if(Datum.is_processing) {\
 		processing = SS##X.processing; \
 	}\
 }\
+/datum/controller/subsystem/processing/##X/fire() {..() /*just so it shows up on the profiler*/}\
 /datum/controller/subsystem/processing/##X
+
+#define TIMER_SUBSYSTEM_DEF(X) var/global/datum/controller/subsystem/timer/##X/SS##X;\
```
MistakeNot4892: Believe this is unused with the subsystem rolled back.

code/datums/runechat.dm
MistakeNot4892: Looking at `strip_html_properly()` I cannot see why it wouldn't be appropriate here. It just removes anything between `<>`

code/__defines/lists.dm
MistakeNot4892: Suggested in Discord convo: `LAZYASSOCLISTADD` and `LAZYASSOCLISTREMOVE`

MistakeNot4892:
Is this a draft or is it ready for review?

quardbreak:
This will be ready for review only when the current revision of runechat subsystem will be fixed.

afterthought2 (changes requested):
In its current form, I have concerns about this system's performance and robustness under load. I think it's not viable with this backend approach. I would be willing to reconsider either with a full backend rewrite to cut down on datum churn, or if it comes with some form of automated stress-testing. Consider this blocking.

PR #2257, opened 2021-12-15
sstimer: fix potential bucket corruption in timer reset_buckets

quardbreak:
## Description of changes
ports semoro's fix related to potential SStimer bucket corruption
> Hi! The essence of the fix is that earlier timers with a built linkedlist could get into the second queue, which could cause an incorrect state
It works super stupidly, resets the state to the original correct one

this should fix potential error

## Authorship
semoro (https://github.com/ss220-space/Paradise/pull/511)

### DIFF
--- a/code/controllers/subsystems/timer.dm
+++ b/code/controllers/subsystems/timer.dm
@@ -259,6 +259,12 @@ SUBSYSTEM_DEF(timer)
 	// Add all timed events from the secondary queue as well
 	alltimers += second_queue
 
+	for (var/datum/timedevent/t in alltimers)
+		t.bucket_joined = FALSE
+		t.bucket_pos = -1
+		t.prev = null
+		t.next = null
+
 	// If there are no timers being tracked by the subsystem,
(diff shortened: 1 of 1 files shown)


code/controllers/subsystems/timer.dm
```diff
@@ -259,6 +259,12 @@ SUBSYSTEM_DEF(timer)
 	// Add all timed events from the secondary queue as well
 	alltimers += second_queue
 
+	for (var/datum/timedevent/t in alltimers)
+		t.bucket_joined = FALSE
+		t.bucket_pos = -1
+		t.prev = null
+		t.next = null
```
afterthought2, line 266: Don't really get why we do this here; unless I'm misreading the logic all of this is set again below except on deleted or irrelevant timers.

afterthought2 (approved):

afterthought2:
We're merging this because it's unlikely to make the thing worse. I'm skeptical of whether this is a fix, and it's clearly not fixing underlying logic issues if they cause this code to matter.

PR #2262, opened 2021-12-21
Network access rework

NataKilar:
## Description of changes
* Reworks network access, supplanting the former grant based system tied to crew records with one based off user account membership in groups.
* Replaces e-mail accounts with more generic user accounts which are now tied to the network access system.
* Adds distinct access requirements to computer files for reading, writing, and modifying access requirements which can be tied either to specific user accounts, or membership in groups.
* Removes the program access system previously on the network ACL.

**User Accounts**
E-mail accounts have been replaced by user accounts tied to a particular network. In terms of e-mail, they function almost identically. Rather than logging into the e-mail account through the client program, accounts can be logged into from the main GOOSE screen. GOOSE programs will always be aware of access tied to particular user accounts as long as the account is logged in on the OS.

User accounts are stored on account mainframes, and accounts can be created or recovered from a backup using the account management program. The required access to create accounts is based off either the access on the account mainframe itself, or optionally through parent group authorization, described below.

**Groups**
Grants have been replaced with user account membership in groups. On the backend, membership in a group is represented by an access string tied to the user account in the form "[Group Name].[Network ID]", similar to how grants acted before. Groups are created using the Network ACL, but can be managed remotely using the new account management program.

Groups are split into two types: parent groups and their child groups. Membership in a child group grants access to any object which requires access to the parent group. In addition there are options on the network ACL which allows for members of a parent group to create new accounts and manage account membership in the parent group's respective child groups. Essentially, the distinction allows for sub management along department lines etc. For example, a chief engineer may be a member in both the Engineering parent group and its AtmosTechs child group, while a technician may only be a member of the AtmosTechs group. Both would be able to access an airlock with its access restricted to "Engineering", but only the chief engineer would be permitted to add or remove the technician's membership in the AtmosTechs group.

This distinction is explained in-game in the UI of the ACL and Network Lock.

**Computer File Access**
Computer files now have individual access requirements for read/write/access mod. Programs have had their previous access requirements removed and replaced with read/write access where appropriate. In addition to group membership, computer files can have their access requirements restricted to particular users, which uses the access string "[Account Login]@[Network ID]". Currently, computer file access can only be modified using the terminal command 'permmod'. There's some special handling for reports such as crew records, which should ideally be made a bit more flexible in the future.

This PR replaces the previous grant system with a system that is much more intuitive and in line with real-life expectations. User Accounts will allow for interesting expansions of the modular computer system in the future, as will individual computer file access requirements. While this system doesn't fully supplant the current "mapped" access system, it does allow for near identical behavior where desired.

## Authorship
Myself

## Changelog

:cl:
rscadd: Adds user accounts and network groups which allows for custom access systems tied to computer networks
rscadd: Adds access requirements tied to computer files for read/write
rscdel: Removes previous grant based computer network access system
/:cl:


code/game/machinery/_machines_base/stock_parts/network_lock.dm
```diff
@@ -39,18 +42,16 @@
...
+	var/list/resulting_access = list()
+	for(var/group in groups)
+		if(!(group in access_controller.get_all_groups()))
+			groups -= groups // This group doesn't exist anymore - delete it.
+			continue
+		resulting_access |= "[group].[D.network_id]"
+	if(!resulting_access.len)
 		return
-	return list(resulting_grants) // List of lists is an OR type access configuration.
+	return OR_mode ? list(resulting_access) : resulting_access // List of lists is an OR type access configuration.
```
afterthought2: Needs to be or by default to remain compatible with other logic. For the record, I hate that this thing doesn't work with full access format (nested lists), and have hated that from the time of its introduction.

code/game/objects/items/weapons/cards_ids.dm
```diff
@@ -469,53 +472,121 @@ var/global/const/NO_EMAG_ACT = -50
...
+		account_lookup_failed = TRUE
+	else if(check_account.login != login || check_account.password != password) // The most likely case - login or password were changed.
+		account_lookup_failed = TRUE
+	// Check if the account can be located on the network in case it was moved.
+	else if(!(check_account in network.get_accounts()))
+		account_lookup_failed = TRUE
+	
+	if(account_lookup_failed)
+		current_account = null
+		visible_message(SPAN_WARNING("\The [src] emits a short buzz, indicating a network or login error"),null,null,1)
```
afterthought2: Should probably indicate what went wrong.

code/modules/modular_computers/file_system/computer_file.dm
```diff
@@ -34,8 +43,118 @@ var/global/file_uid = 0
...
+			else if(change == "-")
+				if(!LAZYLEN(write_access))
+					return TRUE
+				write_access[1] -= access_key
+				if(!length(write_access[1]))
+					write_access[1] = null
+					write_access = null
+				return TRUE
+			else
+				return FALSE
```
afterthought2: Think about a way of writing this that doesn't duplicate code.

code/modules/modular_computers/file_system/reports/report.dm
```diff
@@ -20,40 +22,24 @@
...
-
-//Strongly recommended to use these procs to check for access. They can take access values (numbers) or lists of values.
-/datum/computer_file/report/proc/verify_access(given_access)
-	return has_access_pattern(access, given_access)
-
-/datum/computer_file/report/proc/verify_access_edit(given_access)
-	if(!verify_access(given_access))
-		return //Need access for access_edit
-	return has_access_pattern(access_edit, given_access)
+			field.set_access(read_access, write_access, override, TRUE)
```
afterthought2: TBH should rewrite this to use the standard list(list())-accepting notation everywhere else. It's been on my todo for ages now.

afterthought2 (commented):
Need to review this more carefully.

code/game/machinery/_machines_base/stock_parts/network_lock.dm
NataKilar: These now work with nested lists as requested.

code/game/objects/items/weapons/cards_ids.dm
NataKilar: Done.

code/modules/modular_computers/file_system/computer_file.dm
NataKilar: Done, as far as I was able. Mod access still needs to have separate logic.

code/modules/modular_computers/file_system/reports/report.dm
NataKilar: Done

code/_helpers/text.dm
```diff
@@ -132,26 +132,32 @@
 
 //Used to strip text of everything but letters and numbers, make letters lowercase, and turn spaces into .'s.
 //Make sure the text hasn't been encoded if using this.
-/proc/sanitize_for_email(text)
+/proc/sanitize_for_account(text)
 	if(!text) return ""
 	var/list/dat = list()
-	var/last_was_space = 1
+	var/last_was_fullstop = 1
```
MistakeNot4892, line 138: What's going on here? Was the previous code lying?

NataKilar: The behavior has changed slightly. Before spaces were turned to full stops, but you could not manually add full stops.  Now you can, so the name of the var has been changed to reflect that the actual character in the output is a full stop,

code/game/objects/items/weapons/cards_ids.dm
```diff
@@ -299,7 +302,7 @@ var/global/const/NO_EMAG_ACT = -50
 	return
 
 /obj/item/card/id/GetAccess()
-	return access
+	return access.Copy()
```
MistakeNot4892, line 305: Might be nice to lazylist `access` and `req_access` at some point.

code/modules/modular_computers/file_system/programs/generic/reports.dm
```diff
@@ -115,16 +115,16 @@
 	if(href_list["save"])
 		if(!selected_report)
 			return 1
-		if(!selected_report.verify_access(get_access(user)))
+		if(!(selected_report.get_file_perms(get_access(user), user) & OS_READ_ACCESS))
```
MistakeNot4892: Should this be `WRITE`? The href is 'save'.

code/game/machinery/_machines_base/stock_parts/network_lock.dm
```diff
@@ -83,15 +98,48 @@
...
-		)))
-	data["grants"] = grants_data
+
+	data["patterns"] = list()
+	var/pattern_index = 0
+	for(var/list/pattern in groups)
+		pattern_index++
+		data["patterns"].Add(list(list(
+			"index" = "[pattern_index]",
+			"groups" = english_list(pattern, "No groups assigned!")
```
comma: should pass and_text="or" here, otherwise result is misleading in UI (groups are OR based but it lists them as AND in text)

comma:
Groups with # in the name cause issues
They're not interactible in access edit programs, and in network locks you can sorta pick them resulting in broken state it seems 
image

code/modules/modular_computers/file_system/programs/generic/reports.dm
NataKilar: Technically no since this currently always saves a copy. However looking at the UI it seems like saving as a copy should just be an option, so I'll try to fix this up a bit while I'm in the area.

afterthought2: If I recall correctly, the issue with modifying in place is that the file can be changed by other things while this UI is active but not updated, so then when you save you get problematic behavior. ntos has no concept of file locks.

Possible this isn't an issue anymore or can be worked around.

code/game/machinery/_machines_base/stock_parts/network_lock.dm
```diff
@@ -107,7 +107,7 @@
 		pattern_index++
 		data["patterns"].Add(list(list(
 			"index" = "[pattern_index]",
-			"groups" = english_list(pattern, "No groups assigned!") 
+			"groups" = english_list(pattern, "No groups assigned!", and_text = "or")
```
comma: need to pass " or " with spaces here

code/modules/modular_computers/file_system/programs/generic/reports.dm
NataKilar: I don't think this should be an issue, since the reports program uses a copy of the report it's editing, and the save function saves it locally. In the future some sort of file lock system is probably a good idea.

comma (approved):

code/modules/modular_computers/file_system/programs/generic/reports.dm
afterthought2: Well, local as in on the local harddrive--but multiple users might still be accessing it at once from there.

PR #2267, opened 2021-12-26
keybind tweaks

quardbreak:
## Description of changes
In a few words:
* Removed carbon category.
* Added mob and living category.
* Variable documentation.
* Re-implemented classic hotkeys which I still don't know how they work but.
* Moved button which resets all keybinds up in control preferences
* None -> Unbound. Remind me to replace it with define later.
* New keybinds:
	* LOOC - Unbound.
	* F5 - Msay
	* ShiftF5 - Asay
	* Tab - Now can be used as key. Currently used for hotkey mode switch because I needed to add more features.
	* Screenshot is Unbound by default.
	* G - Give for humans.
	* Space - turns throw mode on while you hold it.

One day byond become drunk and split `auto` argument for `.screenshot` as separate command `.auto`. Auto one don't ask directory location from you to save a screenshot. Probably it defined by default by byond with F2 or ShiftF2 key. 

Keybinds are still meh but slowly heals self.

## Authorship
me me me


code/modules/client/preference_setup/controls/01_keybindings.dm
```diff
@@ -23,7 +24,7 @@
 		if(length(user_binds[kb.name]))
 			continue // key is unbound and or bound to something
 		var/addedbind = FALSE
-		if(hotkeys)
+		if(client.prefs?.hotkeys)
```
afterthought2: Why?

afterthought2 (commented):
Probably need to test.

code/modules/client/preference_setup/controls/01_keybindings.dm
quardbreak: Idk I running on this place twice and probably mass-replaced this.

MistakeNot4892 (approved):
Let's give it a go.

PR #2268, opened 2021-12-27
ventcrawl: diagonal direction no longer breaks movement

quardbreak:
## Description of changes
Strips diagonal directions on move.

Bugfix. Mobs should not leave pipes because of diagonal move.

## Authorship
https://github.com/ss220-space/Baystation12/pull/95

### DIFF
--- a/code/modules/ventcrawl/ventcrawl_atmospherics.dm
+++ b/code/modules/ventcrawl/ventcrawl_atmospherics.dm
@@ -15,7 +15,10 @@
 /obj/machinery/atmospherics/relaymove(mob/living/user, direction)
 	if(user.loc != src || !(direction & initialize_directions)) //can't go in a way we aren't connecting to
 		return
-	ventcrawl_to(user,findConnecting(direction),direction)
+
+	// Only cardinals allowed.
+	direction = FIRST_DIR(direction)
+	ventcrawl_to(user,findConnecting(direction), direction)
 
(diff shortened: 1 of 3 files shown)


afterthought2:
This doesn't really work: imagine you are on a 4-way manifold with simple pipes going north, south, east, west, and those are the only pipes on the map. From the manifold turf you call move northeast, which does not get stripped (north and east are initialize dirs because there is a pipe east and a pipe north), but there is no pipe to the northeast.

I suggest the simpler solution of just blocking all diagonal moves (if diagonal return) while ventcrawling.

code/modules/ventcrawl/ventcrawl_atmospherics.dm
```diff
@@ -15,7 +15,9 @@
 /obj/machinery/atmospherics/relaymove(mob/living/user, direction)
 	if(user.loc != src || !(direction & initialize_directions)) //can't go in a way we aren't connecting to
 		return
-	ventcrawl_to(user,findConnecting(direction),direction)
+
+	direction &= ~(direction & ~initialize_directions)
```
Vallat: Remark accepted
image
```suggestion
	direction = (direction & NORTH) || (direction & SOUTH) || (direction & WEST) || (direction & EAST)
```

Vallat (changes requested):

afterthought2 (approved):

PR #2270, opened 2021-12-28
remaps example map

quardbreak:
## Description of changes
Example map has been remapped to new map - Test Site (or just debug):
* Based on tgstation's multiz debug map.
* Bigger than before. Map size is 52x52x3.
* More place for testing. Third floor now specified for zmimic/multiz tests.
* New lobby icon for Test Site.
* This can make someone unhappy, but init time for debug map is somewhere around 11.5sec now.

Other changes:
* Added color hexagon icon asset. Unsure why exactly in this PR, but this can be used for button which involves color change.

Example map is good for testing, but it's size don't always allow to test big things. There also lack of proper place to test multiz.
I decided update example map and rename it to debug, so it will represent more about its purpose.

## Authorship
tgstation, me


MistakeNot4892:
Is this ready for review?

maps/example/example_areas.dm
```diff
@@ -33,23 +32,18 @@
 	arrival_sound = null
 	lift_announce_str = null
 
-/area/turbolift/alert_on_fall(var/mob/living/carbon/human/H)
-	if(H.client && SSpersistence.elevator_fall_shifts > 0)
```
MistakeNot4892, line 37: Why remove this?

quardbreak: I don't think this needs to be here permanently on debug map.

MistakeNot4892 (approved):

PR #2271, opened 2021-12-31
New year's eve fixes

quardbreak:
## Description of changes
Some fixes to things I found and able to resolve with my skills:
* removes infinity loop on modular computer (devices) boot
* removes client view jitter after interact with helm consoles (https://github.com/ss220-space/NebulaHestia/issues/88).
* widescreen overmap support.
* fixes occupant overlay transform in sleepers (part of https://github.com/ss220-space/NebulaHestia/issues/24).
* reduces winset calls in fullscreen toggle function (https://github.com/ss220-space/NebulaHestia/issues/9).
* shuffles generate image function for cameras. Didn't fix anything, but I touch it.
* call icon update chain on lay button press (https://github.com/ss220-space/NebulaHestia/issues/36).
* overmap sectors now will pick empty turf, otherwise random one (https://github.com/ss220-space/NebulaHestia/issues/37)
* storage contents offset reset (https://github.com/ss220-space/NebulaHestia/issues/25).
* alt-click shortcut to open storage (https://github.com/ss220-space/NebulaHestia/issues/55).
* discarding UI interface edits reverts to preference version (https://github.com/ss220-space/NebulaHestia/issues/65)
* ~~tweaks fabricator interface for better information view.~~
* nanoui button highlight on mouseover
* human matrix translation offset tweaked to respect matrix for resting humans uses angle to select correct offset
* advanced global callproc work again
* fixes storage inventory contents offset (https://github.com/NebulaSS13/Nebula/issues/2302)

It fixes some things

## Authorship
https://github.com/Baystation12/Baystation12/pull/31825
~~KekMan632 for advising with one part of fabricator ui,~~ other stuff by me.


code/modules/mob/living/simple_animal/friendly/mouse.dm
```diff
@@ -54,10 +54,6 @@
 		else if(prob(5))
 			INVOKE_ASYNC(src, .proc/audible_emote, "snuffles.")
 
-/mob/living/simple_animal/mouse/lay_down()
-	..()
-	icon_state = resting ? "mouse_[body_color]_sleep" : "mouse_[body_color]"
-
```
kelgaras, line 60: Why did you delete this? Just rename the names responsible for the textures
https://github.com/ss220-space/NebulaHestia/pull/72/commits/061291afb61e627018e6e3ad3948e54868a3a9b3

code/modules/client/ui_style.dm
```diff
@@ -52,18 +52,24 @@ var/global/all_tooltip_styles = list(
...
 		prefs.UI_style_color = UI_style_color_new
 		SScharacter_setup.queue_preferences_save(prefs)
 		to_chat(usr, "Your UI settings were saved.")
+	else
+		// Reverting back.
+		for(var/obj/screen/I in icons)
+			if(!(I.name in list(I_HELP, I_HURT, I_DISARM, I_GRAB)))
+				I.icon = all_ui_styles[prefs.UI_style]
+				I.color = prefs.UI_style_color
+				I.alpha = prefs.UI_style_alpha
```
kelgaras: you can create `client/proc` and call it to update ui
```
client/proc/update_icons(var/list/icons, var/alpha = prefs.UI_style_alpha, var/color = prefs.UI_style_color,var/style = all_ui_styles[prefs.UI_style])
	for(var/obj/screen/I in icons)
		if(I.name in list(I_HELP, I_HURT, I_DISARM, I_GRAB)) continue
		I.icon = style
		I.color = color
		I.alpha = alpha

File: code/modules/client/ui_style.dm [hunk 3]
```diff
@@ -52,18 +52,24 @@ var/global/all_tooltip_styles = list(
...
 	for(var/obj/screen/I in icons)
-		if(I.name in list(I_HELP, I_HURT, I_DISARM, I_GRAB)) continue
-		I.icon = ic
-		I.color = UI_style_color_new
-		I.alpha = UI_style_alpha_new
-
+		if(!(I.name in list(I_HELP, I_HURT, I_DISARM, I_GRAB)))
+			I.icon = UI_style_icon_new
+			I.color = UI_style_color_new
+			I.alpha = UI_style_alpha_new
```
kelgaras: call the `client/proc` created below
```
update_icons(icons, UI_style_alpha_new, UI_style_color_new, ic)

File: code/modules/client/ui_style.dm [hunk 4]
```diff
@@ -52,18 +52,24 @@ var/global/all_tooltip_styles = list(
...
 
 	if(alert("Like it? Save changes?",,"Yes", "No") == "Yes")
 		prefs.UI_style = UI_style_new
 		prefs.UI_style_alpha = UI_style_alpha_new
 		prefs.UI_style_color = UI_style_color_new
 		SScharacter_setup.queue_preferences_save(prefs)
 		to_chat(usr, "Your UI settings were saved.")
+	else
+		// Reverting back.
```
kelgaras: after `else`
```
update_icons(icons)

**Review by kelgaras — CHANGES_REQUESTED** [review]

File: code/modules/mob/living/simple_animal/friendly/mouse.dm (hunk 1 above)
Comment (quardbreak) [reply code/modules/mob/living/simple_animal/friendly/mouse.dm:60]: on_udate_icon already does same, no need you duplicate code.

Comment (kelgaras) [reply code/modules/mob/living/simple_animal/friendly/mouse.dm:60]: Without this, the mouse does not change the texture when it lies down, but it does not disappear, but it does not lie down either.
> on_udate_icon already does same, no need you duplicate code.

image

Comment (quardbreak) [reply code/modules/mob/living/simple_animal/friendly/mouse.dm:60]: The override should not even be existed after simple animal icons refactor:
https://github.com/NebulaSS13/Nebula/blob/a7776ca81f3d46b46fe264f0e617771f4617041e/code/modules/mob/living/simple_animal/simple_animal.dm#L122-L123

Comment (kelgaras) [reply code/modules/mob/living/simple_animal/friendly/mouse.dm:60]: my bad

**Review by kelgaras — DISMISSED** [review]

File: code/_helpers/icons.dm [hunk 5]
```diff
@@ -872,53 +872,67 @@ The _flatIcons list is a cache for generated icon files.
...
+	ASSERT(range)
+
+	// - Collecting list of turfs to render -
+
+	var/list/render_turfs = list()
+	for(var/x_offset = 0 to range)
+		for(var/y_offset = 0 to range)
+			var/turf/T = locate(target_x + x_offset, target_y + y_offset, target_z)
+			if(!T)
+				CRASH("proc/generate_image: capture includes non-existan turfs.")
```
afterthought2: I don't think this is reasonable; should handle map edge gracefully.

code/_helpers/icons.dm
```diff
@@ -872,53 +872,67 @@ The _flatIcons list is a cache for generated icon files.
...
+			render_atoms.Add(A)
+
+	// - Performing rendering with collected atoms in list -
+
+	render_atoms = sort_atoms_by_layer(render_atoms)
+	var/icon/capture = icon('icons/effects/96x96.dmi', "")
+	capture.Scale(range * world.icon_size, range * world.icon_size)
+	capture.Blend(COLOR_BLACK, ICON_OVERLAY)
+	for(var/atom/A as anything in render_atoms)
+		var/icon/image = getFlatIcon(A)
```
afterthought2: Don't use image for var names (at least ones which aren't holding images).

code/game/machinery/Sleeper.dm
```diff
@@ -128,16 +129,25 @@
...
 
 /obj/machinery/sleeper/on_update_icon()
-	overlays.Cut()
+	cut_overlays()
 	icon_state = "med_pod"
+
 	if(occupant)
-		var/image/pickle = new
-		pickle.appearance = occupant
+		pickle = new /mutable_appearance(occupant)
```
afterthought2: Why do we cache it on the instance if we are just recreating it here anyway?

code/game/objects/items/weapons/storage/storage.dm
```diff
@@ -38,7 +38,7 @@
 	. = (loc == user && istype(over, /obj/screen)) || ..()
 
 /obj/item/storage/handle_mouse_drop(var/atom/over, var/mob/user)
-	if(canremove && (ishuman(user) || isrobot(user)))
+	if(canremove && (ishuman(user) || isrobot(user) || issmall(user)) && !user.incapacitated(INCAPACITATION_DISRUPTED))
```
afterthought2: Why does being small give us the same rights over inventories as being human? Seems pretty unreasonable.

code/game/objects/items/weapons/storage/storage.dm
```diff
@@ -135,7 +150,7 @@
 			return 0
 
 	//If attempting to lable the storage item, silently fail to allow it
-	if(istype(W, /obj/item/hand_labeler) && user && user.a_intent != I_HELP)
+	if(istype(W, /obj/item/hand_labeler) || istype(W, /obj/item/forensics) && user && user.a_intent != I_HELP)
```
afterthought2: Why? Super unintuitive interaction that should be reconsidered, but if this is something you really want you should (a) prompt the user with mechanical help and (b) have this be an item proc (`blocked_from_storage_on_help_intent` or whatever).

code/modules/admin/buildmode/areas.dm
```diff
@@ -73,7 +73,7 @@ Right Click       - List/Create Area
 	var/used_colors = 0
 	var/list/max_colors = length(distinct_colors)
 	var/list/vision_colors = list()
-	for (var/turf/T in range(get_effective_view(user.client), user))
+	for (var/turf/T in range(user?.client?.view || world.view, user))
```
afterthought2, line 76: Isn't this what this proc does?

code/modules/client/client_procs.dm
```diff
@@ -529,17 +529,18 @@ var/global/const/MAX_VIEW = 41
...
+
+	winset(src, null, {"
+	mainwindow.is-maximized = false;
+	mainwindow.can-resize = [(value == PREF_BASIC) || (value == PREF_FULL) ? "false" : "true"];
+	mainwindow.titlebar = [(value == PREF_BASIC) || (value == PREF_FULL) ? "false" : "true"];
+	mainwindow.menu = [value == PREF_FULL ? "null" : "menu"];
+	mainwindow.statusbar = [value == PREF_FULL ? "false" : "true"];
+	mainwindow.split.pos = [(value == PREF_BASIC) || (value == PREF_FULL) ? "0x0" : "3x0"];
+	"})
+	winset(src, null, "mainwindow.is-maximized = true;")
```
afterthought2, line 543: Why batch all of them other than this one?

code/modules/mob/living/simple_animal/simple_animal.dm
```diff
@@ -119,7 +119,7 @@
 	icon_state = ICON_STATE_WORLD
 	if(stat == DEAD && (mob_icon_state_flags & MOB_ICON_HAS_DEAD_STATE))
 		icon_state += "-dead"
-	else if(!incapacitated() && resting && (mob_icon_state_flags & MOB_ICON_HAS_REST_STATE))
+	else if((resting || incapacitated()) && (mob_icon_state_flags & MOB_ICON_HAS_REST_STATE))
```
afterthought2: Doesn't make sense; clashes with the logic two lines down.

code/modules/nano/interaction/default.dm
```diff
@@ -42,7 +42,7 @@ var/global/datum/topic_state/default/default_topic_state = new
 		return STATUS_CLOSE
 
 	// If an object is in view then we can interact with it
-	if(src_object in view(get_effective_view(client), src))
+	if(src_object in view(client.view, src))
```
afterthought2, line 45: Again, not sure what the point of this is.

code/modules/overmap/sectors.dm
```diff
@@ -96,9 +96,27 @@ var/global/list/known_overmap_sectors
...
 	var/datum/overmap/overmap = global.overmaps_by_name[overmap_id]
-	start_x = start_x || rand(OVERMAP_EDGE, overmap.map_size_x - OVERMAP_EDGE)
-	start_y = start_y || rand(OVERMAP_EDGE, overmap.map_size_y - OVERMAP_EDGE)
-	forceMove(locate(start_x, start_y, overmap.assigned_z))
+	var/location
+
+	if(start_x || start_y)
+		location = locate(
+			start_x || rand(OVERMAP_EDGE, overmap.map_size_x - OVERMAP_EDGE),
+			start_y || rand(OVERMAP_EDGE, overmap.map_size_y - OVERMAP_EDGE),
```
afterthought2: Doesn't make sense.

code/modules/reagents/reagent_containers/glass.dm
```diff
@@ -89,14 +89,14 @@
...
 	if(standard_dispenser_refill(user, target)) //Are they clicking a water tank/some dispenser?
 		return TRUE
-	if(standard_pour_into(user, target)) //Pouring into another beaker?
+	else if(standard_pour_into(user, target)) //Pouring into another beaker?
 		return TRUE
-	if(standard_feed_mob(user, target))
+	else if(standard_feed_mob(user, target))
 		return TRUE
-	if(user.a_intent == I_HURT)
+	else if(user.a_intent == I_HURT)
```
afterthought2: Why? They all return.

code/modules/overmap/sectors.dm
quardbreak: Does. There is reason for it.
If one of them not set, use random one, otherwise it will spawn on null place or 0.

code/modules/nano/interaction/default.dm
quardbreak: When checking view() and using user view, the proc I removed returned non-wide supported view (10, instead 21x15 for example).

code/modules/overmap/sectors.dm
afterthought2: If one of them is not set, use the logic below to select the other one, then do all the setting as previously. The way you broke it up both duplicates code and for some reason uses different selection logic depending on whether one or both is unset.

code/modules/client/client_procs.dm
quardbreak: Because byond issue: proper updating can be done only if you change is-maximized after other parameters have been set.

code/modules/admin/buildmode/areas.dm
quardbreak: This proc returns square view, and it can't return widescreen value.

code/game/objects/items/weapons/storage/storage.dm
quardbreak: I heard about issue when forensics items could be placed when you want to get swabs, or otherwise.

code/game/objects/items/weapons/storage/storage.dm
quardbreak: This done for nymphs and mouses, since they theoretically could look inside boxes.

code/game/machinery/Sleeper.dm
quardbreak: I wanted to cache it, but don't know is this possible to set appearance on existent MA.

code/_helpers/icons.dm
quardbreak: I have no faith to rewrite this proc to make it work better/fix it's work. All what I did is cut garbage for it, leaving bare minimum which work same as before.

code/modules/overmap/sectors.dm
quardbreak: I don't know how to make it like you suggest. This is how I wanted it to be designed:
`[ Custom start --> Random unchecked start  ] |> [ Random checked start --> Random Unchecked start ]`

quardbreak:
More likely I will strip some fixes because I don't care to address feedback for them. I did what I did. If I have no idea how to make it proper and I don't care about it at all - it will be removed and only someone who cares about it more than me could do it better.

code/_helpers/icons.dm
afterthought2: Just drop this line; that's what it did before with the error suppression arg (which was always true).

code/game/machinery/Sleeper.dm
afterthought2: Yes, but it seems risky in this context (and likely not much faster; image ops are fast enough that none of this matters here).

code/game/objects/items/weapons/storage/storage.dm
afterthought2: It's done for literally all small mobs and it's unacceptable.

EDIT: I'm open to having this be `isliving` or something as long as it doesn't break stuff, just the small check is terrible.

code/modules/admin/buildmode/areas.dm
afterthought2: OK.

code/game/machinery/Sleeper.dm
quardbreak: You can't manipulate image's matrix, hey just don't have it. This is why I used MA -- I need to get mob's appearance and reset transform and apply other like scaling modifiers, but without lay transform.

quardbreak: Also the var is exist because it allows to change MA in potential situations (jitter, etc).

code/modules/overmap/sectors.dm
afterthought2: A low-effort approach would be to turn this branch into `if(x && y)` and drop these `||`s, and that would be fine with me. Setting just one and not the other in earlier code makes no sense anyway, and seems fine to override.

code/game/machinery/Sleeper.dm
afterthought2: Yeah, I think it's fine to use an MA, just don't cache it. Note that you aren't going to be able to use the MA to modify the overlay: SSoverlays will extract the appearance anyway.

code/_helpers/icons.dm
MistakeNot4892: Please address this change.

code/game/objects/items/weapons/storage/storage.dm
```diff
@@ -38,7 +38,7 @@
 	. = (loc == user && istype(over, /obj/screen)) || ..()
 
 /obj/item/storage/handle_mouse_drop(var/atom/over, var/mob/user)
-	if(canremove && (ishuman(user) || isrobot(user)))
+	if(canremove && (ishuman(user) || isrobot(user) || isanimal(user)) && !user.incapacitated(INCAPACITATION_DISRUPTED))
```
MistakeNot4892, line 41: Probably needs `()` around that `&&`. What is this meant to solve?

code/game/objects/items/weapons/storage/storage_ui/default.dm
```diff
@@ -177,21 +177,21 @@
...
 		O.hud_layerise()
 		cx++
-		if (cx > (4+cols))
-			cx = 4
+		if (cx > (3 + cols))
+			cx = 3
 			cy--
 
-	closer.screen_loc = "LEFT+[4+cols+1]:16,BOTTOM+2:16"
+	closer.screen_loc = "LEFT+[3 + cols + 1]:16,BOTTOM+1.7:16"
```
MistakeNot4892: Considering making these constants or defines somewhere instead of magic numbers?

code/modules/overmap/ships/computers/ship.dm
```diff
@@ -60,7 +60,11 @@ somewhere on that shuttle. Subtypes of these can be then used to perform ship ov
 	if(linked)
 		user.reset_view(linked)
 	if(user.client)
-		user.client.view = world.view + extra_view
+		if(istext(user.client.view))
+			var/list/retrieved_view = splittext(user.client.view, "x")
+			user.client.view = "[text2num(retrieved_view[1]) + extra_view]x[text2num(retrieved_view[2]) + extra_view]"
```
MistakeNot4892, line 65: Probably out of scope since this is a fix for existing code, but extra view should be handled by the client in OnResize() or something. If they user resizes their window while looking through this console they'll lose the extra viewsize.

code/game/objects/items/weapons/storage/storage_ui/default.dm
quardbreak: Closed define in the file is enough?

code/game/objects/items/weapons/storage/storage.dm
quardbreak: I added animals there because passing living will include AI and other mobs I didn't expect.

quardbreak: Actually the idea I wanted is allow Nymphs and mouses to look inside bags.

code/game/objects/items/weapons/storage/storage_ui/default.dm
MistakeNot4892: Yep so long as it's not repeated magic numbers.

MistakeNot4892 (approved):
Not sure how I feel about the get_effective_view removals vs fixing get_effective_view but works for now.

PR #2273, opened 2022-01-03
fix: grabbed dead mobs now able to move after the grabber

Gaxeer:
## Description of changes

It's was impossible to make dead mob move after the grabber. Now it is.

fix the bug with grab-moving dead mob

## Authorship

@Gaxeer

## Changelog

:cl:
bugfix: fix the bug with grab-moving dead mob
/:cl:

### DIFF
--- a/code/datums/movement/mob.dm
+++ b/code/datums/movement/mob.dm
@@ -13,12 +13,15 @@
 		control_object.set_dir(direction)
 
 // Death handling
-/datum/movement_handler/mob/death/DoMove()
-	if(mob.stat != DEAD)
+/datum/movement_handler/mob/death/DoMove(var/direction, var/mob/mover)
+	if(mob != mover || mob.stat != DEAD)
 		return
+
 	. = MOVEMENT_HANDLED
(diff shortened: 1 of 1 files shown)


code/datums/movement/mob.dm
```diff
@@ -14,11 +14,18 @@
 
 // Death handling
 /datum/movement_handler/mob/death/DoMove()
+	if(mob.moved_by_grab)
```
PsiOmegaDelta: The appropriate solution is:
```
/datum/movement_handler/mob/death/DoMove(var/direction, var/mob/mover)
    if(mob != mover || mob.stat != DEAD)
        return
    [...]
```
No other changes should be needed, assuming the move is initiated correctly by the grab.

PsiOmegaDelta (dismissed):

code/datums/movement/mob.dm
Gaxeer: thanks. This what I was looking for

comma (approved):

PR #2276, opened 2022-01-16
Lower GC phase 2 timeout to 30 seconds

MarinaGryphon:
## Description of changes
Experimentally lowers the timeout from 2 minutes (post-TG) to 30 seconds, since Spookerton reported it worked to reduce queue buildup.

Will hopefully reduce GC queue buildup and lag; needs testing.

## Authorship
Spookerton for the suggestion/information.

### DIFF
--- a/code/controllers/subsystems/garbage.dm
+++ b/code/controllers/subsystems/garbage.dm
@@ -17,7 +17,7 @@ SUBSYSTEM_DEF(garbage)
 	runlevels = RUNLEVELS_DEFAULT | RUNLEVEL_LOBBY
 	init_order = SS_INIT_GARBAGE
 
-	var/list/collection_timeout = list(0, 2 MINUTES, 10 SECONDS)	// deciseconds to wait before moving something up in the queue to the next level
+	var/list/collection_timeout = list(0, 30 SECONDS, 10 SECONDS)	// deciseconds to wait before moving something up in the queue to the next level
 
 	//Stat tracking
 	var/delslasttick = 0            // number of del()'s we've done this tick


afterthought2:
Worth testing, but I'm generally averse to changes where I do not understand the mechanism behind them.

Spookerton:
There are two things to note here.

The first is that some long tail refs, mostly client related, are often pushed to the Del queue with lower watch timeouts. That results in occasional slightly longer work hangs. I am of the opinion that a handful of \~150ms hangs across a two hour round is preferable to a 40\~80ms hang for many servers *as a best case* (since this can increase dramatically as the watch queue gets longer and more stuff is pushed to the Del queue) *every two seconds*, though.

The second is that while this often results in empty GC queues and thus almost no time spent on them, it's still possible to over-fill the second queue if some large submap (or etc) is created. When the second queue is *too* full, its length affects performance and GC becomes unable to clear itself, resulting in similar performance (ie, bad) to a 2 minute threshold. From a little observation, this is inevitable after about 40k watch entries at round init on bay's hardware and memory setup, although running GC through the lobby period (as is the case for bay and neb) can help get the length under whatever critical point it actually is.

Notably, nearly every watch entry in that (and the current 2 minute) circumstance is null; it's just not fast enough to clear them anymore.

Fixing that would take a bunch of work reimagining how the queues are structured or accessed, and I haven't gotten around to really looking at it myself yet.

Bay sometimes suffers from the second issue because of dynamic submap picking; some of them are large and involve a load of subsequent GC of random atoms etc. Anyone running smaller maps or less init-time generated stuff should be perfectly fine every time.

I think for the most part it is advantageous to go lower than two minutes. Thirty seconds is where bay is, and what GC was at prior to widely adopted tg changes. One minute might be a nice compromise but I have no live run experience of that threshold.

Late ed: If anything, making all three queue thresholds part of configuration to allow servers to hunt for their own ideal numbers is probably a good idea.

MistakeNot4892:
Cherrypicked out due to Jade noting this is an old account.

PR #2278, opened 2022-01-22
Fix and refactor plant throw_impact

MarinaGryphon:
## Description of changes

Heavily refactors throw_impact and splatter() for plants, to try and reduce spaghetti.

Removes some ancient, kinda-janky code and replaces it with a modern, simpler alternative. The biggest change is that exploding fruits aren't completely ludicrous anymore, but they're rare enough that you could probably remove them entirely and no one would notice.
Also this fixes a runtime due to improper arguments passed to /obj/effect/decal/cleanable/fruit_smudge/Initialize() that made it try to serialise the seed as its age when being made persistent. Yikes!

## Changelog

:cl:
bugfix: Sometimes, thrown fruit smudges would fail to save. That's fixed now.
/:cl:


code/modules/hydroponics/seed.dm
```diff
@@ -199,40 +199,42 @@
 		environment.adjust_gas(/decl/material/gas/carbon_dioxide, -req_CO2_moles, 1)
 		environment.adjust_gas(/decl/material/gas/oxygen, req_CO2_moles, 1)
 
+/datum/seed/proc/make_splat(var/turf/T, var/obj/item/thrown)
+	if(!splat_type || locate(splat_type) in T)
```
afterthought2: Put `()` around any `in` clause in dm, it has weird operator priority.

code/modules/hydroponics/seed.dm
```diff
@@ -199,40 +199,42 @@
...
+		return
+	var/atom/splat = new splat_type(T, src)
+	splat.SetName("[name] [pick("smear","smudge","splatter")]")
```
afterthought2: Is it correct that we are using our name?

code/modules/hydroponics/seed.dm
```diff
@@ -199,40 +199,42 @@
...
+/datum/seed/proc/splatter(var/turf/T, var/obj/item/thrown)
+	make_splat(T, thrown)
+
+	var/datum/reagents/splat_reagents = thrown?.reagents
+	if(!splat_reagents?.maximum_volume) // if thrown doesn't exist or has no reagents, use the seed's default reagents.
+		splat_reagents = new /datum/reagents(INFINITY, global.temp_reagents_holder)
+		var/potency = get_trait(TRAIT_POTENCY)
+		for(var/rid in chems)
+			var/list/reagent_data = chems[rid]
+			if(reagent_data && reagent_data.len)
```
afterthought2: I think this runtimes; reagent data need not be a list.

code/modules/hydroponics/seed.dm
```diff
@@ -199,40 +199,42 @@
...
+				var/rtotal = reagent_data[1]
+				var/list/data = list()
+				if(reagent_data.len > 1 && potency > 0)
+					rtotal += round(potency/reagent_data[2])
```
afterthought2: I don't follow what this code does. Reagent data, if a list, is a key-value mapping where order doesn't matter. The keys are strings, usually.

code/modules/hydroponics/seed.dm
```diff
@@ -199,40 +199,42 @@
...
+				if(rid == /decl/material/liquid/nutriment)
+					data[seed_name] = max(1,rtotal)
+				splat_reagents.add_reagent(rid,max(1,rtotal),data)
```
afterthought2, line 233: Don't really follow here: do you really want to pass an empty list into the `data` position for any non-nutriment reagent? That's not equivalent to passing `null` here.

code/modules/hydroponics/seed_datums.dm
```diff
@@ -390,6 +390,12 @@
 	splat_type = /obj/effect/vine
 	kitchen_tag = "mushroom"
 
+/datum/seed/mushroom/make_splat(var/turf/T, var/obj/item/thrown)
+	if(!splat_type || locate(splat_type) in T)
```
afterthought2: `()` around `in`.

code/modules/hydroponics/seed.dm
MarinaGryphon: This is old code, but no, it should be the plant name, I think.

code/modules/hydroponics/seed.dm
MarinaGryphon: waaaaaaaaaaaaaaaa yeah okay. lazylen it is, but this is ripped directly from /obj/item/chems/food/grown, so it'll have to be fixed there too

code/modules/hydroponics/seed.dm
MarinaGryphon: this is also ripped directly from /obj/item/chems/food/grown. i think. that said yeah i think this was supposed to be `chems`, where `chems[i][1]` -> base amount, `chems[i][2]` -> potency multiplier

code/modules/hydroponics/seed.dm
MarinaGryphon: this is also exactly what /obj/item/chems/food/grown does; i'll take a look at that to see if i missed something

code/modules/hydroponics/seed.dm
MarinaGryphon: It IS chems. reagent_data is just a badly named var.

code/modules/hydroponics/seed.dm
MarinaGryphon: Nevermind. `chems = list(/decl/material/liquid/capsaicin = list(3,5), /decl/material/liquid/nutriment = list(1,25))`; it is always at least a list of len 1, and at most a list of len 2. If it's improperly set, it SHOULD runtime and/or fail a unit test.

afterthought2 (approved):

PR #2284, opened 2022-02-05
Fix character setup + change_species

MarinaGryphon:
## Description of changes
Replaces a number of uses of set_species with change_species, and moves like, one thing from change_species to set_species.
Fixes a bunch of issues with character setup: HearthOfHestia/Nebula#189, as well as the fact that it just flat-out doesn't render properly at all following #2184.

Character setup ought to work, and the proper helper functions ought to be used.

## Authorship

Me.


code/modules/species/species.dm
```diff
@@ -458,6 +457,10 @@ var/global/const/DEFAULT_SPECIES_HEALTH = 200
 
 //Checks if an existing limbs is the specie's default
 /decl/species/proc/is_default_limb(var/obj/item/organ/external/E)
+	if((species_flags & SPECIES_FLAG_CRYSTALLINE) != BP_IS_CRYSTAL(E))
+		return FALSE
+	if((species_flags & SPECIES_FLAG_SYNTHETIC) != BP_IS_PROSTHETIC(E))
```
afterthought2: You can't check for equality safely like this; they are bitflags and might differ. Add `!`s or use logic operators.

code/modules/client/preferences.dm
```diff
@@ -340,7 +340,7 @@ var/global/list/time_prefs_fixed = list()
 	// Sanitizing rather than saving as someone might still be editing when copy_to occurs.
 	player_setup.sanitize_setup()
 	character.personal_aspects = list()
-	character.set_species(species)
+	character.change_species(species)
```
PsyCommando, line 343: Good catch!

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -695,6 +695,7 @@
 	if(!istype(move_intent))
 		set_next_usable_move_intent()
 	update_emotes()
+	species.update_appearence_descriptors(src)
```
PsyCommando: Don't call this here. Its called later in apply_species_appearance() during normal setup().
Otherwise this causes the mob to do a full deep update twice during normal init and etc. I'd suggest calling apply_species_appearance() lower down the call chain if you manually have to call set_species.

code/modules/mob/living/carbon/human/human_damage.dm
```diff
@@ -322,6 +322,7 @@ In most cases it makes more sense to use apply_damage() instead! And make sure t
 This function restores all organs.
 */
 /mob/living/carbon/human/restore_all_organs(var/ignore_prosthetic_prefs)
+	species?.create_missing_organs(src)
```
PsyCommando, line 325: Hmm, I think this might not work as expected depending on the value of 'ignore_prosthetic_prefs'

code/modules/organs/external/_external.dm
```diff
@@ -396,13 +396,13 @@
 		//
 		for(var/obj/item/organ/internal/I in owner.get_internal_organs())
 			if(I.parent_organ == organ_tag)
-				LAZYDISTINCTADD(I, internal_organs)
+				LAZYDISTINCTADD(internal_organs, I)
```
PsyCommando, line 399: Well, shit.. That's a good catch! I can't believe that slipped in ^^;

PsyCommando (changes requested):
Looks great to me! Besides one or two proc calls I listed. Otherwise, you've fixed a lot of things I didn't catch in my Pr!

code/modules/mob/living/carbon/human/human.dm
MarinaGryphon: Yeah, turns out it wasn't necessary after changing `mannequin.set_species` for species preview icons to `mannequin.change_species`! I'll remove this line.

afterthought2 (approved):

code/modules/mob/living/carbon/human/human_damage.dm
PsyCommando: Just leaving this for the record, but putting this here can cause infinite recursion in some conditions.. Pretty sure.

PR #2264, opened 2021-12-24
window: removes border check flag from fulltile types

quardbreak:
## Description of changes
Removes ATOM_FLAG_CHECKS_BORDER from fulltile windows.

## Why and what will this PR improve
* fixes https://github.com/NebulaSS13/Nebula/issues/2211

Didn't test, so can't 100% be sure that this fix the issue.

## Authorship
Myaself

PR #2287, opened 2022-02-11
Couple small fixes

comma:
## Description of changes

Fixes two bugs

Fixes good

## Authorship

Chairs is me, windows is #2264 but moved to set_dir and tested (it worked)

## Changelog

:cl:
bugfix: Cannot phase onto low walls through solid windows anymore.
bugfix: Shuttle chairs now display their bars in raised state too
/:cl:


code/game/objects/structures/window.dm
```diff
@@ -333,6 +333,13 @@
 	set_dir(turn(dir, 90))
 	update_nearby_tiles(need_rebuild=1)
 
+/obj/structure/window/set_dir(ndir)
+	..()
```
afterthought2: `. = ..()`; it returns true if dir was changed, I believe

comma: Fixed

afterthought2 (approved):

PR #2288, opened 2022-02-11
Fixes #2004

### LINKED ISSUE #2004: Some devices do not reconnect to their network after a power outage unless reconnected manually.
#### Description of issue
Reporting this secondhand so will be a bit vague.
Some devices (specifically the software server in this case) will lose connection to the network if the network suffers a power outage, and will not reconnect unless you go into their network settings and forcibly refresh them.
This was also observed when testing the network comms branch (#733).

Things should drop their network properly on a power outage and reconnect next time they need their network as happens with a newly spawned device.

#### Steps to reproduce
Cycle the APC on the network mainframe.
Try to use the software vendor.

Nothing springs to mind.

Unclear, probably since the network rewrites.

Current dev branch.

#### Issue bingo

comma:
It seems to be mostly caused by relays never reconnecting and therefore entire decks losing network connectivity forever.
Now if network is shutting down, relays will start trying to reconnect (every 30 seconds), so if router goes up, relays will follow soon.

## Description of changes

Makes network relays attempt reconnection after network goes down, so when router is up again they join the net.

Ideally router would tell them to do that, but I don't think there's a nice way to get all broadcasters in the area, so here we are.

Will make dealing with network power issues slightly less of a nightmare since you don't need to go and manually reboot every relay now.

## Authorship

Me

## Changelog

:cl:
bugfix: Network relays will now attempt reconnect on their own, so if router had a blackout, they'll rejoin on their own without manual reboot.
/:cl:

### DIFF
--- a/code/modules/modular_computers/networking/device_types/relay.dm
+++ b/code/modules/modular_computers/networking/device_types/relay.dm
@@ -1,6 +1,7 @@
 /datum/extension/network_device/broadcaster/relay
 	connection_type = NETWORK_CONNECTION_STRONG_WIRELESS
 	expected_type = /obj/machinery
+	var/reconnect_time
 
 /datum/extension/network_device/broadcaster/relay/get_nearby_networks()
 	if(long_range)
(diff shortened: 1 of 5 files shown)


code/modules/modular_computers/networking/device_types/relay.dm
```diff
@@ -21,7 +22,22 @@
...
+	. = ..()
+	// Router went down, we should try to hook back up when it's up
+	if(net_down)
+		reconnect_time = world.time + 30 SECONDS
+		START_PROCESSING(SSprocessing, src)
+
+/datum/extension/network_device/broadcaster/relay/Process()
+	if(world.time > reconnect_time)
+		if(connect())
+			return PROCESS_KILL
```
afterthought2, line 39: Does everything work fine if it somehow connected (e.g. manually) between ticks? I.e. will connect return true and do nothing if already connected?

comma: yeah, it checks that device is already in devices and returns true early

afterthought2 (approved):

PR #2289, opened 2022-02-12
Fabricator UI refactor

PsyCommando:
## Description of changes
What was supposed to be just adding an option to fabricators to create organ and prosthetics with the proper specie turned into a UI refactor..

To sum up the changes:
* Refactoring fabricator UI to handle per sub-class options for what a given fab is building. So that organ fabs can actually make organs from DNA and etc.. 
* Refactors the fabricator UI to waste less space and be easier to read. 
* Refactors the UI code for fabricators so sub classes can make their own UI template.
* Added support for shared template files to fabricators, so its very easy to reuse code between fabricator UI templates.

## Changelog

:cl:
tweak: Refactored fabricator UI.
/:cl:


PsyCommando:
Alright, things look good. And all bugs seems to be ironed out now. If you guys want screenshots from the UI changes hit me up on discord. Because I can't figure how to display them in here.

code/modules/fabrication/fabricator_bioprinter.dm
```diff
@@ -9,28 +11,75 @@
...
-	var/datum/dna/loaded_dna_datum
+	var/datum/dna/loaded_dna //DNA for biological organs
+
+/obj/machinery/fabricator/bioprinter/get_nano_template()
+	return "fabricator_bioprinter.tmpl"
+
+/obj/machinery/fabricator/bioprinter/make_order(datum/fabricator_recipe/recipe, multiplier)
+	var/datum/fabricator_build_order/order = ..()
+	//Keep these in the order so changing settings while queueing things up won't screw up older orders in the queue
+	LAZYSET(order.data, "dna", loaded_dna)
```
afterthought2: I think if you are going to have a wrapper datum for this stuff, you should be using a proc on the datum to set `data` here and elsewhere.

code/modules/fabrication/fabricator_bioprinter.dm
```diff
@@ -9,28 +11,75 @@
...
-			var/loaded_dna = REAGENT_DATA(S.reagents, /decl/material/liquid/blood)
-			if(islist(loaded_dna))
-				var/weakref/R = loaded_dna["donor"]
+			var/sample = REAGENT_DATA(S.reagents, /decl/material/liquid/blood)
+			if(islist(sample))
+				var/weakref/R = sample["donor"]
 				var/mob/living/carbon/human/H = R.resolve()
 				if(H && istype(H) && H.species && H.dna)
-					loaded_dna_datum = H.dna && H.dna.Clone()
+					loaded_dna = H.dna && H.dna.Clone()
```
afterthought2: `H.dna &&` is redundant here.

code/modules/fabrication/fabricator_bioprinter.dm
```diff
@@ -9,28 +11,75 @@
...
 	. = ..()
+
+/obj/machinery/fabricator/bioprinter/OnTopic(user, href_list, state)
+	if(href_list["flush_dna"])
+		if(fab_status_flags & FAB_BUSY)
+			state("Can't flush DNA while printing in progress!")
+		else
+			loaded_dna = null
+		. = TOPIC_REFRESH
+	. = ..()
```
afterthought2: By not returning in the previous block, you will never set `TOPIC_REFRESH` and potentially make further calls. Generally it's simplest to return after this kind of block, unless sending multiple commands at once is necessary on the front end.

code/modules/fabrication/fabricator_bioprinter.dm
```diff
@@ -9,28 +11,75 @@
...
+		"real_name" = loaded_dna.real_name,
+		"UE"        = loaded_dna.unique_enzymes,
+		"species"   = loaded_dna.species,
+		"btype"     = loaded_dna.b_type,
+	) 
+
+/obj/machinery/fabricator/bioprinter/ui_draw_config(mob/user, ui_key)
+	return TRUE //Always draw it for us
+
+/obj/machinery/fabricator/bioprinter/ui_data_config(mob/user, ui_key)
```
afterthought2, line 75: Proc name typo?

code/modules/fabrication/fabricator_books.dm
```diff
@@ -14,8 +14,13 @@
...
 
-/obj/machinery/fabricator/book/do_build(datum/fabricator_recipe/recipe, amount)
-	. = recipe.build(get_turf(src), amount, selected_color)
+/obj/machinery/fabricator/book/make_order(datum/fabricator_recipe/recipe, multiplier)
+	var/datum/fabricator_build_order/order = ..()
+	LAZYSET(order.data, "selected_color", selected_color)
+	return order
+
+/obj/machinery/fabricator/book/do_build(datum/fabricator_build_order/order)
+	. = order.target_recipe.build(get_turf(src), order.multiplier, LAZYACCESS(order.data, "selected_color"))
```
afterthought2: I think at the minimum this should be like `order.build(blah)`, and then `/order/proc/build(blah){target_recipe.build(blah)}`. Likely the superior pattern would be something like `target_recipe.build(order){looks for appropriate metadata on order}`, but may be out of scope.

code/modules/fabrication/fabricator_robotics.dm
```diff
@@ -19,8 +19,36 @@
...
+			O.set_species(S)
 		O.status |= ORGAN_CUT_AWAY
+
+/obj/machinery/fabricator/robotics/OnTopic(user, href_list, state)
+	if(href_list["pick_species"])
+		var/chosen_specie = input(user, "Choose a specie to produce prosthetics for", "Target Species", null) in get_playable_species()
+		if(chosen_specie)
+			picked_prosthetic_species = chosen_specie
+		. = TOPIC_REFRESH
+	. = ..()
```
afterthought2: Similar issue with Topic.

code/modules/fabrication/fabricator_topic.dm
```diff
@@ -39,6 +39,17 @@
...
+	//Tab expanding/collapsing
+	if(href_list["toggle_resources"])
+		ui_expand_resources = !ui_expand_resources
+		. = TOPIC_REFRESH
+	if(href_list["toggle_queue"])
+		ui_expand_queue = !ui_expand_queue
+		. = TOPIC_REFRESH
+	if(href_list["toggle_config"])
+		ui_expand_config = !ui_expand_config
+		. = TOPIC_REFRESH
```
afterthought2, line 52: For what it's worth, this kind of global state is problematic. I know other nanoui stuff does this (I probably wrote some) so I won't be too critical.

code/modules/fabrication/fabricator_ui.dm
```diff
@@ -1,94 +1,211 @@
...
+			multiplier += list(list("label" = "x[mult]", "multiplier" = mult))
+	return multiplier
+
+//
+// UI sections data
+//
+/obj/machinery/fabricator/ui_data(mob/user, ui_key)
+	var/list/data = ..()
+	//Common fab data
+	data += ui_data_status(user, ui_key)
```
afterthought2: There are actually small issues with doing this: these procs return lazylists (or some of them do, at any rate) so they might return `null`, which here then adds a `null` literal to the list; this will then lead to potential parsing errors in the front end. Make sure they are returning an actual, possibly empty, list, not a lazy list.

code/modules/fabrication/fabricator_ui.dm
```diff
@@ -1,94 +1,211 @@
...
-				build_option["multiplier"] = list()
-				for(var/i = 1 to FLOOR(min(R.max_amount, max_sheets)/PRINT_MULTIPLIER_DIVISOR))
-					var/mult = i * PRINT_MULTIPLIER_DIVISOR
-					build_option["multiplier"] += list(list("label" = "x[mult]", "multiplier" = mult))
-			data["build_options"] += list(build_option)
+	return data
+
+/obj/machinery/fabricator/proc/ui_data_resources(mob/user, ui_key)
+	if(!is_functioning())
+		return
```
afterthought2: Do we really need to check this every time? Are these actually ever called individually in a way where the check matters?

afterthought2 (commented):
Will need to actually test this.

code/modules/fabrication/fabricator_bioprinter.dm
PsyCommando: True

code/modules/fabrication/fabricator_bioprinter.dm
PsyCommando: Can you explain what you mean here by wrapper?
Because orders are temporary objects and not really meant to be tempered with. And I'm not sure if its worth it creating datum subobjects for that list?

code/modules/fabrication/fabricator_bioprinter.dm
PsyCommando: True.. Probably should just run the parent proc before instead.. Because technically, href_list can have more than one entry per call.

code/modules/fabrication/fabricator_bioprinter.dm
afterthought2: Well, the core change here is that instead of passing around the recipe + multiplier + optional args, you are now passing around instances of `datum/fabricator_build_order`. My point is that it's better form to not manipulate this datum's internal structure from the outside, as that makes it easy to modify that internal structure in the future. If you really don't care about this and are using this purely to work around lack of multiple returns, then you can ignore this + the comment below about moving build logic.

code/modules/fabrication/fabricator_books.dm
PsyCommando: Yeah, I thought that as well, but adding a proc that just calls another proc just rubs me the wrong way. 
Also, technically, an order is metadata for queued recipes already. So dunno what you mean about looking for metadata on order?

code/modules/fabrication/fabricator_topic.dm
PsyCommando: Yeah, not a fan either. Its just we got limited tools to work with that. Would need cookies or something to store the prefs client side. And my knowledge of JS is kinda limited XD

code/modules/fabrication/fabricator_ui.dm
PsyCommando: Good catch!

code/modules/fabrication/fabricator_ui.dm
PsyCommando: That's a good point. I can probably just remove most of those checks.

code/modules/fabrication/fabricator_bioprinter.dm
PsyCommando: pushed

code/modules/fabrication/fabricator_bioprinter.dm
PsyCommando: pushed the change!

code/modules/fabrication/fabricator_bioprinter.dm
PsyCommando: done

code/modules/fabrication/fabricator_ui.dm
PsyCommando: fixed

code/modules/fabrication/fabricator_ui.dm
PsyCommando: fixed

PsyCommando:
@afterthought2 Any more issues?

afterthought2 (approved):

MistakeNot4892:
Could you elaborate on why you added species setting for the prosthetic limbs? Prosthetics are restricted based on manufacturer settings, having selectable species appears to just be confusing people.

PR #2290, opened 2022-02-12
Fix new linter switch warnings

MarinaGryphon:
## Description of changes

Fixes a number of issues with switch statements that SpacemanDMM picks up now.

Warnings bad.
## Authorship
Me.


code/game/machinery/syndicatebeacon.dm
```diff
@@ -61,6 +61,8 @@
 				src.updateUsrDialog()
 				spawn(rand(50,200)) selfdestruct()
 				return
+			if(2)
+				break
```
afterthought2: Doesn't compile; dm does not use that syntax.

comma: just replace with prob(50) tbh

afterthought2 (approved):

PR #2291, opened 2022-02-13
Resets player view after exiting overmap interface

kelgaras:
## Description of changes

Bug fix

Bug fix is good
## Authorship

Spookerton
## Changelog

:cl:
bugfix: Fix bug where player's viewport offsets after exiting overmap interface
/:cl:

### DIFF
--- a/code/modules/overmap/ships/computers/helm.dm
+++ b/code/modules/overmap/ships/computers/helm.dm
@@ -249,6 +249,9 @@ var/global/list/overmap_helm_computers
 /obj/machinery/computer/ship/helm/unlook(mob/user)
 	. = ..()
 	if (current_operator?.resolve() == user)
+		if (user.client)
+			user.client.pixel_x = 0
+			user.client.pixel_y = 0
 		set_operator(null)
 
 


afterthought2:
Can't do it like that, at least not here. Do it in the mob-level proc (if this is actually an issue on neb; I wasn't aware it was?)

quardbreak:
Screen shaking/jitter issue is common here and your fix don't handle specific view cases. Here is my fix which also unlocks overmap wide-screen: https://github.com/NebulaSS13/Nebula/commit/730ced2414b78975911769ef6a4eef5402796f38

kelgaras:
> Screen shaking/jitter issue is common here and your fix don't handle specific view cases. Here is my fix which also unlocks overmap wide-screen: [730ced2](https://github.com/NebulaSS13/Nebula/commit/730ced2414b78975911769ef6a4eef5402796f38)

If you're going to finish that PR, then I close this one, and I leave it to you.

kelgaras:
> Can't do it like that, at least not here. Do it in the mob-level proc (if this is actually an issue on neb; I wasn't aware it was?)

On heistia already stella nova this error exists, and there is. In any case, I'll try using the qb code

Issue #2295, opened 2022-02-21
I can't reattach lungs after I detached them


kelgaras:
#### Description of issue
After detaching an organ, such as a lung, the game assumes that the character still has that lung and prevents it from being put back in.

#### Difference between expected and actual behavior
If I have separated the lung, I am expected to be able to put it back in later.

#### Steps to reproduce
Detach out the lung
Try to stick it back with a tool

#### Issue bingo

- [x] Issue could be reproduced at least once
- [x] Issue could be reproduced by different players
- [x] Issue could be reproduced in multiple rounds
- [x] Issue happened in a recent (less than 7 days ago) round
- [x] [Couldn't find an existing issue about this](https://github.com/NebulaSS13/Nebula/issues)

PsyCommando:
You'll have to be a bit clearer here. 
Did you remove the lungs from the mob after detaching them and trying to attach them back? 
And what message did it give you? (That would help finding where it breaks in the code)

Because depending on how you did it, it could either be a problem in the surgery code transitions. Or in the organ code.

PsyCommando:
Nevermind that, I found the problem. 
Surgery was bypassing the organ code all along and removing the detached organ straight from the implants list.. Working on a fix.

kelgaras:
> Nevermind that, I found the problem. 
> Surgery was bypassing the organ code all along and removing the detached organ straight from the implants list.. Working on a fix. 

it didn't reattach because it wasn't deleted, I think your fixed it

but if it's something else and you found a problem, everything is fine

PsyCommando:
Organs don't get deleted during surgery. They're stored inside the mob and get dropped out or inserted in when you perform surgery. What was happening is that the organ were being dumped outside the mob, without letting the mob know the organ was actually removed. Because the way detached organs work is really cruddy.
But anyways, should be fixed soonish. Just gotta find a relatively non ugly way to fix it..

PR #2300, opened 2022-03-21
Fix delete_organs() not actually deleting organs

PsyCommando:
## Description of changes
Fix delete_organs() not actually deleting organs.

Deleting is better when you delete the things.

## Authorship
Someone who might be me.

## Changelog

:cl:
bugfix: Organs now properly deleted in some cases where they wouldn't be.
/:cl:


afterthought2 (approved):

kelgaras:
link #2295 to pr, if you can

PsyCommando:
> link #2295 to pr, if you can

I think #2295 is a separate issue. This fixes only completely deleting all organs. 
That re-attaching issue is pretty weird though.

PR #2301, opened 2022-03-21
Fix de-braining not renaming the brain

PsyCommando:
## Description of changes
Renaming the brain during removal from mob now happens before its removed, so it actually happens. Used to be done after it was removed which meant the check done before renaming was always false.

## Changelog

:cl:
bugfix: Brains removed from someone's head now properly gets renamed to whoever the brain owner was.
/:cl:


afterthought2 (approved):

keIgaras:
link #2300 to pr, if you can

PR #2303, opened 2022-03-27
Fix installing organs via surgery, and organ removal

### LINKED ISSUE #2295: I can't reattach lungs after I detached them

PsyCommando:
## Description of changes
The organ wouldn't get cleared from the mob's organ lists properly, so fixed that.
* ~~Instead of slapping in some hardcoded snowflake code for the mmi_holder I've made some procs to override the behavior of organ detach during surgery.~~ 
* Removed the contents of the vital_organs list in the species. Since removing some organs would cause unintended death and brain damage.

EDIT:
* Remade the logic behind removing/adding external organs, since it was causing some serious issues.
* Partially remade the logic behind the internal organs in order to handle surgeries better.

Fixes #2295 

## Changelog

:cl:
bugfix: Fix organ surgery.
/:cl:


PsyCommando:
Okay, after messing around with dismembering and adding and removing organs, I think this is ready for review.

code/modules/mob/living/carbon/carbon_organs.dm
```diff
@@ -61,3 +59,19 @@
...
+	remove_organ(O, FALSE, TRUE)
+	O.forceMove(src)
+	LAZYDISTINCTADD(parent.implants, O.get_detached_organ()) //#FIXME: get_detached_organ() is a placeholder for handling the weird behavior of the mmi_holder
+
+//Called by surgery when placing an organ during organ attach surgery. Happens before attaching
+/mob/living/carbon/surgical_place_organ(var/obj/item/organ/O, var/obj/item/organ/external/parent)
+	LAZYDISTINCTADD(parent.implants, O)
+	O.forceMove(src)
+
+/mob/living/carbon/get_contained_external_atoms()
```
MistakeNot4892: This is redundant, human get_contained_external_atoms already subtracts the external organ list, which stumps should be in.

code/modules/mob/living/carbon/carbon_organs.dm
```diff
@@ -61,3 +59,19 @@
 	//Check if we should die
 	if(species.is_vital_organ(src, O))
 		death()
+
+//Called by surgeries when detaching an organ during organ detach surgery
+/mob/living/carbon/surgical_detach_organ(var/obj/item/organ/O, var/obj/item/organ/external/parent)
```
MistakeNot4892: This is _exactly_ why I was against moving the removal and addition procs onto the mob.

code/modules/mob/living/carbon/carbon_organs.dm
PsyCommando: It doesn't, because I had to put that there in order to prevent the stumps from coming out. I couldn't find another implementation.

code/modules/mob/living/carbon/carbon_organs.dm
PsyCommando: Its more of an helper than anything. Even before the organ changes, it was just a long snippet copy pasted a million times to do that.
Also, I'm not entirely sure what you're really getting to anyways?

code/modules/mob/living/carbon/carbon_organs.dm
PsyCommando: Actually, after thinking about it, I know what's up. Stumps were getting installed on initialize before.. Going to fix that.

PsyCommando: Nah, they still get dumped. Gonna need to mess with the gib proc methinks.

MistakeNot4892:
````
/mob/living/carbon/get_contained_external_atoms()
	. = ..()
	LAZYREMOVE(., get_organs())
````
Current `dev`, `living.dm`.

File: code/modules/mob/living/carbon/carbon_organs.dm (hunk 2 above)
Comment (MistakeNot4892) [reply code/modules/mob/living/carbon/carbon_organs.dm]: Metastatizing new specialized procs on the mob is exactly what I was expecting to happen from your changes to scope and encapsulation of organ behavior. It's going to be more brittle, it's going to harder to override for specific organs, everything I spoke about at length prior to your human init PR being merged.

It's moot, but I felt obligated to say something. You can disregard this comment.

File: code/modules/mob/living/carbon/carbon_organs.dm (hunk 1 above)
Comment (PsyCommando) [reply code/modules/mob/living/carbon/carbon_organs.dm]: > ```
> /mob/living/carbon/get_contained_external_atoms()
> 	. = ..()
> 	LAZYREMOVE(., get_organs())
> ```
> 
> Current `dev`, `living.dm`.

Yeah, realized that earlier when I fixed stumps not getting added to the external_organ list.
But the line does nothing in our particular case. Might be because the organ list are cleared before dump_contents() is called in destroy. So they still get dumped anyways. Could also be the gib proc kicking them out the body on gib. ~~But it shouldn't since stumps cannot be amputated.~~

EDIT: They can be amputated, but only for surgery.. But seems like the check to dismember only looks at whether the amputation flag is on.. Seems like the flag is being used with possibly mutually exclusive states in mind.

File: code/modules/mob/living/carbon/carbon_organs.dm (hunk 2 above)
Comment (PsyCommando) [reply code/modules/mob/living/carbon/carbon_organs.dm]: I'm gonna be honest here. Its hard to be more brittle than it was already. And I actually noticed it was holding up pretty well in that regard, even seeing an improvement.

Also, I did try to integrate more of your approach from your PR, but a lot of it didn't work, or didn't take things into account. It didn't seem like you tested/investigated a lot of it before. (EX: things like mobs not having eyes for example because get_organ would be used in too many cases and override organ zones and return the head when passed BP_EYES) 
It also touched to so many different systems that I didn't feel like spending time to fill in the blanks, and deal with the massive amount of testing myself, since that's was way out of scope for me. And you should keep in mind my changes were never meant to be the be all, end all of organs. I fully expected there would be a need for further refactors as humans are overhauled. If you think you can do it better, by all means be my guest! I'm just doing what I can with what I got.

Also, concerning your comment, I integrated the surgery behavior into add_organ and remove_organ now via a boolean param. I think its less clear and less overridable than having a proc to handle it, but I guess it means less specialized procs. 
Will be pushing once I test everything.

File: code/modules/mob/living/carbon/carbon_organs.dm (hunk 1 above)
Comment (PsyCommando) [reply code/modules/mob/living/carbon/carbon_organs.dm]: I added a is_droppable() proc to decide if an organ should be allowed to exist outside the mob. Otherwise it just gets deleted instead of dropped. Seems to be working well to prevent stumps and the root limb from just getting kicked out. 

Only weird thing I'm noticing is that gib() doesn't seem to run consistently. Sometimes everything gets kicked out like it should, some other times, the lower body organs don't get kicked out for some reasons. Not entirely sure what's happening there, since everything seems 100% fine structure wise. Its like gib is called in parallel with destruction almost.

File: code/modules/organs/external/_external.dm [hunk 3]
```diff
@@ -272,39 +265,52 @@
 			E.update_icon()
 			return
 
+	//Allow removing sub-limbs
+	if(istype(W,/obj/item/circular_saw) && LAZYLEN(children))
+		var/list/removables = get_limbs_recursive(TRUE)
+		if(LAZYLEN(removables))
+			var/obj/item/organ/external/removing = pick(removables)
```
noelle-lavenza: Uh... No? If anything, this should be a radial menu. If a dismembered limb has several sub-parts, either branching or linear, you should be able to choose which to cut off rather than blindly hacking at it.

code/modules/organs/external/_external.dm
```diff
@@ -272,39 +265,52 @@
...
+				compile_icon()
+				update_icon()
+				removing.compile_icon()
+				removing.update_icon()
+				if(user.get_empty_hand_slot())
+					user.put_in_hands(removing)
+				user.visible_message(SPAN_DANGER("<b>[user]</b> cuts off \the [removing] from [src] with [W]!"))
+				return
+			else
+				user.visible_message(SPAN_DANGER("<b>[user]</b> stops trying to cut \the [removing]."))
```
noelle-lavenza: No "starts trying to cut" message was ever sent. That's probably a big oversight; you should be able to see if someone is hacking up a dismembered organ *before* they finish or stop doing it.

code/modules/organs/external/_external.dm
```diff
@@ -325,6 +331,15 @@
 
 	return all_items
 
+/obj/item/organ/external/proc/get_limbs_recursive(var/no_stumps = FALSE)
+	var/list/all_limbs = list()
+	for(var/obj/item/organ/external/child in children)
+		if(no_stumps && child.is_stump())
+			continue
+		all_limbs += child
```
noelle-lavenza, line 358: Are there assurances that children can't be cyclic? It might be good to check if it's already in the list, and only add it and call get_limbs_recursive if it isn't.

noelle-lavenza: (Also, would kind of like to see either separate functions instead of arguments like `no_stumps`, or the ability to pass predicates/procs/callbacks in. A whitelist and blacklist argument would probably suffice, so to replicate `no_stumps` you'd pass a callback to `is_stump`, or the path to is_stump, as the blacklist argument. Kind of a shame there are no inline functions in DM, but I think it's a better design than control coupling.)

code/modules/organs/external/_external.dm
```diff
@@ -788,7 +816,7 @@ Note that amputating the affected organ does in fact remove the infection from t
 		var/dam_type = BRUTE
 		if(W.damage_type == BURN)
 			dam_type = BURN
-		if(owner.can_autoheal(dam_type))
+		if(owner && owner.can_autoheal(dam_type))
```
noelle-lavenza:
```suggestion
		if(owner?.can_autoheal(dam_type))
```

code/modules/organs/external/_external.dm
```diff
@@ -1305,33 +1332,42 @@ Note that amputating the affected organ does in fact remove the infection from t
...
+				if(QDELETED(organ))
+					LAZYREMOVE(internal_organs, organ)
+					continue
+				organ.do_install(null, src, FALSE, update_icon, FALSE) //Forcemove the organ and properly set it up in our internal data
+
+	//Note that we don't need to change our own hierarchy when not removing from a mob
 
 	// Remove parent references
 	if(parent)
 		LAZYREMOVE(parent.children, src)
```
noelle-lavenza, line 1402: I'd suggest maybe moving this to a proc on the parent, so subtypes can control additional references to remove. Something like `/obj/item/organ/external/proc/remove_child(obj/item/organ/to_remove)`?

code/modules/organs/organ.dm
```diff
@@ -538,3 +550,10 @@ var/global/list/ailment_reference_cache = list()
 /obj/item/organ/proc/is_internal()
 	return FALSE
 
+//Used to tell stumps from other organs. Stumps don't behave like regular organs, and require special handling.
+/obj/item/organ/proc/is_stump()
+	return FALSE
+
+//Used to override organ drop behavior, so we don't drop organs that shouldn't be dropped into the world, like stumps and root limbs, or wrappers
+/obj/item/organ/proc/is_dropable()
```
noelle-lavenza: Nitpick, but technically it's "droppable".

code/modules/organs/organ.dm
```diff
@@ -538,3 +550,10 @@ var/global/list/ailment_reference_cache = list()
 /obj/item/organ/proc/is_internal()
 	return FALSE
 
+//Used to tell stumps from other organs. Stumps don't behave like regular organs, and require special handling.
+/obj/item/organ/proc/is_stump()
```
noelle-lavenza, line 554: Is there a reason stumps aren't a subtype that we can do an istype on?

code/modules/species/species.dm
```diff
@@ -479,6 +473,8 @@ var/global/const/DEFAULT_SPECIES_HEALTH = 200
 	return FALSE
 
 /decl/species/proc/is_vital_organ(var/mob/living/carbon/human/H, var/obj/item/organ/O)
+	if(!vital_organs)
```
noelle-lavenza:
```suggestion
	if(!LAZYLEN(vital_organs))
```

code/modules/surgery/limb_reattach.dm
```diff
@@ -152,15 +143,12 @@
...
-	E.status &= ~ORGAN_CUT_AWAY
-	if(E.children)
-		for(var/obj/item/organ/external/C in E.children)
-			C.status &= ~ORGAN_CUT_AWAY
-	target.update_body()
-	target.updatehealth()
-	target.UpdateDamageIcon()
+
+	//This time we call add_organ but we want it to install in a non detached state
+	target.add_organ(E, P, FALSE, TRUE, FALSE)
```
noelle-lavenza, line 151: This will runtime, won't it? `if(existing && O != existing) CRASH("mob/living/carbon/add_organ(): '[O]' tried to overwrite [src]'s existing organ '[existing]' in slot '[O.organ_tag]'!")` Or is it fine because it's technically the same organ?

code/modules/surgery/organs_internal.dm
```diff
@@ -275,11 +276,14 @@
...
 	if(istype(O) && user.unEquip(O, target))
-		LAZYDISTINCTADD(affected.implants, O) //move the organ into the patient. The organ is properly reattached in the next step
-		if(!(O.status & ORGAN_CUT_AWAY))
-			log_debug("[user] ([user.ckey]) replaced organ [O], which didn't have ORGAN_CUT_AWAY set, in [target] ([target.ckey])")
-			O.status |= ORGAN_CUT_AWAY
-		playsound(target.loc, 'sound/effects/squelch1.ogg', 15, 1)
+		//Place the organ but don't attach it yet
+		target.add_organ(O, affected, detached = TRUE)
+
+		if(!BP_IS_PROSTHETIC(affected))
```
noelle-lavenza, line 282: Personally, I'd move this into something like `/obj/item/organ/proc/get_manipulate_sound()` or something, so that it can be used for attaching, detaching, and other surgery interactions that need sounds to differ based on prosthetic status and other bits.

code/unit_tests/organ_tests.dm
```diff
@@ -251,4 +251,141 @@
...
+
+// ==============================================================================
+// Stumps shall not drop
+// ==============================================================================
+/datum/unit_test/stumps_shall_not_drop
+	name = "ORGAN: Stumps Shall Not Drop From a Gibbed Mob or Severed Limbs."
+
+/datum/unit_test/stumps_shall_not_drop/proc/find_stumps()
+	//Look for stumps that aren't deleted
+	var/list/found_stumps = (locate(/obj/item/organ/external/stump) in world)
```
noelle-lavenza: I don't think this works; you'd need to use a for loop if you want a list of stumps. As it stands, the body of the for loop will never run. However, since you check LAZYLEN, LAZYLEN of a single datum is always 0; therefore, the test will always pass even if stumps drop.
```suggestion
```

code/unit_tests/organ_tests.dm
```diff
@@ -251,4 +251,141 @@
...
+	for(var/obj/item/organ/external/stump/O in found_stumps)
```
noelle-lavenza:
```suggestion
	for(var/obj/item/organ/external/stump/O in world)
```

noelle-lavenza (changes requested):
A few complaints/concerns, but bravo for replacing those span strings with the SPAN_ defines; that had really been bothering me.

Additionally, have you tested that braindeath and decapitation are still lethal?

code/modules/organs/external/_external.dm
PsyCommando: Actually, I'm not sure what's going on with that. The do_after call behaves really weirdly and doesn't pop a progressbar like it should and I couldn't figure why. I tried several different ways, and tried to compare it with other calls to it elsewhere and can't figure out why it doesn't behave.. So its either a bug, or I'm doing something stupid I can't see.

code/modules/organs/external/_external.dm
PsyCommando: Well, the behavior is consistent with the removal of internal organs that's right below. 
Also, the radial menu is kind of a bit annoying to use. But, I guess I could change it to that relatively easily.

code/modules/organs/external/_external.dm
PsyCommando: > Are there assurances that children can't be cyclic? It might be good to check if it's already in the list, and only add it and call get_limbs_recursive if it isn't.

Childrens have to be unique in the mob's list and organs list. If there were cyclic references, this proc would be the last of your worries, cause everything would start breaking up. (True story XD )

> (Also, would kind of like to see either separate functions instead of arguments like `no_stumps`, or the ability to pass predicates/procs/callbacks in. A whitelist and blacklist argument would probably suffice, so to replicate `no_stumps` you'd pass a callback to `is_stump`, or the path to is_stump, as the blacklist argument. Kind of a shame there are no inline functions in DM, but I think it's a better design than control coupling.)

I guess that could be a future improvement. But its a bit out of scope right now. This is mainly a bugfix PR XD
Truth be told, a lot of the human code is still really wonky and will need work, so I'm not too much in a hurry to start polishing stuff like that, and having it get changed anyways a few weeks later when someone refactors all this.

code/modules/organs/external/_external.dm
PsyCommando: true

code/modules/organs/external/_external.dm
PsyCommando: That's something I considered. But, its not a very common case to have special behavior when a limb is outside a mob. I don't really have any applicable use cases for those in the codebase rn, so someone would probably just delete the proc since its unused. Limb behavior outside mobs currently is basically just ceasing any processing and just being in a proper hierarchy so icon updates and stuff works.

But, yeah, its a bit out of scope for this bugfix imho. But it would make sense to add that in the future.

code/modules/organs/organ.dm
PsyCommando: That's actually a proc from /obj/item/organ/external that I brought down to /obj/item/organ. Needed it here to avoid doing extra type check + cast since a lot of proc work on /organ in general.

Honestly I'm not sure if doing a type check is faster than using a proc. But I'm assuming if it was there someone probably checked it before?

code/modules/organs/organ.dm
PsyCommando: Nice catch XD

code/modules/surgery/limb_reattach.dm
PsyCommando: I fixed that already. It will only runtime if we try to replace a different organ in the same slot. Not if its the same one.

code/modules/surgery/organs_internal.dm
PsyCommando: True, but that's again out of scope. I was mostly just updating the surgery code to work properly, and I noticed that prosthetics made the squelch sound when installing them just like organic limbs, so I fixed it. 
But yeah, that's a good idea.

code/unit_tests/organ_tests.dm
PsyCommando: You're right! Not sure why I thought it returned a list..

code/unit_tests/organ_tests.dm
PsyCommando: (Had to commit it in two part since apparently I didn't add it to the batch properly.)

noelle-lavenza (approved):
All the things I found before are now fixed, so it seems good to me.

code/modules/organs/external/_external.dm
```diff
@@ -325,6 +331,15 @@
 
 	return all_items
 
+/obj/item/organ/external/proc/get_limbs_recursive(var/no_stumps = FALSE)
+	var/list/all_limbs = list()
+	for(var/obj/item/organ/external/child in children)
+		if(no_stumps && child.is_stump())
+			continue
+		all_limbs += child
+		all_limbs.Add(child.get_limbs_recursive())
```
noelle-lavenza:
```suggestion
		all_limbs.Add(child.get_limbs_recursive(no_stumps))
```

noelle-lavenza (changes requested):
Ackshually, forgot one.

code/modules/organs/external/_external.dm
PsyCommando: Good catch!

noelle-lavenza (dismissed):
*Now* it seems good.

code/modules/mob/living/living_organs.dm
```diff
@@ -20,19 +20,31 @@
...
-/mob/living/proc/remove_organ(var/obj/item/organ/O, var/drop_organ = TRUE, var/detach = TRUE, var/ignore_children = FALSE, var/in_place = FALSE, var/update_icon = TRUE)
-	if(!in_place)
+//Can be called when the organ is detached or attached. 
+/mob/living/proc/remove_organ(var/obj/item/organ/O, var/drop_organ = TRUE, var/detach = FALSE, var/ignore_children = FALSE, var/in_place = FALSE, var/update_icon = TRUE)
+	//Only run effects if we're not already detached, and we're not doing a in-place removal
+	if(!in_place && !(O.status & ORGAN_CUT_AWAY)) //Gotta check the flag here, because of prosthetics handling detached state differently
 		on_lost_organ(O)
 	. = O.do_uninstall(in_place, detach, ignore_children, update_icon)
 
+	if(!in_place && !(O.status & ORGAN_CUT_AWAY)) //Gotta check the flag here, because of prosthetics handling detached state differently
```
MistakeNot4892: Is this check duplicated?

code/modules/organs/external/_external.dm
```diff
@@ -272,39 +265,52 @@
 			E.update_icon()
 			return
 
+	//Allow removing sub-limbs
+	if(istype(W,/obj/item/circular_saw) && LAZYLEN(children))
+		var/list/removables = get_limbs_recursive(TRUE)
+		if(LAZYLEN(removables))
+			var/obj/item/organ/external/removing = pick(removables)
+			if(do_after(user, 3 SECONDS, removing, FALSE))
```
MistakeNot4892: Probably should check if the organ is still attached and present in the mob.

code/modules/organs/external/_external.dm
MistakeNot4892: Radial menu or a modal selector would be preferable if the target doll isn't usable.

code/modules/organs/external/_external.dm
MistakeNot4892: Jade is referring to the absence of a 'Dave starts cutting off thingo' before all of this processing logic fires.

code/modules/organs/external/_external.dm
```diff
@@ -325,6 +331,15 @@
 
 	return all_items
 
+/obj/item/organ/external/proc/get_limbs_recursive(var/no_stumps = FALSE)
+	var/list/all_limbs = list()
+	for(var/obj/item/organ/external/child in children)
+		if(no_stumps && child.is_stump())
+			continue
+		all_limbs += child
+		all_limbs.Add(child.get_limbs_recursive(no_stumps))
```
MistakeNot4892: This is going to insert nulls if nothing is returned from the recursive limb proc.

code/modules/mob/living/living_organs.dm
PsyCommando: It is because its the same condition but it needs to happen after do_uninstall, and its too simple of a check for me to bother to place the result into a var.

code/modules/organs/external/_external.dm
PsyCommando: It inserts nulls if the list passed to Add is empty?

code/modules/organs/external/_external.dm
PsyCommando: That's not something you can do on a mob. Its for loose limbs lying on the ground. Added a check if its in the parent limb instead.

code/modules/organs/external/_external.dm
PsyCommando: Tested it on wrench and it does not insert nulls at all?

code/modules/organs/external/_external.dm
PsyCommando: Alright. Fixed that, pushing it soon.

code/modules/organs/external/_external.dm
PsyCommando: Added a radial menu to that and internal organ removing too.

code/modules/mob/living/living_organs.dm
PsyCommando: Changed the second condition.

PsyCommando:
All changes are in. Just waiting for approval.

code/modules/mob/living/living.dm
```diff
@@ -635,6 +635,10 @@ default behaviour is:
 /mob/living/carbon/get_contained_external_atoms()
 	. = ..()
 	LAZYREMOVE(., get_organs())
+	//Since organs are cleared on destroy, add this separate check here
+	for(var/obj/item/organ/O in .)
+		if(!O.is_droppable())
+			. -= O
```
MistakeNot4892: Needs to be LAZYREMOVE so that `.` is nulled if the list is cleared, to line up with previous implementation.

code/modules/organs/external/_external.dm
```diff
@@ -108,26 +108,17 @@
...
-		LAZYREMOVE(parent.children, src)
-		parent = null
-	QDEL_NULL_LIST(children)
-	QDEL_NULL_LIST(internal_organs)
-
+	//Update the hierarchy BEFORE clearing all the vars and refs
+	. = ..()
+	//Clear all leftover refs
+	splinted = null //Splints got deleted in parent proc
+	parent = null
```
MistakeNot4892, line 115: Unless the logic has been moved somewhere, `parent.children` is not being properly updated.

code/modules/organs/external/_external.dm
```diff
@@ -272,42 +265,74 @@
 			E.update_icon()
 			return
 
+	//Remove sub-limbs
+	if(istype(W,/obj/item/circular_saw) && LAZYLEN(children) && try_saw_off_child(W, user))
```
MistakeNot4892: Please check for `TOOL_SAW` (ideally use the tool procs for the interaction).

code/modules/organs/external/_external.dm
```diff
@@ -272,42 +265,74 @@
...
+//Handles removing child limbs from the detached limb.
+/obj/item/organ/external/proc/try_saw_off_child(var/obj/item/W, var/mob/user)
+	var/list/removables = get_limbs_recursive(TRUE)
+	if(!LAZYLEN(removables))
+		return 
+	var/obj/item/organ/external/removing = show_radial_menu(user, src, removables, radius = 42, require_near = TRUE, use_labels = TRUE, check_locs = list(src))
+	if(!istype(removing))
+		return TRUE
+
+	user.visible_message(SPAN_DANGER("<b>[user]</b> starts cutting off \the [removing] from [src] with \the [W]!"))
```
MistakeNot4892: As noted above, please either manually check for tool quality, or use `do_tool_interaction` a la
````
if(W.do_tool_interaction(TOOL_WELDER, user, src,        \
	  max(5, round((stored_max_health-stored_health) / 5)), \
	  "repairing the damage to", "repairing the damage to", \
	  "You fail to patch the damage to \the [src].",        \
	  fuel_expenditure = 1
))
````

File: code/modules/organs/external/_external.dm [hunk 23]
```diff
@@ -1503,3 +1563,10 @@ Note that amputating the affected organ does in fact remove the infection from t
 
 /obj/item/organ/external/is_internal()
 	return FALSE
+
+/obj/item/organ/external/get_contained_external_atoms()
+	. = ..()
+	//Prevent stumps and things that shouldn't be dropped from getting dumped out
+	for(var/obj/item/organ/O in .)
+		if(!O.is_droppable())
+			. -= O
```
MistakeNot4892, line 1572: Needs to use `LAZYREMOVE`.

code/modules/organs/organ.dm
MistakeNot4892: Typechecks should generally be avoided. Class shouldn't be used as a categorization system.

code/modules/organs/external/standard.dm
```diff
@@ -38,6 +38,10 @@
 		return
 	. = ..()
 
+//Can't drop root limb
+/obj/item/organ/external/chest/is_droppable()
```
MistakeNot4892, line 42: Will this interfere with the amputation mechanic that destroys the human mob when all limbs other than the chest are removed?

code/modules/organs/organ.dm
```diff
@@ -28,16 +28,18 @@
...
 
 /obj/item/organ/Destroy()
 	if(owner)
-		owner.remove_organ(src, FALSE, FALSE, TRUE, TRUE, FALSE) //Tell our parent we're unisntalling silently
+		owner.remove_organ(src, FALSE, FALSE, TRUE, TRUE, FALSE) //Tell our parent we're unisntalling in place
 	else
 		do_uninstall(TRUE, FALSE, FALSE, FALSE) //Don't ignore children here since we might own/contain them
 	owner = null
 	dna = null
+	species = null
```
MistakeNot4892, line 41: Why are these being nulled?

code/modules/organs/external/_external.dm
PsyCommando: /obj/item/organ/Destroy() calls do_uninstall when stored inside a limb. So no need to do that here.

code/modules/organs/external/standard.dm
PsyCommando: That has nothing to do with that. That mechanic is handled in its own conditional that checks if only a single limb is left then gib:

_external.dm, line 1001
```	owner.remove_organ(src, TRUE, FALSE, ignore_children)
	if(istype(last_owner) && !QDELETED(last_owner) && LAZYLEN(last_owner.get_external_organs()) <= 1)
		last_owner.physically_destroyed(FALSE, disintegrate)
```

This is solely to categorize things as droppable or not, so that when we gib the mob, or individual limbs, its not going to drop say the chest or stumps. I made a proc for it so its no hardcoded and doing type checks or expecting things to get deleted in time at the same time the mob is gibbed like it used to, since it was unreliable after testing.

code/modules/organs/organ.dm
PsyCommando: Just because its a reference. Even if species are singletons, I got the reflex to clear any dangling refs.

code/modules/organs/external/_external.dm
PsyCommando: done

code/modules/organs/external/_external.dm
PsyCommando: Done. 
But. kinda uncomfortable to add in something that's barely used anywhere else in a bugfix tbh.
Also the defines for the return values are defined too late to be used in here. So I used literals instead for handling the no message case, and success case.

code/modules/organs/external/_external.dm
PsyCommando: Done

code/modules/mob/living/living.dm
PsyCommando: Fixed

code/modules/organs/organ.dm
MistakeNot4892: It's a bit untidy, and in this case it's not a circular ref so it's unnecessary for GC. Not a huge issue.

MistakeNot4892 (approved):
Not sure why we need unit tests for stumps, but they're harmless at worst.

Nothing else springs out at me as wrong but I am also not signing off on the fixes working as I am (as demonstrated) largely unfamiliar with the new organ code.

PsyCommando:
> Not sure why we need unit tests for stumps, but they're harmless at worst.

Cause people keep breaking them. And now we know when they do. Its been a reoccurring issue since at least 2016 on bay.

> Nothing else springs out at me as wrong but I am also not signing off on the fixes working as I am (as demonstrated) largely unfamiliar with the new organ code.

Did a lot of surgery tests in the last month. They work. And if something is still broken, I'll just make another PR to fix them.

Also, about the changes. Its all essentially very similar to the old system, except additon/removal of organs has been streamlined into general purpose procs instead of being re-implemented by anything using organs. Also now you can make sure installing/removing organs won't call dangerous things on the parent when you really don't want them to.

In a nutshell:

add_organ() calls:
- do_install()
- on_install_effects()

Here's what they do:
* add_organ() do installing stuff on the mob, organ lists etc.. 
* do_install() do installing stuff on the limb/organ itself, its internal state, and the limb hierarchy whether its inside a mob, or outside, without messing with the mob as much as possible. Can be called to attach loose limbs to other loose limbs as well.
* on_install_effects() handles the fluff of organ installation, like interactive elements, messages, things you wouldn't want to happen if you were say swapping 2 organs in place via code that aren't necessary for the organ to work for example. Its a separate proc because we can skip and override those steps at will.

Those 3 procs and their parameters covers most of the use cases for dealing with organs in the entire codebase.
The removal procs are similar. 
Its is a huge plus compared to having to manually dig through organs, figure out what var checks you have to do to safely retrieve the organs and change the proper lists, and in what limbs, every single times you do something slightly different with organs..

comma (approved):

PR #2308, opened 2022-04-23
Faz-World Sync 04/22/2022

noelle-lavenza:
## Description of changes

Ports a number of fixes from downstream.

Various bugfixes from downstream.

## Authorship
Me.

## Changelog

:cl:
bugfix: Fixed prebuilt floor lights starting with the wrong icon/status.
bugfix: Fixed base wood tables and chairs not having materials set.
bugfix: Fixed an improper usage of DEFAULT_WALL_MATERIAL in place of "steel".
bugfix: Made shuttle windows block air
bugfix: Fixed things without hearts bleeding/making blood splatters
bugfix: Fixed "something something on their hands" message
bugfix: Fixed runtimes with species cultural info and appearance being called before organs
bugfix: Fixes MMI installation
bugfix: Fixes certain limbs not being set up properly on spawn
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human_defense.dm
+++ b/code/modules/mob/living/carbon/human/human_defense.dm
@@ -217,6 +217,9 @@ meteor_act
 	if(W.damtype != BRUTE)
 		return
 
+	if(!should_have_organ(BP_HEART))
+		return
+
 	//make non-sharp low-force weapons less likely to be bloodied
 	if(W.sharp || prob(effective_force*4))
 		if(!(W.atom_flags & ATOM_FLAG_NO_BLOOD))
--- a/code/game/machinery/floor_light.dm
+++ b/code/game/machinery/floor_light.dm
@@ -25,6 +25,10 @@ var/global/list/floor_light_cache = list()
 /obj/machinery/floor_light/prebuilt
 	anchored = 1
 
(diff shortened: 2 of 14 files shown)


code/modules/mob/living/carbon/human/human_defense.dm
```diff
@@ -217,6 +217,9 @@ meteor_act
 	if(W.damtype != BRUTE)
 		return
 
+	if(!should_have_organ(BP_HEART))
+		return
```
afterthought2, line 221: Is this a good proxy for bloodiness across commonly used species? I seem to recall mechanical species using blood overlays to hack in oil spills and stuff.

MistakeNot4892: It's a standard check for other species code involving blood, dating back to one of the foundational organ PRs. Should be changed to its own proc most likely but that would be a broader refactor.

afterthought2: fair enough.

afterthought2 (approved):

code/modules/mob/living/carbon/human/human_defense.dm
noelle-lavenza: Yeah, I used it specifically because other similar code did.

PR #2309, opened 2022-04-23
Add extensions for getting radios and IDs/access

noelle-lavenza:
## Description of changes
- Things that provide IDs can now register to their parent mob and be included in `GetIdCard(s)` and `GetAccess`. This is best-suited for things like organs and implants, rather than handheld items.
- Also adds an equivalent extension and method for radios, which allows the potential for robotic species to have a built-in radio organ.
- Cleans up a little bit of radio code in say.dm
- Fixes an issue with the encryption_keys list on headsets.

More extensibility, and it fixes the broken and formerly-janky Ascent ID controller. Also, needed for downstream stuff that does the same as that organ, as well as a similar organ to provide a built-in radio.
Oh, and I did some sliiiight say.dm refactoring too, specifically with the code that gets radios.
Also fixed a runtime with the encryption_keys list, which just outright didn't work sometimes.
## Authorship

Me.
## Changelog

:cl:
bugfix: Fixes the Ascent ID controller organ.
/:cl:


noelle-lavenza:
Depends on #2303.

code/game/jobs/access.dm
```diff
@@ -7,14 +7,23 @@
...
-		. += id.GetAccess()
+		. = id.GetAccess()
 
 /atom/movable/proc/GetIdCard()
-	return null
+	return LAZYACCESS(GetIdCards(), 1)
+
+/atom/movable/proc/GetIdCards()
+	var/datum/extension/access_provider/our_provider
+	if((our_provider = get_extension(src, /datum/extension/access_provider)))
```
MistakeNot4892: Would prefer the following for readability:
````
	var/datum/extension/access_provider/our_provider = get_extension(src, /datum/extension/access_provider)
	if(our_provider)
````

File: code/game/jobs/access.dm [hunk 2]
```diff
@@ -7,14 +7,23 @@
 		return FALSE
 	return check_access_list(M.GetAccess())
 
-/atom/movable/proc/GetAccess()
+/atom/movable/proc/GetAccess(var/union = FALSE)
```
MistakeNot4892, line 10: What is the purpose of changing the default behavior of GetAccess() to not be a union of all available accesses?

code/game/jobs/access.dm
```diff
@@ -224,21 +233,21 @@ var/global/list/priv_region_access
...
-	var/list/id_cards = get_held_items()
-	LAZYDISTINCTADD(id_cards, wear_id)
-	for(var/obj/item/I in id_cards)
-		if(is_type_in_list(I, exceptions))
+	return LAZYACCESS(GetIdCards(exceptions), 1)
+
+/mob/living/carbon/human/GetIdCards(exceptions = null)
+	. = ..()
+	var/list/candidates = get_held_items()
+	LAZYDISTINCTADD(candidates, wear_id)
```
MistakeNot4892: This will add a null if `wear_id` is null.

code/game/jobs/access.dm
```diff
@@ -224,21 +233,21 @@ var/global/list/priv_region_access
...
+	for(var/atom/movable/candidate in candidates)
+		if(!candidate || is_type_in_list(candidate, exceptions))
```
MistakeNot4892, line 240: Won't this prevent multiple access cards being tracked?

code/game/jobs/access.dm
```diff
@@ -224,21 +233,21 @@ var/global/list/priv_region_access
...
+		if(istype(id_card))
+			LAZYDISTINCTADD(., id_card)
 
-/mob/living/carbon/human/GetAccess()
-	. = list()
-	var/list/id_cards = get_held_items()
-	LAZYDISTINCTADD(id_cards, wear_id)
-	for(var/obj/item/I in id_cards)
-		. |= I.GetAccess()
+/mob/living/carbon/human/GetAccess(var/union = TRUE)
```
MistakeNot4892, line 246: Appears to be a completely redundant override?

code/game/objects/items/devices/radio/headset.dm
```diff
@@ -23,9 +23,10 @@
 /obj/item/radio/headset/Initialize()
 	. = ..()
 	internal_channels.Cut()
-	for(var/T in encryption_keys)
+	for(var/i in 1 to LAZYLEN(encryption_keys))
```
MistakeNot4892, line 26: Why does this need to be numerically indexed? Bit confused about the changes.

MistakeNot4892:
Bit confused about the extensions. If I'm reading this correctly nothing uses them except the controller uses the access provider, which in theory would be achieved by having the controller return an access list.

code/game/jobs/access.dm
noelle-lavenza: It isn't a change. It was only ever the first ID card's access. Now, there's the option to make it the union of all accesses on all ID cards.

code/game/jobs/access.dm
noelle-lavenza: Changes union to true by default for humans.

code/game/jobs/access.dm
noelle-lavenza: EDIT: It *was* originally a hardcoded union but only on humantypes. Now, humantypes use the argument and have it default to true.

code/game/objects/items/devices/radio/headset.dm
noelle-lavenza: Need to replace each typepath with an instantiated datum from that typepath. Could remove and readd, but that'd be worse, IMO.

code/game/jobs/access.dm
noelle-lavenza: I don't believe it will? It just skips nulls and excluded types. For what it's worth, I'm not a big fan of the exceptions argument, but existing things use it, so.

code/game/objects/items/devices/radio/headset.dm
noelle-lavenza: Also, before it'd set it to a single instance instead of a list—setting encryption_keys just never functioned before.

noelle-lavenza:
> Bit confused about the extensions. If I'm reading this correctly nothing uses them except the controller uses the access provider, which in theory would be achieved by having the controller return an access list.

It's for, well, extensibility. Ideally some stuff that currently uses bespoke overrides (silicons come to mind, for example) would eventually just have components that register to the extension instead. Could probably do the same with holders.

code/game/jobs/access.dm
MistakeNot4892: Oh I somehow completely misread exceptions as being a list of previously observed types. Disregard.

code/datums/extensions/access_provider.dm
```diff
@@ -0,0 +1,18 @@
...
+	. = ..()
+
+/datum/extension/access_provider/proc/register_id(atom/movable/to_register)
+	LAZYDISTINCTADD(registered_ids, to_register)
+
+/datum/extension/access_provider/proc/unregister_id(atom/movable/to_unregister)
+	LAZYREMOVE(registered_ids, to_unregister)
+
+/datum/extension/access_provider/proc/GetIdCards()
+	for(var/atom/movable/registered_id in registered_ids)
```
MistakeNot4892, line 17: Microopt: could be `as anything`?

code/datums/extensions/radio_provider.dm
```diff
@@ -0,0 +1,29 @@
...
+	for(var/atom/movable/registered_radio in registered_radios)
+		LAZYDISTINCTADD(., registered_radio.GetRadios(message_mode))
+
+/datum/extension/radio_provider/proc/GetRadio(message_mode)
+	return LAZYACCESS(GetRadios(message_mode), 1)
+
+/atom/movable/proc/GetRadios(message_mode)
+	var/datum/extension/radio_provider/radio_provider = get_extension(src, /datum/extension/radio_provider)
+	if(istype(radio_provider))
+		return radio_provider.GetRadios(message_mode)
```
MistakeNot4892, line 26: Isn't this an infinite loop? GetRadios() called on an atom, it calls it on its extension, extension iterates its atom list and calls it again.

MistakeNot4892: GetRadio() and GetRadios() are also never called anywhere - unclear if this could have been tested.

noelle-lavenza: No, that's just wrong.
- The extension holder *does not* register to the extension, and in fact should not.
- GetRadio is called in say.dm. GetRadios is called by GetRadio.
It also *was* definitely tested, on Ministation.

mods/species/ascent/items/id_control.dm
```diff
@@ -39,26 +39,32 @@
 
 /obj/item/organ/internal/controller/do_install(mob/living/carbon/human/target, obj/item/organ/external/affected, in_place, update_icon, detached)
 	. = ..()
-	if(!detached && owner)
-		owner.set_id_info(id_card)
-		owner.add_language(/decl/language/mantid/worldnet)
+	if(detached || !owner)
+		return
+	var/datum/extension/access_provider/owner_access = get_or_create_extension(owner, /datum/extension/access_provider)
```
MistakeNot4892, line 44: Should this be removing itself from the access holder on uninstall?

MistakeNot4892 (approved):

PR #2315, opened 2022-04-27
Make some of the fields init on carbon and human override easier

PsyCommando:
## Description of changes
Prevent carbon and human from initializing bloodstr, reagents, touching and vessel if they're already set beforehand. So that we don't end up doing the work twice if we override it.

This makes overriding initializing those much easier down the line. Extremely handy downstream to avoid doing temporary var swap bs among other things.

### DIFF
--- a/code/modules/mob/living/carbon/carbon.dm
+++ b/code/modules/mob/living/carbon/carbon.dm
@@ -1,7 +1,8 @@
 /mob/living/carbon/Initialize()
 	//setup reagent holders
-	if(!bloodstr && !reagents)
+	if(!bloodstr)
 		bloodstr = new/datum/reagents/metabolism(120, src, CHEM_INJECT)
+	if(!reagents)
 		reagents = bloodstr
 	if(!touching)
 		touching = new/datum/reagents/metabolism(1000, src, CHEM_TOUCH)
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/carbon/carbon.dm
```diff
@@ -1,8 +1,10 @@
 /mob/living/carbon/Initialize()
 	//setup reagent holders
-	bloodstr = new/datum/reagents/metabolism(120, src, CHEM_INJECT)
-	touching = new/datum/reagents/metabolism(1000, src, CHEM_TOUCH)
-	reagents = bloodstr
+	if(!bloodstr && !reagents)
```
MistakeNot4892: Should split this into two `if` blocks to avoid a null reagents or bloodstr due to a partial init.

MistakeNot4892 (dismissed):
See comment.

code/modules/mob/living/carbon/carbon.dm
PsyCommando: Alright. I also modified the destroy proc to qdel bloodstr, and null out reagents. It was the other way around before and it could cause some confusion.

comma (approved):

PR #2316, opened 2022-04-29
Stirling Engines

NataKilar:
## Description of changes
Adds stirling engines to the game, a type of heat engine that generates power from a temperature differential. This is intended to be a backup power source, or a power source usable on shuttle craft and medium sized ships. It can generate up to 100 kW, but actually achieving this for more than a few seconds would be extremely difficult without a large enough cooling setup that using a TEG would make more sense.

The cycle is idealized, with perfect regeneration. The working volume in the stirling engine is a small piston (a tank subtype) which can be charged with gas as with any other tank.

Other miscellaneous changes are that heat exchangers can now connect to fuel pipes, and air tanks have a material and empty versions can be produced in fabricators.

Stirling engines are neat! This also provides a nice middle ground for power generation that works on shuttles and small ships without being as simple as a PACMAN. Getting full use out of these engines requires some additional heat pipe options, but those will be added in a separate PR.

## Authorship
Myself

## Changelog
:cl:
rscadd: Adds stirling engines to the game. These engines produce power from a temperature differential of two gas sources.
/:cl:


comma (dismissed):

code/modules/power/stirling.dm
```diff
@@ -0,0 +1,218 @@
...
+	icon = 'icons/obj/power.dmi'
+	icon_state = "stirling"
+	density = 1
+	anchored = 1
+	layer = STRUCTURE_LAYER
+	use_power = POWER_USE_OFF
+	base_type = /obj/machinery/atmospherics/binary/stirling
+	construct_state = /decl/machine_construction/default/panel_closed
+	uncreated_component_parts = null
+	stat_immune = NOSCREEN | NOINPUT | NOPOWER // Mechanical machine and display.
```
afterthought2, line 16: It's not correct to use `NOPOWER` immune together with `use_power` stuff, I believe: `NOPOWER` immunity will circumvent certain logic. My guess is that if you set your machine to have `0` as power draw (likely default) and take off `NOPOWER` nothing will change (you may need to also toggle `interact_offline` or similar, but probably not).

EDIT: alternatively you might just want to use a separate `on` toggle var, to avoid going though all the power stuff.

code/modules/power/stirling.dm
```diff
@@ -0,0 +1,218 @@
...
+	if(heat_transfer)
+		air1.add_thermal_energy(-heat_transfer)
+		air2.add_thermal_energy(heat_transfer)
+
+	if(!istype(inserted_cylinder))
+		return
+
+	if(use_power != POWER_USE_ACTIVE)
+		update_icon()
+		update_sound()
```
afterthought2: Do we really need to do this? Is it enough to do it once when power is toggled?

code/modules/power/stirling.dm
```diff
@@ -0,0 +1,218 @@
...
+			update_use_power(POWER_USE_ACTIVE)
+		return
+	. = ..()
+
+/obj/machinery/atmospherics/binary/stirling/on_update_icon()
+	cut_overlays()
+	if (stat & (BROKEN))
+		return
+	if (genlev != 0)
+		add_overlay(emissive_overlay('icons/obj/power.dmi', "stirling-op[genlev]"))
```
afterthought2, line 187: As you're doing this every tick, you probably do want to track state and only do these operations if the state has changed. There are a bunch of machines that do this (see e.g. APC), but here it's pretty simple, just cache genlev and if genlev == last_genlev do nothing (with e.g. null or -1 corresponding to BROKEN).

code/modules/power/stirling.dm
```diff
@@ -0,0 +1,218 @@
...
+		add_overlay(emissive_overlay('icons/obj/power.dmi', "stirling-op[genlev]")) 
+
+/obj/machinery/atmospherics/binary/stirling/proc/update_sound()
+	if(!sound_id)
+		sound_id = "[type]_[sequential_id(/obj/machinery/atmospherics/binary/stirling)]"
+	if(use_power == POWER_USE_ACTIVE)
+		var/volume = 10 + 15*genlev
+		if(!sound_token)
+			sound_token = play_looping_sound(src, sound_id, 'sound/machines/engine.ogg', volume = volume)
+		sound_token.SetVolume(volume)
```
afterthought2, line 196: Same here maybe, unless SetVolume is a null op when the volume is the same.

code/modules/power/stirling.dm
```diff
@@ -0,0 +1,218 @@
...
+		update_icon()
+
+/obj/machinery/atmospherics/binary/stirling/dismantle()
+	if(inserted_cylinder)
+		inserted_cylinder.dropInto(get_turf(src))
+	. = ..()
+
+/obj/item/tank/stirling
+	name = "stirling piston cylinder"
+	desc = "A piston cylinder designed for use in a stirling engine. It must be charged with gas before it can be used. Rated for temperatures up to 1000 C"
```
MistakeNot4892: Needs a full stop. Also might be worth embedding `failure_temp` instead of saying flat 1000 C.

code/modules/power/stirling.dm
MistakeNot4892: IMO it's better to clear and regen on call so the function will work if proccalled or from outside contexts where appearance has been manipulated without `genlev` being updated necessarily. Gating the icon update each tick would likely be a better approach.

code/modules/power/stirling.dm
NataKilar: I replaced references to power_use with a seperate ``active`` var.

code/modules/power/stirling.dm
NataKilar: Removed.

code/modules/power/stirling.dm
NataKilar: The second solution is what I went with, sound and icons are only updated during process if the gen level has changed.

code/modules/power/stirling.dm
NataKilar: Added

MistakeNot4892:
Conflicted.

code/modules/power/stirling.dm
```diff
@@ -0,0 +1,222 @@
+#define HEAT_TRANSFER 5000
+#define MAX_FREQUENCY 60
+#define MIN_DELTA_T 5
+
+/obj/machinery/atmospherics/binary/stirling
+	name = "stirling engine"
+	desc = "A mechanical stirling generator. It generates power dependent on the temperature differential between input gas and a radiative heatsink"
+	icon = 'icons/obj/power.dmi'
+	icon_state = "stirling"
+	density = 1
```
MistakeNot4892: Nit: `TRUE` here and below.

code/modules/power/stirling.dm
```diff
@@ -0,0 +1,222 @@
...
+		air2.add_thermal_energy(heat_transfer)
+
+	if(!istype(inserted_cylinder))
+		return
+	if(!active)
+		return
+
+	var/datum/gas_mixture/working_volume = inserted_cylinder.air_contents
+
+	if(stat & (BROKEN) || !air1.total_moles || !air2.total_moles || !working_volume.total_moles)
```
MistakeNot4892: `(stat & BROKEN)`

code/modules/power/stirling.dm
```diff
@@ -0,0 +1,222 @@
...
+		stop_engine()
+		return
+	
+	// A rough approximation of a Stirling cycle with perfect regeneration e.g. Carnot efficiency.
+	var/working_heatcap = working_volume.heat_capacity()
```
MistakeNot4892, line 68: Can `air_contents` on the cylinder be null?

code/modules/power/stirling.dm
MistakeNot4892: It's more or less a noop.

code/modules/power/stirling.dm
NataKilar: Air tanks create their air_contents on Initialize, so unless something very strange is happening, it shouldn't occur. Nonetheless I added a shortcircuit check above.

code/modules/power/stirling.dm
NataKilar: Fixed

code/modules/power/stirling.dm
NataKilar: Fixed

code/modules/power/stirling.dm
```diff
@@ -0,0 +1,222 @@
...
+		if(!inserted_cylinder)
+			to_chat(user, SPAN_WARNING("You must insert a stirling piston cylinder into \the [src] before you can start it!"))
+			return
+		to_chat(user, "You start trying to manually rev up \the [src].")
+		if(do_after(user, 2 SECONDS, src) && !active && inserted_cylinder && !(stat & BROKEN))
+			visible_message("[user] pulls on the starting cord of \the [src], revving it up!", "You pull on the starting cord of \the [src], revving it up!")
+			playsound(src.loc, 'sound/machines/engine.ogg', 35, 1)
+			active = TRUE
+		return
+	. = ..()
```
Gaxeer: What do you think about switching conditions to:
```
if ((stat & BROKEN) || active)
	return ..()
... code that was in "if (!(stat & BROKEN) && !active)" block
```
will be more readable, as I think

code/modules/power/stirling.dm
```diff
@@ -0,0 +1,222 @@
...
+			stop_engine()
+		skipped_cycle = TRUE
+		return
+	skipped_cycle = FALSE
+
+	// Positive/negative Work from volume expansion/compression. Maximum volume is 1.5 tank volume.	The volume of the tank is not actually adjusted.
+	var/work_coefficient = working_volume.get_total_moles()*R_IDEAL_GAS_EQUATION*log(1.5)
+	var/work_done = work_coefficient*abs(air1.temperature - air2.temperature)*cycle_frequency
+
+	var/reversed = air1.temperature < air2.temperature ? 1 : -1
```
Gaxeer: This `reversed` var naming intuitively makes me think, that it will store boolean type, but it assigned with 1 or -1 which is confusing a bit.

Lower in the code, there is condition, that will be always true, because, as I can see, there is no other changes to this var. 1 and -1 both return TRUE in boolean equivalent.
```
if(reversed)
   air1_dq += 0.75*(max_power - power_generated)
else
   air2_dq += 0.75*(max_power - power_generated)
```
Also, what do you think about moving 1 and -1 to local #defines? I suppose, they must mean something

Gaxeer (changes requested):

code/modules/power/stirling.dm
NataKilar: Sure, done.

code/modules/power/stirling.dm
NataKilar: Good catch. At some point ``reversed`` was a boolean, which is why this check was there, and I probably forgot to change it. I renamed the variable to ``heat_dir`` and swapped the sign so it makes more sense. ``heat_dir = 1`` corresponds to heat transfer from ``air_1`` to ``air_2``, and vice versa.

Gaxeer (dismissed):

code/modules/power/stirling.dm
```diff
@@ -0,0 +1,224 @@
...
+		return ..()
+	
+	if(!inserted_cylinder)
+		to_chat(user, SPAN_WARNING("You must insert a stirling piston cylinder into \the [src] before you can start it!"))
+		return
+	to_chat(user, "You start trying to manually rev up \the [src].")
+	if(do_after(user, 2 SECONDS, src) && !active && inserted_cylinder && !(stat & BROKEN))
+		visible_message("[user] pulls on the starting cord of \the [src], revving it up!", "You pull on the starting cord of \the [src], revving it up!")
+		playsound(src.loc, 'sound/machines/engine.ogg', 35, 1)
+		active = TRUE
```
afterthought2, line 180: This stuff should be in `physical_attack_hand` without the parent call. The initial check will still be needed.

code/modules/power/stirling.dm
```diff
@@ -0,0 +1,224 @@
...
+		stop_engine()
+		return TRUE
+
+	if(panel_open && IS_WRENCH(W))
+		var/target_frequency = input(user, "Enter the cycle frequency you would like \the [src] to operate at ([MAX_FREQUENCY/4] - [MAX_FREQUENCY] Hz)", "Stirling Frequency", cycle_frequency) as num | null
+		if(!CanPhysicallyInteract(user) || !target_frequency)
+			return
+		cycle_frequency = round(clamp(target_frequency, MAX_FREQUENCY/4, MAX_FREQUENCY))
+		to_chat(usr, SPAN_NOTICE("You adjust \the [src] to operate at a frequency of [cycle_frequency] Hz."))
+		return TRUE
```
afterthought2: There are going to be part interaction conflicts with using crowbar and wrench here. Both are fine if panel is closed.

NataKilar: Addressed, changed to panel closed.

code/modules/power/stirling.dm
NataKilar: Done so

afterthought2 (approved):

PR #2322, opened 2022-05-04
Particles support

quardbreak:
## Description of changes
Adds particles support for zcopy and some particles from baystation12.

Nice client-side effects.

## Authorship
Lohikar for zcopy particles code.
Most featuring code from https://github.com/Baystation12/Baystation12/pull/31947


MistakeNot4892:
To clarify after the confusion on Discord, I am not blocking this PR. I have said before that I would prefer the particles to be in a modpack or behind a config, so if this goes in as-is I will PR something to do that.

MistakeNot4892:
Given that Quard has moved on and this has no major interest among the other maints with respect to signing off or merging it, I'm going to close it for now.

PR #2323, opened 2022-05-05
Lower GC phase 2 timeout to 30 seconds

MistakeNot4892:
Alternative to #2276


MistakeNot4892:
Courtesy ping to @Spookerton as this is partially a response to previous comment and @jade-lavenza as this was your PR.

General sentiment among maints who cared enough to comment was that this is a very indirect change, and it's hard to assess meaningful impact from it. Tick checking should handle anything short of hard del lockup when the GC queue is processing, and the other stuff is so subjective that it's hard to accord impact to this change specifically.

Personally I also think that this is a fairly trivial change to make downstream as desired, and isn't one that is going to cause conflicts later, so might as well not be upstreamed. As spook mentioned on the last PR, making it configurable is probably a good middle ground.

PR #2325, opened 2022-05-05
shuffles character preview code

quardbreak:
## Description of changes
The current positions for the character preview mannequins on preview map didn't satisfy me and I got stuck for awhile rewriting it.
I can say I finally start to understand a code I ported a year ago!

изображение

The only noticeable differences is that closing the window won't clear render objects; the backgrounds list become more flexible and can contain icon, icon state and color; and a pAI chassis will display on the map window as a mannequins, somehow.

## Authorship
https://github.com/Haven-13/Haven-Urist/pull/85 gave me inspiration to kick this code and update, so I basically rewrited it after investigating how this feature is work.

### DIFF
--- a/code/modules/client/preference_setup/global/01_ui.dm
+++ b/code/modules/client/preference_setup/global/01_ui.dm
@@ -12,6 +12,25 @@ var/global/list/valid_icon_sizes = list(32, 48, 64, 96, 128)
 	//Style for popup tooltips
 	var/tooltip_style = "Midnight"
 
+	var/list/background_options = list(
+		"Void" = list(
+			"icon" = 'icons/turf/areas.dmi',
+			"icon_state" = "dark",
+			"color" = null
+		),
+		"Dark" = list(
+			"icon" = 'icons/turf/flooring/techfloor.dmi',
+			"icon_state" = "techfloor_gray",
+			"color" = null
+		),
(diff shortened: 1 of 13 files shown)


MistakeNot4892:
Pulled this down to test with it, works well. Actually feels more responsive than the current version.

One issue though, if you go look at your pAI, it does't reset back to your crew character for the other tabs.

code/modules/client/preference_setup/global/01_ui.dm
```diff
@@ -12,6 +12,25 @@ var/global/list/valid_icon_sizes = list(32, 48, 64, 96, 128)
 	//Style for popup tooltips
 	var/tooltip_style = "Midnight"
 
+	var/list/background_options = list(
```
MistakeNot4892, line 15: `var/static` may be appropriate here.

MistakeNot4892 (changes requested):
Code generally looks fine other than possibly using `var/static` on the list as mentioned, and fixing the pAI preview not resetting to crew issue.

MistakeNot4892:
Obviated by #2434 unless you want to update this one instead.

MistakeNot4892:
Feel free to reopen this one or or object on the open PR when/if you come back to this one.

PR #2330, opened 2022-05-06
Rejuvenating the heart should now fix it stopping

PsyCommando:
## Description of changes
Since we don't always recreate all organs on rejuv anymore, the heart wouldn't have its pulse state reset on rejuv.

## Changelog

:cl:
bugfix: Fix heart not restarting on rejuv.
/:cl:

### DIFF
--- a/code/modules/organs/internal/heart.dm
+++ b/code/modules/organs/internal/heart.dm
@@ -218,4 +218,7 @@
 
 /obj/item/organ/internal/heart/rejuvenate(ignore_prosthetic_prefs)
 	. = ..()
-	pulse = PULSE_NORM
+	if(!BP_IS_PROSTHETIC(src))
+		pulse = PULSE_NORM
+	else 
+		pulse = PULSE_NONE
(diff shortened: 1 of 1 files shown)


code/modules/organs/internal/heart.dm
```diff
@@ -214,4 +214,11 @@
 	. = "[pulsesound] pulse"
 
 /obj/item/organ/internal/heart/get_mechanical_assisted_descriptor()
-	return "pacemaker-assisted [name]"
\ No newline at end of file
+	return "pacemaker-assisted [name]"
+
+/obj/item/organ/internal/heart/rejuvenate(ignore_prosthetic_prefs)
+	. = ..()
+	if(!BP_IS_PROSTHETIC(src))
```
MistakeNot4892, line 221: Might be better to emulate resuscitate() and do `pulse = PULSE_NORM; handle_pulse()`

PsyCommando: I don't know. I'd rather let the organ be updated later. Because, if the mob is still not completely rejuvenated, it could just immediately kill the heartbeat again.

comma (approved):

PR #2333, opened 2022-05-07
Add feedback when wiring zcable + tweaks

PsyCommando:
## Description of changes
* Added a message when connecting a upward pointing zcable, to let users know they connected it properly.
* Made it so connecting a upward pointing zcable calls turf_place instead, to make them a bit more coherent with the regular cables.
* Made all turfs hit with a cable coil use the same code, condition check, and error message procs This makes the checks usable elsewhere. (aka in my tweaks)
* Did some misc text string cleanup.

## Changelog

:cl:
tweak: Upward pointing cable stubs now give proper feedback, and install properly when clicked on with a cable coil in hands. You used to need to click on the turf instead before, which was a bit confusing.
/:cl:

### DIFF
--- a/code/game/turfs/simulated/floor_attackby.dm
+++ b/code/game/turfs/simulated/floor_attackby.dm
@@ -173,15 +173,19 @@
 	update_icon()
 	return 1
 
-/turf/simulated/floor/try_build_cable(obj/item/stack/cable_coil/C, mob/user)
-	if(!can_build_cable(user))
-		if(!is_plating() || flooring)
+/turf/simulated/floor/why_cannot_build_cable(var/mob/user, var/cable_error)
+	switch(cable_error)
+		if(0)
+			return
+		if(1)
 			to_chat(user, SPAN_WARNING("Removing the tiling first."))
-		if(broken || burnt)
+		if(2)
 			to_chat(user, SPAN_WARNING("This section is too damaged to support anything. Use a welder to fix the damage."))
-		return FALSE
-	return ..()
+		else //Fallback
+			. = ..()
 
-/turf/simulated/floor/can_build_cable(var/mob/user)
-	//Need to check 'flooring' again since some floor types override is_plating to check for something else than flooring..
-	return is_plating() && !flooring && !broken && !burnt
+/turf/simulated/floor/cannot_build_cable()
+	if(broken || burnt)
+		return 2
+	if(!is_plating())
+		return 1
--- a/code/game/turfs/simulated/floor_attackby.dm
+++ b/code/game/turfs/simulated/floor_attackby.dm
@@ -173,11 +173,15 @@
 	update_icon()
 	return 1
 
+/turf/simulated/floor/try_build_cable(obj/item/stack/cable_coil/C, mob/user)
+	if(!can_build_cable(user))
+		if(!is_plating() || flooring)
+			to_chat(user, SPAN_WARNING("Removing the tiling first."))
+		if(broken || burnt)
+			to_chat(user, SPAN_WARNING("This section is too damaged to support anything. Use a welder to fix the damage."))
+		return FALSE
+	return ..()
+
 /turf/simulated/floor/can_build_cable(var/mob/user)
-	if(!is_plating() || flooring)
-		to_chat(user, "<span class='warning'>Removing the tiling first.</span>")
-		return 0
-	if(broken || burnt)
-		to_chat(user, "<span class='warning'>This section is too damaged to support anything. Use a welder to fix the damage.</span>")
-		return 0
-	return 1
+	//Need to check 'flooring' again since some floor types override is_plating to check for something else than flooring..
+	return is_plating() && !flooring && !broken && !burnt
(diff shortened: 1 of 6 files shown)


code/game/turfs/simulated/floor_attackby.dm
```diff
@@ -173,11 +173,15 @@
 	update_icon()
 	return 1
 
+/turf/simulated/floor/try_build_cable(obj/item/stack/cable_coil/C, mob/user)
+	if(!can_build_cable(user))
+		if(!is_plating() || flooring)
+			to_chat(user, SPAN_WARNING("Removing the tiling first."))
+		if(broken || burnt)
+			to_chat(user, SPAN_WARNING("This section is too damaged to support anything. Use a welder to fix the damage."))
+		return FALSE
```
afterthought2: I've found the superior pattern is to have the check proc return either an error string, possibly an unspecified error signal (like TRUE), or a success (like FALSE or null). The issue with the approach here is that I can potentially override to have a turf not require e.g. !burnt for cabling, but it's then not clear how to override this proc without duplicating code.

Not sure others agree with me here; I haven't seen anyone else use that.

code/game/turfs/simulated/floor_attackby.dm
```diff
@@ -173,11 +173,15 @@
 	update_icon()
 	return 1
 
+/turf/simulated/floor/try_build_cable(obj/item/stack/cable_coil/C, mob/user)
+	if(!can_build_cable(user))
```
MistakeNot4892: Looks like you're duplicating a bunch of checks by using this proc. What's the reasoning?

code/game/turfs/simulated/floor_attackby.dm
PsyCommando: I don't think fixing it by making some complex return code stuff is really the best way here. This should be very simple and concise.
And its more likely that if they changed things so much as to need to change those pretty basic conditions they'd probably need to make a new proc to handle that on the cable coil. And then they'd need to bypass the parent proc to call their own anyways.

code/game/turfs/simulated/floor_attackby.dm
PsyCommando: The reasoning is I haven't found a way not to duplicate at least one check somewhere. And its all in the same proc, so it doesn't cause maintenance issues. And its also a very short proc. 

I could always stub out the try_build_cable() proc on /turf but then I'd copy the code three times for each /turf branches.. Or I could just build the cable without checking in the base class, but that might cause issues. 
I could use some suggestions tbh.

afterthought2: I gave one way to avoid duplicating code in my comment above: have the check proc return a fail message if the check failed. Alternatively, the old-code variant of this pattern: have the check proc alert the user unless it's run with a silent arg.

The point is that by separating the check logic and the failure alerts into different places like this, you make it much harder to override partially.

PsyCommando: Okay, so I changed it and made it easier to override error messages and conditions.
I also noticed during testing that the message about the floor being too damaged is basically never reached, because the turf intercept attackby when its damaged anyways. which kind of makes me wonder why that's there now..

code/modules/power/cable.dm
```diff
@@ -187,19 +187,19 @@ By design, d1 is the smallest direction and d2 is the highest
 		return
 
 	if(d1 == UP || d2 == UP)
-		to_chat(user, "<span class='warning'>You must cut this cable from above.</span>")
+		to_chat(user, SPAN_WARNING("You must cut this [src] from above."))
```
MistakeNot4892: `this the cable` will result from this, you want `[name]` or `cut \the [src]`.

code/modules/power/cable.dm
```diff
@@ -187,19 +187,19 @@ By design, d1 is the smallest direction and d2 is the highest
...
 		return
 
 	if(breaker_box)
-		to_chat(user, "<span class='warning'>This cable is connected to a nearby breaker box. Use the breaker box to interact with it.</span>")
+		to_chat(user, SPAN_WARNING("This [src] is connected to a nearby breaker box. Use the breaker box to interact with it."))
```
MistakeNot4892: As above.

code/modules/power/cable.dm
```diff
@@ -602,13 +602,13 @@ By design, d1 is the smallest direction and d2 is the highest
 	if(ishuman(M) && !M.incapacitated())
 		if(!isturf(usr.loc)) return
 		if(!src.use(15))
-			to_chat(usr, "<span class='warning'>You need at least 15 lengths to make restraints!</span>")
+			to_chat(usr, SPAN_WARNING("You need at least 15 [singular_name] of cable to make restraints!"))
```
MistakeNot4892: `plural_name`

code/modules/power/cable.dm
PsyCommando: fixed

code/modules/power/cable.dm
PsyCommando: fixed

code/modules/power/cable.dm
PsyCommando: fixed

MistakeNot4892 (approved):
I'm ok with this, the message passing is a bit goofy but works.

PR #2324, opened 2022-05-05
Codex ID overlap unit test.

MistakeNot4892:
## Description of changes
- Adds a unit test for codex ID being substrings of each other.
- Adds a unit test for codex names and associated strings not being unique, as this will lead to orphaned codex pages.
- Moves spell codex pages into a category.
- Adds disambiguators and prunes redundant paths from existing codex entries.
- Adds a codex_name var to substances to allow for disambiguating them in the codex without making them display strangely in general play.

## Why and what will this PR improve
Currently if you search 'chemistry' you get the skill page specifically rather than a list of the 12 matching pages. If you search 'robot' you get a page about 'The Ascent', as 'robot' is the initial atom name of the Ascent flying bot. This is generally desirable behavior for extremely specific names but not for general names.

## Authorship
Me.

## Changelog
Not player facing.

PR #2334, opened 2022-05-07
Adds multi-init checking to decls. [MDB IGNORE] [IDB IGNORE]

MistakeNot4892:


MistakeNot4892:
Probably depends on #2324 to pass CI.

afterthought2 (approved):

PR #2337, opened 2022-05-07
Limb dislocation fix + improvements

PsyCommando:
## Description of changes
Dislocation effects wouldn't work properly, so fixed that. Now dislocated limbs show in medical scans again. 
Dislocating people's limbs using grabs, or anything else really, didn't cause any pain at all. Now it does once again.
Now use a limb_flag to tell if a limb can be dislocated instead of a magic number, so it works a bit more like other limb "capabilities".
Added a message during grabs when trying to dislocate or joint lock someone if you don't have the necessary skills.

## Changelog

:cl:
bugfix: Dislocated limbs now show in medical scans again.
bugfix: Dislocating someone's limb now causes pain again.
tweak: Now get feedback when trying to use a jointlock or dislocate someone's limbs and don't have the required skills.
/:cl:

### DIFF
--- a/code/modules/mob/grab/normal/grab_normal.dm
+++ b/code/modules/mob/grab/normal/grab_normal.dm
@@ -52,7 +52,7 @@
 		return FALSE
 
 	if(!assailant.skill_check(SKILL_COMBAT, SKILL_ADEPT))
-		to_chat(assailant, SPAN_WARNING("You attempt clumsily but struggle to do a proper jointlock with \the [affecting]'s [O.name]!"))
+		to_chat(assailant, SPAN_WARNING("You clumsily attempt to jointlock \the [affecting]'s [O.name], but fail!"))
 		return FALSE
 
 	assailant.visible_message(SPAN_DANGER("\The [assailant] begins to [pick("bend", "twist")] \the [affecting]'s [O.name] into a jointlock!"))
@@ -85,7 +85,7 @@
 		return  FALSE
 
 	if(!assailant.skill_check(SKILL_COMBAT, SKILL_ADEPT))
-		to_chat(assailant, SPAN_WARNING("You attempt clumsily but can't seem to dislocate \the [affecting]'s [O.joint]!"))
+		to_chat(assailant, SPAN_WARNING("You clumsily attempt to dislocate \the [affecting]'s [O.name], but fail!"))
 		return FALSE
 
 	if(!O.is_dislocated() && (O.limb_flags & ORGAN_FLAG_CAN_DISLOCATE))
--- a/code/modules/mob/grab/normal/grab_normal.dm
+++ b/code/modules/mob/grab/normal/grab_normal.dm
@@ -98,7 +98,7 @@
 			return TRUE
 		affecting.visible_message(SPAN_WARNING("\The [assailant] fails to dislocate \the [affecting]'s [O.joint]."))
 		return FALSE
-		
+
 	if(O.limb_flags & ORGAN_FLAG_CAN_DISLOCATE)
 		to_chat(assailant, SPAN_WARNING("\The [affecting]'s [O.joint] is already dislocated!"))
 	else
--- a/code/modules/mob/grab/normal/grab_normal.dm
+++ b/code/modules/mob/grab/normal/grab_normal.dm
@@ -90,8 +90,8 @@
 			return TRUE
 		affecting.visible_message(SPAN_WARNING("\The [assailant] fails to dislocate \the [affecting]'s [O.joint]."))
 		return FALSE
-
-	if (O.dislocated > 0)
+		
+	if(O.limb_flags & ORGAN_FLAG_CAN_DISLOCATE)
 		to_chat(assailant, SPAN_WARNING("\The [affecting]'s [O.joint] is already dislocated!"))
 	else
 		to_chat(assailant, SPAN_WARNING("You can't dislocate \the [affecting]'s [O.joint]!"))
--- a/code/modules/mob/grab/normal/grab_normal.dm
+++ b/code/modules/mob/grab/normal/grab_normal.dm
@@ -43,12 +43,16 @@
 		return FALSE
 
 	var/mob/living/assailant = G.assailant
-	if(!assailant || !assailant.skill_check(SKILL_COMBAT, SKILL_ADEPT))
+	if(!assailant)
 		return FALSE
 
 	var/obj/item/organ/external/O = G.get_targeted_organ()
 	if(!O)
 		to_chat(assailant, SPAN_WARNING("\The [affecting] is missing that body part!"))
+		return FALSE
+
+	if(!assailant.skill_check(SKILL_COMBAT, SKILL_ADEPT))
+		to_chat(assailant, SPAN_WARNING("You attempt clumsily but struggle to do a proper jointlock with \the [affecting]'s [O.name]!"))
 		return FALSE
 
 	assailant.visible_message(SPAN_DANGER("\The [assailant] begins to [pick("bend", "twist")] \the [affecting]'s [O.name] into a jointlock!"))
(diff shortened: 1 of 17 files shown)


code/modules/organs/external/_external.dm
```diff
@@ -58,7 +58,7 @@
 	// Joint/state stuff.
 	var/joint = "joint"                // Descriptive string used in dislocation.
 	var/amputation_point               // Descriptive string used in amputation.
-	var/dislocated = 0                 // If you target a joint, you can dislocate the limb, causing temporary damage to the organ.
+	var/dislocated = 0                 // If you target a joint, you can dislocate the limb, causing temporary damage to the organ. Boolean.
```
MistakeNot4892: Could use a `limb_status` flag for this considering it's no longer a trinary condition.

code/modules/mob/grab/normal/grab_normal.dm
```diff
@@ -43,14 +43,18 @@
...
+	if(!assailant)
 		return FALSE
 
 	var/obj/item/organ/external/O = G.get_targeted_organ()
 	if(!O)
 		to_chat(assailant, SPAN_WARNING("\The [affecting] is missing that body part!"))
 		return FALSE
 
+	if(!assailant.skill_check(SKILL_COMBAT, SKILL_ADEPT))
+		to_chat(assailant, SPAN_WARNING("You attempt clumsily but struggle to do a proper jointlock with \the [affecting]'s [O.name]!"))
```
MistakeNot4892: Would suggest:
`You clumsily attempt to jointlock \the [affecting]'s [O.name], but fail!`

code/modules/mob/grab/normal/grab_normal.dm
```diff
@@ -72,26 +76,30 @@
...
 		return FALSE
 
 	var/obj/item/organ/external/O = G.get_targeted_organ()
 	if(!O)
 		to_chat(assailant, SPAN_WARNING("\The [affecting] is missing that body part!"))
 		return  FALSE
 
-	if(!O.dislocated)
+	if(!assailant.skill_check(SKILL_COMBAT, SKILL_ADEPT))
+		to_chat(assailant, SPAN_WARNING("You attempt clumsily but can't seem to dislocate \the [affecting]'s [O.joint]!"))
```
MistakeNot4892: Would suggest:
`You clumsily attempt to dislocate \the [affecting]'s [O.name], but fail!`

code/modules/organs/external/_external.dm
```diff
@@ -365,44 +365,47 @@
 	return all_limbs
 
 /obj/item/organ/external/proc/is_dislocated()
-	if(dislocated > 0)
-		return 1
-	if(is_parent_dislocated())
-		return 1//if any parent is dislocated, we are considered dislocated as well
-	return 0
+	return (dislocated > 0) || is_parent_dislocated() //if any parent is dislocated, we are considered dislocated as well
```
MistakeNot4892: Could just be `dislocated || is_parent_dislocated()` now.

code/modules/organs/external/_external.dm
```diff
@@ -365,44 +365,47 @@
...
 
 /obj/item/organ/external/proc/is_parent_dislocated()
 	var/obj/item/organ/external/O = parent
-	while(O && O.dislocated != -1)
-		if(O.dislocated == 1)
-			return 1
+	while(O && (O.limb_flags & ORGAN_FLAG_CAN_DISLOCATE))
+		if(O.dislocated > 0)
```
MistakeNot4892: `if(O.dislocated)`

code/modules/organs/external/_external.dm
```diff
@@ -365,44 +365,47 @@
...
+	return FALSE
 
 /obj/item/organ/external/proc/update_internal_organs_cost()
 	internal_organs_size = 0
 	for(var/obj/item/organ/internal/org in internal_organs)
 		internal_organs_size += org.get_storage_cost()
 
 /obj/item/organ/external/proc/dislocate()
-	if(dislocated == -1)
+	if(owner && owner.status_flags & GODMODE)
```
MistakeNot4892: Need `()` for operator clarity.

code/modules/organs/external/_external.dm
```diff
@@ -365,44 +365,47 @@
...
+		return
+	if(!(limb_flags & ORGAN_FLAG_CAN_DISLOCATE))
 		return
 
-	dislocated = 1
+	dislocated = TRUE
 	if(owner)
+		if(owner.can_feel_pain(src))
```
MistakeNot4892: Believe `add_pain` already takes can_feel_pain into account and will handle stuff like limb-specific numbness better than `owner.can_feel_pain()`. Checking code, calling owner.can_feel_pain with src is strictly redundant in view of it just calling src.can_feel_pain() anyway.

code/modules/organs/external/_external.dm
PsyCommando: That's a good idea! 
Speaking of limb status, some of the state in there probably shouldn't be a status, like prosthetic and assisted. Because those are more like permanent changes, and also because rejuv resets them on rejuv, cause they happen to be set in the status flag.

code/modules/mob/grab/normal/grab_normal.dm
PsyCommando: You're right, that does sounds a lot better XD

code/modules/organs/external/_external.dm
PsyCommando: True

code/modules/organs/external/_external.dm
PsyCommando: By limb specific numbness what do you mean? Like does it adds extra effects?
Also, the main reason for the check is for the owner.apply_effect(). Putting add_pain() in there alongside just means it'll do the check a single time instead of twice if I've had placed add_pain() outside.

code/modules/organs/external/_external.dm
PsyCommando: To be clearer, I feel like ORGAN_ASSISTED, ORGAN_PROSTHETIC, and ORGAN_CRYSTAL probably should be limb flags and not status flags. Since they define permanent capabilities and not transient states? I could probably do something on another PR if you feel it makes sense.

PsyCommando:
Aagh rebase why ;n;
Should have merged...

code/modules/organs/external/_external.dm
```diff
@@ -549,18 +551,22 @@
...
 	burn_dam = 0
 	germ_level = 0
-	pain = 0
 	genetic_degradation = 0
+
 	for(var/datum/wound/wound in wounds)
 		qdel(wound)
 	number_wounds = 0
 
+	dislocated = initial(dislocated)
```
PsyCommando: @MistakeNot4892 Seems like you merged a fix to something I'm fixing before my fix was fixed O_o

code/modules/organs/external/_external.dm
PsyCommando: Also made dislocation a status flag now.

code/modules/organs/external/_external.dm
MistakeNot4892: I think that came in with one of the GC fixes whoops

code/modules/organs/external/_external.dm
MistakeNot4892: Conceptually, if we later want to add some kind of per-limb pain response, pain handling, etc. then this change wouldn't be ideal due to prioritizing mob-level responses over limb-level. The main reason I think you should use limb-level pain procs is that as discussed on Discord the mob level procs just wrap the limb level ones anyway.

code/modules/organs/external/_external.dm
MistakeNot4892: I don't disagree but I also don't personally think it's a huge priority currently.

MistakeNot4892:
Looks good other than the pain proc comment.

code/modules/organs/external/_external.dm
PsyCommando: Yeah, I'll change it to the limb level proc in a bit.

MistakeNot4892 (approved):

Issue #132, opened 2020-03-14
Blast and fire doors don't do automatic direction set


quardbreak:
Currently every mapped blast and fire door need to manually direction changed, without it, horizontally they're will look like here:
![image](https://user-images.githubusercontent.com/44920739/76682123-56c9e700-660a-11ea-967e-a9a5c3607ce1.png)
When they must look like here:
![image](https://user-images.githubusercontent.com/44920739/76682167-cb9d2100-660a-11ea-835e-fa33cdb234e4.png)

afterthought2:
I don't believe any doors do this automatically; it may be reasonable, but it does make certain mapping assumptions. Are you suggesting that there is code for them to autoset like this in the repo, but it doesn't work, or are you proposing this as a feature?

quardbreak:
More like it's doesn't work now. You can check for example the airlock code where it was able to automatically set direction based on list of objects or turfs around.

quardbreak:
Also, I think recent blending refactor could have influenced this.

PR #2341, opened 2022-05-07
Blast doors now autoset direction from neighbors.

MistakeNot4892:
Implements #132, firedoors already do this.
Airlocks will also now respect set_dir_on_update.

### DIFF
--- a/code/game/machinery/doors/airlock.dm
+++ b/code/game/machinery/doors/airlock.dm
@@ -295,10 +295,12 @@ About the new airlock wires panel:
 		return 0
 
 /obj/machinery/door/airlock/on_update_icon(state=0, override=0)
-	if(connections & (NORTH|SOUTH))
-		set_dir(EAST)
-	else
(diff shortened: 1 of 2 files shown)


MistakeNot4892:
Noting that they are still going to look weird on tradeship due to mapping and smoothing lists.

image

Work fine if left to their own devices.

comma (approved):

afterthought2:
I don't actually buy they should be doing this in `update_icon` at all. I think they should do this in `Initialize`.

MistakeNot4892:
They need to be able to update when connections change.

afterthought2:
I'm not sure that's true; they should likely not be changing dir just because their neighbors changed. But even if that's the case, it should be done in the connections update proc, which is separate from `update_icon`.

MistakeNot4892:
Well, I'm unwilling to touch or refactor any connection or direction setting code and this change is consistent with the other door types.

Issue #183, opened 2020-03-19
Grab and drag does not work well with multiz movement.


ghostsheet:
~~First issue: ctrl-click grabbing an object then clicking on openspace turf with the grab, doesn't drag that object to the openspace turf (unlike other turfs). 
Clicking with the grab on a turf, normally move the object to the turf.~~

Second issue: ctrl-click grabbing objects doesn't like Multi Z movement, it doesn't allow you to pull objects up or down through space or through ladders.
Pulling objects should come up with you for space at the very least. 

~~Third issue: ctrl-click grabbing objects doesn't count atheletics or item size for movespeed delay. It should have some way to count and apply the encumbrance properly~~

MistakeNot4892:
Encumberance is factored into move delay these days.
Will put a PR up with the first issue.

PR #2342, opened 2022-05-07
Generalizes grab-carrying across z-levels.

MistakeNot4892:
## Description of changes
Implements remaining aspects of #183.
- Grab move logic has been centralized in Move(), which should mean it behaves more consistently between methods of traversing z-levels.
- Falling now uses Move() rather than forceMove() to propagate grabs. Pulled mobs are still injured by the fall.
- Ladders will allow you to carry 75% of your pull strength up them (100% if you are in zero-G).

Tested stairs, ladders, falling through several levels, and spacemoves. Everything appears to work.

Consistent handling of grabs.

## Authorship
Myself.

## Changelog
:cl:
tweak: You can now carry things up ladders and through space with grabs.
/:cl:

### DIFF
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -513,7 +513,6 @@ default behaviour is:
 		if(G.assailant_reverse_facing())
 			set_dir(global.reverse_dir[direction])
 		G.assailant_moved()
-		G.adjust_position()
 		if(QDELETED(G) || QDELETED(G.affecting))
 			mygrabs -= G
 
@@ -561,7 +560,7 @@ default behaviour is:
 /mob/living/Move(NewLoc, Dir)
 	if (buckled)
 		return
-	var/old_loc = loc
+	var/turf/old_loc = loc
 	. = ..()
 	if(.)
 		handle_grabs_after_move(old_loc, Dir)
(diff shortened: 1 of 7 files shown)


code/modules/mob/living/living.dm
```diff
@@ -493,21 +493,79 @@ default behaviour is:
...
+			continue
+
+	if(length(mygrabs) && !skill_check(SKILL_MEDICAL, SKILL_BASIC))
+		for(var/obj/item/grab/grab AS_ANYTHING in mygrabs)
 			var/mob/living/affecting_mob = grab.get_affecting_mob()
 			if(affecting_mob)
 				affecting_mob.handle_grab_damage()
 
-/mob/living/Move(a, b, flag)
+/mob/living/Move(NewLoc, Dir)
```
MistakeNot4892, line 560: Noting that these are the args stated in the docco, and last edit was Cael 9 years ago.

afterthought2:
Did you test this with grabbed dense non-mobs, and with dense non-mobs (and I guess mobs) blocking movement at various points? On one hand, I think most of your logic is the same as it was before; on the other hand, just examining the code would suggest the logic might break with block checking on up/down-and-sideways diagonal movement, like what happens on stairs.

MistakeNot4892:
I tested with Poppy, a wrench, a closed closet, and an open closet. Also tried pulling around diagonals, from awkward angles, etc. There's a pre-existing bug with pulling mobs on a diagonal dropping the grab that is still occurring in this one, but my main goal was making the z-transit code consistent rather than fixing that one.

afterthought2 (approved):
Good enough for me, then.

PR #2348, opened 2022-05-08
Refactoring skill book use to allow for adjacent reading.

MistakeNot4892:
## Description of changes
You can now continue to reference a skill book so long as it is either in your direct inventory (hands, pocket, belt as applicable) or on a tile adjacent to you (put it on a table, etc).

This will make skillbooks more broadly useful for things like piloting or construction as they won't take up an entire hand slot permanently. It will also allow Baxxid to use skillbooks properly on downstream scav.

## Changelog
:cl:
tweak: Skill books can now be used from adjacent tiles.
/:cl:

### DIFF
--- a/code/game/objects/items/books/skill_book.dm
+++ b/code/game/objects/items/books/skill_book.dm
@@ -38,16 +38,16 @@ Skill books that increase your skills while you activate and hold them
 	author = "The Oracle of Bakersroof"
 	icon_state = "book2"
 	force = 4
-	w_class = ITEM_SIZE_LARGE //Skill books are THICC with knowledge. Up one level from regular books to prevent library-in-a-bag silliness.
+	w_class = ITEM_SIZE_LARGE            // Skill books are THICC with knowledge. Up one level from regular books to prevent library-in-a-bag silliness.
 	unique = TRUE
 	material = /decl/material/solid/plastic
 	matter = list(/decl/material/solid/wood = MATTER_AMOUNT_REINFORCEMENT)
 
-	var/decl/hierarchy/skill/skill // e.g. SKILL_LITERACY
-	var/skill_req = SKILL_NONE //The level the user needs in the skill to benefit from the book, e.g. SKILL_PROF
-	var/reading = FALSE //Tto check if the book is actively being used
-	var/custom = FALSE //To bypass init stuff, for player made textbooks and weird books. If true must have details manually set
-	var/ez_read = FALSE //Set to TRUE if you can read it without basic literacy skills
+	var/decl/hierarchy/skill/skill       // e.g. SKILL_LITERACY
+	var/skill_req = SKILL_NONE           // The level the user needs in the skill to benefit from the book, e.g. SKILL_PROF
+	var/weakref/reading                  // To check if the book is actively being used
+	var/custom = FALSE                   // To bypass init stuff, for player made textbooks and weird books. If true must have details manually set
+	var/ez_read = FALSE                  // Set to TRUE if you can read it without basic literacy skills
 
 	var/skill_name = "missing skill name"
 	var/progress = SKILLBOOK_PROG_FINISH // used to track the progress of making a custom book. defaults as finished so, you know, you can read the damn thing
@@ -55,6 +55,8 @@ Skill books that increase your skills while you activate and hold them
 /obj/item/book/skill/Initialize()
 
 	. = ..()
+
+	global.events_repository.register(/decl/observ/moved, src, src, .proc/check_buff)
 
 	if(!custom && skill && skill_req)// custom books should already have all they need
 		skill_name = initial(skill.name)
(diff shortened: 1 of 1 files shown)


code/game/objects/items/books/skill_book.dm
```diff
@@ -82,64 +82,106 @@ Skill books that increase your skills while you activate and hold them
...
 	if(user.get_skill_value(skill) > skill_req)
 		to_chat(user, SPAN_WARNING("You already know everything [title] has to teach you!"))
-		return
+		return FALSE
+
+	return TRUE
+
+/obj/item/book/skill/attack_self(mob/user)
+
+	if(reading && reading == weakref(user))
```
afterthought2: I believe the correct check is always `user && user == reading?.resolve()` or equivalent. Avoids creating unnecessary weakrefs and is independent of the weakref implementation.

code/game/objects/items/books/skill_book.dm
```diff
@@ -82,64 +82,106 @@ Skill books that increase your skills while you activate and hold them
...
+		//Close book, get rid of buffs
+		src.unlearn(user)
```
afterthought2: Extraneous `src`.

code/game/objects/items/books/skill_book.dm
```diff
@@ -82,64 +82,106 @@ Skill books that increase your skills while you activate and hold them
...
+		to_chat(user, SPAN_NOTICE("You close the [name]. That's enough learning for now."))
```
afterthought2: Probably `\the [name]`?

code/game/objects/items/books/skill_book.dm
```diff
@@ -82,64 +82,106 @@ Skill books that increase your skills while you activate and hold them
...
+	if(!reading)
+		return PROCESS_KILL
+	check_buff()
+
+/obj/item/book/skill/proc/check_buff()
+	var/mob/R = reading?.resolve()
+	if(!istype(R) || R.incapacitated())
+		remove_buff()
+		return
+	if(loc != R && !R.Adjacent(src))
```
afterthought2: I would at least consider `CanPhysicallyInteract` (possibly for both this and the incapacitated check above).

code/game/objects/items/books/skill_book.dm
```diff
@@ -82,64 +82,106 @@ Skill books that increase your skills while you activate and hold them
...
-	if(reading)
-		to_chat(user, SPAN_DANGER("You lose the page you were on! You can't cross-reference using the [name] like this!"))
-		var/mob/M = user
-		if(istype(M) && M.fetch_buffs_of_type(/datum/skill_buff/skill_book, 0))
-			src.unlearn(user)
-		reading = FALSE
+/obj/item/book/skill/Move()
 	. = ..()
+	if(.)
+		check_buff()
```
afterthought2: This override won't work with forceMove, if you e.g. store the book in a bag. Is this intentional? If not, consider moved events instead.

code/game/objects/items/books/skill_book.dm
```diff
@@ -82,64 +82,106 @@ Skill books that increase your skills while you activate and hold them
...
+
+/obj/item/book/skill/dropped()
+	check_buff()
```
afterthought2: That seems to answer above question with `forceMove`. I'd consider subscribing to your own moved event.

code/game/objects/items/books/skill_book.dm
MistakeNot4892: `\the [src]` or `the [name]` - `\the [name]` gives `the the foo`

MistakeNot4892:
Made them more valuable, implemented suggestions.

code/game/objects/items/books/skill_book.dm
```diff
@@ -82,64 +84,102 @@ Skill books that increase your skills while you activate and hold them
...
-		user.buff_skill(buff, buff_type = /datum/skill_buff/skill_book)
-		reading = TRUE
-		to_chat(user, SPAN_NOTICE("You find the information you need! Better keep the page open to reference it."))
-	else
+	if(!user.do_skilled(4 SECONDS, SKILL_LITERACY, src, 0.75))
 		to_chat(user, SPAN_DANGER("Your perusal of the [name] was interrupted!"))
-		return
+		return TRUE
+
+	if(QDELETED(user) || user.incapacitated() || !CanPhysicallyInteract(user) || (loc != user && !user.Adjacent(src)))
```
afterthought2: CanPhysicallyInteract generally renders Adjacent and incapacitated checks redundant. Occasionally that's not the case (incapacitated flags, CanPhysicallyInteract overrides) but sort of doubting it's the case here?

code/game/objects/items/books/skill_book.dm
```diff
@@ -82,64 +84,102 @@ Skill books that increase your skills while you activate and hold them
...
-		return
+		return FALSE
 
-	if(reading) //Close book, get rid of buffs
-		src.unlearn(user)
-		to_chat(user, SPAN_NOTICE("You close the [name]. That's enough learning for now."))
-		reading = FALSE
-		return
+	if(reading)
+		if(reading != weakref(user))
```
afterthought2: Commented on this before; feel it should be `reading.resolve() != user`.

MistakeNot4892:
Changes addressed - have also pointed this to staging because it was initially a fix for downstream and they are now basing on staging instead of dev. Abuse of powers oh no.

afterthought2 (approved):

PR #2355, opened 2022-05-08
Lowers the threshold for cleaners to act on turfs.

MistakeNot4892:
## Description of changes
Allows materials to set the minimum volume at which they will interact with turfs. Sets this to 0.1 on cleaners.

Mops can currently be ruined by introducing any amount of other reagents to the mop bucket or mop.

## Authorship
Me.

## Changelog
Nothing player facing.


MistakeNot4892:
Chucked a couple of other small fixes on here.

afterthought2 (approved):

PR #2357, opened 2022-05-09
Hearth GC fixes.

MistakeNot4892:
## Description of changes
Ported a variety of GC changes from Hearth, tweaked a few of them and added some additional GC/Destroy() stuff to 

Fewer harddels from GC fails.

## Authorship
Jade/Hearth for the base commits, myself for the tweaks.

## Changelog
Nothing player-facing.

### DIFF
--- a/code/game/atoms_init.dm
+++ b/code/game/atoms_init.dm
@@ -73,6 +73,8 @@
 	LAZYCLEARLIST(our_overlays)
 	LAZYCLEARLIST(priority_overlays)
 
+	LAZYCLEARLIST(climbers)
+
 	QDEL_NULL(light)
 
 	if(opacity)
--- a/code/game/atoms_init.dm
+++ b/code/game/atoms_init.dm
@@ -95,7 +95,17 @@
 		loc.Entered(src, null)
 
 /atom/movable/Destroy()
-	. = ..()
+	// These must be done before event cleanup in the parent call
+	if(LAZYLEN(movement_handlers) && !ispath(movement_handlers[1]))
+		QDEL_NULL_LIST(movement_handlers)
+
+	if (bound_overlay)
+		QDEL_NULL(bound_overlay)
+
+	if(virtual_mob && !ispath(virtual_mob))
+		qdel(virtual_mob)
+		virtual_mob = null
+
 #ifdef DISABLE_DEBUG_CRASH
 	// meh do nothing. we know what we're doing. pro engineers.
 #else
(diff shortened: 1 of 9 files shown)


code/game/atoms_init.dm
```diff
@@ -106,12 +117,7 @@
...
-
-	if (bound_overlay)
-		QDEL_NULL(bound_overlay)
+	vis_locs = null //clears this atom out of all vis_contents
+	clear_vis_contents(src)
 
-	if(virtual_mob && !ispath(virtual_mob))
-		qdel(virtual_mob)
-		virtual_mob = null
+	. = ..() // called last so that events are unregistered before the parent call
```
afterthought2: Confusing comment: events are unregistered _in_ the parent call, what does this mean?

One thing this change does is re-order when the destroyed event is raised vs. everything else in this proc. Now it is raised after being moved to nullspace and deleting everything in `src`, which makes it impossible to preempt being deleted while inside something via destroyed events. I don't know how much code relies on this, but I also wouldn't be surprised if this breaks things (e.g. power).

code/modules/hydroponics/spreading/spreading_growth.dm
```diff
@@ -143,6 +143,8 @@
 		if(!istype(check_turf))
 			continue
 		for(var/obj/effect/vine/neighbor in check_turf.contents)
+			if(QDELETED(neighbor))
```
afterthought2, line 146: Should comment on what's going on here (presumably some kind of deletion loop), as it looks wrong at a glance.

code/game/atoms_init.dm
MistakeNot4892: I think(?) that they mean movement handlers rather than events. Not seeing anything else in Destroy() that relates. Unsure about the ordering question.

code/modules/hydroponics/spreading/spreading_growth.dm
MistakeNot4892: It's a refresh proc to wake up vines that may have fallen asleep. Not actually sure if the loop needs to happen at all.

code/game/atoms_init.dm
noelle-lavenza: No, it has to be moved to nullspace prior to the parent call so that `cleanup_events()` doesn't run unnecessarily, since it's very slow. Moving to nullspace unregisters several events that are registered on `Entered()`. IMO a different pre-destroy event would be better for what you're talking about; IIRC in some PR I left a comment with a TODO relating to exactly that.

MistakeNot4892: Two questions from the Discord convo are -
- Where is the human registering to the crypod's events?
- Why not just dump everything into nullspace or unregister the events manually before qdel?

afterthought2 (approved):

PR #2360, opened 2022-05-09
Lazylists bad_external_organs, applies ref fixes from Hearth.

MistakeNot4892:
## Description of changes
- bad_external_organs no longer has as many dangling references.
- bad_external_organs is now lazylisted.

tidy tidy tidy

## Authorship
@jade-lavenza, myself.

## Changelog
Nothing player-facing.

### DIFF
--- a/code/modules/mob/living/carbon/human/human_organs.dm
+++ b/code/modules/mob/living/carbon/human/human_organs.dm
@@ -8,7 +8,7 @@
 	var/obj/item/organ/external/E = get_organ(zone)
 	if(E) . = E.name
 
-/mob/living/carbon/human/proc/recheck_bad_external_organs()
+/mob/living/carbon/human/proc/should_recheck_bad_external_organs()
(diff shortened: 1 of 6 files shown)


code/modules/mob/living/carbon/human/human_organs.dm
```diff
@@ -17,31 +17,38 @@
 		. = TRUE
 	last_dam = damage_this_tick
 
+/mob/living/carbon/human/proc/recheck_bad_external_organs()
+	LAZYCLEARLIST(bad_external_organs)
+	for(var/obj/item/organ/external/E in organs)
```
noelle-lavenza:
```suggestion
	for(var/obj/item/organ/external/E in get_external_organs())
```

noelle-lavenza (changes requested):
Think this is the only change needed to update it?

comma (approved):

PR #2361, opened 2022-05-09
Adds qdeleted check to atom temperature queuing.

MistakeNot4892:
## Description of changes
Adds QDELETED check to atom temperature queuing. Unsure if this will do anything, but does seem like a safe change.

## Why and what will this PR improve
May clear a dangling ref.

## Authorship
@jade-lavenza 

## Changelog
Nothing player-facing.

PR #2365, opened 2022-05-10
Fix GC issues with temperature (+ maybe reactions)

noelle-lavenza:
## Description of changes
Supersedes #2361.
Atoms will now remove themselves from `SStemperature.processing` when qdeleted.
Qdeleted reagent holders will no longer be queued for updating.
Renames and rewrites a few temperature macros.

Fixes that cascade of GC fails approximately 2 minutes after the atoms info repository inits.
May also fix a bunch of other GC fails as well.

## Authorship
Me.


MistakeNot4892 (approved):
Looks good to me and discussion on Discord seemed logical.

PR #2373, opened 2022-05-13
fixing several bugs

kelgaras:
## Description of changes
* Blood on objects that can be attacked now has DNA, according to which the detective can work.
* Welding is spent when crafting a flamethrower
* Nutrients in the biogenerator are now displayed correctly
* Now you can use fingerprint card (when fingerprints taken from mobs hands) in microscope and receive correct fingerprint data in report.

bug fix is good
## Authorship
@keIgaras  and @SmellsLikeAFox
## Changelog
:cl:
bugfix: Nutrients in the biogenerator are now displayed correctly.
bugfix: Welding is spent when crafting a flamethrower
bugfix: Blood on objects that can be attacked now has DNA
bugfix: Now you can use fingerprint card in microscope and receive correct fingerprint data in report
/:cl:

### DIFF
--- a/code/game/objects/item.dm
+++ b/code/game/objects/item.dm
@@ -700,19 +700,26 @@ var/global/list/slot_flags_enumeration = list(
 		blood_overlay.color = COLOR_LUMINOL
 		update_icon()
 
-/obj/item/add_blood(mob/living/carbon/human/M, amount = 2, blood_data)
+/obj/item/add_blood(mob/living/carbon/human/human, amount = 2, blood_data)
 	if (!..())
-		return 0
+		return FALSE
 
 	if(istype(src, /obj/item/energy_blade))
-		return
+		return FALSE
 
-	if(!blood_data && istype(M))
-		blood_data = REAGENT_DATA(M.vessel, /decl/material/liquid/blood)
-	var/datum/extension/forensic_evidence/forensics = get_or_create_extension(src, /datum/extension/forensic_evidence)
-	forensics.add_data(/datum/forensics/blood_dna, LAZYACCESS(blood_data, "blood_DNA"))
-	add_coating(/decl/material/liquid/blood, amount, blood_data)
-	return 1 //we applied blood to the item
+	if(blood_data || !istype(human))
+		return TRUE
+
+	blood_data = REAGENT_DATA(human.vessel, /decl/material/liquid/blood)
+
+	if(!LAZYACCESS(blood_DNA, human.dna.unique_enzymes))
+		LAZYSET(blood_DNA, human.dna.unique_enzymes, human.dna.b_type)
+		LAZYSET(blood_data, human.dna.unique_enzymes, REAGENT_DATA(human.vessel, human.species.blood_reagent))
+		var/datum/extension/forensic_evidence/forensics = get_or_create_extension(human, /datum/extension/forensic_evidence)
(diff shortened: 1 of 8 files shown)


code/modules/reagents/reagent_containers/glass/bottle.dm
```diff
@@ -100,13 +100,15 @@
 	initial_reagents = list(/decl/material/gas/ammonia = 60)
 
 /obj/item/chems/glass/bottle/eznutrient
+	name = "EZ NUtrient"
```
MistakeNot4892, line 103: This isn't a fix, these names will be cooked as soon as the holder updates from reagents.
The actual fix is making the biogenerator use the atom_info_repository instead of initial() in ui_interact() a la `atom_info_repository.get_name_for(path)`.

code/game/objects/item.dm
```diff
@@ -700,19 +700,26 @@ var/global/list/slot_flags_enumeration = list(
 		blood_overlay.color = COLOR_LUMINOL
 		update_icon()
 
-/obj/item/add_blood(mob/living/carbon/human/M, amount = 2, blood_data)
+/obj/item/add_blood(mob/living/carbon/human/human, amount = 2, blood_data)
```
MistakeNot4892, line 703: Why change the arg name? This is now different to every other add_blood proc.

MistakeNot4892:
We're about to go into a staging period, would it be possible for you to point these fixes at `stable` rather than `dev`?

code/game/objects/item.dm
keIgaras: a more descriptive name for a variable than just a single letter, I guess

kelgaras:
> We're about to go into a staging period, would it be possible for you to point these fixes at `stable` rather than `dev`?

no problem as soon as I get home

code/modules/reagents/reagent_containers/glass/bottle.dm
kelgaras: This way it is displayed more pleasantly, and the cache will not store unnecessary information.
the second screenshot with my option
image
image

code/game/objects/item.dm
kelgaras: > Why change the arg name? This is now different to every other add_blood proc.

a more convenient name for an argument than just one letter, and in a good way there should not be arguments consisting of one letter at all

afterthought2: There are stupid byond issues with locally renaming proc arguments. Either rename it everywhere or don't rename it.

code/modules/reagents/reagent_containers/glass/bottle.dm
MistakeNot4892: But the bottles themselves are going to be broken.

MistakeNot4892: You could use `initial(label_text)` instead of the atom repo, but ultimately you need to change it on the biogenerator not the bottle.

MistakeNot4892: Nvm, looks like `base_name` overrides name in Init so this is fine. Not a fan of the solution but biogenerator in general needs a rewrite so w/e.

MistakeNot4892:
Needs a rebase/conflict resolution.

PR #2381, opened 2022-05-14
Attempting to make organ init more consistent.

### LINKED ISSUE #2376: Surgery is inconsistently broken.
#### Description of issue
Surgey will sometimes decide that mobs that joined via the lobby can have infinite incisions made on them, and will not accept any of those incisions as valid for further surgery steps.

Surgery should find an open incision and use it.

#### Steps to reproduce
Take someone who joins from the lobby into medbay and try to cut them open.

Definitely seems to be specific to mobs joining via the lobby, so likely related to the use or ordering of set_species procs.

Presumably since the big organ rewrite.

Client Version: 514
Server Revision: 1364caab858d0fcb1c02d0fc2280fca879c05484 - dev - 2022-05-13
Game ID: chV-arBV

#### Issue bingo

MistakeNot4892:
## Description of changes
Organ init now calls the same consistent logic regarding setup as organic/robotic. Appears to fix #2376.

Surgery works for birds

## Authorship
Me @afterthought2 and @PsyCommando 

## Changelog
Nothing player facing.

### DIFF
--- a/code/modules/organs/internal/_internal.dm
+++ b/code/modules/organs/internal/_internal.dm
@@ -15,8 +15,6 @@
 /obj/item/organ/internal/Initialize(mapload, material_key, datum/dna/given_dna)
 	if(!alive_icon)
 		alive_icon = initial(icon_state)
-	if(max_damage)
-		min_bruised_damage = FLOOR(max_damage / 4)
 	. = ..()
 
 /obj/item/organ/internal/set_species(species_name)
--- a/code/modules/mob/living/carbon/human/human_organs.dm
+++ b/code/modules/mob/living/carbon/human/human_organs.dm
@@ -224,7 +224,10 @@
 
 /mob/living/carbon/human/proc/sync_organ_dna()
 	for(var/obj/item/organ/O in get_organs())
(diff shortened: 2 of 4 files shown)


code/modules/organs/internal/_internal.dm
```diff
@@ -13,11 +13,11 @@
 	var/damage_reduction = 0.5     //modifier for internal organ injury
 
 /obj/item/organ/internal/Initialize(mapload, material_key, datum/dna/given_dna)
+	. = ..()
```
PsyCommando: Dunno if you tried it out, but you probably want to leave the parent call at the end, since icon_updates are going to get triggered with the alive icon not set in some organ's case.

PsyCommando: Its probably worth just moving those to set_species though.

MistakeNot4892: Damage threshold set turned out to be redundant, it's done in set_max_damage() already. Moved parent call back to the end.

afterthought2 (approved):

deathride58 (approved):

PR #2387, opened 2022-05-14
fixing several bugs

kelgaras:
## Description of changes
* Blood on objects that can be attacked now has DNA, according to which the detective can work.
* Welding is spent when crafting a flamethrower
* Scrubber icon "In" are now displayed correctly
* Now you can use fingerprint card (when fingerprints taken from mobs hands) in microscope and receive correct fingerprint data in report.

bug fix is good
## Authorship
@keIgaras  and @SmellsLikeAFox
## Changelog

bugfix: fix scrubber animation swap
bugfix: Welding is spent when crafting a flamethrower
bugfix: Blood on objects that can be attacked now has DNA
bugfix: Now you can use fingerprint card in microscope and receive correct fingerprint data in report


MistakeNot4892:
Currently doesn't compile.

code/game/objects/item.dm
```diff
@@ -684,17 +684,24 @@ var/global/list/slot_flags_enumeration = list(
...
 	if(istype(src, /obj/item/energy_blade))
 		return
 
-	if(!blood_data && istype(M))
-		blood_data = M.vessel.reagent_data[/decl/material/liquid/blood]		
-	var/datum/extension/forensic_evidence/forensics = get_or_create_extension(src, /datum/extension/forensic_evidence)
-	forensics.add_data(/datum/forensics/blood_dna, blood_data["blood_DNA"])
-	add_coating(/decl/material/liquid/blood, amount, blood_data)
-	return 1 //we applied blood to the item
+	if(blood_data || !istype(M))
```
MistakeNot4892: Won't this prevent extra blood DNA being added to `blood_data`?

code/game/objects/items/weapons/tools/weldingtool.dm
```diff
@@ -98,14 +98,14 @@
...
 
-	if((!status) && (istype(W,/obj/item/stack/material/rods)))
+	if(!status && istype(W, /obj/item/stack/material/rods))
 		var/obj/item/stack/material/rods/R = W
 		R.use(1)
-		var/obj/item/flamethrower/F = new/obj/item/flamethrower(user.loc)
-		user.drop_from_inventory(src, F)
-		F.weldtool = src
-		master = F
 		add_fingerprint(user)
```
MistakeNot4892, line 112: Unrelated to your code, but I'm a bit curious why they were adding fingerprints to something they immediately qdeleted. I wonder if they meant to do `F.add_fingerprint(user)`.

code/game/objects/item.dm
kelgaras: No, tested in game

MistakeNot4892: I got this around the wrong way - this code means that if a non-null blood_data is supplied to the proc (which is done in at least one spot), nothing will ever be added to it. This isn't the same as the previous code where blood_data not being provided to the proc results in it being set to the reagent data.

Also I suspect that the lazyset on blood_data is mutating the reagent data list.

kelgaras: image
stabbed two different mobs, got two different blood.

If I understand you correctly, of course

MistakeNot4892: There's a specific interaction when you click on a blood puddle that supplies a list to `blood_data`. In the previous code, if `blood_data` was unsupplied, it was set to the reagent data, but otherwise kept going and updated forensics info. In your new code, it will abort early if `blood_data` is supplied.

code/game/objects/item.dm
```diff
@@ -684,17 +684,24 @@ var/global/list/slot_flags_enumeration = list(
...
+		return TRUE
+
+	blood_data = REAGENT_DATA(M.vessel, /decl/material/liquid/blood)
+
+	if(!LAZYACCESS(blood_DNA, M.dna.unique_enzymes))
+		LAZYSET(blood_DNA, M.dna.unique_enzymes, M.dna.b_type)
+		LAZYSET(blood_data, M.dna.unique_enzymes, REAGENT_DATA(M.vessel, M.species.blood_reagent))
```
MistakeNot4892, line 713: Why are you setting `blood_data` at all here? This line will modify the blood data of the source mob while it's inside their body.

PR #2391, opened 2022-05-14
Tidying up a half-finished movable screen object removal.

### LINKED ISSUE #2255: Ability buttons are gone
#### Description of issue
The ability buttons on your HUD do not appear anywhere. Only the button to show/hide them is visible. Item-specific buttons seem to work though. Bay's version works.

Buttons for things like spells do not appear on your HUD when given.

#### Steps to reproduce
Load game.
Spawn a contract, spellbook, or use give spell command.
Try to find where the corresponding button to use it is.

N/A

Found out the day before while testing spells.

Client Version: 514
Server Revision: 6145049cb4eb27878ecdf216aacb9080b2cea1f8 - dev - 2021-12-14
Game ID: cfu-cSf3

#### Issue bingo

MistakeNot4892:
## Description of changes
- Restores some overzealous removal of screen_loc decode/encode procs.
- Updates procs to use TOP/BOTTOM LEFT/RIGHT.
- Removes remnants of /obj/screen/movable path; this behavior needs rewriting to be implemented with our scaling view and the first botched removal is what broke the above.
- Closes #2255

Fixes screen elements not showing up.

## Authorship
Me.

## Changelog
Nothing player facing.


code/_onclick/hud/ability_screen_objects.dm
```diff
@@ -1,4 +1,56 @@
-/obj/screen/movable/ability_master
+/// Encode a mob's client view size into a screen_loc X component.
+/proc/encode_screen_X(var/X, var/mob/viewer)
```
afterthought2: Sorry but I hate the casing. The `x` is not capitalized even at dm-level in general, and the next line calls a proc with x not capitalized. I mean specifically the proc name, not the arg, and similarly for the other procs below here.

code/_onclick/hud/ability_screen_objects.dm
```diff
@@ -60,9 +112,28 @@
 		overlays.Add(open_state)
 	update_icon()
 
-/obj/screen/movable/ability_master/proc/open_ability_master()
+/obj/screen/ability_master/proc/open_ability_master()
+	var/list/screen_loc_xy = splittext(screen_loc,",")
+
+	//Create list of X offsets
+	var/list/screen_loc_X = splittext(screen_loc_xy[1],":")
+	var/x_position = decode_screen_X(screen_loc_X[1], my_mob)
```
afterthought2: I feel like somewhere, maybe here or maybe in the decoder proc, you need to check to make sure this is a legitimate x position in the sense that when it's re-encoded you get something sane. Maybe it already does the fallback correctly and I'm paranoid, maybe it already fields everything it should and so if this is `null` or whatever you should crash; I'm not sure.

afterthought2 (commented):
I don't have high confidence in checking the math here, so going to assume you tested it.

code/_onclick/hud/ability_screen_objects.dm
MistakeNot4892: I think I am going to sit down and implement a set of helpers for encoding and deencoding arbitrary screen locs so I can use them in other UI stuff beyond this, as well as stuff like the click catcher horseshit we were discussing the other day. This code is stuff I've restored from commit history rather than something I wrote, for reference. I did test the PR overall but I have no confidence that this is the best way or even an appropriate way to handle screen locs.

afterthought2: I think that's a good idea. With the way the client screen is set up currently, we'll be seeing more and more need to mess with client-dependent screen locations.

code/_onclick/hud/ability_screen_objects.dm
MistakeNot4892: Casing fixed.

code/_onclick/hud/ability_screen_objects.dm
MistakeNot4892: Ended up stripping the screen loc decoding stuff out since it relates to the old movable ability master.

MistakeNot4892:
Bumped it over 1 tile to avoid it overlapping with the 'look up' button, but revised version seems to work:
image

code/_onclick/hud/ability_screen_objects.dm
```diff
@@ -60,9 +61,15 @@
 		overlays.Add(open_state)
 	update_icon()
 
-/obj/screen/movable/ability_master/proc/open_ability_master()
+/obj/screen/ability_master/proc/open_ability_master()
+	for(var/i = 1 to length(ability_objects))
+		var/obj/screen/ability/A = ability_objects[i]
+		var/row = round(i/abilities_per_row)
+		A.screen_loc = "RIGHT-[(i-(row*abilities_per_row))+2]:16,TOP-[row+1]:16"
```
MistakeNot4892, line 68: Would be nice to move this to vis_contents in the future but I'm leaving it as screen_loc for now.

afterthought2 (approved):

PR #2393, opened 2022-05-14
Fixes anchored mobs being impossible to mousedrag in any circumstances.

### LINKED ISSUE #1609: Removing all worn items on another player stops after first item is removed
#### Description of issue
You must have an empty hand to remove equipment from another character, and you can only remove one at a time.

When attempting to remove another character's equipment, you are able to select multiple entries like normal; however as soon as the first object is removed, it goes into your open hand and then all progress bars are immediately removed with the objects still equipped to the downed player.

I would expect all items to be unequipped in the order which they were clicked

#### Steps to reproduce
1. Click and drag from another character to yourself
2. Ensure you have an empty hand selected
3. Quickly click all the equipment you can 
4. Watch as the first item is removed and see all of the progress bars fade out

Likely wherever the "remove equipment" screen is housed. That may not actually be the name of the screen.

Dunno!

Client Version: 513
Server Revision: 6ce3da8783adf9318c59cbbddaaaafe0542efa9a - staging - 2021-05-06
Game ID: cbU-dqMu

#### Issue bingo

### LINKED ISSUE #2113: cant take off clothes from buckled human
#### Description of issue
cant take off clothes from buckled human

?

#### Steps to reproduce
1) buckle someone to something like a chair or rollerbed 
2)try to take off it clothes

#### Issue bingo

MistakeNot4892:
## Description of changes
- Removes anchored check from base mouse drop proc, adds it back in where appropriate.
- Adds a hand check to do_mob, renames arg in do_after to match.
- Fixes #2113.
- Fixes #1609.

Allows for mousedragging anchored mobs in some circumstances and restores stripping to expected Bay behavior.

## Authorship
Myself.

## Changelog
Nothing player-facing.

### DIFF
--- a/code/_helpers/mobs.dm
+++ b/code/_helpers/mobs.dm
@@ -71,7 +71,7 @@
 /proc/get_exposed_defense_zone(var/atom/movable/target)
 	return pick(BP_HEAD, BP_L_HAND, BP_R_HAND, BP_L_FOOT, BP_R_FOOT, BP_L_ARM, BP_R_ARM, BP_L_LEG, BP_R_LEG, BP_CHEST, BP_GROIN)
 
-/proc/do_mob(mob/user , mob/target, time = 30, target_zone = 0, uninterruptible = 0, progress = 1, var/incapacitation_flags = INCAPACITATION_DEFAULT)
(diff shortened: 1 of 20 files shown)


afterthought2 (approved):
Pretty risky approach, but does feel "correct" to some extent. The alternative would be an extra breakout proc for `can_mouse_drop`, like `can_receive_mouse_drop` on the `over` atom, which would default to `!anchored` but could be overridden (or vice versa I guess).

deathride58 (approved):

PR #2395, opened 2022-05-15
Fixes click catcher coverage.

MistakeNot4892:
## Description of changes
Extends click catchers to cover the entire possible view.

Currently on widescreen click catchers do not cover the left side of the screen.

## Authorship
Myself.

## Changelog
Nothing player-facing.

### DIFF
--- a/code/_helpers/turfs.dm
+++ b/code/_helpers/turfs.dm
@@ -41,23 +41,9 @@
 	if(turfs.len)
 		return pick(turfs)
 
-/proc/screen_loc2turf(text, turf/origin)
-	if(!origin)
-		return null
-	var/tZ = splittext(text, ",")
-	var/tX = splittext(tZ[1], "-")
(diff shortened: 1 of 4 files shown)


afterthought2 (dismissed):

MistakeNot4892:
Click catchers are behaving very oddly with regards to mapping to the appropriate turf, but this does extend coverage to the entire screen at least.

afterthought2:
Probably need to check the client's actual view size when reading clicks, as e.g. the top-right 0, 0 click catcher will be on different turfs for different clients with different views. Alternatively, offset them from screen center instead of from top right.

MistakeNot4892:
Didn't end up doing much with screen loc, but this PR is now working and responds to rescaling windows.

afterthought2 (approved):
Concept seems fine. Whether or not this makes off-by-one errors in some configurations, no idea, but hey.

PR #2396, opened 2022-05-16
Splitting serpentids into their own modpack.

MistakeNot4892:
## Description of changes
Moves serpentid code into a modpack separate to the Ascent.

I want to do some work on both species as separate concepts and at the moment you can't really mess with one without messing with the other.

## Authorship
Me.

## Changelog
Nothing player-facing.


afterthought2 (approved):
I hope they enjoy their new habitat.

Issue #1803, opened 2021-07-01
Inconsistent unit test failure on dev


MistakeNot4892:
````
## UNIT_TEST ##: [POWER: All areas must have accurate power use values.]: The area Surface of Trovato-Upsilon had improper power use values on the equip channel: was 2000 but should be 1000.
## UNIT_TEST ##: !!! FAILURE !!! [POWER: All areas must have accurate power use values.]: At least one area had improper power use values
````
Appears to be related to things loading maps out of the correct order, possibly?

MistakeNot4892:
Observed this today in #2340

PR #2405, opened 2022-05-17
Apparent fix to power usage; harsher unit test

afterthought2:
## Description of changes

Testing suggests #1679 completely broke power for any machines spawned after mapload (e.g. constructed in round). This is a theoretical fix. It should resolve issues with the exoplanet power unit test failures. The unit test has been upgraded to catch this type of issue more consistently.

Aiming at staging because it seems power is quite broken there; would suggest aggressive testing (monitor APC power usage while building/destroying/moving machines around; move things in shuttles).

The change in #1679 feels inconsistent (should either do it unconditionally or not at all, not just during not mapload), but I don't have the ability to test enough right now to comfortably PR something that touches that line. Changing it would allow for more robust fixes.

Power broke, I try fix.

## Authorship

me

### DIFF
--- a/code/unit_tests/power_tests.dm
+++ b/code/unit_tests/power_tests.dm
@@ -71,22 +71,37 @@
 
 /datum/unit_test/area_power_tally_accuracy
 	name = "POWER: All areas must have accurate power use values."
+	var/list/channel_names = list("equip", "light", "environ")
+/obj/machinery/test_machine
+	power_channel = EQUIP
+	idle_power_usage = 1000
+	use_power = POWER_USE_IDLE
+
+/datum/unit_test/area_power_tally_accuracy/proc/check_power(var/area/A)
+	var/list/old_values = list(A.used_equip, A.used_light, A.used_environ)
+	A.retally_power()
+	var/list/new_values = list(A.used_equip, A.used_light, A.used_environ)
+	for(var/i in 1 to length(old_values))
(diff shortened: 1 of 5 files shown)


code/unit_tests/power_tests.dm
```diff
@@ -71,22 +71,37 @@
...
+	power_channel = EQUIP
+	idle_power_usage = 1000
+	use_power = POWER_USE_IDLE
+
+/datum/unit_test/area_power_tally_accuracy/proc/check_power(var/area/A)
+	var/list/old_values = list(A.used_equip, A.used_light, A.used_environ)
+	A.retally_power()
+	var/list/new_values = list(A.used_equip, A.used_light, A.used_environ)
+	for(var/i in 1 to length(old_values))
+		if(abs(old_values[i] - new_values[i]) > 1) // Round because there can in fact be roundoff error here apparently.
```
MistakeNot4892, line 85: Not seeing a round here.

MistakeNot4892:
Will this potentially address #1803?

code/unit_tests/power_tests.dm
afterthought2: The unrounded check would have been old_values == new_values; the way here accounts for roundoff error in math elsewhere.

afterthought2:
Yes, it should fix #1803 (though probably not the more recent scrubbers error).

MistakeNot4892 (approved):

PR #2411, opened 2022-05-19
Small changes from #2204

MistakeNot4892:
## Description of changes
Splitting some helper procs/macros from #2204.
- Adds a coloured square HTML macro.
- Adds a z-level naming system.
- Adds an equipped proc on items for robots.
- Tweaked some filter setting to use proc.

Prep work for #2204, also generally nice little things to have.

## Authorship
Myself.

## Changelog
Nothing player-facing.


afterthought2 (approved):
Should `equipped_robot` also send in the slot name (or whatever the hell the appropriate robot concept is) as a second arg?

MistakeNot4892:
Don't think it matters, robit slots are interchangeable.

PR #2414, opened 2022-05-20
fix: fix blast door opening/closing issue. Adjust opening/closing to animation time

Gaxeer:
## Description of changes

Move open_final and close_final procs to /obj/machinery/door. Fix bug with blast door won't open/close if you click too fast.
Move open_final and close_final procs to /obj/machinery/door.

Fix annoying visual bug with blast door, shatters (animation is too fast, so closing/opening finishes before actual door closing/opening). Add proper check for blast door state (closed/open) via calling can_close and can_open.

## Authorship

Until now only me:
@Gaxeer 

## Changelog

:cl:
bugfix: Adjust calling to update of Door icon to right after the animation finish
bugfix: Fix bug with blast door won't open/close if you click too fast
refactor: Move open_final and close_final procs to /obj/machinery/door. 
/:cl:

### DIFF
--- a/code/game/machinery/doors/_door.dm
+++ b/code/game/machinery/doors/_door.dm
@@ -393,52 +393,57 @@
 				flick("door_deny", src)
 				playsound(src.loc, 'sound/machines/buzz-two.ogg', 50, 0)
 
-/obj/machinery/door/proc/open(var/forced = 0)
+/obj/machinery/door/proc/open(forced = FALSE)
 	if(!can_open(forced))
 		return
+
 	operating = 1
 
 	do_animate("opening")
 	icon_state = "door0"
-	set_opacity(0)
-	sleep(3)
-	src.set_density(0)
+	set_opacity(FALSE)
+
+	sleep(0.5 SECONDS)
+	src.set_density(FALSE)
 	update_nearby_tiles()
-	sleep(7)
+
+	sleep(0.5 SECONDS)
 	src.layer = open_layer
 	update_icon()
-	set_opacity(0)
+	set_opacity(FALSE)
 	operating = 0
 
 	if(autoclose)
 		close_door_at = next_close_time()
 
-	return 1
+	return TRUE
 
 /obj/machinery/door/proc/next_close_time()
 	return world.time + (normalspeed ? 150 : 5)
 
-/obj/machinery/door/proc/close(var/forced = 0)
+/obj/machinery/door/proc/close(forced = FALSE)
 	if(!can_close(forced))
 		return
+
 	operating = 1
 
 	close_door_at = 0
 	do_animate("closing")
-	sleep(3)
-	src.set_density(1)
-	src.layer = closed_layer
+
+	sleep(0.5 SECONDS)
+	src.set_density(TRUE)
 	update_nearby_tiles()
-	sleep(7)
+	src.layer = closed_layer
+
+	sleep(0.5 SECONDS)
 	update_icon()
 	if(visible && !glass)
-		set_opacity(1)	//caaaaarn!
+		set_opacity(TRUE)
 	operating = 0
 
 	//I shall not add a check every x ticks if a door has closed over some fire.
 	var/obj/fire/fire = locate() in loc
-	if(fire)
-		qdel(fire)
+	qdel(fire)
 
 /obj/machinery/door/proc/toggle(to_open = density)
 	if(to_open)
--- a/code/game/machinery/doors/blast_door.dm
+++ b/code/game/machinery/doors/blast_door.dm
@@ -110,11 +110,12 @@
 	operating = 1
 	playsound(src.loc, open_sound, 100, 1)
 	flick(icon_state_opening, src)
-	set_density(0)
+
+	sleep(0.6 SECONDS)
+	set_density(FALSE)
 	update_nearby_tiles()
 	update_icon()
-	set_opacity(0)
-	sleep(15)
+	set_opacity(FALSE)
 	layer = open_layer
 	operating = 0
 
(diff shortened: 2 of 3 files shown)


code/game/machinery/doors/_door.dm
```diff
@@ -396,49 +396,57 @@
...
-		set_opacity(1)	//caaaaarn!
-	operating = 0
+		set_opacity(TRUE)
+	operating = FALSE
 
 	//I shall not add a check every x ticks if a door has closed over some fire.
 	var/obj/fire/fire = locate() in loc
-	if(fire)
-		qdel(fire)
+	QDEL_NULL(fire)
```
afterthought2: both old and new are unnecessarily fancy; `qdel(fire)` is sufficient here.

code/game/machinery/doors/blast_door.dm
```diff
@@ -107,40 +107,44 @@
...
 // Proc: force_toggle()
 // Parameters: None
 // Description: Opens or closes the door, depending on current state. No checks are done inside this proc.
 /obj/machinery/door/blast/proc/force_toggle(to_open = density)
 	if(to_open)
-		force_open()
+		open()
 	else
-		force_close()
+		close()
```
afterthought2: Why this change? I believe this is used with radio stuff to remotely open groups of doors, so having them not force can leave them in weird staggered states.

EDIT: on reviewing the rest I can sort of see why you did this, but I'd suggest either not doing it (assuming it was just wrong before) or renaming the proc.

code/game/machinery/doors/blast_door.dm
```diff
@@ -107,40 +107,44 @@
...
 	flick(icon_state_closing, src)
-	set_density(1)
+	addtimer(CALLBACK(src, .proc/close_final), 0.6 SECONDS, TIMER_UNIQUE)
+
+/obj/machinery/door/blast/close_final()
+	set_density(TRUE)
 	update_nearby_tiles()
 	update_icon()
-	set_opacity(1)
-	sleep(15)
```
afterthought2, line 133: Was this time just incorrect relative to the animations?

code/game/machinery/doors/windowdoor.dm
```diff
@@ -103,18 +103,18 @@
...
-/obj/machinery/door/window/proc/open_final()
+/obj/machinery/door/window/open_final()
 	explosion_resistance = 0
 	set_density(FALSE)
 	update_icon()
 	update_nearby_tiles()
 
-	if(operating == 1) //emag again
-		operating = 0
+	if(operating) //emag again
```
afterthought2: I believe the reason it says emag here is that emag sets `operating` to `-1` or some other value and this is meant to not un-emag the door. Please double-check, and if that's wrong and the `==1` is pointless here then update the comment.

code/game/machinery/doors/_door.dm
```diff
@@ -396,49 +396,57 @@
...
 	icon_state = "door0"
-	set_opacity(0)
-	sleep(3)
-	src.set_density(0)
+	set_opacity(FALSE)
+	sleep(0.3 SECONDS)
+	src.set_density(FALSE)
 	update_nearby_tiles()
-	sleep(7)
+	addtimer(CALLBACK(src, .proc/open_final), 0.7 SECONDS, TIMER_UNIQUE)
```
afterthought2: Am a bit confused; why do we _both_ sleep and then create a timer? Won't one of them have the same issue as you're trying to resolve in the current code (and which in theory should be resolved already via the `operating` var?

code/game/machinery/doors/windowdoor.dm
```diff
@@ -103,18 +103,18 @@
 	icon_state = "[src.base_state]open"
 	flick("[src.base_state]opening", src)
 	playsound(src.loc, 'sound/machines/windowdoor.ogg', 100, 1)
-	addtimer(CALLBACK(src, .proc/open_final), 10, TIMER_UNIQUE | TIMER_OVERRIDE)
+	addtimer(CALLBACK(src, .proc/open_final), 0.9 SECONDS, TIMER_UNIQUE | TIMER_OVERRIDE)
```
afterthought2: Why do we override here but not in the other places?

code/game/machinery/doors/blast_door.dm
Gaxeer: yes it was

Gaxeer: The animation takes 0.6 seconds

code/game/machinery/doors/_door.dm
Gaxeer: I am not really sure we need timer here. I would remove it completely for doors, because we have this 2 sleep logic 
(If door is slightly opened, we can already move through it)

code/game/machinery/doors/windowdoor.dm
Gaxeer: I am not sure we need timer override. The logic of this proc won't let `addtimer` while the door is `operating`,
so it's unreachable, until the timer end.

Gaxeer:
I made a decision to remove those timers and extra procs. I think sleeps are enough here.
Also I've adjusted door time to become dense and vice versa to 0.5 SECONDS. The gap we have in door on this timing, is size of player model:D 0.2 secs won't make much sense, but looks slightly better

afterthought2 (dismissed):

MistakeNot4892:
Could you retarget this at staging please?

Gaxeer:
> Could you retarget this at staging please?

Done

afterthought2 (approved):

PR #2415, opened 2022-05-20
Adding basic secret loading subsystem.

MistakeNot4892:
## Description of changes
Adds an ~~extremely simple~~ early-load subsystem that scrapes a data directory for json files and loads them in as datums in a secrets cache that can be retrieved from elsewhere in the code. I'm adding this now because I want to start working on hotloaded maps and this seemed like an ideal place to start.

Adds a super basic framework for non-AGPL content to be hotloaded instead of compiled.

## Authorship
Myself.

## Changelog
Nothing player-facing.


code/controllers/subsystems/initialization/secrets.dm
```diff
@@ -0,0 +1,109 @@
...
+		"data/secrets/",
+		"maps/",
+		"mods/"
+	)
+	var/list/secrets = list()
+
+// Copied from customitems loading; scrapes entire file tree for json files.
+/datum/controller/subsystem/secrets/Initialize()
+	
+	. = ..()
```
afterthought2: iirc this needs to go last for it to report time correctly, but may be wrong.

code/controllers/subsystems/initialization/secrets.dm
```diff
@@ -0,0 +1,109 @@
...
+			if(secrets[secret_key])
+				log_warning("Duplicate secret_key [secret_key] in [checkfile]!")
+				continue
+			var/datum_path = text2path(loaded_data["path"])
+			if(!ispath(datum_path))
+				log_warning("Invalid or empty datum path ([datum_path || "NULL"]) loaded from [checkfile]!")
+				continue
+
+			loaded_data["_source_dir"] = checkdir
+			loaded_data["_source_file"] = relative_checkfile
```
afterthought2, line 197: I think you should throw errors if these are already set in `loaded_data`.

code/controllers/subsystems/initialization/secrets.dm
```diff
@@ -0,0 +1,109 @@
...
+			secrets[secret_key] = new datum_path(loaded_data)
```
afterthought2: Just noting that this is extremely unsafe. Obviously this system requires some level of trust, but it's also possible to just crash the server with this line. Ideally you would sanity-check the datum_path against some reasonable set of datums (/atom is fine but will maul the loc, /decl is not fine, controllers are bad, admin holders are bad). If you are planning to use this with various non-atoms and don't feel further filtering is reasonable here, I think you should log the type this is about to create.

If the plan is like secret chems/reactions/codex entries, I'd consider something like
```
/datum/secret_maker

/datum/secret_maker/chems/New(loaded_data)
    chem making logic goes here
```
and likewise subtypes for other secret types, and then check `ispath(datum_path, /datum/secret_maker)` above. This is much safer.

I also think you should log the data payload. It's possible it should remain secret from admins below some level, but it should be possible to log in secret in that case. It's a relevant debug/grief prevention measure.

code/controllers/subsystems/initialization/secrets.dm
```diff
@@ -0,0 +1,109 @@
...
+	name = "Secret Content"
+	init_order = SS_INIT_EARLY
+	flags = SS_NO_FIRE
+	/// Root location of content to load; terminating / is important for example dir check.
+	var/list/load_directories = list(
+		"data/secrets/",
+		"maps/",
+		"mods/"
+	)
+	var/list/secrets = list()
```
afterthought2: Consider VV-protecting this, depending on your goals.

afterthought2 (commented):
I would like to see your example secret be moved/copied out of the example dir (and possibly renamed), actually loaded, and then accessed in a unit test.

I also strongly recommend a "nuke-all-and-reload" verb for debugging.

code/controllers/subsystems/initialization/secrets.dm
MistakeNot4892: I'm not inclined towards a wrapper datum just because the idea behind this system was already going to be loading datum structures rather than trying to load atoms or anything that is going to be exposed directly to users. It feels like it would redundant and costly to do a factory-style intermediary pattern.

I've added blacklist/whitelist types that forbid any atom or non-datum paths for the moment.

Where would you want the payload logged? I'm really unsure how this can be a debugging or grief prevention measure considering anyone with write access to the server directory is already going to be able to check what's being ingested.

MistakeNot4892:
PR now includes unit tests, an admin verb to nuke and reload, and a fairly extended example implementation in the form of a secret note. If I keep letting scope creep out it's never going to be finished so I'm going to leave it alone for now.

MistakeNot4892:
Icon setting works: 
image

code/controllers/subsystems/initialization/secrets.dm
MistakeNot4892: This actually helped me spot a serious problem with `cached_json_decode()`.

code/controllers/subsystems/initialization/secrets.dm
MistakeNot4892: Done.

code/controllers/subsystems/initialization/secrets.dm
MistakeNot4892: I've now specifically only listed the implemented example secret path as permitted, and I have added a dangerous types logging line.

afterthought2 (approved):

PR #2417, opened 2022-05-21
Fix slurred speech.

noelle-lavenza:
## Description of changes

BYOND doesn't use `break` to exit switches. Oops. My bad.

Before, it would only get a few characters in before stopping prematurely.
## Authorship
Me for the bug and me for the fix.

## Changelog

:cl: Noelle Lavenza
bugfix: Slurred speech now works again.
/:cl:

### DIFF
--- a/code/modules/mob/mob_helpers.dm
+++ b/code/modules/mob/mob_helpers.dm
@@ -247,7 +247,7 @@ var/global/list/global/organ_rel_size = list(
 			if(1,3,5,8)	newletter="[lowertext(newletter)]"
 			if(2,4,6,15)	newletter="[uppertext(newletter)]"
 			if(7)	newletter+="'"
-			if(9 to 14)	break
+			//if(9 to 14) do nothing
 			//if(9,10)	newletter="<b>[newletter]</b>"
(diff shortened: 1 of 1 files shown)


MistakeNot4892 (approved):
Why the comments?

noelle-lavenza:
> Why the comments?

I, uh, I'm not sure. I think it was just to make it explicit that 9 to 14 did nothing? Since the ranges used here are kind of weird; it alternates odd-even for no good reason.

The other comments were already present.

Gaxeer:
> > Why the comments?
> 
> I, uh, I'm not sure. I think it was just to make it explicit that 9 to 14 did nothing? Since the ranges used here are kind of weird; it alternates odd-even for no good reason.
> 
> The other comments were already present.

You've got `else` in switch/case

MistakeNot4892:
`else` in a switch is valid in DM.

Gaxeer:
> `else` in a switch is valid in DM.

I mean, it's possible to replace this `if (9 to 14)` with `else`

Gaxeer:
> `else` in a switch is valid in DM.

I am blind. I thought Jade `add` if (9 to 14)

PR #2418, opened 2022-05-21
Prevent inserting a part twice into component_parts

PsyCommando:
## Description of changes
If a part is already in the component_parts list, don't let anyone add another reference to it. 
It doesn't happen in regular usage, but downstream its been quite annoying at times. 
Plus, it probably should be clearly enforced that this list must not have duplicates.

### DIFF
--- a/code/game/machinery/_machines_base/machinery_components.dm
+++ b/code/game/machinery/_machines_base/machinery_components.dm
@@ -148,7 +148,7 @@ var/global/list/machine_path_to_circuit_type
 		. = part
 
 	if(istype(part))
-		LAZYADD(component_parts, part)
+		LAZYDISTINCTADD(component_parts, part)
 		part.on_install(src)
 		events_repository.register(/decl/observ/destroyed, part, src, .proc/component_destroyed)
 	else if(ispath(part))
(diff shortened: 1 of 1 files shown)


afterthought2:
I don't like this; I'd rather it CRASH or ASSERT instead. You are right that this shouldn't happen, but if it does happen it's best to find out when/why and fix it.

PsyCommando:
> I don't like this; I'd rather it CRASH or ASSERT instead. You are right that this shouldn't happen, but if it does happen it's best to find out when/why and fix it.

So you'd want me to add a check to see if every newly installed components were already in the components list?

afterthought2:
Do I want you to? Not particularly; it's not something that I have evidence of happening. You might also consider a unit test that looks for duplicated parts in all machines.

But I don't want you to do what you are doing here, while a check + crash would be something I am OK with.

PsyCommando:
> Do I want you to? Not particularly; it's not something that I have evidence of happening. You might also consider a unit test that looks for duplicated parts in all machines.
> 
> But I don't want you to do what you are doing here, while a check + crash would be something I am OK with.

You're being a bit ambiguous. You're saying you're ok with a check and a crash. But you're also saying you don't want me to do a check and a crash?

afterthought2 (approved):

PR #2419, opened 2022-05-21
Prevents pipe from runtiming during atmos_init in nullspace

PsyCommando:
## Description of changes
Prevents pipe from runtiming during atmos_init in nullspace.
Normally this doesn't happen. But in some special cases it can. As when you create a pipe but cannot create it in its place yet. and want to move it later in position. Or downstream when we create pipes in nullspace before moving them in position.


afterthought2:
I think it's better to nullcheck loc in `Initialize` and then basically just return `..()` if loc is null, avoiding all init completely.

For what it's worth: objs which are created or placed in nullspace runtiming is always a bug, and should be fixed. It's not just a downstream issue; it means that stuff like the atoms repository and unit tests can start picking this up upstream as well. But generally the fix should be to bypass all surrounding-dependent logic, which for pipes is most of it.

PsyCommando:
> I think it's better to nullcheck loc in `Initialize` and then basically just return `..()` if loc is null, avoiding all init completely.
> 
> For what it's worth: objs which are created or placed in nullspace runtiming is always a bug, and should be fixed. It's not just a downstream issue; it means that stuff like the atoms repository and unit tests can start picking this up upstream as well. But generally the fix should be to bypass all surrounding-dependent logic, which for pipes is most of it.

Initialize isn't the only thing calling atmos_init though.

afterthought2:
Other calls to `atmos_init` cannot be made while the thing is in nullspace; they come from construction code or from adjacent pipes.

PsyCommando:
> Other calls to `atmos_init` cannot be made while the thing is in nullspace; they come from construction code or from adjacent pipes.

The controller does call it on everything blindly:

```dm

/datum/controller/subsystem/machines/proc/setup_atmos_machinery(list/machines)
	set background=1

	report_progress("Initializing atmos machinery")
	for(var/obj/machinery/atmospherics/A in machines)
		A.atmos_init()
		CHECK_TICK

	report_progress("Initializing pipe networks")
	for(var/obj/machinery/atmospherics/machine in machines)
		machine.build_network()
		CHECK_TICK
```
Plus its async...

afterthought2:
OK, fair enough, forgot it does it there rather than in Initialize. Retarget and it's good to go.

PsyCommando:
Forgot to push something. And I'm a bit worried about rebasing, because its mentioning 25 commits when this pr has only 2?

PsyCommando:
doesn't seem like the rebasing helped much...

PsyCommando:
Okay, nvm this pr is broken. Use the new one. Closing this.

PR #2423, opened 2022-05-22
Fix bad limb max damage scaling

PsyCommando:
## Description of changes
Fix the problem with limbs getting inconsistent damage values depending on how they were created. 
Turns out the values were scaled down on each call of set_species.
Also fix neo-avian no spawning with any languages when created in-game or when changing a mob's species.

I dunno if it'll be useful since loaf has been fixing it too. But here it is in any cases.

## Changelog

:cl:
bugfix: Fix incoherent max damage scaling on human mob limbs. Also fix things depending on it.
/:cl:

### DIFF
--- a/code/game/dna/dna2_helpers.dm
+++ b/code/game/dna/dna2_helpers.dm
@@ -167,7 +167,7 @@
 
 		//Base skin and blend
 		for(var/obj/item/organ/external/E in H.get_external_organs())
-			E.set_dna(E.dna)
+			E.set_dna(H.dna)
 
 		//Hair
 		var/list/hair_subtypes = subtypesof(/decl/sprite_accessory/hair)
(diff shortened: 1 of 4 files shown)


MistakeNot4892 (approved):

MistakeNot4892:
Oops, fixes need to target staging. Could you rebase?

MistakeNot4892:
I had to pull your commits into staging for my PR so if you'd prefer we can just merge them with that one rather than rebasing here.

PsyCommando:
> I had to pull your commits into staging for my PR so if you'd prefer we can just merge them with that one rather than rebasing here.

Sure, you can just merge!

MistakeNot4892:
Alrighty cheers.

PR #2424, opened 2022-05-22
fix: adjust mob's move direction, only if it is an initiator of movement

Gaxeer:
## Description of changes

You can now grab-move mob with STAT_CONFUSE, without him moving in random directions

Handling of confused mob moving is now work correct, I suppose.

## Authorship

@Gaxeer

## Changelog

:cl:
prefix:
/:cl:

### DIFF
--- a/code/datums/movement/mob.dm
+++ b/code/datums/movement/mob.dm
@@ -25,9 +25,9 @@
 	mob.ghostize()
 
 // Incorporeal/Ghost movement
-/datum/movement_handler/mob/incorporeal/DoMove(var/direction)
+/datum/movement_handler/mob/incorporeal/DoMove(var/direction, var/mob/mover)
 	. = MOVEMENT_HANDLED
-	direction = mob.AdjustMovementDirection(direction)
+	direction = mob.AdjustMovementDirection(direction, mover)
 	mob.set_glide_size(0)
 
 	var/turf/T = get_step(mob, direction)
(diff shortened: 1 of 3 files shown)


code/datums/movement/mob.dm
```diff
@@ -178,7 +178,9 @@
 	//We are now going to move
 	mob.moving = 1
 
-	direction = mob.AdjustMovementDirection(direction)
+	if (mob == mover)
+		direction = mob.AdjustMovementDirection(direction)
```
afterthought2: I think the check needs to be inside this proc, not outside it--it's called from several places and might plausibly contain other logic. That might mean you need to pass mover as an arg to the proc.

Gaxeer: Yes, you are right. I've made adjustments

code/datums/movement/mob.dm
```diff
@@ -25,9 +25,9 @@
 	mob.ghostize()
 
 // Incorporeal/Ghost movement
-/datum/movement_handler/mob/incorporeal/DoMove(var/direction)
+/datum/movement_handler/mob/incorporeal/DoMove(var/direction, var/mob/mover)
```
Gaxeer, line 28: forgot to pass mover, I suppose

Gaxeer: No, that's fine. But I am confused, why do we need to AdjustMovementDirection for ghosts. They can be shocked or drunk :D?

MistakeNot4892: Doubledead

MistakeNot4892 (approved):

PR #2427, opened 2022-05-23
Lowered gibbing throwforce.

### LINKED ISSUE #2306: Limbs thrown when dismembering a mob with a melee weapon can hit you and throw you away
#### Description of issue
When dismembering a mob, if you stand in the path of the limb (~~stand on top for example~~ standing one tile next) you get kicked around with a lot of force. You get knocked down and if you hit a wall, you can actually cause some brain damage and "YOU DON'T KNOW WHICH WAY IS FORWARD" to happen.

Chopping off a limb and getting hit with it definitely shouldn't send you flying with that amount of force.

#### Steps to reproduce
* Take a weapon that does edge damage
* Spawn a human, preferably in a tight corridor
* Stand on the tile next to the mob as its laying down
* Attack the limb zone on the human until the limb flies off and it hits you
* Notice how it hits you, and sends you flying

#### Issue bingo

MistakeNot4892:
## Description of changes
Lowers throwforce of gibs to be about on par with throwing something from your hand.

Resolves #2306.
Should prevent gibbed bodies from damaging walls and doors.

## Authorship
Myself.

## Changelog
:cl:
tweak: Gibbed bodies now throw the limbs around with less force.
/:cl:

### DIFF
--- a/code/__defines/items_clothing.dm
+++ b/code/__defines/items_clothing.dm
@@ -147,7 +147,7 @@
 #define FIRE_MAX_STACKS           25
 #define FIRE_MAX_FIRESUIT_STACKS  20 // If the number of stacks goes above this firesuits won't protect you anymore. If not, you can walk around while on fire like a badass.
 
-#define THROWFORCE_GIBS 10             // Throw speed for gibbed or dismembered organs.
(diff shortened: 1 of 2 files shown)


afterthought2 (approved):

Gaxeer:
My favourite bug. So much fun I had with thouse killing limbs:D

PR #2431, opened 2022-05-24
cached_json_code() returns a copy to avoid cache mutation.

MistakeNot4892:
## Description of changes
`cached_json_decode()` now returns a copy of any cached json.

Previously anything messing with the returned json would mutate the cache, which is undesirable.

## Authorship
Myself.

## Changelog
Nothing player-facing.

### DIFF
--- a/code/_helpers/lists.dm
+++ b/code/_helpers/lists.dm
@@ -767,12 +767,15 @@ proc/dd_sortedObjectList(list/incoming)
 
 var/global/list/json_cache = list()
 /proc/cached_json_decode(var/json_to_decode)
-	if(!json_to_decode || !length(json_to_decode))
-		return list()
-	try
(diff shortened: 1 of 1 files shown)


code/_helpers/lists.dm
```diff
@@ -772,7 +772,8 @@ var/global/list/json_cache = list()
 	try
 		if(isnull(global.json_cache[json_to_decode]))
 			global.json_cache[json_to_decode] = json_decode(json_to_decode)
-		. = global.json_cache[json_to_decode]
+		var/list/decoded = global.json_cache[json_to_decode]
+		return decoded?.Copy() || list() // To prevent cache mutation.
```
afterthought2: Not an expert on the system, but aren't there valid nonlist outputs of `json_decode`, like a single string, number, or bool for example?

afterthought2 (approved):

PR #2432, opened 2022-05-25
SSmapping rework.

MistakeNot4892:
## Description of changes
- Reorders overmap init and moves it onto SSmapping.
- Tweaks overmap creation a bit to allow for easier downstream overrides.
- Removes the map_template id var and instead uses a parent type and the name var. Checks uniqueness.
- Map templates now use their name as their unique identifier and should always hit the SSmapping cache instead of creating templates ad hoc.
- SSmapping has been reworked a bit. .dmm blacklists have been unified in a single JSON list.
- Ruin templates now use a category instead of typechecking.
- Removes unused ruins landmark and global list (also unused on Bay?)
- Makes the majority of template stuff type agnostic.
- Moves templates into the existing module folder for mapping stuff.
- Adds a unit test for special role base maps.

## TODO
- ~~Test that everything is working.~~ No runtimes, templates all seem present in the list, ruins are spawning on the exoplanet.
- ~~Move abstract template types into a modules folder of their own.~~ `code/modules/maps` already existed, have relocated them there.
- ~~Work out why cost sorting was done during init and readd a sort where appropriate.~~ No idea. Cost is checked elsewhere for the practical use, it was just sorting the type list for no reason.

Cleaner, saner code. Cleaner, saner map templates. Easier to convert them to hotloaded template form down the track.

## Authorship
Myself.

## Changelog
Nothing player-facing.


MistakeNot4892:
Helping thrash out some issues with Lohi's PRs using this one, DNM till those are in.

MistakeNot4892:
This is ready to go.

MistakeNot4892:
Pls review.

code/controllers/subsystems/mapping.dm
```diff
@@ -4,71 +4,76 @@ SUBSYSTEM_DEF(mapping)
...
+	for(var/temple_cat in map_template.template_categories) // :3
+		LAZYINITLIST(map_templates_by_category[temple_cat])
+		LAZYSET(map_templates_by_category[temple_cat], map_template.name, map_template)
+	return TRUE
 
-	if (!banned_exoplanet_dmms || !banned_space_dmms || !banned_away_site_dmms)
-		report_progress("One or more map blacklist files are not present in the config directory!")
+/datum/controller/subsystem/mapping/proc/validate_map_template(var/datum/map_template/map_template)
+	if(!istype(map_template))
+		return FALSE
```
afterthought2, line 59: May as well crash here.

code/controllers/subsystems/mapping.dm
```diff
@@ -4,71 +4,76 @@ SUBSYSTEM_DEF(mapping)
...
+		PRINT_STACK_TRACE("Duplicate map name '[map_template.name]' on type [map_template.type]!")
+		return FALSE
+	return TRUE
+	
+/datum/controller/subsystem/mapping/proc/get_all_template_instances()
+	. = list()
+	for(var/template_type in subtypesof(/datum/map_template))
+		var/datum/map_template/template = template_type
+		if(initial(template.template_parent_type) != template_type && initial(template.name))
+			. += new template_type(name) // send name as a param to catch people doing illegal ad hoc creation
```
afterthought2: You're sending our name, i.e. the controller's? I don't get it.

EDIT: send either the controller itself, its type path, or a string literal instead; seems unreasonable to tie this to its name.

code/game/antagonist/antagonist.dm
```diff
@@ -86,8 +86,8 @@
 		Think through your actions and make the roleplay immersive! <b>Please remember all \
 		rules aside from those without explicit exceptions apply to antagonists.</b>"
 	
-	// Map template that antag needs to load before spawning. Nulled after it's loaded.
-	var/datum/map_template/base_to_load
+	// Map template name that antag needs to load before spawning. Nulled after it's loaded.
+	var/base_to_load
```
afterthought2, line 90: Not loving this change; paths are checked (for existence at least) at compile time, here you can make a typo and have it fail.

code/game/antagonist/antagonist_add.dm
```diff
@@ -2,14 +2,15 @@
...
-		if(SSmapping.map_templates[initial(base.name)])
-			base = SSmapping.map_templates[initial(base.name)]
-		if(!base)
-			base = new base()
-		else if(base.loaded > 0)
+		if(istext(base))
+			base = SSmapping.get_template(base)
+		if(!istype(base))
+			log_warning("Map template '[base]' could not be found, skipping load for [name].")
+			return FALSE
```
afterthought2, line 12: I'd crash here so it fails unit tests. I know you have a unit test below which may also check this, though.

code/modules/maps/_map_template_unit_testing.dm
```diff
@@ -9,8 +9,8 @@
 	var/list/area_coherency_test_exempt_areas = list()
 	var/list/area_coherency_test_subarea_count = list()
 
-/datum/map_template/New(var/list/paths = null, var/rename = null)
-	..()
+/datum/map_template/New(var/created_ad_hoc)
+	..(created_ad_hoc)
```
afterthought2, line 13: Why? Surely `..()` doesn't care about this var.

code/modules/maps/helper_landmarks.dm
```diff
@@ -1,25 +1,30 @@
...
-/obj/abstract/landmark/map_load_mark/non_template/Initialize()
-	. = ..()
-	LAZYADD(SSmapping.compile_time_map_markers, src)
+	forceMove(null)
+	if(istype(spawn_loc))
+		if(istext(template))
+			template = SSmapping.get_template(template)
+		if(istype(template))
+			template.load(spawn_loc, TRUE)
+	if(!QDELETED(src))
```
afterthought2: I feel that if you want to run a qdeleted check, you should do it earlier, e.g. before moving yourself.

Also, do these actually work? I'd think you should be failing the atoms slept in Initialize criteria on them.

code/controllers/subsystems/mapping.dm
MistakeNot4892: It doesn't matter what it sends, the point is just for a bare New() to not have the same value.

code/game/antagonist/antagonist.dm
MistakeNot4892: The point of this change is to allow moving templates off paths and into JSON. Using types means we're limited to subtyping for creating template variants and makes it harder or at least more convoluted to hotload from JSON files or the equivalent in the future. The unit tests I've added plus the away site testing build should handle the validation aspect.

code/modules/maps/helper_landmarks.dm
MistakeNot4892: Added a `waitfor = FALSE` and moved the qdel up before the template is loaded.

code/game/antagonist/antagonist_add.dm
MistakeNot4892: Added a `PRINT_STACK_TRACE` instead of a `log_warning`. Unit test will fail already unless this is happening midround due to varediting.

code/controllers/subsystems/mapping.dm
MistakeNot4892: Passing type in instead.

code/controllers/subsystems/mapping.dm
MistakeNot4892: Added a stack trace - I don't think crashing during a validation proc is necessarily helpful.

afterthought2 (approved):
Can't decide if making it waitfor FALSE makes it better or worse.

PR #2434, opened 2022-05-26
#2325 with tweaks.

MistakeNot4892:
## Description of changes
Commit from #2325 with the requested changes. Quard seems to be gone from the Discord so unsure if he will implement them, figured I'd do it myself.

Character preview in prefs is larger and cleaner, and pAIs get their own preview when Global is selected.

## Authorship
@quardbreak for the main changes, I just tweaked a couple things.

## Changelog
:cl:
tweak: Character preview is a bit bigger and nicer.
/:cl:

### DIFF
--- a/code/modules/client/preference_setup/global/03_pai.dm
+++ b/code/modules/client/preference_setup/global/03_pai.dm
@@ -2,10 +2,7 @@
 	name = "pAI"
 	sort_order = 3
 
-	var/icon/pai_preview
 	var/datum/paiCandidate/candidate
-	var/icon/bgstate = "steel"
-	var/list/bgstate_options = list("FFF", "steel", "white")
 
 /datum/category_item/player_setup_item/player_global/pai/load_preferences(datum/pref_record_reader/R)
 	if(!candidate)
@@ -30,8 +27,7 @@
 	if(!candidate)
 		candidate = new()
 
-	if (!pai_preview)
-		update_pai_preview(user)
+	update_pai_preview()
 
 	. += "<b>pAI:</b><br>"
 	if(!candidate)
@@ -43,8 +39,6 @@
 	. += "OOC Comments: <a href='?src=\ref[src];option=ooc'>[candidate.comments ? TextPreview(candidate.comments, 40) : "None Set"]</a><br>"
 	. += "<div>Chassis: <a href='?src=\ref[src];option=chassis'>[candidate.chassis]</a><br>"
 	. += "<div>Say Verb: <a href='?src=\ref[src];option=say'>[candidate.say_verb]</a><br>"
-	. += "<table><tr style='vertical-align:top'><td><div class='statusDisplay'><center><img src=pai_preview.png width=[pai_preview.Width()] height=[pai_preview.Height()]></center><a href='?src=\ref[src];option=cyclebg'>Cycle Background</a></div>"
-	. += "</td></tr></table>"
 
 /datum/category_item/player_setup_item/player_global/pai/OnTopic(var/href,var/list/href_list, var/mob/user)
 	if(href_list["option"])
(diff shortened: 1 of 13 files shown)


code/modules/client/preference_setup/preference_setup.dm
```diff
@@ -55,6 +55,8 @@ var/global/const/CHARACTER_PREFERENCE_INPUT_TITLE = "Character Preference"
 	sort_order = 10
 	category_item_type = /datum/category_item/player_setup_item/player_global
 
+/datum/category_group/player_setup_category/global_preferences/selected(var/datum/preferences/prefs)
+	prefs.update_pai_preview()
```
afterthought2, line 59: This call seems illegal (no args, so it still makes a blank mutable appearance and does something with it).

code/modules/client/preference_setup/preference_setup.dm
```diff
@@ -115,9 +117,10 @@ var/global/const/CHARACTER_PREFERENCE_INPUT_TITLE = "Character Preference"
 		return 1
 
 	if(href_list["category"])
-		var/category = locate(href_list["category"])
-		if(category && (category in categories))
+		var/datum/category_group/player_setup_category/category = locate(href_list["category"])
+		if(istype(category) && (category in categories))
 			selected_category = category
+			category.selected(user.client.prefs)
```
afterthought2, line 123: I think `user.client.prefs` is actually wrong; these retain a ref to their prefs datum.

code/modules/client/preference_setup/preference_setup.dm
```diff
@@ -248,7 +254,7 @@ var/global/const/CHARACTER_PREFERENCE_INPUT_TITLE = "Character Preference"
 	if(!pref_mob || !pref_mob.client)
 		return 1
 	if(. & TOPIC_UPDATE_PREVIEW)
-		pref_mob.client.prefs.update_preview_icon()
+		pref_mob.client.prefs.update_character_preview_map()
```
afterthought2, line 257: Again, should just be the instance's held prefs ref.

code/modules/mob/new_player/preferences_setup.dm
```diff
@@ -98,12 +98,21 @@
...
-/datum/preferences/proc/update_preview_icon()
+/// Updates character mannequin without rebuilding the inventory or overlays of the mannequin.
+/datum/preferences/proc/refresh_character_preview()
+	var/mob/living/carbon/human/dummy/mannequin/mannequin = get_mannequin(client_ckey)
+	if(mannequin)
+		update_character_preview_mannequin(new /mutable_appearance(mannequin))
+
+/// Updates character preview map.
+/// If you don't need to update renders, use [/datum/preferences/proc/update_character_preview_background].
+/datum/preferences/proc/update_character_preview_map()
```
afterthought2, line 109: Why rename this? It's confusing; this proc doesn't actually do anything with the map control that any of the other split out ones don't do; the main thing it does is the expensive step of deleting and rebuilding the mannequin inventory.

The number of procs involved in this whole business feels needlessly confusing.

afterthought2 (commented):
I can't actually follow what's happening in the code relative to when the pAI vs. the mob is shown to the user because of how it's set up: whichever call you make last will win, so it just depends on the ordering of the calls. As such, I have no idea if this is logically sound.

I think it would be much saner to have a single call (kind of like we do in atom code with `update_icon`) that you make to correctly display whatever should be shown based on prefs state, be that the pAI, the mannequin, nothing, whatever. It might be coupled with possibly one (1) auxiliary public function that's used to rebuild the preview mannequin (i.e. delete and rebuild the inventory) to alleviate performance concerns.

afterthought2 (changes requested):
Actually I understated my objections in the previous comment. Looking over this more carefully and testing locally, the _only reason it actually works_ is the pAI update call in the html content getter for the global prefs tab. This should not even be there; that proc is not a refresh proc and should not be messing with client state (and defeats the purpose of the system). It also obfuscates that some of the other refresh calls are broken (see my comment), it obfuscates the fact that some calls from the global state menu will currently rebuild the mannequin, display the mannequin briefly (because of the way prefs/Topic and base Topic are set up, and it's visible to me even testing locally with no latency), and then replace that with the pAI appearance afterwards when the callstack arrives at that `contents` proc.

code/modules/client/preference_setup/global/03_pai.dm
```diff
@@ -30,8 +27,7 @@
 	if(!candidate)
 		candidate = new()
 
-	if (!pai_preview)
-		update_pai_preview(user)
+	pref.update_pai_preview(candidate.chassis)
```
afterthought2, line 30: This is what I was objecting to; this does not belong in this proc.

MistakeNot4892:
Honestly I was adopting this because I thought it was alright, I'm not willing to revise it to the extent requested.

PR #2408, opened 2022-05-18
Stole a bunch of tweaks from Naxodile's wyverm branch.

MistakeNot4892:
## Description of changes
Not sure if I can make these associated with @Naxodile to avoid stealing authorship, but in any case these are ganked from their wyverm branch. I did a reset and individual commit after a large rebase, hence new commits.

## Why and what will this PR improve
- Implements missing voidsuit refit behavior for icons.
- Changes AI core spawn to use a landmark.
- Removes a redundant organ check from defibrilators.
- Tweaks medicine paths (adrenaline to stabilizer, etc) in a few spots.
- Handful of grammar/name tweaks.
- Adds a skill_transferrable check to prefs setup.
- Tweaks forensics to use SKILL_BASIC for taking samples.
- Fixes metal tile recipe creation.
- Restricts monkey language.
- Prevents prosthetic chests and groins becoming necrotic (ORGAN_DEAD).
- Adds air sensors to the RUST field exception list.
- Fixes beaker overlay positions on chem dispensers.
- Adds a global map setting for shuttle atmos checking in CI.

## Authorship
@Naxodile other than a few cleanup/tweak edits.

## Changelog
:cl:
tweak: Forensic samples now require basic skill.
tweak: Some medical storage contents have been tweaked to be more appropriate.
/:cl:

PR #2436, opened 2022-05-26
fix: put beaker in hands, when ejected from ChemDispenser, or drop if used by silicon

Gaxeer:
## Description of changes

Put beaker in hands instead of drop to ChemDispenser turf, when ejecting from it

QoL

## Authorship

@Gaxeer 

## Changelog

:cl:
tweak: put beaker in hands instead of drop to ChemDispenser turf, when ejecting from it
/:cl:

### DIFF
--- a/code/modules/reagents/dispenser/dispenser2.dm
+++ b/code/modules/reagents/dispenser/dispenser2.dm
@@ -30,7 +30,7 @@
 	var/static/list/acceptable_containers = list(
 		/obj/item/chems/glass,
 		/obj/item/chems/condiment,
-		/obj/item/chems/drinks		
+		/obj/item/chems/drinks
 	)
 
 /obj/machinery/chemical_dispenser/Initialize(mapload, d=0, populate_parts = TRUE)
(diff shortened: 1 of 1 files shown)


afterthought2 (commented):
Not sure about the random x thing.

code/modules/reagents/dispenser/dispenser2.dm
```diff
@@ -181,7 +181,7 @@
 	else if(href_list["ejectBeaker"])
 		if(container)
 			var/obj/item/chems/B = container
-			B.dropInto(loc)
+			user.put_in_hands(B)
```
afterthought2, line 187: This is actually problematic. If the user is not near the machine, this will teleport the item to their location and leave it there. Since this is in Topic, there's no control over where the user is. It leads to common borg and AI bugs. You need to check `CanPhysicallyInteract` if you want to do this.

MistakeNot4892:
Random overlay stuff is fixed in #2408

Gaxeer:
> Random overlay stuff is fixed in #2408

Okay. Thanks

code/modules/reagents/dispenser/dispenser2.dm
Gaxeer: Understood

afterthought2:
This is better, but now it just does nothing if you cannot physically interact. It should likely `dropInto(loc)` in that case. Borgs can interact with the UI within view, and it generally makes sense to have the machine respond to the eject command even if not adjacent. If you think that's not the case, you probably need to modify the UI to grey out the eject button if physical interaction is impossible (several ways to do that of varying levels of sophistication).

Gaxeer:
@afterthought2 
image
It's alredy `grey`

afterthought2:
It's grey because there's no beaker loaded.

MistakeNot4892 (approved):

PR #2453, opened 2022-05-28
Removes powered() checks

afterthought2:
## Description of changes

Removes a bunch of `powered()` checks. Reworks how lights/lightswitches work a bit.

You should (almost) never use this proc to check for whether a machine is powered, the correct thing to do is to just check the flag. Some of the code touched here is kind of crazy.

This is pointing at dev because while maybe using this proc is a bug (depending on your perspective), some of the changes are substantial.

## Authorship

me


code/modules/power/lighting.dm
```diff
@@ -420,8 +420,14 @@
 
 // called when area power state changes
 /obj/machinery/light/power_change()
-	spawn(10)
-		seton(powered())
+	. = ..()
+	if(.)
+		delay_and_set_on(expected_to_be_on(), 1 SECOND)
```
afterthought2, line 425: I decoupled this interaction (power status changes, we turn lights on/off) from lightswitches turning lights on/off (they used to basically force the whole area to recompute power status, and then that would fall through to this proc and do the work).

I have no idea why this and the lightswitch interaction wait 1 second. If it's not desirable in one or both cases, I can easily make this happen immediately.

MistakeNot4892: Seems fine to me, gives a bit of tension, keeps behavior equivalent. Easy to remove if someone wants to down the track.

MistakeNot4892 (approved):

code/modules/power/lighting.dm
PsyCommando: iirc the wait is just to simulate neon lights powering up to match the sound effect.

PR #2458, opened 2022-05-28
Partially reverts vote UI changes.

afterthought2:
## Description of changes

Partially reverts #150 to resolve formatting issues for non-admins. Makes a couple of other fixes. The vote obfuscation/highlighting and anything that exposes the raw weighted vote totals is reverted; the per-datum time option and datum/browser usage is kept.

Should now be formatted properly for both admins and non-admins. If people really love the obfuscation I can keep it, but I feel that (a) the way the coloring is done is illogical and (b) exposing raw weighted vote totals is bad practice.

Can be pointed at dev instead.

## Authorship

me


MistakeNot4892 (approved):
Seems fine to me, I did find the numbers attached to the votes a bit weird.

PR #2459, opened 2022-05-28
experimental: Safe pre-Initialize() qdel

Lohikar:
This adds a separate `EarlyDestroy()` proc, and allows atoms to be qdeleted before they have initialized.

Uninitialized atoms will call `EarlyDestroy()` instead of `Destroy()` -- overriding this proc should only be necessary if that atom is doing something in `New()` _and is not using INITIALIZE_IMMEDIATE_.

Additional changes:
- Turfs also now short-circuit most of `ChangeTurf()` if they haven't been initialized yet thus have no state to copy.
- Removes the `AdjustInitializeArguments()` SSatoms hook.
- Removes the unused `turf_changed` event.


MistakeNot4892:
Pulled this into https://github.com/NebulaSS13/Nebula/pull/2432 while testing and it addresses the pre-SSatoms runtimes/warnings with no apparent problems presented.

MistakeNot4892:
Testing this on my other branches revealed no apparent problems, I would be happy for this to go in.

MistakeNot4892 (dismissed):

MistakeNot4892 (approved):

deathride58 (approved):

PR #2466, opened 2022-05-30
Airlock panel opens with wrench

SolatK:
## Description of changes

Airlock panel opens with wrench instead crowbar

to open the panel with a crowbar, the airlock had to be welded or boled. Wrench doesn't have this problem.
## Authorship

me
## Changelog

:cl:
tweak: Airlock components panel now opens with wrench
/:cl:

### DIFF
--- a/code/game/machinery/_machines_base/machine_construction/airlock.dm
+++ b/code/game/machinery/_machines_base/machine_construction/airlock.dm
@@ -27,7 +27,7 @@
 /decl/machine_construction/default/panel_closed/door/mechanics_info()
 	. = list()
 	. += "Use a screwdriver to open a small hatch and expose some logic wires."
-	. += "Use a crowbar on the secured (bolted, welded or braced) airlock to pry open the main cover."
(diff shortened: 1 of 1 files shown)


afterthought2:
Need to also modify the mechanics text below.

The issue with this kind of interaction is that you can't do the reverse, closing the panel with a wrench, because that does something else.

SolatK:
> The issue with this kind of interaction is that you can't do the reverse, closing the panel with a wrench, because that does

the components panel is closed with a screwdriver before my change and after

SolatK:
> 



> Need to also modify the mechanics text below.

"You open the main cover panel on \the [machine], exposing the internals." to "You unwrench locking bolts and open the main cover panel on \the [machine], exposing the internals."?

afterthought2:
No, there's a proc below, `mechanics_info`, which explains what tool to use for what.

comma (approved):
Sure, it was requiring welding etc just to mirror old deconstruction steps

PR #2470, opened 2022-05-31
Adjust suit refitting code to be less broken.

MistakeNot4892:
## Description of changes
- Refitting now expects both bodytype and body flags.
- Refitting now copies sprite_sheets.
- Refitting now only sets icons if the icon has a world/inventory state to use.
- Removes a merge artifact.

Accidentally included some fixes:
- Allows girders to be built from wood, bone and plastic.
- Exterior turfs now count as floors for the purposes of movement and crafting.

Existing code is busted all to heck.

## Authorship
Myself.

## Changelog
Nothing player-facing.

### DIFF
--- a/code/game/machinery/suit_cycler.dm
+++ b/code/game/machinery/suit_cycler.dm
@@ -43,7 +43,7 @@
 	)
 
 	//Bodytypes that the suits can be configured to fit.
-	var/list/available_bodytypes = list(BODYTYPE_HUMANOID = BODY_FLAG_HUMANOID)
+	var/list/available_bodytypes = list(BODYTYPE_HUMANOID)
 
 	var/decl/item_modifier/target_modification
 	var/target_bodytype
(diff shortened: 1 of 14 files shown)


MistakeNot4892:
Still need to test this.

MistakeNot4892:
Ready to go.

code/game/machinery/suit_cycler.dm
```diff
@@ -458,21 +458,15 @@
...
-	var/target_flags = available_bodytypes[target_bodytype]
-	if(helmet) helmet.refit_for_bodytype(target_flags)
-	if(suit)   suit.refit_for_bodytype(target_flags)
-	if(boots)  boots.refit_for_bodytype(target_flags)
-
 
 	if(helmet)
 		target_modification.RefitItem(helmet)
 		helmet.refit_for_bodytype(target_bodytype)
+		helmet.SetName("refitted [helmet.name]")
```
afterthought2, line 465: If I refit it repeatedly won't this give me `refitted refitted blah` and so on? Maybe `initial`, or there's some fancy extension thing I believe.

MistakeNot4892: `RefitItem` already sets the item name to a base name.

afterthought2: In that case, should probably change boots below, which, strangely, are not copy-paste but use `initial`.

MistakeNot4892: Magboots are not visually refitted currently as there are no voidsuit-specific booties.

afterthought2 (approved):

PR #2481, opened 2022-06-03
Makes dmm reader use ChangeArea

afterthought2:
## Description of changes

dmm suite changes the area of turfs being loaded over (some of the time). When it does, it changes area directly without firing any of the listeners currently located in ChangeArea (and is, I believe, the only remaining thing which does this).

ChangeArea doesn't really do anything itself, it just fires a lot of listeners and hooks. This might have secondary effects, but it's unclear what if any. It does come at some cost.

Power machines require perfectly accurate notification of area changes. In this case, loading one template over another in a way which changes area breaks power tracking. This is really uncommon; almost nothing does this by default apart from a weird edge case with exoplanent ruins and artifact finds. It's possible to do with adminbus, though.

## Authorship

me

### DIFF
--- a/code/modules/maps/reader.dm
+++ b/code/modules/maps/reader.dm
@@ -338,7 +338,7 @@ var/global/dmm_suite/preloader/_preloader = new
 			instance = new atype(null)
 			initialized_areas_by_type[atype] = instance
 		if(crds)
-			instance.contents.Add(crds)
+			ChangeArea(crds, instance)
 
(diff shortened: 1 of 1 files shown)


MistakeNot4892 (approved):

afterthought2:
I'm open to the theory that this change should actually be gated to only trigger if `no_changeturf` is `FALSE`. Unlike ChangeTurf there are no obvious secondary effects to ChangeArea, but perhaps there are legitimate consistency arguments (and it would be more performant).

PR #2482, opened 2022-06-03
Overmap movement is now handled by a rapid processing subsystem.

MistakeNot4892:
## Description of changes
`SSovermap` is a processing ticker that runs five times per second, and moving overmap objects use it to move. I am sure someone will tell me this is an absolutely miserable idea so please let me know what you'd prefer. The actual rate of moment is adjusted to the previous wait delay so there should be no overall impact on how fast the ship actually moves.

Example of how it looks:
https://cdn.discordapp.com/attachments/492799900145418270/982256412946214982/dreamseeker_n0VfGbAgrq.mp4

Currently the overmap ticks once per second using SSobj, which was fine for the previous turf based system, but is unbearably stodgy for pixel movement. This makes movement feel much more fluid and responsive. It is also prep work for down the track when the overmap can be changed to use pixel adjacency/collision for events so careful maneuvering is possible.

~~Depends on #2204.~~ No more deps.

## Authorship
Myself.

## Changelog
:cl:
tweak: Overmap position updates ten times faster now. Exciting.
/:cl:


code/controllers/subsystems/processing/overmap.dm
```diff
@@ -0,0 +1,5 @@
+PROCESSING_SUBSYSTEM_DEF(overmap)
+	name = "Overmap"
+	priority = SS_PRIORITY_OVERMAP
+	wait = 1
```
Lohikar: This is too fast. If you want to make things smooth, find a way to do it with `animate()`.

Lohikar (dismissed):

code/controllers/subsystems/processing/overmap.dm
MistakeNot4892: Smoothness relates to the stepping of position, not the visual offsets. Animate isn't going to help here. A lot of the time individual steps are less than a pixel.

MistakeNot4892: Animate also works very strangely and inconsistently with client offsets when tried previously.

code/controllers/subsystems/processing/overmap.dm
```diff
@@ -0,0 +1,25 @@
...
+/datum/controller/subsystem/overmap/fire(resumed = 0)
+	if(!resumed)
+		src.current_run = moving_entities.Copy()
+	var/list/current_run = src.current_run
+	var/wait = src.wait
+	var/times_fired = src.times_fired
+	while(current_run.len)
+		var/obj/effect/overmap/entity = current_run[current_run.len]
+		current_run.len--
+		if(QDELETED(entity) || entity.ProcessOvermap(wait, times_fired) == PROCESS_KILL)
```
Lohikar, line 22: Does `ProcessOvermap` have sanity checks for being called on a stationary object?

code/modules/overmap/overmap_object.dm
```diff
@@ -113,8 +127,28 @@ var/global/list/overmap_unknown_ids = list()
...
+
+/obj/effect/overmap/proc/update_moving()
+	if(is_still())
+		if(is_moving)
+			is_moving = FALSE
+			SSovermap.moving_entities -= src
+	else
+		if(!is_moving)
+			is_moving = TRUE
+			SSovermap.moving_entities[src] = TRUE
```
Lohikar: This list is deduplicated via `is_moving` already, why assoc?

MistakeNot4892:
Need to retest this when I am at home and not on a laptop at work.

code/modules/overmap/overmap_object.dm
MistakeNot4892: `is_moving` was an artifact of using a processing subsystem, I've changed it to just check list presence now instead.

code/controllers/subsystems/processing/overmap.dm
MistakeNot4892: It doesn't need them, a stationary object will return PROCESS_KILL or simply not move.

MistakeNot4892:
````
/obj/effect/overmap/proc/ProcessOvermap(wait, tick)

	if(halted || is_still())
		return PROCESS_KILL

	if(!can_move)
		return
````

**Comment by MistakeNot4892** [comment]
All working.

File: code/modules/overmap/overmap_object.dm [hunk 4]
```diff
@@ -128,26 +154,40 @@ var/global/list/overmap_unknown_ids = list()
...
+			// Can probably be tidied up at some point but math is spooky.
+			position[i] += (deltas[i] > 0) ? -1 : 1
+
+	if(moved)
 		var/turf/newloc = locate(x + deltas[1], y + deltas[2], z)
 		if(newloc && loc != newloc)
 			Move(newloc)
 			handle_wraparound()
-	handle_overmap_pixel_movement()
+		update_icon()
```
afterthought2: Is the reason you are doing this that the underlay being applied here has its scaling transform overridden by the pixel shifts? If so, can you use `KEEP_TOGETHER` or some other method instead?

afterthought2 (commented):
Looked through it as Lohikar hasn't touched it in a while.

code/modules/overmap/overmap_object.dm
MistakeNot4892: I believe this is called here because adjacency radius may have changed since the last call, so the underlay needs to be rescaled. I'll see if I can break that logic out into a more event driven setup.

MistakeNot4892: Adjacency radius is unused, I have no idea why this logic was put in here.

MistakeNot4892:
This is ready to go.

afterthought2 (approved):

PR #2483, opened 2022-06-04
Allow removing/building some props

PsyCommando:
## Description of changes
* Reworked /obj/structure/flora a bit, and made plants destructible and able to be dismantled for materials with the right tool (axe). 
* Merged /obj/structure/flora/ausbushes with /obj/structure/flora/bush since they're really the same dang thing.
* Added 5 sound effects for interacting with the plants (aka destroying them). All of them are CC0 or public domain, with their details in their file's tags. (All from freesound.org, with minor tweaks)
* Made tree stumps icons for all trees.
* Made a generic plant remains cleanable icon used when smashing plants to pieces.
* Made the mundane fountain buildable and able to be dismantled, with the proper dropped materials and etc.. Also made it spawn on mapload with water inside, otherwise it spawns empty. The idea would be to make the sprite reflect what's in the fountain reagent-wise. But, I'm not entirely sure how to go about that yet.
* Made the bedsheet bin less buggy, made it able to be disassembled, gave it a material, and made it give more feedback.

## Changelog

:cl:
tweak: Generic plants can now be cut. Trees can be chopped down with an axe.
soundadd: Added several sound effects when dealing with generic plants.
/:cl:


code/game/objects/structures/bedsheet_bin.dm
```diff
@@ -77,49 +78,87 @@ LINEN BINS
...
+	var/obj/item/hidden         //Object hidden amidst the bedsheets
+
+/obj/structure/bedsheetbin/mapped/Initialize(ml, _mat, _reinf_mat)
+	stored = max_stored //Mapped ones start with some unspawned sheets
+	. = ..()
 
-/obj/structure/bedsheetbin/examine(mob/user)
+/obj/structure/bedsheetbin/dump_contents()
+	//Dump all sheets, even unspawned ones
+	for(var/i = 0, i < get_amount(), i++)
```
MistakeNot4892: `for(var/i = 1 to get_amount())` is faster in DM.

code/game/objects/structures/__structure.dm
```diff
@@ -134,8 +134,9 @@
 		last_damage_message = 0.75
 
 /obj/structure/physically_destroyed(var/skip_qdel)
-	. = ..(TRUE)
-	dismantle()
+	if(!(. = ..(TRUE)))
```
MistakeNot4892: Why?
Adding a `if(.)` before the `dismantle()` would have done the same job.

code/game/objects/structures/bedsheet_bin.dm
PsyCommando: oh, gotcha, will fix that.

code/game/objects/structures/__structure.dm
PsyCommando: Just to annoy lohikar tbh :P
But, yeah, not really any reasons. It just fits on one line.
Looking at it now, I'm not even really using the return value anyways, so I probably could just do if(!..(TRUE))

code/game/objects/structures/__structure.dm
```diff
@@ -134,8 +134,9 @@
 		last_damage_message = 0.75
 
 /obj/structure/physically_destroyed(var/skip_qdel)
-	. = ..(TRUE)
-	dismantle()
+	if(!..(TRUE))
```
MistakeNot4892:
````
. = ..(TRUE)
if(.)
  return dismantle()
````
or
````
if(..(TRUE))
  return dismantle()
````

File: code/game/objects/structures/seaweed.dm [hunk 4]
```diff
@@ -29,3 +29,9 @@
 	desc = "Damp and mossy plant life."
 	icon_state = "lichen"
 	icon = 'icons/obj/structures/plants.dmi'
+
+/obj/effect/decal/cleanable/lichen/attackby(obj/item/I, mob/user)
+	if(I.sharp && I.force > 1)
+		qdel(src)
+		return
```
MistakeNot4892: `return TRUE`

code/game/objects/structures/pit.dm
```diff
@@ -54,7 +54,14 @@
...
-	for(var/atom/movable/A in src.loc)
+
+	//If we close the pit without anything inside, just leave the soil undisturbed
+	var/turf/T = get_turf(src)
+	if(!length((T.contents - src)))
+		qdel(src)
+		return
+
+	for(var/atom/movable/A in T)
 		if(!A.anchored && A != user)
```
MistakeNot4892: Probably should check `!A.simulated` as well.

code/game/objects/structures/fountain.dm
```diff
@@ -59,10 +59,23 @@
...
+	w_class                = ITEM_SIZE_STRUCTURE
+	material               = /decl/material/solid/stone/marble 
+	used                   = TRUE
+
+/obj/structure/fountain/mundane/Initialize(ml, _mat, _reinf_mat)
+	. = ..()
+	atom_flags |= ATOM_FLAG_OPEN_CONTAINER | ATOM_FLAG_CLIMBABLE
+	setup_reagents(ml)
+
+/obj/structure/fountain/mundane/proc/setup_reagents(var/should_fill = TRUE)
```
MistakeNot4892, line 75: Why is this a separate proc rather than just `if(ml)` in Initialize()?

code/game/objects/structures/bedsheet_bin.dm
```diff
@@ -77,49 +78,87 @@ LINEN BINS
...
+/obj/structure/bedsheetbin/dump_contents()
+	//Dump all sheets, even unspawned ones
+	for(var/i = 1 to get_amount())
+		remove_sheet()
 	. = ..()
 
-	if(amount < 1)
+/**Returns the total amount of sheets contained, including unspawned ones. */
+/obj/structure/bedsheetbin/proc/get_amount()
+	return LAZYLEN(sheets)? stored + length(sheets) : stored
```
MistakeNot4892: You could just return `stored + length(sheets)` since `+0` is valid.

code/game/objects/structures/flora.dm
```diff
@@ -1,204 +1,320 @@
...
+	w_class                = ITEM_SIZE_NORMAL                       //Size determines material yield
+	material               = /decl/material/solid/plantmatter       //Generic plantstuff
+	tool_interaction_flags = 0 
+	hitsound               = 'sound/effects/hit_bush.ogg'
+	var/tmp/snd_cut        = 'sound/effects/plants/brush_leaves.ogg' //Sound to play when cutting the plant down
+	var/remains_type       = /obj/effect/decal/cleanable/plant_bits //What does the plant leaves behind in addition to the materials its made out of. (part_type is like this, but it drops instead of materials)
+
+/obj/structure/flora/Initialize(ml, _mat, _reinf_mat)
+	. = ..()
+	init_appearance()
```
MistakeNot4892, line 17: Is this separate from Initialize() for serialization purposes downstream? Seems like an odd change otherwise.

code/game/objects/structures/flora.dm
```diff
@@ -1,204 +1,320 @@
...
+/**What to do when the can_cut_down check returns true. Normally simply calls dismantle. */
+/obj/structure/flora/proc/cut_down(var/obj/item/I, var/mob/user)
+	if(snd_cut)
+		playsound(src, snd_cut, 40, TRUE)
+	dismantle()
+	return TRUE
+
+//Drop some bits when destroyed
+/obj/structure/flora/physically_destroyed(skip_qdel)
+	var/turf/T = get_turf(src)
```
MistakeNot4892: Could move this after `..()`

code/game/objects/structures/flora.dm
```diff
@@ -1,204 +1,320 @@
...
+	if(!..(TRUE)) //Tell parents we'll delete ourselves
+		return
+	if(T)
+		. = create_remains() != null
```
MistakeNot4892: `!isnull(create_remains())` or just `!create_remains()`

code/game/objects/structures/flora.dm
```diff
@@ -1,204 +1,320 @@
...
+		if(snd_cut)
+			playsound(src, snd_cut, 60, TRUE)
+	//qdel only after we do our thing, since we have to access members
+	if(!skip_qdel)
+		qdel(src)
+
+/**Returns an instance of the object the plant leaves behind when destroyed. Null means it leaves nothing. */
+/obj/structure/flora/proc/create_remains()
+	var/turf/T = get_turf(src)
```
MistakeNot4892: Why cast the turf?

code/game/objects/structures/fountain.dm
PsyCommando: Mainly because its easier to override this way than just piling up more things directly into initialize. And also it makes my life easier downstream.

code/game/objects/structures/seaweed.dm
PsyCommando: done!

code/game/objects/structures/pit.dm
PsyCommando: You mean only let simulated atoms in right?

code/game/objects/structures/__structure.dm
PsyCommando: done!

code/game/objects/structures/bedsheet_bin.dm
PsyCommando: Alright, done!

code/game/objects/structures/flora.dm
PsyCommando: Its because of random icon states. Some children don't have random icons, so instead of just overwriting the icon inside initialize() for each children I did that. Also, not all children may have the same icon state names for picking a random state.

code/game/objects/structures/flora.dm
PsyCommando: No chances of it getting moved into nullspace after qdeling?

code/game/objects/structures/flora.dm
PsyCommando: Second one doesn't work though, you'd need ``!!create_remains()``.
But I just went with isnull.

code/game/objects/structures/flora.dm
PsyCommando: That's a good question...
I probably just forgot to remove that var.

code/game/objects/structures/flora.dm
PsyCommando: Actually nevermind. Going to remove that XD

code/game/objects/structures/flora.dm
MistakeNot4892: Yep forgot a !

code/game/objects/structures/pit.dm
MistakeNot4892: Yep

code/game/objects/structures/bedsheet_bin.dm
Gaxeer: btw why not using LAZYLEN again in TRUE case?

MistakeNot4892 (approved):

code/game/objects/structures/bedsheet_bin.dm
MistakeNot4892: LAZYLEN is just a wrapper around length() anyway.

MistakeNot4892:
Needs a squash merge.

PR #2486, opened 2022-06-04
Random Ministation Asteroid Field [MDB IGNORE]

gyurka66:
## Description of changes
Added an asteroid wall spawner obj under obj/abstract/ministation/asteroid_spawner. It has a 1% chance to spawn and asteroid and then expand it  randomly with a recursive algorithm.

I've also filled in the area where the old asteroid field was completely with /obj/abstract/ministation/asteroid_spawner as to accomodate the changes in the code.

I thought having the same asteroid field every round was kind of boring. Randomness is more interesting.

## Authorship
Gyurka66 - Me
Loaf - Fixed parts of my code

## Changelog

:cl:
maptweak: Randomised Ministation Asteroid Field
/:cl:


maps/ministation/ministation_objects.dm
```diff
@@ -14,6 +19,35 @@
...
 			. = SSmaterials.weighted_minerals_rich
 
+/turf/exterior/wall/random/ministation/Initialize(var/ml, var/materialtype, var/rmaterialtype)
+	..(ml, materialtype, rmaterialtype)
+	if(ml)
+		starting = TRUE
+	. = INITIALIZE_HINT_LATELOAD
+
+/turf/exterior/wall/random/ministation/LateInitialize()
+	if(starting) //only need to call generate_asteroid here if this is a first node, the others are called in the generate_asteroid function
```
MistakeNot4892: Why not only lateload on mapload and remove the starting var entirely?

maps/ministation/ministation_objects.dm
```diff
@@ -14,6 +19,35 @@
...
+//tries to convert the space turfs around the asteroid into asteroids as well and tries to keep already existing asteroids around it from deleting in mapload.
+/turf/exterior/wall/random/ministation/proc/generate_asteroid(var/probability)
+	var/turf/exterior/wall/random/ministation/newasteroid
+	for(var/ndir in global.cardinal)
+		var/turf/ad = get_step(src, ndir)
+		if(prob(probability))
+			if(istype(ad, /turf/space)) //if it's space turn it into asteroid
+				newasteroid = ad.ChangeTurf(/turf/exterior/wall/random/ministation)
+				newasteroid.generate_asteroid(max(10,probability-10)) //reduce the probability of conversion with 10 percent for each turf away from the starting one.
+			if(istype(ad, /turf/exterior/wall/random/ministation)) //if it's asteroid prevent it from being deleted in it's own LateInitialize proc.
```
MistakeNot4892: This feels a bit strange - aren't you going to run into ordering issues where you're modifying the vars on an asteroid that has already fired its LateInitialize()? With the space change, aren't you going to get cases of double-calling generate_asteroid()? Apologies, it's a bit hard to follow the expected chain of events here.

maps/ministation/ministation_objects.dm
gyurka66: /turf/exterior/wall/ has it's own lateload logic that i didn't want to interfere with.

maps/ministation/ministation_objects.dm
gyurka66: If the wall is already initialized the modification of vars have no effect. generate_asteroid() on an already existing wall will only be called by it's LateInitialize method so no double calling. However it's true that it could be solved more clearly and elegantly so brb.

maps/ministation/ministation_objects.dm
afterthought2: For what it's worth, `LateInitialize` is called with the same args as `Initialize`, including `mapload` as the first parameter.

MistakeNot4892 (dismissed):
To be honest I don't think this is a fantastic pattern - I would prefer the use of mask turfs and a random map generator rather than per turf logic like this. But for the sake of a single map's asteroids I don't have any serious objections.

gyurka66:
> To be honest I don't think this is a fantastic pattern - I would prefer the use of mask turfs and a random map generator rather than per turf logic like this. But for the sake of a single map's asteroids I don't have any serious objections.

What do you think about this newer implementation. I've split the spawn logic into a separate spawner object.

maps/ministation/ministation_objects.dm
```diff
@@ -1,5 +1,30 @@
...
+	name = "random asteroid spawner"
+	icon = 'icons/misc/mark.dmi'
+	icon_state = "X"
+	color = COLOR_PURPLE
+
+/obj/ministation/random_asteroid_spawner/Initialize()
+	..()
+	. = INITIALIZE_HINT_LATELOAD
+
+/obj/ministation/random_asteroid_spawner/LateInitialize(var/ml)
```
MistakeNot4892: You might want to use the `/obj/abstract` type for this, if only for categorization purposes.

gyurka66: okay

gyurka66:
This is ready to be merged, imo.

MistakeNot4892:
Sorry, this fell off my radar. Will give it a look over today.

MistakeNot4892 (approved):

PR #2487, opened 2022-06-04
Fix structures not dropping their contents on dismantle. [IDB IGNORE] [MDB IGNORE]

PsyCommando:
## Description of changes
Some older structures, like the bedsheet bin have to dump their contents when dismantled. But that was removed from the base /obj/structure class for some reasons. Its a bit odd it would have been removed, since dumped contents can always be managed by overriding get_contained_external_atoms().

### DIFF
--- a/code/datums/extensions/assembly/assembly.dm
+++ b/code/datums/extensions/assembly/assembly.dm
@@ -52,7 +52,7 @@
 	else
 		var/atom/movable/H = holder
 		P.dropInto(H.loc)
-	if(P.type in critical_parts)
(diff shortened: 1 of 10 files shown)


MistakeNot4892:
Fixes to staging please.

I assume this was lost by accident.

PsyCommando:
> Fixes to staging please.
> 
> I assume this was lost by accident.

Gotcha! Without looking at blame, I'd guess its probably something to do with some of the other structures containing things having their own dropping code for their contents.

PR #2488, opened 2022-06-04
Let the tool interactions properly play the defined tool sounds

PsyCommando:
## Description of changes
The tool extension would completely ignore the use_sound defined in the tool archetypes and instead just grab the overrides. Changed that so if there's no overrides it grabs the sound from the archetype. 

## Changelog

:cl:
bugfix: Fixed how some tool interactions wouldn't play tool sounds.
/:cl:

### DIFF
--- a/code/modules/tools/tool_extension.dm
+++ b/code/modules/tools/tool_extension.dm
@@ -59,6 +59,8 @@
 			use_sound = pick(use_sound)
 		else 
 			use_sound = null
+	if(use_sound)
+		playsound(user.loc, use_sound, 100)
 
 	if(!do_after(user, max(5, CEILING(delay * get_tool_speed(archetype))), holder))
 		return TOOL_USE_FAILURE_NOMESSAGE
--- a/code/modules/tools/tool_extension.dm
+++ b/code/modules/tools/tool_extension.dm
@@ -51,10 +51,14 @@
 
 	user.visible_message(SPAN_NOTICE("\The [user] begins [start_message || tool_archetype.use_message] \the [target] with \the [holder]."), SPAN_NOTICE("You begin [start_message || tool_archetype.use_message] \the [target] with \the [holder]."))
 	var/use_sound = LAZYACCESS(tool_use_sounds, archetype)
-	if(islist(use_sound) && length(use_sound))
-		use_sound = pick(use_sound)
-	if(use_sound)
-		playsound(user.loc, use_sound, 100)
+	 //If no sound overrides, grab the archetype's sound/sound list
+	if(!use_sound) 
+		use_sound = tool_archetype.use_sound
+	if(islist(use_sound))
+		if(length(use_sound))
+			use_sound = pick(use_sound)
+		else 
+			use_sound = null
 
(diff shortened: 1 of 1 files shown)


code/modules/tools/tool_extension.dm
```diff
@@ -51,10 +51,14 @@
...
-	if(use_sound)
-		playsound(user.loc, use_sound, 100)
+	 //If no sound overrides, grab the archetype's sound/sound list
+	if(!use_sound) 
+		use_sound = tool_archetype.use_sound
+	if(islist(use_sound))
+		if(length(use_sound))
+			use_sound = pick(use_sound)
+		else 
+			use_sound = null
```
MistakeNot4892, line 61: Why null it?

PsyCommando: Because the rest of the code only checks if the var is non-null. If its just an empty list its gonna go through the check even if there's nothing in it.

MistakeNot4892 (dismissed):
I think there is almost certainly a better way to handle this, but I can't think of one that doesn't come with a significant memory cost (populating the sound list on all instances).

code/modules/tools/tool_extension.dm
```diff
@@ -51,10 +51,14 @@
 
 	user.visible_message(SPAN_NOTICE("\The [user] begins [start_message || tool_archetype.use_message] \the [target] with \the [holder]."), SPAN_NOTICE("You begin [start_message || tool_archetype.use_message] \the [target] with \the [holder]."))
 	var/use_sound = LAZYACCESS(tool_use_sounds, archetype)
-	if(islist(use_sound) && length(use_sound))
-		use_sound = pick(use_sound)
-	if(use_sound)
-		playsound(user.loc, use_sound, 100)
```
MistakeNot4892, line 57: You removed the first `playsound()` without replacing it, ftr.

PsyCommando: I removed it because I wasn't sure why there were 2 play sound really?
Most common tool interactions only play their sound once. So I wasn't sure if that was part of some merge artifact or something? I can add it back if you want.

PsyCommando:
> I think there is almost certainly a better way to handle this, but I can't think of one that doesn't come with a significant memory cost (populating the sound list on all instances).

There's probably a better way to do this yeah. I thought the way it worked was kinda silly. 
Would be nice to have some sort of  abstraction system for sound effects. So say a machine plays "beep.ogg", and its child would want to play "boop.ogg" instead, you could just replace that by having a proc both call with a generic name for the sound effect. And the sound effect that actually plays when the generic name is used can change across all children. Plus, that might allow some sound caching stuff to be done maybe?(Not too sure if you can actually cache sounds in byond though)

code/modules/tools/tool_extension.dm
MistakeNot4892: It's a sound for starting and a sound for ending. Most tool interactions in current code don't have a do_after(). Readding it would be ideal.

MistakeNot4892 (approved):

PR #2492, opened 2022-06-05
June 5, 2022 Bugfixes

noelle-lavenza:
## Description of changes
- Fix incorrect comments, argument definitions, and missing qdeleted checks in sound_tokens
- Add additional debugging info for events that aren't cleaned up properly
- PDAs will no longer give a critical shutdown message when equipped on a character that goes to cryo, because assemblies now shut down properly (and quietly) on destroy instead of remaining online.
- Offline assemblies can no longer have critical shutdowns.
- Adds several get_contained_external_atom overrides to fix things like cryo pulling internal components out of assemblies. (This *also* fixes the PDA cryo issue, I'm pretty sure.)
- Fix hanging refs to detonators and beakers on chem grenades.
- Fix hanging refs to owner/target on objectives.
- Fix improperly ordered arguments to event_repository.unregister in air alarms.
- Fix coating preventing items from GCing.
- Fix leaking pipes continuing to process after GC.
- Add some notes via comment on future refactors.

Bugfix roundup!

## Authorship
Me, on Hearth of Hestia several months ago.

### DIFF
--- a/code/datums/extensions/assembly/assembly.dm
+++ b/code/datums/extensions/assembly/assembly.dm
@@ -52,7 +52,7 @@
 	else
 		var/atom/movable/H = holder
 		P.dropInto(H.loc)
-	if(P.type in critical_parts)
+	if(enabled && (P.type in critical_parts))
 		critical_shutdown()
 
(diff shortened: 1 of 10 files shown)


MistakeNot4892 (approved):
Kinda hate the assembly boilerplate but heigh ho.

PR #2494, opened 2022-06-06
Swaps out all direct inventory references for get_equipped_item().

MistakeNot4892:
## Description of changes
This is going to be an absolute bastard to review. Swaps out all direct references to inventory vars (`wear_suit` and such) for the slot-based getter. Rewrites a bunch of logic to use slot loops instead of nested checks.

- A lot of this code is very old and needed a dust off.
- This will allow me to implement inventory slots a la hand slots down the track.

## Testing
- [x] Screen loc on equipment is not being cleared when inventory is hidden.
- [x] Uniforms aren't being equipped during spawn.
- [x] Tons of lag when equipping during spawn.
- [x] Handcuffs are almost certainly broken.

## Authorship
Myself.

## Changelog
Nothing player-facing.


MistakeNot4892:
Lag went away, not sure what the cause was. May have just been something else on my system.

MistakeNot4892:
I've done some limited testing of this one but frankly I'm not going to find all the edge cases regardless of how much time I spend on it. Would prefer to just get this in ASAP and then work on bugs as they crop up, since I will be slapping in a total rewrite of the core inventory system after this one.

afterthought2 (dismissed):

afterthought2 (approved):

PR #2495, opened 2022-06-06
Add pen tool archetype

PsyCommando:
## Description of changes
Make pens a tool archetype to standardize and make abstract what is considered a pen. So far only flashlight pens benefit from this. But if any other kinds of writing implements are made, this can take care of them pretty comprehensively.

I added a tool specific properties handling, for things that making a proc in the tool_extension was going to get too messy. So essentially, things like color, color name, pen state and uses left are all stored as properties that can be fetched or modified in the tool_extension. This way, more things can be added to the tool system later and have their properties abstracted like this.

I also rewrote a lot of the pen code to make better use of the extension, and also to clean it up. Managed to get rid of a lot of boolean vars by adding a pen flags field instead. 

Suggestions are welcome!

Helps make more pens work as pen when interacting with objects and the paperwork system. 

## Changelog

:cl:
tweak: More things are now considered pens. Like the flashlight pen.
/:cl:


code/__defines/paperwork.dm
```diff
@@ -0,0 +1,5 @@
+#define PEN_FLAG_ACTIVE     BITFLAG(1) //If the pen is expanded and ready to write
```
MistakeNot4892: Bitflags start at 0 just fyi.

PsyCommando: Oh, gonna change that then!

PsyCommando:
Note to self, DON'T FORGET TO TEST.

PsyCommando:
After testing, it seems like its pretty solid right now.

code/modules/paperwork/pen/multi_pen.dm
```diff
@@ -1,20 +1,31 @@
...
 		'icons/obj/items/pens/pen_red.dmi',
 		'icons/obj/items/pens/pen_green.dmi',
 	)
 
+/obj/item/pen/multi/Initialize(ml, material_key)
+	. = ..()
+	change_colour(colour_idx)
+
+/obj/item/pen/multi/make_pen_description()
+	desc = "It's a [istype(material)?"[material.name] ":""]pen with multiple colors of ink! Its currently set to [stroke_colour_name] [medium_name]."
```
MistakeNot4892: `It's`
Also possibly `[istype(material) ? "[ADD_ARTICLE([material.name])] pen" : "a pen"]`

MistakeNot4892: Hang on I think I'm having a brainfart, not sure if any material name would need 'an'

PsyCommando: I guess maybe osmium?

PsyCommando:
Btw, I tried applying some of the changes we discussed on discord, aka moving things over to a /datum/extension/tool/pen class. But, then I remembered /datum/extension/tool/variable.. I kinda feel it would be a really weird ass override to make when the only other sub class is something generic.. And if someone ever makes a pen that needs custom sounds overrides that could be problematic..

So I went with just making the tool properties list lazy, and literally only ever store the properties differing from the base params defined in the tool_acrchetype. So basically none of the existing tools creates lists during init, only pens. And I gathered that was your main issue with the previous version?

code/modules/paperwork/pen/multi_pen.dm
MistakeNot4892: Ahah yes.

code/game/objects/structures/door_assembly.dm
```diff
@@ -124,12 +124,12 @@
 
 /obj/structure/door_assembly/attackby(obj/item/W, mob/user)
 
-	if(IS_PEN(W))
+	if(IS_PEN(W) && CanPhysicallyInteractWith(user, src))
 		var/t = sanitize_safe(input(user, "Enter the name for the door.", src.name, src.created_name), MAX_NAME_LEN)
-		if((!t) || (!in_range(src, usr) && src.loc != usr))
```
MistakeNot4892: Need to readd a `CanInteract` or something post-input().

PsyCommando: Yeah, I see it now. Gonna have to take a look at the other ones too.

code/game/objects/structures/morgue.dm
```diff
@@ -80,22 +80,14 @@
 	if(Adjacent(user))
 		return attack_hand(user)
 
-/obj/structure/morgue/attackby(P, mob/user)
-	if(istype(P, /obj/item/pen))
-		var/new_label = sanitize_safe(input(user, "What would you like the label to be?", capitalize(name), null) as text|null, MAX_NAME_LEN)
-
-		if((!Adjacent(user) || loc == user))
+/obj/structure/morgue/attackby(obj/item/P, mob/user)
+	if(IS_PEN(P) && CanPhysicallyInteractWith(user, src) && loc != user)
```
MistakeNot4892: I feel like in an ideal world we could move this attackby checking down to the base level and have it check for the presence of the extension, call generic procs, etc.

PsyCommando: Honestly, the way this thing uses label is non-standard. We could probably just flush that stuff here, and let the hand labeler do it. 
I wouldn't move the pen interaction to a base class because a lot of things just use pens to rename themselves in some cases and in some others to do other things.  Ideally, I feel we probably should phase out using the pen on things to rename them. Machines could do it via multitool interactions or something, and the rest could be done through some alt extension?

PsyCommando: Anyways, I just removed labeling by pen on this and the crematorium. So now its using the standard way with the labeler.

code/game/objects/structures/pit.dm
```diff
@@ -201,3 +201,17 @@
 	var/died = max(cur_year - rand(0,70),born)
 
 	message = "Here lies [nam], [born] - [died]."
+
+/obj/structure/gravemarker/attackby(obj/item/W, mob/user)
+	if(IS_HATCHET(W))
+		visible_message("<span class = 'warning'>\The [user] starts hacking away at \the [src] with \the [W].</span>")
+		if(do_after(user, 30))
```
MistakeNot4892: Is this an accidental inclusion? If not, it should probably be using do_tool_interaction since you're already in the area.

PsyCommando: yeah, its crap leftover from the rebase. Going to get rid of it.

code/game/objects/items/books/_book.dm
```diff
@@ -133,9 +133,11 @@
 /obj/item/book/proc/formatpencode(var/mob/user, var/t, var/obj/item/pen/P)
 	. = t
 	if(findtext(t, "\[sign\]"))
-		. = replacetext(t, "\[sign\]", "<font face=\"[signfont]\"><i>[P ? P.get_signature(user) : "Anonymous"]</i></font>")
+		var/decl/tool_archetype/pen/parch = GET_DECL(TOOL_PEN)
+		var/signature = parch.get_signature(user, P) 
+		. = replacetext(t, "\[sign\]", "<font face=\"[signfont]\"><i>[signature]</i></font>")
 	if(P)
-		if(P.iscrayon)
+		if(P.get_tool_property(TOOL_PEN, TOOL_PROP_PEN_FLAG) & PEN_FLAG_CRAYON)
```
MistakeNot4892: Probably want to retrieve these property values and keep them rather than retrieving them repeatedly in this chain.

code/game/objects/structures/morgue.dm
```diff
@@ -10,10 +10,14 @@
 
 	var/obj/structure/morgue_tray/connected_tray
 
+/obj/structure/morgue/get_mechanics_info()
+	return "[..()]<BR>Can be labeled once with a hand labeler."
```
MistakeNot4892, line 14: Should this be down somewhere lower behind a `has_extension(src, /datum/extension/labels/single)`?

code/game/objects/structures/pit.dm
```diff
@@ -175,9 +175,9 @@
 		if(W.do_tool_interaction(TOOL_HATCHET, user, src, 3 SECONDS, "hacking away at", "hacking at"))
 			physically_destroyed(FALSE)
 		return TRUE
-	if(istype(W,/obj/item/pen))
+	if(IS_PEN(W) && CanPhysicallyInteractWith(user, src))
```
MistakeNot4892: You only need `CanPhysicallyInteractWith` if you're specifying something other than `src` as the target. `CanPhysicallyInteract` will do the job here.

code/modules/paperwork/pen/pen.dm
```diff
@@ -1,49 +1,34 @@
...
-	icon = 'icons/obj/items/pens/pen_green.dmi'
-	colour = "green"
-
-/obj/item/pen/invisible
-	desc = "It's an invisble pen marker."
-	colour = "white"
-	color_description = "transluscent ink"
+	set_extension(src, /datum/extension/tool,
+		list(
+				TOOL_DRILL = TOOL_QUALITY_WORST,
```
MistakeNot4892: Double tabbing here? Not a blocking issue, just looks a bit weird.

code/modules/paperwork/paper.dm
```diff
@@ -307,15 +309,11 @@
...
-
-		if(P.iscrayon)
-			iscrayon = TRUE
-
-		if(P.isfancy)
-			isfancy = TRUE
+		if(!(I.get_tool_property(TOOL_PEN, TOOL_PROP_PEN_FLAG) & PEN_FLAG_ACTIVE))
+			var/decl/tool_archetype/pen/parch = GET_DECL(TOOL_PEN)
+			parch.toggle_active(usr, I)
+		iscrayon = I.get_tool_property(TOOL_PEN, TOOL_PROP_PEN_FLAG) & PEN_FLAG_CRAYON
```
MistakeNot4892: As above, you should keep the value of get_tool_property when checking it repeatedly instead of getting it repeatedly.

code/modules/paperwork/clipboard.dm
```diff
@@ -33,156 +35,182 @@
...
+		attack_self(user)
 		update_icon()
-	return
 
 /obj/item/clipboard/ebony
 	material = /decl/material/solid/wood/ebony
 
 /obj/item/clipboard/steel
 	material = /decl/material/solid/metal/steel
-	material = /decl/material/solid/metal/steel
```
MistakeNot4892, line 176: Wonder how this happened? Merge artifact/find and replace mistake?

MistakeNot4892:
Apologies for the delay on getting to this one, have been ill.

PsyCommando:
> Apologies for the delay on getting to this one, have been ill.

No worries! I've been kinda busy anyways.

code/game/objects/items/books/_book.dm
PsyCommando: True!

code/game/objects/structures/morgue.dm
PsyCommando: Tbh, I probably don't need to mention it really. Its just because it used to be done with a pen. And I wanted to communicate that to people trying to do it with a pen again and wondering what was going on.
If we move that to a base class, that would need to be in its own PR though.

code/modules/paperwork/clipboard.dm
PsyCommando: Seems like a merge artifact to me. Sometimes during rebase it just can't seem to be able to tell when two competing changes are actually identical. Or maybe its just a human error.

code/game/objects/structures/pit.dm
PsyCommando: I actually also goofed that one. The interaction check was meant to be after the player input. But noted!

code/modules/paperwork/pen/pen.dm
PsyCommando: True, I'll fix that.

code/game/objects/structures/pit.dm
```diff
@@ -175,9 +175,12 @@
 		if(W.do_tool_interaction(TOOL_HATCHET, user, src, 3 SECONDS, "hacking away at", "hacking at"))
 			physically_destroyed(FALSE)
 		return TRUE
-	if(istype(W,/obj/item/pen))
+	if(IS_PEN(W))
 		var/msg = sanitize(input(user, "What should it say?", "Grave marker", message) as text|null)
-		if(msg)
+		if(!CanPhysicallyInteract(user))
```
MistakeNot4892, line 180: You could probably put this check in `can_use_tool()` on the tool archetype. Not blocking though.

code/modules/paperwork/clipboard.dm
```diff
@@ -33,156 +35,182 @@
 			return TRUE
 	. = ..()
 
+/obj/item/clipboard/examine(mob/user, distance, infix, suffix)
+	. = ..()
+	if(stored_pen)
+		to_chat(user, "It's holding \a [stored_pen].")
+	if(!LAZYLEN(papers))
+		to_chat(user, "It contains [length(papers)] / [max_papers] paper(s).")
```
MistakeNot4892: ` [length(papers)]/[max_papers] paper\s`

code/modules/paperwork/clipboard.dm
```diff
@@ -33,156 +35,182 @@
...
+	else 
+		to_chat(user, "It has room for [max_papers] paper(s).")
+
+/obj/item/clipboard/proc/top_paper()
+	return LAZYACCESS(papers, 1)
+
+/obj/item/clipboard/proc/push_paper(var/obj/item/P)
+	papers = papers? list(P) + papers : list(P) //Push at the top
```
MistakeNot4892:
````
LAZYINITLIST(papers)
papers.Insert(1, P)
````

File: code/modules/paperwork/clipboard.dm [hunk 15]
```diff
@@ -33,156 +35,182 @@
...
+
 /obj/item/clipboard/on_update_icon()
 	..()
-	if(toppaper)
-		overlays += overlay_image(toppaper.icon, toppaper.icon_state, flags=RESET_COLOR)
-		overlays += toppaper.overlays
-	if(haspen)
+	var/obj/item/top_paper = top_paper()
+	if(top_paper)
+		overlays += overlay_image(top_paper.icon, top_paper.icon_state, flags=RESET_COLOR)
```
MistakeNot4892, line 61: Would be nice if we could refactor overlays here to use SSoverlays but I appreciate that would mean refactoring every /obj/item that uses ..()

PsyCommando: I'm not sure I follow?

MistakeNot4892: Mostly just waffling, disregard.

MistakeNot4892 (approved):

PR #2496, opened 2022-06-06
Fix network power cycle race conditions

noelle-lavenza:
## Description of changes
Routers will now broadcast when they try to connect to a nonexistent network.
Other machines will instead queue their connection.

No more power cycle race conditions, which means #733 has one less blocking issue.

## Authorship
Me.

## Changelog

:cl: Noelle Lavenza
bugfix: Network devices now work after you turn them off and back on again.
/:cl:


MistakeNot4892 (dismissed):

MistakeNot4892:
Likely will fix #2004 as well.

MistakeNot4892 (approved):

noelle-lavenza:
Hoooold on. Might need additional changes; it looks like the issue might still occasionally occur with non-machine network devices, it's just far less likely because of the broadcast tweak.

PR #2500, opened 2022-06-07
RE:Fix structures not dropping their contents on dismantle. [IDB IGNORE][MDB IGNORE]

PsyCommando:
## Description of changes
Remake of #2487 after a git accident.


noelle-lavenza:
You don't need IDB ignore or MDB ignore on this because it doesn't touch icons or maps...

PsyCommando:
> You don't need IDB ignore or MDB ignore on this because it doesn't touch icons or maps...

Eh. Just in case I mess up with git again. Because it can start generating icon diff for the like 50 last commits sometimes.

MistakeNot4892 (approved):

PR #2502, opened 2022-06-07
Make shovel use the tool system

PsyCommando:
## Description of changes
Relies on #2483 for the flora interaction stuff.

* Makes the shovel into a tool, and change all shovel interactions to use the tool system.
* Added digging sound. CC0 and details in the file.
* Pits can be filled up when empty to delete them.

Transition old code using type checks to use the tool archetype instead.


MistakeNot4892 (changes requested):
Code looks good, just fix up the linter fails and we're all set.

PsyCommando:
> Code looks good, just fix up the linter fails and we're all set.

The fail is just because the PR's dependency isn't in yet.

PsyCommando:
Should be good now that the dependency got merged.

code/modules/mining/mine_items.dm
```diff
@@ -175,6 +179,12 @@
 	throwforce = 7
 	w_class = ITEM_SIZE_SMALL
 
+/obj/item/shovel/spade/Initialize(ml, material_key)
```
MistakeNot4892: Would be preferable to have a var on /shovel containing the tool quality, pass that in on /shovel Initialize, then change it on /spade.

MistakeNot4892:
Other than minor change above to avoid the double override, this code looks fine.

code/modules/mining/mine_items.dm
PsyCommando: I can do that.

MistakeNot4892 (approved):

PR #2503, opened 2022-06-08
Removed stumps, cleaned up some dropping limb code.

MistakeNot4892:
## Description of changes
- Removed stumps as a concept.
- Removed `is_stump()` and `is_droppable()` procs.
- Reworked a few organ dropping procs to try and avoid some runtimes observed during testing.

Stumps are a fairly outdated mechanic and were causing a lot of problems/code smell, no need for them now that we have arteries as a consequence for a traumatic amputation.

## Authorship
Myself.

## Changelog
:cl:
tweak: Limb stumps no longer exist!
/:cl:


code/modules/organs/external/_external.dm
```diff
@@ -1003,90 +1024,83 @@ Note that amputating the affected organ does in fact remove the infection from t
...
+				qdel(src)
 
-			for(var/obj/item/I in src)
-				I.dropInto(loc)
-				I.throw_at(get_edge_target_turf(src,pick(global.alldirs)), rand(1,3), THROWFORCE_GIBS)
-
-			qdel(src)
+/obj/item/organ/external/proc/handle_async_post_dismemberment(var/mob/living/carbon/human/victim)
+	set waitfor = FALSE
+	sleep(1)
```
afterthought2: It was awful before...and it's still awful. Just do it synchronously at the end. If you have performance concerns, add an arg to the proc that suppresses this stuff and then call it at the end of batch removal outside of qdel instead (similarly for recursive calls).

Notably, this spawn approach does _not_ address any performance concerns _because this is queued for every organ removed anyway_.

MistakeNot4892: I don't have performance concerns I just had no idea what this would impact if I didn't replicate the previous behavior.

code/modules/mob/living/carbon/human/update_icons.dm
```diff
@@ -377,6 +383,9 @@ var/global/list/damage_icon_parts = list()
 	else
 		//BEGIN CACHED ICON GENERATION.
 		var/obj/item/organ/external/chest = get_organ(BP_CHEST)
+		if(!chest)
+			return // This should not be happening, but occasionally there is an ordering issue or a late call to update icon.
```
PsyCommando: Imo, you probably shouldn't just ignore that happening. Its something I've been trying to track down for a while, and it hadn't happened to me in a long time.

code/modules/mob/mob.dm
```diff
@@ -1006,6 +1006,7 @@
 
 /mob/physically_destroyed(var/skip_qdel)
 	SHOULD_CALL_PARENT(FALSE)
+	dump_contents()
```
PsyCommando: That probably breaks things being thrown around in human/gib(). There's a loop to throw around the mob's contents. So instead of throwing the contents its just going to drop it down.

code/modules/organs/external/_external.dm
PsyCommando: I actually never notice that spawn before... There's gotta be a way to get rid of it?

code/modules/mob/living/carbon/human/update_icons.dm
MistakeNot4892: It was happening during testing and I'm really not clear on why. I assume because some part of droplimb is forcing a body update - it's possible now that it's queued it won't be an issue. I might add a PRINT_STACK_TRACE before returning.

code/modules/mob/mob.dm
MistakeNot4892: ~~Mob gear is stripped and thrown before parent is called so nope it works fine.~~ Good catch.

code/modules/mob/living/carbon/human/update_icons.dm
MistakeNot4892: Turns out we can completely remove the need for a root organ.

code/modules/organs/external/_external.dm
MistakeNot4892: It wasn't actually needed.

afterthought2 (dismissed):

code/modules/mob/death.dm
```diff
@@ -8,13 +8,15 @@
 	set_invisibility(101)
 	UpdateLyingBuckledAndVerbStatus()
 	remove_from_dead_mob_list()
+	dump_contents()
```
PsyCommando, line 11: Might be good to move the code in humans that throw the contained items around over here? So all gibbing does the same thing?

code/modules/mob/living/carbon/human/update_icons.dm
```diff
@@ -347,15 +347,15 @@ var/global/list/damage_icon_parts = list()
 	var/obj/item/organ/internal/eyes/eyes = get_organ(species.vision_organ || BP_EYES)
 	icon_key += istype(eyes) ? eyes.eye_colour : COLOR_BLACK
 
-	for(var/organ_tag in species.has_limbs)
+	for(var/organ_tag in species.has_limbs) // This means non-species limb slots won't draw, todo
```
PsyCommando: Good catch!

code/modules/mob/living/carbon/human/update_icons.dm
```diff
@@ -347,15 +347,15 @@ var/global/list/damage_icon_parts = list()
...
 		var/obj/item/organ/external/part = get_organ(organ_tag)
-		if(isnull(part) || part.is_stump() || part.organ_tag == BP_TAIL)
+		if(isnull(part) || part.organ_tag == BP_TAIL)
 			icon_key += "0"
 			continue
 		for(var/M in part.markings)
 			icon_key += "[M][part.markings[M]]"
 		if(part)
-			icon_key += "[part.bodytype.get_icon_cache_uid(part.owner)][part.render_alpha]"
+			icon_key += "[part.bodytype?.get_icon_cache_uid(part.owner)][part.render_alpha]"
```
PsyCommando: I'm a bit against ignoring the bodytype being null there since its a major fault and probably should crash. 

Maybe we should define some expectations for when update_body should run? Because In my case my expectation were that all limbs must be initialized when its called. That's why I haven't been just blanket calling it each times something changes in the limb hierarchy. Its because otherwise you end up with limbs silently not updating along with the rest because of just timing, and the state after update_body runs now becomes uncertain and unreliable.

code/modules/organs/external/head.dm
```diff
@@ -122,6 +122,9 @@
 
 /obj/item/organ/external/head/proc/get_hair_icon()
 	var/image/res = image(bodytype.icon_template,"")
+	if(!owner)
```
PsyCommando, line 125: Won't that mean decapitated heads are going to turn bald once they get dumped?

code/modules/mob/death.dm
MistakeNot4892: When I redo inventory slots I might generalize the behavior but I'll think about it in another PR

code/modules/organs/external/head.dm
MistakeNot4892: Yep. Heads need reworking to track the overlays directly instead of relying on owner.

code/modules/mob/living/carbon/human/update_icons.dm
MistakeNot4892: I can probably remove this change, it was before I worked out why things were double-updating during removal.

MistakeNot4892: Yep, fixed.

code/modules/mob/living/carbon/human/update_icons.dm
MistakeNot4892: Changed this to use the global limb flag list instead.

code/modules/mob/death.dm
PsyCommando: kk!

code/modules/mob/living/carbon/human/update_icons.dm
PsyCommando: Great, thanks!

afterthought2 (approved):

PR #2507, opened 2022-06-10
Fix dynamic lights on ambiently-lit turfs

noelle-lavenza:
## Description of changes
Fix dynamic lights on ambiently-lit turfs by calling lighting_build_overlay instead of generate_missing_corners; this still calls that eventually, but also does important other steps too, like setting the corners to active.

Flashlights not working on exoplanets at night was... pretty weird.

## Changelog
:cl: Noelle Lavenza
bugfix: Dynamic lights now work on ambiently-lit turfs.
/:cl:

### DIFF
--- a/code/modules/lighting/lighting_turf.dm
+++ b/code/modules/lighting/lighting_turf.dm
@@ -39,7 +39,7 @@
 		ambient_b = (HEX_BLUE(ambient_light) / 255) * ambient_light_multiplier
 
 	if (!corners || (null in corners))
-		generate_missing_corners()
+		lighting_build_overlay()
 
 	for (var/thing in corners)
 		var/datum/lighting_corner/C = thing


Lohikar:
I'm already working on a set of more comprehensive fixes to ambient lights, and this isn't the correct way to fix it anyways.

MistakeNot4892:
Bit of friction around this one so to lay it out, having chatted with Lohi - closure out of hand was a misstep, but blocking is valid if a maint believes it's not an acceptable fix. I would suggest Lohi requests specific changes, opens his alternative PR in the immediate near future, or Noelle asks for a maint consensus on the fix.

MistakeNot4892:
This has been definitively fixed in the ambient lighting rewrite PR so going ahead and closing it.

PR #2509, opened 2022-06-10
Ambient lights v2

Lohikar:
This is essentially a rewrite (whoops) of ambient lights with different init behavior -- it should generally behave a lot saner.

Changes:
- Ambient lights can now propagate down Z-levels
- Dynamic lighting is no longer broken by ambient lights
- Ambient lights no longer generate pointless corners on space
- Adjacent ambient lights will properly mix colors
- Ambient lights are now initialized during SSlighting init rather than SSatoms


Lohikar:
This is not done (it has some bugs left), but it does fix some major bugs and isn't broken enough to block.

MistakeNot4892 (approved):
Happy to get this merged and debug further in future PRs or during staging.

PR #2510, opened 2022-06-11
Swapping get_organ() out for safer procs.

MistakeNot4892:
## Description of changes
- `get_organ()` can now be supplied with an optional type to check against before returning a value.
- Many uses of `get_organ()` now use `GET_EXTERNAL_ORGAN` and `GET_INTERNAL_ORGAN` macros to enforce typing.

`get_organ()` used to only return external organs, now returns all organs, some calls expected a type that they don't receive.

## Authorship
Myself.

## Changelog
Nothing player-facing.


afterthought2:
I don't actually like this. There are maybe a few places where additional checks are needed (e.g. surgery code), but for the most part if you are calling this proc and supplying the organ tag, you already know the kind of organ you are getting. And even the surgery code either shouldn't care about the organ type or should be doing more aggressive checks after getting it.

MistakeNot4892:
Redid this as discussed.

MistakeNot4892:
By virtue of how wide-reaching this PR is, I can't effectively test it, but I have spawned in, roamed around touching all the objects, fallen down the elevator and blown myself up with no apparent issues or runtimes.

afterthought2 (approved):

PsyCommando:
I'm fine with it, but I feel like it could probably be more easily solved by type checking the result you get to whatever you need? Because at the end of the day, this isn't going to fix problems, but make them manifest differently? Like instead of a runtime for bad casts, you'll get a possibly silent failed nullcheck and things will just break silently?

MistakeNot4892:
Not sure what you mean, this absolutely solves the problem of get_organ() changing behavior. It's not masking problems, it's replicating the previous behavior and making it a lot more predictable and stringent overall.

MistakeNot4892:
Previous behavior:
You call get_organ(), you get an external organ or null, you typecheck it.

Post-rewrite:
You call get_organ(), you get an organ or null, even if the context you want it in is nonsensical like grabbing someone by the eyes. There is no way to specify other than typechecking yourself.

New system:
You provide an expected type to get_organ() based on the context you want it in, or you call get_organ() as previously if you really want to for some reason.

Any bugs this introduces are due to me cocking up what I chose to swap in for get_organ(), it's not implicitly going to hide problems.

PR #2515, opened 2022-06-12
Removed accidental organ qdel.

### LINKED ISSUE #2513: Surgery lost limbs
#### Description of issue
Amputation cause limb to dissapear

#### Steps to reproduce
1) Cut off an arm with surgery
2) Install it back to place 
3) Your patient has arm without hand

Server Revision: 65c7b877a39455309a43a2685d61b565c8351453 - staging - 2022-06-12

#### Issue bingo

MistakeNot4892:
Fixes #2513. @jade-lavenza found my fuckup, ty.

### DIFF
--- a/code/modules/organs/external/_external_icons.dm
+++ b/code/modules/organs/external/_external_icons.dm
@@ -1,10 +1,7 @@
 var/global/list/limb_icon_cache = list()
 
-/obj/item/organ/external/set_dir(var/direction, var/forced)
-	SHOULD_CALL_PARENT(FALSE)
-	if(forced)
-		return ..(direction)
-	return FALSE
+/obj/item/organ/external/set_dir()
+	return ..(SOUTH)
 
 /obj/item/organ/external/proc/compile_icon()
 	overlays.Cut()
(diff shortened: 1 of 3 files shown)


code/modules/organs/external/_external_icons.dm
```diff
@@ -1,10 +1,7 @@
 var/global/list/limb_icon_cache = list()
 
-/obj/item/organ/external/set_dir(var/direction, var/forced)
-	SHOULD_CALL_PARENT(FALSE)
-	if(forced)
-		return ..(direction)
-	return FALSE
+/obj/item/organ/external/set_dir()
+	return ..(SOUTH)
```
afterthought2, line 4: I mean, things do call this with second arg, not sure what's going on here.

MistakeNot4892: The second arg doesn't exist in any other set_dir override.
EDIT: Forcing dir set here is also undesirable, limbs have directions with no state.

MistakeNot4892: Pushed an amended commit, nothing uses the forced arg now.

afterthought2 (approved):

PR #2519, opened 2022-06-14
Fix post rejuv +adherents

PsyCommando:
## Description of changes
The main fix is preventing adherents from dying instantly on spawn from the main screen.
That in turn required changing how organ properties are kept on organs.
Which in turn meant I had to define some subtypes for some races not behaving like everything else.

Moved organ properties such as prosthetic, assisted and crystal into their own var instead of being stored into the organ status flag to avoid complications with having to constantly approximately restoring the original state in post-rejuv.

Just made limbs overrides for the mantid to set their property to crystal, since not everything in the mob is crystaline.
Also made sub-types for the head and eyes of utility frames to set their glowing eye and eye icon instead of doing it in post-rejuv.

## Changelog
:cl:
fix: Adherents no longer die immediately on spawn.
fix: No more random runtime spam with some backpacks. 
fix: Synthetic races now properly show they don't have blood on the health scanner.
fix: Fixed some synthetic eyes not having the right icon or name.
/:cl:

### DIFF
--- a/code/__defines/mobs.dm
+++ b/code/__defines/mobs.dm
@@ -206,14 +206,12 @@
 
 // Prosthetic helpers.
 #define BP_IS_PROSTHETIC(org) (!QDELETED(org) && (org.organ_properties & ORGAN_PROP_PROSTHETIC))
-#define BP_IS_ASSISTED(org)   (!QDELETED(org) && (org.organ_properties & ORGAN_PROP_ASSISTED))
 #define BP_IS_BRITTLE(org)    (!QDELETED(org) && (org.status           & ORGAN_BRITTLE))
 #define BP_IS_CRYSTAL(org)    (!QDELETED(org) && (org.organ_properties & ORGAN_PROP_CRYSTAL))
 
 //Prosthetics setter
 //Since assisted and prosthetic are mutually exclusive set them using this helper for convenience
-#define BP_SET_PROSTHETIC(org) org.organ_properties = ((org.organ_properties & (~ORGAN_PROP_ASSISTED))   | ORGAN_PROP_PROSTHETIC);
-#define BP_SET_ASSISTED(org)   org.organ_properties = ((org.organ_properties & (~ORGAN_PROP_PROSTHETIC)) | ORGAN_PROP_ASSISTED);
+#define BP_SET_PROSTHETIC(org) org.organ_properties |= ORGAN_PROP_PROSTHETIC;
 #define BP_SET_CRYSTAL(org)    org.organ_properties |= ORGAN_PROP_CRYSTAL;
 
 // Limb flag helpers
--- a/code/__defines/mobs.dm
+++ b/code/__defines/mobs.dm
@@ -209,8 +209,7 @@
 #define BP_IS_BRITTLE(org)    (!QDELETED(org) && (org.status           & ORGAN_BRITTLE))
 #define BP_IS_CRYSTAL(org)    (!QDELETED(org) && (org.organ_properties & ORGAN_PROP_CRYSTAL))
 
-//Prosthetics setter
-//Since assisted and prosthetic are mutually exclusive set them using this helper for convenience
+//Organ Properties Setters
 #define BP_SET_PROSTHETIC(org) org.organ_properties |= ORGAN_PROP_PROSTHETIC;
 #define BP_SET_CRYSTAL(org)    org.organ_properties |= ORGAN_PROP_CRYSTAL;
 
--- a/code/__defines/mobs.dm
+++ b/code/__defines/mobs.dm
@@ -205,10 +205,16 @@
 #define AUGMENTATION_ORGANIC  2
 
 // Prosthetic helpers.
-#define BP_IS_PROSTHETIC(org) (!QDELETED(org) && (org.status & ORGAN_PROSTHETIC))
-#define BP_IS_ASSISTED(org)   (!QDELETED(org) && (org.status & ORGAN_ASSISTED))
-#define BP_IS_BRITTLE(org)    (!QDELETED(org) && (org.status & ORGAN_BRITTLE))
-#define BP_IS_CRYSTAL(org)    (!QDELETED(org) && (org.status & ORGAN_CRYSTAL))
+#define BP_IS_PROSTHETIC(org) (!QDELETED(org) && (org.organ_properties & ORGAN_PROP_PROSTHETIC))
+#define BP_IS_ASSISTED(org)   (!QDELETED(org) && (org.organ_properties & ORGAN_PROP_ASSISTED))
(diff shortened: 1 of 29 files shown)


code/__defines/mobs.dm
```diff
@@ -205,10 +205,14 @@
...
-#define BP_IS_PROSTHETIC(org) (!QDELETED(org) && (org.status & ORGAN_PROSTHETIC))
-#define BP_IS_ASSISTED(org)   (!QDELETED(org) && (org.status & ORGAN_ASSISTED))
-#define BP_IS_BRITTLE(org)    (!QDELETED(org) && (org.status & ORGAN_BRITTLE))
-#define BP_IS_CRYSTAL(org)    (!QDELETED(org) && (org.status & ORGAN_CRYSTAL))
+#define BP_IS_PROSTHETIC(org) (!QDELETED(org) && (org.organ_properties & ORGAN_PROP_PROSTHETIC))
+#define BP_IS_BRITTLE(org)    (!QDELETED(org) && (org.status           & ORGAN_BRITTLE))
+#define BP_IS_CRYSTAL(org)    (!QDELETED(org) && (org.organ_properties & ORGAN_PROP_CRYSTAL))
+
+//Prosthetics setter
+//Since assisted and prosthetic are mutually exclusive set them using this helper for convenience
```
MistakeNot4892: Comment is now out of date ftr

PsyCommando: I wonder if the setters are actually needed anymore though. I only made them because assisted / prosthetic are exclusive. But with assisted gone, its not doing much.

code/modules/organs/internal/_internal.dm
```diff
@@ -3,10 +3,12 @@
 ****************************************************/
 /obj/item/organ/internal
 	scale_max_damage_to_species_health = TRUE
+	var/tmp/prosthetic_name //Name given to the organ when robotized. Null means no changes
 	var/tmp/alive_icon //icon to use when the organ is alive
 	var/tmp/dead_icon // Icon to use when the organ has died.
 	var/tmp/prosthetic_icon //Icon to use when the organ is robotic
 	var/tmp/prosthetic_dead_icon //Icon to use when the organ is robotic and dead
+	var/tmp/prosthetic_icon_file //The path to the icon file for the prosthetic icons. Null means use the current 'icon' value.
```
MistakeNot4892, line 11: Not a fan of every organ in the game having these vars defined on them - they should be handled by a prosthetics model, defaulting to the base type if none is provided.

code/modules/species/species.dm
```diff
@@ -865,9 +865,9 @@ var/global/const/DEFAULT_SPECIES_HEALTH = 200
 		if(31 to 45)	. = 4
 		else			. = 8
 
+/**Runs extra processing after any organs member of this species goes through rejuvenation. */
```
MistakeNot4892, line 868: Why the block comment syntax for a single line comment?

mods/species/utility_frames/species.dm
```diff
@@ -62,7 +63,11 @@
 	)
 	has_organ = list(
 		BP_POSIBRAIN = /obj/item/organ/internal/posibrain,
-		BP_EYES = /obj/item/organ/internal/eyes/robot
+		BP_EYES      = /obj/item/organ/internal/eyes/robot/utility_frame, //Eyes got a different icon file 
+		BP_CELL      = /obj/item/organ/internal/cell,
+	)
+	override_limb_types = list(
+		BP_HEAD = /obj/item/organ/external/head/utility_frame, //Needed for glowing eyes shennanigans
```
MistakeNot4892, line 70: Comments after list entries will fail the indentation check. Also trailing comma here and above.

mods/species/bayliens/adherent/organs/organs_internal.dm
```diff
@@ -10,13 +10,17 @@
...
+	. = ..()
+
+/obj/item/organ/internal/posibrain/adherent/refresh_action_button()
 	. = ..()
 	if(.)
 		action.button_icon_state = "adherent-brain"
 		if(action.button) action.button.UpdateIcon()
 
-/obj/item/organ/internal/brain/adherent/attack_self(var/mob/user)
+/obj/item/organ/internal/posibrain/adherent/attack_self(var/mob/user)
```
MistakeNot4892, line 23: Adherent are not positronic, this is not a good repath imo.

mods/species/ascent/mobs/bodyparts.dm
```diff
@@ -88,3 +88,28 @@
 /obj/item/organ/external/head/insectoid/mantid/proc/reset_cooldown()
 	cooldown = FALSE
 	refresh_action_button()
+
+//Basically appends the crystal flag to all the types specified
+#define MANTIDIFY_ORGAN(ORGAN_PATH) ##ORGAN_PATH/Initialize(){BP_SET_CRYSTAL(src) . = ..();}
+//External organs
+MANTIDIFY_ORGAN(/obj/item/organ/external/chest/insectoid/mantid)
```
MistakeNot4892, line 95: What on earth? Why? Nothing else uses the mantid organs and this really isn't saving lines.

code/modules/organs/internal/_internal.dm
PsyCommando: Internal organs have no prosthetic models. 
Also, adding prosthetic models to internal organs is like a brand new feature, and its out of scope for a fix PR. I probably could do it, but not on this pr.

code/modules/species/species.dm
PsyCommando: Cause this is the only comment format that properly shows proc description in the auto-complete.

mods/species/utility_frames/species.dm
PsyCommando: Oh okay.

mods/species/bayliens/adherent/organs/organs_internal.dm
PsyCommando: Well, its hard to tell from the weird conflicting info around the code. Some of the stuff says they're piezo-electronic. I can set them back to an organic brain though. With the changes in here it won't instantly kill them anymore at least.

mods/species/ascent/mobs/bodyparts.dm
PsyCommando: The point is only have one line to change instead of ten million organ defines.

PsyCommando:
Closed since loaf just decided to do it for me apparently

MistakeNot4892:
I told you outright last night that I was going to do an alternate commit as a compromise as I had serious objections to the subtyping. If you prefer next time I'll just tell you I'm blocking your PR and then do nothing.

PR #2521, opened 2022-06-14
Moving exoplanet building to SSmapping.

MistakeNot4892:
## Description of changes
Moves exoplanet generation down to pre-SSatoms alongside the other away site and map template based init behaviors.

Centralizes map loading logic and minimizes potential for ordering problems.

## Authorship
Myself.

## Changelog
Nothing player-facing.

### DIFF
--- a/code/modules/overmap/exoplanets/_exoplanet.dm
+++ b/code/modules/overmap/exoplanets/_exoplanet.dm
@@ -93,12 +93,9 @@
 	if(length(possible_strata))
 		crust_strata = pick(possible_strata)
 
-/obj/effect/overmap/visitable/sector/exoplanet/Initialize(mapload, z_level)
-	if(global.overmaps_by_name[overmap_id])
-		forceMove(locate(1, 1, z_level))
-	return ..()
-
 /obj/effect/overmap/visitable/sector/exoplanet/proc/build_level(max_x, max_y)
+
+	find_z_levels()
 
 	maxx = max_x || world.maxx
 	maxy = max_y || world.maxy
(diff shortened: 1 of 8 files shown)


MistakeNot4892:
image
Appears to be working, it's registering the correct map levels, moving itself appropriately on the overmap, initializing name correctly.

code/modules/overmap/exoplanets/_exoplanet.dm
```diff
@@ -93,15 +93,12 @@
...
-/obj/effect/overmap/visitable/sector/exoplanet/Initialize(mapload, z_level)
-	if(global.overmaps_by_name[overmap_id])
-		forceMove(locate(1, 1, z_level))
-	return ..()
-
 /obj/effect/overmap/visitable/sector/exoplanet/proc/build_level(max_x, max_y)
 
-	maxx = max_x ? max_x : world.maxx
-	maxy = max_y ? max_y : world.maxy
+	find_z_levels()
```
afterthought2, line 98: Not following this change. It will now ignore the passed `z_level` arg in Initialize, which is fine because you are now supplying a loc. But it calls `find_z_levels` in a parent Initialize call already, why are you calling it again? Generally it's dangerous to call this without then re-registering everything.

If it's because this actually fires before Initialize, I would consider `INITIALIZE_IMMEDIATE` on this obj.

MistakeNot4892: I actually didn't see the /visitable override on this, will have a closer look.

MistakeNot4892:
This is going to need a ton of work and I'm not sure when I'll get to it. Possibly early August when I have leave.

PR #2522, opened 2022-06-14
Comments out waitfor=FALSE in submap template load.

MistakeNot4892:
## Description of changes
Temporary workaround for a problem I introduced with the SSmapping rework.

Stops an assert failing due to overlapping map template loads.

## Authorship
@afterthought2 pointed out the issue.

## Changelog
Nothing player-facing.


afterthought2 (approved):

afterthought2:
Going to bypass the usual wait time because this is making everything fail unit testing.

PR #2525, opened 2022-06-14
Fixes adherent spawning and reworks their icon generation.

MistakeNot4892:
## Description of changes
- `post_organ_rejuvenate` has been renamed and cleaned up for clarity.
- Species status resetting is now centralized in one proc.
- Adherent icon stuff is handled by a prosthetics model. Other adherent code also cleaned up.
- Old FBP hooks that replace brain with MMI on `robotize()` removed.
- robotize() no longer supplies or expects a default arg to the procs, but rather grabs it during applying robotize().
- Prosthetic manufacturers now use a proc to retrieve the icon.

Alternative to a big chunk of #2519.

Adherent are currently very broken. This makes them work.

## Authorship
Me, @PsyCommando and @jade-lavenza.

## Changelog
Nothing player-facing.


code/modules/organs/organ.dm
```diff
@@ -287,13 +289,16 @@
...
 	if(ignore_prosthetic_prefs && ishuman(owner) && owner.client && owner.client.prefs && owner.client.prefs.real_name == owner.real_name)
 		for(var/decl/aspect/aspect as anything in owner.personal_aspects)
 			if(aspect.applies_to_organ(organ_tag))
 				aspect.apply(owner)
-	if(species)
-		species.post_organ_rejuvenate(src, owner)
+	reset_status()
+
+/obj/item/organ/proc/reset_status()
+	status = initial(status)
```
PsyCommando, line 301: This isn't very reliable. Because initial is the state as defined in the definition. It doesn't take into account things done in init.

Also, its a bit silly to do all that extra processing just because you don't wanna move mostly immutable flags to their own var, and thus literally not have to do any processing at all besides when reseting the mob completely. Like prosthetic and crystal definitely shouldn't be stored along with mutable flags like dislocated, artery cut, cut away, etc..

Also shouldn't you reset the status var before applying aspects? Because you're just gonna clear whatever you've applied to the status?

code/modules/species/species.dm
```diff
@@ -865,8 +862,10 @@ var/global/const/DEFAULT_SPECIES_HEALTH = 200
 		if(31 to 45)	. = 4
 		else			. = 8
 
-/decl/species/proc/post_organ_rejuvenate(var/obj/item/organ/org, var/mob/living/carbon/human/H)
-	if(org && (org.species ? (org.species.species_flags & SPECIES_FLAG_CRYSTALLINE) : (species_flags & SPECIES_FLAG_CRYSTALLINE)))
+// This should only ever be called via the species set on the organ; calling it across species will cause weirdness.
+/decl/species/proc/apply_species_organ_modifications(var/obj/item/organ/org, var/mob/living/carbon/human/H)
+	SHOULD_CALL_PARENT(TRUE)
+	if(species_flags & SPECIES_FLAG_CRYSTALLINE)
```
PsyCommando, line 868: I'm not sure why you're so fixated on that. Its not because its working now that its not going to break something later, and it'll take ages to track down that this thing was forcing crystal on everything.

On spawn some species got bits that aren't meant to be crystalline. And you're making them all crystaline in here. That's just asking for trouble.

code/modules/organs/organ.dm
afterthought2: I also don't like this, though maybe for a different reason. I don't think you need two vars for the flags, but you probably should be doing something like `status &= (ROBOTIC | CRYSTAL)` here. If you have a prosthetic organ, won't this de-robotize it otherwise?

code/modules/species/species.dm
afterthought2: I disagree with that. I think the species should be free to do basically whatever to organs. Whether that's at `/species` level with this flag or done at `/adherent` level doesn't really matter to me (though I'd probably have done the latter personally). It's easy to debug: "The species modified the organs, did it do so correctly?" If this is _currently inaccurate as written_ then it's a reasonable objection though.

code/modules/organs/organ.dm
MistakeNot4892: I am replicating the existing behavior, not adding new behavior. If you look at the diff you'll see that previously it set it to 0, which would have cleared any init behavior as well.

Good point about aspects, I actually thought I moved it up higher.

I sincerely have no clue what you mean by `all that extra processing`. This PR has absolutely nothing to do with splitting properties and status into two args, that was your PR and I don't object to it.

code/modules/species/species.dm
MistakeNot4892: The purpose of renaming the proc was to make it more obvious what it was doing. I don't think it's unclear now.

I really, really don't appreciate being called fixated for fixing bugs with existing code rather than removing it wholesale. It is not an unstable dangerous thing that's going to break other things, it's slapping a flag on some organs. 

Species that have bits that shouldn't be crystalline shouldn't be applying the crystalline flag in this proc and I have yet to see it happen. I've said about nine times now that the entire function of this proc is to apply things that are universal to organs that belong to the species.

The species level flag exists because on Bay there were 3+ species using crystal organs. It could be removed here and the flag put into adherent and mantid modpacks.

afterthought2 (approved):

code/modules/organs/organ.dm
PsyCommando: Well, its just you've been working on exactly the same part as I've been and fixing the same issue. So I've assumed you probably don't want the changes you've excluded from my stuff? 

> I also don't like this, though maybe for a different reason. I don't think you need two vars for the flags, but you probably should be doing something like `status &= (ROBOTIC | CRYSTAL)` here. If you have a prosthetic organ, won't this de-robotize it otherwise?

I think its important to separate mutable from immutable stuff? Just because its easy to accidentally decide to clear that var named "status" assuming it actually only contains status effects, and not properties? But whatever.

code/modules/species/species.dm
PsyCommando: > I disagree with that. I think the species should be free to do basically whatever to organs. Whether that's at `/species` level with this flag or done at `/adherent` level doesn't really matter to me (though I'd probably have done the latter personally). It's easy to debug: "The species modified the organs, did it do so correctly?" If this is _currently inaccurate as written_ then it's a reasonable objection though.

It should, but it probably shouldn't just change the nature of what it was defined to have at definition, without getting some guarantees first? 

There's no mechanism to make an organ aware that if its flag gets reset, it should go back to being crystal/prosthetics besides this external proc that is called by the parent mob only when the parent modifies the organ. 
Thing is, its perfectly legal for organs to clear their status flag when their damage proc is called or for whatever other reasons. But it cannot know its supposed to reset to being crystal or prosthetic unless it was defined like that? And here we're just willingly not defining the organs with the flag. So we're just removing our only shaky guarantee.

This is what I mean here. This is the kind of things that cause really "fun" bugs from experience..

> The purpose of renaming the proc was to make it more obvious what it was doing. I don't think it's unclear now.
> 
> I really, really don't appreciate being called fixated for fixing bugs with existing code rather than removing it wholesale. It is not an unstable dangerous thing that's going to break other things, it's slapping a flag on some organs.
> 

Well, I really really don't like just getting treated like a dummy, and having a maintainer open a competing PR instead of working things out with me, or just asking if you could just take over it or something? That was a bit rude from you imo.
And yeah I used the word fixated. It wasn't meant to be offensive, and I'm sorry if it was, but I didn't know what else to call it?

> Species that have bits that shouldn't be crystalline shouldn't be applying the crystalline flag in this proc and I have yet to see it happen. I've said about nine times now that the entire function of this proc is to apply things that are universal to organs that belong to the species.
> 
> The species level flag exists because on Bay there were 3+ species using crystal organs. It could be removed here and the flag put into adherent and mantid modpacks.

But, unless I'm wrong, there's not only fully crystalline species with the crystalline flag in this PR? Look at the organs lists. Some of the species with the crystalline flag got at least a random machine or biological part. 

For example mantids got this, and you're setting the species flag to crystalline:
```dm
/obj/item/organ/internal/controller
	name = "system controller"
	desc = "A fist-sized lump of complex circuitry."
	icon = 'mods/species/ascent/icons/ascent.dmi'
	icon_state = "plant"
	parent_organ = BP_CHEST
	organ_tag = BP_SYSTEM_CONTROLLER
	surface_accessible = TRUE
	status = ORGAN_PROSTHETIC
	var/obj/item/card/id/id_card = /obj/item/card/id/ascent
```

Also, the crystal flag comes with different damage handling. It could easily make some parts break when they shouldn't?

afterthought2: It's true that this proc maybe shouldn't crystalize robotic organs, etc. Worth looking over.

For clearing `status` here and in the other comment: I think it's pretty clear that doing `status = 0` is basically always wrong, if that's what you mean. Maybe it would be cleaner to separate the vars, but I'm not so sure about that; there are more dependencies between the flags than might first appear.

code/modules/organs/organ.dm
MistakeNot4892: The only changes I don't want from your PR are the ones I mentioned as blocking on Discord, which I was making this alternative PR for, specifically the subtyping of all organs and the removal of the post-species modifier proc. Everything else in your PR is fine, I don't disagree conceptually with any of your thoughts on the flag splitting/clustering.

code/modules/species/species.dm
MistakeNot4892: Any species with the flag expects to apply it to all of their organs including internal. The controller is fine as a crystalline organ, it means they can repair it with the tools on hand. Anything that wants to have non-crystalline organs should be handling it post-flag or not applying the flag at all.

The status flag stuff is a totally separate discussion to this PR. I am not opposed to reworking how flags and limb status are handled. I feel like I'm repeating myself a lot at the moment.

PR #2528, opened 2022-06-14
Fix dropping inventory on species change

PsyCommando:
## Description of changes
Now everything in non-existent slots gets dropped.

The old proc was dropping exclusively clothing, and it would obey user inventory access rules, which made some things undroppable. 

## Changelog
:cl:
fix: Fixed items staying in non-existent inventory slots when changing species.
/:cl:


afterthought2 (dismissed):

MistakeNot4892:
I believe this should be going to staging?

PsyCommando:
> I believe this should be going to staging?

I was afraid you'd say that like a second after I clicked the create PR button ^^;
Alright, lets do this then X,D

afterthought2 (approved):

PR #2530, opened 2022-06-14
Fix species with no blood [IDB IGNORE] [MDB IGNORE]

PsyCommando:
## Description of changes
Human mobs with no blood don't show up as if they lost all their blood on the health scanner.

## Changelog
:cl:
fix: Species with no circulatory system now properly say so in the health scanner report, and don't show up as being in shock constantly.
/:cl:


MistakeNot4892:
Please point this at staging.

PsyCommando:
Alright will do.

MistakeNot4892 (approved):

PR #2533, opened 2022-06-15
Inventory slots.

MistakeNot4892:
## Description of changes
This is a hefty one. 
- Removes all direct references to inventory stuff on mobs.
- Moves almost all inventory stuff to /mob/living or /mob.
- Expands the held item slot datums to work as generic inventory slots and plugs them in as appropriate.

## TODO ##
- [X] Implement slot datums.
- [X] Collapse procs.
- [X] Work out why inv slots are duplicated.
- [x] Work out a good place to refresh inv slots on the mob, using HUD instantiation sucks.
- [x] Work out why removing any item causes everything you're wearing to fall off.
- [x] Get basic inventory condensing/rewrite actually working in practice.
- [x] Get hand slots to generate and order dynamically.
- [x] Fix up hand slots so they aren't replacing themselves and losing their held item on client relog.
- [x] Check if species setting properly disposes/rebuilds slots.
- [x] Convert species gear lists to have appropriate ui locs.
- [x] Convert human examine() to be generic and use inv slots.
- [x] Convert nymph inventories to use a slot and bespoke behavior.
- [x] Get nymph hat wearing to work.
- [x] Restrict nymph ClickOn behavior to stop them using items held in their mouth.
- [x] Add nymph throw mode UI.
- [x] Add nymph drop button.
- [x] Fix hand slots :(
- [x] Convert stripping and equipping to mob level
- [ ] Sort out GC fails on inv slot datums.
- [x] Bring Ascent nymph code into line with doona nymphs.
- [x] Convert Ascent nymph health icons from the diona version.
- [ ] Fix grabs by nymphs not being droppable/showing in HUD.
- [x] Add held slot sorting so multihands arrange themselves nicely.
- [ ] Test and debug internals setting.
- [ ] Work out/fix reverse pickpocketing.
- [ ] Convert parrots to use inventories.
- [ ] Convert crows to use inventories.
- [ ] Convert synthetic inventories to use slots and bespoke behavior.
- [ ] Remove stub and properly fix linter fails from slot_tie_str attackby call.
- [ ] Maybe straight up merge the inventory slots with the screen elements and kill two birds with one stone.

Unifies inventory code in one spot and makes expanding it later trivial.

## Authorship
Myself.

## Changelog
TODO proper changelog.
- Diona nymphs have a throw and drop UI button, and their inventory and equip behavior has changed.
- Likewise with crows and possums.


code/modules/mob/living/living.dm
```diff
@@ -751,6 +746,8 @@ default behaviour is:
 			add_overlay(A)
 
 /mob/living/Destroy()
+	clear_inventory_slots()
+	. = ..()
```
out-of-phaze, line 727: Why are you calling the parent's Destroy() prior to removing auras? Shouldn't almost all Destroy() behavior on non-top-level types be done prior to the parent call?

code/modules/mob/living/living_inventory.dm
```diff
@@ -0,0 +1,99 @@
...
+		LAZYSET(inventory_slots, inv_slot.slot_id, inv_slot)
+
+/mob/living/remove_inventory_slot(var/slot_id)
+	var/datum/inventory_slot/inv_slot = LAZYACCESS(inventory_slots, slot_id)
+	if(inv_slot)
+		var/held = inv_slot.get_equipped_item()
+		if(held)
+			drop_from_inventory(held)
+		qdel(inv_slot)
+	LAZYREMOVE(inventory_slots, slot_id)
```
out-of-phaze, line 42: Does this ever remove it from `held_item_slots`? That feels like it'd cause hand slots to have a hanging reference that makes them fail GC.

out-of-phaze (changes requested):
This review is only addressing this todo item: `Sort out GC fails on inv slot datums.`

I think I found a potential cause, though.

code/modules/mob/living/living.dm
MistakeNot4892: Clearing the inventory slots here means they don't get deleted in the `/atom/movable` contents clearing logic.

code/modules/mob/living/living_inventory.dm
MistakeNot4892: It shouldn't ever be present in `held_item_slots` as far as I recall. This might be outdated/wrong, it has been awhile since I touched this logic, but the lists should be distinct.

code/_onclick/click.dm
```diff
@@ -48,6 +48,9 @@
 	* item/afterattack(atom,user,adjacent,params) - used both ranged and adjacent
 	* mob/RangedAttack(atom,params) - used only ranged, only used for tk and laser eyes but could be changed
 */
+/mob/proc/can_use_held_item(var/obj/item/held)
```
MistakeNot4892, line 51: Note to self to split this out.

code/modules/mob/living/living_inventory.dm
MistakeNot4892: `held_item_slot_strings` is a linear list of slot strings and doesn't hold any references.

MistakeNot4892:
This has been split up into several other PRs.

PR #2541, opened 2022-06-17
KeIgaras' fix PR.

MistakeNot4892:
Opening by request, alternative to other busted PR.
Original PR: #2387

### DIFF
--- a/code/game/objects/items/weapons/tools/weldingtool.dm
+++ b/code/game/objects/items/weapons/tools/weldingtool.dm
@@ -113,7 +113,7 @@
 		user.drop_from_inventory(src)
 		var/obj/item/flamethrower/F = new(get_turf(src), src)
 		user.put_in_hands(F)
-		return
+		qdel(src)		return
 
 	if (istype(W, /obj/item/welder_tank))
 		if(tank)
--- a/code/game/objects/items/weapons/tools/weldingtool.dm
+++ b/code/game/objects/items/weapons/tools/weldingtool.dm
@@ -109,11 +109,10 @@
 	if(!status && istype(W, /obj/item/stack/material/rods))
 		var/obj/item/stack/material/rods/R = W
 		R.use(1)
-		add_fingerprint(user)
 		user.drop_from_inventory(src)
-		var/obj/item/flamethrower/F = new(get_turf(src), src)
-		user.put_in_hands(F)
-		qdel(src)		return
+		user.put_in_hands(new /obj/item/flamethrower(get_turf(src), src))
+		qdel(src)
+		return TRUE
 
 	if (istype(W, /obj/item/welder_tank))
 		if(tank)
--- a/code/modules/atmospherics/components/unary/vent_scrubber.dm
+++ b/code/modules/atmospherics/components/unary/vent_scrubber.dm
@@ -77,8 +77,10 @@
 		icon_state = "weld"
 	else if((stat & NOPOWER) || !use_power)
 		icon_state = "off"
+	else if(scrubbing == "exchange")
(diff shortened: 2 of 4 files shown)


comma (dismissed):

code/modules/atmospherics/components/unary/vent_scrubber.dm
```diff
@@ -77,8 +77,10 @@
 		icon_state = "weld"
 	else if((stat & NOPOWER) || !use_power)
 		icon_state = "off"
+	else if(scrubbing == "exchange")
```
afterthought2: While it's true that the current code is nonsensical, there do seem to be defines you should be using for this.

code/game/objects/items/weapons/tools/weldingtool.dm
```diff
@@ -109,11 +109,12 @@
 	if(!status && istype(W, /obj/item/stack/material/rods))
 		var/obj/item/stack/material/rods/R = W
 		R.use(1)
-		add_fingerprint(user)
 		user.drop_from_inventory(src)
 		var/obj/item/flamethrower/F = new(get_turf(src), src)
+		F.add_fingerprint(user)
```
afterthought2: It's highly unlikely to be correct to manually add fingerprints like this.

afterthought2 (commented):
I find the blood data manipulation concerning: is this really the way to do it?

Other comments are obviously minor.

code/game/objects/items/weapons/tools/weldingtool.dm
keIgaras: now prints are also being added everywhere, for an alternative option it needs to be refactored

kelgaras: there is no choice, they are added so everywhere

code/modules/atmospherics/components/unary/vent_scrubber.dm
kelgaras: you mean, i need to swap ``"exchange"`` for ``SCRUBBER_EXCHANGE`` define?

code/game/objects/items/weapons/tools/weldingtool.dm
afterthought2: I'm not actually sure this one is replicated in equip code, but it likely should be.

kelgaras: looked at process ``put_in_hands``, ``drop_from_inventory``, and ``remove_from_mob`` and didn't found where fingerprints has be added

MistakeNot4892: Added an add_fingerprint() call to base item equipped().

code/modules/atmospherics/components/unary/vent_scrubber.dm
MistakeNot4892: Done.

MistakeNot4892:
Reviewed the blood data manipulation and I agree - the main issue appeared to be the original code returning early rather than trying to retrieve data if it wasn't provided, rather than the blood data being null. `reset_blood()` should be populating the mob's DNA hashes etc in their blood data already.

code/modules/atmospherics/components/unary/vent_scrubber.dm
afterthought2: If you did, you didn't push the commit.

MistakeNot4892: Shit

afterthought2 (approved):

PR #2543, opened 2022-06-18
Minimal support for mutable areas.

afterthought2:
## Description of changes

Attempts to add minimal support to changing the area of a turf during normal gameplay (as opposed to only in pre-init stages or during shuttle movement).

Likely breaks machinery-shuttle interactions. Has some minor side effects for non-movable machines being moved (e.g. by legacy effects like sm pull).

Deleting areas with turfs in them should be considered unsupported. Deleting areas without turfs should "work," though they will almost certainly not GC.

Blueprints clearly assume such support exists, whereas it does not. Note that this PR does not attempt to fix blueprints, though it will make them slightly less buggy some of the time.

## Authorship

me

### DIFF
--- a/code/datums/repositories/areas.dm
+++ b/code/datums/repositories/areas.dm
@@ -1,15 +1,9 @@
 var/global/repository/area/area_repository = new()
 
 /repository/area
-	var/list/by_name_coords_cache_data
-	var/list/by_name_cache_data
-	var/list/by_z_level_cache_data
-
-/repository/area/New()
-	by_name_coords_cache_data = list()
-	by_name_cache_data        = list()
-	by_z_level_cache_data     = list()
-	..()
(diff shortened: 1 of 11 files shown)


afterthought2:
Some minor testing suggests that this is at least somewhat functional with shuttles. Power (in particular, apcs) are reset on shuttle move, meaning they will briefly cycle to battery mode before starting to charge from whatever available powernet power there is. This is possibly undesirable but probably makes little difference in practice. Vents/scrubbers seem to operate, though more extensive testing would be useful here.

I would suggest merging this to dev (after code review) and then identifying any issues that are considered serious/disruptive.

MistakeNot4892 (approved):
I don't fully understand the repercussions of this PR but I'm willing to see what happens.

PR #2550, opened 2022-06-24
fix organs not being removed properly during surgery STAGING

PsyCommando:
Things would not get removed from the implants list on the second call to do_uninstall.

## Changelog
:cl:
fix: fixed organs not being removed correctly during surgery.
/:cl:


MistakeNot4892 (approved):
You get to resolve the conflict when we merge staging back to dev. >:[

PsyCommando:
> You get to resolve the conflict when we merge staging back to dev. >:[

Isn't it going to be essentially the exact same commit being merged?

PR #2552, opened 2022-06-25
Paperwork rework and QoL

PsyCommando:
## Description of changes
Depends on #2495 

A bunch of tweaks to bring the very dated stuff in the paperwork items up to date along with a lot of quality of life updates.
* Adds paper material
* Makes paper, carbon paper, folder, paper_bundle, clipboard handle interactions like other non nanoui things.
* Remade tape rolls into stacks.
* Barricade tape rolls now place down structures instead of items behaving like structures. 
* Barricade tape structure now is handling neighbors and its icon state better.
* Barricade tape types details are now stored into some /decl
* Slightly tweaked barricade tape icons to have more directional range and look better with neighbours.
* Made the hand labeler refillable, and use the alt interaction system.
* And a lot more small things.

## Changelog
:cl:
fixed: Photocopier now works, can be built, and has a new UI.
tweak: Tweaked police tape. Now called barricade tape, and the tape itself now handles neighbours better.
tweak: Photo cameras now have skilled interactions. And can eject their film cartridge.
tweak: Hand labeler can now be refilled with paper sheets, or paper material sheets. It can also now remove labels. And has a new menu for changing mode.
tweak: Fax machine is now buildable, and make use of the network system.
tweak: Fax machine has a new UI, and new functionalities. Like history logging and contact saving. The data is saved to a data disk and can be transferred from machine to machine. 
tweak: Numerous paperwork simple browser UIs have had some touch up and polish. 
added: Paper material.
added: Paper refill bundle.
added: Paper planes.
added: Material cuttings.
added: Printer stock part. (Machine version)
added: Disk drive stock part. (Machine version)
added: Card reader stock part. (Machine version)
/:cl:


PsyCommando:
I'm pretty satisfied with how things work now. So I'll set it up for review. 
There's still a couple of things I need to look over again, but it should be mostly good to go.

MistakeNot4892:
Jesus Christ that diff size.
Will try to review, but you might want to break this into a series of PRs.

maps/ministation/ministation_objects.dm
```diff
@@ -86,4 +86,30 @@
...
+	preset_channels = list("Satellite")
+
+//Detective bs
+/obj/item/camera_film/high
+	name = "high capacity film cartridge"
+	max_uses  = 30
+	uses_left = 30
+
+/obj/item/camera/detective
+	name = "detectives camera"
```
MistakeNot4892: `detective's`

maps/away/liberia/liberia.dmm
```diff
@@ -5019,10 +5019,8 @@
 /area/liberia/bridge)
 "jD" = (
 /obj/structure/table/glass,
-/obj/machinery/photocopier/faxmachine{
-	department = "Merchant Outpost";
+/obj/machinery/faxmachine{
 	req_access = list("ACCESS_MERCHANT");
```
MistakeNot4892, line 5023: Why keep the req_access here?

code/modules/persistence/noticeboards.dm
```diff
@@ -168,37 +171,32 @@
...
+	P = new(src, null, 
+		"how many times do i have to tell you people, these xeno-arch samples are del-i-cate, and should be handled so! careful application of a focussed, concentrated heat or some corrosive liquids should clear away the extraneous carbon matter, while application of an energy beam will most decidedly destroy it entirely - like someone did to the chemical dispenser! W, <b>the one who signs your paychecks</b>",
+		"READ ME! Before you people destroy any more samples")
+	P.apply_custom_stamp(overlay_image('icons/obj/bureaucracy.dmi', icon_state = "paper_stamped_rd", flags = RESET_COLOR), "Research Department")
 	add_paper(P, skip_icon_update = TRUE)
 
-	P = new()
-	P.SetName("Reminder regarding the anomalous material suits")
-	P.info = "Do you people think the anomaly suits are cheap to come by? I'm about a hair trigger away from instituting a log book for the damn things. Only wear them if you're going out for a dig, and for god's sake don't go tramping around in them unless you're field testing something, R"
-	P.stamped = list(/obj/item/stamp/rd)
```
MistakeNot4892, line 190: Note to self to convert these into secrets for the example secrets folder.

code/modules/persistence/noticeboards.dm
```diff
@@ -61,62 +65,60 @@
...
 
-/obj/structure/noticeboard/on_update_icon()
-	..()
-	icon_state = "[base_icon_state][LAZYLEN(notices)]"
+/obj/structure/noticeboard/handle_default_screwdriver_attackby(mob/user, obj/item/screwdriver/S)
+	var/choice = input("Which direction do you wish to place the noticeboard?", "Noticeboard Offset") as null|anything in list("North", "South", "East", "West")
+	if(choice && S.do_tool_interaction(TOOL_SCREWDRIVER, user, src, 1 SECOND))
+		default_pixel_x = 0
+		default_pixel_y = 0
+		switch(choice)
```
MistakeNot4892: Please preserve the directional setting behavior, if only for consistency with things like persistence noteboards.

code/datums/trading/traders/goods.dm
```diff
@@ -207,7 +207,7 @@ Sells devices, odds and ends, and medical stuff
 								/obj/item/uv_light                  = TRADER_THIS_TYPE,
 								/obj/item/mmi                       = TRADER_ALL,
 								/obj/item/robotanalyzer             = TRADER_THIS_TYPE,
-								/obj/item/toner                     = TRADER_THIS_TYPE,
+								/obj/item/toner_cartridge                     = TRADER_THIS_TYPE,
```
MistakeNot4892: Spacing oh no

code/datums/uplink/devices and tools.dm
```diff
@@ -15,7 +15,7 @@
 	name = "Duct Tape"
 	desc = "A roll of duct tape. changes \"HELP\" into sexy \"mmm\"."
```
MistakeNot4892: Christ. While you're in the region could you make this less awful?

code/game/gamemodes/cult/talisman.dm
```diff
@@ -1,7 +1,9 @@
 /obj/item/paper/talisman
-	icon_state = "paper_talisman"
-	var/imbue = null
 	info = "<center><img src='talisman.png'></center><br/><br/>"
+	var/imbue = null
+
+/obj/item/paper/talisman/update_contents_overlays()
+	add_overlay(global.cached_paper_overlays["paper_talisman"])
```
MistakeNot4892: Not sure I like this. Should ideally use a getter, or check that there's something cached before passing it to add_overlay().

code/game/machinery/_machines_base/machinery.dm
```diff
@@ -329,7 +329,14 @@ Class Procs:
 		text = "\The [src] pings."
 
 	state(text, "blue")
-	playsound(src.loc, 'sound/machines/ping.ogg', 50, 0)
+	playsound(src.loc, 'sound/machines/ping.ogg', 50, FALSE)
+
+/obj/machinery/proc/buzz(text=null)
```
MistakeNot4892: `text` without `=null` is equivalent.

code/game/objects/item.dm
```diff
@@ -992,3 +985,11 @@ modules/mob/living/carbon/human/life.dm if you die, you will be zoomed out.
 
 /obj/item/proc/gives_weather_protection()
 	return FALSE
+
+//#TODO: Stub implementation. Probably should be unified into /obj along with health..
+/obj/item/proc/take_damage(var/amount, var/damtype, var/silent = FALSE)
```
MistakeNot4892, line 1009: This is at least a step closer to unification.

code/game/objects/item_materials.dm
```diff
@@ -1,10 +1,10 @@
 /obj/item/on_update_icon()
-	overlays.Cut()
+	cut_overlays()
```
MistakeNot4892: How certain are you that you got every /obj/item override of on_update_icon()?

code/game/objects/items/stacks/stack.dm
```diff
@@ -258,9 +258,7 @@
 
 //creates a new stack with the specified amount
 /obj/item/stack/proc/split(var/tamount, var/force=FALSE)
-	if (!amount)
-		return null
-	if(uses_charge && !force)
+	if (!can_split() || !amount || (uses_charge && !force))
```
MistakeNot4892: Shouldn't `can_split()` already check `amount`?

MistakeNot4892:
53/127 files in. I would strongly suggest splitting the repathing and tape stuff into its own PR, if nothing else. The /obj/item/ update icon change also worries me, since every override down from that needs to make sure it uses the same procs, or the overlays will break. Not reviewing further tonight as my eyes are crossing.

code/datums/uplink/devices and tools.dm
PsyCommando: Wow, that's a description I guess. 
I'll slap a duck joke over it.

maps/away/liberia/liberia.dmm
PsyCommando: Well, access is still used for being able to use the thing. And that was what it was set to for this map.

code/modules/persistence/noticeboards.dm
PsyCommando: Well, the issue is that when you build one, it just floats middair. A bunch of wall mounted things are like that too. Machinery has its frame installation code taking care of it, but structures don't.

code/datums/trading/traders/goods.dm
PsyCommando: gawd, that's definitely something using tabs for spacing. -_-

code/game/objects/item_materials.dm
PsyCommando: Well, most of the things using the old overlay method basically don't call their parent.

But yeah it could mess with stuff's overlays. I could take a deeper look and see if anything in particular does it.

code/game/objects/items/stacks/stack.dm
PsyCommando: Well, when I thought about it, it was just a capability check. Aka "can you ever split a stack of this type". But I guess with there being a split proc I could always just merge the whole conditional statement into it as well?

code/modules/persistence/noticeboards.dm
MistakeNot4892: Noticeboards should be offsetting in set_dir, it was part of a persistence fix PR I did months ago.

PsyCommando: Alright, I guess I can do it in set_dir.

code/game/gamemodes/cult/talisman.dm
PsyCommando: I can do that.

code/game/machinery/_machines_base/machinery.dm
PsyCommando: Probably some leftover from copy pasting ping..

code/game/objects/item.dm
PsyCommando: Hopefully XD

PsyCommando:
> 53/127 files in. I would strongly suggest splitting the repathing and tape stuff into its own PR, if nothing else. The /obj/item/ update icon change also worries me, since every override down from that needs to make sure it uses the same procs, or the overlays will break. Not reviewing further tonight as my eyes are crossing.

Icon update issues was solved in #2608. 
Also, at this point if I have to spend a week splitting this into separate PRs for you to merge it I think I'll just pass.

afterthought2:
Discussed this briefly; it's just too big in this state to reasonably review. Try splitting out some isolated content from it. Alternatively reorganize the commit structure so each commit corresponds to a separate change/feature. @MistakeNot4892 and I will prioritize reviewing PRs related to this, I know it's been sitting for a bit.

PsyCommando:
> Discussed this briefly; it's just too big in this state to reasonably review. Try splitting out some isolated content from it. Alternatively reorganize the commit structure so each commit corresponds to a separate change/feature. @MistakeNot4892 and I will prioritize reviewing PRs related to this, I know it's been sitting for a bit.

Commits already correspond to features/changes. You want me to squash all related stuff together?
But whatever, I split off the tape and paper shredder stuff. The rest is just too tightly coupled together because of the old paperwork code being a spaghetti.

MistakeNot4892:
We don't have a conflict label currently so I'm using the rebase one.

PsyCommando:
Squashed a bunch of commits together.

PsyCommando:
Desperado is crashing unit tests again. So I got sick of it and made a PR to fix it: #2670

code/modules/persistence/noticeboards.dm
MistakeNot4892: This hasn't been actioned yet.
For reference:
````
/obj/structure/noticeboard/set_dir(var/ndir)
	. = ..()
	if(dir & SOUTH)
		default_pixel_y = 32
	else // NORTH is also 0-offset due to the icon.
		default_pixel_y = 0
	if(dir & WEST)
		default_pixel_x = 32
	else if(dir & EAST)
		default_pixel_x = -32
	else
		default_pixel_x = 0
	reset_offsets(0)
````
It's possible this needs to be called in Initialize() to fix the spawn issue.

**Comment by MistakeNot4892** [comment]
Sorry, this is still really hard to review. I can't review it commit per commit because the final state is the changeset that is actually going in, and this is like six different rewrites at once. I don't know if I'll manage to review it fully at all given time and effort constraints. Maybe @afterthought2 or @comma will have more success.

File: code/game/machinery/_machines_base/stock_parts/card_reader.dm [hunk 12]
```diff
@@ -0,0 +1,120 @@
...
+
+/obj/item/stock_parts/card_reader/proc/insert_card(var/obj/item/card/C, var/mob/user)
+	if(inserted_card)
+		if(user)
+			to_chat(user, SPAN_WARNING("There is already a card in \the [src]!"))
+		return
+
+	if(user)
+		to_chat(user, SPAN_NOTICE("You insert \the [C] into \the [src]."))
+		user.unEquip(C, src)
```
afterthought2: You have to check the return value and exit if it fails.

code/game/machinery/_machines_base/stock_parts/disk_reader.dm
```diff
@@ -0,0 +1,104 @@
+/obj/item/stock_parts/disk_reader
```
afterthought2: Combine/subtype with the card reader.

code/modules/fabrication/designs/imprinter/designs_misc_circuits.dm
```diff
@@ -477,4 +477,10 @@
 	path = /obj/item/stock_parts/circuitboard/inertial_damper
 
 /datum/fabricator_recipe/imprinter/circuit/docking_beacon
-	path = /obj/item/stock_parts/circuitboard/docking_beacon
\ No newline at end of file
+	path = /obj/item/stock_parts/circuitboard/docking_beacon
+
+/datum/fabricator_recipe/imprinter/circuit/fax_machine
+	path = /obj/item/stock_parts/circuitboard/fax_machine
```
afterthought2, line 483: Historically these were deliberately unconstructable due to poor gameplay patterns.

code/modules/paperwork/carbonpaper.dm
```diff
@@ -1,53 +1,43 @@
...
+	LAZYSET(copy.metadata, "is_copy", TRUE)
+	copy.set_color("#ffccff")
+
+	//Silly hack to make all the text grayscale, since nobody is using standard css classes for pens we could override instead. Also, all using deprecated tags as well.
+	var/copycontents = copy.info
+	copycontents = replacetext(copycontents, "<font face=\"[original.deffont]\" color=", "<font face=\"[original.deffont]\" nocolor=")
+	copycontents = replacetext(copycontents, "<font face=\"[original.crayonfont]\" color=", "<font face=\"[original.crayonfont]\" nocolor=")
+	copy.set_content("<font color = #101010>[copycontents]</font>", "Copy - [original.name]")
 
+	user.unEquip(src) //#TODO: Set playdropsound to false whenever it gets merged
```
afterthought2: Either check return value and stop on fail, qdel yourself first, or use `drop_from_inventory`.

code/modules/paperwork/clipboard.dm
```diff
@@ -72,16 +72,26 @@
 			return
 		push_paper(W)
 		to_chat(user, SPAN_NOTICE("You clip the [W] onto \the [src]."))
+		if(user.machine)
+			attack_self(user)
```
afterthought2: I don't know what this check means; `user.machine` is completely arbitrary and internal.

code/modules/paperwork/clipboard.dm
```diff
@@ -72,16 +72,26 @@
...
 		update_icon()
 		return TRUE
 
-	else if(top_paper)
-		top_paper.attackby(W, user)
+	else if(top_paper?.attackby(W, user))
+		if(user.machine)
+			attack_self(user)
```
afterthought2: Same here.

code/modules/paperwork/clipboard.dm
```diff
@@ -94,42 +104,45 @@
...
 		if(i == 1)
-			dat += "<A href='?src=\ref[src];write=\ref[P]'>Write</A> "
-		dat += "<A href='?src=\ref[src];remove=\ref[P]'>Remove</A> <A href='?src=\ref[src];rename=\ref[P]'>Rename</A></DIV><BR>"
+			dat += "<TD/><TD><A href='?src=\ref[src];write=\ref[P]'>Write</A></TD>"
+		else
+			dat += "<TD/><TD/>"
+		dat += "<TD><A href='?src=\ref[src];remove=\ref[P]'>Remove</A></TD><TD><A href='?src=\ref[src];rename=\ref[P]'>Rename</A></TD></TR>"
+	dat += "</TABLE>"
 
+	user.set_machine(src)
```
afterthought2, line 125: Do not do this manually.

code/modules/paperwork/clipboard.dm
```diff
@@ -94,42 +104,45 @@
...
 /obj/item/clipboard/proc/remove_pen(var/mob/user)
-	if(stored_pen && user.put_in_hands(stored_pen))
-		to_chat(user, SPAN_NOTICE("You pull your trusty [stored_pen] from your [src]."))
+	if(stored_pen && user.get_empty_hand_slot())
+		to_chat(user, SPAN_NOTICE("You pull \the [stored_pen] from your [src]."))
+		user.put_in_hands(stored_pen)
 		. = stored_pen
 		stored_pen = null
+		if(user.machine)
+			attack_self(user)
```
afterthought2: Same here.

code/modules/paperwork/faxmachine.dm
```diff
@@ -1,235 +1,675 @@
...
+	. = ..()
+	if(. != INITIALIZE_HINT_QDEL)
+		global.allfaxes += src
+		set_extension(src, /datum/extension/network_device/fax)
+		if(ml && printer)
+			printer.make_full()
+
+/obj/machinery/faxmachine/Destroy()
+	global.allfaxes -= src
 	. = ..()
```
afterthought2, line 103: Are all the component references wiped via callbacks? Would suggest avoiding these component references.

code/modules/paperwork/faxmachine.dm
```diff
@@ -1,235 +1,675 @@
...
+		/decl/stock_part_preset/access_lock, //Empty access list
+	)
+	uncreated_component_parts = null
+	public_methods = list(
+		/decl/public_access/public_method/fax_receive_document,
+	)
+	var/obj/item/stock_parts/disk_reader/disk_reader  //Cached ref to the disk_reader component. Used for handling data disks.
+	var/obj/item/stock_parts/card_reader/card_reader  //Cached ref to the card_reader component. Used for scanning a user's id card.
+	var/obj/item/stock_parts/printer/printer          //Cached ref to the printer component. Used for printing things.
+	var/obj/item/scanner_item                         //Item to fax
```
afterthought2, line 81: Noting to check possible GC issues.

code/modules/paperwork/faxmachine.dm
```diff
@@ -1,235 +1,675 @@
...
+	// --- Sending Fax ---
+	if(href_list["send"])
+		if(!scanner_item)
+			return TOPIC_NOACTION
+		if(world.timeofday < time_cooldown_end)
+			to_chat(user, SPAN_WARNING("\The [src] isn't ready yet for sending again! [max(0, time_cooldown_end - world.timeofday) / (1 SECOND)] second\s left."))
+			return TOPIC_REFRESH
+
+		//Prioritize the quick dial value
+		dest_uri = replacetext(sanitize(href_list["quick_dial"]), " ", "_")
```
afterthought2: Are you sure the html encoding state here is valid for sanitize use?

code/modules/paperwork/faxmachine.dm
```diff
@@ -1,235 +1,675 @@
...
-			if (has_access(send_access, scan.GetAccess()))
-				authenticated = 1
+	// --- Quick Dial Operations  ---
+	if(href_list["add_qd"])
+		var/qduri = uppertext(replacetext(sanitize(href_list["network_uri"]), " ", "_"))
+		if(!length(qduri))
+			to_chat(user, SPAN_WARNING("Please specify a destination URI!"))
+			return TOPIC_NOACTION
+
+		var/inputname = input(user, "Name for the new quick dial contact?", "New quick dial contact")
```
afterthought2, line 306: Check interactivity status after input.

code/modules/paperwork/faxmachine.dm
```diff
@@ -1,235 +1,675 @@
...
+
+/obj/machinery/faxmachine/proc/insert_scanner_item(var/obj/item/I, var/mob/user)
+	if(scanner_item)
+		if(user)
+			to_chat(user, SPAN_WARNING("\The [src] already has something being scanned!"))
+		return FALSE
+	
+	if(user)
+		to_chat(user, SPAN_NOTICE("You place \the [I] into \the [src]'s scanner."))
+		user.unEquip(I, src)
```
afterthought2: Check return value and exit if failed.

code/modules/paperwork/faxmachine.dm
```diff
@@ -1,235 +1,675 @@
...
+
+/obj/machinery/faxmachine/proc/eject_scanner_item(var/mob/user)
+	if(!scanner_item)
+		if(user)
+			to_chat(user, SPAN_WARNING("There's nothing to eject from \the [src]."))
+		return FALSE 
+	
+	if(user)
+		to_chat(user, SPAN_NOTICE("You eject \the [scanner_item] from \the [src]."))
+		user.put_in_hands(scanner_item)
```
afterthought2, line 372: Not 100% certain but this proc appears accessible from Topic. If that's the case, you must do a CanPhysicallyInteract or adjacent check before `put_in_hands`.

code/modules/paperwork/faxmachine.dm
```diff
@@ -1,235 +1,675 @@
...
+	//Tell the printer to do its job
+	printer.start_printing_queue()
+	use_power_oneoff(active_power_usage)
+	var/datum/extension/network_device/fax/ND = get_extension(src, /datum/extension/network_device)
+	log_history("Incoming", "'[incoming]', from '[source_name]'@'[source_URI]' to '[get_area(src)]'s [src]'@'[ND.get_network_URI()]'.")
+	return TRUE
+
+/**Cause the nanoui to update, also updates the icon of the machine. */
+/obj/machinery/faxmachine/proc/update_ui()
+	log_debug("updating [src] ui and icon")
```
afterthought2: Left from testing?

code/modules/paperwork/folders.dm
```diff
@@ -1,118 +1,95 @@
...
-	for(var/obj/item/paper/P in src)
-		dat += "<A href='?src=\ref[src];remove=\ref[P]'>Remove</A> <A href='?src=\ref[src];rename=\ref[P]'>Rename</A> - <A href='?src=\ref[src];read=\ref[P]'>[P.name]</A><BR>"
-	for(var/obj/item/photo/Ph in src)
-		dat += "<A href='?src=\ref[src];remove=\ref[Ph]'>Remove</A> <A href='?src=\ref[src];rename=\ref[Ph]'>Rename</A> - <A href='?src=\ref[src];look=\ref[Ph]'>[Ph.name]</A><BR>"
-	for(var/obj/item/paper_bundle/Pb in src)
-		dat += "<A href='?src=\ref[src];remove=\ref[Pb]'>Remove</A> <A href='?src=\ref[src];rename=\ref[Pb]'>Rename</A> - <A href='?src=\ref[src];browse=\ref[Pb]'>[Pb.name]</A><BR>"
+	for(var/obj/item/I in src)
+		dat += "<A href='?src=\ref[src];remove=\ref[I]'>Remove</A> <A href='?src=\ref[src];rename=\ref[I]'>Rename</A> - <A href='?src=\ref[src];examine=\ref[I]'>[I.name]</A><BR>"
+
+	user.set_machine(src)
```
afterthought2, line 64: Don't do this manually.

code/modules/paperwork/folders.dm
```diff
@@ -1,118 +1,95 @@
...
+	return TRUE
 
-/obj/item/folder/Topic(href, href_list)
-	..()
-	if((usr.stat || usr.restrained()))
-		return
+/obj/item/folder/OnTopic(mob/user, href_list, datum/topic_state/state)
+	if(href_list["remove"])
+		var/obj/item/P = locate(href_list["remove"])
+		user.put_in_hands(P)
```
afterthought2, line 75: Check PhysicallyInteract or similar.

code/modules/paperwork/handlabeler.dm
```diff
@@ -1,66 +1,279 @@
...
+		HAND_LABELER_SAFETY_TOGGLE    = btn_power,
+		HAND_LABELER_MODE_ADD_NAME    = btn_set_label,
+		HAND_LABELER_MODE_REM_NAME    = btn_rem_one,
+		HAND_LABELER_MODE_REMALL_NAME = btn_rem_all,
+	)
+	return show_radial_menu(user, user, choices, use_labels = TRUE)
+
+/obj/item/hand_labeler/attack_self(mob/user)
+	var/choice = show_action_radial_menu(user)
+	switch(choice)
```
afterthought2, line 125: May need interactivity check here.

code/modules/paperwork/handlabeler.dm
```diff
@@ -1,66 +1,279 @@
...
+		if(((incoming_amt + current_amt) / LABEL_MATERIAL_COST) > max_labels)
+			to_chat(user, SPAN_WARNING("There's not enough room in \the [src] for the [label_added] label(s) \the [P] is worth."))
+			return
+		if(!user.do_skilled(2 SECONDS, SKILL_LITERACY, src) || (QDELETED(W) || QDELETED(src)))
+			return
+
+		to_chat(user, SPAN_NOTICE("You slice \the [P] into [label_added] small strips and insert them into \the [src]'s paper feed."))
+		add_paper_labels(label_added)
+		user.unEquip(W)
+		qdel(W)
```
afterthought2, line 174: Do not unequip, just delete.

code/modules/paperwork/helpers.dm
```diff
@@ -0,0 +1,73 @@
+/**Tries to find a readily accessible pen in the user's held items, and in some of its inventory. */
+/proc/get_accessible_pen(var/mob/user)
```
afterthought2: Why is this not a mob proc?

code/modules/paperwork/helpers.dm
```diff
@@ -0,0 +1,73 @@
...
+/proc/clone_paper_work_item(var/obj/item/I)
+	if(istype(I, /obj/item/paper))
+		var/obj/item/paper/P = I
+		return P.Clone()
+	if(istype(I, /obj/item/photo))
+		var/obj/item/photo/P = I
+		return P.Clone()
+	if(istype(I, /obj/item/paper_bundle))
+		var/obj/item/paper_bundle/P = I
+		return P.Clone()
```
afterthought2, line 47: Why is this not an instance proc with overrides, or alternatively an extension?

code/modules/paperwork/paper.dm
```diff
@@ -1,160 +1,189 @@
...
+
+/**Caches the paper overlays/underlays we use frequently.*/
+/proc/cache_paper_overlays()
+	if(LAZYLEN(cached_paper_overlays))
+		return
+	LAZYSET(cached_paper_overlays, "paper_words",       overlay_image('icons/obj/bureaucracy.dmi', icon_state = "paper_words"))
+	LAZYSET(cached_paper_overlays, "paper_stack_words", overlay_image('icons/obj/bureaucracy.dmi', icon_state = "paper_stack_words"))
+	LAZYSET(cached_paper_overlays, "paper_talisman",    overlay_image('icons/obj/bureaucracy.dmi', icon_state = "paper_talisman", flags = RESET_COLOR))
+	LAZYSET(cached_paper_overlays, "sticky_words",      overlay_image('icons/obj/stickynotes.dmi', icon_state = "paper_words"))
+	LAZYSET(cached_paper_overlays, "photo_underlay",    overlay_image('icons/obj/bureaucracy.dmi', icon_state = "photo"))
```
afterthought2: What's the point of this? SSoverlays probably will cache their appearances already, and even if not these images are not expensive.

code/modules/paperwork/paper.dm
```diff
@@ -1,160 +1,189 @@
...
+	P.updateinfolinks()
+	P.update_icon()
+	return P
+
+/obj/item/paper/get_matter_amount_modifier()
+	return 0.2
 
 /obj/item/paper/proc/set_content(text,title)
-	set waitfor = FALSE
+	set waitfor = FALSE //Uh, why?
```
afterthought2: Indeed, why?

code/modules/paperwork/paper.dm
```diff
@@ -1,160 +1,189 @@
...
+			is_honking = TRUE
+			playsound(loc, 'sound/items/bikehorn.ogg', 50, TRUE)
 			spawn(20)
-				spam_flag = 0
+				is_honking = FALSE
+	return TRUE
 
 /obj/item/paper/attack_ai(mob/living/silicon/ai/user)
-	show_content(user)
+	interact(user)
```
afterthought2: Everyone else is read-only but not us?

code/modules/paperwork/paper.dm
```diff
@@ -279,212 +306,265 @@
...
+/obj/item/paper/proc/try_bundle_with(var/obj/item/paper/other, var/mob/user)
+	if(!can_bundle_with(other))
+		if(user)
+			to_chat(user, SPAN_WARNING("You can't bundle those!"))
+		return
+
+	var/obj/item/paper_bundle/B = new(loc)
+	if(user)
+		user.unEquip(src,   B)
+		user.unEquip(other, B)
```
afterthought2, line 413: Thou shalt never invoke `unEquip` without checking the return value.

code/modules/paperwork/paper_bundle.dm
```diff
@@ -1,79 +1,178 @@
...
+	pickup_sound      = 'sound/foley/paperpickup2.ogg'
+	item_flags        = ITEM_FLAG_CAN_TAPE
+	health            = 10
+	max_health        = 10
+	var/tmp/cur_page  = 1           // current page
+	var/tmp/max_pages = 100         //Maximum number of papers that can be in the bundle
+	var/list/pages                  // Ordered list of pages as they are to be displayed. Can be different order than src.contents.
+	var/static/list/cached_overlays //Cached images used by all paper bundles for generating the overlays and underlays
+
+/**Creates frequently used images globally, so we can re-use them. */
```
afterthought2, line 29: Again, if you do this right SSoverlays will cache them for you. Even if not, it's not worth caching images.

code/modules/paperwork/paper_bundle.dm
```diff
@@ -98,145 +197,354 @@
...
-		dat += "<div> <img src='tmp_photo.png' width = '180'[P.scribble ? "<div> Written on the back:<br><i>[P.scribble]</i>" : null ]</body></html>"
+		dat += {"
+			<html><head><title>[P.name]</title></head><body style='overflow:hidden'>
+			<div> <img src='tmp_photo.png' width = '180'[P.scribble ? "<div> Written on the back:<br><i>[P.scribble]</i>" : null ]</body></html>
+		"}
 		send_rsc(user, P.img, "tmp_photo.png")
-		show_browser(user, JOINTEXT(dat), "window=[name]")
+		show_browser(user, dat, "window=[name]")
+		onclose(user, name)
+	user.set_machine(src)
```
afterthought2, line 257: Don't do this manually. Also: what do we need onclose for?

code/modules/paperwork/paper_bundle.dm
```diff
@@ -98,145 +197,354 @@
...
+
+	//Merge lone paper
+	if(istype(I, /obj/item/paper) || istype(I, /obj/item/photo))
+		var/obj/item/paper/P = I
+		if(is_full() || (istype(P) && !P.can_bundle())) //Only paper check if they can be bundled apparently
+			return FALSE
+		//Merge the thing
+		insert_sheet_at(user, I, at_hint)
+		if(user)
+			user.unEquip(I, src)
```
afterthought2: Check return.

code/modules/paperwork/photocopier.dm
```diff
@@ -1,223 +1,301 @@
...
+	var/tmp/busy          = FALSE              //Whether we should allow people to mess with the settings and contents
+	var/accept_refill     = FALSE              //Whether we should handle attackby paper to be sent to the paper bin, or to the scanner slot
+	var/total_printing    = 0                  //The total number of pages we are printing in the current run
 
-/obj/machinery/photocopier/interface_interact(mob/user)
-	interact(user)
+/obj/machinery/photocopier/Initialize(ml)
+	. = ..()
+	if(.!= INITIALIZE_HINT_QDEL && ml && printer)
+		//Mapped photocopiers shall spawn with ink and paper
```
afterthought2, line 50: If so, the correct check is the `populate_parts` arg, not mapload.

code/modules/paperwork/photocopier.dm
```diff
@@ -1,223 +1,301 @@
...
+	obj_flags             = OBJ_FLAG_ANCHORABLE
+	construct_state       = /decl/machine_construction/default/panel_closed
+	maximum_component_parts = list(
+		/obj/item/stock_parts/printer = 1,
+		/obj/item/stock_parts         = 10,
+	)
+	uncreated_component_parts = null
+	var/tmp/insert_anim   = "photocopier_animation"
+	var/obj/item/scanner_item                  //what's in the scanner
+	var/obj/item/stock_parts/printer/printer   //What handles the printing queue
```
afterthought2, line 41: Check GC handling.

code/modules/paperwork/photocopier.dm
```diff
@@ -1,223 +1,301 @@
...
+
+	else if(istype(I, /obj/item/photo))
+		var/obj/item/photo/Ph = I
+		LAZYADD(., Ph.Clone())
+
+	else if(istype(I, /obj/item/paper_bundle))
+		var/obj/item/paper_bundle/B = I
+		for(var/obj/item/page in B.pages)
+			LAZYADD(., scan_item(page))
+
```
afterthought2: Another thing that could be handled via the helper or an extension.

afterthought2 (commented):
Hard to review. Seems like the natural structure for some of this is an extension. I don't like the explicit refs on machines to some of their parts: they make GC bookkeeping needlessly complicated. Just retrieve the part, or write a getter proc if you use it a lot.

code/game/machinery/_machines_base/stock_parts/card_reader.dm
PsyCommando: Nice catch. I didn't see that one..

code/game/machinery/_machines_base/stock_parts/disk_reader.dm
PsyCommando: That's kind of vague?

code/modules/paperwork/clipboard.dm
PsyCommando: Currently, that's the way most things that don't use nano check if they got someone using them. Its pretty awful, but that's the only way you can do it currently. 
I had a rework for properly handling users that was a lot less gross in the works, but I didn't really get around to finishing it, and this PR is already enormous, soooo.

code/modules/paperwork/clipboard.dm
PsyCommando: I need to do it manually? Take a look at the code behind it? Its nothing new, look at the code for ``/obj/item/paicard`` for instance. 
If you wanna fix it and make a better system for keeping track of open non-nano guis then be my guest? But this isn't in the scope.

code/modules/paperwork/faxmachine.dm
PsyCommando: Not sure what you mean by that? Sanitize just gets rid of all sort of crap character a user might input?

code/modules/paperwork/faxmachine.dm
PsyCommando: Yeah, that's a good catch!

code/modules/paperwork/faxmachine.dm
PsyCommando: Take a look at line 198. Proximity is already checked in the topic state.

code/modules/paperwork/faxmachine.dm
PsyCommando: Woops, yeah. ^^;

code/modules/paperwork/handlabeler.dm
PsyCommando: attack_self is validated for proximity, restrains, incapacitation.

code/modules/paperwork/handlabeler.dm
PsyCommando: That would cause a bug. Qdeling something in the hands slot without dropping would cause some item to persist. 
I'm not aware if it was fixed since then.

code/modules/paperwork/helpers.dm
PsyCommando: Because why?
That's an awfully niche proc to have on the base mob?

code/modules/paperwork/helpers.dm
PsyCommando: Look at where its used?
The admin verb needs it as well. Don't think overrides or extensions works too well for global procs..

code/modules/paperwork/paper.dm
PsyCommando: I literally have no ideas why they did this. :/

code/modules/paperwork/paper.dm
PsyCommando: Not entirely sure about this? Its been a long time. I know that at the time I had been told to cache things on another PR. Not sure what's the maintainer team's consensus on caching?
But yeah, using add_overlay with the state name should be sufficient for a lot of those.

code/modules/paperwork/paper.dm
PsyCommando: What? attack_self isn't read only either?

code/modules/paperwork/faxmachine.dm
PsyCommando: Gonna clear on destroy.

code/modules/paperwork/faxmachine.dm
PsyCommando: kk

code/modules/paperwork/paper.dm
PsyCommando: True,. Handling the error condition in this case is a bit annoying though. 
I'll just pop a generic warning that you can't drop a paper or something I guess.

code/modules/paperwork/paper_bundle.dm
PsyCommando: Paper bundles are a bit different though. Pretty sure there's a gain to using only a single source icon when used with mutable appearance.

code/modules/paperwork/paper_bundle.dm
PsyCommando: Look at how set_machine works. Then look at what the verb called by byond when onclose is set does.

code/modules/paperwork/clipboard.dm
PsyCommando: Also take a look at ``/obj/proc/updateUsrDialog()`` and ``/obj/proc/updateDialog()``.
That's how browser windows that aren't nano handle state updates without just calling attack_self willy-nilly.

code/modules/paperwork/photocopier.dm
PsyCommando: kk

code/modules/paperwork/photocopier.dm
PsyCommando: True

code/modules/paperwork/photocopier.dm
PsyCommando: Gotcha

code/modules/paperwork/folders.dm
PsyCommando: I'll set a better default topic state instead.

code/modules/paperwork/clipboard.dm
afterthought2: If you need to actually check this, store a weakref to the user in CouldUseTopic or interact, unwrap here, listen for on-close and remove the weakref there. But generally there is no need to check for this, you can format the interaction flow to just not need to know whether the UI is open.

The absolute bare minimum here is to check `user.machine == src`, checking `user.machine` is meaningless. But I'd encourage just reconsidering these checks in general.

code/modules/paperwork/clipboard.dm
afterthought2: UIs that do this do not manually set machine for the most part (yes, I know there are random exceptions). Machinery, the level at which `user.machine` is somewhat standardized, handle setting machine and searching for users using `CouldUseTopic`, the on-close listener, etc. In particular they are somewhat accurate at _clearing_ this var, which is relevant for GC. If you really need this functionality, you may wish to move that handling up a level or abstracting it to a flag, extension, etc. I don't like lots of ad hoc setting of it because there are in fact some best practices to follow here.

That said, yes, obviously the UI update system for non-nano UIs is pretty shit. Another alternative besides keeping weakrefs to users as I mentioned above is keeping refs to browser datums.

EDIT: If you really insist on using these I'll swallow it, but at least consider alternatives.

code/modules/paperwork/faxmachine.dm
afterthought2: Sanitize by default html-encodes the input. This is often correct (e.g. for input from an actual `input` call), but if it's not it can lead to double-encoding and player-facing &number stuff. I can't tell if it's correct here or not. There is a `sanitize` argument for not encoding if that's necessary.

code/modules/paperwork/handlabeler.dm
afterthought2: Qdeling an item in hands is always valid, it triggers appropriate code in the `Destroy` chain. If there are bugs related to that report them on the tracker.

code/modules/paperwork/faxmachine.dm
PsyCommando: Those ref got a pretty well contained lifetime. When removing a part, it will call refresh_parts, which will clear the refs. It seems that qdeleting a part while inside a machine doesn't have any handling, which might be a bug, since then anything calculated from parts would be wrong. But that seems to be more of a wide-scale implementation oversight?

The main reason for those refs is that they're very much central to how the machine operates. And having to constantly fetch it from the components list when we could already determine if we had that thing or not on install/removal would get expensive after a whille. Especially when updating the UI via autoupdates and such. 

I'll clear them on faxmachine/Destroy too I guess.

code/modules/paperwork/helpers.dm
afterthought2: I mean an `/item` proc. If you are doing an `if(istype` chain, there's usually a better way to organize the code, and the simplest approach is a proc with overrides.

code/modules/paperwork/paper.dm
afterthought2: Check with Loaf, if he prefers this then leave it. I would personally never do this though.

code/modules/paperwork/paper.dm
afterthought2: I'm just confused as to why it's not read only for AI: presumably it shouldn't be modifying the paper?

code/modules/paperwork/paper.dm
afterthought2: There's a `canUnEquip` if you need to check it before making the bundle.

code/modules/paperwork/paper_bundle.dm
afterthought2: Run it by Loaf or Lohikar; image manipulation is not very expensive.

code/game/machinery/_machines_base/stock_parts/disk_reader.dm
afterthought2: Well, it's extremely similar to the card reader, with the only major difference being the item it accepts; surely we can find some way to share code?

code/modules/paperwork/clipboard.dm
PsyCommando: Yeah, its true not checking if user.machine == src is a mistake here..

But, yeah, about the weakref, that's what I did in the rework for interface users. But, the problem is, if I write a whole new custom system for this specific thing it'll just convolute things even more.

And, you can't format interactions to account for external factors like adding a page, dropping the item, etc.. You absolutely need the system I'm currently using here for that (althought I should really be using UpdateUsrDialog() instead of calling attack_self())

code/modules/persistence/noticeboards.dm
PsyCommando: What do you mean actioned?
Also, I might need to get rid of that code, since wall offsets are handled differently now.

code/game/machinery/_machines_base/stock_parts/disk_reader.dm
PsyCommando: Well, I guess I could make a parent class for something that handles an item being inserted? Could be use for future parts too.

code/modules/fabrication/designs/imprinter/designs_misc_circuits.dm
PsyCommando: Yeah, but now they've got an actual use.
The network update means you can actually easily send faxes between machines.

code/modules/paperwork/clipboard.dm
PsyCommando: Its not like I didn't try other alternatives first. The only problem I got with /datum/browser is that I'd need to hold refs per user inside the object.  And then handling the cleanup of closed windows would get really annoying, and I'd probably need to use onclose anyways. It seems to have been thought out for things that only one person may look at at a time.

Weakrefs are really the way to go imho. Its just I don't think piling up another custom handling over the current generic one is going to be worth it. At least using the old system, its way simpler to refactor the thing later on when someone makes a proper user handling system for browser window guis?

code/modules/paperwork/faxmachine.dm
PsyCommando: I'm going to take a look.

code/modules/paperwork/handlabeler.dm
PsyCommando: Might have been from the hand slots rework at the time. But fair enough, I'll just qdel.

code/modules/paperwork/helpers.dm
PsyCommando: Yeah, but, I didn't know if I wouldn't meet resistance for creating a Clone() proc at the obj or item level. Also, it would be kind of a big undertaking on its own? Unless I just stub it under /item and call it a day?

code/modules/paperwork/paper.dm
PsyCommando: Hmm, yeah, that's true. I always forgot that there used to be a use and a show_content proc.

code/modules/paperwork/paper.dm
PsyCommando: I found why! There's a very cursed proc in the admin paper object :
```
/obj/item/paper/admin/get_signature()
	return input(usr, "Enter the name you wish to sign the paper with (will prompt for multiple entries, in order of entry)", "Signature") as text|null
```
This gets called by parsepencode, which is called by set_content.. 
Seems like someone didn't want to fix their problem and just brushed it under the rug..

```
 code/game/machinery/nuclear_bomb.dm, line 411, column 55:
error: /obj/item/folder/envelope/nuke_instructions/proc/Initialize calls /obj/item/paper/proc/get_signature which has override child proc that sleeps /obj/item/paper/admin/proc/get_signature
- code/game/atoms_init.dm:34:2: SpacemanDMM_should_not_sleep set here
- 414:3: /obj/item/paper/proc/set_content() called here
- code/modules/paperwork/paper.dm:97:9: /obj/item/paper/proc/parsepencode() called here
- code/modules/paperwork/paper.dm:252:65: /obj/item/paper/proc/get_signature() called here
- code/modules/paperwork/adminpaper.dm:160:9: input() called here

code/modules/paperwork/adminpaper.dm, line 22, column 33:
error: /obj/item/paper/admin/proc/Initialize calls /obj/item/paper/proc/get_signature which has override child proc that sleeps /obj/item/paper/admin/proc/get_signature
- code/game/atoms_init.dm:34:2: SpacemanDMM_should_not_sleep set here
- 23:6: /obj/item/paper/proc/Initialize() called here
- code/modules/paperwork/paper.dm:67:2: /obj/item/paper/proc/set_content() called here
- code/modules/paperwork/paper.dm:97:9: /obj/item/paper/proc/parsepencode() called here
- code/modules/paperwork/paper.dm:252:65: /obj/item/paper/proc/get_signature() called here
- 160:9: input() called here

code/modules/paperwork/paper.dm, line 65, column 27:
error: /obj/item/paper/proc/Initialize calls /obj/item/paper/proc/get_signature which has override child proc that sleeps /obj/item/paper/admin/proc/get_signature
- code/game/atoms_init.dm:34:2: SpacemanDMM_should_not_sleep set here
- 67:2: /obj/item/paper/proc/set_content() called here
- 97:9: /obj/item/paper/proc/parsepencode() called here
- 252:65: /obj/item/paper/proc/get_signature() called here
- code/modules/paperwork/adminpaper.dm:160:9: input() called here
```

PsyCommando: ~~lol, I stg that better not be my code XD~~

EDIT: It wasn't my code. But my changed might not have helped. Seems like people decided to make get_signarture async so that it would wait on the admin before fully creating the object or some pretty "galaxy brain" stuff. If I can find where the admin paper is used, I can probably get rid of that get_signature implementation and just get it in the proc instead so it doesn't block on initialize...

code/modules/paperwork/handlabeler.dm
MistakeNot4892: No, this has been the behavior as far back as Bay.

PsyCommando: I meant the bug.

code/modules/paperwork/helpers.dm
MistakeNot4892: It's an awfully niche proc to have at global scope too.

code/modules/paperwork/helpers.dm
MistakeNot4892: Clone proc at item or obj level would be fine imo. Default behavior would just need to pass appropriate args to new().

code/modules/paperwork/paper.dm
MistakeNot4892: Caching images is unnecessary, I don't think I would have recommended this. May have waffled about caching icon procs and been unclear.

code/modules/admin/admin.dm
```diff
@@ -1458,93 +1472,90 @@ var/global/floorIsLava = 0
...
+	// for(var/obj/machinery/faxmachine/F in possible_targets)
+	// 	var/datum/extension/network_device/N = get_extension(F, /datum/extension/network_device)
+	// 	var/datum/computer_network/CN = N?.get_network()
+	// 	if(!N || !CN)
+	// 		continue
+	// 	var/area/A = get_area(F)
+	// 	html += "<option value=\"\ref[F]\">[CN.network_id].[N.network_tag] [A? "([A])" : ""]</option>"
+	// html += "</select></label></BR>"
+	// html += "<button type=\"submit\">Continue ...</button></BR>"
+	// html += "</form>"
```
afterthought2: What's with the comments?

code/modules/paperwork/faxmachine.dm
afterthought2: The bigger problem is handling the case of `scanner_item` being destroyed.

code/modules/paperwork/faxmachine.dm
afterthought2: Machinery, on part install, registers for the part's destroyed event. When the part is destroyed, it removes references to the part. You have to do this too if you want to keep refs to parts like this.

When I say "handle GC" the issue is not what happens when the machine is destroyed--that's easy, the machine (presumably) is deleted cleanly and the refs it carries become irrelevant (though yes, it's certainly better to wipe them). It's what happens when the thing which is referenced is destroyed.

PsyCommando: > Machinery, on part install, registers for the part's destroyed event. When the part is destroyed, it removes references to the part. You have to do this too if you want to keep refs to parts like this.
> 
> When I say "handle GC" the issue is not what happens when the machine is destroyed--that's easy, the machine (presumably) is deleted cleanly and the refs it carries become irrelevant (though yes, it's certainly better to wipe them). It's what happens when the thing which is referenced is destroyed.

If they get deleted, and their event is called, its gonna call RefreshParts which will clear the ref to the part that was just removed anyways?

code/modules/paperwork/helpers.dm
PsyCommando: I mean, I don't really care that much where it is. It just sounded too niche of a functionality to put on the base mob.

code/modules/paperwork/helpers.dm
PsyCommando: I think I'll do that then. But maybe it should just runtime/print a warning in the base implementation? Because, creating a new instance of an objects with the same parameters is pretty far from what you'd want Clone() to do?

code/modules/paperwork/paper.dm
PsyCommando: Btw, fixed that.

code/modules/admin/admin.dm
PsyCommando: Woops, forgot to delete that before commiting..

code/modules/paperwork/paper.dm
PsyCommando: kk

code/modules/fabrication/designs/imprinter/designs_misc_circuits.dm
MistakeNot4892: Might be worth breaking out or restricting the admin fax capability to an unbuildable subtype.

PsyCommando: Admin fax is done via the admin verbs though. As it was before.

code/modules/paperwork/helpers.dm
PsyCommando: I put the clone stuff in its own PR: #2787.
Mainly because I suspect its gonna grow and get changed a lot.

code/modules/fabrication/designs/imprinter/designs_misc_circuits.dm
MistakeNot4892: Players sending faxes to the admin.

PsyCommando: I added an access check to admin addresses instead. So now you can't send them a fax unless you got the required access on the id card you inserted in the machine.

code/game/machinery/_machines_base/stock_parts/disk_reader.dm
PsyCommando: I made a class called ``/obj/item/stock_parts/item_holder`` that both readers inherit from. It should cover all the shared code.

code/modules/paperwork/faxmachine.dm
afterthought2: I don't think they call RefreshParts, but I'm not sure.

code/modules/paperwork/helpers.dm
afterthought2: It's generally strictly preferable to write instance procs to global procs, as they can be overridden. For this proc, it seems logical to have it be on the instance, as different mob types can then cleanly override with pen-finding behavior specific to them (borg pen module, custom slots, pen organs, whatever).

code/modules/fabrication/designs/imprinter/designs_misc_circuits.dm
afterthought2: OK, this is fine for me. It shouldn't be difficult to either override on the map or later add easier override behavior.

afterthought2:
OK, looking over this again I suggest the following:

- Do a final pass over GC behavior. You can check the wiki guide to GC to see what I would generally recommend. My comments about GC in the diff, given the size, are mostly there to get you to review the GC behavior and make sure that you've throught it through fully, not to find every possible error.
- Loaf has a comment about offsets of noticeboards. Make sure that either that's addressed, it's addressed in another open PR (not sure if that's the case or not), or the change is delayed to some future PR.
- Make the pen proc a mob proc.

@MistakeNot4892, I may be missing other outstanding issues given that there are more than 100 comments in the diff. Speak now if you have further blocking issues.

@PsyCommando, once you indicate you are content with the GC handling and deal with the other two things I mention, I will approve.

code/modules/fabrication/designs/imprinter/designs_misc_circuits.dm
MistakeNot4892: Works for me too.

code/modules/paperwork/faxmachine.dm
PsyCommando:
```
/obj/machinery/proc/component_destroyed(var/obj/item/component)
	events_repository.unregister(/decl/observ/destroyed, component, src)
	LAZYREMOVE(component_parts, component)
	LAZYREMOVE(processing_parts, component)
	power_components -= component
```

Apparently, they don't. But then that's a bug, and pretty messy. Component efficiency won't get recalculated, and uninstall code isn't run either.  So, stat doesn't get updated,
Honestly, the only way a component can be straight up qdeled inside a machine is either if the machine is deleted, or if someone manually VV and delete it.
Even explosion_act skip over components which are inside machines. I don't think its a very well supported or planned use case.

code/modules/paperwork/paperbin.dm
```diff
@@ -21,83 +28,145 @@
...
+			return
+		add_paper(I)
+		to_chat(user, SPAN_NOTICE("You put [I] in [src]."))
+		return TRUE
 
-/obj/item/paper_bin/attackby(obj/item/i, mob/user)
-	if(istype(i, /obj/item/paper))
-		if(!user.unEquip(i, src))
+	else if(istype(I, /obj/item/paper_bundle))
+		if(amount >= max_amount)
```
comma, line 91: you only check if there's at least 1 empty slot here, but put all papers in separately, extra papers will fail to be added

PsyCommando: Yes? So it will only add what fits in?
That's intended, so refilling isn't a pain in the ass.

code/modules/paperwork/faxmachine.dm
PsyCommando: @afterthought2

code/modules/paperwork/helpers.dm
PsyCommando: @afterthought2 
That's true. 
I'll figure out something.

PsyCommando:
> OK, looking over this again I suggest the following:
> 
>     * Do a final pass over GC behavior. You can check the wiki guide to GC to see what I would generally recommend. My comments about GC in the diff, given the size, are mostly there to get you to review the GC behavior and make sure that you've throught it through fully, not to find every possible error.
> 
>     * Loaf has a comment about offsets of noticeboards. Make sure that either that's addressed, it's addressed in another open PR (not sure if that's the case or not), or the change is delayed to some future PR.
> 
>     * Make the pen proc a mob proc.
> 
> 
> @MistakeNot4892, I may be missing other outstanding issues given that there are more than 100 comments in the diff. Speak now if you have further blocking issues.
> 
> @PsyCommando, once you indicate you are content with the GC handling and deal with the other two things I mention, I will approve.

Not sure what the noticeboard thing is?

PsyCommando:
> OK, looking over this again I suggest the following:
> 
>     * Do a final pass over GC behavior. You can check the wiki guide to GC to see what I would generally recommend. My comments about GC in the diff, given the size, are mostly there to get you to review the GC behavior and make sure that you've throught it through fully, not to find every possible error.
> 
>     * Loaf has a comment about offsets of noticeboards. Make sure that either that's addressed, it's addressed in another open PR (not sure if that's the case or not), or the change is delayed to some future PR.
> 
>     * Make the pen proc a mob proc.
> 
> 
> @MistakeNot4892, I may be missing other outstanding issues given that there are more than 100 comments in the diff. Speak now if you have further blocking issues.
> 
> @PsyCommando, once you indicate you are content with the GC handling and deal with the other two things I mention, I will approve.

@afterthought2
Btw, idk if you've read that: https://github.com/NebulaSS13/Nebula/pull/2552#discussion_r1008728691
But seems like the event on components being deleted is going to cause problem in its current state, since it doesn't refresh the components of the parent machine.

afterthought2:
My main concern with component deletion is GC handling. Generally components will be deleted due to the machine being deleted or some backend purpose; if they are deleted for some other reason then whatever deletes them should handle uninstall logic as appropriate (usually by uninstalling first). But the machine, and the component, still has to delete cleanly regardless of the reason--hence the event.

So no, them not calling RefreshParts from the deletion listener is not a bug per se.

MistakeNot4892:
Noticeboard appears to be fine now, it's handled by `directional_offset` and updated appropriately in `set_dir()`.

MistakeNot4892:
I don't mind if this goes in. I'm not signing off on it, but it's going to dev/staging so bugs should be found fairly rapidly. Further reviewing really seems to be diminishing returns given how stupid huge it is.

afterthought2 (approved):
There's some stuff here I don't love (would prefer more code sharing between the "paper" entities) and I'm trusting you on the GC stuff, but we may as well go for it.

PsyCommando:
> There's some stuff here I don't love (would prefer more code sharing between the "paper" entities) and I'm trusting you on the GC stuff, but we may as well go for it.

I would too, but at the time people were making huge changes to the computer files, so I couldn't get around to sharing text contents between things. 
Ideally, I want a datum for sharing document contents, so its about a million time easier to scan and copy text documents.

PR #2553, opened 2022-06-25
Sorts the spawn material stack verb type list

PsyCommando:
## Description of changes
The material list shown when clicking the "spawn material stack" verb is sorted now. So you can actually find something in that overcrowded listbox.
A browser window might work better in the future though.

## Changelog
:cl:
admin: The list for spawning material stacks is now sorted by type name. So its a bit more useful.
/:cl:

### DIFF
--- a/code/modules/admin/verbs/debug.dm
+++ b/code/modules/admin/verbs/debug.dm
@@ -472,7 +472,8 @@
 	var/material = input("Select material to spawn") as null|anything in mat_types
 	if(!material)
 		return
-	SSmaterials.create_object(material.type, get_turf(mob), 50)
+	var/decl/material/M = SSmaterials.materials_by_name[material]
+	SSmaterials.create_object(M.type, get_turf(mob), 50)
 
 /client/proc/force_ghost_trap_trigger()
 	set category = "Debug"
--- a/code/modules/admin/verbs/debug.dm
+++ b/code/modules/admin/verbs/debug.dm
@@ -468,7 +468,8 @@
 	set name = "Spawn Material Stack"
 	if(!check_rights(R_DEBUG)) return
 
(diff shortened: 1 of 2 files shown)


code/modules/admin/verbs/debug.dm
```diff
@@ -468,10 +468,12 @@
 	set name = "Spawn Material Stack"
 	if(!check_rights(R_DEBUG)) return
 
-	var/decl/material/material = input("Select material to spawn") as null|anything in SSmaterials.materials
```
MistakeNot4892, line 471: This is a regression - `SSmaterials.materials` is an instance list that uses the `name` var.
I would suggest adding a sort to the end of `SSmaterials` init rather than sorting the entire list of several hundred materials every time someone uses the verb.

MistakeNot4892 (changes requested):
Open to using the `materials_by_name` list instead, but the sort should definitely be done on SSmaterials init.

code/modules/admin/verbs/debug.dm
PsyCommando: Not sure what you mean there about it being "an instance list that use the name var"? Its not associative, so there's no name?
And I guess I could sort it after material init. But do we know if anything creates material decl on the fly?

MistakeNot4892: You can't create material decls on the fly by definition.
The code you are replacing does not use the assoc list.

PsyCommando: True.

PsyCommando:
Going to check back on this soon. Just gotta finish up the other prs and bugfixes downstream.

MistakeNot4892 (approved):

PR #2556, opened 2022-06-25
do_after() is now centered on the target atom regardless of icon size.

MistakeNot4892:
Thanks to WatchCat on Discord for discussing possible solutions to this.

### DIFF
--- a/code/datums/progressbar.dm
+++ b/code/datums/progressbar.dm
@@ -9,6 +9,8 @@
 	var/mob/user
 	var/client/client
 	var/listindex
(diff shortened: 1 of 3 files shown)


afterthought2:
Is it really worth it creating an icon each time to do this?

MistakeNot4892:
The icon is never shown to anyone so as far as I am aware we're not running into the network overhead that makes icon generation an issue.

MistakeNot4892:
Added a bool that you need to set on the atom before it will use this behavior.

afterthought2 (approved):

PR #2557, opened 2022-06-25
Polaris beams port.

MistakeNot4892:
Port of Polaris's port of Paradise's port of tg's rewrite of `Beam()`, last touched by @Leshana a number of years ago as far as I can tell. I've trimmed out the subtyping/extra behavior, as the previous `Beam()` was just cosmetic and that's all I need out of it, and I've added some caching for the icon drawing. Otherwise I haven't tinkered with it much. Tested with the ninja net launcher and appears to work as expected.


code/datums/beam.dm
```diff
@@ -0,0 +1,142 @@
...
+		X.icon = II
+		X.transform = rot_matrix
+
+		//Calculate pixel offsets (If necessary)
+		var/Pixel_x
+		var/Pixel_y
+		if(DX == 0)
+			Pixel_x = 0
+		else
+			Pixel_x = round(sin(Angle)+32*sin(Angle)*(N+16)/32)
```
afterthought2: Isn't this just `sin(Angle) * (N + 17)`? May need to check the math here.

code/datums/beam.dm
```diff
@@ -0,0 +1,142 @@
...
+		if(DY == 0)
+			Pixel_y = 0
+		else
+			Pixel_y = round(cos(Angle)+32*cos(Angle)*(N+16)/32)
```
afterthought2: Same here.

code/datums/beam.dm
MistakeNot4892: I have no idea.

MistakeNot4892: Math checks out in practice, implemented.

code/datums/beam.dm
MistakeNot4892: As above.

afterthought2 (approved):
I'm more concerned that the math is wrong, but don't really want to think about it, it would have to do with how Translate analogues round floats. I do find it confusing/wrong that the pixel offsets are not used to compute the angle.

MistakeNot4892:
Frankly it works in testing and that's all I care about. The previous version was worse than any problems this version might demonstrate.

PR #2562, opened 2022-07-08
Attempting to fix issues with non-owned limb robotize().

MistakeNot4892:
- Prosthetic recipes and crafting now supply a species.
- Species and bodytype checking on robotize() is a bit more nuanced.
- Robotics fabricators will not display prosthetic models that are invalid with the selected species (previously would cause them to spawn as unbranded).


afterthought2 (approved):
Might want to add some instructions to the UI if it's confusing.

Issue #2518, opened 2022-06-13
Mouse interactions


Qumefox:
#### Description of issue

Mice can interact with far more items in game than they should be able to.

#### Difference between expected and actual behavior

Mice shouldn't be unscrewing lightbulbs and setting off fire alarms.

#### Steps to reproduce

Be a mouse. Try to do things.

#### Specific information for locating

Items i've (so far) found that mice can meaningfully manipulate and interact with:

Elevator call buttons
Fire alarms
Portable air pumps
Portable scrubbers
Light bulbs

#### Length of time in which bug has been known to occur

Since I started hosting scavstation, but probably longer. At least ~6 weeks.

#### Client version, Server revision & Game ID

Client Version: 514
Server Revision: 013e91177a69345c0a692db0b6f73915933075ce - dev - 2022-06-09
Game ID: cip-aakB
Current map: Tradeship Ivenmoth

#### Issue bingo

- [x] Issue could be reproduced at least once
- [x] Issue could be reproduced by different players
- [x] Issue could be reproduced in multiple rounds
- [x] Issue happened in a recent (less than 7 days ago) round
- [x] [Couldn't find an existing issue about this](https://github.com/NebulaSS13/Nebula/issues)

MistakeNot4892:
Appears to be related to Chinsky's natural weapon PR giving animals a way to call `attack_hand()`.

MistakeNot4892:
Closing for now, please reopen with specific interactions you think are inappropriate if they pop up again. Mice should be very restricted in what they can interact with following the dexterity changes.

Issue #2544, opened 2022-06-19
warp ladder


SolatK:
#### Description of issue
ladder with wet floor can trow you through z levels

#### Steps to reproduce
1) find a ladder
2) make a floor on the other side wet. Best to change "wet" of the turf manually to 100
3) climb the ladder
#### Client version, Server revision & Game ID

Server Revision: 2788d4f07cd92bf2ceb4e8389f29fc873d30758e - dev - 2022-06-19

#### Issue bingo

- [x] Issue could be reproduced at least once
- [x] Issue could be reproduced by different players
- [x] Issue could be reproduced in multiple rounds
- [x] Issue happened in a recent (less than 7 days ago) round
- [x] [Couldn't find an existing issue about this](https://github.com/NebulaSS13/Nebula/issues)

SolatK:
banana peel didnt cause warp effect

PR #2563, opened 2022-07-08
Added some restrictions to base mob dexterity.

MistakeNot4892:
Partially addresses #2518 - mice can no longer freely interact with all machines and items. Some interactions do not check dexterity (such as unloading emergency lockers) are are still exploitable though.


MistakeNot4892:
Slapped a fix for #2544 in there too.

comma (approved):

PR #2567, opened 2022-07-11
Remove all uses of UNLINT, define-gates unit tests

noelle-lavenza:
## Description of changes
Removes every usage of UNLINT.
- Adds SpacemanDMM variable protection macros with exceptions for unit tests, since it accounts for a majority of the uses of those macros to begin with. They're postfixed with `_UT`.
- Separates the alert portion of `find_references` into `user_find_references` to avoid false positives for sleeping/blocking calls.
- Removes UNLINTs that were made redundant after the update to SpacemanDMM 1.3.
To facilitate the above changes:
- Moves all unit tests to be gated by the UNIT_TEST define.
- Changes a few map-related unit test files to gate UT-dependent code behind the UNIT_TEST define.

We should use UNLINT as little as possible.

## Authorship
Me.


code/controllers/subsystems/garbage.dm
```diff
@@ -393,9 +393,14 @@ SUBSYSTEM_DEF(garbage)
 	set name = "Find References"
 	set src in world
 
-	find_references(FALSE)
+	user_find_references()
 
-/datum/proc/find_references(skip_alert)
+/datum/proc/user_find_references()
+	if(alert("Running this will lock everything up for about 5 minutes.  Would you like to begin the search?", "Find References", "Yes", "No")) == "No")
```
MistakeNot4892: General view on alerts is to lead with "No" so you can't accidentally hit yes by having the alert pop up and take focus.

noelle-lavenza: Just moving it from the other proc, but I guess I can change it anyway.

code/modules/multiz/map_data.dm
```diff
@@ -6,8 +6,8 @@
 	var/height = 1     ///< The number of Z-Levels in the map.
 	var/turf/edge_type ///< What the map edge should be formed with. (null = world.turf)
 
-	VAR_PROTECTED/UT_turf_exceptions_by_door_type // An associate list of door types/list of allowed turfs
-
+	VAR_PROTECTED_UT/UT_turf_exceptions_by_door_type // An associate list of door types/list of allowed turfs
+z
```
MistakeNot4892: `z`

MistakeNot4892 (approved):

PR #2574, opened 2022-07-19
Renames all skills to match their in-game representation

PipeWizard:
## Description of changes

Renames all skills to match their in-game representation to prevent confusion while working in code.  That's it, nothing more.
I was quite thorough in my search to make sure I didn't miss anything, but it could be possible.

Decreases the confusion on devs when adjusting skills and testing in game when the skill levels don't seem to match up.

## Authorship

Me.

## Changelog
:cl:
refactor: Renames skill names in code to match in-game references.
/:cl:


code/__defines/skills.dm
```diff
@@ -1,9 +1,9 @@
-#define SKILL_NONE     1
+#define SKILL_UNSKILLED     1
 #define SKILL_BASIC    2
-#define SKILL_ADEPT    3
-#define SKILL_EXPERT   4
-#define SKILL_PROF     5
-#define HAS_PERK       SKILL_NONE + 1
+#define SKILL_TRAINED    3
+#define SKILL_EXPERIENCED   4
```
MistakeNot4892, line 4: pls fix spacing

code/__defines/skills.dm
```diff
@@ -18,7 +18,7 @@
 #define SKILL_EVA           /decl/hierarchy/skill/general/eva
 #define SKILL_MECH          /decl/hierarchy/skill/general/eva/mech
 #define SKILL_PILOT         /decl/hierarchy/skill/general/pilot
-#define SKILL_HAULING       /decl/hierarchy/skill/general/hauling
+#define SKILL_ATHLETICS       /decl/hierarchy/skill/general/athletics
```
MistakeNot4892: pls fix spacing

MistakeNot4892:
Other than the borked spacing, looks fine to me. Waiting to see afterthought's thoughts as this is a ton of largely non-functional diff noise.

code/__defines/skills.dm
PipeWizard: Spacing issues fixed

code/__defines/skills.dm
PipeWizard: Spacing issues also fixed.  Not sure why it didn't mark this as also being outdated?

MistakeNot4892:
As discussed on Discord, the macros are intended to be generic in case downstream or some future contrib wants to rename the ingame representation without having to find and replace 500 uses. I would prefer the macros to be changed to SKILL_NONE to SKILL_MAX rather than using subjective terms if you're going to be making this set of changes anyway. I do not have strong feelings either way though and neither does afterthought so personally I am just going to bow out of approving or merging this one.

afterthought2:
I personally don't find this worth it, especially given how our tooling now lets you see what the macro resolves to with hover or right click. If you do find them a serious problem for workflow, how about adding comments after the defines with their display names instead?

PR #2575, opened 2022-07-20
Attempt at item damage [MDB IGNORE]

PsyCommando:
## Description of changes
Basically, I got sick of reimplementing damage handling for any individual item types. I moved the take_damage proc down to the /obj/item level, so things are a bit more flexible and more ordered. 

Makes writing items that take damage simpler, and with less quirks due to implementation choices by the parents.

## Changelog
:cl:
code: Standardized items damage handling.
/:cl:


PsyCommando:
From what testing I could do it seems okay. Might need to write some unit tests for testing damage...

code/game/machinery/_machines_base/stock_parts/_stock_parts.dm
```diff
@@ -86,18 +89,19 @@
 	SetName("broken [name]")
 
 /obj/item/stock_parts/proc/is_functional()
-	return health > 0
+	return (health == -1) || (health > 0)
```
MistakeNot4892: Should probably have a macro for `-1` to make it a bit more descriptive.

PsyCommando: That's probably true. But, I'm afraid having an HEALTH_NO_DAMAGE or something among the line as a constant would cause confusion with the other existing health systems that may or may not handle -1 as being invulnerable. Also, -1 is usually used as an invalid/nonsensical value that's not null. 

I'd say that might be best reserved for whoever reworks the health system as a whole. I just needed items to handle health properly to make my life easier, so I'd rather not get into the massive headache that reworking all those would be. At least not for now XD

MistakeNot4892: A given constant meaning something else in another system isn't a good reason not to make it clear in this system. I am not in favour of leaving a bunch of bare-assed `-1`s in this code.

PsyCommando: Alright then, made a constant.

PsyCommando:
Note to myself: 
Removing health in explosion_act() by a proportional amount to an object's max health is not working for very low health object.

Vizoplez:
> Note to myself: Removing health in explosion_act() by a proportional amount to an object's max health is not working for very low health object.

This may be due to how objects are actually being handled by explosives.

IIRC it may be down to how 'Deleting' an object may actually work, last I checked object health would sometimes be skipped if an explosion is supposed to outright remove a specific object, but that might be fixed.

Something to go off of, however. Also-

`/obj/item/explosion_act(severity)
	if(QDELETED(src))
		return`

may be part of the offending code.

PsyCommando:
> > Note to myself: Removing health in explosion_act() by a proportional amount to an object's max health is not working for very low health object.
> 
> This may be due to how objects are actually being handled by explosives.
> 
> IIRC it may be down to how 'Deleting' an object may actually work, last I checked object health would sometimes be skipped if an explosion is supposed to outright remove a specific object, but that might be fixed.
> 
> Something to go off of, however. Also-
> 
> `/obj/item/explosion_act(severity) if(QDELETED(src)) return`
> 
> may be part of the offending code.

That's not really related. What I meant was that since I used proportional damage, items with like 2HP would have the damage from the explosion either rounded off or take no damage or a very tiny amount. It was just a really dumb idea I had.

PsyCommando:
The CL recipe issue is being fixed in #2673.

PsyCommando:
Should be good now?

code/_global_vars/lists/objects.dm
```diff
@@ -25,3 +25,6 @@ var/global/list/alphabet_capital = list("A","B","C","D","E","F","G","H","I","J",
 var/global/list/numbers = list("0","1","2","3","4","5","6","7","8","9")
 
 var/global/list/meteor_list = list()
+
+/// List of types that shouldn't be tested in the item damage test.
+var/global/list/item_damage_test_examptions = list()
```
MistakeNot4892: `exemptions`

code/game/machinery/_machines_base/stock_parts/_stock_parts.dm
```diff
@@ -63,18 +63,21 @@
 // RefreshParts has been called, likely meaning other componenets were added/removed.
 /obj/item/stock_parts/proc/on_refresh(var/obj/machinery/machine)
 
-/obj/item/stock_parts/proc/take_damage(amount, damtype)
-	if(!is_functional() || (damtype in ignore_damage_types))
+/obj/item/stock_parts/take_damage(damage, damage_type, damage_flags, inflicter, armor_pen)
+	if(!is_functional() || (damage_type in ignore_damage_types) || (istype(loc, /obj/machinery) && (part_flags & PART_FLAG_NODAMAGE)) )
```
MistakeNot4892: Should this set of conditions be in `can_take_damage()` instead?

code/game/objects/item.dm
```diff
@@ -118,6 +117,8 @@
 	UNSETEMPTY(matter)
 
 /obj/item/Initialize(var/ml, var/material_key)
+	if(!health)
```
MistakeNot4892: `isnull(health)` possibly makes more sense here.

code/game/objects/objs.dm
```diff
@@ -252,6 +252,22 @@
 ////////////////////////////////////////////////////////////////
 // Interactions
 ////////////////////////////////////////////////////////////////
+/**Returns a text string to describe the current damage level of the item, or null if non-applicable. */
+/obj/proc/show_examined_damage(var/health_ratio)
```
MistakeNot4892: Probably should rename this to `get_examined_damage_string` or something like that, since it isn't doing any showing currently.

code/game/objects/items/weapons/material/misc.dm
```diff
@@ -68,7 +68,7 @@
 	set_extension(src, /datum/extension/tool, list(TOOL_HATCHET = TOOL_QUALITY_DEFAULT))
 
 /obj/item/hatchet/unbreakable
-	unbreakable = TRUE
```
MistakeNot4892, line 71: Hrm this is a bit of an interesting case. The point of `unbreakable` was so that material weapons didn't degrade and destroy the item on use as a weapon, rather than the intent being for them to opt entirely out of damage simulation.

code/game/objects/items/weapons/lighter.dm
```diff
@@ -17,6 +17,7 @@
 
 /obj/item/flame/lighter/Initialize()
 	. = ..()
+	item_flags |= ITEM_FLAG_HOLLOW
```
MistakeNot4892: Why apply this in Initialize()?

code/game/objects/items/weapons/handcuffs.dm
```diff
@@ -13,16 +12,28 @@
...
+/obj/item/handcuffs/physically_destroyed(skip_qdel)
+	if(istype(loc, /obj/item/clothing/shoes))
+		loc.visible_message(SPAN_WARNING("\The [src] attached to \the [loc] snap and fall away!"), range = 1)
+	. = ..()
+
 /obj/item/handcuffs/examine(mob/user)
 	. = ..()
-	if (health)
-		var display = health / initial(health) * 100
+	if (health > 0 && max_health > 0)
```
MistakeNot4892, line 35: Won't they be qdeleted on health <= 0 anyway?

MistakeNot4892: Nvm, I see the weirdness now.

code/game/objects/items/weapons/cosmetics.dm
```diff
@@ -14,6 +14,7 @@
 
 /obj/item/lipstick/Initialize()
 	. = ..()
+	item_flags |= ITEM_FLAG_HOLLOW
```
MistakeNot4892: Same question as lighters.

code/game/objects/item_materials.dm
```diff
@@ -73,8 +96,12 @@
 	if(new_material)
 		material = GET_DECL(new_material)
 	if(istype(material))
-		health = round(material_health_multiplier * material.integrity)
-		max_health = health
+		//Only set the health if health is null. Some things define their own health value.
+		if(isnull(max_health))
```
MistakeNot4892, line 106: `max_health` should be multiplied by the material health modifiers imo. Doesn't make sense that a cardboard machete will be as durable as a titanium one.

code/game/objects/item_materials.dm
```diff
@@ -20,22 +20,45 @@
...
+	if(lastdamtype == BRUTE)
+		if(material?.is_brittle())
+			shatter(consumed)
+			return
+	else if(lastdamtype == BURN)
+		melt()
+		return
+	physically_destroyed()
+
+/obj/item/melt()
```
MistakeNot4892, line 48: This should probably be on `/obj` since that is where matter is defined.

code/game/objects/item_materials.dm
```diff
@@ -20,22 +20,45 @@
...
+		if(!M)
+			log_warning("[src] ([type]) has a bad material path in its matter var.")
+			continue
+		var/turf/T = get_turf(src)
+		//TODO: Would be great to just call a proc to do that, like "Material.place_burn_product(loc, amount_matter)" so no need to care if its a gas or something else
+		var/datum/gas_mixture/environment = T?.return_air()
+		if(M.burn_product)
+			environment.adjust_gas(M.burn_product, M.fuel_value * (matter[mat] / SHEET_MATERIAL_AMOUNT))
+
+	new /obj/effect/decal/cleanable/ash(src)
```
MistakeNot4892: There's a 'melted item' dummy you could use for this.

MistakeNot4892: `/obj/effect/decal/cleanable/molten_item`

code/game/objects/item_materials.dm
```diff
@@ -20,22 +20,45 @@
...
 			health--
 		check_health()
 
-/obj/item/proc/check_health(var/consumed)
-	if(health<=0)
-		shatter(consumed)
+/obj/item/proc/check_health(var/lastdamage = null, var/lastdamtype = null, var/lastdamflags = 0, var/consumed = FALSE)
+	if(health > 0 || !can_take_damage())
+		return //If invincible, or if we're not dead yet, skip
+	if(lastdamtype == BRUTE)
```
MistakeNot4892, line 39: I feel like
````
switch(lastdamtype)
	if(BRUTE)
		if(material?.is_brittle())
			shatter(consumed)
	if(BURN)
		melt()
	else
		physically_destroyed()
````
would generally read better and be equivalent logically.

File: code/game/objects/items/weapons/extinguisher.dm [hunk 14]
```diff
@@ -44,6 +44,7 @@
 
 /obj/item/extinguisher/Initialize()
 	. = ..()
+	item_flags |= ITEM_FLAG_HOLLOW
```
MistakeNot4892: Same question as lighters.

code/modules/clothing/shoes/_shoes.dm
```diff
@@ -57,101 +57,106 @@
 	if(!.)
 		. = ..()
 
+/obj/item/clothing/shoes/proc/update_cuffed()
```
MistakeNot4892: This proc is nonsensical.
- You could be initializing the slowdown list outside of the condition.
- You are trying to check the elastic var of cuffs after falsy-checking the cuffs, this will runtime without fail.
- This proc has no safeguards against being called repeatedly, and the naming convention for `update_foo` tends to imply it's safe to call at arbitrary times. If this is called at any time after the cuffs are added or removed you'll end up adding or removing magic slowdown.

code/modules/clothing/shoes/_shoes.dm
```diff
@@ -178,4 +183,34 @@
 		S.alpha = 127 * shine / 100
 		S.blend_mode = BLEND_ADD
 		overlay.overlays += S
-	. = ..()
\ No newline at end of file
+	. = ..()
+
+// Interactions
```
MistakeNot4892: If you're going to rework how cuffing is handled I would suggest doing it in a separate PR, as it has a lot of noodly bits to review that will slow the item damage stuff down (the slowdown comment for example)

code/modules/mechs/components/legs.dm
```diff
@@ -51,7 +51,7 @@
 /obj/item/mech_component/propulsion/return_diagnostics(mob/user)
 	..()
 	if(motivator)
-		to_chat(user, SPAN_NOTICE(" Actuator Integrity: <b>[round((((motivator.max_dam - motivator.total_dam) / motivator.max_dam)) * 100)]%</b>"))
+		to_chat(user, SPAN_NOTICE(" Actuator Integrity: <b>[round((((motivator.max_health - motivator.health) / motivator.max_health)) * 100)]%</b>"))
```
MistakeNot4892: I feel like we could do with a percent_health() helper at some point.

mods/species/vox/_vox.dm
```diff
@@ -9,6 +9,10 @@
 	credits_crew_names = list("THE VOX")
 	credits_topics = list("VOX RITUAL DUELS", "NECK MARKINGS", "ANCIENT SUPERCOMPUTERS")
 
+/decl/modpack/vox/post_initialize()
+	. = ..()
+	global.item_damage_test_examptions |= /obj/item/gun/energy/darkmatter
```
MistakeNot4892: What is special about this gun?

code/game/objects/structures/__structure.dm
```diff
@@ -37,25 +37,18 @@
...
-	else
-		to_chat(user, SPAN_DANGER("It looks heavily damaged."))
+	. = ..()
 
 /obj/structure/examine(mob/user, var/distance)
 	. = ..()
 	if(distance <= 3)
 
-		show_examined_damage(user, (health/maxhealth))
+		var/damage_desc = show_examined_damage(health / maxhealth)
```
MistakeNot4892: Could move this down to /obj

code/game/machinery/_machines_base/stock_parts/_stock_parts.dm
PsyCommando: It probably could be yeah!

code/game/objects/item.dm
PsyCommando: True.

code/game/objects/items/weapons/material/misc.dm
PsyCommando: That's not really accurate though? All the unbreakable things are used in borg inventories. They shouldn't ever need to take damages since they're not supposed to exist outside the borg.

code/game/objects/items/weapons/lighter.dm
PsyCommando: Because I'm on the fence between doing this above, or just copying over the previous flag values and slapping them in the class definition. The way above means that any previously set bits in the parent would carry over unmolested.

code/game/objects/item_materials.dm
PsyCommando: The thing is, the health is normally set by the material by default anyways? 
If you specify a max_health, you override the material specific health. This is for things that don't or can't have a material, but need their health set, or if they have to override the material set health.

code/game/objects/item_materials.dm
PsyCommando: Can't because abstract objects.

PsyCommando: Also, this is a step towards the unified system. Its out of scope for now. Eventually I'll get things setup at the object level, or someone else will.

code/game/objects/item_materials.dm
PsyCommando: I guess.

code/game/objects/item_materials.dm
PsyCommando: Alright.

code/game/objects/item_materials.dm
PsyCommando: Actually, can't do that, because if the material isn't brittle, it needs to call physically_destroyed.

code/game/objects/items/weapons/lighter.dm
PsyCommando: Also, I forgot to remove those in Initialize after I opted for the class definition. ^^;

code/game/objects/objs.dm
PsyCommando: alright

code/game/objects/structures/__structure.dm
PsyCommando: Like I said earlier, that's a whole other can of worms. Its why I'm doing items separately first.

mods/species/vox/_vox.dm
PsyCommando: Not entirely sure. But it would runtime during unit testing when created on its own. Couldn't figure why from glancing over its really simple definition. Might have to do with the extension?

code/modules/clothing/shoes/_shoes.dm
PsyCommando: Yeah, looking at this, I realize I had a brainfart.

code/game/objects/structures/__structure.dm
PsyCommando: Also, to add to this, /obj doesn't define the health vars. And there's some name collision and bunch of other crap to fix. Its just simpler to keep those implemented separately on structures and items for now.

code/modules/clothing/shoes/_shoes.dm
PsyCommando: I can't completely avoid messing with the cuffs a bit in this PR though. Because, handcuffs use the health var to tell if someone has broken out of the cuffs or not. But, the sec handcuffs can't be broken out of, and yet should be able to take damage.. So this is kind of cursed.

code/_global_vars/lists/objects.dm
PsyCommando: fixed

code/unit_tests/obj_damage_tests.dm
```diff
@@ -0,0 +1,185 @@
...
+	global.item_damage_test_exemptions |= /obj/item/paper/monitorkey               //Crash on null turf
+	global.item_damage_test_exemptions |= /obj/item/chems/food/grown               //Base type, delete itself
+	global.item_damage_test_exemptions |= /obj/item/chems/food/grown/mushroom      //Base type, delete itself
+	global.item_damage_test_exemptions |= /obj/item/chems/food/fruit_slice         //Base type, delete itself
+	global.item_damage_test_exemptions |= /obj/item/chems/pill/pouch_pill          //Base type, causes many runtimes
+	global.item_damage_test_exemptions |= /obj/item/gun/energy/gun/secure/mounted  //Can't exist outside something
+	global.item_damage_test_exemptions |= /obj/item/energy_blade/ninja             //Delete itself on init
+	global.item_damage_test_exemptions |= /obj/item/grenade/flashbang/instant      //Delete itself
+	global.item_damage_test_exemptions |= /obj/item/crafting_holder                //Can't exist outside of something
+	global.item_damage_test_exemptions |= /obj/item/lock_construct                 //Can't exist outside of something
```
MistakeNot4892: This doesn't seem accurate, this type is crafted directly by a recipe.

code/unit_tests/obj_damage_tests.dm
```diff
@@ -0,0 +1,185 @@
...
+	global.item_damage_test_exemptions |= /obj/item/frame_holder                   //Abstract
+	global.item_damage_test_exemptions |= /obj/item/drill_head                     //Can't exist outside of something
```
MistakeNot4892: Same here, this is craftable. I think it just doesn't have a default material set.

code/unit_tests/obj_damage_tests.dm
PsyCommando: Which one do you mean?

code/unit_tests/obj_damage_tests.dm
PsyCommando: Which one? The drill_head?

MistakeNot4892:
I'm pulling the item instantiation logic out into #2725 as the exception list really doesn't sit right with me. I need to go over it again but currently it has the core of your unit test and is passing CI, so seems to be working.

PsyCommando:
> I'm pulling the item instantiation logic out into #2725 as the exception list really doesn't sit right with me. I need to go over it again but currently it has the core of your unit test and is passing CI, so seems to be working.

What do you mean? The unit test one?

PsyCommando:
Rebased on @MistakeNot4892 's branch.

PsyCommando:
reeeeeeeeee why did you rename the file ;n;

PsyCommando:
Found a problem with the implementation for organs. It doesn't really disrupt anything as far as I can tell, but it makes it so organs have 2 different health.
The problem with organs is that they have a damage amount which is their living health which determines if the organ is dead or alive.
Then there's the structural health that I've been adding in this PR, which determines if an object should be physically destroyed or not.

I'm not entirely sure what would be the best way to address that. I'm thinking its probably a good idea to keep them separate. But the risk for collision and for confusing one for the other is kinda high. Any suggestions @MistakeNot4892 @Lohikar ?

code/game/objects/items/weapons/material/misc.dm
MistakeNot4892: It is accurate. Shards, armblades and folding knives are `unbreakable = TRUE` and aren't borg modules.

PsyCommando: Fair point. I'm guessing its to prevent shattering on use? I could probably set some flag for preventing damage from wear instead of making them completely invulnerable?

MistakeNot4892: Keeping the existing var would also work.

code/game/objects/item_materials.dm
MistakeNot4892: There's nothing in this proc that will break for abstract items. Just check `!INSTANCE_IS_ABSTRACT(src) && simulated`.

code/game/objects/item_materials.dm
MistakeNot4892: Not sure what I meant by my comment as material_health_multiplier is still being used. May have been intended for a different line.

mods/species/ascent/items/id_control.dm
```diff
@@ -48,10 +48,9 @@
 	owner?.add_language(/decl/language/mantid/worldnet)
 
 /obj/item/organ/internal/controller/do_uninstall(in_place, detach, ignore_children)
-	if(owner)
-		var/datum/extension/access_provider/owner_access = get_extension(owner, /datum/extension/access_provider)
-		owner_access?.unregister_id(src)
 	var/mob/living/carbon/H = owner
+	var/datum/extension/access_provider/owner_access = owner? get_extension(owner, /datum/extension/access_provider) : null
```
MistakeNot4892: Please change this back.

code/modules/power/fuel_assembly/fuel_assembly.dm
```diff
@@ -4,6 +4,7 @@
 	icon_state = "fuel_assembly"
 	layer = 4
 	abstract_type = /obj/item/fuel_assembly
+	material = /decl/material/solid/metal/uranium
```
MistakeNot4892: This is going to make all custom fuel rods be made of uranium.

MistakeNot4892: It's also completely unrelated to the stated purpose of this PR. Same with the archaeological find change below.

code/modules/mechs/equipment/utility.dm
```diff
@@ -350,9 +350,6 @@
 	to_chat(user, "It [get_visible_durability()].")
 
 
-/obj/item/drill_head/steel
```
MistakeNot4892, line 353: Why are you doing these random material changes in this PR?

code/modules/materials/definitions/solids/materials_solid_organic.dm
```diff
@@ -94,7 +94,7 @@
 	destruction_desc        = "tears"
 	icon_base               = 'icons/turf/walls/solid.dmi'
 	icon_reinf              = 'icons/turf/walls/reinforced.dmi'
-	integrity               = 3
+	integrity               = 10 //Probably don't wanna go below 10, because of the health multiplier on things, that would result in a value smaller than 1.
```
MistakeNot4892, line 97: If this is a concern it should be unit tested.

code/modules/integrated_electronics/core/integrated_circuit.dm
```diff
@@ -4,7 +4,11 @@
 	icon = 'icons/obj/assemblies/electronic_components.dmi'
 	icon_state = "template"
 	w_class = ITEM_SIZE_TINY
-	matter = list()				// To be filled later
+	material = /decl/material/solid/silicon
```
MistakeNot4892: More random material changes.

code/modules/clothing/shoes/_shoes.dm
MistakeNot4892: That's fine. My issue is that you're rewriting a shitload of cuff interaction in a PR about item damage. Alt interactions and verb changes don't rely on item damage.

mods/species/ascent/items/id_control.dm
PsyCommando: Will do. Not sure where that even comes from..

code/modules/integrated_electronics/core/integrated_circuit.dm
PsyCommando: No, that one isn't random. Health and other properties are calculated from material. And integrated circuit didn't have a proper material set anywhere.

code/modules/materials/definitions/solids/materials_solid_organic.dm
PsyCommando: Yeah, it really should have been when the health multiplier was added. Should I add that to material tests or something else ?

code/modules/mechs/equipment/utility.dm
PsyCommando: Because items are tested whether they're properly setting their health. Things without materials do not set their health, unless you specify it manually? Its not random.

code/modules/power/fuel_assembly/fuel_assembly.dm
PsyCommando: Don't the custom ones set their material?

code/game/objects/items/weapons/material/misc.dm
PsyCommando: That would mean people could reasonably assume that unbreakable is synonym with not taking damage. And then use it by mistake instead of the actual way of making things not take damage?

code/modules/power/fuel_assembly/fuel_assembly.dm
MistakeNot4892: They supply a composition list to Initialize, this doesn't override material.

code/modules/materials/definitions/solids/materials_solid_organic.dm
MistakeNot4892: Adding to the material or item tests would make sense I think.

code/game/objects/items/weapons/material/misc.dm
MistakeNot4892: So rename the var. I just mean keeping the bool on the item doing the current job it's doing.

code/unit_tests/obj_damage_tests.dm
MistakeNot4892: This was all resolved by the big abstract type PR stuff.

code/modules/mechs/equipment/utility.dm
MistakeNot4892: It would be better to change materials here, for circuits, etc. in a separate PR so nothing sneaks in without review (such as the fuel rod stuff above). They're not strictly related to the item damage changes.

code/modules/materials/definitions/solids/materials_solid_organic.dm
PsyCommando: Added a test to items since, its mainly a problem in obj/item code.

code/modules/games/cards.dm
```diff
@@ -416,6 +416,7 @@ var/global/list/card_decks = list()
 /*** A special thing that steals a card from a deck, probably lost in maint somewhere. ***/
 /obj/item/hand/missing_card
 	name = "missing playing card"
+	abstract_type = /obj/item/hand/missing_card //Can't spawn this for tests because it kills itself
```
MistakeNot4892: This is an issue with the card object that isn't solved by marking it as an abstract type (it isn't an abstract type). Better fix would be moving the qdel logic out of update_icon of all places and making it return a hint to Initialize if `!length(cards)`.

code/modules/mechs/equipment/utility.dm
PsyCommando: done for the drill

code/modules/power/fuel_assembly/fuel_assembly.dm
PsyCommando: done for the fuel assembly

code/modules/integrated_electronics/core/integrated_circuit.dm
PsyCommando: Reverted the material changes in circuit stuff

code/modules/games/cards.dm
PsyCommando: You just told me not to add mildly unrelated changes though? Wouldn't fixing fundamental issues with the card game be part of that? Because looking at the code and what depends on that behaviour, it seems like its not gonna be a simple change.

Also, there's no test exclusion list anymore, so that's the only way I can exclude buggy things from testing. Plus, I'm not actually sure this item is even meant to really exist in the world or if its just some pseudo item for pulling off some hack for the card game visuals. So I thought classifying it as abstract made sense.

PsyCommando: Like, its not just in init that update_icon is called. Its called in the base class, and its called in several other procs of the card stuff. Its probably going to need some code diving and debugging for properly changing that and not breaking anything.

code/game/objects/item_materials.dm
PsyCommando: That's going to go in the obj damage PR.

code/modules/clothing/shoes/_shoes.dm
PsyCommando: Alright. I removed the changes in the shoes file.

code/game/objects/items/weapons/material/misc.dm
PsyCommando: Had to do a bit more than that since the var turned out to be handling both wear and shattering on destruction/usage.
But I restored the functionality without making those things invulnerable to damage.

PsyCommando:
> Found a problem with the implementation for organs. It doesn't really disrupt anything as far as I can tell, but it makes it so organs have 2 different health. The problem with organs is that they have a damage amount which is their living health which determines if the organ is dead or alive. Then there's the structural health that I've been adding in this PR, which determines if an object should be physically destroyed or not.
> 
> I'm not entirely sure what would be the best way to address that. I'm thinking its probably a good idea to keep them separate. But the risk for collision and for confusing one for the other is kinda high. Any suggestions @MistakeNot4892 @Lohikar ?

Update on this. After looking at it a whole bunch. Merging organ damage together with the health system is going to need its own PR. There's over 300 reference to organ damage vars that would need to be wrapped and have their behaviour altered to work with the system and I think you'll agree this PR is big enough. 
One of the big issue is that a lot of things are hardcoded to check if damage == max_damage to tell if the organ is dead, and etc. And a bunch of things are working with that assumption in parallel. 

In its current state, organs will still work just like they used to. They're just going to have one health, and one damage value that are independent of eachothers, which should be fine for the time being. Just a bit confusing to coders.

code/modules/games/cards.dm
MistakeNot4892: >You just told me not to add mildly unrelated changes though? Wouldn't fixing fundamental issues with the card game be part of that? Because looking at the code and what depends on that behaviour, it seems like its not gonna be a simple change.

So you would do it in a separate PR as a dependency for this PR. I'm not going to sign off on marking a non-abstract object as abstract because it causes an error in a unit test you are adding dude, that's nonsensical. Mark it as an unspawnable type if you need it to be skipped by the unit test and don't want to fix it, don't use `abstract_type`.

MistakeNot4892:
The situation with organ health seems fine to me. Damage is more abstract than physical state anyway since toxins etc. cause it.

PsyCommando:
Alright. So I'll keep it at that for now. And we'll see what we want to do with organ damage later I guess?

code/modules/games/cards.dm
PsyCommando: Sold! It doesn't seem much better to me, but I'll mark it as unspawnable. 

I just think the rabbit hole is deep enough as it is, without trying to fix every single problems along the way and burning out. I need to focus on one thing at a time. And I frankly don't want to have to figure out how the card games work, or if they even still work, in order to fix that.

PsyCommando:
The one time I push some code to try to figure out what's going on with the desperado stuff, it stops doing it...

code/game/objects/item_materials.dm
PsyCommando: #2729 Just gonna reference it here to make sure.

code/modules/mechs/components/legs.dm
PsyCommando: Done! Also, this seemed to be calculating the % of damage and not the integrity, so I fixed that.

MistakeNot4892 (approved):
Let's give it a go.

PR #2582, opened 2022-07-29
Fix masks making people bald, and hats inv_flags

PsyCommando:
## Description of changes
* Masks wouldn't properly update human overlays when dropped. Fixed that.
* Fixed the top hat not having a sprite showing on the mob.
* Fixed a lot of hats not hiding hairs properly.
* Fixing hats hiding people's facial hair for no reasons.

## Changelog
:cl:
bugfix: Fixed top hat not showing up on the mob sprite.
bugfix: Fixed balaclava and other masks turning people bald.
bugfix: Fixed a lot of hats not hiding hair under them.
bugix: Fixed various hats hiding people's facial hair arbitrarily.
/:cl:


code/modules/clothing/head/_head.dm
```diff
@@ -1,11 +1,13 @@
...
-	body_parts_covered = SLOT_HEAD
-	slot_flags = SLOT_HEAD
-	w_class = ITEM_SIZE_SMALL
-	blood_overlay_type = "helmetblood"
+	name                = "head"
+	icon_state          = ICON_STATE_WORLD
+	icon                = 'icons/clothing/head/softcap.dmi'
+	blood_overlay_type  = "helmetblood"
+	w_class             = ITEM_SIZE_SMALL
+	flags_inv           = BLOCKHEADHAIR
```
MistakeNot4892: This should not be the default behavior for all head items.

code/modules/clothing/head/hardhat.dm
```diff
@@ -25,6 +23,7 @@
 	material = /decl/material/solid/plastic
 	matter = list(/decl/material/solid/metal/steel = MATTER_AMOUNT_REINFORCEMENT)
 	origin_tech = "{'materials':1,'engineering':1,'combat':1}"
+	protects_against_weather = TRUE
```
MistakeNot4892: Why would a hardhat protect you against the weather?

MistakeNot4892:
Titling this a fix PR is misleading, this is a ton of largely unrelated changes mashed into one PR and fixing baldness isn't the primary one.

code/modules/mob/living/carbon/human/update_icons.dm
```diff
@@ -648,6 +648,14 @@ var/global/list/damage_icon_parts = list()
 /mob/living/carbon/human/update_inv_wear_mask(var/update_icons=1)
 	var/obj/item/mask = get_equipped_item(slot_wear_mask_str)
 	var/obj/item/head = get_equipped_item(slot_head_str)
+	update_hair(0)	//rebuild hair
+	update_inv_ears(0)
+
+	if(!(head && (head.item_flags & ITEM_FLAG_AIRTIGHT)))
+		set_internals(null)
+	if(!(mask && (mask.item_flags & ITEM_FLAG_AIRTIGHT)))
```
MistakeNot4892, line 654: This should be `||`, you can have an airtight mask or helmet and preserve your internals.

code/modules/clothing/head/hardhat.dm
MistakeNot4892: Per discussion, cowboy hats have this too. No idea why I put it on cowboy hats.

code/modules/mob/living/carbon/human/update_icons.dm
PsyCommando: It might need to be changed in other places too then. I think there's some head code that still does it like that.

code/modules/clothing/head/_head.dm
PsyCommando: The flag_inv you mean? Because the rest is just copy-pasted.

MistakeNot4892: Correct.

code/modules/mob/living/carbon/human/update_icons.dm
MistakeNot4892: Well, still needs to be changed here.

code/modules/clothing/head/_head.dm
PsyCommando: Can you explain why though? Because 90% of all headgear hides only the hair, and not the facial hair? It logically avoids having to type ``flags_inv = BLOCKHEADHAIR`` for most of them?

code/modules/mob/living/carbon/human/update_icons.dm
PsyCommando: Wait, the && is being a nullguard for the flag check afterwards. It doesn't make sense to slap || in there? 

Also, if I can be honest, I really don't think checking for internals should be done in the inventory updates to begin with.. It probably should be handled via dropped() or something?

MistakeNot4892: I misspoke - what I meant was:
````
	if(!(head && (head.item_flags & ITEM_FLAG_AIRTIGHT)) && !(mask && (mask.item_flags & ITEM_FLAG_AIRTIGHT)))
		set_internals(null)
````

Comment (PsyCommando) [reply code/modules/mob/living/carbon/human/update_icons.dm:654]: Oh alright!

File: code/modules/clothing/head/_head.dm (hunk 1 above)
Comment (MistakeNot4892) [reply code/modules/clothing/head/_head.dm]: The basic hat behavior should not be to hide hair, the majority of hats don't. Helmets and space helmets can have the hair hiding flags set easily.

Comment (MistakeNot4892) [reply code/modules/clothing/head/_head.dm]: Ok, misunderstood the flags.

**Comment by MistakeNot4892** [comment]
Would request something like:
```
#define BLOCK_HEAD_HAIR BITFLAG(4)    // Hides the user's head hair (rather than facial hair).
#define BLOCK_FACE_HAIR BITFLAG(5)    // Hides the user's facial hair.
#define BLOCK_ALL_HAIR (BLOCK_HEAD_HAIR|BLOCK_FACIAL_HAIR) // Blocks all hair.
```
while you're changing these, the existing defines can be quite confusing.
EDIT: feel free to just say 'no' to this, just a request.

PsyCommando:
> Would request something like:
> 
> ```
> #define BLOCK_HEAD_HAIR BITFLAG(4)    // Hides the user's head hair (rather than facial hair).
> #define BLOCK_FACE_HAIR BITFLAG(5)    // Hides the user's facial hair.
> #define BLOCK_ALL_HAIR (BLOCK_HEAD_HAIR|BLOCK_FACIAL_HAIR) // Blocks all hair.
> ```
> 
> while you're changing these, the existing defines can be quite confusing. EDIT: feel free to just say 'no' to this, just a request.

I could take a look, just to see if everything is going to break if I change it. Then if everything breaks it might need its own PR.  If not I'll push it to this PR.

PsyCommando:
Just an update, but I've been a bit busy with outreach stuff in the last 2 weeks, so replies/changes on my PRs might be slow for a bit.

code/modules/mob/living/carbon/human/update_icons.dm
PsyCommando: Looking at this again, I think the helmet check is actually wrong. That was part of what I moved over, but it makes no sense. So gonna just leave the mask one I added.

MistakeNot4892 (approved):

PR #2588, opened 2022-08-01
Allow /datum/mind/get_client() to actually return the client

PsyCommando:
## Description of changes
Allow mind datum to return their client with get_client(). It seems to have been a weird oversight, considering that get_client is defined at the datum level..

## Changelog
:cl:
code: /datum/mind/get_client() now returns the mind's client instead of null.
/:cl:

### DIFF
--- a/code/modules/client/client_helpers.dm
+++ b/code/modules/client/client_helpers.dm
@@ -1,6 +1,9 @@
 /datum/proc/get_client()
 	return null
 
+/datum/mind/get_client()
+	return current?.client
+
 /client/get_client()
 	return src
 


afterthought2 (approved):
I mean, I guess this is fine, but this is only used as a mob helper as far as I'm aware.

MistakeNot4892:
Yeah the reason this was never defined is that there's no instance where you want to get the client of a mind.

PsyCommando:
@MistakeNot4892 
> Yeah the reason this was never defined is that there's no instance where you want to get the client of a mind.

Well, it happened downstream. Honestly, I guess the better solution might have been to have the proc crash if it wasn't overriden.

MistakeNot4892:
````
Well, it happened downstream. 
````
I meant there is no extant use on Neb where it was needed.

Issue #2440, opened 2022-05-27
APC charge mode


SolatK:
#### Description of issue
Disable APC charge prevents it from charge it cell, but it used energy from wire, not from battery. 
![изображение](https://user-images.githubusercontent.com/29740451/170622718-60a8043a-123b-4ef6-ac91-c029cfc00e05.png)

#### Difference between expected and actual behavior
It should use cell stored energy when charge mode is off

#### Steps to reproduce
1) set apc charge mode to off
2) it dont use battery power

#### Client version, Server revision & Game ID

Client Version: 514
Server Revision: c2377f69454640262b020b26407fc27de8786968 - dev - 2022-05-27
Game ID: ch8-aBbe

#### Issue bingo

- [x] Issue could be reproduced at least once
- [x] Issue could be reproduced by different players
- [x] Issue could be reproduced in multiple rounds
- [x] Issue happened in a recent (less than 7 days ago) round
- [x] [Couldn't find an existing issue about this](https://github.com/NebulaSS13/Nebula/issues)

afterthought2:
I think that's just not a correct interpretation of how the interaction should work; charge off disables the battery from charging, but not the APC drawing power from the net. I guess it could be changed, though I don't relish the prospect of doing it.

SolatK:
> I think that's just not a correct interpretation of how the interaction should work; charge off disables the battery from charging, but not the APC drawing power from the net. I guess it could be changed, though I don't relish the prospect of doing it.

I'm not sure. On the one hand, everything is logically correct, on the other hand, on any other build, this switch just disconnects the APC from the network.

MistakeNot4892:
I would argue the purpose of the charge mode option is to stop APCs draining the powernet/SMES to charge themselves, which is still a valid use of it currently. Wouldn't call this a bug necessarily.

afterthought2:
I feel that since the APC is so extremely complicated as far as the possible interactions go, and there are ways of obtaining both of these effects with basic, completely universal _hardware_ interactions (cut terminal for the "legacy" one, remove battery for the "current" one; alternatively, remove the corresponding stock part), I'm not too concerned with this. If anything I'd consider just removing this software toggle, or renaming it, or explaining it better somehow.

MistakeNot4892:
Calling this fixed for now unless someone wants to articulate the issue and a better fix.

Issue #2118, opened 2021-09-26
Blobs spread ludicrously fast.


chaoko99:
#### Description of issue
When spawning, blobs will spread and heal at untenable speeds.

#### Difference between expected and actual behavior
Was expecting it to conceivably be combat-able. 

#### Steps to reproduce
Spawn blob or have it randomspawn
Be immediately completely overwhelmed.

#### Specific information for locating

Blob core

#### Client version, Server revision & Game ID

Client Version: 514
Server Revision: Revision Unknown
Game ID: cec-dL43
Current map: ISEO Endeavour

This was done downstream but I don't believe this is a downstream issue.
#### Issue bingo

- [ ] Issue could be reproduced at least once
- [ ] Issue could be reproduced by different players
- [ ] Issue could be reproduced in multiple rounds
- [x] Issue happened in a recent (less than 7 days ago) round
- [x] [Couldn't find an existing issue about this](https://github.com/NebulaSS13/Nebula/issues)

MistakeNot4892:
`ISEO Endeavour` This isn't a Neb map, have you reproduced the bug on Neb main?

chaoko99:
https://streamable.com/k7rilv here is an example of it happening on example

MistakeNot4892:
Noting after discussion with Noelle that this appears to be due to us fixing the delays in blob processing on SSobj and that previously blobs were slowing down object processing for the entire subsystem.

chaoko99:
lmao my favorite kind of timing issue, fair enough. It may be wise to add
some delays there—It's been some time since I played, but they could
consume the entire ship in like ~5 minutes.

MistakeNot4892:
Marking this as closed for now as the blob subsystem can be slowed down further if needed.

PR #2591, opened 2022-08-02
Staging fixes [IDB IGNORE] [MDB IGNORE]

### LINKED ISSUE #2545: lube is not slip
#### Description of issue
lube didnt make the floor slippery

it should

#### Steps to reproduce
1) splash some lube to the floor
2) walk on it

Server Revision: 2788d4f07cd92bf2ceb4e8389f29fc873d30758e - dev - 2022-06-19

#### Issue bingo

### LINKED ISSUE #2584: Utility frames spawn with no cell
#### Description of issue
Utility frames spawn on join in crit due to having no cell.

They should have a cell.

#### Steps to reproduce
Join as a utility frame.

Unclear.

Scav dev branch, but this is effectively Neb staging.

#### Issue bingo

### LINKED ISSUE #2589: Borgs cannot use several of their modules
#### Description of issue
Borgs cannot use several of their modules due to dexterity checks.

Borgs should be able to use their built-in modules.

#### Steps to reproduce
* Get miner borg. 
* Equip drill module. 
* Try drilling.
* Notice how it says "you're not dexterous enough" when using the drill. 

This also happens with a lot of other items that require dexterity to use.

At least 2 weeks.

6a1dcd035e5c940b65cf582e3722a057726fa7c0

#### Issue bingo

MistakeNot4892:
- Fixes #2589
- Prevents erroneous energy gun charge overlays rendering on mobs.
- Fixes missing attack selector icon from some themes, and makes larger/centered to properly show text.
- Removes scaling from fullscreen overlays - a proper fix for this is needed but this way they are at least visible and obvious rather than almost impossible to see.
- Moved blob processing to a subsystem so that timing can be controlled - semi-fix for #2118
- Adding some logic around explosions damaging limbs due to torsos being destroyed and bugging out the mob.
- Fixes #2545
- Fixes https://github.com/ScavStation/ScavStation/issues/584
- Various tweaks from reported Scav runtimes.
- Amends APC UI to make #2440 less confusing.
- Fixes #2584.


code/datums/movement/mob.dm
```diff
@@ -222,6 +222,10 @@
 	return T && !((mob_flags & MOB_FLAG_HOLY_BAD) && check_is_holy_turf(T))
 
 /mob/proc/AdjustMovementDirection(var/direction)
+
+	if(!direction || !isnum(direction))
```
PsiOmegaDelta, line 226: Feels like https://github.com/NebulaSS13/Nebula/blob/27a6f06844510b7a5dcdc0ae3908c443e73de816/code/datums/movement/movement.dm#L81 should be rejecting attempts to move in invalid directions before we even get here

code/modules/mob/living/carbon/human/human_defense.dm
```diff
@@ -468,37 +468,38 @@ meteor_act
...
-				SET_STATUS_MAX(src, STAT_PARA, 10)
-
-	// focus most of the blast on one organ
-	apply_damage(0.7 * b_loss, BRUTE, null, DAM_EXPLODE, used_weapon = "Explosive blast")
-	apply_damage(0.7 * f_loss, BURN, null, DAM_EXPLODE, used_weapon = "Explosive blast")
-
-	// distribute the remaining 30% on all limbs equally (including the one already dealt damage)
-	apply_damage(0.3 * b_loss, BRUTE, null, DAM_EXPLODE | DAM_DISPERSED, used_weapon = "Explosive blast")
-	apply_damage(0.3 * f_loss, BURN, null, DAM_EXPLODE | DAM_DISPERSED, used_weapon = "Explosive blast")
+	if(!QDELETED(src))
```
PsiOmegaDelta: Since I've already made another comment suppose I might as well ask, why not
```
if (QDELETED(src))
    return
[...]
```
in order to avoid indentation changes?

code/modules/mob/mob.dm
```diff
@@ -1006,10 +1006,18 @@
...
+/mob/get_contained_external_atoms()
 	. = ..()
-	if(!blinded)
-		flash_eyes()
+	if(.)
+		LAZYREMOVE(., get_organs())
+
+/mob/explosion_act(var/severity)
+	. = ..()
+	if(!QDELETED(src))
```
PsiOmegaDelta, line 1016: Is this something that can happen because an exploded object both deletes its contents, and also explosion_acts them?

code/datums/movement/mob.dm
PsiOmegaDelta, line 226: Feels like https://github.com/NebulaSS13/Nebula/blob/27a6f06844510b7a5dcdc0ae3908c443e73de816/code/datums/movement/movement.dm#L81 should be rejecting movements in invalid directions (and perhaps even log/crash them) before we even get here

code/modules/mob/mob.dm
MistakeNot4892: Previously things have qdeleted lower down in the chain due to calling physically_destroyed, I don't think it's likely with /mob but it's not an expensive check.

code/modules/mob/living/carbon/human/human_defense.dm
MistakeNot4892: Initially I had some logic happening outside of this block, looks like I removed the need for it.

code/modules/mob/mob.dm
MistakeNot4892: Okay yeah went back and checked. If atom-level explosion act triggers a reagent explosion, the atom might be qdeleted by the time it returns control to the parent proc.

code/datums/movement/mob.dm
MistakeNot4892: Added similar checks to DoMove() as discussed, as wheelchairs and janicarts call AdjustMovementDirection() directly during relaymove.

PsiOmegaDelta (approved):

PR #2592, opened 2022-08-03
Macroizes get_area().

MistakeNot4892:
By request of @Lohikar. Has apparently been live on Aurora for years.

### DIFF
--- a/code/_helpers/game.dm
+++ b/code/_helpers/game.dm
@@ -16,11 +16,6 @@
 					return TRUE
 	return FALSE
 
-/proc/get_area(O)
-	RETURN_TYPE(/area)
-	var/turf/loc = get_turf(O)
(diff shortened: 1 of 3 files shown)


code/_helpers/game.dm
```diff
@@ -16,11 +16,6 @@
 					return TRUE
 	return FALSE
 
-/proc/get_area(O)
-	RETURN_TYPE(/area)
-	var/turf/loc = get_turf(O)
-	return loc?.loc
```
afterthought2, line 22: The new version clearly misses the subtle historical reference here to the superior implementation `O.loc:loc`. We are forgetting our heritage.

afterthought2 (approved):
But not really; if anything this is overdue.

PR #2593, opened 2022-08-03
Makes firedoors slightly more performant with area changing.

afterthought2:
## Description of changes

What it says on the tin. The event it was using didn't filter for what had its area changed, so if e.g. you mass-changed an area with a lot of turfs and 5 things per turf, it would fire 5 * (number of turfs) * (number of doors) times. This hook will fire 5 * number of turfs changed, doing most of the work only when a door is adjacent, so something like 5 * (number of doors) + (small constant) * (number of turfs). I'd guess this is like ~10 to 100 times less with shuttles, but haven't tested.

That said, it's not pretty and it's not super efficient.

Apparently current version is awful. This might be slightly better.

## Authorship

me

### DIFF
--- a/code/game/area/areas.dm
+++ b/code/game/area/areas.dm
@@ -120,8 +120,18 @@ var/global/list/areas = list()
 	for(var/obj/machinery/M in T)
 		M.area_changed(old_area, A) // They usually get moved events, but this is the one way an area can change without triggering one.
 
+	T.update_registrations_on_adjacent_area_change()
+	for(var/direction in global.cardinal)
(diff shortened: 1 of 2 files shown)


afterthought2:
I should note that an alternative might be to partially or fully revert #2543, possibly making firedoors deconstruct when a blueprint is used and maybe force-updating them post shuttle move manually in shuttle code if accuracy is valued. That felt a bit too gross but may be the only efficient option.

MistakeNot4892 (approved):

PR #2599, opened 2022-08-06
Computer directories

NataKilar:
## Description of changes
Reworks computer file systems to support directories.
**File Storage**
* The ``stored_files`` variable on hard drives is now a dictionary, with the key being each file (as before), and the value being the directory it is contained within, if it exists.
* Directories are subtypes of computer files, with the same permissions system. Directories hold weakrefs to their held files, but may also temporarily hold actual references to their held files while they are being transferred between hard drives, or before they are stored post creation.
* Most procs for storing, creating, updating, or moving files now requires a passed directory as well. For most procs, the passed directory can either be a string representing the path e.g. ``/local/documents` or the directory file itself.
* Certain files are now stored in default directories, such as ``programs``
* Most file system procs no longer return boolean values, but either ``OS_FILE_SUCCESS`` or various error codes. Since there's a variety of ways in which these procs can fail, this makes it much easier to return feedback to the player.
* Instances of ``/datum/file_storage`` are now semi-permanent, acting as a layer between the OS and a hard drive somewhere in the world. They are removed/added on reboot.
* are stored in the ``mounted_storage`` variable on OSs. This is a dictionary keyed by the "root name" of the storage, with the value being the datum itself. By default, the local hard drive has the root name ``local``
* Any path referencing files managed by the datum is prefixed by the root name. For instance, ``/local/documents`` vs ``/media/documents``. The hard drive itself is unaware of the root name, and so this is only relevant when parsing directories on the OS procs.

**File Manager and File Browser**
* The file manager has been completely reworked. The file manager is now able to open multiple directories at once, and transfer between them freely.
* The file manager can now mount instances of ``/datum/file_storage/network`` to access multiple network drives simultaneously.
* A generic file browser utility has been added as a seperate nanomodule, accessible by programs. This makes it much easier to add player facing interactions with the file system.

**Terminal**
* Adds directory parsing to terminals. Terminals are able to use either absolute or relative paths.

There's a few things that haven't been cinched quite yet. Most programs have been integrated to use directories and/or the file browser, but emails have not yet, since that will be a larger change. The concept of mainframe roles could potentially be done away with, but regardless various network functions should probably be looking at specific directories rather than every file on a mainframe. As of yet, there's still not a great way to modify permissions on files en masse. A way to do this in the file manager is a good next step. Finally, the UI for both the file browser and file manager could use some work to be more user friendly.

**Screenshots:**
image
image
image

Current file systems in game are becoming extremely crowded, particularly on mainframes. It's difficult to set access in a reasonable manner or find desired files quickly. The file manager program is unintuitive, and is bad at transferring files between local and network file systems. The new file manager fixes these issues. Furthermore, the new file browser module cuts down on repeated code and difficulty with user facing selection of files. Directories are an important part of a convincing computer simulation, and this brings us closer to that goal.

## Authorship
Myself


MistakeNot4892:
We need to review this. :(

comma (approved):
Sorry for hyper delay, actually looked through it and looks fine to me

PR #2607, opened 2022-08-09
Make small common machines less annoying to build/repair

PsyCommando:
## Description of changes
After a lot of complaints downstream, and after a station-wide EMP I decided to make some of these things be less of a massive pain in the ass to repair/build.
* Alarms have a built-in unbreakable tesla-link and don't need a console screen and keyboard part anymore to be usable.
* Pipe meters have unbreakable built-in components now.
* Buttons have unbreakable tesla-link in them now.
* Lightswitches have unbreakable tesla-links.
* Lights have unbreakable tesla-links in them now.
* Airlock sensors, and airlock access buttons have unbreakable tesla-links now.

It should help people from giving up on fixing things because it turns into a massive chore nobody wants to do. 

## Changelog
:cl:
tweak: Alarms have a built-in unbreakable tesla-link and don't need a console screen and keyboard part anymore to be usable.
tweak: Pipe meters have unbreakable built-in components now.
tweak: Buttons have unbreakable tesla-link and transmitters/receiver in them now.
tweak: Light switches have unbreakable tesla-links.
tweak: Lights have unbreakable tesla-links in them now.
tweak: Airlock sensors, and airlock access buttons have unbreakable parts now.
/:cl:

### DIFF
--- a/code/game/machinery/alarm.dm
+++ b/code/game/machinery/alarm.dm
@@ -58,8 +58,7 @@
 
 	base_type = /obj/machinery/alarm
 	frame_type = /obj/item/frame/air_alarm
-	stat_immune = 0
-	uncreated_component_parts = null
+	uncreated_component_parts = list(/obj/item/stock_parts/power/apc = 1)
 	construct_state = /decl/machine_construction/wall_frame/panel_closed
 	wires = /datum/wires/alarm
 
@@ -831,8 +830,7 @@ FIRE ALARM
 
 	base_type = /obj/machinery/firealarm
 	frame_type = /obj/item/frame/fire_alarm
-	stat_immune = 0
-	uncreated_component_parts = null
+	uncreated_component_parts = list(/obj/item/stock_parts/power/apc = 1)
 	construct_state = /decl/machine_construction/wall_frame/panel_closed
 
 	var/detecting =    TRUE
--- a/code/game/machinery/buttons.dm
+++ b/code/game/machinery/buttons.dm
@@ -32,7 +32,7 @@
 
 /obj/machinery/button/buildable
 	uncreated_component_parts = list(
-		/obj/item/stock_parts/radio/transmitter/basic = 1,
+		/obj/item/stock_parts/power/apc = 1,
 	)
 
 /obj/machinery/button/Initialize()
--- a/code/game/machinery/buttons.dm
+++ b/code/game/machinery/buttons.dm
@@ -17,10 +17,10 @@
 	public_methods = list(/decl/public_access/public_method/toggle_input_toggle)
 	stock_part_presets = list(/decl/stock_part_preset/radio/basic_transmitter/button = 1)
 	uncreated_component_parts = list(
-		/obj/item/stock_parts/power/apc/buildable,
-		/obj/item/stock_parts/radio/transmitter/basic/buildable
+		/obj/item/stock_parts/power/apc = 1,
+		/obj/item/stock_parts/radio/transmitter/basic = 1
 	)
-	base_type = /obj/machinery/button/buildable
+	base_type = /obj/machinery/button
 	construct_state = /decl/machine_construction/wall_frame/panel_closed/simple
 	frame_type = /obj/item/frame/button
 	required_interaction_dexterity = DEXTERITY_SIMPLE_MACHINES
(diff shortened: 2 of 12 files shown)


code/game/machinery/alarm.dm
```diff
@@ -58,8 +58,7 @@
 
 	base_type = /obj/machinery/alarm
 	frame_type = /obj/item/frame/air_alarm
-	stat_immune = 0
```
MistakeNot4892, line 61: Why remove stat_immune here?

code/game/machinery/buttons.dm
```diff
@@ -17,10 +17,10 @@
 	public_methods = list(/decl/public_access/public_method/toggle_input_toggle)
 	stock_part_presets = list(/decl/stock_part_preset/radio/basic_transmitter/button = 1)
 	uncreated_component_parts = list(
-		/obj/item/stock_parts/power/apc/buildable,
-		/obj/item/stock_parts/radio/transmitter/basic/buildable
+		/obj/item/stock_parts/power/apc = 1,
```
MistakeNot4892, line 20: The changes to the same list above aren't assoc, what's the significance of these being assoc here?

MistakeNot4892:
Would appreciate if @afterthought2 could have a look at this one as I don't understand the significance/impact of some of these changes.

code/game/machinery/alarm.dm
PsyCommando: Because the keyboard and screen components are indestructible anyways on the alarm. So its redundant.

code/game/machinery/buttons.dm
PsyCommando: There's no difference. It just assumes 1 when there's no assoc values. But its meant to be an assoc list. So its probably better to make sure it always is one.

PsyCommando:
> Would appreciate if @afterthought2 could have a look at this one as I don't understand the significance/impact of some of these changes.

In essence, putting parts into uncreated_components means they spawn with the machine always even when built from parts. And replacing the parts with their non buildable types means they can't be removed and don't take damage. 
Essentially what I've done is make things like light switches always spawn with an indestructible and unremovable tesla link, so you don't need to slap one inside each times you build a lightswitch, or have to replace the tesla link in all lightswitches that got emp-ed..

code/game/machinery/alarm.dm
afterthought2: I think the preferred solution is to let stat_immune default to being screen/input immune but then not add non-interactive parts.

code/game/machinery/buttons.dm
```diff
@@ -17,10 +17,10 @@
...
+		/obj/item/stock_parts/radio/transmitter/basic = 1
```
afterthought2: This I'm against; there's a massive difference between an interactive and non-interactive part here. A button with a non-interactive transmitter is basically worthless unless it's pre-mapped.

afterthought2 (commented):
I don't mind the power/keyboard/screen changes. I do mind the non-interactive radio components. I think you could look into having them printed with interactive components (`part_stash` extension and similar on certain other entities), but you have to configure those parts to have them function, and for that you need to be able to take them out.

I think the code-level approach here is OK for the most part.

code/game/machinery/alarm.dm
PsyCommando: Pretty sure that's what I did? By non-interactive you mean parts that have the flag to prevent them from being removable by hands?

code/game/machinery/buttons.dm
PsyCommando: Yeah, that was my bad. Fixed it.

code/game/machinery/buttons.dm
```diff
@@ -31,7 +31,9 @@
 	var/cooldown = 1 SECOND
 
 /obj/machinery/button/buildable
-	uncreated_component_parts = null
+	uncreated_component_parts = list(
+		/obj/item/stock_parts/radio/transmitter/basic = 1,
```
afterthought2: Still seeing things with noninteractive radio components.

PsyCommando: Well, actually that was supposed to be an apc component, so I fixed that XD

PsyCommando:
Seems like the damn desperado ship is having that weird issue where its failing atmos checks again.

PsyCommando:
Adding a random space to force testing again worked :D

afterthought2 (approved):

PR #2608, opened 2022-08-11
Update item update icon

PsyCommando:
## Description of changes
I decided to move part of the suggested fixes for #2552 into its own PR.
This makes sure that all items overriding the on_update_icon() proc call the parent and properly use the overlay system.

## Changelog
:cl:
refactor: Refactored item on_update_icon() so it all uses the overlay system.
/:cl:


code/game/objects/items/devices/transfer_valve.dm
```diff
@@ -128,22 +128,24 @@
...
-		overlays += "[tank_one.icon_state]"
+		LAZYADD(cur_overlays, new /mutable_appearance(tank_one))
 	if(tank_two)
 		var/icon/J = new(icon, icon_state = "[tank_two.icon_state]")
 		J.Shift(WEST, 13)
 		underlays += J
 	if(attached_device)
-		overlays += "device"
+		LAZYADD(cur_overlays, overlay_image(icon, "device"))
+	add_overlay(cur_overlays)
```
MistakeNot4892, line 148: `set_overlays`

MistakeNot4892: As mentioned on Discord, SSoverlays implicitly avoids appearance churn, there's no benefit to collecting overlays before adding them.

code/game/objects/items/devices/transfer_valve.dm
```diff
@@ -128,22 +128,24 @@
...
 
 	if(!tank_one && !tank_two && !attached_device)
 		icon_state = "valve_1"
 		return
 	icon_state = "valve"
 
+	var/list/cur_overlays
 	if(tank_one)
-		overlays += "[tank_one.icon_state]"
+		LAZYADD(cur_overlays, new /mutable_appearance(tank_one))
```
MistakeNot4892: We have a `mutable_appearance()` proc similar to `overlay_image()` for this.

code/game/objects/items/devices/transfer_valve.dm
MistakeNot4892: As per Discord, you don't need to generate images for string states if they're using the root icon of the object, SSoverlays handles that itself.

code/game/objects/items/weapons/implants/implanter.dm
```diff
@@ -21,10 +21,8 @@
 	. = ..()
 
 /obj/item/implanter/on_update_icon()
-	if (imp)
-		icon_state = "implanter1"
-	else
-		icon_state = "implanter0"
+	. = ..()
+	icon_state = "implanter[imp == TRUE]"
```
MistakeNot4892, line 25: `!!imp` would also work here

code/game/objects/items/weapons/swords_axes_etc.dm
```diff
@@ -61,14 +61,13 @@
 	update_held_icon()
 
 /obj/item/telebaton/on_update_icon()
+	if(length(blood_DNA))
```
MistakeNot4892, line 64: Isn't this redundant with `..()`?

MistakeNot4892:
Generally looks good, feel free to remove the collecting or not (per Discord discussion with Lohi). Some of the overlay_image() calls could be tidied to bare strings but that should probably be done separately as it's a broader issue on the codebase.

code/game/objects/items/devices/transfer_valve.dm
PsyCommando: The proc doesn't accept an object though.

code/game/objects/items/weapons/swords_axes_etc.dm
PsyCommando: Nah, its generating the blood overlay overlay, the parent only applies the blood overlay.

PsyCommando:
Related to adding a SHOULD_CALL_PARENT check, this is proving a bit more complicated..
Going back over the PR I noticed that I'm not handling chems and a few other items types. So gonna fix that really quick.
The only annoying thing is that chems do a lot of weirdness and generally seem to disregard parent implementations of on_update_icon and do their own thing..

code/game/objects/items/devices/transfer_valve.dm
PsyCommando: I removed code needlessly creating lists for adding overlays in a bunch of places.

code/game/objects/items/weapons/implants/implanter.dm
PsyCommando: My beef with using !! is that its easy to misread it and/or consider it a typo. Maybe if we could somehow make a macro for it it would be more readable?

MistakeNot4892: I don't mind either way for this PR, I think a macro is probably overkill for such a simple/foundational operation though.

code/game/objects/item_materials.dm
```diff
@@ -1,10 +1,12 @@
 /obj/item/on_update_icon()
-	overlays.Cut()
+	. = ..()
+	SHOULD_CALL_PARENT(TRUE)
```
MistakeNot4892, line 3: For clarity's sake, linter macros should probably go at the absolute top of the proc.
Also a bit unclear on what is happening in `..()` here, on my local copy there's no `/obj/on_update_icon()` and you're setting the linter proc to SHOULD_CALL_PARENT(FALSE) in the base proc.

code/modules/reagents/reagent_containers/food/sliceable/pizza.dm
```diff
@@ -162,10 +161,9 @@
 			icon_state = "pizzabox_open"
 
 		if( pizza )
-			var/image/pizzaimg = image("food.dmi", icon_state = pizza.icon_state)
+			var/mutable_appearance/pizzaimg = new(pizza)
```
MistakeNot4892, line 164: Why switch this to a mutable appearance?

code/modules/detectivework/tools/sample_kits/fingerprinting.dm
```diff
@@ -20,8 +20,11 @@
 	possible_evidence_types = list(/datum/forensics/fingerprints)
 
 /obj/item/forensics/sample/print/on_update_icon()
+	. = ..()
 	if(length(evidence))
 		icon_state = "fingerprint1"
+	else
+		icon_state = "fingerprint0"
```
MistakeNot4892, line 27: Could just be `fingerprint[!!length(evidence)]`

code/modules/detectivework/tools/rag.dm
```diff
@@ -57,11 +57,8 @@
 		SetName("dry [initial(name)]")
 
 /obj/item/chems/glass/rag/on_update_icon()
-	if(on_fire)
-		icon_state = "raglit"
-	else
-		icon_state = "rag"
-
+	. = ..()
+	icon_state = "rag[on_fire? "lit" : ""]"
```
MistakeNot4892, line 61: `[on_fire && "lit"]` might read a bit cleaner.

MistakeNot4892: Nvm that'd produce `0`

code/modules/reagents/reagent_containers/food/sliceable/pizza.dm
MistakeNot4892: Nvm I think I follow, it's just reusing the icon/state of the parent so might as well use appearance.
My concern there would be whether or not this is layering the icon properly when dropped/picked up.

code/game/objects/item_materials.dm
PsyCommando: That's because there's a problem here with the idea of slapping in a SHOULD_CALL_PARENT like you suggested last time in discord. I mentioned it in the previous comment on the PR. 
/atom/on_update_icon() is a stub. We don't want to call the stub since its a waste of time. But the SHOULD_CALL_PARENT macro makes it so /obj/item/on_update_icon() has to call the parent stub.. The objective was just to force children of /obj/item to call their parents to avoid problems.

code/modules/reagents/reagent_containers/food/sliceable/pizza.dm
PsyCommando: why wouldn't it?

MistakeNot4892: Plane/layer stuff when moving between inventory and world. I'll assume it works, want to get this PR merged.

MistakeNot4892 (approved):

PR #2609, opened 2022-08-15
Quadruped species supporting code.

MistakeNot4892:
## Description of changes
- `bodytype_flag` is now `bodytype_flags`.
- Bodytypes can now override species direction when prone. This is because east/west facing long sprites look absurd when prone.
- Mapping of arms/legs to hands/feet is now handled on the species datum, allowing overriding for nonstandard mobs downstream.
- Legs/feet (stance limbs in general) are handled by lists on the species datum so that they can be overridden. May move all this to bodytypes at some point.

Adds some support for species with four legs/long bodies.

## Authorship
Myself.

## Changelog
No player facing changes hopefully.


MistakeNot4892:
Going to come back to this one after #2610

PR #2610, opened 2022-08-15
Moving organ and limb lists onto bodytypes.

MistakeNot4892:
## Description of changes
- Moves a bunch of organ tracking onto the bodytype instead of the species.
- Also removes the deprecated vital organ tracking logic.
- Unifies prosthetic models with bodytypes, as they share a fair chunk amount of code/vars.

Species is less of a monolith, bodytype is more fleshed out.

## Authorship
Myself.

## Changelog
Nothing player-facing.


MistakeNot4892:
This is not a proper and complete move-over and merge of the involved code, but it appears to be working in testing and is good groundwork for more work, so I'm calling this ready for review.

MistakeNot4892:
Binning this for now, it will introduce a lot of issues and doesn't achieve anything functional. To the todo pile with you.

PR #2614, opened 2022-08-16
Immutable gasses

MistakeNot4892:
## Description of changes
- Mirrors a staging fix to exoplanet gas overlay retrieval.
- Adds some unit test-gated linting around direct modification or access of gas mix lists. This includes wrapping every extant read with a macro due to the lack of a read-only flag in the linter.
- Converts exoplanet and global exterior atmos to a new immutable gas mixture type. This should in theory cut down on unnecessary generation of new gas mixtures (though the temperature modification code still generates a new copy due to needing to mutate the temperature value).

Less memory use/overhead from exterior atmos, hopefully. 

## Authorship
Myself for code, Kapu and Noelle for discussion.

## Changelog


MistakeNot4892:
Will reopen when this is ready for merge.

PR #2618, opened 2022-08-20
ZAS CSRFZ Webedit

Kapu1178:
## Description of changes

Port of my Daedalus Dock and Baystation PRs.

Basically, if a turf is only connected to a zone by 1 direction, it is ALWAYS a safe removal. This greatly improves performance, because airlocks no longer trigger zone rebuilds erroneously.

## Authorship

Lohikar for the actual bitwise logic, myself for the "idea".

## Changelog
:cl: Lohikar, Kapu1178
code: Greatly reduced the amount of zone rebuilds.
/:cl:

### DIFF
--- a/code/modules/ZAS/Turf.dm
+++ b/code/modules/ZAS/Turf.dm
@@ -56,6 +56,10 @@
 	var/check_dirs = get_zone_neighbours(src)
 	var/unconnected_dirs = check_dirs
 
+	//src is only connected to the zone by a single direction, this is a safe removal.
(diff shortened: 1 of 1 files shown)


comma (approved):

MistakeNot4892:
@Lohikar had some issues with this one, unsure if they were blocking.

Lohikar:
Only issues I have with this is that the entire proc should just be replaced with the correct version instead of making fixes to the old version that was supposed to be replaced years ago -- actual changes are fine.

MistakeNot4892:
Going to go ahead and merge this one then, a full replacement can happen when someone can be bothered.

PR #2619, opened 2022-08-20
Fixes to runtimes on Scav.

MistakeNot4892:
## Description of changes
- Fixes some alarm issues resulting from switching to material identifiers for gasses.
- Moves virtual_mob cleanup to before event cleanup.
- Cleans up some special role name retrieval code.
- Fixes atoms not cleaning up their contents in Destroy().
- ~~Makes event cleanup more sensitive for unit testing purposes.~~ Can't chase down some of these to get this to pass CI, so reverting for now.

Fixes.

## Authorship
Myself, Psi and Noelle advised on virtual mob stuff.

## Changelog
Nothing player-facing.


code/datums/mind/mind.dm
```diff
@@ -570,6 +570,9 @@
 	mind.assigned_special_role = "Cultist"
 
 /datum/mind/proc/get_special_role_name()
```
out-of-phaze: This can still return null, is that intended? Maybe you could make it `get_special_role_name(default = null)` to avoid having to pepper a dozen versions of `|| "No special role"`/`|| "unknown role"` all over the place. Having it be null by default would prevent it from messing with the one case that expects it to be a null, which is the hud icon stuff.

MistakeNot4892: Current behavior accounts for null returns, but a default fail result might be worth adding to streamline it.

out-of-phaze (approved):

comma (approved):

afterthought2:
Slightly hesitant on the Destroy reordering to staging, but let's try it.

PR #2621, opened 2022-08-20
Rewrite cocktails to use an impurity tolerance var

out-of-phaze:
## Description of changes
- Adds an `impurity_tolerance` var to cocktails, which takes the place of having the ratio sum to a number less than 1.0; 0.3 means 70%
- Improves comment formatting for cocktail vars, so that the right comments show up when you mouse over it with the langserver running.
- Normalises cocktail ratios so that all the ingredient amounts are integers.
- Makes the codex show the original, unmodified, integer ratio for cocktails, and changes some of the wording for it.
- Tweaks some cocktails to be more consistent/accurate, have nicer numbers, or show off some of the cocktail features (like 'to taste' ingredients).

It's really annoying having to manually handle the impurity tolerance. This improves both the dev experience when setting the numbers and the user experience when viewing them, since you know that whatever ratio you set is what the user will see in the codex.

## Authorship
Loaf/MistakeNot4892 for most of the code, me for the cocktail numbers wiggling, documentation updates, and codex integration.

## Changelog
:cl:
tweak: Broadly retools cocktail ratios, changing some cocktail ingredient amounts and improving their display in the codex.
/:cl:

### DIFF
--- a/code/modules/reagents/cocktails.dm
+++ b/code/modules/reagents/cocktails.dm
@@ -1,19 +1,43 @@
 // This is a system used to change the name and description of a glass of
-// alcohol if it meets certain minimum proportions of ingredients. It 
+// alcohol if it meets certain minimum proportions of ingredients. It
 // replaces the previous system, which used chemical reactions.
 
 /decl/cocktail
-	var/name                   // Cocktail name, applied to the glass.
-	var/description            // Cocktail description, applied to the glass.
-	var/list/ratios            // Associative list of reagents to a minimum percentage of mix (<1).
-	                           // Reagents with no assoc value will count as valid for any amount (even 0.001u)
-	var/order_specific = FALSE // If set, cocktail will fail if ingredients were added out of ratio order.
-	var/hidden_from_codex      // Doesn't generate a codex entry.
+	/// Cocktail name, applied to the glass.
+	var/name
+	/// Cocktail description, applied to the glass.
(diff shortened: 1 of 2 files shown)


code/modules/reagents/cocktails.dm
```diff
@@ -1,19 +1,43 @@
...
+	/// If TRUE, cocktail ingredients must be added in the order they're specified in the ratio.
+	var/order_specific = FALSE
+	/// If TRUE, doesn't generate a codex entry.
+	var/hidden_from_codex
+
+	// Impurity tolerance gives a buffer for imprecise mixing, avoiding finnicky measurements
+	// and allowing for things like spiked drinks. The default is 0.3, meaning aside from ice,
+	// the drink can be at most 30% other reagents not part of the cocktail recipe.
+
+	/// What ratio of the total volume of the drink (ignoring ice) can be unrelated chems?
```
MistakeNot4892: Nitpick but this should be 'what percentage', I think this wording was what I wrote on Discord so apologies for that

MistakeNot4892 (dismissed):

code/modules/reagents/cocktails.dm
out-of-phaze: a percentage is from 0-100, this is technically a fraction ;)

MistakeNot4892 (approved):

PR #2622, opened 2022-08-21
Fix QDELETED looping timers being reinserted

### LINKED ISSUE #2517: Timer runtimes
#### Description of issue

Frequently timer runtimes during rounds. 
The runtime is always the same other than the timer number.

Timers shouldn't be runtiming.

#### Steps to reproduce

Watch the console output during rounds

````
<!-- e.g. an object name, paste specific message outputs... -->
[11:14:47] Runtime in timer.dm,494: Cannot read null.object
  proc name: bucketJoin (/datum/timedevent/proc/bucketJoin)
  src: Timer: 123 (\[0x21006695]), TT... (/datum/timedevent)
  call stack:
  Timer: 123 (\[0x21006695]), TT... (/datum/timedevent): bucketJoin()
  Timer (/datum/controller/subsystem/timer): fire(0)
  Timer (/datum/controller/subsystem/timer): ignite(0)
````

At least since i've started hosting scavstation, so at minimum ~6 weeks, but likely far longer

Client Version: 514
Server Revision: 013e91177a69345c0a692db0b6f73915933075ce - dev - 2022-06-09
Game ID: cip-aakB

#### Issue bingo

out-of-phaze:
## Description of changes
Fixes a timer runtime that can currently be reproduced by mouse highlighting an object one tile away and moving out of range while still holding shift. This was because the timer was qdeleted inside its callback, which wasn't handled properly.

Closes #2517.
Also adds a more informative message if something else causes the same problem through a different method.

## Authorship
LemonInTheDark from /tg/station for pointing out the issue, me for the fix.

### DIFF
--- a/code/controllers/subsystems/timer.dm
+++ b/code/controllers/subsystems/timer.dm
@@ -172,7 +172,9 @@ SUBSYSTEM_DEF(timer)
 				callBack.InvokeAsync()
 				last_invoke_tick = world.time
 
-			if (timer.flags & TIMER_LOOP) // Prepare looping timers to re-enter the queue
+			if (timer.flags & TIMER_LOOP) // Prepare valid looping timers to re-enter the queue
+				if(QDELETED(timer)) // If we were deleted in the callback, don't re-insert.
+					continue
 				timer.spent = 0
 				timer.timeToRun = world.time + timer.wait
 				timer.bucketJoin()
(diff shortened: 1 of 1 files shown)


Lohikar (approved):
*Technically* this could cause the timer SS to overrun if there's a pile of QDELETED timers being skipped, but I think that's unlikely enough to not be worth handling.

PR #2629, opened 2022-08-21
Attempting to address some performance issues seen on Scav testing.

MistakeNot4892:
## Description of changes
- Breaks temperature queuing and proximity updates into their own procs for profiling purposes.
- Removes a `spawn(0)` from proximity updates in favour of `waitfor = FALSE`.
- Refactors some pain handling code with a view towards optimizing it somewhat.
- Condenses /turf/Entered() overrides into a single proc to avoid multiple parent call overhead and make it a bit easier to read.
- Human pain now uses a cache var to avoid multiple loops over organs each tick. The var is cleared in human Life() and updated on getHalLoss() call if null.

Hoping it addresses some lagspikes and general overhead issues on Scav.

## Authorship
Myself.

## Changelog
Nothing player-facing.

### DIFF
--- a/code/game/turfs/turf_enter.dm
+++ b/code/game/turfs/turf_enter.dm
@@ -1,0 +1,54 @@
+#define ENTER_PROXIMITY_LOOP_SANITY 100
+// Splitting this into its own proc for profiling purposes.
+/turf/proc/handle_proximity_update(var/atom/movable/mover)
+	var/objects = 0
+	if(!istype(mover) || !(mover.movable_flags & MOVABLE_FLAG_PROXMOVE))
+		return
+	for(var/atom/movable/neighbor in range(1))
+		if(objects > ENTER_PROXIMITY_LOOP_SANITY) 
+			break // Don't let ore piles kill the server as well as the client.
+		objects++
+		mover.HasProximity(neighbor)
+		if(!QDELETED(neighbor) && !QDELETED(mover) && (neighbor.movable_flags & MOVABLE_FLAG_PROXMOVE))
+			neighbor.HasProximity(mover)
+
+#undef ENTER_PROXIMITY_LOOP_SANITY
+/turf/Entered(var/atom/atom, var/atom/old_loc)
+
+	..()
+
+	if(!istype(atom, /atom/movable))
+		return
+
+	if(ishuman(atom))
+		var/mob/living/carbon/human/H = atom
+		H.handle_footsteps()
+
+	queue_temperature_atoms(atom)
+
+	var/atom/movable/A = atom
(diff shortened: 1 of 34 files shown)


code/game/turfs/turf_enter.dm
```diff
@@ -0,0 +1,54 @@
...
+	for(var/atom/movable/neighbor in range(1))
+		if(objects > ENTER_PROXIMITY_LOOP_SANITY) 
+			break // Don't let ore piles kill the server as well as the client.
+		objects++
+		mover.HasProximity(neighbor)
+		if(!QDELETED(neighbor) && !QDELETED(mover) && (neighbor.movable_flags & MOVABLE_FLAG_PROXMOVE))
+			neighbor.HasProximity(mover)
+
+#undef ENTER_PROXIMITY_LOOP_SANITY
+/turf/Entered(var/atom/atom, var/atom/old_loc)
```
Lohikar: It would be better to type this as a movable instead; none of the logic actually expects this to not be a movable, so the early exit code being mistyped is fine.

code/game/turfs/turf_enter.dm
```diff
@@ -0,0 +1,54 @@
...
+// Splitting this into its own proc for profiling purposes.
+/turf/proc/handle_proximity_update(var/atom/movable/mover)
+	var/objects = 0
+	if(!istype(mover) || !(mover.movable_flags & MOVABLE_FLAG_PROXMOVE))
+		return
+	for(var/atom/movable/neighbor in range(1))
+		if(objects > ENTER_PROXIMITY_LOOP_SANITY) 
+			break // Don't let ore piles kill the server as well as the client.
+		objects++
+		mover.HasProximity(neighbor)
```
afterthought2: Quickly looking over the implementation we find:
```
// If you want to use this, the atom must have the PROXMOVE flag, and the moving
// atom must also have the PROXMOVE flag currently to help with lag. ~ ComicIronic
/atom/proc/HasProximity(atom/movable/AM)
```
This suggests to me that calling this on neighbors without the flag set (i.e. before the check on the line following) is not necessary for the system to function. Maybe that's false, but it's worth checking as it would be a much better optimization.

afterthought2 (commented):
For what it's worth, `turf/Entered` overhead has historically been almost entirely due to moved/entered event listeners. Those contain a bunch of logic found in other places on other servers, but also systems with inherently poor performance like the sound player. Actually registering for turf/Entered (especially with nontrivial logic) is very expensive, as when something enters a turf each event will be called on that thing _and its contents_; this is less true for moved events as there the caller is the moving entity instead, but moved events are also resolved through Entered.

afterthought2 (approved):

PR #2277, opened 2022-01-16
Add inhalers

MarinaGryphon:
## Description of changes
Adds inhalers, which can be used to administer chemicals in a less-invasive way than autoinjectors.
- [x] Adds inhalers
    - [x] Tested
- [x] Adds oxygel inhalers to pouches
    - [x] Tested
- [x] Refactor lungs, rags, and cigarettes to use CHEM_INHALE
    - [x] Tested
- [x] Add lung lavage to sleepers to filter out chemicals
    - [ ] Tested
- [x] Make coughing clear some reagents out of the lungs
    - [x] Tested
- [x] Add a breathalyser that can detect intoxication, lung health, etc.
    - [x] Sprite
    - [x] Tested

## Why and what will this PR improve
Roleplay-wise, it's a good alternative to needles for characters who don't like them. It also adds more variety for administering chemicals, and it makes lung chemical metabolism make more sense.

## Authorship
Aurorastation, for most of the original code. I think the original implementation was by Burger? Me, for porting it to Nebula. I think most of this implementation is novel, except for autoinhalers.

## Changelog

:cl:
rscadd: Adds inhalers.
rscadd: Sleepers can now remove chemicals from the lungs via a lung lavage.
tweak: Autoinjectors in oxygen deprivation pouches have been replaced with oxygel inhalers.
tweak: Chemicals can now be properly inhaled into the lungs.
tweak: Coughing removes one to two units of reagents from your lungs.
tweak: Coughing can now only be manually done once every three seconds.
/:cl:

PR #2642, opened 2022-08-25
Adding inhalers.

MistakeNot4892:
## Description of changes
Mirroring #2277 as Noelle has moved on from it and isn't up for revival.

## Authorship
@noelle-lavenza for the vast majority of the code, which is noted as an Aurora port, I'm just fiddling with it to get it merged.

## Changelog
:cl: noelle-lavenza
rscadd: Adds inhalers.
rscadd: Sleepers can now remove chemicals from the lungs via a lung lavage.
tweak: Autoinjectors in oxygen deprivation pouches have been replaced with oxygel inhalers.
tweak: Chemicals can now be properly inhaled into the lungs.
tweak: Coughing removes one to two units of reagents from your lungs.
tweak: Coughing can now only be manually done once every three seconds.
/:cl:


Vizoplez:
This may have grand balance implications for overdosing/etc if we're adding another method to 'store' reagents in the body.

Might want to add an on_inhale special overdose for a few chems, especially opiates/other strong painkillers, especially as these would function as hyposprays pretty much anyone can use, but with a different location other than the blood.


`tweak: Autoinjectors in oxygen deprivation pouches have been replaced with oxygel inhalers.`

I personally think this may be a poor idea, but we'll see. It just feels like if someone gets decompressed their lungs would have already been burst.

MistakeNot4892:
> This may have grand balance implications for overdosing/etc if we're adding another method to 'store' reagents in the body.
> 
> Might want to add an on_inhale special overdose for a few chems, especially opiates/other strong painkillers, especially as these would function as hyposprays pretty much anyone can use, but with a different location other than the blood.

Not quite sure I follow here - do you mean people may split reagents between injected and inhaled with the intent of avoiding overdosing? A fix for that might be making overdosing use a summary value instead of going per holder.

> `tweak: Autoinjectors in oxygen deprivation pouches have been replaced with oxygel inhalers.`
>
> I personally think this may be a poor idea, but we'll see. It just feels like if someone gets decompressed their lungs would have already been burst.

This is a good point.

Vizoplez:
> Not quite sure I follow here - do you mean people may split reagents between injected and inhaled with the intent of avoiding overdosing? A fix for that might be making overdosing use a summary value instead of going per holder.

This is exactly what I meant. Before you could split reagents between the blood and stomach, and avoid certain overdoses by being 'smart' with how you did it, hence how people could drink 10u of oxycodone, and inject 15u and have a far-longer lasting effect, atop of avoiding overdoses at a higher dosage.

MistakeNot4892:
Going to do some additional work on reagents (building on #2654) before revisiting this one.

MistakeNot4892:
This has been ready for review for ages.

MistakeNot4892:
Pls review.

code/game/objects/items/devices/scanners/breath.dm
```diff
@@ -0,0 +1,118 @@
...
+		if(unknown)
+			print_reagent_default_message = FALSE
+			. += "<span class='scan_warning'>Non-medical reagent[(unknown > 1)?"s":""] found in subject's breath.</span>"
+
+	if(print_reagent_default_message)
+		. += "No results."
+
+	header = jointext(header, null)
+	. = jointext(.,"<br>")
+	. = jointext(list(header,.),null)
```
afterthought2, line 118: For what it's worth string operations were optimized in some DM update, this kind of stuff is no longer really needed.

afterthought2 (approved):
Was hoping @comma would review it for actual feedback on the system, but it's been sitting for a while. Seems fine in principle; med players tend to like this kind of stuff. Might need tuning based on play feedback as I have no sense of how it will actually feel in play.

PR #2645, opened 2022-08-25
Paperwork Rework - Taperoll rework

PsyCommando:
## Description of changes
PR split off from #2552
Tried to isolate only the part handling tape rolls. Might be some things missing still.

## Changelog
:cl:
tweak: Tweaked police tape. Now called barricade tape, and the tape itself now handles neighbours better.
/:cl:


comma (dismissed):
Could use a squash, otherwise looks fine

maps/tradeship/tradeship-1.dmm
```diff
@@ -3337,7 +3337,7 @@
 	},
 /obj/item/t_scanner,
 /obj/item/cell/high,
-/obj/item/ducttape,
+/obj/item/duct_tape,
```
MistakeNot4892: Is this a separate path to `/obj/item/stack/tape_roll/duct_tape`?

MistakeNot4892: Okay, this is now the individual tape piece, makes sense. Seems like you have find and replaced the roll item to the single item in the maps though which doesn't make much sense.

code/modules/clothing/spacesuits/breaches.dm
```diff
@@ -239,13 +239,22 @@
 		if(!target_breach)
 			to_chat(user, "There are no open breaches to seal with \the [W].")
 		else
-			playsound(src, 'sound/effects/tape.ogg',25)
 			var/mob/living/carbon/human/H = user
-			if(!istype(H)) return
+			if(!istype(H)) 
+				return
+
+			var/obj/item/stack/tape_roll/duct_tape/D = W
```
MistakeNot4892, line 246: I think it would probably be fairer on players to let them partially seal the damage with what's left of the roll, rather than blocking them from repairing at all if they don't have enough tape to cover the entire breach. Not a blocking request, but would be a nice improvement on the interaction now that tape is a consumable.

code/game/objects/items/weapons/policetape.dm
```diff
@@ -1,397 +1,555 @@
...
+	dir              = SOUTH                          //This structure will try to turn its icon depending on what neighbors it has
+	layer            = ABOVE_DOOR_LAYER
+	pass_flags       = PASS_FLAG_TABLE                //About the height of table
+	anchored         = TRUE
+	material         = /decl/material/solid/plastic
+	var/neighbors    = 0                              //Contains all the direction flags of all the neighboring tape_barricades
+	var/nb_neighbors = 0                              //Keep track of our cached neighbors number
+	var/is_lifted    = 0                              //Whether the tape is lifted and we're allowing everyone passage.
+	var/is_crumpled  = 0                              //Whether the tape was damaged
+	var/decl/barricade_tape_template/tape_template   //Details about the behavior and looks of the barricade
```
MistakeNot4892, line 206: Not blocking, but comments like these might be good on the line above with `///` so they work with dmdoc.

MistakeNot4892:
Code generally looks alright but I think that tape mapping thing is worth fixing. Not sure if it's something introduced by this PR or not, might be an older copypaste issue.

maps/tradeship/tradeship-1.dmm
PsyCommando: That's possible. Will take a look again.

code/modules/clothing/spacesuits/breaches.dm
PsyCommando: Truue, but, can you even partially repair a breach?

PsyCommando:
Squashed and fixed the paths on the maps.

comma (approved):

PR #2646, opened 2022-08-25
Paperwork rework - Paper shredder rework

PsyCommando:
## Description of changes
PR split off from #2552
This is the part about the paper shredder. Had to put stubs here because everything is just too deeply interconnected.

## Changelog
:cl:
add: Added a circuit board for assembling paper shredders.
tweak: Paper shredder gives better user feedback.
code: Reworked how the paper shredder works.
/:cl:

### DIFF
--- a/code/modules/materials/definitions/solids/materials_solid_organic.dm
+++ b/code/modules/materials/definitions/solids/materials_solid_organic.dm
@@ -85,6 +85,7 @@
 	. += new/datum/stack_recipe_list("folders", create_recipe_list(/datum/stack_recipe/folder))
 
 /decl/material/solid/paper
+	uid = "solid_paper"
 	//STUB: Waiting on papwerork PR
 
 /decl/material/solid/cloth //todo
--- a/code/modules/materials/definitions/solids/materials_solid_organic.dm
+++ b/code/modules/materials/definitions/solids/materials_solid_organic.dm
@@ -84,6 +84,9 @@
 	. += new/datum/stack_recipe/cardborg_helmet(src)
 	. += new/datum/stack_recipe_list("folders", create_recipe_list(/datum/stack_recipe/folder))
 
+/decl/material/solid/paper
+	//STUB: Waiting on papwerork PR
+
 /decl/material/solid/cloth //todo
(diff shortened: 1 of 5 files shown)


comma (approved):

code/modules/materials/definitions/solids/materials_solid_organic.dm
```diff
@@ -84,6 +84,10 @@
 	. += new/datum/stack_recipe/cardborg_helmet(src)
 	. += new/datum/stack_recipe_list("folders", create_recipe_list(/datum/stack_recipe/folder))
 
+/decl/material/solid/paper
```
MistakeNot4892, line 87: Surprised this passes unit testing without a name or anything defined. I guess materials aren't uniqueness tested for names.

code/modules/paperwork/papershredder.dm
```diff
@@ -1,148 +1,252 @@
...
-/obj/machinery/papershredder/verb/empty_contents()
-	set name = "Empty bin"
-	set category = "Object"
-	set src in range(1)
-
-	if(usr.incapacitated())
+	if(is_bin_full())
+		visible_message(SPAN_WARNING("The \"bin full\" warning light is flashing on \the [src]!"))
 		return
+	//#TODO: Uncomment this once the bug is fixed, so we check for available power before actually working
```
MistakeNot4892, line 44: What is `the bug` referred to here?

code/modules/materials/definitions/solids/materials_solid_organic.dm
PsyCommando: Well it did trigger the uid uniqueness check until I slapped the uid in

code/modules/paperwork/papershredder.dm
PsyCommando: If I recall correctly it was something with can_use_power_oneoff not returning true when there was actually sufficient power on the power net. But its been a while since I wrote that so I might be wrong here.

But from what I remember there was something funky with how power usage is checked for machines that don't have an idle power use, and just pull power one off?

PR #2649, opened 2022-08-28
say_understands refactor, animals can talk again.

MistakeNot4892:
## Description of changes
- Passes speaker to several language procs.
- Cleans up/refactors say_understands and a few language retrieval procs so simple animals can use them.
- Adds an Animal Noises language that all simple animals get, which handles understanding/verbs/scrambling based on mob type.

Animals won't just `babble aimlessly` when speaking.

## Authorship
Myself.

## Changelog
Nothing player-facing.


MistakeNot4892:
Corgis can be understood by mice even with universal_* unset, not sure why.

out-of-phaze:
> Corgis can be understood by mice even with universal_* unset, not sure why.

Mice have `universal_understand` set and this PR doesn't seem to touch it at all.

MistakeNot4892:
I unset it for testing.

MistakeNot4892:
Working now.

code/modules/mob/language/animal.dm
```diff
@@ -0,0 +1,32 @@
...
+	name = "Animal Noises" // translate them!
+	desc = "Some varieties of animal can communicate amongst themselves, apparently."
+	colour = "say_quote"
+	key = "a"
+	shorthand = "A"
+	hidden_from_codex = TRUE
+
+/decl/language/animal/can_be_understood_by(var/mob/living/speaker, var/mob/living/listener)
+	if(!istype(listener))
+		return FALSE
```
afterthought2: Possibly this should be `TRUE` instead? Do non-living mobs like ghosts and various special mobs get hit by this proc?

code/modules/mob/language/animal.dm
```diff
@@ -0,0 +1,32 @@
...
+	if(listener.universal_understand || listener.universal_speak)
+		return TRUE
+	if(!istype(speaker) || speaker.universal_speak)
+		return TRUE
```
afterthought2, line 13: Confusing and probably illegal check; if `speaker` is mistyped, you shouldn't be checking vars on it.

code/modules/mob/language/animal.dm
```diff
@@ -0,0 +1,32 @@
...
+	if(istype(speaker, listener.type) || istype(listener, speaker.type))
+		return TRUE
```
afterthought2, line 15: This is questionable at best. Imagine you have `/cat/tabby` and `/cat/calico` talking to each other; they just failed this check.

I would seriously consider checking `faction`, even though that's not ideal either.

afterthought2 (commented):
Refraining from commenting on some of the incidentally touched oldcode.

code/modules/mob/language/animal.dm
MistakeNot4892: Special mobs have universal_understand generally. I'll doublecheck if this interferes with ghosts or not.

code/modules/mob/language/animal.dm
MistakeNot4892: If `speaker` is mistyped it will short circuit before checking vars. Not sure what you mean, this isn't an uncommon pattern.

code/modules/mob/language/animal.dm
MistakeNot4892: I could add a root_species_type var or such.

code/modules/mob/language/animal.dm
afterthought2: I was wrong here, sorry.

code/modules/mob/language/animal.dm
MistakeNot4892: Added `base_animal_type`.

code/modules/mob/language/animal.dm
MistakeNot4892: Returned true for a null listener.

afterthought2 (approved):

PR #2650, opened 2022-08-29
Fix cocktail descriptions, add sprite support

out-of-phaze:
## Description of changes
Refactors `show_reagent_name` into `presentation_flags`, with support for name and description (currently).

Adds helpers and functionality for showing cocktail and reagent descriptions on containers, and adds a framework for cocktails to have sprites.

Needed for downstream, and also cocktail descriptions weren't used, only set.

## Authorship
Me.

## Changelog
:cl:
add: Drinking glasses now show the description of either their cocktail or their primary reagent, if applicable.
/:cl:

### DIFF
--- a/code/modules/materials/_materials.dm
+++ b/code/modules/materials/_materials.dm
@@ -687,6 +687,19 @@ INITIALIZE_IMMEDIATE(/obj/effect/gas_overlay)
 	if(prop.reagents.has_reagent(/decl/material/solid/ice))
 		. = "iced [.]"
 
+/decl/material/proc/get_presentation_desc(var/obj/item/prop)
+	. = glass_desc
+	if(prop?.reagents?.total_volume)
+		. = build_presentation_desc_from_reagents(prop, .)
+
+/decl/material/proc/build_presentation_desc_from_reagents(var/obj/item/prop, var/supplied)
+	. = supplied
+
+	if(cocktail_ingredient)
+		for(var/decl/cocktail/cocktail in SSmaterials.get_cocktails_by_primary_ingredient(type))
+			if(cocktail.matches(prop))
+				return cocktail.get_presentation_desc(prop)
+
 /decl/material/proc/neutron_interact(var/neutron_energy, var/total_interacted_units, var/total_units)
 	. = list() // Returns associative list of interaction -> interacted units
 	if(!length(neutron_interactions))
--- a/code/modules/reagents/cocktails.dm
+++ b/code/modules/reagents/cocktails.dm
@@ -18,6 +18,12 @@
 	var/order_specific = FALSE
 	/// If TRUE, doesn't generate a codex entry.
 	var/hidden_from_codex
+	/// The icon to use for the cocktail. May be null, in which case no custom icon is used.
+	var/icon/glass_icon
+	/// The icon_state to use for the cocktail. May be null, in which case the first state in the icon is used.
+	var/glass_icon_state
+	/// A list of types (incl. subtypes) to display this cocktail's glass sprite on.
+	var/display_types = list(/obj/item/chems/drinks/glass2)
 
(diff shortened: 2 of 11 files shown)


code/modules/materials/_materials.dm
```diff
@@ -687,6 +687,19 @@ INITIALIZE_IMMEDIATE(/obj/effect/gas_overlay)
...
+/decl/material/proc/get_presentation_desc(var/obj/item/prop)
+	. = glass_desc
+	if(prop?.reagents?.total_volume)
+		. = build_presentation_desc_from_reagents(prop, .)
+
+/decl/material/proc/build_presentation_desc_from_reagents(var/obj/item/prop, var/supplied)
+	. = supplied
+
+	if(cocktail_ingredient)
+		for(var/decl/cocktail/cocktail in SSmaterials.get_cocktails_by_primary_ingredient(type))
```
MistakeNot4892, line 699: Possibly minimal benefit, but you could slap `as anything` in here since this will be called repeatedly and may be iterating a large list.

code/modules/reagents/cocktails.dm
```diff
@@ -18,6 +18,12 @@
 	var/order_specific = FALSE
 	/// If TRUE, doesn't generate a codex entry.
 	var/hidden_from_codex
+	/// The icon to use for the cocktail. May be null, in which case no custom icon is used.
+	var/icon/glass_icon
```
MistakeNot4892, line 22: This is somewhat misleading typing unless I'm misunderstanding the purpose of this var. `'icon/foo.dmi'` is not actually typed as `/icon`.

code/modules/reagents/reagent_containers/drinkingglass/drinkingglass.dm
```diff
@@ -127,9 +128,17 @@ var/global/const/DRINK_ICON_NOISY = "noise"
 /obj/item/chems/drinks/glass2/on_update_icon()
 	. = ..()
 	underlays.Cut()
+	icon_state = base_icon
 
 	if (LAZYLEN(reagents?.reagent_volumes) > 0)
 		var/decl/material/R = reagents.get_primary_reagent_decl()
```
MistakeNot4892, line 134: Nitpick here since it's a pre-existing assumption but we might need to think about how this relies on the assumption that you can't have a glass where the primary ingredient isn't part of the cocktail. Thinking about one of the multi-layered ones you've added recently - if we assume 30% wiggle room, and they fill it with 30% whatever random reagent, that might leave you with like 15-20% for each of the other ingredients.

code/modules/materials/_materials.dm
MistakeNot4892: Just realized this is filtering by type so may not be worth caring about, disregard.

code/modules/materials/_materials.dm
```diff
@@ -687,6 +687,19 @@ INITIALIZE_IMMEDIATE(/obj/effect/gas_overlay)
 	if(prop.reagents.has_reagent(/decl/material/solid/ice))
 		. = "iced [.]"
 
+/decl/material/proc/get_presentation_desc(var/obj/item/prop)
```
MistakeNot4892, line 690: Curious why this is on the material rather than the reagents holder.

code/modules/materials/_materials.dm
```diff
@@ -687,6 +687,19 @@ INITIALIZE_IMMEDIATE(/obj/effect/gas_overlay)
...
+	. = glass_desc
+	if(prop?.reagents?.total_volume)
+		. = build_presentation_desc_from_reagents(prop, .)
+
+/decl/material/proc/build_presentation_desc_from_reagents(var/obj/item/prop, var/supplied)
+	. = supplied
```
MistakeNot4892, line 696: As above.

MistakeNot4892 (approved):
Code generally looks good, my comments are mainly questions/curiosity rather than blocking change requests.

PR #2652, opened 2022-08-29
Matter lists now use references instead of types as the index.

MistakeNot4892:
## Description of changes
- [X] Matter lists use material instances as the assoc key instead of types.
- [x] Gas lists use material instances as the assoc key instead of types.
- [x] Reagent lists use material instances as the assoc key instead of types.
- [ ] Chemistry reactions use material instances as the assoc key instead of types.
- [ ] Recipes etc etc etc
- [ ] Update `/decl/material/liquid/random` stuff
- [ ] Debug gas extractor
- [ ] Debug fuel compressor
- [ ] Debug microwave

Current code relies on material paths as unique identifiers mapped to immutable singletons. Down the track it would be good to allow for runtime creation of new materials, or materials loaded from secrets, but the current code does not support that at all as you can't create types post-compile. Using materials as the keys allows for easier translation of /decl/material back to /datum/material in the future, and also moves the overhead for material lookup onto `Initialize()`, which should speed up hot code like fluids and chemistry due to removing the need for `GET_DECL(X)`.

## Authorship
Myself.

## Changelog
Nothing player-facing.


MistakeNot4892:
Will revisit these down the track.

PR #2653, opened 2022-08-29
No nameless material decls pls.

MistakeNot4892:
Fixing a unit test failure on dev.


MistakeNot4892:
Wanted to fix this on staging but the code responsible isn't _on_ staging, whoops.

comma (approved):

PR #2654, opened 2022-08-30
Overdosing is now cumulative across reagent holders.

MistakeNot4892:
## Description of changes
- Unique metabolism holders will only call `metabolize()` once each during life ticks.
- Overdosing values are checked against total reagents across all holders, not per holder.

Thanks to @yawet330 for pointing out the issue with the inhaled reagent holder.

Simplifies some reagent processing logic and removes the balance problems with splitting dosage between holders.

## Authorship
Myself.

## Changelog
:cl:
tweak: Overdosing is now calculated across all reagent holders (blood and ingested) rather than per holder.
/:cl:

### DIFF
--- a/code/modules/materials/_materials.dm
+++ b/code/modules/materials/_materials.dm
@@ -513,24 +513,28 @@ INITIALIZE_IMMEDIATE(/obj/effect/gas_overlay)
 			T.assume_gas(vapor, (volume * vapor_products[vapor]), temperature)
 		holder.remove_reagent(type, volume)
 
-/decl/material/proc/on_mob_life(var/mob/living/M, var/metabolism_class, var/datum/reagents/holder) // Currently, on_mob_life is called on carbons. Any interaction with non-carbon mobs (lube) will need to be done in touch_mob.
+/decl/material/proc/on_mob_life(var/mob/living/M, var/metabolism_class, var/datum/reagents/holder, var/list/life_dose_tracker)
+
 	if(QDELETED(src))
 		return // Something else removed us.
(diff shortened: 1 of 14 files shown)


MistakeNot4892:
Note for the morning, maybe overdose value should be scaled based on where the reagent is metabolizing a la inhaled reagents in Noelle's PR being 75% as effective as injected.

Vizoplez:
> Note for the morning, maybe overdose value should be scaled based on where the reagent is metabolizing a la inhaled reagents in Noelle's PR being 75% as effective as injected.

This would add downsides to using inhalers, and smoke overall. It's more convenient but you sacrifice some efficiency, I suppose.

It'd also nerf the popular method of vecromium bromide smoke which existed on main bay, and would pose a major threat to combat here too with /similar/ chems, and this would go a long way to fixing it.

I think inhalation should definitely scale based on what you're wearing though, coifs/bandanas/etc should massively reduce its effectiveness when covering the mouth- though that's a discussion for another time.

As for OD's scaling from where it was injected, absolutely.

Inhaling reagents should make them metabolise quickly to blood and have a higher risk of overdosing compared to just drinking something, with injection having the lowest (base) OD threshold, as its directly in your blood.

E.G real life example:

Vaporising and inhaling alcohol.
PROS: You get Drunk.
CONS: You can die due to acute alcohol poisoning with no way to purge it from your bloodstream.

While tragic, it's a good example of how dangerous vaporising chemicals can be, as it can absorb into the blood, far far easier.

afterthought2 (approved):

PR #2655, opened 2022-09-01
Fix ChangeTurf overwriting ambient lights improperly

out-of-phaze:
## Description of changes
Splits the code that calls `setup_environmental_lighting()` up into two separate locations:
- /turf/exterior/Initialize() when `mapload` is `TRUE` (compiled or loaded maps)
- `/obj/effect/overmap/visitable/sector/exoplanet/proc/generate_daycycle()` for exoplanet generation

This prevents improper overwrites of ambient lights due to ChangeTurf calls in map loading, which caused desyncs of turf and corner lighting values and ugly lighting artifacts.

Fixes lighting artifacts and improper `setup_environmental_lighting()` calls.

## Authorship
Me.

### DIFF
--- a/code/game/turfs/exterior/_exterior.dm
+++ b/code/game/turfs/exterior/_exterior.dm
@@ -36,7 +36,8 @@
 			ChangeArea(src, owner.planetary_area)
 
 	. = ..(mapload)	// second param is our own, don't pass to children
-	setup_environmental_lighting()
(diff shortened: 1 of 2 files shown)


MistakeNot4892 (approved):

MistakeNot4892:
Good stuff.

PR #2658, opened 2022-09-03
Seed optimization pass.

MistakeNot4892:
## Description of changes

This should probably go to dev. Willing to be told to point it there. If nobody objects too loudly I'd like to see if this helps Scav with their performance issues.

- Reorganizes flora.dm into its own subfolder.
- Splits seed appearance generation out of hydrotrays onto the seed datum.
- Adds a flora subtype for exoplanets that replicates some hydroponics tray behaviors.

Halves the number of atoms created by exoplanet flora, and removes them from the processing list.

## Authorship
Myself.

## Changelog
:cl:
tweak: Exoplanet flora have been simplified and may behave differently.
/:cl:


code/game/objects/structures/flora/plant.dm
```diff
@@ -0,0 +1,99 @@
...
+	// ask our environment for light
+	// update health
+	// advance age
+	// update icon/harvestability as appropriate
+*/
+
+/obj/structure/flora/plant/examine(mob/user, distance)
+	. = ..()
+	if(dead)
+	else if(length(harvestable))
```
afterthought2, line 28: Weird stylistic choice? Missing block?

code/game/objects/structures/flora/plant.dm
```diff
@@ -0,0 +1,99 @@
...
+		to_chat(user, SPAN_NOTICE("You can see [length(harvestable)] harvestable fruit\s."))
+
+/obj/structure/flora/plant/Initialize(ml, _mat, _reinf_mat, datum/seed/_plant)
+	if(!plant && _plant)
+		plant = _plant
+	if(!plant)
+		return INITIALIZE_HINT_QDEL
+	name = plant.display_name
```
afterthought2, line 36: Is the theory that subtypes might populate `plant` before calling `..()`? Otherwise hard to see how it can be an instance here except through the 4th arg method, in which case the block above wouldn't need the `if(!plant...` condition.

code/game/objects/structures/flora/plant.dm
MistakeNot4892: Missing block, not sure how that happened.

code/game/objects/structures/flora/plant.dm
MistakeNot4892: The idea is that down the track mappable subtypes might be desired.

comma (approved):

PR #2669, opened 2022-09-08
Automatically set wall offset for wall mounted objects

PsyCommando:
## Description of changes
Added a json list for specifying an offset for each directions for any subclasses of ``/obj``.
The offsets in the list are overridden by mapper set values.

No more need to write a bunch of custom handling for wall offsets. Anything built in game is automatically correctly offset. Mappers don't need to set the pixel offsets in every single wall mounted obj now.

## Changelog
:cl:
code: Objects can now specify an offset for each direction they may be facing, via a json encoded list. The offset is automatically applied to the object and updated. Unless the object has any mapper set offset.
/:cl:

### DIFF
--- a/code/game/atoms.dm
+++ b/code/game/atoms.dm
@@ -22,6 +22,7 @@
 	var/tmp/default_pixel_x
 	var/tmp/default_pixel_y
 	var/tmp/default_pixel_z
+	var/tmp/default_pixel_w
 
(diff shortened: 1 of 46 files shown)


comma (approved):

PsyCommando:
One thing to note here btw. The offset and directions mapped onto thing on tradeship are really weird and some were never updated to match some of the newer icons directions. 
And I'm working on a PR to fix maps.

PR #2670, opened 2022-09-09
Fix air alarm connectivity test with desperado

PsyCommando:
## Description of changes
This modifies the air alarm connectivity test to check if the vent or scrubber being tested is actually allowed to be controlled by air alarms.
Seems to work locally.

For reference, uncontrolled vents would get listed by the air alarms. But then when the alarm would handle received message it would have its area uid set to area "NONE" which would get the messages ignored and cause the alarm to not fill the vent info list, which would fail the test.
```
/obj/machinery/atmospherics/unary/vent_pump/area_uid()
	return controlled ? ..() : "NONE"
```

The area uid is sent as a public var to the air alarm via radio signal, and ends up in this check, where it compares the area name with the one of the air alarm. That check fails so it never writes the message's data to the info list in the air alarm.
```
/obj/machinery/alarm/receive_signal(datum/signal/signal)
	if(!signal || signal.encryption)
		return
	if(alarm_id == signal.data["alarm_id"] && signal.data["command"] == "shutdown")
		mode = AALARM_MODE_OFF
		report_danger_level = FALSE
		apply_mode()
		return

	var/id_tag = signal.data["tag"]
	if (!id_tag)
		return
	if (signal.data["area"] != area_uid)
		return
```

## Changelog
:cl:
code: Made the air alarm connectivity unit test check if the vents and scrubbers it tests are allowed to be controlled via air alarms.
/:cl:

### DIFF
--- a/code/modules/atmospherics/components/unary/vent_pump.dm
+++ b/code/modules/atmospherics/components/unary/vent_pump.dm
@@ -112,7 +112,7 @@
 	. = ..()
 
 /obj/machinery/atmospherics/unary/vent_pump/reset_area(area/old_area, area/new_area)
-	if(controlled)
+	if(!controlled)
 		return
 	if(old_area == new_area)
 		return
@@ -232,7 +232,7 @@
 	return controlled ? ..() : "NONE"
 
 /obj/machinery/atmospherics/unary/vent_pump/proc/change_area_name(var/area/A, var/old_area_name, var/new_area_name)
-	if(controlled)
+	if(!controlled)
 		return
 	if(get_area(src) != A)
 		return
--- a/code/modules/atmospherics/components/unary/vent_pump.dm
+++ b/code/modules/atmospherics/components/unary/vent_pump.dm
@@ -112,6 +112,8 @@
 	. = ..()
 
 /obj/machinery/atmospherics/unary/vent_pump/reset_area(area/old_area, area/new_area)
+	if(controlled)
+		return
 	if(old_area == new_area)
 		return
 	if(old_area)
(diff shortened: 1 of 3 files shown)


MistakeNot4892:
Needs to go to `staging` if possible.
If this was the root cause wouldn't the issue be reliable? As it stands the Desperado failure only happens occasionally, but antag maps are loaded every run.

afterthought2:
If non-controlled air alarms are registering their ids on the area list, that in itself is a bug.

PsyCommando:
> If non-controlled air alarms are registering their ids on the area list, that in itself is a bug.

Yeah, I had been thinking about that actually. I'll try to get them not to register with the air alarm to begin with.

PsyCommando:
> Needs to go to `staging` if possible. If this was the root cause wouldn't the issue be reliable? As it stands the Desperado failure only happens occasionally, but antag maps are loaded every run.

Yeah, that was weird to me as well. But, its possible there's some sort of race condition in there. All the 3 vents causing issues are all uncontrolled vents, 2 exposed to space, and one in a air tank. So there's definitely a link there.
In any case uncontrolled air vents shouldn't end up in the unit test to begin with.

PsyCommando:
Also, just wondering, does that bug even happens on staging? 
@MistakeNot4892

PsyCommando:
Alright, so I think I got all bases covered now. 
Uncontrolled vents can't accidentally register with the area anymore. And the unit test waits on SSmachinery and SSalarm too now.

I don't know if I really should target staging since it mainly is a problem on dev rn?

MistakeNot4892:
@afterthought2 would appreciate your views on this one.

PsyCommando:
Well crap. I tried to rebase to staging, but then I noticed that air vent code changed compared to staging and I'd have to re-apply the changes in the PR twice.

MistakeNot4892 (approved):
@afterthought2

MistakeNot4892:
Going ahead and merging this since Afterthought is unavailable.

PR #2674, opened 2022-09-13
Fix telescreen installation and offset + refactor into a console

PsyCommando:
## Description of changes
* Telescreen now properly installs on walls from the direction the user is facing.
* Added basic directional icon states for the telescreen and its overlays.
* ~~Fix telescreen being usable when held.~~ (Not relevant anymore)
* ~~Made a mapped preset so the mapped telescreens start anchored.~~  (Not relevant **anymore)
* ~~Prevented objects from automatically being offset if not anchored.~~
* Modified all map references to the telescreen to use the auto-offset.
* Made the telescreen into a subclass of the ``/obj/machinery/computer/modular`` class.
* Added a wall mountable frame item for the telescreen, along with a kit.
* Added a circuitboard for the telescreen.

## Changelog
:cl:
bugfix: Telescreen can no longer be used when held, as it was intended.
bugfix: Telescreen now properly install on walls.
imageadd: Added basic directional icon states for the telescreen.
imageadd: Added construction state icons for the telescreen.
added: Added a wall frame for the telescreen along with a kit.
added: Added a circuitboard for the telescreen.
code: Refactored the telescreen so that its a subclass of /obj/machinery/computer/modular now.
tweak: Changed all instances of telescreens in the map files to use auto directional offset, and face proper direction.
/:cl:


code/game/objects/objs.dm
```diff
@@ -208,22 +208,22 @@
...
 /obj/set_dir(ndir)
 	. = ..()
-	if(directional_offset)
+	if(directional_offset && anchored)
 		update_directional_offset()
 
 /obj/Move()
 	. = ..()
-	if(directional_offset)
+	if(directional_offset && anchored)
```
MistakeNot4892: Why would something anchored be moved?
Generally this looks like it would be better to have `update_direction_offset()` null offsets if not anchored, rather than doing `n` separate checks.

PsyCommando: Yeah, that's a brainfart on my part. I'll remove that line in Move().

And, if I null the offsets when not anchored, it means anything else that might set the default offset would get overwritten when unanchored. So I'd rather take a more prudent approach to it and let whatever unanchor the thing handle its new default offset. That's especially tricky for items, since they tend to mess with pixel offsets a bunch.

code/game/objects/objs.dm
```diff
@@ -208,22 +208,17 @@
 
 /obj/set_dir(ndir)
 	. = ..()
-	if(directional_offset)
-		update_directional_offset()
-
-/obj/Move()
-	. = ..()
-	if(directional_offset)
+	if(directional_offset && anchored)
```
afterthought2: I'm with Loaf on this: what actually moves/forceMoves anchored entities, except e.g. shuttle move and qdel?

code/game/objects/objs.dm
```diff
@@ -208,22 +208,17 @@
 
 /obj/set_dir(ndir)
 	. = ..()
-	if(directional_offset)
-		update_directional_offset()
```
afterthought2: Isn't the whole point to update when the dir is changed?

code/game/objects/objs.dm
PsyCommando: Got rid of the stuff in move a while ago. Also I don't think you're commenting on the right line?

code/game/objects/objs.dm
PsyCommando: The thing is, some items act differently when they're not anchored. 
So they only get offset when they actually are anchored in place, since that's the common case when all things that uses this system actually always want to be offset.

code/game/objects/objs.dm
afterthought2: Well, it's reasonable to point this comment at any of the three. I just don't buy that this event is in the right place (anchored + forceMove/dir set). Things being moved or rotated while anchored are inherently ad hoc/outside the world model and will vary obj by obj, so it's not clear what you are intercepting here. I get that it probably works for this obj in particular, but this is an /obj-level forceMove override, it needs to be well thought out. I think Loaf's suggestion to have some ability to activate/deactivate offsets makes a lot more sense.

PsyCommando: The comment on the check being in move() made sense, since an anchored object never moves anyways, so I fixed that. ForceMove on the other hand doesn't work the same way, and will move the thing whether its anchored or not, that's how things are moved into the inventory for one. 
Also, you already can activate/deactivate the offset? Just null out the directional_offset var?

Then, the only objects using this are wall-mounted objects and items. I could check all of them if they want to have a directional offset, by calling a proc in the if condition, but that's getting a bit overkill when the only current, and unlikely to change, use case is wall-mounted objects.. I just don't see how that would be worth it, or not annoy people because of an extra unavoidable proc call inside a busy code path?

code/game/objects/objs.dm
```diff
@@ -208,22 +208,17 @@
...
-/obj/Move()
-	. = ..()
-	if(directional_offset)
+	if(directional_offset && anchored)
 		update_directional_offset()
 
 /obj/forceMove(atom/dest)
 	. = ..()
-	if(directional_offset)
+	if(directional_offset && anchored)
```
MistakeNot4892: `directional_offset` and `anchored` are both checked in `update_directional_offset()`, why check them at all here and in `set_dir`

PsyCommando: To prevent the proc call overhead for literally every obj ever.

MistakeNot4892: `directional_offset` is only set on 64 types, largely structures and machines, and anchored objects shouldn't be moving frequently.

PsyCommando: Oh so you meant only the ``anchored`` check? Yeah, that one might be too much here.

code/game/objects/objs.dm
afterthought2: If you can activate/deactivate it, then why not do it here as appropriate? That would be natural, and there are anchored setters that you can put the activation into.

On the one hand, you are arguing that `anchored` is the right condition for these wall objs and they are the only things using the system, so it's the right condition overall. On the other hand, you are writing the system at `/movable` level. I just don't think you can have it both ways.

PsyCommando: Changed the code since then.

code/game/objects/objs.dm
```diff
@@ -211,19 +211,14 @@
 	if(directional_offset)
 		update_directional_offset()
 
-/obj/Move()
```
MistakeNot4892: Can you clarify why this is being removed? I looked over the PR and I can't see an equivalent of it being readded.

code/game/objects/objs.dm
MistakeNot4892: Currently the code seems to be unchanged from base so assuming this is resolved.

code/modules/modular_computers/computers/subtypes/dev_telescreen.dm
```diff
@@ -1,61 +1,65 @@
...
+		set_dir(global.reverse_dir[user.dir])
+		set_anchored(TRUE)
+		visible_message(SPAN_NOTICE("\The [user] secures \the [src]."), SPAN_NOTICE("You secure \the [src]."), range = 3)
+		playsound(src, 'sound/effects/metalhit.ogg', 25, TRUE)
+
+/obj/item/modular_computer/telescreen/proc/set_anchored(var/state)
+	var/datum/extension/assembly/modular_computer/telescreen/assembly = get_extension(src, /datum/extension/assembly)
+	anchored = state
+	if(anchored)
+		obj_flags |= OBJ_FLAG_MOVES_UNSUPPORTED
```
MistakeNot4892: Why does this flag need to be toggled? As far as I can tell it's solely used for shuttle moves.

code/modules/modular_computers/computers/subtypes/dev_telescreen.dm
```diff
@@ -1,61 +1,65 @@
...
+	desc               = "A wall-mounted touchscreen computer."
+	icon               = 'icons/obj/modular_computers/modular_telescreen.dmi'
+	icon_state         = "telescreen"
+	anchored           = FALSE
+	density            = FALSE
+	light_strength     = 4
+	w_class            = ITEM_SIZE_HUGE
+	computer_type      = /datum/extension/assembly/modular_computer/telescreen
+	obj_flags          = 0
+	directional_offset = "{'NORTH':{'y':-20}, 'SOUTH':{'y':24}, 'EAST':{'x':-24}, 'WEST':{'x':24}}"
```
MistakeNot4892, line 59: These don't line up with the offsets in the code below, is that intentional?

code/game/objects/objs.dm
```diff
@@ -211,19 +211,14 @@
...
-
 /obj/forceMove(atom/dest)
 	. = ..()
 	if(directional_offset)
 		update_directional_offset()
 
 /** Applies the offset stored in the directional_offset json list depending on the current direction.  */
 /obj/proc/update_directional_offset()
-	if(!length(directional_offset))
+	if(!length(directional_offset) || !anchored)
```
MistakeNot4892: I don't think it makes sense to check anchored here. It's still not clear to me why you need to universally stop atoms from updating their offsets if they're not anchored. You could override update_directional_offset() on the telescreen to return early if not anchored, or make it an object flag.

code/modules/modular_computers/computers/subtypes/dev_telescreen.dm
PsyCommando: If it gets wall-mounted, it has to act like it is. So if someone bolts a telescreen on the outside of a shuttle, it should move along with the shuttle even if its on an open turf.

code/game/objects/objs.dm
PsyCommando: Because afterthought really wanted me to change that.
https://github.com/NebulaSS13/Nebula/pull/2674#discussion_r980650767

code/modules/modular_computers/computers/subtypes/dev_telescreen.dm
MistakeNot4892: Why does it need to be _toggled_? It was previously just set on the atom.

code/modules/modular_computers/computers/subtypes/dev_telescreen.dm
PsyCommando: Uh, what code below? I looked at the whole file but I can't find another place changing offsets?
Those on line 12 are the proper offsets I got after tweaking them live with the current icon.

MistakeNot4892: The ones you deleted.

code/game/objects/objs.dm
PsyCommando: I don't really know anymore tbh. I remember having a talk with you and afterthought and it didn't make much sense to intercept move since the objects updating their offsets were all anchored. 
And there too : https://github.com/NebulaSS13/Nebula/pull/2674#discussion_r980650767

code/modules/modular_computers/computers/subtypes/dev_telescreen.dm
PsyCommando: The ones I deleted were completely off?

code/modules/modular_computers/computers/subtypes/dev_telescreen.dm
PsyCommando: Because once you detach the screen from the wall it behaves like any other items and should fall off the shuttle if it moves?

code/game/objects/objs.dm
PsyCommando: I guess I'll work out something else then.

PsyCommando:
Alright, so since everyone seemed annoyed at having special cases for handling items that behave like machines, I just turned the telescreen into a machine, so that we don't have to debate about that anymore or make exception for items that act weird.

code/game/objects/objs.dm
PsyCommando: I removed the anchored checks, and implemented the checks for being on the wall or not into the concerned objects.

code/modules/modular_computers/computers/subtypes/dev_telescreen.dm
PsyCommando: Now it turns into a frame item when in deconstructed state. So no need for special handling anymore.

code/game/objects/objs.dm
PsyCommando: Restored that code to what it was before.

code/modules/modular_computers/os/visuals.dm
```diff
@@ -20,7 +20,10 @@
 	if(screen_icon_file)
 		var/image/I
 		if(active_program)
-			I = image(screen_icon_file, active_program.program_icon_state)
+			if(active_program.program_icon_state in icon_states(screen_icon_file))
```
MistakeNot4892, line 23: `check_state_in_icon(active_program.program_icon_state, screen_icon_file)`

MistakeNot4892 (approved):

PR #2683, opened 2022-09-17
Remade wrapping paper and parcels/gifts

PsyCommando:
## Description of changes
They basically both were very similar in the way they worked, so I made them inherit from the same base class. 
I added some  fairly basic recolors for some of the missing gift item sizes. And an icon for wrapping humans in packaging paper as well. 

I've also changed the way stacks init their matter list to allow the list to contain matter that's not used when using a stack "sheet". Mainly used by the cardboard roll in the package wrappers, but its fairly generalized and could be used by other stacks. 

The icon handling on parcels and gifts is a bit rough around the edges frankly speaking, but I don't know if it would be worth it going all in and making a whole custom system for what is essentially a one time use object that will only update its icon maybe once or twice at most.

## Changelog
:cl:
imageadd: Added some fairly bad wrapped package icons, and wrapped gifts icons.
refactor: Refactored package wrapper and gift wrapper.
/:cl:


PsyCommando:
conflict fixed @MistakeNot4892

code/datums/supplypacks/nonessent.dm
```diff
@@ -24,7 +23,8 @@
 		/obj/item/chems/glass/paint/black,
 		/obj/item/chems/glass/paint/white,
 		/obj/item/contraband/poster,
-		/obj/item/wrapping_paper = 3
+		/obj/item/stack/package_wrap/fifty      = 2,
+		/obj/item/stack/package_wrap/gift/fifty = 2
 	)
 	containername = "arts and Crafts crate"
```
out-of-phaze: While you're in the area, mind fixing this?
```suggestion
	containername = "arts and crafts crate"
```

code/modules/recycling/package_wrapper.dm
```diff
@@ -0,0 +1,174 @@
...
+	max_amount       = 50
+	material         = /decl/material/solid/paper
+	matter           = list(/decl/material/solid/cardboard = MATTER_AMOUNT_PRIMARY * HOLLOW_OBJECT_MATTER_MULTIPLIER) //For the tube
+	matter_per_piece = list(/decl/material/solid/paper = MATTER_AMOUNT_PRIMARY)
+	force            = 1
+	throwforce       = 1
+	throw_range      = 5
+	throw_speed      = 3
+	///Types that the wrapper cannot wrap, ever
+	var/static/list/wrap_blacklist = list(
```
out-of-phaze: I would really suggest making this a proc so that it can be modified by subtypes. Static vars defined on types can't be overridden on subtypes, while if you did something like
```
/obj/item/stack/package_wrap/proc/get_wrap_blacklist()
	var/static/list/wrap_blacklist = list(
		/obj/item/parcel,
		/obj/item/parcel/gift,
		/obj/item/evidencebag,
		/obj/item/stack/package_wrap,
	)
	return wrap_blacklist
```
you could then do
```
/obj/item/stack/package_wrap/recursive/get_wrap_blacklist()
	var/static/list/wrap_blacklist = (..() - list(/obj/item/parcel, /obj/item/parcel/gift, /obj/item/stack/package_wrap))
	return wrap_blacklist
```
and modify it on a subtype. This also late-initialises the list, and ensures it's only modified/created once. (Make sure nothing mutates it, though!!)

code/modules/recycling/sort_tag.dm
```diff
@@ -0,0 +1,26 @@
...
+/datum/extension/sorting_tag
+	base_type     = /datum/extension/sorting_tag
+	expected_type = /atom/movable //Both mob and objects can have it
+	var/destination
+	var/tag_icon_state
+	var/tag_x = 0
+	var/tag_y = 0
+
+/datum/extension/sorting_tag/New(datum/holder, var/_destination, var/_icon_state, var/_tag_x = 0, var/_tag_y = 0)
+	. = ..(holder)
```
out-of-phaze: New() has no return value. Don't confuse it with Initialize().
```suggestion
	..(holder)
```

out-of-phaze (dismissed):
`/obj/item/stack/package_wrap::wrap_blacklist` and `/obj/item/parcel::type_whitelist` seem redundant?

That said, I don't seen any blocking issues, just some suggestions and also that one apparent redundancy.

code/modules/recycling/package_wrapper.dm
PsyCommando: Yeah, that's a good idea. I'll try that.

code/modules/recycling/sort_tag.dm
PsyCommando: Yeah, that's just vscode automatically adding .=..() when autocompleting overriding a proc and me forgetting about it XD.

out-of-phaze (dismissed):
Thank you for addressing those comments, seems good!

afterthought2:
I suggest splitting the stacks changes out; I will not be reviewing the stacks changes, which is why this has sat without review for so long.

PsyCommando:
> I suggest splitting the stacks changes out; I will not be reviewing the stacks changes, which is why this has sat without review for so long.

I mean, its literally just a bug fix, and a callback when destroying the empty stack? What's with the attitude?

PsyCommando:
> I suggest splitting the stacks changes out; I will not be reviewing the stacks changes, which is why this has sat without review for so long.

https://github.com/NebulaSS13/Nebula/pull/2802

Can't remove any of it from this PR since everything depends on that. But there you go.

afterthought2:
No attitude, I just want to get this PR reviewed/merged while having the stack changes reviewed by the person who understands the relevant code (not me).

PsyCommando:
Alright, so after a lot of debating yesterday, I removed most of the stack changes. Only kept the on_last_used() callback as discussed. So there's no dependencies anymore. And everything stack related is mostly confined to the wrapping paper stack.

code/modules/recycling/package_wrapper.dm
```diff
@@ -102,23 +102,27 @@
...
-/obj/item/stack/package_wrap/proc/on_used_last(var/oldloc)
+/obj/item/stack/package_wrap/on_used_last()
 	//Drop the cardboard tube after we're emptied out
-	var/obj/item/c_tube/tube = new(get_turf(oldloc))
-	if(ismob(oldloc))
-		var/mob/M = oldloc
+	var/obj/item/c_tube/tube = new(get_turf(loc))
+	if(ismob(loc))
+		var/mob/M = loc
+		M.unEquip(src, null, FALSE)
```
MistakeNot4892: This will be handled by qdel afaik

PsyCommando: Yeah, its just that I need to empty the hand before I can replace the stack with the cardboard tube. The stack isn't deleted until the base proc gets called.

MistakeNot4892: Would be better to cache the holder and call parent rather than unequipping twice imo.

MistakeNot4892:
````
/obj/item/stack/package_wrap/on_used_last()
    var/last_loc = loc
    ..()
    if(last_loc)
        var/obj/tube = new(get_turf(last_loc))
        if(ismob(last_loc))
            var/mob/M = last_loc
            M.put_in_hands(tube)
````

Comment (PsyCommando) [reply code/modules/recycling/package_wrapper.dm]: done

File: code/game/objects/structures/crates_lockers/closets/__closet.dm [hunk 5]
```diff
@@ -410,9 +410,10 @@ var/global/list/closets = list()
 		make_broken()
 
 	//Do this to prevent contents from being opened into nullspace
-	if(istype(loc, /obj/structure/bigDelivery))
-		var/obj/structure/bigDelivery/BD = loc
-		BD.unwrap()
+	//#TODO: There's probably a better way to do this?
```
afterthought2, line 417: While maybe there are technical improvements to be made here (e.g. move contained mobs using `dropInto(loc)` or possibly even `forceMove(get_turf(src))` to avoid screwing over client-controlled mobs under weird interactions like nested parcels), I'm not sure how you are going to get around unwrapping the parcel cleanly.

code/modules/mob/living/living.dm
```diff
@@ -301,33 +301,13 @@ default behaviour is:
...
-			L += D.wrapped
-			if(istype(D.wrapped, /obj/item/storage)) //this should never happen
-				L += get_contents(D.wrapped)
 		return L
 
 	else
 
 		L += src.contents
 		for(var/obj/item/storage/S in src.contents)	//Check for storage items
 			L += get_contents(S)
```
afterthought2, line 310: I'm kind of lost here: why is there a blocking arg at all, i.e. why not
```
/mob/living/get_contents()
		. = contents.Copy()
		for(var/obj/item/storage/S in src)	//Check for storage items
				. += S.return_inv()
```

code/modules/mob/living/silicon/robot/drone/drone_abilities.dm
```diff
@@ -4,17 +4,27 @@
...
+	else if(choice == "List")
+		new_tag = input("Select the desired destination.", "Set Mail Tag", null) as null|anything in global.tagger_locations
+	else
 		return
 
-	to_chat(src, "<span class='notice'>You configure your internal beacon, tagging yourself for delivery to '[new_tag]'.</span>")
-	mail_destination = new_tag
+	if(!new_tag)
+		remove_extension(src, /datum/extension/sorting_tag)
+		return
```
afterthought2, line 21: Not actually familiar with this UI, but is this intuitive behavior? Doesn't this mean that canceling out of the second prompt will remove an existing tag?

code/modules/recycling/destination_tagger.dm
```diff
@@ -0,0 +1,78 @@
...
+	var/list/last_used_tags
+
+/obj/item/destTagger/proc/clear_previous_tags()
+	LAZYCLEARLIST(last_used_tags)
+
+///Add a previous destionation tag to the list and perform rotation of the tag history if neccessary.
+/obj/item/destTagger/proc/add_previous_tag(var/text)
+	if(LAZYLEN(last_used_tags) >= MAX_DEST_TAGGER_PREVIOUS_TAGS)
+		last_used_tags.Cut(1, MAX_DEST_TAGGER_PREVIOUS_TAGS - 1)
+		last_used_tags.Insert(1, text)
```
afterthought2: Is this really doing what you want? Possibly I'm misreading, but I believe this cuts the first 23 (!) entries of the list, and then inserts the new thing in the first entry, ahead of the other 2. On the other hand, the other logic below just inserts the new thing at the end.

code/modules/recycling/destination_tagger.dm
```diff
@@ -0,0 +1,78 @@
...
+	var/cnt = 1
+	for(var/prevdest in last_used_tags)
+		if(cnt % 4 == 0)
+			dat += "</tr><tr>"
+		dat += "<td><a href='?src=\ref[src];set_tag=[prevdest]'>[prevdest]</a></td>"
+		++cnt
+	dat += "</tr></table></tt>"
+
+	show_browser(user, dat, "window=destTagScreen;size=450x375")
+	onclose(user, "destTagScreen")
```
afterthought2: Why do you need an `onclose` here? You don't seem to do anything on close.

code/modules/recycling/package_wrapper.dm
```diff
@@ -0,0 +1,181 @@
+///Returns the amount of sheets of wrapping paper that the given object would need to be wrapped.
+/proc/wrapping_paper_needed_for_obj(var/obj/O)
+	if(istype(O))
+		return min(O.w_class, 8) //At most, since 20 sheets is a bit much for structures
+
+	else if(ismob(O))
```
afterthought2: You know what I'm going to say about this, right? Make it a movable proc.

code/modules/recycling/package_wrapper.dm
```diff
@@ -0,0 +1,181 @@
...
+	if(currently_wrapping && user)
+		return
+	var/paper_to_use = wrapping_paper_needed_for_obj(AM)
+	if(!can_use(paper_to_use))
+		if(user)
+			to_chat(user, SPAN_WARNING("There isn't enough [plural_name] in \the [src] to wrap \the [AM]!"))
+		return
+	currently_wrapping = TRUE
+	if(!user?.do_skilled(clamp(paper_to_use, 2, 6) SECONDS, SKILL_HAULING, AM))
+		currently_wrapping = FALSE
```
afterthought2, line 78: Slightly confused by the logic chain. We are carefully engineering this to work fine with no user...only to do nothing and exit immediately right here? Would be easier to understand if there's just `if(!user) return` (maybe setting `currently_wrapping = FALSE` if that's desirable) on the first line.

code/modules/recycling/package_wrapper.dm
```diff
@@ -0,0 +1,181 @@
...
+		return
+	. = create_wrapper(AM, user)
+	if(.)
+		use(paper_to_use)
+	else if(user)
+		to_chat(user, SPAN_WARNING("You cannot wrap \the [AM]!"))
```
afterthought2, line 84: `user` is true here, and in any case `to_chat(null, ...)` is safe and does nothing.

code/modules/recycling/package_wrapper.dm
```diff
@@ -0,0 +1,181 @@
...
+			to_chat(user, SPAN_WARNING("\The [target] moving around too much. Restrain or incapacitate them first."))
+
+/obj/item/stack/package_wrap/afterattack(var/obj/target, mob/user, proximity_flag, click_parameters)
+	if(!proximity_flag || !can_wrap(target))
+		return
+	return wrap(target, user) || ..()
+
+/obj/item/stack/package_wrap/attack(mob/living/M, mob/living/user, target_zone, animate)
+	if(user.a_intent != I_HELP)
+		return ..()
```
afterthought2: Is it transparent that this is a "help" interaction? I would have thought it should be made on any intent, or possibly only on _non_-help intent.

code/modules/recycling/package_wrapper.dm
```diff
@@ -0,0 +1,181 @@
...
+		matter[mat] = (matter_per_piece[mat] * amount)
+	matter[/decl/material/solid/cardboard] = cardboard_amount
+
+///Types that the wrapper cannot wrap, ever
+/obj/item/stack/package_wrap/proc/get_blacklist()
+	var/static/list/wrap_blacklist = list(
+		/obj/item/parcel,
+		/obj/item/parcel/gift,
+		/obj/item/evidencebag,
+		/obj/item/stack/package_wrap,
```
afterthought2, line 144: Is this just for deconflicting the stack merge interaction? Otherwise would seem fine.

code/modules/recycling/wrapped_package.dm
```diff
@@ -0,0 +1,283 @@
...
+	item_flags        = ITEM_FLAG_HOLLOW
+	material          = /decl/material/solid/paper
+	attack_verb       = list("delivered a hit", "expedited on", "shipped at", "went postal on")
+	base_parry_chance = 40 //Boxes tend to be good at parrying
+	///A text note attached to the parcel that shows on examine
+	var/attached_note
+	///The first part of the icon_state names
+	var/tmp/icon_state_prefix = "parcel"
+	///Color of the dropped paper trash on unwrapping.
+	var/tmp/trash_color = "#b97644"
```
afterthought2, line 22: Are these meaningfully `tmp`? They seem to be normal state vars, afaik we use `tmp` for internal auxiliary vars used for buffering, optimization, etc.

code/modules/recycling/wrapped_package.dm
```diff
@@ -0,0 +1,283 @@
...
+
+	if(user)
+		add_fingerprint(user)
+		AM.add_fingerprint(user)
+
+	//Notify properly whatever we're taking the thing from
+	//And then put it in the user's hands if it makes sense
+	if(ismob(AM.loc))
+		var/mob/M = AM.loc
+		M.unEquip(AM, src)
```
afterthought2: If you are going to do this, `drop_from_inventory`, as `unEquip` can in principle fail to actually move `AM`. You could I guess alternatively check the return and abort, but presumably that should have happened earlier in the chain.

Along similar lines, check locally: can borgs wrap their own modules?

code/modules/recycling/wrapped_package.dm
```diff
@@ -0,0 +1,283 @@
...
+		if(ITEM_SIZE_SMALL)
+			parcel_name = "small [initial(name)]"
+		if(ITEM_SIZE_NORMAL)
+			parcel_name = "normal-sized [initial(name)]"
+		if(ITEM_SIZE_LARGE)
+			parcel_name = "large [initial(name)]"
+		if(ITEM_SIZE_HUGE, ITEM_SIZE_GARGANTUAN)
+			parcel_name = "huge [initial(name)]"
+		else
+			parcel_name = "enormous [initial(name)]"
```
afterthought2, line 119: Is it intentional that mobs will generally fall in this category? Mobs and objs have different "scales" for size to some degree, usually people convert in this situation somehow (e.g. with log).

code/modules/recycling/wrapped_package.dm
```diff
@@ -0,0 +1,283 @@
...
+			var/obj/item/storage/S = loc
+			S.handle_item_insertion(AM, TRUE)
+		else
+			AM.dropInto(loc)
+
+		if(ismob(AM))
+			var/mob/M = AM
+			if(M.client)
+				M.client.eye = M.client.mob
+				M.client.perspective = MOB_PERSPECTIVE
```
afterthought2: I'm 99% sure you are supposed to call `reset_view` or a related proc on the mob instead.

PlayerDeer:
image
Made you some new sprites that might go with the current style of the game better.

https://www.dropbox.com/s/sf7xmio4bb02rlz/deliverypackage.dmi?dl=0
https://www.dropbox.com/s/k5kav63imqepdn4/gift_wrapped.dmi?dl=0

PsyCommando:
> image Made you some new sprites that might go with the current style of the game better.
> 
> https://www.dropbox.com/s/sf7xmio4bb02rlz/deliverypackage.dmi?dl=0 https://www.dropbox.com/s/k5kav63imqepdn4/gift_wrapped.dmi?dl=0

Oooh, thanks! Looks much better than my cruddy placeholders XD

code/game/objects/structures/crates_lockers/closets/__closet.dm
PsyCommando: I'm really unsure what you mean here? That's strictly for wrapped closets. It's in the closet's procs?

code/modules/mob/living/living.dm
PsyCommando: Dunno, I didn't write that?

code/modules/mob/living/silicon/robot/drone/drone_abilities.dm
PsyCommando: That's to keep the behaviour the same as it was before.
Cancelling clears the tag.

code/modules/recycling/destination_tagger.dm
PsyCommando: Yeah, you may be right here. I'll fix that one.

code/modules/recycling/destination_tagger.dm
PsyCommando: Look at what onclose does in its default implementation. We already had a convo about that same issue on the paperwork PR.

code/modules/recycling/package_wrapper.dm
PsyCommando: You know this was written before we had that convo on the paperwork stuff? :P
If I'd make an atom proc I'd make a proc to get an approximation of the surface area of an atom and work from there in the package_wrapper code.

code/modules/recycling/package_wrapper.dm
PsyCommando: That's actually a bug. It should keep going without a user.

code/modules/recycling/wrapped_package.dm
MistakeNot4892: I don't think we have any consistent pattern for usage of `tmp` honestly.

code/modules/recycling/wrapped_package.dm
MistakeNot4892: Tiny and minuscule mobs will fall under 'huge', for reference. Mob and item sizes should be the same units these days.

code/modules/recycling/package_wrapper.dm
PsyCommando: Maybe, but why bother even concatenating that string and call the << operator if we know it's not necessary?

code/modules/recycling/package_wrapper.dm
PsyCommando: This is solely to allow people to insert the package_wrapper into storage items, put them on tables and inside crates. That's how its done with other things that got special interactions with things they hit as well.

code/modules/recycling/package_wrapper.dm
PsyCommando: Its probably partly for that, and avoiding to wrap a gift a million times I guess. That's mostly just preserved behaviour.

code/modules/recycling/wrapped_package.dm
PsyCommando: Tmp does nothing besides making the proc returning whether a var is saved or not return false. 
tmp is for things that shouldn't be saved or are just temporary. Like constants and the like. Since constants cannot be changed by subclass, might as well mark them tmp, since they never change from their definition and will never need to be saved.

code/modules/recycling/wrapped_package.dm
PsyCommando: That's a bug, I forgot to check the return from unEquip.

PsyCommando: I don't think borg module slots allow any interactions between eachothers that way. Its an entirely different inventory system.

MistakeNot4892: This is incorrect. Borg modules call attackby on each other just fine in inventory.

code/modules/recycling/wrapped_package.dm
MistakeNot4892: Constants should be marked const, not tmp.

code/modules/recycling/wrapped_package.dm
PsyCommando: I wouldn't call that the same units.

```dm
#define ITEM_SIZE_TINY           1
#define ITEM_SIZE_SMALL          2
#define ITEM_SIZE_NORMAL         3
#define ITEM_SIZE_LARGE          4
#define ITEM_SIZE_HUGE           5
#define ITEM_SIZE_GARGANTUAN     6
#define ITEM_SIZE_NO_CONTAINER   10 // Use this to forbid item from being placed in a container.
#define ITEM_SIZE_STRUCTURE      20
```
```dm
#define MOB_SIZE_LARGE  		40
#define MOB_SIZE_MEDIUM 		20
#define MOB_SIZE_SMALL 		10
#define MOB_SIZE_TINY 		5
#define MOB_SIZE_MINISCULE	1
```
Honestly, I'd have made small mobs the size of a normal/large sized object. 

But the problem is that the w_class is also used to prevent things from being inserted into storage as well as give its size. So when you don't want something to be inserted into storage, you also give up on any meaningful size value. Which is kinda annoying when you're trying to give a visual representation of a wrapped object. 

So I just sorta gave up because there's no good solution to this currently short of making some type checking mess.

code/modules/recycling/wrapped_package.dm
PsyCommando: In the borg inventory, or in a non-borg inventory?

code/modules/recycling/wrapped_package.dm
PsyCommando: Could you get that missing 1% so I can just replace that line? :P

code/modules/recycling/wrapped_package.dm
MistakeNot4892: >I wouldn't call that the same units.

We've had this conversation before. If you disagree with the scale of sizes, sure, but as they are currently, they are intended to be the same units. If you look at the places in code where both values are used, they are equivalent. You could make `w_class` and `mob_size` into a single var on `/atom/movable` like `atom_size` and it wouldn't impact much in the code.

>So when you don't want something to be inserted into storage, you also give up on any meaningful size value. 

Make it an atom flag.

code/modules/recycling/wrapped_package.dm
MistakeNot4892: Borg inventory. Testing locally now and package wrappers do not wrap other module items, not sure of the distinction. If you look at engineering bots, attackby is definitely passed around normally:
````
The steel rod is fused together by Engineering Robot-311 with the welding tool.
The steel rod is fused together by Engineering Robot-311 with the welding tool.
You add a new sheet to the stack. It now contains 2 sheets.
The steel rod is fused together by Engineering Robot-311 with the welding tool.
You add a new sheet to the stack. It now contains 3 sheets.
````
I'm assuming there's an is_robot_module check somewhere to prevent the wrapper working. It's worth testing on this PR in any case.

File: code/modules/recycling/wrapped_package.dm (hunk 18 above)
Comment (MistakeNot4892) [reply code/modules/recycling/wrapped_package.dm]: Go and look at the proc After mentioned. He's correct.

File: code/modules/recycling/wrapped_package.dm (hunk 17 above)
Comment (PsyCommando) [reply code/modules/recycling/wrapped_package.dm:119]: > > I wouldn't call that the same units.
> 
> We've had this conversation before. If you disagree with the scale of sizes, sure, but as they are currently, they are intended to be the same units. If you look at the places in code where both values are used, they are equivalent. You could make `w_class` and `mob_size` into a single var on `/atom/movable` like `atom_size` and it wouldn't impact much in the code.
> 
> > So when you don't want something to be inserted into storage, you also give up on any meaningful size value.
> 
> Make it an atom flag.

Yeah, I remember, its just the scale has big gaps between mob sizes. So a gargantuan sized object is actually smaller than a crab, cat or corgi, or something. Which is a bit incoherent when you have to reconcile those together visually?

And yeah, an atom flag would make sense for that. I'll make a PR for that then I guess.

File: code/modules/recycling/wrapped_package.dm (hunk 15 above)
Comment (PsyCommando) [reply code/modules/recycling/wrapped_package.dm:22]: The problem here, is that constants cannot be changed in sub-classes. 
So nobody uses the const keyword for that, unless the var is also static.

Comment (MistakeNot4892) [reply code/modules/recycling/wrapped_package.dm:22]: If you're changing the value, it's not a constant. :/

File: code/modules/recycling/wrapped_package.dm (hunk 18 above)
Comment (PsyCommando) [reply code/modules/recycling/wrapped_package.dm]: I asked nicely? I'm already working on like 20 other suggestions, so I'm trying to save some time.

Comment (afterthought2) [reply code/modules/recycling/wrapped_package.dm]: `reset_view` is a proc which exists, can be called here, and does what this does in a saner way. It also has other potential effects, and I have no idea how those interact with the mob being in a closet in a parcel in a chute or whatever is going on here, so look it over and/or test locally before using.

File: code/modules/recycling/wrapped_package.dm (hunk 17 above)
Comment (afterthought2) [reply code/modules/recycling/wrapped_package.dm:119]: I don't care about mob vs. obj sizes in general, I just want to make sure this piece of UI code is doing what you want it to be doing.

File: code/modules/recycling/wrapped_package.dm (hunk 15 above)
Comment (afterthought2) [reply code/modules/recycling/wrapped_package.dm:22]: I would just let them be normal vars, I feel that's the most descriptive and idiomatic approach. I don't care strongly if you insist on tmp, I guess.

File: code/modules/recycling/destination_tagger.dm (hunk 9 above)
Comment (afterthought2) [reply code/modules/recycling/destination_tagger.dm]: In that PR it interacted with the machine var, which afaik is not relevant here. I commented because I don't actually see what it's doing here.

File: code/game/objects/structures/crates_lockers/closets/__closet.dm (hunk 5 above)
Comment (afterthought2) [reply code/game/objects/structures/crates_lockers/closets/__closet.dm:417]: I was answering the TODO, suggesting that no, I don't know of a better way, and I doubt the future reader will either.

File: code/modules/recycling/wrapped_package.dm (hunk 15 above)
Comment (PsyCommando) [reply code/modules/recycling/wrapped_package.dm:22]: > I would just let them be normal vars, I feel that's the most descriptive and idiomatic approach. I don't care strongly if you insist on tmp, I guess.

That's a bit of an oxymoron, no? Why would not marking a var as unsaved would be more idiomatic and descriptive than not? 
For the idiomatic part, that's literally how the language wants you to use it. So you can hardly be more idiomatic, no? 
Its also descriptive, in indicating that it shouldn't be saved? Which implies that its either temporary or never changes from definition?


> If you're changing the value, it's not a constant. :/

I meant constant as in it doesn't change for the lifetime of the object. Its defined once and doesn't change.
Those probably would have been better as procs I guess. But, then there's the overhead. So I dunno.

File: code/modules/recycling/wrapped_package.dm (hunk 17 above)
Comment (PsyCommando) [reply code/modules/recycling/wrapped_package.dm:119]: I was replying to loaf.

File: code/modules/recycling/wrapped_package.dm (hunk 18 above)
Comment (PsyCommando) [reply code/modules/recycling/wrapped_package.dm]: Well, thanks for clarifying I guess.

File: code/game/objects/structures/crates_lockers/closets/__closet.dm (hunk 5 above)
Comment (PsyCommando) [reply code/game/objects/structures/crates_lockers/closets/__closet.dm:417]: Oh! Gotcha. 
I think I meant type checking the loc in this case. Maybe having some kind of generic break out proc for atoms be something to look into I guess?

File: code/modules/recycling/destination_tagger.dm (hunk 9 above)
Comment (PsyCommando) [reply code/modules/recycling/destination_tagger.dm]: Actually, my bad. Looking back at this again, I don't know what its for.  It doesn't set the machine var like it probably should first.
That's all code I moved from sorting_machinery.dm into its own file, and besides some renaming of procs, that bit is as-is.

File: code/modules/recycling/wrapped_package.dm (hunk 15 above)
Comment (afterthought2) [reply code/modules/recycling/wrapped_package.dm:22]: Idiomatic means using it the way it's used in the codebase or broader ss13 context; it's not used like this in the codebase. Idioms are about active usage, not some kind of language-design concepts. It's also not descriptive, as it's not temporary in any meaningful sense--it's the opposite, you are trying to indicate that it's immutable.

I agree it can't be made const. You can wrap it in a proc if you want but that's way afield from why I commented on this in review.

File: code/modules/recycling/wrapped_package.dm (hunk 17 above)
Comment (PsyCommando) [reply code/modules/recycling/wrapped_package.dm:119]: But yeah, the name thing is a problem and its known. Its just there's no clean and reliable way to do that currently, short of just removing that adjective in front of the name afaik.

File: code/modules/recycling/destination_tagger.dm (hunk 8 above)
Comment (PsyCommando) [reply code/modules/recycling/destination_tagger.dm]: done

File: code/modules/recycling/wrapped_package.dm (hunk 15 above)
Comment (PsyCommando) [reply code/modules/recycling/wrapped_package.dm:22]: I mean, its not like ss13 codebases will go out of their way to support a feature of DM they never use. 
Also, I can't help it that this is what they picked as name for their variable qualifier. It literally just means "not a constant, not a system variable, but still unsaved". Its not too far fetched to think someone that knows the language would be expected to know what that's for considering its in the doc?

File: code/modules/recycling/wrapped_package.dm (hunk 17 above)
Comment (PsyCommando) [reply code/modules/recycling/wrapped_package.dm:119]: Also, now that I look at this again, it probably won't be a problem for now. Since only human mobs are allowed to be wrapped currently.

File: code/modules/recycling/wrapped_package.dm (hunk 15 above)
Comment (afterthought2) [reply code/modules/recycling/wrapped_package.dm:22]: I looked up `tmp` in the dm documentation. Here is what it says:
```
The tmp type modifier indicates that an object variable should not be automatically written to the save file. This could mean that the variable is transient—that is, it is calculated at run-time and need not be saved. It could also indicated that the designer will handle saving of that variable specially and wishes to bypass the automated routine.
```
I don't buy your basic premise that this paragraph suggests that one should use `tmp` to designate immutable but type-specific vars. A literal reading would indicate that it should be used for any state var holding a reference, as none of those should be "written to a save file" in any reasonable context (indeed, the--admittedly ridiculous--examples following this in the dm docs make this very point). Along similar chains of reasoning, other vars that should have tmp are vars computed at runtime, like in `Initialize` or `Process`, and vars holding privileged information. But state vars pointing to static types are safe to store in savefiles, and so I see no reason why they would receive the `tmp` designation. You could argue that they are useless to store in savefiles if they are immutable (fine, sure)--but then you have to follow the actual logic and use `tmp` in all of the cases I just described, which will apply to most instance vars in the codebase.

This is of course all moot, because _we don't use savefiles_.

As I said, if you insist on `tmp` here, fine. But don't tell me your usage is specifically in line with the language architecture's intent, because that's not any more or less true than with any other sane usage pattern, such as e.g. never using it.

code/modules/recycling/destination_tagger.dm
PsyCommando: Removed that onclose.

code/modules/recycling/package_wrapper.dm
PsyCommando: Rewrote into an atom proc.

code/modules/recycling/package_wrapper.dm
PsyCommando: Fixed

code/modules/recycling/package_wrapper.dm
PsyCommando: Got rid of that proc anyways.

afterthought2 (approved):

PsyCommando:
@PlayerDeer Btw, if you ever want to make variants for wooden shipping crates(for the mule bot, animals and the like) and large metal crates(Like the hydroponics one), I wouldn't mind adding those variants. Since right now it looks a bit weird when wrapping those.

PR #2684, opened 2022-09-18
Fix duplicate datum regarding steel tiles

Qumefox:
## Description of changes
Changes the roofing tile datum name from /datum/stack_recipe/tile/metal/floor to /datum/stack_recipe/tile/metal/roof

Duplicate datum name was preventing regular steel floor tiles from being craftable. Only the last entry with that datum name was showing up in the list (which was the roofing tile)

## Authorship
Qumefox

## Changelog
:cl:
bugfix: roofing tile datum name
/:cl:

### DIFF
--- a/code/modules/materials/recipes_stacks.dm
+++ b/code/modules/materials/recipes_stacks.dm
@@ -32,7 +32,7 @@
 	title = "regular floor tile"
 	result_type = /obj/item/stack/tile/floor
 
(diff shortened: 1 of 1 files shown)


MistakeNot4892:
Whoops.
Wonder if we can lint this somehow.

MistakeNot4892 (approved):

PR #2692, opened 2022-09-21
Item Dmg: Give material or health to a million items

PsyCommando:
## Description of changes
This is part of the item damage PR #2575 with only the material changes, and the changes to the health values. In order to make the main PR less cluttered.

* Added materials to many many things, so they can properly determine their health, and other properties. 
* Also set the health on a few items as well. However, it won't have an effect until the damage PR is merged.

## Changelog
:cl:
code: Gave a material value, and matter list to a lot of items.
code: Gave health to a bunch of things that didn't specify it.
/:cl:

### DIFF
--- a/code/datums/uplink/services.dm
+++ b/code/datums/uplink/services.dm
@@ -61,6 +61,7 @@
 	w_class = ITEM_SIZE_TINY
 	icon = 'icons/obj/items/device/flash_synthetic.dmi'
 	icon_state = ICON_STATE_WORLD
+	health = ITEM_HEALTH_NO_DAMAGE
 	var/state = AWAITING_ACTIVATION
 	var/service_label = "Unnamed Service"
 	var/service_duration = 0 SECONDS
--- a/code/game/objects/items/weapons/cosmetics.dm
+++ b/code/game/objects/items/weapons/cosmetics.dm
@@ -7,11 +7,13 @@
 	w_class = ITEM_SIZE_TINY
 	slot_flags = SLOT_EARS
 	color = "#e00606"
+	material = /decl/material/solid/plastic
 	var/color_desc = "ruby"
 	var/open = FALSE
 
 /obj/item/lipstick/Initialize()
 	. = ..()
+	item_flags |= ITEM_FLAG_HOLLOW
 	if(color_desc)
 		desc += " This one is in [color_desc]."
 	update_icon()
--- a/code/game/objects/items/weapons/cosmetics.dm
+++ b/code/game/objects/items/weapons/cosmetics.dm
@@ -4,6 +4,7 @@
 	desc = "An unbranded tube of lipstick."
 	icon = 'icons/obj/items/lipstick.dmi'
 	icon_state = "lipstick_0"
+	item_flags = ITEM_FLAG_HOLLOW
 	w_class = ITEM_SIZE_TINY
 	slot_flags = SLOT_EARS
 	color = "#e00606"
@@ -13,7 +14,6 @@
 
 /obj/item/lipstick/Initialize()
 	. = ..()
-	item_flags |= ITEM_FLAG_HOLLOW
 	if(color_desc)
 		desc += " This one is in [color_desc]."
 	update_icon()
--- a/code/game/objects/items/weapons/explosives.dm
+++ b/code/game/objects/items/weapons/explosives.dm
@@ -8,6 +8,11 @@
 	item_flags = ITEM_FLAG_NO_BLUDGEON
 	w_class = ITEM_SIZE_SMALL
 	origin_tech = "{'esoteric':2}"
+	material = /decl/material/solid/plastic
+	matter = list(
+		/decl/material/solid/silicon = MATTER_AMOUNT_TRACE,
+		//No RDX or explosive stuff to put in there.. :c
+	)
 	var/datum/wires/explosive/c4/wires = null
 	var/timer = 10
 	var/atom/target = null
--- a/code/game/objects/items/weapons/explosives.dm
+++ b/code/game/objects/items/weapons/explosives.dm
@@ -11,7 +11,7 @@
 	material = /decl/material/solid/plastic
 	matter = list(
 		/decl/material/solid/silicon = MATTER_AMOUNT_TRACE,
-		//No RDX or explosive stuff to put in there.. :c
+		/decl/material/liquid/anfo = MATTER_AMOUNT_REINFORCEMENT, //#TODO: Slap RDX in here
 	)
 	var/datum/wires/explosive/c4/wires = null
 	var/timer = 10
--- a/code/game/objects/items/weapons/extinguisher.dm
+++ b/code/game/objects/items/weapons/extinguisher.dm
@@ -43,6 +43,7 @@
 
 /obj/item/extinguisher/Initialize()
 	. = ..()
+	item_flags |= ITEM_FLAG_HOLLOW
 	create_reagents(max_water)
 	if(starting_water > 0)
 		reagents.add_reagent(/decl/material/liquid/water, starting_water)
--- a/code/game/objects/items/weapons/extinguisher.dm
+++ b/code/game/objects/items/weapons/extinguisher.dm
@@ -6,6 +6,7 @@
 	item_state = "fire_extinguisher"
 	hitsound = 'sound/weapons/smash.ogg'
 	obj_flags = OBJ_FLAG_CONDUCTIBLE
+	item_flags = ITEM_FLAG_HOLLOW
 	throwforce = 10
 	w_class = ITEM_SIZE_NORMAL
 	throw_speed = 2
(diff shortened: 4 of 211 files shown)


code/datums/uplink/services.dm
```diff
@@ -61,6 +61,7 @@
 	w_class = ITEM_SIZE_TINY
 	icon = 'icons/obj/items/device/flash_synthetic.dmi'
 	icon_state = ICON_STATE_WORLD
+	health = -1
```
MistakeNot4892: Would prefer a macro for this, as discussed on the other PR.

PsyCommando: There's a macro for it. Its just the commit that replaces it is on the other PR rn.

code/game/objects/items/weapons/explosives.dm
```diff
@@ -8,6 +8,11 @@
 	item_flags = ITEM_FLAG_NO_BLUDGEON
 	w_class = ITEM_SIZE_SMALL
 	origin_tech = "{'esoteric':2}"
+	material = /decl/material/solid/plastic
+	matter = list(
+		/decl/material/solid/silicon = MATTER_AMOUNT_TRACE,
+		//No RDX or explosive stuff to put in there.. :c
```
MistakeNot4892: Could put in ANFO as a placeholder :U

code/game/objects/items/weapons/extinguisher.dm
```diff
@@ -43,6 +43,7 @@
 
 /obj/item/extinguisher/Initialize()
 	. = ..()
+	item_flags |= ITEM_FLAG_HOLLOW
```
MistakeNot4892: Shouldn't this be set on the definition?

code/game/objects/items/weapons/lighter.dm
```diff
@@ -213,6 +219,12 @@
 //Spawn using the colour list in the master type
 /obj/item/flame/lighter/zippo/random
 	random_colour = TRUE
+/obj/item/flame/lighter/zippo/random/Initialize()
+	. = ..()
+	if(color == COLOR_BRASS)
```
MistakeNot4892: Probably better to have the zippo pick a material and apply material colour rather than switching based on colour.

mods/content/xenobiology/slime/items_potion.dm
```diff
@@ -3,6 +3,8 @@
 	desc = "A potent chemical mix that will nullify a slime's powers, causing it to become docile and tame."
 	icon = 'icons/obj/items/chem/bottle.dmi'
 	icon_state = "bottle19"
+	material = /decl/material/solid/glass
+	item_state = ITEM_FLAG_HOLLOW
```
MistakeNot4892: `item_state` is not the right var for this.

code/modules/paperwork/paper_bundle.dm
```diff
@@ -11,6 +11,8 @@
 	throw_speed = 1
 	layer = ABOVE_OBJ_LAYER
 	attack_verb = list("bapped")
+	health = 10
+	max_health = 10
```
MistakeNot4892, line 15: Possibly not in scope for this PR, but I feel like we should set `max_health` only, and `health` should autoset from that in `Initialize()`.

code/game/objects/structures/stool_bed_chair_nest_sofa/wheelchair.dm
```diff
@@ -135,3 +136,9 @@
 		user.visible_message(SPAN_NOTICE("<b>[user]</b> lays out \the [W.name]."))
 		W.add_fingerprint(user)
 		qdel(src)
+
+/obj/item/wheelchair_kit/physically_destroyed(skip_qdel)
```
MistakeNot4892, line 140: Idle note, not in scope for this PR, we need a generic 'itemized carry form' thing for this and rollerbeds, like the mob holder item. Maybe a generalized holder.

code/modules/paperwork/paper_bundle.dm
PsyCommando: That's done in the actual item damage PR even if I actually left both.

code/datums/uplink/services.dm
PsyCommando: I pushed the commit with a bunch of crap removed

code/game/objects/items/weapons/explosives.dm
PsyCommando: I mean, I guess XD

code/game/objects/items/weapons/extinguisher.dm
PsyCommando: Could probably move it there yeah.

code/game/objects/items/weapons/cosmetics.dm
```diff
@@ -7,11 +7,13 @@
 	w_class = ITEM_SIZE_TINY
 	slot_flags = SLOT_EARS
 	color = "#e00606"
+	material = /decl/material/solid/plastic
 	var/color_desc = "ruby"
 	var/open = FALSE
 
 /obj/item/lipstick/Initialize()
 	. = ..()
+	item_flags |= ITEM_FLAG_HOLLOW
```
MistakeNot4892: Same as extinguisher, should this be on the definition?

code/game/objects/items/weapons/lighter.dm
PsyCommando: What's really annoying with that is, the base lighter is a cheap plastic lighter that also comes in several colors. So they both share a system to randomize their color. So I'd have to make several implementations for each types of materials for the zippo?

mods/content/xenobiology/slime/items_potion.dm
PsyCommando: Nice catch!

code/game/objects/items/weapons/lighter.dm
```diff
@@ -31,6 +33,7 @@
 
 /obj/item/flame/lighter/Initialize()
 	. = ..()
+	item_flags |= ITEM_FLAG_HOLLOW
```
MistakeNot4892: As per extinguishers.

code/game/objects/items/weapons/lighter.dm
MistakeNot4892: Per Discord I would suggest something a la https://github.com/MistakeNot4892/Nebula/commit/306517604a819198c91ceb5816a33c26abc6d872 since the randomization is restricted to two specific `/random` classes.

code/game/objects/items/weapons/soap.dm
```diff
@@ -9,6 +9,9 @@
 	throwforce = 0
 	throw_speed = 4
 	throw_range = 20
+	material = /decl/material/liquid/cleaner
```
MistakeNot4892, line 12: This doesn't make a huge amount of sense. Should probably use a fat or oil reagent if we have it. Maybe a nutrient subtype. If we ever add phase transitions soap is going to instantly melt in this case.

code/game/objects/items/weapons/staff.dm
```diff
@@ -9,16 +9,21 @@
...
 /obj/item/staff/gentcane
 	name = "gentleman's cane"
 	desc = "An ebony can with an ivory tip."
 	icon = 'icons/obj/items/cane.dmi'
 	icon_state = "cane"
-	item_state = "stick"
\ No newline at end of file
+	item_state = "stick"
+	material = /decl/material/solid/wood/ebony
+	matter = list(/decl/material/solid/plastic = MATTER_AMOUNT_TRACE) //No ivory material :c
```
MistakeNot4892: Could just use bone?

code/game/objects/structures/stool_bed_chair_nest_sofa/wheelchair.dm
```diff
@@ -135,3 +136,9 @@
...
+	//Make sure if the kit is destroyed to drop the same stuff as the actual wheelchair
+	var/obj/structure/S = new structure_form_type(get_turf(src)) 
+	S.physically_destroyed()
```
MistakeNot4892, line 143: Only worry here is whether or not the holder item generates its own debris in `physically_destroyed()`

code/modules/locks/key.dm
```diff
@@ -17,6 +18,7 @@
 /obj/item/key/soap
 	name = "soap key"
 	desc = "a fragile key made using a bar of soap."
+	material = /decl/material/liquid/cleaner
```
MistakeNot4892, line 21: Definitely feel like we need a placeholder soap material now.

code/game/objects/items/weapons/soap.dm
PsyCommando: We don't have anything else than cleaner right now. :c

code/game/objects/items/weapons/staff.dm
PsyCommando: I guess.

code/game/objects/structures/stool_bed_chair_nest_sofa/wheelchair.dm
PsyCommando: It doesn't. Most items without the brittle material flag do not create debris when destroyed. Unless they reimplement physically destroyed.

code/game/objects/items/weapons/lighter.dm
PsyCommando: Done!

MistakeNot4892 (approved):

PR #2694, opened 2022-09-21
Cleaning up matter initialization.

MistakeNot4892:
## Description of changes
- Moves `material` and `set_material()` down to `/obj`.
- Cleans up some Initialize() overrides on `/obj/item` and `/obj/structure`.
- Reworks some matter gen to make recipes less brittle.

Fixes issues with the inconsistent material from crafting recipes unit test, works towards standardized material handling.

## Authorship
Myself and @PsyCommando.

## Changelog
Nothing player-facing.


code/game/machinery/_machines_base/stock_parts/network_receiver.dm
```diff
@@ -6,7 +6,7 @@
 	icon_state = "net_lock"
 	part_flags = PART_FLAG_QDEL
 
-/obj/item/stock_parts/network_receiver/Initialize(ml, material_key)
+/obj/item/stock_parts/network_receiver/Initialize()
```
PsyCommando, line 9: If you remove those the compiler won't let you refer to them as named parameters with subclasses. And the auto-complete won't list them as parameters. It really is not a good idea to remove them imo.

MistakeNot4892: I'm removing them so I can mass find and replace Initialize() later.

PsyCommando: Alright then.

MistakeNot4892:
Will revisit these down the track.

PR #2266, opened 2021-12-26
interface: skin cleanup

quardbreak:
## Description of changes
изображение

* Cleanups skin.dmf and interface related things:
  * All skins elements renamed in flavor of name standardization.
  * Fixes weird elements offset on the right side of the screen.
  * Most buttons are gone now, only four left.
изображение
  * Replaces default status bar with alternative implementation using label element.
  * Prefix in the info panel use blue color now.

* Not skin related changes:
  * Kills discord button.
  * Renames forum button to community.
  * Shuffles Show Server Revision. Now it prints everything, but if server don't have the repository link set, the link won't be created. The output style will look like `branch-example | 2022-04-28 | hash`.
изображение
  * Fixes tooltips offsets. Removes clockcult theme. Updates jquery script link.

## Why and what will this PR improve
I wanted to redesign interface to be more Antimonium-like fullscreen interface with dark layout without borders, yet I'm only able to cleanup things right now!

## Authorship
tgstation: status bar label code.
me: everything else.

PR #2695, opened 2022-09-21
Couple of interface fixes

quardbreak:
## Description of changes
Part of https://github.com/NebulaSS13/Nebula/pull/2266
Used some changes for staging fixes.

Fixes tooltips for non-square view.
A bit nicer `/client/proc/toggle_fullscreen()` and `getviewsize()`.

## Authorship
Me.

### DIFF
--- a/code/modules/tooltip/tooltip.dm
+++ b/code/modules/tooltip/tooltip.dm
@@ -3,31 +3,33 @@ Tooltips v1.1 - 22/10/15
 Developed by Wire (#goonstation on irc.synirc.net)
 - Added support for screen_loc pixel offsets. Should work. Maybe.
 - Added init function to more efficiently send base vars
+
 Configuration:
 - Set control to the correct skin element (remember to actually place the skin element)
 - Set file to the correct path for the .html file (remember to actually place the html file)
 - Attach the datum to the user client on login, e.g.
-
 /client/New()
 	src.tooltips = new /datum/tooltip(src)
 
 Usage:
 - Define mouse event procs on your (probably HUD) object and simply call the show and hide procs respectively:
-
 /obj/screen/hud
 	MouseEntered(location, control, params)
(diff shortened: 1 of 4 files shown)


code/modules/tooltip/tooltip.html
```diff
@@ -93,7 +93,7 @@
 	<div id="wrap" class="wrap">
 		<div id="content" class="content"></div>
 	</div>
-	<script type="text/javascript" src="https://code.jquery.com/jquery-1.11.3.min.js"></script>
+	<script type="text/javascript" src="https://code.jquery.com/jquery-1.12.4.min.js"></script>
```
quardbreak, line 96: I think this should not be depended to external distributors. What about keeping this locally in codebase and compile with all assets?

MistakeNot4892: I am against bundling other libraries with our codebase. I'm honestly not clear why we need jquery for tooltips at all.

code/modules/tooltip/tooltip.dm
```diff
@@ -3,113 +3,121 @@ Tooltips v1.1 - 22/10/15
...
-	if(queueHide)
+	if (queueHide)
 		hide()
 
 	return TRUE
 
+
 /datum/tooltip/proc/hide()
-	if(queueHide)
+	queueHide = showing ? TRUE : FALSE
```
MistakeNot4892: `queueHide = !!showing`

code/modules/tooltip/tooltip.html
afterthought2: I think the target IE versions don't support sane js async handling, so while you aren't forced to use jquery per se it's not unreasonable to. We do bundle jquery with assets already (version shouldn't be super relevant), so could access it from there; the method here might be more reliable due to shitty dm asset distribution though.

MistakeNot4892 (approved):

PR #2696, opened 2022-09-23
Stackabled ore piles + accompanying tweaks

PsyCommando:
## Description of changes
* Made ores into stacks.
* Modified ore box to handle ore stacks. And updated some of the code used by it.
* Since the default stone type is invalid, I made a granite stone material, which is more of a generic kind of stone, and not sedimentary. Mainly for the default ore material
* Updated type path of ores all over the code.
* Turned xenoarch strangerocks into items since they have no real reasons to inherit from ores.
* Fixed some calls to the ore constructor with the parameters in the wrong order.

## Changelog
:cl:
tweak: Ores now stack.
tweak: Can now pick a random ore stack from the ore box.
tweak: Ore box has alt interactions.
refactor: Refactored ore, ore box, and strangerocks.
/:cl:


PsyCommando:
I was wondering if anyone had some suggestions for generating stack icon states? Because so far it seems I'm doing something stupid and I haven't spotted it yet..

PsyCommando:
Okay, managed to fix it. Was just the way I indexed rendered stacks..

code/game/objects/items/stacks/stack.dm
```diff
@@ -65,6 +65,7 @@
 	. = ..()
 	if(!isturf(loc))
 		var/image/I = image(null)
+		I.layer = POINTER_LAYER
```
MistakeNot4892: What is the purpose of this change?

code/modules/materials/materials_ore.dm
```diff
@@ -1,27 +1,106 @@
...
+	max_amount                 = ORE_MAX_AMOUNT
+	material                   = /decl/material/solid/stone/granite //By default is just a rock
 	material_health_multiplier = 0.5
+	stack_merge_type           = /obj/item/stack/material/ore
+	randpixel                  = 8
+	applies_material_name      = FALSE //Handled in override
+	randpixel                  = 6
+	///Associative list of cache key to the generate icons for the ore piles. We pre-generate a pile of all possible ore icon states, and make them available
+	var/static/list/cached_ore_icon_states
+	///A list of all the existing ore icon states in the ore file
```
MistakeNot4892, line 22: You could actually pull this straight out of the ore file with `icon_states()`.

code/modules/materials/materials_ore.dm
```diff
@@ -1,27 +1,106 @@
...
+	cached_ore_icon_states = list()
+	for(var/IS in ore_icon_states)
+		var/list/states = list(null) //First index is null since we're creating overlays
+		var/image/scrapboard = image(null)
+		//Generate base image
+		for(var/i = 2 to ORE_MAX_OVERLAYS)
+			//Randomize the orientation and position of each ores in the image
+			var/matrix/M = matrix()
+			M.Translate(rand(-7, 7), rand(-8, 8))
+			M.Turn(pick(-72, -58, -45, -27.-5, 0, 0, 0, 0, 0, 27.5, 45, 58, 72))
```
MistakeNot4892, line 47: This is going to look super weird with the `dust` state.

code/modules/mining/satchel_ore_boxdm.dm
```diff
@@ -2,91 +2,148 @@
...
+	var/total_ores = 0
+	///A list with all the ores of every types that are currently inside
+	var/list/stored_ore
+
+/obj/structure/ore_box/attack_hand(mob/user)
+	if(total_ores > 0)
+		to_chat(user, SPAN_NOTICE("You grab a random ore pile from \the [src]."))
+		var/obj/item/stack/material/ore/O = pick(get_contained_external_atoms())
+		if(remove_ore(O, user))
+			return TRUE
```
MistakeNot4892, line 25: Wouldn't gate the return here personally, or rather I would move the to_chat down into the gate.

code/game/objects/items/stacks/stack.dm
PsyCommando: To prevent the numbers from getting covered by the ore's overlays.

code/modules/materials/materials_ore.dm
PsyCommando: How would I do that?

code/modules/materials/materials_ore.dm
PsyCommando: Probably, but its not really meant to be permanent. Its more like a place holder until we get some real icons made.

code/modules/mining/satchel_ore_boxdm.dm
PsyCommando: Well, its just if I don't gate it, it'll eat all attack hands whether it actually can't or can remove any ores.

code/game/objects/items/stacks/stack.dm
MistakeNot4892: Hadn't thought of that, in my brain it was still using atom maptext.
I would suggest:
````
I.plane = HUD_PLANE
I.layer = HUD_ABOVE_ITEM_LAYER
````
as the POINTER_LAYER value will mean something different when the atom is not on DEFAULT_PLANE. If I remember my own code properly you shouldn't see the marker except when the atom is in the inventory anyway.

code/modules/materials/materials_ore.dm
MistakeNot4892: `cached_ore_icon_states = icon_states('icons/obj/materials/ore.dmi')` - it's constant too, can chuck it straight on the var definition.

code/modules/mining/satchel_ore_boxdm.dm
MistakeNot4892: Then move the message down? My issue is that if you fail to remove the ore you'll still get 'You grab a random ore'.

code/modules/materials/materials_ore.dm
MistakeNot4892: A comment to that effect would be good - also personally I would like the rotation to be gated behind an icon state check, though I won't block over that.

code/modules/materials/materials_ore.dm
PsyCommando: Alright, thanks!

code/modules/mining/satchel_ore_boxdm.dm
PsyCommando: Ah, gotcha.

MistakeNot4892 (approved):

PR #2697, opened 2022-09-23
More work on chem rounding/remove_any().

MistakeNot4892:
## Description of changes
- Swaps out `round()` in reagent rounding for a macro that always floors.
- Reworks `remove_any()` to avoid failing to remove anything, which may cause loops 

Trying to hammer out the recursion crashes in SSmaterials.

## Authorship
Myself and @out-of-phaze

## Changelog
Nothing player-facing.

### DIFF
--- a/code/__defines/maths.dm
+++ b/code/__defines/maths.dm
@@ -13,3 +13,6 @@
 // Will filter out extra rotations and negative rotations
 // E.g: 540 becomes 180. -180 becomes 180.
 #define SIMPLIFY_DEGREES(degrees) (MODULUS_FLOAT((degrees), 360))
+
+// Float-aware floor since round() will round upwards when given a second arg
+#define FLOAT_FLOOR(X, N) (round(X/N)*N)
--- a/code/__defines/maths.dm
+++ b/code/__defines/maths.dm
@@ -1,7 +1,11 @@
 // Macro functions.
 #define RAND_F(LOW, HIGH) (rand() * (HIGH - LOW) + LOW)
 #define CEILING(x) (-round(-(x)))
+
+// Float-aware floor and ceiling since round() will round upwards when given a second arg.
+#define NONUNIT_FLOOR(x, y)    (round( (x) / (y)) * (y))
 #define NONUNIT_CEILING(x, y) (-round(-(x) / (y)) * (y))
+
 #define MULT_BY_RANDOM_COEF(VAR,LO,HI) VAR =  round((VAR * rand(LO * 100, HI * 100))/100, 0.1)
 
 #define ROUND(x) (((x) >= 0) ? round((x)) : -round(-(x)))
@@ -13,6 +17,3 @@
 // Will filter out extra rotations and negative rotations
 // E.g: 540 becomes 180. -180 becomes 180.
 #define SIMPLIFY_DEGREES(degrees) (MODULUS_FLOAT((degrees), 360))
-
-// Float-aware floor since round() will round upwards when given a second arg
-#define FLOAT_FLOOR(X, N) (round(X/N)*N)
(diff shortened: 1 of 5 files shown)


code/modules/reagents/Chemistry-Holder.dm
```diff
@@ -269,13 +270,47 @@ var/global/obj/temp_reagents_holder = new
...
+		return
+
+	var/removing = FLOAT_FLOOR(Clamp(amount, 0, max(0, total_volume)), MINIMUM_CHEMICAL_VOLUME) // not ideal but something is making total_volume become NaN
+	if(!removing || total_volume <= 0)
+		. = 0
+		clear_reagents()
+		return
+
+	var/part = . / total_volume
+	if(part < MINIMUM_CHEMICAL_VOLUME)
```
MistakeNot4892: This is flat-out wrong but my brain meat is struggling to work out the appropriate check. Basically want to short circuit here if there's no capacity for us to remove anything. Possibly `total_volume/length(reagent_volumes)` or something along those lines.

code/modules/reagents/Chemistry-Holder.dm
```diff
@@ -269,13 +270,47 @@ var/global/obj/temp_reagents_holder = new
...
+			. += removing_amt
+			remove_reagent(current, removing_amt, TRUE, TRUE)
+
+		// We failed to remove anything, so nuke the list and return.
+		if(all_too_low)
+			. += total_volume
+			clear_reagents()
+			return
+
+	if(!defer_update)
```
MistakeNot4892, line 300: It might be worth inlining the `remove_any()` logic into `update_total()` to avoid the recursion in the future.

MistakeNot4892:
Starting to wonder if I should just strip out all of the rounding and accept the floating point issues.

Alceris:
Honestly not sure what was the reason for changing all this from when there were no apparent rounding issues.

MistakeNot4892:
As noted on Discord, there were floating point math issues with the previous implementation. There were also incidents of reagent holders becoming NaN but that may have been related to the rounding or to the `defer_update` logic in the fluid system.

MistakeNot4892:
This does appear to fix the MC lockups on Scav at least.

MistakeNot4892:
Was reverted on Scav, and actually causes the blood bug, not the commit following it. :(

MistakeNot4892:
Cut this right back and tested it, appears to work now.

code/__defines/maths.dm
```diff
@@ -13,3 +13,6 @@
 // Will filter out extra rotations and negative rotations
 // E.g: 540 becomes 180. -180 becomes 180.
 #define SIMPLIFY_DEGREES(degrees) (MODULUS_FLOAT((degrees), 360))
+
+// Float-aware floor since round() will round upwards when given a second arg
+#define FLOAT_FLOOR(X, N) (round(X/N)*N)
```
afterthought2: See also `NONUNIT_CEILING` on line 4, should probably be adjacent and similarly named for consistency.

code/modules/reagents/Chemistry-Holder.dm
```diff
@@ -269,13 +270,35 @@ var/global/obj/temp_reagents_holder = new
...
-	. = Clamp(amount, 0, max(0, total_volume)) // not ideal but something is making total_volume become NaN
-	if(.)
-		var/part = . / total_volume
+
+	if(amount >= total_volume)
+		. = total_volume
+		clear_reagents()
+		return
+
+	var/removing = Clamp(FLOAT_FLOOR(amount, MINIMUM_CHEMICAL_VOLUME), 0, max(0, total_volume)) // not ideal but something is making total_volume become NaN
```
afterthought2: The `max(0, ` appears redundant.

MistakeNot4892: Yeah this was an attempt at fixing negative total_volumes, which turned out to be related to the remove_any() issues.

afterthought2 (approved):

PR #2704, opened 2022-09-27
Fix issues found by OpenDream

out-of-phaze:
## Description of changes

- Replaces three redundant clamp helpers with the BYOND builtin
- Removes unnecessary line ending escapes in ZAS code
- Fixes a space in `to_chat (usr, "foo")` in eyepatch code
- Fix a number of non-constant switch conditions (these silently evaluate to null and break in DM)
- Adds a semicolon to the COLORED_FOLDER define

From 7,700 errors in OpenDream compilation down to just 9.
Also, some of these are actual errors that DM simply doesn't report and breaks silently on.

## Authorship
Me.


code/controllers/subsystems/jobs.dm
```diff
@@ -285,16 +285,17 @@ SUBSYSTEM_DEF(jobs)
 				var/age = V.client.prefs.get_character_age()
 				if(age < job.minimum_character_age) // Nope.
 					continue
-				switch(age)
-					if(job.minimum_character_age to (job.minimum_character_age+10))
-						weightedCandidates[V] = 3 // Still a bit young.
-					if((job.minimum_character_age+10) to (job.ideal_character_age-10))
-						weightedCandidates[V] = 6 // Better.
-					if((job.ideal_character_age-10) to (job.ideal_character_age+10))
+				switch(age - job.ideal_character_age)
```
MistakeNot4892, line 288: Note to self, make ideal_character_age respect species age variance.

code/modules/mob/living/silicon/robot/life.dm
```diff
@@ -179,10 +179,11 @@
 						src.healths.icon_state = "health3"
 					if(0 to 50)
 						src.healths.icon_state = "health4"
-					if(config.health_threshold_dead to 0)
```
MistakeNot4892, line 182: Are nonconstant switches an issue in DM or just in OD?

out-of-phaze: In DM. The non-constant parts just silently evaluate to null IIRC.

MistakeNot4892 (approved):

PR #2705, opened 2022-09-29
Add cocktail tastes

out-of-phaze:
## Description of changes
- Adds `var/list/tastes` to cocktails, an associative list with the format `taste_description = taste_strength`.
- The strength is relative to the number of ingredients, which allows for basic reasoning about relative strength assuming each ingredient has a taste_mult of 1. That means a flavor with a strength of 5 is roughly 5x the intensity of the ingredients, 0.5 is half as strong, etc.

Necessary for downstream stuff, and also (optionally) replicates the old functionality of cocktails with the benefits of the new system.

## Authorship
Me.

### DIFF
--- a/code/modules/mob/living/carbon/taste.dm
+++ b/code/modules/mob/living/carbon/taste.dm
@@ -5,7 +5,7 @@
 	if(last_taste_time + 50 < world.time)
 		var/datum/reagents/temp = new(amount, global.temp_reagents_holder) //temporary holder used to analyse what gets transfered.
 		from.trans_to_holder(temp, amount, multiplier, 1)
-		var/text_output = temp.generate_taste_message(src)
+		var/text_output = temp.generate_taste_message(src, from)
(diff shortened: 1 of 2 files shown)


code/modules/mob/living/carbon/taste.dm
```diff
@@ -47,6 +47,19 @@ calculate text size per text.
 				else
 					tastes[taste_desc] = taste_amount
 
+		var/decl/material/primary_ingredient = get_primary_reagent_decl()
+		if(primary_ingredient?.cocktail_ingredient && source_holder?.my_atom)
+			for(var/decl/cocktail/cocktail in SSmaterials.get_cocktails_by_primary_ingredient(primary_ingredient.type))
+				if(!cocktail.matches(source_holder.my_atom))
+					continue
+				if(!LAZYLEN(cocktail.tastes))
```
MistakeNot4892, line 53: Probably should do this before `matches` as a microopt.

MistakeNot4892 (dismissed):
Not hugely fussed about the optimization mentioned, will reapprove if you feel like committing it.

MistakeNot4892 (approved):

PR #2707, opened 2022-10-01
Enable full diagnostics suite

PsiOmegaDelta:
## Description of changes

Adds the settings documented in https://github.com/SpaceManiac/SpacemanDMM/blob/master/CONFIGURING.md at the error level

Ensures that some code standards are enforced via automation, with the exception of redefined_proc

## Authorship

Borrowed from https://github.com/Baystation12/Baystation12/pull/32504

### DIFF
--- a/SpacemanDMM.toml
+++ b/SpacemanDMM.toml
@@ -10,3 +10,44 @@ disallow_relative_proc_definitions = true
 
 [debugger]
 engine = "auxtools"
+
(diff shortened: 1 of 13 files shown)


MistakeNot4892 (approved):
I think most or all of this could be pointed at staging, but looks good generally. The override collapsing has been something I have vaguely thought we should do for awhile.

PR #2708, opened 2022-10-02
Fix ice being named water

out-of-phaze:
## Description of changes
- Fixes ice being named water.
- Makes the codex, chem dispensers, and chemmasters/condimasters use `use_name` rather than `name`.
Targeting dev due to making potentially-breaking changes to things like the codex/dispenser/chemmaster; despite being told to use `use_name` it doesn't seem like much uses or sets it, so this might make things have funky names.

ice should be named ice and not water

## Authorship
Me.

### DIFF
--- a/code/modules/codex/categories/category_cocktails.dm
+++ b/code/modules/codex/categories/category_cocktails.dm
@@ -22,7 +22,7 @@
 			// which we normalize internally for calculations.
 			var/decl/material/mixer = GET_DECL(rtype)
 			var/minimum_amount = cocktail.display_ratios[rtype]
-			var/ingredient = "<span codexlink='[mixer.codex_name || mixer.name] (substance)'>[mixer.name]</span>"
+			var/ingredient = "<span codexlink='[mixer.codex_name || mixer.name] (substance)'>[mixer.use_name]</span>"
 			if(minimum_amount)
 				ingredient = "[minimum_amount] part\s [ingredient]"
 			else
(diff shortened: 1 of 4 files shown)


code/modules/codex/categories/category_cocktails.dm
```diff
@@ -22,7 +22,7 @@
 			// which we normalize internally for calculations.
 			var/decl/material/mixer = GET_DECL(rtype)
 			var/minimum_amount = cocktail.display_ratios[rtype]
-			var/ingredient = "<span codexlink='[mixer.codex_name || mixer.name] (substance)'>[mixer.name]</span>"
+			var/ingredient = "<span codexlink='[mixer.codex_name || mixer.use_name] (substance)'>[mixer.use_name]</span>"
```
MistakeNot4892: Invalid, codex page gen uses `name`. Either need to change it in `substances/Populte()` or use `name` in the link here.

code/modules/reagents/dispenser/dispenser2.dm
```diff
@@ -150,7 +150,7 @@
 	if(LAZYLEN(container?.reagents?.reagent_volumes))
 		for(var/rtype in container.reagents.reagent_volumes)
 			var/decl/material/R = GET_DECL(rtype)
-			beakerD[++beakerD.len] = list("name" = R.name, "volume" = REAGENT_VOLUME(container.reagents, rtype))
+			beakerD[++beakerD.len] = list("name" = R.use_name, "volume" = REAGENT_VOLUME(container.reagents, rtype))
```
MistakeNot4892, line 153: This should probably be `liquid_name` given that it's a beaker and reagents are handled as liquids regardless of state.

MistakeNot4892: Nvm, just grokked that this would mess with ice.

MistakeNot4892 (approved):

PR #2716, opened 2022-10-06
Fix APC being deleted in power alarm clear

### LINKED ISSUE #2715: APCs get deleted when their power alarm is cleared.
#### Description of issue
APCs get deleted when a power alarm they raise is cleared.

APCs should not disappear when their power alarm clears.

#### Steps to reproduce
Drain an APC's battery until it raises a power alarm, then fix the alarm by giving it charge (i.e. a new battery). You might even be able to just pull the battery out and put it back in to cause it.

N/A

The past four months, apparently.

Current dev

#### Issue bingo

out-of-phaze:
## Description of changes
Deletes the alarm source datum instead of the APC when a power alarm is cleared.

Fixes #2715.

## Authorship
Me.

### DIFF
--- a/code/modules/alarm/alarm.dm
+++ b/code/modules/alarm/alarm.dm
@@ -65,7 +65,7 @@
 	var/datum/alarm_source/AS = sources_assoc[source]
 	sources -= AS
 	sources_assoc -= source
-	qdel(source)
+	qdel(AS)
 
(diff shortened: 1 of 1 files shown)


MistakeNot4892 (approved):
Powerful code.

PR #2722, opened 2022-10-08
Porting SStyping and speech bubbles from Polaris.

MistakeNot4892:
## Description of changes
- Typing indicator is now handled by a subsystem and can report typing in the client text bar. This has been tested with a decent pop on Polaris (15 people) and showed no noticeable performance impacts.
- Mobs can now override the typing indicator/speech indicator appearance.

Generally more immersive and informative speech.

## Authorship
@Spookerton wrote SStyping for Polaris, which was based on/inspired by work by @Atermonera as far as I understand. The ghost and synth speech icons are also from Polaris, but I'm not sure if they originally came from Virgo or such.

## Changelog
:cl:
tweak: Using 'say' or 'me' in the client text bar will now show the typing indicator as though you used the F3 hotkey.
tweak: Synthetic mobs and ghosts now have their own speech overlay appearance.
/:cl:


MistakeNot4892:
Mostly working other than typing indicator not removing itself after speech.

code/controllers/subsystems/typing.dm
```diff
@@ -0,0 +1,208 @@
...
+
+/datum/controller/subsystem/typing/Recover()
+	status = list()
+	queue = list()
+
+/datum/controller/subsystem/typing/fire(resumed, no_mc_tick)
+	if(!resumed)
+		queue = list()
+		for (var/client/client as anything in global.clients)
+			queue += client
```
afterthought2: Why not `clients.Copy()`?

code/controllers/subsystems/typing.dm
```diff
@@ -0,0 +1,208 @@
...
+		), ...)
+		See .proc/get_entry for details.
+	*/
+	var/static/list/status = list()
+
+	/// Matches input bar verbs that should set TYPING_STATE_INPUT.
+	var/static/regex/match_verbs = regex("^(Me|Say) +\"?\\w+")
+
+	/// A list of clients waiting to be polled for input state.
+	var/static/list/client/queue = list()
```
afterthought2: Why is this (and status) static? It only affects Recover...but does so negatively, in that you need extra Recover handling because for some reason these are static.

MistakeNot4892:
Ready for review.

code/controllers/subsystems/typing.dm
```diff
@@ -0,0 +1,208 @@
...
+			icon_state = "[speech_state_modifier]_typing"
+		else
+			icon_state = initial(icon_state)
+
+	if(ismovable(master))
+		var/atom/movable/owner = master
+		add_vis_contents(owner, src)
+
+/atom/movable/overlay/typing_indicator/SetInitLoc()
+	forceMove(get_turf(master))
```
afterthought2: Doesn't this make it visible twice, once through vis contents and the second time just due to being at this loc? I would have assumed it should be positioned in null.

code/controllers/subsystems/typing.dm
```diff
@@ -0,0 +1,208 @@
...
+		if(!islist(entry) || !entry[INDEX_PREFERENCE])
+			continue
+		if(!entry[INDEX_INFLIGHT])
+			update_indicator_state_from_winget(client, entry)
+		else if(world.time < entry[INDEX_TIMEOUT])
+			entry[INDEX_INFLIGHT] = FALSE
+		if(no_mc_tick)
+			CHECK_TICK
+		else if(MC_TICK_CHECK)
+			queue.Cut(1, cut_until)
```
afterthought2, line 69: Feel like this is a needlessly complicated queue design. Why not just while .len/pop like dozens of other SS, given that you clear the whole queue each time anyway?

code/controllers/subsystems/typing.dm
```diff
@@ -0,0 +1,208 @@
...
+	if(display == target.is_typing)
+		return
+	if(display)
+		if(!target.typing_indicator)
+			target.typing_indicator = new(target)
+		target.is_typing = TRUE
+		target.typing_indicator.show_typing_indicator()
+	else if(target.typing_indicator)
+		target.is_typing = FALSE
+		target.typing_indicator.hide_typing_indicator()
```
afterthought2, line 137: You may need some way to deconflict multiple calls to these async procs. In theory you are calling one of the two roughly every half-second, and hotkey say can call them too, so multiple calls might be active at once depending on server load, network timing, etc. It's not clear they will be executed sequentially in the right order, or whether that's even desirable.

code/controllers/subsystems/typing.dm
```diff
@@ -0,0 +1,208 @@
...
+	alpha = 0
+	animate(src, transform = 0, alpha = 255, time = 0.2 SECONDS, easing = EASE_IN)
+
+	if(ismob(master))
+		var/mob/owner = master
+		var/speech_state_modifier = owner.get_speech_bubble_state_modifier()
+		if(speech_state_modifier)
+			icon_state = "[speech_state_modifier]_typing"
+		else
+			icon_state = initial(icon_state)
```
afterthought2, line 190: I'm pretty confused why this (and below) logic is after the animate.

nebula.dme
```diff
@@ -5,6 +5,447 @@
 // END_INTERNALS
 // BEGIN_FILE_DIR
 #define FILE_DIR .
+#define FILE_DIR "code"
```
afterthought2: What's going on with this file?

code/controllers/subsystems/typing.dm
```diff
@@ -0,0 +1,208 @@
...
+	PRIVATE_PROC(TRUE)
+	set waitfor = FALSE
+	var/timeout = world.time + INFLIGHT_TIMEOUT
+	entry[INDEX_INFLIGHT] = TRUE
+	entry[INDEX_TIMEOUT] = timeout
+	var/content = winget(client, INPUT_HANDLE, "text")
+	if(timeout != entry[INDEX_TIMEOUT]) // We're stale. Touch nothing.
+		return
+	entry[INDEX_INFLIGHT] = FALSE
+	if(QDELETED(client) || !isliving(client.mob))
```
afterthought2: This living check is made here but nowhere else.

afterthought2 (commented):
Well, this is a way of doing it.

I would at least consider storing all state on mobs instead of this client stuff. Clients are awful to work with and it's not any harder going mob -> client to poll than going client -> mob to update indicators.

nebula.dme
MistakeNot4892: Does it every time I open DM and accidentslly committed it.

code/controllers/subsystems/typing.dm
MistakeNot4892: Spook has some opinions on the performance of the 'standard' list iter pattern afaik.

code/controllers/subsystems/typing.dm
MistakeNot4892: Yeah this is an artefact of merging SStyping with our indicator code. Will tidy up.

MistakeNot4892:
Not sure I like the idea of storing this state on mob when only clients can show the typing indicator.

code/controllers/subsystems/typing.dm
afterthought2: Well, if you like it you can do it like this, or ask Lohikar for opinions. My go-to more performant variant is just an index you iterate using, no list ops at all. But that probably would have an improvement on the same order as e.g. removing the client data getter proc: not much.

code/controllers/subsystems/typing.dm
MistakeNot4892: Done.

code/controllers/subsystems/typing.dm
MistakeNot4892: Removed.

code/controllers/subsystems/typing.dm
MistakeNot4892: Tidied up.

code/controllers/subsystems/typing.dm
MistakeNot4892: Reorganized this block a bit.

nebula.dme
MistakeNot4892: Removed.

code/controllers/subsystems/typing.dm
MistakeNot4892: Removed.

code/controllers/subsystems/typing.dm
MistakeNot4892: The worst case for this would be potentially something like hitting F3 while you have a say command in the taskbar, which would hide the indicator until the next SStyping tick. I'm not sure it's worth further complexity for a brief visual flicker in an edge case. Happy to adjust if you have any recommendations/suggestions for how to disambiguate it cleanly.

afterthought2:
Was thinking: say I hit t for hotkey say, speech bubble shows up. But I'm not typing into the chatbar, so now it goes away on the next poll. There's even an `is_typing` var that you can probably use to avoid this, it's just the hotkey verbs don't seem to use it.

Along similar lines: does actually submitting a say via verb bar-enter hide the chat indicator? This should be possible because this is a detectable event, though I don't know exactly which verb is triggered.

If you do detect typing activity when polling, it may make sense to lower the repoll timeout until the mob is no longer typing, though maybe not needed in practice (can't tell without testing).

MistakeNot4892:
Okay, picked through this.
`INDEX_VERB_STATE` is set when the verbs are invoked and cleared when they finish (or when the mob state changes). 
`INDEX_INPUT_STATE` tracks the chatbar.
They shouldn't work across each other, if either is set then the indicator will still be presented.

MistakeNot4892:
Checked and confirmed they don't interfere with each other.

afterthought2 (approved):
OK, let's try it. Try having it tested post-merge somewhere; I can't test it and it's not clear that local testing is sufficient here anyway.

MistakeNot4892:
Polaris has had it live for a few 10+ rounds for what that's worth as data.

PR #2723, opened 2022-10-09
Add Javascript-free darkmode

out-of-phaze:
## Description of changes
Adds darkmode to both the chat and skin, using only winset and CSS.
Adds a pref to toggle it. Toggling it clears your chat so that the changes take effect instantly. (Alternatively, I could make it so that it'd only take effect on the next login, but that's kind of annoying.)

it's a good feature to have so that my eyes don't melt when I'm testing something at 3am and the light of God hits my retinas at full force

## Authorship
Me for code; Baystation12 for certain CSS color values and the basic darkmode code.

## Changelog
:cl:
add: Added dark mode to the UI, which can be toggled in global preferences and is on by default.
/:cl:


code/modules/client/darkmode.css
```diff
@@ -0,0 +1,133 @@
...
+.selenian				{color: #22228b;}
+
+.interface				{color: #750e75;}
+
+.good					{color: #4f7529; font-weight: bold;}
+.bad					{color: #ee0000; font-weight: bold;}
+
+BIG IMG.icon 			{width: 32px; height: 32px;}
+
+/* Codex */
```
MistakeNot4892: Shouldn't need this stuff unless common.css isn't passed to browsers when darkmode is set.

code/modules/client/darkmode.dm
```diff
@@ -0,0 +1,132 @@
+//Darkmode preference by Kmc2000//
+
+/*
+This lets you switch chat themes by using winset and CSS loading, you must relog to see this change (or rebuild your browseroutput datum)
+Things to note:
+If you change ANYTHING in interface/skin.dmf you need to change it here:
```
MistakeNot4892, line 9: Wonder if we can unit test this somehow.

code/modules/client/lightmode.css
```diff
@@ -0,0 +1,129 @@
...
+.selenian				{color: #22228b;}
+
+.interface				{color: #330033;}
+
+.good					{color: #00ff00; font-weight: bold;}
+.bad					{color: #ee0000; font-weight: bold;}
+
+BIG IMG.icon 			{width: 32px; height: 32px;}
+
+/* Codex */
```
MistakeNot4892: As above, looks like you don't need the codex classes.

code/modules/client/darkmode.dm
```diff
@@ -0,0 +1,132 @@
...
+Format:
+winset(src, "window as appears in skin.dmf after elem", "var to change = currentvalue;var to change = desired value")
+*/
+
+/client/proc/deactivate_darkmode(clear_chat = TRUE) //There's no way round it. We're essentially changing the skin by hand. It's painful but it works, and is the way Lummox suggested.
+	//Main windows
+	winset(src, "infowindow", "background-color = [COLOR_DARKMODE_DARKBACKGROUND];background-color = none")
```
MistakeNot4892, line 16: Not going to request this for your PR, but I feel like we could convert a lot of this into a set of static lists categorized by winset field and arguments.

code/modules/client/lightmode.css
out-of-phaze: This was just copied from client/script, so yeah.

MistakeNot4892:
Per convo on Discord, there's a couple of unstyled strings sent in client/New prior to CSS being sent.

MistakeNot4892 (approved):

PR #2713, opened 2022-10-05
Adding a unit test for guns due to busted firemod init.

MistakeNot4892:
## Description of changes
- Adds a unit test for gun init.
- Will fix things that break, currently opening for CI.

## Why and what will this PR improve
Guns bork.

## Authorship
Myself.

## Changelog
Nothing player facing.

PR #2725, opened 2022-10-11
Various path/type cleanups. [MDB IGNORE] [IDB IGNORE]

MistakeNot4892:
## Description of changes
Sorts out out the various exception types in #2575 as best as possible.
Implements abstract type checking on /datum and adds a var to prevent manual/unit test spawning of types.

Safer init, more systematic handling of abstract objects.

## Authorship
Myself.

## Changelog
Nothing player-facing.


MistakeNot4892:
Has dependencies:
~~#2726~~
~~#2720~~
~~#2727~~
~~#2713~~

code/unit_tests/items.dm
```diff
@@ -0,0 +1,63 @@
...
+		volatile_tests += new test_path(src)
+
+	//Run tests on each objects
+	for(var/objpath in obj_test_instances)
+		//Run constant tests first
+		for(var/datum/item_unit_test/T in constant_tests)
+			T.run_test(obj_test_instances[objpath])
+		//Run volatile tests second
+		for(var/datum/item_unit_test/T in volatile_tests)
+			T.run_test(obj_test_instances[objpath])
```
afterthought2: Probably should at least have them check the return value for pass/fail, even if logging happens in the test itself. Also, are there tests being added with this system?

code/modules/materials/materials_ore.dm
```diff
@@ -113,49 +114,70 @@
...
+	is_spawnable_type = TRUE
 /obj/item/stack/material/ore/neon
 	material = /decl/material/solid/ice/hydrate/neon
+	is_spawnable_type = TRUE
 /obj/item/stack/material/ore/krypton
 	material = /decl/material/solid/ice/hydrate/krypton
+	is_spawnable_type = TRUE
 /obj/item/stack/material/ore/xenon
 	material = /decl/material/solid/ice/hydrate/xenon
+	is_spawnable_type = TRUE
```
afterthought2: You can't do this at /ore level or under another common type?

code/unit_tests/items.dm
MistakeNot4892: No, I'm only interested in whether or not all the types init without runtimes.

code/modules/materials/materials_ore.dm
MistakeNot4892: Now that I check it looks like `/obj/item/stack/material/ore` has a default material, so should be fine to move it down to that.

code/unit_tests/items.dm
MistakeNot4892: Happy to leave the individual test failure stuff to Psy_Commando's PR as it's not really needed here.

afterthought2: If the idea is that the mini-tests modify the big test's `failures` list themselves (not how I'd do it but sure, fine), document that expectation below.

MistakeNot4892: Made failure reporting explicit.

afterthought2 (approved):

PR #2745, opened 2022-10-19
Attempting to generalize ambient light.

MistakeNot4892:
## Description of changes
Moves ambient light checking on `Initialize()` down to `/turf`.
Adds an ambient light update to `set_outside()`.
Adds a subsystem to queue and handle ambient lighting updates based on z-level.

This has some performance considerations for `Initialize()` due to the number of turfs, so if anyone has any suggestions for the following, I'm all ears:
- How to gate the proc call to avoid proc overhead.
- How to streamline z-level ambience to not require a list lookup/istype for exoplanets.

Testing on Exodus hasn't shown a variance in init time, but the ambient light procs do profile very highly for self CPU.

Currently non-space and non-exterior turfs don't handle any kind of ambient light (exoplanet or starlight). Building on an exoplanet or looking at the highest z-level of the tradeship results in black holes.

## Authorship
Myself.

## Changelog
:cl:
tweak: Ambient light will now apply to simulated turfs.
/:cl:

### DIFF
--- a/code/__defines/subsystem-priority.dm
+++ b/code/__defines/subsystem-priority.dm
@@ -20,6 +20,7 @@
 #define SS_PRIORITY_SPACEDRIFT     40  // Drifting things.
 #define SS_PRIORITY_INPUT          20  // Input things.
 #define SS_PRIORITY_ICON_UPDATE    20  // Queued icon updates. Mostly used by APCs and tables.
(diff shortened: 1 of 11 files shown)


MistakeNot4892:
Initializing a 255x255x2 map with global ambient lighting takes <20s on my machine currently, I think the performance impacts of this are not significant.

afterthought2 (approved):

PR #2756, opened 2022-10-22
Nullchecks/typechecks status marker updates due to runtime on scav test.

MistakeNot4892:
Unclear on what caused this on Scav, but it was a pretty severe problem. I suspect it's a hard del or a qdel not removing a marker from the global.


MistakeNot4892:
Chucked the `let` to `var` change on here since that slipped to scav without hitting staging. Makes it slightly more compatible with old versions of IE.

comma (approved):

PR #2757, opened 2022-10-23
October Downstream Sync - Glitch City RP

out-of-phaze:
## Description of changes
Contains a number of fixes and helper/observation additions made downstream that I'm too lazy to bundle into separate PRs.
Spoooooky PR! Will be undrafted on Halloween, probably.

- Adds an `is_dense_assoc` helper for checking if something is an associative list with no null values + at least one value. This is mainly useful for things like checking if a list has already been converted to a typecache, for example.
- Adds a helper for outputting human-readable durations, like "10 hours, 3 minutes, and 59 seconds".
- Adds four observations:
    - 'Atom Examined', for when the registered atom is examined by a mob.
    - 'Mob Examining', for when the registered mob examines an atom.
    - 'Entering', a counterpart to the Entered observation but fired on the movable doing the entering, instead of the entered atom.
    - 'Ingested', for when the registered living mob ingests reagents.
- Adds flavored syrups to the flask reagent exception list.
- Fixes a runtime when `glass2` drinking glasses are examined.
- Fixes fruits and other foods losing name changes made after Initialize, i.e. "apple" becoming "fruit".

## Authorship
Me.

### DIFF
--- a/code/_helpers/lists.dm
+++ b/code/_helpers/lists.dm
@@ -775,7 +775,11 @@ var/global/list/json_cache = list()
 			if(islist(decoded)) // To prevent cache mutation.
 				return deepCopyList(decoded)
 			else if(decoded)
-				return decoded	
+				return decoded
 		catch(var/exception/e)
 			log_error("Exception during JSON decoding ([json_to_decode]): [e]")
 	return list()
+
+/// Is this a dense (all keys have non-null values) associative list with at least one entry?
+/proc/is_dense_assoc(var/list/L)
+	return length(L) > 0 && !isnull(L[L[1]])
--- a/code/datums/observation/examine.dm
+++ b/code/datums/observation/examine.dm
@@ -1,0 +1,25 @@
+//	Observer Pattern Implementation: Atom Examined
+//		Registration type: /atom
+//
+//		Raised when: A mob examines an atom.
+//
+//		Arguments that the called proc should expect:
+//			/mob/examiner:  The mob that examined the atom.
+//			/atom/examined: The examined atom.
+
+/decl/observ/atom_examined
+	name = "Atom Examined"
+	expected_type = /atom
+
+//	Observer Pattern Implementation: Mob Examining
+//		Registration type: /mob
+//
+//		Raised when: A mob examines an atom.
+//
+//		Arguments that the called proc should expect:
+//			/atom/examined: The examined atom.
+//			/mob/examiner:  The mob that examined the atom.
+
+/decl/observ/mob_examining
+	name = "Mob Examining"
+	expected_type = /mob
--- a/code/modules/reagents/reagent_containers/food/lunch.dm
+++ b/code/modules/reagents/reagent_containers/food/lunch.dm
@@ -60,7 +60,12 @@ var/global/list/lunchables_drink_reagents_ = list(
 											/decl/material/liquid/drink/dry_ramen,
 											/decl/material/liquid/drink/hell_ramen,
 											/decl/material/liquid/drink/hot_ramen,
-											/decl/material/liquid/drink/mutagencola
+											/decl/material/liquid/drink/mutagencola,
+											/decl/material/liquid/drink/syrup,
+											/decl/material/liquid/drink/syrup/mint,
+											/decl/material/liquid/drink/syrup/chocolate,
+											/decl/material/liquid/drink/syrup/caramel,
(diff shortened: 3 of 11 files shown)


code/_helpers/lists.dm
```diff
@@ -775,7 +775,11 @@ var/global/list/json_cache = list()
...
 			else if(decoded)
-				return decoded	
+				return decoded
 		catch(var/exception/e)
 			log_error("Exception during JSON decoding ([json_to_decode]): [e]")
 	return list()
+
+/// Is this a dense (all keys have non-null values) associative list with at least one entry?
+/proc/is_dense_assoc(var/list/L)
+	return length(L) > 1 && !isnull(L[L[1]])
```
MistakeNot4892: `>=`?

code/datums/observation/examine.dm
```diff
@@ -0,0 +1,25 @@
+//	Observer Pattern Implementation: Atom Examined
+//		Registration type: /atom
+//
+//		Raised when: A mob examines an atom.
+//
+//		Arguments that the called proc should expect:
+//			/mob/examiner:  The mob that examined the atom.
+//			/atom/examined: The examined atom.
+
+/decl/observ/atom_examined
```
MistakeNot4892, line 10: What is the expected use case of these versus just overriding examine() on the atom?

code/modules/reagents/reagent_containers/food/lunch.dm
```diff
@@ -60,7 +60,12 @@ var/global/list/lunchables_drink_reagents_ = list(
 											/decl/material/liquid/drink/dry_ramen,
 											/decl/material/liquid/drink/hell_ramen,
 											/decl/material/liquid/drink/hot_ramen,
-											/decl/material/liquid/drink/mutagencola
+											/decl/material/liquid/drink/mutagencola,
+											/decl/material/liquid/drink/syrup,
+											/decl/material/liquid/drink/syrup/mint,
+											/decl/material/liquid/drink/syrup/chocolate,
```
MistakeNot4892, line 66: Fond memories of being sent to school with a lunchbox full of chocolate syrup

code/datums/observation/examine.dm
afterthought2: Presumably for e.g. contained atoms to subscribe to and append messages when the loc is examined, regardless of what loc is.

code/datums/observation/entered.dm
```diff
@@ -13,10 +13,26 @@
...
+	expected_type = /atom/movable
+
+/****************************
+* Entered/Entering Handling *
+****************************/
 
 /atom/Entered(atom/movable/enterer, atom/old_loc)
 	..()
 	events_repository.raise_event(/decl/observ/entered, src, enterer, old_loc)
+	events_repository.raise_event(/decl/observ/entering, enterer, src, old_loc)
```
afterthought2: You may want to include an extra trigger in `/movable/forceMove` if the destination is `null`, for better overall accuracy and consistency with the moved event. The entered event does not do this mostly because having null as the source is technically problematic (or at any rate inefficient).

code/_helpers/lists.dm
out-of-phaze: Oh shit lol

code/datums/observation/examine.dm
out-of-phaze: When it's "examine this specific item," not "examine an atom of this type." The use-case downstream is for an impulse to examine a certain thing, which can be any structure (chosen just for practical reasons given how it uses locate(), but I'm considering refactoring it to allow for any atom) and not just one particular type. Adding it to /obj/structure/examine or /atom/examine feels silly, and isn't a pattern I like.

code/datums/observation/entered.dm
out-of-phaze: Oh, good point. I've set a reminder to do that tomorrow.

out-of-phaze: I actually just completely removed it, since it's redundant.

code/_helpers/lists.dm
MistakeNot4892: Looks like this still needs to be addressed.

out-of-phaze: Went with > 0 instead.

MistakeNot4892 (approved):

PR #2759, opened 2022-10-24
medical scan table chart fix

Astolaf:
## Description of changes
"Analyze human health" option for mob is a bit broken.
Fixing  table chart 

Before
image
After
image


MistakeNot4892:
Could you point this one at `stable`? The bug is present there too.

Astolaf:
I'll redo it

PR #2761, opened 2022-10-25
Removed unused mule crate, added animal crate.

MistakeNot4892:
- Swaps out some live cargo decls to use an animal crate subtype, adds some mapped subtypes.
- Removes a now-unused mule crate.


afterthought2 (commented):
I'm actually super confused; doesn't this produce two animals, one from the supply decl and one from the crate's Initialize?

MistakeNot4892:
Animal crate defaults to not having an animal type set, this is the type that cargo uses. The ones with the type set are for mapping/spawning.

afterthought2 (approved):
Oh. 

Is the mule crate icon worth reusing for something? No idea what it looks like.

PR #2763, opened 2022-10-25
Adding clotting agent.

MistakeNot4892:
## Description of changes
Adds a new chemical, clotting agent, based on the clotting agents from Polaris.

Currently the only med with this effect is overdosing on bicaridine, which has its own issues.

## Authorship
Myself for the code, Polaris for the idea.

## Changelog
:cl:
add: Added a clotting agent medication.
/:cl:

### DIFF
--- a/code/modules/reagents/chems/chems_medicines.dm
+++ b/code/modules/reagents/chems/chems_medicines.dm
@@ -360,6 +360,41 @@
 	M.add_chemical_effect(CE_OXYGENATED, 1)
 	holder.remove_reagent(/decl/material/gas/carbon_monoxide, 2 * removed)
 
+/decl/material/liquid/clotting_agent
+	name = "clotting agent"
+	uid = "chem_clotting_agent"
+	lore_text = "A medication used to rapidly clot internal hemorrhages by increasing the effectiveness of platelets."
+	metabolism = REM * 0.5
+	overdose = REAGENTS_OVERDOSE * 0.5
+	color = "#4246c7"
+	scannable = TRUE
+
+/decl/material/liquid/clotting_agent/affect_blood(mob/living/M, removed, datum/reagents/holder)
+	SET_STATUS_MAX(M, STAT_BLURRY, 30)
+	M.add_chemical_effect(CE_BLOCKAGE, (15 + REAGENT_VOLUME(holder, type))/100)
+	for(var/obj/item/organ/external/E in M.get_external_organs())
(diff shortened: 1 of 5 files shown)


code/modules/reagents/chems/chems_medicines.dm
```diff
@@ -359,3 +359,38 @@
...
+	uid = "chem_clotting_agent"
+	lore_text = "A medication used to rapidly clot internal hemorrhages by increasing the effectiveness of platelets."
+	metabolism = REM * 0.5
+	overdose = REAGENTS_OVERDOSE * 0.5
+	color = "#4246c7"
+	scannable = TRUE
+
+/decl/material/liquid/clotting_agent/affect_blood(mob/living/M, removed, datum/reagents/holder)
+	SET_STATUS_MAX(M, STAT_BLURRY, 30)
+	M.add_chemical_effect(CE_BLOCKAGE, (15 + REAGENT_VOLUME(holder, type))/100)
```
Vizoplez, line 374: This seems way too stronk, especially since all three reagents needed to make it are pretty common.

If I am reading this correct, this only blurs your vision, yet fixes artery cuts and organ bleeds (+ external cutting) atop of adding the CE_BLOCKAGE status effect.

Pretty godtier.

MistakeNot4892: Carpotoxin isn't particularly common.
Willing to look at reworking this if you have thoughts on balance for it.

MistakeNot4892: For clarity's sake, the CE_BLOCKAGE effect is detrimental.

code/modules/reagents/chems/chems_medicines.dm
```diff
@@ -359,3 +359,38 @@
...
+	for(var/obj/item/organ/external/E in M.get_external_organs())
```
comma, line 375: Maybe only check `bad_external_organs` so it doesn't loop over all organs for victim? Bleedings should put organ in that list

out-of-phaze (dismissed):
Was going to comment on how you could just use the same value for both in populate_reagents, but then I noticed the rounding and realized that means a 15u autoinjector would have 8u of clotting agent and 7u of stabilizer, so you have to do it this way.

Other than that, nothing jumps out at me as something to note or change. Looks fine to me.

code/modules/reagents/chems/chems_medicines.dm
MistakeNot4892: Done.

out-of-phaze:
You seem to have added power tools by accident?

afterthought2 (approved):
Can tune balance after playtest feedback.

PR #2764, opened 2022-10-25
Ported chess toys from Polaris.

MistakeNot4892:
## Description of changes
Ports chess props from Polaris, used for map-scale chess games.

image

I'm picking apart another large Polaris feature parity PR and this is part of it. I'm not actually sure if this should go into dev or not. It's fun for mapping but that's about it.

## Authorship
Original is from Polaris, not sure if it originated there.
New icons are from Doe, who is very good.

## Changelog
Nothing player-facing.


afterthought2 (commented):
Possibly "duplicating" `/obj/item/chems/food/checker/pawn` and other pieces.

MistakeNot4892:
Different use-case, these are like, a foot tall. Not chessboard pieces.

out-of-phaze (dismissed):
Looks good, but I have a question: why have separate blue/purple sprites instead of just using the `color` var?

MistakeNot4892:
Because those are the sprites that were on Polaris and this is a minimum effort PR.

Doe is actually redoing them to be isometric and gresycaled - if that goes ahead I'll gank them to replace these.

out-of-phaze (dismissed):
Neat. I like the new sprites even though the knight looks a little scale-y.

code/game/objects/items/toys.dm
```diff
@@ -768,3 +768,95 @@
...
+
+//////////////////////////////////////////////////////
+//					Chess Pieces					//
+//////////////////////////////////////////////////////
+
+/obj/item/toy/chess
+	name = "chess piece"
+	desc = "This should never display."
+	icon = 'icons/obj/items/chess.dmi'
+	w_class = ITEM_SIZE_SMALL
```
afterthought2: I don't actually know how obvious it is that these are big and the others are small; there are lots of not-to-scale sprites. I've never seen the others in game so could be off the mark, but would argue that either (a) replace the others or (b) make it more transparent that these are big via item var values and desc.

out-of-phaze: Isn't there a long-term goal of making world sprites to-scale? That's why the icon system is the way it is.
That said, updating the w_class value might be good, yes.

PlayerDeer:
> Neat. I like the new sprites even though the knight looks a little scale-y.

Thanks! I do see what you mean... If I can think of a way to make it look better I'll open a new PR for the change! Thanks for the feedback.

code/game/objects/items/toys.dm
MistakeNot4892: Renamed them to 'oversized'. A more detailed integration can happen in another PR, I just want these for mapping compatibility.

afterthought2 (approved):

PR #2769, opened 2022-10-25
Added ERT and ninja PDA subtypes.

MistakeNot4892:
## Description of changes
Adds some colour-customised subtype PDAs for special roles.

Special roles have more distinct gear.

## Authorship
Myself.

## Changelog
:cl:
tweak: Antag PDAs have been customised.
/:cl:


Vizoplez:
This has massive balance implications when and if we re-add NTNet hacking, since the AI/other roles can possibly trace the ninja via their PDA ID.

MistakeNot4892:
I'm not sure that I agree it's a massive balance consideration. Offmap PDAs also won't be part of the station network anyway as far as I know.

afterthought2 (approved):

PR #2772, opened 2022-10-25
Rewriting tank dispensers.

MistakeNot4892:
## Description of changes
Rewrites some crufty old dispenser code and makes them craftable.

Better code, more crafting options.

## Authorship
Myself.

## Changelog
:cl:
add: Tank dispensers are now craftable.
/:cl:

### DIFF
--- a/code/game/objects/structures/tank_dispenser.dm
+++ b/code/game/objects/structures/tank_dispenser.dm
@@ -43,7 +43,7 @@
 /obj/structure/tank_rack/dump_contents()
 	hydrogen_tanks = null
 	oxygen_tanks = null
-	. = ..()
+	return ..()
 
 /obj/structure/tank_rack/on_update_icon()
 	..()
@@ -62,10 +62,9 @@
 		if(5 to INFINITY)
 			add_overlay("hydrogen-5")
 
-/obj/structure/tank_rack/attack_ai(mob/living/silicon/ai/user)
-	if(user.Adjacent(src))
+/obj/structure/tank_rack/attack_robot(mob/user)
+	if(CanPhysicallyInteract(user))
 		return attack_hand(user)
-	..()
 
 /obj/structure/tank_rack/attack_hand(mob/user)
 	var/list/dat = list()
@@ -81,18 +80,21 @@
 	if(istype(I, /obj/item/tank))
 		var/list/adding_to_list
 		if(istype(I, /obj/item/tank/oxygen) || istype(I, /obj/item/tank/air))
+			LAZYINITLIST(oxygen_tanks)
 			adding_to_list = oxygen_tanks
 		else if(istype(I, /obj/item/tank/hydrogen))
+			LAZYINITLIST(hydrogen_tanks)
 			adding_to_list = hydrogen_tanks
 		if(LAZYLEN(adding_to_list) >= 10)
 			to_chat(user, SPAN_WARNING("\The [src] is full."))
+			UNSETEMPTY(adding_to_list)
 			return TRUE
 		if(!user.unEquip(I, src))
 			return TRUE
 		LAZYADD(adding_to_list, weakref(I))
 		to_chat(user, SPAN_NOTICE("You put [I] in [src]."))
-		if(LAZYLEN(adding_to_list) < 5)
-			update_icon()
+		update_icon()
+		attack_hand(user)
 		return TRUE
 	return ..()
 
--- a/code/game/objects/structures/tank_dispenser.dm
+++ b/code/game/objects/structures/tank_dispenser.dm
@@ -1,117 +1,131 @@
-/obj/structure/dispenser
-	name = "tank storage unit"
+/obj/structure/tank_rack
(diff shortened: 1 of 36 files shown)


code/game/objects/structures/tank_dispenser.dm
```diff
@@ -1,117 +1,135 @@
...
+	return ..()
+
+/obj/structure/tank_dispenser/Topic(href, href_list)
+
+	. = ..()
+	if(. || !usr || usr.incapacitated() || usr.restrained())
+		return
+
+	if(!Adjacent(usr))
 		close_browser(usr, "window=dispenser")
```
afterthought2: Check the topic state, use OnTopic, at least something along those lines.

code/game/objects/structures/tank_dispenser.dm
```diff
@@ -1,117 +1,135 @@
...
+		var/obj/item/tank/H = hydrogen_tanks[1]
+		LAZYREMOVE(hydrogen_tanks, H)
+		H.dropInto(loc)
+		to_chat(usr, SPAN_NOTICE("You take \the [H] out of \the [src]."))
+		. = TOPIC_REFRESH
+
+	if(. == TOPIC_REFRESH)
+		usr.set_machine(src)
+		update_icon()
+		add_fingerprint(usr)
```
afterthought2: Don't add fingerprints. Setting machine here is a giant hack but apparently I'm in a forgiving mood when it comes to the machine var at the moment.

code/game/objects/structures/tank_dispenser.dm
```diff
@@ -1,117 +1,135 @@
...
+		for(var/i in 1 to spawn_oxy)
+			oxygen_tanks += new /obj/item/tank/oxygen(src)
+	else
+		oxygen_tanks = null
+
+	if(isnum(hydrogen_tanks) && hydrogen_tanks > 0)
+		var/spawn_hyd = hydrogen_tanks
+		hydrogen_tanks = list()
+		for(var/i in 1 to spawn_hyd)
+			hydrogen_tanks += new /obj/item/tank/hydrogen(src)
```
afterthought2: I wouldn't complain about this if it was oldcode, but you actually changed it to work like this: do we really need to spawn 20 tanks whenever one of these is mapped in or created?

code/game/objects/structures/tank_dispenser.dm
```diff
@@ -1,117 +1,135 @@
...
+	switch(hydrocount)
 		if(1 to 4)
-			add_overlay("hydrogen-[hydrogentanks]")
+			add_overlay("hydrogen-[hydrocount]")
 		if(5 to INFINITY)
 			add_overlay("hydrogen-5")
 
-/obj/structure/dispenser/attack_ai(mob/living/silicon/ai/user)
+/obj/structure/tank_dispenser/attack_ai(mob/living/silicon/ai/user)
 	if(user.Adjacent(src))
```
afterthought2: This check is unreasonable, use topic states for sane interactivity checks.

afterthought2 (commented):
- It should really delete all of its tanks in Destroy. 
- Please double-check that in the physical destroy proc it doesn't dump its tanks while retaining refs to them.
- If the physical destroy handling involves dumping tanks, suggest reconsidering to destroy them instead for performance reasons.
- If it's buildable it needs to be deconstructable; maybe it is due to parent code but can't tell.

code/game/objects/structures/tank_dispenser.dm
MistakeNot4892: I would argue yes, if we're intending down the track to do physics sim etc. We spawn the contents of crates and lockers, we spawn organs, we populate with chems. I don't think tanks are that much more of an ask.

MistakeNot4892: I was thinking of dropping the tank count though, I've never seen a tank dispenser get emptied in a round.

code/game/objects/structures/tank_dispenser.dm
MistakeNot4892: I might just completely rewrite the interaction and topic handling to be honest.

code/game/objects/structures/tank_dispenser.dm
MistakeNot4892: Done.

code/game/objects/structures/tank_dispenser.dm
MistakeNot4892: Removed.

code/game/objects/structures/tank_dispenser.dm
MistakeNot4892: Tank count has been dropped to six of each. Willing to discuss this further.

code/game/objects/structures/tank_dispenser.dm
MistakeNot4892: Done.

afterthought2 (approved):

PR #2780, opened 2022-10-25
Broke random.dm up into a subtypes folder, added a bunch of new random types.

MistakeNot4892:
## Description of changes
- Breaks random.dm up into a subtype folder.
- Ported a bunch of maintenance and miscellaneous random spawns from Polaris.

Easier to read and update the folder, more stuff for mappers.

## Authorship
Polaris for the maintenance and misc random spawners.

## Changelog
Nothing player-facing.


code/game/objects/random/subtypes/closets.dm
```diff
@@ -0,0 +1,28 @@
...
+	icon = 'icons/obj/closets/bases/closet.dmi'
+	icon_state = "base"
+	var/vermin_chance = 0.1
+	var/list/locker_vermin = list(
+		/mob/living/simple_animal/mouse,
+		/mob/living/simple_animal/opossum
+	)
+
+/obj/random/closet/spawn_choices()
+	return typesof(/obj/structure/closet) - (typesof(/obj/structure/closet/crate) | typesof(/obj/structure/closet/body_bag) | typesof(/obj/structure/closet/secure_closet))
```
out-of-phaze: This could be a nontrivial amount of init time. Maybe make this a static list? Something like:
```suggestion
	var/static/list/spawn_types = typesof(/obj/structure/closet) - (typesof(/obj/structure/closet/crate) | typesof(/obj/structure/closet/body_bag) | typesof(/obj/structure/closet/secure_closet))
	return spawn_types
```

code/game/objects/random/subtypes/closets.dm
```diff
@@ -0,0 +1,28 @@
...
+		new vermin_type(.)
+
+/obj/random/crate
+	name = "random crate"
+	desc = "This is a random crate"
+	icon = 'icons/obj/closets/bases/crate.dmi'
+	icon_state = "base"
+
+/obj/random/crate/spawn_choices()
+	. = typesof(/obj/structure/closet/crate) - typesof(/obj/structure/closet/crate/secure)
```
out-of-phaze: Likewise with above; if we're going to be repeatedly generating these lists deterministically, please cache them with static proc vars.

code/game/objects/random/subtypes/clothing.dm
```diff
@@ -0,0 +1,246 @@
+
+/obj/random/masks
+	name = "random mask"
+	desc = "This is a random face mask."
+	icon = 'icons/clothing/mask/gas_mask.dmi'
+	icon_state = ICON_STATE_WORLD
+
+/obj/random/masks/spawn_choices()
+	return list(
```
out-of-phaze: Again, this could be made a static var in this proc and only allocated once, saving init time and memory.

code/game/objects/random/subtypes/plants.dm
```diff
@@ -0,0 +1,24 @@
...
+		/obj/structure/flora/seaweed/glow =  1
+	)
+
+/obj/random/pottedplant
+	name = "random potted plant"
+	desc = "This is a random potted plant."
+	icon = 'icons/obj/structures/potted_plants.dmi'
+	icon_state = "plant-26"
+	spawn_nothing_percentage = 0
+	var/static/list/blacklisted_plants = list(/obj/structure/flora/pottedplant/unusual)
```
out-of-phaze: Do note that this means you won't be able to modify this on subtypes. It might be better to make this a proc like `/obj/random/pottedplant/proc/blacklisted_plants()`, so that it can be overridden.

out-of-phaze (changes requested):
I'd really like it if all these deterministic `spawn_choices` overrides were made to cache their results in a static var, so that we don't have to repeatedly re-allocate/re-generate them, especially the large lists and the ones that use `typesof()` and list operators.

code/game/objects/random/subtypes/plants.dm
MistakeNot4892: I get the difference but it tickles me a bit that you recommend static vars half a dozen times then recommend the exact opposite here.

out-of-phaze: Proc-level static vars can be overridden by overriding the proc.

MistakeNot4892:
They all use static local lists now.

out-of-phaze (dismissed):
Thanks! With large maps, that can actually be a decent source of init time (at least by TG's standards).

afterthought2:
Appears that there are remaining artifacts from the merge conflicts; does not compile.

MistakeNot4892:
Painful. I compile tested it locally, not sure what would have changed.

code/game/objects/random/subtypes/plants.dm
MistakeNot4892: This turned out to break world init due to static vars being initialized in world/New before the instance existed to retrieve this list. People can move them onto globals or something down the track if they want to override it.

afterthought2 (approved):

PR #2781, opened 2022-10-25
Expanding on the SMES setup logic.

MistakeNot4892:
Some minor tweaks and an additional landmark ported from the Polaris engine setup logic.

### DIFF
--- a/code/modules/supermatter/setup_supermatter.dm
+++ b/code/modules/supermatter/setup_supermatter.dm
@@ -195,8 +195,16 @@ var/global/list/engine_setup_markers = list()
 
 
 // Tries to enable the SMES on max input/output settings. With load balancing it should be fine as long as engine outputs at least ~500kW
-/obj/effect/engine_setup/smes/
+/obj/effect/engine_setup/smes
 	name = "SMES Marker"
+
+	var/target_input_level		//These are in watts, the display is in kilowatts. Add three zeros to the value you want.
+	var/target_output_level		//These are in watts, the display is in kilowatts. Add three zeros to the value you want.
+
+/obj/effect/engine_setup/smes/main
+	name = "Main SMES Marker"
+	target_input_level =  INFINITY
+	target_output_level = INFINITY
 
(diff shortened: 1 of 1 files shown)


code/modules/supermatter/setup_supermatter.dm
```diff
@@ -206,13 +214,24 @@ var/global/list/engine_setup_markers = list()
...
 	S.output_attempt = 1
-	S.input_level = S.input_level_max
-	S.output_level = S.output_level_max
+	if(target_input_level)
+		if(target_input_level > S.input_level_max)
+			S.input_level = S.input_level_max
+		else
+			S.input_level = target_input_level
+	else
+		S.input_level = S.input_level_max
```
afterthought2: How about `S.input_level = min(target_input_level, S.input_level_max)` and default `target_input_level` to infinity?

code/modules/supermatter/setup_supermatter.dm
```diff
@@ -247,6 +266,27 @@ var/global/list/engine_setup_markers = list()
 	F.update_use_power(POWER_USE_IDLE)
 	return SETUP_OK
 
+// Closes the monitoring room shutters so the first Engi to show up doesn't get microwaved
```
afterthought2, line 256: I don't really get it; can't you just map them closed?

MistakeNot4892: I kind of mindlessly copied this over from Polaris when converting the map, this seems like a valid point.

MistakeNot4892: Thinking about it - maybe you want closing the shutters to be part of the routine, I guess. I don't think it causes any harm to include this landmark.

afterthought2 (approved):

code/modules/supermatter/setup_supermatter.dm
afterthought2: I mean, it doesn't even close them, it toggles them. But OK.

PR #2786, opened 2022-10-28
Reworks handle_message_mode

MistakeNot4892:
## Description of changes
- Moves the main body of `handle_message_mode()` down to `/mob/living` other than `whisper_say()`.
- Renames the radio version of the proc to `get_connection_from_message_mode()` for clarity/searchability.
- Migrates `GetRadio()` to `/atom/proc/get_radio()` next to `get_cell()`, overrides it on radios to return the radio and reworks `handle_message_mode` to be consistent in how it's called.

Fixes an issue Andrew identified with use_mode/message_mode, fixes unintended clobbering of the GetRadio() changes to handle_message_mode, integrates `get_radio()` in a way that will make the extension more or less drag and drop downstream if desired (such as on Faz-World).

## Authorship
@Andrew-Fall for spotting the use_mode issue.

## Changelog
Nothing player-facing.

### DIFF
--- a/code/game/atoms.dm
+++ b/code/game/atoms.dm
@@ -494,6 +494,13 @@
 /atom/proc/get_cell()
 	return
 
+// This proc will retrieve any radios associated with this atom,
+// for use in handle_message_mode or other radio-based logic.
+// The message_mode argument is used to determine what subset of
+// radios are relevant to the current call (ie. intercoms or ear radios)
+/atom/proc/get_radio(var/message_mode)
+	return
+
 /atom/proc/building_cost()
 	. = list()
 
--- a/code/modules/mob/living/carbon/human/say.dm
+++ b/code/modules/mob/living/carbon/human/say.dm
@@ -114,46 +114,10 @@
 	return ..(message_data)
 
 /mob/living/carbon/human/handle_message_mode(message_mode, message, verb, speaking, used_radios, alt_name)
-	var/use_mode = null
-	switch(message_mode)
-
-		if("intercom")
-			if(!restrained())
-				for(var/obj/item/radio/I in view(1))
-					if(I.intercom_handling)
-						used_radios += I
-
-		if("right ear", "left ear")
-			var/use_right = message_mode == "right ear"
-			var/obj/item/radio/R = get_equipped_item(use_right ? slot_r_ear_str : slot_l_ear_str)
-			if(!istype(R))
-				R = null
-				var/datum/inventory_slot/inv_slot = LAZYACCESS(held_item_slots, (use_right ? BP_R_HAND : BP_L_HAND))
-				if(istype(inv_slot?.holding, /obj/item/radio))
-					R = inv_slot.holding
-			if(R)
-				used_radios += R
-
-		if("whisper") //It's going to get sanitized again immediately, so decode.
-			whisper_say(html_decode(message), speaking, alt_name)
-			return 1
-
-		else
-			// Headsets are default.
-			if(message_mode)
-				var/obj/item/radio/R
-				for(var/slot in global.ear_slots)
-					R = get_equipped_item(slot)
-					if(istype(R))
-						break
-				if(!istype(R))
-					R = GetRadio()
-				if(istype(R))
-					used_radios += R
-
-	for(var/obj/item/radio in used_radios)
-		radio.add_fingerprint(src)
-		radio.talk_into(src,message,use_mode,verb,speaking)
+	if(message_mode == "whisper") //It's going to get sanitized again immediately, so decode.
+		whisper_say(html_decode(message), speaking, alt_name)
+		return TRUE
+	return ..()
 
 /mob/living/carbon/human/handle_speech_sound()
 	if(species.speech_sounds && prob(species.speech_chance))
--- a/code/modules/mob/living/say.dm
+++ b/code/modules/mob/living/say.dm
@@ -139,13 +139,49 @@ var/global/list/channel_to_radio_key = new
 	message_data[1] = message
 	message_data[2] = verb
 
+// Grabs any radios equipped to the mob, with message_mode used to
+// determine relevancy. See handle_message_mode below.
+/mob/living/proc/get_radios(var/message_mode)
+
+	if(!message_mode)
+		return
+
+	var/list/possible_radios
+	if(message_mode == "right ear" || message_mode == "left ear")
+		var/use_right = message_mode == "right ear"
+		var/obj/item/thing = get_equipped_item(use_right ? slot_r_ear_str : slot_l_ear_str)
+		if(thing)
+			LAZYDISTINCTADD(possible_radios, thing)
+		else
+			thing = get_equipped_item(use_right ? BP_R_HAND : BP_L_HAND)
+			if(thing)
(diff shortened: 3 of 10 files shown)


MistakeNot4892:
Bugfixes don't need to be frozen ftr

code/game/atoms.dm
```diff
@@ -493,6 +493,9 @@
 /atom/proc/get_cell()
 	return
 
+/atom/proc/get_radio(var/message_mode)
```
out-of-phaze, line 501: I'd suggest a small comment explaining this proc, since I was originally going to suggest moving the `"left ear"` etc. logic in handle_message_mode to here, to 'get' a mob's radio, but I'm pretty sure this proc is supposed to be the equivalent to 'casting' the atom to a radio, i.e. "this atom *is* a radio"?

out-of-phaze (dismissed):
I probably would have written it differently, but not in any way that makes a difference.

code/game/atoms.dm
MistakeNot4892: It's meant to retrieve the radio associated with the atom, in the same way that get_cell is supposed to retrieve the cell (and returns `src` on cells themselves). It would make sense for this to be overridden on mobs for use in handle_message_mode, yeah.

code/modules/mob/living/say.dm
```diff
@@ -140,12 +140,40 @@ var/global/list/channel_to_radio_key = new
...
 
 /mob/living/proc/handle_message_mode(message_mode, message, verb, speaking, used_radios, alt_name)
-	if(message_mode == "intercom")
-		for(var/obj/item/radio/I in view(1, null))
-			if(I.intercom_handling)
-				I.talk_into(src, message, verb, speaking)
-				used_radios += I
-	return 0
+
+	// Check the mob's get_radio() in case of inbuilt radios in the future.
```
afterthought2: I don't follow: we aren't checking the mob's radio here as far as I can tell.

code/modules/mob/living/carbon/human/say.dm
```diff
@@ -114,46 +114,10 @@
...
-				if(istype(R))
-					used_radios += R
-
-	for(var/obj/item/radio in used_radios)
-		radio.add_fingerprint(src)
-		radio.talk_into(src,message,use_mode,verb,speaking)
+	if(message_mode == "whisper") //It's going to get sanitized again immediately, so decode.
+		whisper_say(html_decode(message), speaking, alt_name)
+		return TRUE
+	return ..()
```
afterthought2, line 120: Does this proc return things? The parent doesn't seem to.

afterthought2 (commented):
Would recommend commenting on what `handle_message_mode` actually does and what it should returns by the definition.

code/game/atoms.dm
afterthought2: I think the most useful documentation you could provide for this proc is what `message_mode` is, as it seems to be one of several arbitrary string keys. What it does and the return value seems obvious to me.

code/modules/mob/living/say.dm
MistakeNot4892: Referring to the `list(src)` immediately below.

MistakeNot4892: Noting I am going to move all the radio retrieval into mob get_radio()

code/game/atoms.dm
MistakeNot4892: I've added some comments to it.

code/modules/mob/living/say.dm
MistakeNot4892: Comment's gone, logic has moved.

code/modules/mob/living/carbon/human/say.dm
MistakeNot4892: They should return a consistent value for whether or not they handled the message now.

code/modules/mob/living/say.dm
```diff
@@ -139,13 +139,49 @@ var/global/list/channel_to_radio_key = new
...
+			if(istype(radio))
+				LAZYADD(., radio)
+
+// This proc takes in a string (message_mode) which maps to a radio key in global.department_radio_keys
+// It then processes the message_mode to implement an additional behavior needed for the message, such
+// as retrieving radios or looking for an intercom nearby.
 /mob/living/proc/handle_message_mode(message_mode, message, verb, speaking, used_radios, alt_name)
-	if(message_mode == "intercom")
-		for(var/obj/item/radio/I in view(1, null))
+	var/list/assess_items_as_radios = get_radios(message_mode)
```
out-of-phaze, line 175: Much nicer! Now all the radio-getting logic is in one place.

code/modules/mob/living/say.dm
```diff
@@ -139,13 +139,49 @@ var/global/list/channel_to_radio_key = new
...
+	if(message_mode == "intercom" && !restrained())
```
out-of-phaze, line 176: Iiii feel like maybe the usability should be handled on the intercom/radio itself? I don't know about explicitly checking `!restrained()` here instead of doing something like `I.intercom_handling && I.can_use(src)`. On the other hand, that'd be a decent amount more expensive.

Not blocking though, just comments for consideration.

code/modules/mob/living/say.dm
```diff
@@ -139,13 +139,49 @@ var/global/list/channel_to_radio_key = new
...
+		for(var/slot in global.ear_slots)
+			var/thing = get_equipped_item(slot)
+			if(thing)
+				LAZYDISTINCTADD(possible_radios, thing)
+
+	if(length(possible_radios))
+		for(var/atom/movable/thing as anything in possible_radios)
+			var/obj/item/radio/radio = thing.get_radio(message_mode)
+			if(istype(radio))
+				LAZYADD(., radio)
```
out-of-phaze: You use LAZYDISTINCTADD earlier, which is slow, but only LAZYADD here. Is it not possible that get_radio on two of the distinct objects in `possible_radios` might return the same value somehow? This would have all the expense of using LAZYDISTINCTADD but lose the benefits by not doing it here.

Again, not blocking, just a comment.

out-of-phaze (dismissed):
I like the way this is set up now. A few thoughts on potential optional improvements, but it's good as-is. No blocking issues, as far as I can tell.

code/modules/mob/living/say.dm
MistakeNot4892: I supposed it is theoretically possible, though offhand I can't think of what would do that. Maybe some kind of radio backpack with multiple listening points? I'll change that to LAZYDISTINCTADD anyway.

code/modules/mob/living/say.dm
MistakeNot4892: I think I'd like to leave it as-is for now rather than implementing a new radio-level proc for it. The intercom handling in general could be optimized a bit rather than using view() I think.

out-of-phaze (approved):
No additional comments this time, you've addressed everything I can think of and the changes don't seem to have broken anything.

By the way, what are your thoughts on creating unit tests to ensure that radio networks work on maps? 🤔

afterthought2 (approved):

afterthought2:
Reasonable unit tests are always welcome, but seems way out of scope for this PR.

PR #2789, opened 2022-10-29
Fixes for mob offset adjustment procs.

MistakeNot4892:
## Description of changes
- Makes offset procs call ..() last.
- Reworks the bodytype offset proc so it collects overlays in the same icon and preserves state information.
 
Fixes issues with offset icons losing access to icon and icon_state information useful for applying mob overlays.

## Authorship
Myself.

## Changelog
Nothing player-facing.


afterthought2 (approved):
Only have a vague idea of what this does.

MistakeNot4892:
Existing code will generate a single individual icon file with a nameless state for the offset icon. This causes issues with mob overlays that rely on the icon state of the base overlay.

New code generates/inserts a named state to a shared icon and swaps it in for the overlay generation instead.

PR #2790, opened 2022-10-29
medscan chart fix

Astolaf:
"Analyze human health" option for mob is a bit broken.
Fixing table chart
Before

image

After

image


MistakeNot4892 (dismissed):

MistakeNot4892:
This is conflicted - you need to rebase onto the most recent `stable` and resolve them.

Astolaf:
Messing around with git. What about now?

afterthought2 (approved):

PR #2794, opened 2022-11-02
Organ healing.

MistakeNot4892:
## Description of changes
- Livers and brains now share their regeneration behavior with the base internal organ type. In other words, all internal organs will naturally slowly heal damage to the closest threshold based on their vars.
- Since the behavior has been migrated down the chain, a fair bit of other code is now type-agnostic when dealing with the brain.

Still need to test this, tweak numbers, and clean up the redundant checking in the procs.

Minor organ damage being permanent has been an issue for awhile (welding eye damage, for example) and we had three different methods of organs healing themselves. This also clears the field for some changes I want to make in #1447, namely MMIs being an organ subtype.

## Authorship
Myself.

## Changelog
:cl:
tweak: Liver and brain regeneration has been tweaked, and all internal organs will now regenerate slowly to a point.
/:cl:

### DIFF
--- a/code/modules/organs/internal/_internal.dm
+++ b/code/modules/organs/internal/_internal.dm
@@ -3,6 +3,17 @@
 ****************************************************/
 /obj/item/organ/internal
 	scale_max_damage_to_species_health = TRUE
+
+	// Damage healing vars (moved here from brains)
+	/*
+	/// Number of gradiations of damage we can recover in. ie. set to 10, we can recover up to the most recent 10% damage. Leave null to disable regen.
+	var/damage_threshold_count
+	/// Actual amount of health constituting one gradiation.
+	var/damage_threshold_value
+	/// The last gradiation we healed to.
+	var/healed_threshold = 1
+	*/
+
 	var/tmp/alive_icon //icon to use when the organ is alive
 	var/tmp/dead_icon // Icon to use when the organ has died.
 	var/tmp/prosthetic_icon //Icon to use when the organ is robotic
@@ -93,6 +104,8 @@
 	max_damage = FLOOR(ndamage)
 	min_broken_damage = FLOOR(0.75 * max_damage)
 	min_bruised_damage = FLOOR(0.25 * max_damage)
+	if(damage_threshold_count > 0)
+		damage_threshold_value = round(max_damage / damage_threshold_count)
 
 /obj/item/organ/internal/take_general_damage(var/amount, var/silent = FALSE)
 	take_internal_damage(amount, silent)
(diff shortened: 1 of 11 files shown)


code/modules/organs/internal/_internal.dm
```diff
@@ -135,22 +148,45 @@
 	. = "[.][name]"
 
 /obj/item/organ/internal/Process()
+	SHOULD_CALL_PARENT(TRUE)
 	..()
```
PsyCommando, line 153: Shouldn't it be returning the base class return value?
```suggestion
	. = ..()
```

MistakeNot4892: No?

MistakeNot4892: Checked, organ Process() doesn't return a value.

MistakeNot4892:
This needs testing but is otherwise ready for review.

code/modules/organs/internal/_internal.dm
PsyCommando: kk

MistakeNot4892:
Healing isn't stopping when the threshold is hit. Need to fix that.

MistakeNot4892:
Tested, all appears to be working now.

code/modules/organs/internal/_internal.dm
```diff
@@ -135,22 +149,50 @@
...
-	if(damage < 0.1*max_damage)
-		heal_damage(0.1)
+	if(owner && damage && !(status & ORGAN_DEAD))
+		handle_damage_effects()
+
+/obj/item/organ/internal/proc/handle_damage_effects()
+	SHOULD_CALL_PARENT(TRUE)
+	if(organ_can_heal())
+		// Some logic to prevent us healing straight past the threshold just by virtue of the heal amount being larger than the remaining damage in the current threshold.
+		var/min_heal_val = (min_regeneration_cutoff_threshold && !past_damage_threshold(min_regeneration_cutoff_threshold)) ? 0 : (get_current_damage_threshold() * damage_threshold_value)
```
afterthought2: Can we make this more readable? Break out the `&&`, avoid the unnecessary negation in the ternary, etc.

code/modules/organs/internal/_internal.dm
```diff
@@ -135,22 +149,50 @@
...
+	// We cannot regenerate, period.
+	if(!damage_threshold_count || !damage_threshold_value || BP_IS_PROSTHETIC(src))
+		return FALSE
+	// Our owner is under stress.
+	if(owner.get_blood_oxygenation() < BLOOD_VOLUME_SAFE || GET_CHEMICAL_EFFECT(owner, CE_TOXIN) || owner.radiation || owner.is_asystole())
+		return FALSE
+	// If we haven't hit the regeneration cutoff point, heal.
+	if(min_regeneration_cutoff_threshold && !past_damage_threshold(min_regeneration_cutoff_threshold))
+		return TRUE
+	// We have room to heal within this threshold, or we're not past our cutoff threshold, or we've got stabilizer in our system.
```
afterthought2: This isn't accurate, no? The first check is an AND.

afterthought2 (commented):
Not going to lie, I find the logic here pretty confusing. I am going to assume you know what you are doing.

code/modules/organs/internal/_internal.dm
afterthought2: Return the parent value if and only if you want to `PROCESS_KILL` unless parent needs to do stuff, i.e. if you have nothing to do yourself.

code/modules/organs/internal/_internal.dm
MistakeNot4892: What unnecessary negation?

code/modules/organs/internal/_internal.dm
MistakeNot4892: Comment predates that, yeah, I'll update it.

code/modules/organs/internal/_internal.dm
afterthought2: `!A ? B : C` is equivalent to `A ? C : B` which is one less logical operation to parse when reading. But you can write it however you prefer, I just had to read this line like 10 times to understand it.

MistakeNot4892: Done.

code/modules/organs/internal/_internal.dm
MistakeNot4892: Done.

afterthought2 (approved):

PR #2799, opened 2022-11-04
Atom Creation And Destruction/GC Unit Tests + GC/EarlyDestroy Improvements

out-of-phaze:
## Description of changes
Ports Del The World from TG, ~~currently only for items~~ for most obj and mob subtypes. This is basically an expansion of the item creation tests to detect harddels as well as runtimes during init.

Harddel detection and regression testing, what's not to love?

## Authorship
https://github.com/tgstation/tgstation/pull/59612


code/game/objects/items/stacks/stack.dm
```diff
@@ -188,7 +197,7 @@
 		list_recipes(usr, text2num(href_list["sublist"]))
 
 	if (href_list["make"])
-		if (src.get_amount() < 1) qdel(src) //Never should happen
+		delete_if_empty() //Never should happen
```
MistakeNot4892: Should probably return here?

code/unit_tests/items.dm
```diff
@@ -1,21 +1,34 @@
...
+			I = path
 			if(!TYPE_IS_SPAWNABLE(I))
 				continue
-			I = new path
+			I = new path(spawn_loc)
 			if(QDELETED(I))
 				continue
 			obj_test_instances[path] = I
+			// Move ourselves to nullspace to avoid interfering with stuff spawned after.
+			I.forceMove(null)
```
MistakeNot4892: Possibly should move the entire contents of spawn_loc here rather than tidying later.

noelle-lavenza: I don't want errors moving or deleting those to cause the object itself to error; they should runtime outside of the try/catch IMO.

noelle-lavenza: That'd be moving it into the try/catch, though—so any runtimes while moving the stuff it might create would be falsely marked as a runtime during the creation of the object itself.

code/unit_tests/_defines.dm
```diff
@@ -1 +1 @@
-#define MAX_UNIT_TEST_RUN_TIME 2 MINUTES
+#define MAX_UNIT_TEST_RUN_TIME 20 MINUTES
```
noelle-lavenza: 👀 lol

code/unit_tests/items.dm
```diff
@@ -36,13 +47,72 @@
 			if(!T.run_test(obj_test_instances[objpath]))
 				failures += "Failed volatile test: [objpath], [T.type]"
 
+	// Run harddel checks on each object.
+	for(var/objpath in obj_test_instances)
+		var/obj/item/curr_obj = obj_test_instances[objpath]
+		if(QDELETED(curr_obj)) // curse you, QDEL_IN
+			continue
+		qdel(curr_obj, force = TRUE)
+		obj_test_instances -= objpath // have to delete all references for it to GC, duh
```
noelle-lavenza: Does it clear out faster if you remove the references *prior* to qdeling it?

noelle-lavenza (commented):
~~tg cargo cult moment~~

being 200% honest here I think this should wait until 515 so that we can just check refcount instead, increasing unit test run times to 20 minutes *per map* is absurd

code/unit_tests/items.dm
out-of-phaze: the cleanup has given me no shortage of issues when i expanded it to cover more than just items. if it runtimes outside of the try/catch the whole test dies, which sucks, so i might just add a second try/catch

code/controllers/subsystems/fluids.dm
```diff
@@ -203,6 +203,8 @@ SUBSYSTEM_DEF(fluids)
 		processing_holders.len--
 		if(!QDELETED(reagent_holder))
 			reagent_holder.handle_update()
+		else
+			holders_to_update -= reagent_holder
```
MistakeNot4892, line 207: I am reminded why I wanted to split this fire() proc up into multiple subsystems.

maps/away/mobius_rift/mobius_rift.dm
```diff
@@ -16,6 +16,7 @@
 	)
 
 /obj/effect/step_trigger/mobius_rift/seamless_portal
```
MistakeNot4892, line 18: These will need to be removed when we do generalized portals.

code/unit_tests/~unit_test_types.dm
```diff
@@ -66,7 +71,9 @@
 	name = "Edge of Map - Only map space turfs here"
 
 /obj/item/paper/secret_note/example
+	is_spawnable_type = FALSE
```
MistakeNot4892, line 79: Do these actually throw a runtime when spawned or is this just for consistency's sake?

code/unit_tests/graph_tests.dm
```diff
@@ -478,11 +478,13 @@
 	src.name = name
 
 /atom/movable/graph_test
+	is_spawnable_type = FALSE
 	var/datum/node/physical/node
 	var/list/neighoursByDirection = list()
+	is_spawnable_type = FALSE
```
MistakeNot4892: Set twice.

code/unit_tests/_defines.dm
MistakeNot4892: This does seem pretty high. Why 20 minutes in particular?

code/modules/recycling/sortingmachinery.dm
```diff
@@ -18,13 +18,17 @@
...
+		unwrap(src)
 	return TRUE
 
-/obj/structure/bigDelivery/proc/unwrap(var/mob/user)
-	if(Adjacent(user))
-		// Destroy will drop our wrapped object on the turf, so let it.
-		qdel(src)
+/obj/structure/bigDelivery/proc/unwrap()
+	if(isloc(loc))
+		if(!QDELETED(wrapped)) //sometimes items can disappear. For example, bombs. --rastaf0
```
MistakeNot4892: Out of scope for this PR but this would be a good use-case for a weakref probably.

code/modules/modular_computers/file_system/directory.dm
```diff
@@ -11,9 +11,10 @@
 /datum/computer_file/directory/Destroy()
 	for(var/weakref/file_ref in held_files)
 		var/datum/computer_file/held_file = file_ref.resolve()
-		if(held_file)
-			holder.remove_file(held_file, forced = TRUE)
-	
+		var/obj/item/stock_parts/computer/hard_drive/hard_drive = holder?.resolve()
+		if(held_file && hard_drive)
```
MistakeNot4892, line 15: Should probably be checking istype()/qdeleted on the holder derefs.

code/modules/mob/observer/ghost/ghost.dm
```diff
@@ -55,8 +56,9 @@ var/global/list/image/ghost_sightless_images = list() //this is a list of images
 		mind = body.mind	//we don't transfer the mind but we keep a reference to it.
 	else
 		spawn(10) // wait for the observer mob to receive the client's key
```
MistakeNot4892, line 58: Would be good if you could split this off into a `set waitfor = FALSE` proc while you're in the area.

MistakeNot4892: Not blocking though.

code/modules/mob/living/silicon/robot/modules/_module.dm
```diff
@@ -112,7 +112,7 @@
 /obj/item/robot_module/proc/finalize_emag()
 	if(istype(emag))
 		emag.canremove = FALSE
-	else
+	else if(emag)
 		log_debug("Invalid var type in [type] emag creation - [emag]")
```
MistakeNot4892, line 116: Is a null emag a valid case here?

code/modules/integrated_electronics/subtypes/input.dm
```diff
@@ -652,7 +652,6 @@
 
 /obj/item/integrated_circuit/input/signaler/Destroy()
 	radio_controller.remove_object(src,frequency)
-	QDEL_NULL(radio_connection)
```
MistakeNot4892, line 655: Should this just null `radio_connection` instead?

code/game/world.dm
```diff
@@ -154,6 +158,7 @@ var/global/world_topic_last = world.timeofday
 	..(reason)
 
 /world/Del()
+	Master.Shutdown()
```
MistakeNot4892, line 156: This should probably be a hook - I though it was a hook already.

MistakeNot4892: Scratch that, the hook I was thinking of was the MC shutdown proc.

Still, could use the shutdown hook. Not blocking either way.

code/datums/mind/mind.dm
```diff
@@ -71,6 +71,8 @@
 	QDEL_NULL_LIST(objectives)
 	QDEL_NULL(changeling)
 	SSticker.minds -= src
+	current = null
+	original = null
```
MistakeNot4892, line 79: Should it be checking/clearing the mind var on `current` and `original` if set to itself?

code/datums/observation/observation.dm
```diff
@@ -57,11 +57,11 @@
...
-	var/list/event_sources = list()     // Associative list of event sources, each with their own associative list. This associative list contains an instance/list of procs to call when the event is raised.
-	var/list/global_listeners = list()  // Associative list of instances that listen to all events of this type (as opposed to events belonging to a specific source) and the proc to call.
-	var/flags                           // See _defines.dm for available flags and what they do
+	var/name = "Unnamed Event"                // The name of this event, used mainly for debug/VV purposes. The list of event managers can be reached through the "Debug Controller" verb, selecting the "Observation" entry.
+	var/expected_type = /datum                // The expected event source for this event. register() will CRASH() if it receives an unexpected type.
+	var/list/list/list/event_sources = list() // Associative list of event sources, each with their own associative list. This associative list contains an instance/list of procs to call when the event is raised.
```
MistakeNot4892, line 62: jesus christ

code/datums/observation/observation.dm
```diff
@@ -133,36 +133,39 @@
 /decl/observ/proc/unregister(var/event_source, var/datum/listener, var/proc_call)
 	// Sanity.
 	if (!event_source || !listener || !event_sources[event_source])
-		return FALSE
+		return 0
 
-	// Return false if nothing is listening for this event.
+	// Return zero if nothing is listening for this event.
```
MistakeNot4892, line 138: What's the significance of explicitly returning an integer here?

MistakeNot4892: Nevermind, saw a comment below. Number of things unregistered.

code/game/objects/effects/decals/Cleanable/humans.dm
```diff
@@ -43,7 +43,7 @@ var/global/list/image/splatter_cache=list()
 		amount = 0
 		STOP_PROCESSING(SSobj, src)
 		remove_extension(src, /datum/extension/scent)
-	 . = ..(ignore = TRUE)
```
MistakeNot4892, line 46: That's a fun one. Why did that compile?

code/game/objects/items/devices/uplink.dm
```diff
@@ -221,9 +221,11 @@
 // Includes normal radio uplink, multitool uplink,
 // implant uplink (not the implant tool) and a preset headset uplink.
 
+/obj/item/radio/uplink/var/tc_amount = DEFAULT_TELECRYSTAL_AMOUNT
```
MistakeNot4892: I was going to ask for this to be moved into the `/obj/item/radio/uplink` declaration but there isn't one. Could you relatively path the var declaration and make this the Official radio uplink declaration?

code/game/world.dm
```diff
@@ -71,7 +71,11 @@ GLOBAL_PROTECTED_UNTYPED(game_id, null)
 	return match
 
 /world/New()
-
+// Remove when https://www.byond.com/forum/post/2833191 is fixed
```
MistakeNot4892: We need to have some super early logic that iterates all available map templates and resizes the world to the largest. Also something that prevents admin hotloading a dmm with a larger footprint than the existing world.

code/unit_tests/del_the_world.dm
```diff
@@ -0,0 +1,146 @@
...
+		var/list/del_candidates = spawn_loc.contents - cached_contents
+		if(length(del_candidates)) // explicit length check is faster here
+			for(var/atom/to_del in del_candidates)
+				if(QDELETED(to_del))
+					continue
+				qdel(to_del, force = TRUE) // I hate borg stacks I hate borg stacks
+		AM = null // this counts as a reference to the last item if we don't explicitly clear it??
+
+	// Check for hanging references.
+	SSticker.delay_end = TRUE // Don't end the round while we wait!
```
MistakeNot4892, line 58: Should this be set prior to starting the init loop? I imagine it might take awhile to run.

code/unit_tests/del_the_world.dm
```diff
@@ -0,0 +1,146 @@
...
+		var/oldest_packet = check_queue[1]
+		//Pull out the time we deld at
+		var/qdeld_at = check_queue[oldest_packet]
+		//If we've found a packet that got del'd later then we finished, then all our shit has been processed
+		if(qdeld_at > start_time)
+			garbage_queue_processed = TRUE
+			break
+
+		if(world.time > start_time + time_needed + 8 MINUTES)
+			fail("Something has gone horribly wrong, the garbage queue has been processing for well over 10 minutes. What the hell did you do??")
```
MistakeNot4892: I know I'm being a joyless husk but I'm not a fan of this kind of message in unit tests.

code/datums/observation/observation.dm
out-of-phaze: yeah, had to do this for the linter to not scream at me for copying it in event cleanup debug logs

code/modules/mob/living/silicon/robot/modules/_module.dm
out-of-phaze: It shouldn't print a warning in the log, if that's what you mean. Far as I know it works fine with syndiborgs and combat borgs, which is what this was giving a warning on. They just don't get an item when they're emagged.

code/modules/modular_computers/file_system/directory.dm
out-of-phaze: `resolve()` ought to do that, I think. Specifically, it checks if the located value's `weakref` var is that specific weakref; that's unset in Destroy() and would be false if the ref was now pointing to a new object.

code/unit_tests/_defines.dm
out-of-phaze: That was back when it was broken and would routinely time out. I think 10 minutes should probably be fine? It has to be at least the sum of the first two queue times, and maybe 20 seconds to a minute extra to process the queue (which is way faster now btw).

code/unit_tests/del_the_world.dm
out-of-phaze: It's actually not that slow compared to the next bit, which sleeps for like three to five minutes. Also, I'm unsure if this is necessary; just copied it from TG.

code/unit_tests/graph_tests.dm
out-of-phaze: Rebase artifact, oops.

code/unit_tests/~unit_test_types.dm
out-of-phaze: Consistency, and also they *were* runtiming because I was too lazy to copy over the example secrets to my local testing. I did that though, so it's technically not an issue. That said, I don't want the create-and-destroy tests to fail if the secrets system breaks, and this is already tested earlier.

code/game/objects/effects/decals/Cleanable/humans.dm
out-of-phaze: No idea! Just saw the red-highlighted whitespace in VSCode and went 'uh oh'.

code/modules/integrated_electronics/subtypes/input.dm
out-of-phaze: No, `remove_object` does that for us.

code/unit_tests/items.dm
out-of-phaze: It's going to take at least one queue run no matter what.

code/game/world.dm
out-of-phaze: Earliest I can safely do is in SSmapping. Think it's worth a shot? Iterate over all registered map templates, pass `measureOnly = TRUE`, resize `maxx` and `maxy` to the largest? Or do you want some bespoke code here that circumvents the map template registration system to resize it before anything else is done?

MistakeNot4892:
@Lohikar would you mind looking at the GC changes in this PR?

MistakeNot4892 (dismissed):

Lohikar:
The GC changes look good to me, but I would like the PR title and/or desc to be updated to indicate that this is also doing fixes/improvements to the GC and ED.

MistakeNot4892 (dismissed):

MistakeNot4892 (approved):

PR #2800, opened 2022-11-05
Prevent materials from generating recipes by default if they aren't solid at room temperature.

MistakeNot4892:
Should drastically reduce the creation and deletion of atoms during startup.

### DIFF
--- a/code/modules/materials/material_recipes.dm
+++ b/code/modules/materials/material_recipes.dm
@@ -17,8 +17,11 @@
 		. += new recipe_type(src)
 
 /decl/material/proc/generate_recipes(var/reinforce_material)
-	. = list()
 
+	if(phase_at_stp() != MAT_PHASE_SOLID)
+		return list()
+
+	. = list()
 	if(opacity < 0.6)
 		. += new/datum/stack_recipe/furniture/borderwindow(src, reinforce_material)
 		. += new/datum/stack_recipe/furniture/fullwindow(src, reinforce_material)


comma (approved):

code/modules/materials/material_recipes.dm
```diff
@@ -17,8 +17,11 @@
 		. += new recipe_type(src)
 
 /decl/material/proc/generate_recipes(var/reinforce_material)
-	. = list()
 
+	if(phase_at_stp() != MAT_PHASE_SOLID)
```
out-of-phaze, line 21: Why not just pass a temperature argument in? If someone's standing in subzero temperatures they might be able to craft a frozen cake batter reinforced window, for example, even if they can't at STP.

out-of-phaze (approved):
Seems good enough for a stopgap to prevent unnecessary recipe generation, and will need to be retouched if we redo material phases in the future anyway.

code/modules/materials/material_recipes.dm
MistakeNot4892: The list is cached on first call.

MistakeNot4892: To elaborate a bit - letting recipes be variable based on environmental factors would require a fairly substantial rework of how recipes are created and cached,

out-of-phaze: Right. To do that we'd need to rewrite a lot anyway, so how it is in the PR is fine.

PR #2802, opened 2022-11-05
Stack changes for wrapping paper

PsyCommando:
## Description of changes
@afterthought2 said he wanted the stack changes from #2683 in another PR, so I put them in another PR.

## Changelog
:cl:
code: /obj/item/stack's create matter is more consistent with the other things that create matter. Namely it deletes the list if its empty, and lazy access the list.
code: /obj/item/stack's matter_per_piece is now definable and preserved past init, so a static amount of matter used per unit in a stack can be used.
code: /obj/item/stack's matter is preserved post init, so that its possible to define extra matter that isn't modified from changing the amount of things in the stack.
code: Added a on_used_last() proc to /obj/item/stack so that its possible to leave a trash item or do other things when the stack is completely spent.
/:cl:

### DIFF
--- a/code/game/objects/items/stacks/stack.dm
+++ b/code/game/objects/items/stacks/stack.dm
@@ -216,16 +216,21 @@
 	return get_amount() >= used
 
 /obj/item/stack/create_matter()
-	matter_per_piece = matter?.Copy() // this is used for refreshing matter amount in update_matter()
+	//Preserve matter_per_piece if already set, so it can be set manually.
+	if(!matter_per_piece)
+		matter_per_piece = matter?.Copy() // this is used for refreshing matter amount in update_matter()
 	if(istype(material))
 		LAZYINITLIST(matter_per_piece)
 		matter_per_piece[material.type] = max(matter_per_piece[material.type], round(MATTER_AMOUNT_PRIMARY * matter_multiplier))
-	. = ..()
+	//Don't call the parent, we handle matter completely differently, and it does a lot of stuff we'd just be overwriting completely
+	update_matter()
 
 /obj/item/stack/proc/update_matter()
-	matter = list()
+	//Preserve the matter contents for stacks that have extra matter not in matter_per_piece
+	LAZYINITLIST(matter)
 	for(var/mat in matter_per_piece)
 		matter[mat] = (matter_per_piece[mat] * amount)
+	UNSETEMPTY(matter)
 
 /obj/item/stack/proc/use(var/used)
 	if (!can_use(used))
@@ -233,7 +238,7 @@
 	if(!uses_charge)
 		amount -= used
 		if (amount <= 0)
-			qdel(src) //should be safe to qdel immediately since if someone is still using this stack it will persist for a little while longer
+			on_used_last()
 		else
 			update_icon()
 			update_matter()
@@ -246,6 +251,10 @@
 			S.use_charge(charge_costs[i] * used) // Doesn't need to be deleted
 		update_icon()
 		return TRUE
+
+///When the last item in the stack has been used up, this is called. Default behavior is deleting itself.
+/obj/item/stack/proc/on_used_last()
+	qdel(src) //should be safe to qdel immediately since if someone is still using this stack it will persist for a little while longer
 
 /obj/item/stack/proc/add(var/extra)
 	if(!uses_charge)


afterthought2:
Declaring this not frozen despite being a feature PR.

code/game/objects/items/stacks/stack.dm
```diff
@@ -216,24 +216,29 @@
...
 	if(istype(material))
 		LAZYINITLIST(matter_per_piece)
 		matter_per_piece[material.type] = max(matter_per_piece[material.type], round(MATTER_AMOUNT_PRIMARY * matter_multiplier))
-	. = ..()
+	//Don't call the parent, we handle matter completely differently, and it does a lot of stuff we'd just be overwriting completely
+	update_matter()
 
 /obj/item/stack/proc/update_matter()
-	matter = list()
+	//Preserve the matter contents for stacks that have extra matter not in matter_per_piece
```
MistakeNot4892, line 229: Stacks should not have matter outside of `matter_per_piece` though? The point of the list is to be consistent for `matter` rebuilds.

code/game/objects/items/stacks/stack.dm
```diff
@@ -216,24 +216,29 @@
 	return get_amount() >= used
 
 /obj/item/stack/create_matter()
-	matter_per_piece = matter?.Copy() // this is used for refreshing matter amount in update_matter()
+	//Preserve matter_per_piece if already set, so it can be set manually.
+	if(!matter_per_piece)
```
MistakeNot4892, line 220: This is going to potentially cause inconsistencies if stuff isn't nulling `matter_per_piece` prior to calling `create_matter()` when doing a rebuild (organs do this in roboticize for example).

code/game/objects/items/stacks/stack.dm
MistakeNot4892: Really don't like how easy it's going to be for matter list inconsistencies to creep in with these changes.

MistakeNot4892: If this is all specifically for the cardboard tube in the other PR, I would suggest wrapping paper should handle that entirely on its own. Making it a general behavior at stack level is just making stack matter more brittle. I've already had to rewrite it due to similar issues from the last version.

PsyCommando: I'd argue that maybe always clearing the entire list of matter probably should only be done in the ``/obj/item/stack/material`` subclass instead? Since only material stacks are just pure material stacks unlike the other stacks. 
Things under ``/obj/item/stack`` tend to be items with a limited usage count, and it's not too far fetched to think it might be desirable to leave a trash item or something similar after it's used up. Basically like the wrapping paper and its carton tube. It definitely could find use in a number of ``/obj/item/stack`` like nanopaste, trauma kits packages etc.. 

Or maybe I could just add a list of matter that would be always added to the recomputed matter list? The reason I reused the matter list was that it was pretty straightforward to only overwrite the specific matter that was in matter per piece, and I don't think material stacks can/should change material at runtime.

code/game/objects/items/stacks/stack.dm
PsyCommando: Can you give me an example of what you mean by inconsistencies?

code/game/objects/items/stacks/stack.dm
MistakeNot4892: `I'd argue that maybe always clearing the entire list of matter probably should only be done in the /obj/item/stack/material subclass instead?` and `Since only material stacks are just pure material stacks unlike the other stacks.` are unrelated though. The list clearing has nothing to do with single-material purity, it is intended to prevent the amount of items in the stack and the amount of matter in the stack from becoming inconsistent.

Wrapping, tubes, etc. should be handled as entirely separate items otherwise when you feed a stack into a lathe, you are going to be feeding in `some static material amount for the tube or whatever / stack amount consumed`, which is conceptually confusing.

I do not personally think you should put additional matter in at all. You could make the stack work like food and drop a 'trash' item on consumption, sure (and add the cost of the tube to the crafting recipe).

code/game/objects/items/stacks/stack.dm
MistakeNot4892: The amount of items in the stack and the amount of matter in the stack being at odds.

code/game/objects/items/stacks/stack.dm
MistakeNot4892: Additional problem with the association of a tube with the stack: what happens when you split a stack? Do both stacks contain a tube?

PsyCommando: You can't split/merge wrapping paper stacks, so that's not an issue.

Also, the problem here is that not accounting for all the matter that "comes with" an item means we're creating new matter out of thin air, which can be abused. I don't know what's so confusing about accounting for the entire matter value of the thing? 

Also, the point is exactly that when you recycle the object you get the entire matter value so it stays consistent. Food also has a similar issue, though its not as bad, since the recipes don't cost matter. Also, a lot of food items leaving trash are things you can't craft, and so don't have a matter cost to them.

Making sure no new matter is being created out of thin air is pretty important if you want to have any sort of sensible economy.

code/game/objects/items/stacks/stack.dm
PsyCommando: Why is that inconsistent? The ``matter`` list has no requirements to match exactly the ``amount * matter_per_piece`` in the base stack class. 
Some of the base stacks don't even have a matter cost (``/obj/item/stack/medical``, ``/obj/item/stack/telecrystal``, some of ``/obj/item/stack/tile``, ``/obj/item/stack/flag``). 

You don't need matter to match exactly ``amount * matter_per_piece`` to tell the matter amount per unit. Because ``matter_per_piece`` is there for that? ``matter`` is the cost for the entire thing. Why couldn't it include some additional matter that's not affected by the amount? Especially in the ``/obj/item/stack`` class that has relaxed requirements on the matter calculation unlike ``/obj/item/stack/material``, the later which implies its an accurate pure matter count?

code/game/objects/items/stacks/stack.dm
MistakeNot4892: >You can't split/merge wrapping paper stacks, so that's not an issue.

You're implementing these changes at stack level. It is absolutely an issue.

>Making sure no new matter is being created out of thin air is pretty important if you want to have any sort of sensible economy.

Items should not create matter out of thin air. I don't disagree. Making stack material tracking more brittle is not fixing that.

code/game/objects/items/stacks/stack.dm
MistakeNot4892: >Why is that inconsistent? The matter list has no requirements to match exactly the amount * matter_per_piece in the base stack class.

Yes it does. This is literally the point of tracking matter per piece and updating the matter var. Stacks with no matter cost are irrelevant to that. The matter list is used for recycling the item in a lathe.

>Especially in the /obj/item/stack class that has relaxed requirements on the matter calculation unlike /obj/item/stack/material, the later which implies its an accurate pure matter count

You are pulling this out of nowhere. Composite stacks exist.

code/game/objects/items/stacks/stack.dm
```diff
@@ -216,24 +216,29 @@
...
+		matter_per_piece = matter?.Copy() // this is used for refreshing matter amount in update_matter()
 	if(istype(material))
 		LAZYINITLIST(matter_per_piece)
 		matter_per_piece[material.type] = max(matter_per_piece[material.type], round(MATTER_AMOUNT_PRIMARY * matter_multiplier))
-	. = ..()
+	//Don't call the parent, we handle matter completely differently, and it does a lot of stuff we'd just be overwriting completely
+	update_matter()
```
MistakeNot4892, line 226: This is completely changing the call chain for stacks compared to every other `/obj`.

MistakeNot4892:
Per discussion I am not at all in favour of these changes for the stated purpose. You're muddying stack matter calc for the sake of adding the concept of a cardboard tube for wrapping paper. If you desperately want the tube accounted for, add it to the full matter list post-update/create for wrapping paper specifically, do not diddle the core matter calc that I already had to rewrite to get floor tiles and rods to properly track their own matter values to implement it.

code/game/objects/items/stacks/stack.dm
PsyCommando: What would you do then? 
Handling that strictly in the wrapping paper would mean a whole lot of copy-pasta, which really makes that option kind of gross. That could be avoided by defining another list in the base stack class that would how a static matter amount to be added to the matter list when its rebuilt, but you said you're against that.

MistakeNot4892: >Handling that strictly in the wrapping paper would mean a whole lot of copy-pasta,

No it doesn't. It means overriding specific procs to add the matter you want to add. I do not care if you add static matter to the list post-update on wrapping paper, I am objecting to making the changes to core stack matter calc for the sake of wrapping paper.

code/game/objects/items/stacks/stack.dm
PsyCommando: > Yes it does. This is literally the point of tracking matter per piece and updating the matter var. Stacks with no matter cost are irrelevant to that. The matter list is used for recycling the item in a lathe.
> 

So you're saying that since list/matter's use is for recycling the thing, it should not have a static amount of matter added to the matter amount calculated from matter_per_piece? 
I really don't know why you're so adamant about something like that? It sounds more like you're just taking a hard stance on that but don't want to tell me. If that's the case tell me so we can stop going back in forth and going nowhere.

> You are pulling this out of nowhere. Composite stacks exist.
>

You mean ``/obj/item/material`` stacks that can be reinforced. That concept doesn't exist at the ``/obj/item/stack`` level. Just go and read the code yourself? I'm not pulling anything out of nowhere?

code/game/objects/items/stacks/stack.dm
PsyCommando: I think I misunderstood what you meant then?
I thought you were completely against stacks having matter that was added on top of what the stack calculated already.

code/game/objects/items/stacks/stack.dm
MistakeNot4892: >So you're saying that since list/matter's use is for recycling the thing, it should not have a static amount of matter added to the matter amount calculated from matter_per_piece?

No.

>I really don't know why you're so adamant about something like that? 

Because I already had to rewrite this code to remove inconsistencies introduced by the matter calc. The current burn-it-down-and-remake it approach is there for a reason.

>It sounds more like you're just taking a hard stance on that but don't want to tell me. If that's the case tell me so we can stop going back in forth and going nowhere.

I don't even know what you're trying to say here but I don't appreciate the tone.

>You mean /obj/item/material stacks that can be reinforced. That concept doesn't exist at the /obj/item/stack level. Just go and read the code yourself? I'm not pulling anything out of nowhere?

You are extrapolating the idea of matter purity or stringent requirements from code that doesn't have them.

'Go read the code yourself'; I wrote the code. Pull your head in.

code/game/objects/items/stacks/stack.dm
MistakeNot4892: Based on what? I've said repeatedly and explicitly that if you want to do this you should modify the matter list after generation. I think that wanting to do it at all is problematic, and you should use nested items, but you don't seem inclined to do that.

EDIT: presumably based on 

>Stacks should not have matter outside of matter_per_piece though? The point of the list is to be consistent for matter rebuilds.

which was 1. a question 2. still accurate, matter rebuilding using matter_per_piece and then having more matter appended is gross but not problematic.

MistakeNot4892:
Last comment I'm going to make on this, as I've said as much as I'm willing to and the general atmosphere has degraded: literally all you need to do is

````
/obj/item/stack/wrapping_paper_path_blah/update_matter()
    ..()
    matter[cardboard] = whatever
````

My issues with this PR are:
- Messing with stack level list rebuilds, which previously had problems that had to be fixed.
- Implementing niche requirements at stack level when only one subtype needs them.
- Pushing a pattern of fusing conceptually distinct items that shouldn't be the same item (wrapping paper and a trash product).

code/game/objects/items/stacks/stack.dm
PsyCommando: > I don't even know what you're trying to say here but I don't appreciate the tone.
> 

There's no "tone" here? What I was trying to say is if you got your idea already all made up, there's no point in debating it, and getting frustrated at each others. You should just say you don't want it and won't change your mind, so we can move onto something else.

> You are extrapolating the idea of matter purity or stringent requirements from code that doesn't have them.
>

I'm just saying that ``/obj/item/stack/material`` must have a matter list generated from its material and reinf_material, and won't in any way ever be compatible with having the matter list tampered with.

But the ``/obj/item/stack`` doesn't have to clear the whole thing because what it's made out of is less defined/automatic, or less implicit maybe? Basically, it comes down to generic vs specialized.

Pure matter count was referring to material stacks just automatically having their matter list filled by their material and reinforced material, while the base stack just makes a copy of the original matter list to be used as matter per piece.

> 'Go read the code yourself'; I wrote the code. Pull your head in.
>

I mean, you can have written the code a while ago and forgot what's written?
But I just looked at it again and "composite stacks" or whatever I understand you mean by those are only a thing at the /obj/item/stack/material level. Which is why I said you should look at the code, because you're telling me the opposite when I can see that's not the case. Its not an insult or anything?

MistakeNot4892: >But I just looked at it again and "composite stacks" or whatever I understand you mean by those are only a thing at the /obj/item/stack/material level.

Again: you are pulling this from nowhere. The fact that only material stacks have a reinforcing material doesn't imply that the base stack system isn't set up for composite stacks. `matter_per_piece` _copies the matter list_, it is not a single-entry material-only construct.

I'm quite frustrated with the conversation in this PR because I feel like I'm repeating myself over and over again, and getting told to read my own code by someone who, from my perspective, is claiming a subjective interpretation of the code as factual and binding is not helping.

````
/obj/proc/create_matter()
	if(length(matter))
		for(var/mat in matter)
			matter[mat] = round(matter[mat] * get_matter_amount_modifier())
	UNSETEMPTY(matter)

/obj/item/stack/create_matter()
	matter_per_piece = matter?.Copy() // this is used for refreshing matter amount in update_matter()
	if(istype(material))
		LAZYINITLIST(matter_per_piece)
		matter_per_piece[material.type] = max(matter_per_piece[material.type], round(MATTER_AMOUNT_PRIMARY * matter_multiplier))
	. = ..()
````

Stacks take an initial copy of the base matter list because their matter_amount multiplier is `amount`, and `amount` is highly variable. The copy is kept and then updated with base `material` because `material` is inserted into matter in the parent call, and the insertion is contiguous with multiplying by the stack amount, which is not desirable for the per-unit matter list.

Note that material stacks _DO NOT_ modify `material_per_piece`, they modify `matter`, which is then passed into the same logic that all stacks use. This is where they insert reinforced material. They don't have any kind of restriction or behavioral constraints around `matter_per_piece`, they play by identical rules to all stacks. If you wanted to make a stack subtype with seven different composite materials, you would be free to do so.

`matter_per_piece` is destroyed and remade because it is nonsensical to modify it in place. `create_matter()` should not be arbitrary called, and is in fact only done in three places currently, two of which are fundamental state changes on the atom. 

````
/obj/item/stack/proc/update_matter()
	matter = list()
	for(var/mat in matter_per_piece)
		matter[mat] = (matter_per_piece[mat] * amount)
````

`update_matter` destroys and rebuilds for the same reasons described above. `matter_per_piece` should not be modified except explicitly.

MistakeNot4892: Verified composite stacks working for myself on `stable` by adding mats to bruise packs:
image

PsyCommando: See, I understood "composite stack" as in when you reinforce material stacks, and end up with essentially a composite of two materials. Which is why I was insisting on that. There's no malice intended.

PsyCommando:
Just gonna close this, and revert the stack changes in the original PR.

PR #2803, opened 2022-11-05
Make cameras deny all access on network outage

PsyCommando:
## Description of changes
By default cameras would completely unlock their interface if there was a network outage. Which was kind of a big issue given those will happen regularly from random events and the like. And literally anyone could just click on a camera and changes all the access restrictions and network stuff. So made mapped cameras default to just denying all access attempts until the network is restored.

Will prevent easily tampering with cameras whenever anything might cut the network temporarily.

## Changelog
:cl:
bugfix: Prevent mapped security cameras from becoming completely unlocked and accessible by anyone whenever there's a network outage.
/:cl:

### DIFF
--- a/code/game/machinery/camera/camera.dm
+++ b/code/game/machinery/camera/camera.dm
@@ -15,6 +15,10 @@
 	uncreated_component_parts = null
 	construct_state = /decl/machine_construction/wall_frame/panel_closed
 	frame_type = /obj/item/frame/camera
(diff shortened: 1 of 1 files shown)


PsyCommando:
Well, shit, looks like the desperado map tempalte is back to ruin my day :o

MistakeNot4892 (approved):

PR #2808, opened 2022-11-10
Cleans up/comments sheet matter creation.

MistakeNot4892:
## Description of changes
- Sheet matter init now skips the parent call and applies multipliers itself, to avoid wasted effort/inconsistencies.
- Sheet matter has been cleaned up a bit. Redundant double material setting has been removed, lists are nulled properly.
- Reinforced materials on material sheets are set up before it tries to init matter, so they actually work now.
- Comments have been added explaining what/why things are doing as it is apparently unclear.
- Random `spawn(0)` has been replaced by a `waitfor=FALSE` proc.

Cleaner clearer code. Functionally shouldn't be a difference except that mult is applied properly and it doesn't populate and multiply the full matter list only to immediately discard it the first time `update_matter()` is called.

## Authorship
Psy_Commando's PR #2802 indicated there were some problems with the clarity of what the code was doing, and reviewing their PR showed some spots that needed optimization/cleanup/fixing.

## Changelog
Nothing player-facing.

### DIFF
--- a/code/game/objects/items/stacks/stack.dm
+++ b/code/game/objects/items/stacks/stack.dm
@@ -21,7 +21,6 @@
 	var/plural_icon_state
 	var/max_icon_state
 	var/amount = 1
-	var/list/initial_matter
 	var/matter_multiplier = 1
 	var/max_amount //also see stack recipes initialisation, param "max_res_amount" must be equal to this max_amount
 	var/stack_merge_type  //determines whether different stack types can merge
--- a/code/game/objects/items/stacks/stack.dm
+++ b/code/game/objects/items/stacks/stack.dm
@@ -176,11 +176,8 @@
 
 		src.produce_recipe(R, multiplier, usr)
 
-	if (src && usr.machine==src) //do not reopen closed window
-		spawn( 0 )
-			src.interact(usr)
-			return
-	return
+	if(!QDELETED(src))
+		interact(usr)
 
 //Return 1 if an immediate subsequent call to use() would succeed.
 //Ensures that code dealing with stacks uses the same logic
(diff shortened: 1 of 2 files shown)


PsyCommando:
That does look cleaner and easier to understand!
Btw, I think you've commited some mob icons by mistake?

MistakeNot4892:
The PR was hitting the runtime they're trying to fix.

code/modules/materials/material_sheets.dm
```diff
@@ -9,23 +9,27 @@
 	max_amount = 60
 	randpixel = 3
 	icon = 'icons/obj/materials.dmi'
-	matter = null
+	matter = null // Material stacks are expected to populate their own matter var as they are generally abstract 'shapes' rather than discrete functional items.
```
Andrew-Fall: I know they don't work right now but any chance this could be a `//!` doc comment?

MistakeNot4892: What does that do?

Andrew-Fall: It makes the comment show up when hovering over the item.
image
image

MistakeNot4892: Done.

code/game/objects/items/stacks/stack.dm
```diff
@@ -177,11 +176,13 @@
...
-			src.interact(usr)
-			return
-	return
+	if(src && usr.machine == src) //do not reopen closed window
+		do_nonblocking_interact(usr)
+
+/obj/item/stack/proc/do_nonblocking_interact(var/mob/user)
+	set waitfor = FALSE
+	sleep(0)
+	interact(user)
```
afterthought2: Unless you are manually calling `Topic` somewhere (never do that), you should be able to just call `interact(user)` at the end. You aren't unblocking for anything in the rest of `Topic` because this is at the end of `Topic`.

`src` will never be null unless you manually null it or are doing `call` abuse. `usr.machine == src` is not a reliable check of anything, but I guess you can keep it if you like.

code/modules/materials/material_sheets.dm
afterthought2: Does not appear done.

afterthought2 (approved):

PR #2811, opened 2022-11-12
Made cryopod despawn checking into a movable atom proc.

MistakeNot4892:
## Description of changes
Makes crypods use an atom/movable proc instead of a list.
Split out of #1447 for ease of review.

Allows modpacks to override this behavior.

## Authorship
Myself.

## Changelog
Nothing player-facing.


afterthought2 (approved):
Could probably just make it not despawn stuff, but I agree this is better.

PR #2812, opened 2022-11-12
Reworking vital organ checking and effects.

MistakeNot4892:
## Description of changes
Split off from #1447.
- Vital organ tags are set per species but checked per organ.
- Missing a vital organ paralyzes you and after a grace period, straight up kills you.

Vital organ code is a bit cleaner and more centralized, and is easier to override and manipulate per species. Also allows for brain transplants and decapitation fixes if the surgeon is quick.

## Authorship
Myself.

## Changelog
:cl:
tweak: There is now a grace period on losing your brain before your mob will die.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human_organs.dm
+++ b/code/modules/mob/living/carbon/human/human_organs.dm
@@ -26,6 +26,22 @@
 // Takes care of organ related updates, such as broken and missing limbs
 /mob/living/carbon/human/proc/handle_organs()
 
+	// Check for the presence (or lack of) vital organs like the brain.
+	// Set a timer after this point, since we want a little bit of
+	// wiggle room before the body dies for good (brain transplants).
+	if(stat != DEAD)
+		if(species.check_vital_organ_missing(src))
+			SET_STATUS_MAX(src, STAT_PARA, 5)
+			if(vital_organ_missing_time)
+				if(world.time >= vital_organ_missing_time)
+					death()
+			else
+				vital_organ_missing_time = world.time + species.vital_organ_failure_death_delay
+		else
+			vital_organ_missing_time = null
+
+
+
 	//processing internal organs is pretty cheap, do that first.
 	for(var/obj/item/organ/I in internal_organs)
 		I.Process()
@@ -291,7 +307,7 @@
 
 /mob/living/carbon/human/on_lost_organ(var/obj/item/organ/O)
 	if(!(. = ..()))
-		return 
+		return
 	//Move some blood over to the organ
 	if(!BP_IS_PROSTHETIC(O) && O.species && O.reagents?.total_volume < 5)
 		vessel.trans_to(O, 5 - O.reagents.total_volume, 1, 1)
--- a/code/game/objects/auras/regenerating_aura.dm
+++ b/code/game/objects/auras/regenerating_aura.dm
@@ -74,7 +74,7 @@
 	if(prob(grow_chance))
 		for(var/limb_type in H.species.has_limbs)
 			var/obj/item/organ/external/E = GET_EXTERNAL_ORGAN(H, limb_type)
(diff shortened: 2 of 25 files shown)


code/modules/mob/living/carbon/human/human_organs.dm
```diff
@@ -26,6 +26,22 @@
 // Takes care of organ related updates, such as broken and missing limbs
 /mob/living/carbon/human/proc/handle_organs()
 
+	// Check for the presence (or lack of) vital organs like the brain.
```
PsyCommando, line 29: I wonder if adding a timed check like that for telling if a mob is unrevivable is a good idea? 
Because it used to be that organs would decay on their own after the containing mob is dead for some time. 

Also, I think it would be important to preserve the existing behaviour and have the mob call death when it loses a vital organ and is clinically dead. Because a lot of things rely on that behaviour. 
For example: 

code/modules/mob/living/carbon/life.dm
```dm
	if(stat != DEAD && !InStasis())
		//Breathing, if applicable
		handle_breathing()

		//Mutations and radiation
		handle_mutations_and_radiation()

		//Chemicals in the body
		handle_chemicals_in_body()

		//Random events (vomiting etc)
		handle_random_events()

		// eye, ear, brain damages
		handle_disabilities()

		handle_immunity()
```
All this stuff will keep ticking during the time you're keeping the mob alive without its vital organs, which may defeat the purpose anyways by causing further damage to existing organs faster than someone can fix them against the revive timer you've set here?

Maybe it would be simpler to just let the mob go into dead state, and rely on organ decay time for telling how long can a corpse be revived?

MistakeNot4892: >Because it used to be that organs would decay on their own after the containing mob is dead for some time.

I don't think this has ever been the case? Organs don't process while inside an owner, they are processed by Life().

Re: the unrevivable timer, we have no concept of corpse revival in code currently. That is the point of adding the delay to the point of death. Having the body being dead representing the hard final cutoff point makes more sense to me conceptually in a medical sim that includes the heart stopping than letting people arbitrarily remove `stat = DEAD`.

>Because a lot of things rely on that behaviour. For example:

I don't think it's an issue that any of the above continues processing while the mob is missing a vital organ. Processing means the body can be medicated, bleeding is still an issue, etc. Whatever specifically should be halted by missing a brain should be checking for the missing brain. If you have specific concerns from the list, I'm happy to add the checks in this PR.

PsyCommando: > > Because it used to be that organs would decay on their own after the containing mob is dead for some time.
> 
> I don't think this has ever been the case? Organs don't process while inside an owner, they are processed by Life().
> 

Yeah, you're right about that one. It was probably a change on old ps13. I looked at it again, and it doesn't seem to handle that.

> Re: the unrevivable timer, we have no concept of corpse revival in code currently. That is the point of adding the delay to the point of death. Having the body being dead representing the hard final cutoff point makes more sense to me conceptually in a medical sim that includes the heart stopping than letting people arbitrarily remove `stat = DEAD`.
> 
> > Because a lot of things rely on that behaviour. For example:
> 
> I don't think it's an issue that any of the above continues processing while the mob is missing a vital organ. Processing means the body can be medicated, bleeding is still an issue, etc. Whatever specifically should be halted by missing a brain should be checking for the missing brain. If you have specific concerns from the list, I'm happy to add the checks in this PR.

Well, I'm not going to get into another big debate on that. Might as well let you try it out.

afterthought2 (dismissed):
I would suggest not caching. Note that the code called by `Process` does not use `vital_to_owner` (it just fetches by tag), while the code in attachment interaction logic can afford to recompute. On the other hand, it's really hard to tell if the caching is accurate.

Your call, though.

afterthought2 (approved):

PR #2813, opened 2022-11-12
Updates heart pulse on death.

MistakeNot4892:
## Description of changes
Adds a holder death proc to organs, causes hearts to unset pulse when it is called.

Prevents mobs showing up as having a pulse after death.

## Authorship
Myself.

## Changelog
Bugfix, nothing player-facing.

### DIFF
--- a/code/modules/mob/death.dm
+++ b/code/modules/mob/death.dm
@@ -54,6 +54,9 @@
 	if(!gibbed && deathmessage != "no message") // This is gross, but reliable. Only brains use it.
 		src.visible_message("<b>\The [src.name]</b> [deathmessage]")
 
+	for(var/obj/item/organ/O in get_organs())
+		O.on_holder_death(gibbed)
+
 	set_stat(DEAD)
 	adjust_stamina(-100)
 	reset_plane_and_layer()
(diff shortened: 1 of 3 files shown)


comma (approved):

code/modules/mob/death.dm
```diff
@@ -54,6 +54,9 @@
 	if(!gibbed && deathmessage != "no message") // This is gross, but reliable. Only brains use it.
 		src.visible_message("<b>\The [src.name]</b> [deathmessage]")
 
+	for(var/obj/item/organ/O in get_organs())
```
PsyCommando, line 57: Shouldn't this be in ``/mob/living`` or ``/mob/living/carbon`` ? 
Because, organs are mostly only relevant at those levels?

PsyCommando: Also, unrelated, but I noticed while code diving for this that sillicons seems to bypass calling death in their ``/mob/living/silicon/robot/handle_regular_status_updates()`` line 102.
Dunno if that's a bug, or intended?

MistakeNot4892: The proc is defined on /mob, so I'm iterating it on /mob. No reason not to.
Dunno about silicons, not something I'm touching in this PR.

afterthought2: I mean, they are just getting a hook they can use if they like, makes sense for any type with organs.

PR #2817, opened 2022-11-13
Fix reattaching external organs with internal organs

PsyCommando:
## Description of changes
Attaching a head via surgery would turn all internal organs in the head into detached organ. And completing the surgery wouldn't attach them. So added a case where when properly attaching the organs during surgery would actually grab the detached organs in the implant list of the head and attach them properly.

## Changelog
:cl:
bugfix: Fix limb reattachment surgery not properly installing any contained internal organs once the last step was completed. 
/:cl:


MistakeNot4892 (approved):
Very much not a fan of this for reasons we discussed Discord, but this is probably the best way to do it prior to a total rework of organ tracking.

PR #2819, opened 2022-11-17
Added dummy states to template icons to try to address a bad icon runtime.

MistakeNot4892:


deathride58 (approved):

MistakeNot4892:
Selfmerging due to scav needing this for the test round tomorrow.

PR #2823, opened 2022-11-18
Added a Python script for replicating player saves for all other maps.

MistakeNot4892:
## Description of changes

Adds a Python script that copies saves for the specified map, for all other maps.

My Python skills are pretty dogshit, mainly from incidental stuff at work, so any suggestions/changes would be appreciated.

For use by Scav since they're trying out a new map and manually copying saves is hell.

## Authorship
Myself.

## Changelog
Nothing player-facing.

### DIFF
--- a/tools/ReplicateSaves/replicate_saves.py
+++ b/tools/ReplicateSaves/replicate_saves.py
@@ -1,0 +1,111 @@
+#!/usr/bin/env python3
+
+import os
+import sys
+import re
+import json
+
+__author__ = "MistakeNot4892"
+__version__ = "0.1.0"
+__license__ = "MIT"
+
+def main():
+
+	# Make sure they gave us a map to replicate.
+	if len(sys.argv) < 2:
+		print("Specify a map to replicate.")
+		return
+
+	# Make sure we can access the maps folder.
+	mapname = sys.argv[1]
+	mapdir =  os.path.join("maps")
+	if not os.path.isdir(mapdir):
+		print("Cannot find directory '" + mapdir + "', make sure you are running this script from the root repository directory.")
+		return
+
+	# Work out what maps we actually need to replicate to.
+	# This should be updated as map directories change, or the script will break.
+	targetmaps = []
+	ignoremaps = [
+		"away",
+		"away_sites_testing",
+		"antag_spawn",
+		"example",
+		"modpack_testing",
+		"random_ruins",
+		"~mapsystem",
+		"~unit_tests"
+	]
+	for dir in os.scandir(mapdir):
+		if os.path.isdir(dir):
+			targetmap = dir.path
+			targetmap = targetmap.replace(mapdir + os.sep, "")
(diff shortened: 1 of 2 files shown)


MistakeNot4892:
Needs to also update `preferences.json`.

noelle-lavenza:
I don't really like this solution. Stella's solution, overriding the proc to get the map save key to be the same for both maps, works a lot better IMO. I guess you'd still need to migrate it to that path so there's nothing wrong with the script itself, but still.

MistakeNot4892:
Making the key the same for all maps means you're saniziting the saves against different criteria when the map changes. Tradeship saves would lose some loadout choices if you opened them on ministation.

This was also needed immediately yesterday, so I banged out an external solution instead of one that needed testing in  DM.

tools/ReplicateSaves/replicate_saves.py
```diff
@@ -0,0 +1,111 @@
...
+		"modpack_testing",
+		"random_ruins",
+		"~mapsystem",
+		"~unit_tests"
+	]
+	for dir in os.scandir(mapdir):
+		if os.path.isdir(dir):
+			targetmap = dir.path
+			targetmap = targetmap.replace(mapdir + os.sep, "")
+			if not targetmap in ignoremaps and targetmap != mapname:
```
out-of-phaze:
```suggestion
			if targetmap not in ignoremaps and targetmap != mapname:
```

tools/ReplicateSaves/replicate_saves.py
```diff
@@ -0,0 +1,111 @@
...
+	# Collect slot names for each user to update their preferences.json
+	for (root, dirs, files) in os.walk(scrapedir):
+		has_prefs = False
+		has_saves = []
+		for file in files:
+			if file == "preferences.json":
+				has_prefs = True
+				continue
+			match = re.match(filename_regex, file, re.I)
+			if not match:
```
out-of-phaze:
```suggestion
			if match is None:
```

noelle-lavenza:
> Making the key the same for all maps means you're saniziting the saves against different criteria when the map changes. Tradeship saves would lose some loadout choices if you opened them on ministation.
> 
> This was also needed immediately yesterday, so I banged out an external solution instead of one that needed testing in DM.

Both tradeship and the new map would have the same key on Scavstation so that they'd share the same saves, and presumably they'd already be compatible anyway. I didn't mean make it the same for *all* maps.

out-of-phaze (dismissed):
Could be cleaned up a bit but I don't see any actual blocking issues.

out-of-phaze (approved):
Assuming you've tested it. If so, looks good.

MistakeNot4892:
Tested and working yep.

afterthought2 (approved):
I don't love the concept--saves are likely to break when you do this, or at any rate there is no guarantee that they won't. But if there is demand for it, OK I guess.

PR #2826, opened 2022-11-22
Fix ladders breaking open turfs

PsyCommando:
## Description of changes
Ladders would always call T.ReplaceWithLattice() on init on the turf they're on, even if there is a lattice there already. But the bad part in this is that ReplaceWithLattice() also changes the turf it's on to the base open turf if it isn't already the same type. Which can have pretty bad side effects.

Normally the only thing a ladder should require is to be on an open turf of any type and have a lattice under itself.

## Changelog
:cl:
bugfix: Mapped ladders no longer change the turf they're on to the map default's open turf if they're on a lattice and any open turf types already during init. This could cause exterior/unsimulated open turfs to replace a simulated open turf and cause serious atmos issues.
/:cl:


MistakeNot4892:
I would recommend checking `is_open` as well.

MistakeNot4892:
I didn't mean alternatively to the lattice, I meant instead of the lattice. A ladder on a non-open turf with a lattice on it still needs to create an open turf under itself.

PsyCommando:
> I didn't mean alternatively to the lattice, I meant instead of the lattice. A ladder on a non-open turf with a lattice on it still needs to create an open turf under itself.

Yes? The condition check for either a closed turf, or a missing lattice?

MistakeNot4892 (approved):

PR #2828, opened 2022-11-22
Readds the cake hat, adds some missing zmangle flag setting and removal.

MistakeNot4892:
Cake hat was removed at some point, was requested by SpookyJumpscare on Scav, so here it is again.


code/modules/clothing/head/misc_special.dm
```diff
@@ -165,3 +165,70 @@
 	icon = 'icons/clothing/head/richard.dmi'
 	body_parts_covered = SLOT_HEAD|SLOT_FACE
 	flags_inv = BLOCK_ALL_HAIR
+// Return of the cake hat.
```
Lohikar, line 169: Mind adding a newline here to separate it from the item above?

code/modules/clothing/head/misc_special.dm
```diff
@@ -165,3 +165,70 @@
...
+	. = ..()
+	update_icon()
+
+/obj/item/clothing/head/cakehat/on_update_icon(mob/user)
+	. = ..()
+	if(is_on_fire && check_state_in_icon("[icon_state]-flame", icon))
+		if(plane == HUD_PLANE)
+			add_overlay("[icon_state]-flame")
+		else
+			add_overlay(emissive_overlay(icon, "[icon_state]-flame"))
```
Lohikar, line 194: Should probably get `ZMM_MANGLE_PLANES`.

code/modules/clothing/head/misc_special.dm
MistakeNot4892: Conflict resolution ate it, will readd.

afterthought2 (approved):

PR #2836, opened 2022-11-24
'Fixes' waddling.

MistakeNot4892:
## Description of changes
Comments out the line that sets the moved event to unavailable for global registration. This should not be a long-term fix, waddling needs to be rewritten to use some other method.

Fixes waddling.

## Authorship
Myself.

## Changelog
Nothing player-facing.


MistakeNot4892:
Changed this to just comment out the flag setting in view of #2837.

PR #2840, opened 2022-11-25
Rewriting mob examine().

MistakeNot4892:
## Description of changes
- Split out of #2533.
- Breaks mob examine up into several procs and relocates inventory display from /human down to /living.
- Adds some helpers around retrieving held item slots and hand slots.

Supporting work for the broader inventory rewrite, and allows mobs other than humans to display equipment.

## Authorship
Myself.

## Changelog
Nothing player-facing.


code/modules/mob/examine.dm
```diff
@@ -0,0 +1,161 @@
+/mob/proc/get_visible_pronouns(hideflags)
+	//suits/masks/helmets make it hard to tell their gender
+	if((hideflags & HIDEJUMPSUIT) && (hideflags & HIDEFACE))
+		return GET_DECL(/decl/pronouns)
+	return get_pronouns()
+
+/mob/proc/get_equipment_visibility()
+	. = 0
+	for(var/hideflag in global.equipment_visibility_flags)
```
out-of-phaze: I'm not a fan of the worst-case performance of this loop and it's kind of an icky pattern. The intent here appears to be "get all equipment visibility flags set by at least one piece of equipment." This code would work, if so:
```
#define EQUIPMENT_VISIBILITY_FLAGS (HIDEGLOVES | HIDESUITSTORAGE | HIDEJUMPSUIT | HIDESHOES | HIDEMASK | HIDEEYES | HIDEEARS | HIDEFACE)

/mob/proc/get_equipment_visibility()
	. = 0
	for(var/obj/item/thing in get_equipped_items(include_carried = FALSE))
		. |= things.flags_inv
	return . & EQUIPMENT_VISIBILITY_FLAGS
```

code/__defines/items_clothing.dm
```diff
@@ -27,14 +27,25 @@
...
-#define BLOCK_FACIAL_HAIR  BITFLAG(5)    // Hides the user's hair, facial and otherwise.
+#define HIDEMASK           BITFLAG(5)
+#define HIDEEARS           BITFLAG(6) // Headsets and such.
+#define HIDEEYES           BITFLAG(7) // Glasses.
+#define HIDEFACE           BITFLAG(8) // Dictates whether we appear as "Unknown".
+#define BLOCK_HEAD_HAIR    BITFLAG(9)    // Hides the user's hair overlay, and replace it with short hairs. Leaves facial hair.
+#define BLOCK_FACIAL_HAIR  BITFLAG(10)    // Hides the user's hair, facial and otherwise.
 #define BLOCK_ALL_HAIR     (BLOCK_HEAD_HAIR | BLOCK_FACIAL_HAIR)
 
+var/global/list/equipment_visibility_flags = list(
```
out-of-phaze: I'd probably replace this with something like `BLOCK_ALL_HAIR`:
```
#define EQUIPMENT_VISIBILITY_FLAGS (HIDEGLOVES | HIDESUITSTORAGE | HIDEJUMPSUIT | HIDESHOES | HIDEMASK | HIDEEYES | HIDEEARS | HIDEFACE)
```
That way we can make use of bit math.

code/modules/mob/inventory.dm
```diff
@@ -223,21 +203,24 @@ var/global/list/slot_equipment_priority = list( \
...
 		return 1 //already unequipped, so success
 	return I.mob_can_unequip(src, slot)
 
-/mob/proc/get_inventory_slot(obj/item/I)
+/mob/proc/get_equipped_slot_for_item(obj/item/I)
 	for(var/s in global.all_inventory_slots)
 		if(get_equipped_item(s) == I)
 			return s
 
+/mob/proc/get_inventory_slot_datum(var/slot)
```
out-of-phaze, line 232: Should this check `get_held_item_slots()` by default?

code/modules/mob/living/inventory.dm
```diff
@@ -88,11 +97,7 @@
 /mob/living/get_equipped_item(var/slot)
 	. = ..()
 	if(!.)
-		var/datum/inventory_slot/inv_slot
-		if(LAZYLEN(inventory_slots))
-			inv_slot = inventory_slots[slot]
-		else if(LAZYLEN(held_item_slots))
-			inv_slot = held_item_slots[slot]
+		var/datum/inventory_slot/inv_slot = LAZYACCESS(inventory_slots, slot)
```
out-of-phaze: Should this use `get_inventory_slots()` instead?

code/modules/projectiles/ammunition.dm
```diff
@@ -65,15 +65,12 @@
 	if(G)
 		put_residue_on(G)
 		var/mob/living/carbon/human/H = G.get_recursive_loc_of_type(/mob/living/carbon/human)
```
out-of-phaze, line 67: Not to be too much of a /tg/ coder, but like, can we please not use variable names like `H` and `G` and `bp`? This is kind of annoying to follow at a glance.

code/modules/projectiles/ammunition.dm
```diff
@@ -65,15 +65,12 @@
...
-		if(H)
-			for(var/bp in H.held_item_slots)
-				var/datum/inventory_slot/inv_slot = H.held_item_slots[bp]
-				if(G == inv_slot?.holding)
-					var/target = H.get_covering_equipped_item_by_zone(bp)
-					if(!target)
-						target = GET_EXTERNAL_ORGAN(H, bp)
+		if(H && (G in H.get_held_items()))
+			var/bp = H.get_equipped_slot_for_item(G)
```
out-of-phaze: This will loop over `all_inventory_slots`, so why check that it's in `get_held_items()` as well? Why not make a `get_held_slot_for_item` proc that just runs a similar check on `get_held_item_slots()`?

out-of-phaze (changes requested):
I want to say that I like how `/mob/examine` is structured now, so that if something in one of the examine procs runtimes, it won't break the chat with an unclosed `<quote>` tag.

I don't like the big block in `show_examined_worn_held_items` but you have plans to replace that anyway.

Have you thought of using some kind of dedicated keyword-replacement system for text processing to avoid all of that text interpolation with `pronouns.He`, `pronouns.is`, etc.? Something like `"%He% %is% wearing [uniform.get_examine_line()]."

code/modules/projectiles/ammunition.dm
MistakeNot4892: I can see the argument against `H` and `G` which honestly I only use by habit, happy to avoid using those (though this specific line isn't my code) - what's the issue with `bp` though? `bodypart` is fairly long and it's a temporary var.

MistakeNot4892:
````
I don't like the big block in show_examined_worn_held_items but you have plans to replace that anyway.
````
You can actually go look at the inventory PR to see the replacement logic, I just couldn't include it in this one because it relies on the rest of the inventory datum system.

````
Have you thought of using some kind of dedicated keyword-replacement system for text processing to avoid all of that text interpolation with pronouns.He, pronouns.is, etc.? Something like `"%He% %is% wearing [uniform.get_examine_line()]."
````
Well, we'd still be doing the interpolation, just via replacetext(). I'm not sure if that benchmarks as faster or slower, I would have assumed slower due to the string searching. Not opposed to the idea in principle, some other systems like emotes and ailments use `$USER_HIS$` for token replacement so that could be generalized.

out-of-phaze:
> Well, we'd still be doing the interpolation, just via replacetext(). I'm not sure if that benchmarks as faster or slower, I would have assumed slower due to the string searching. Not opposed to the idea in principle, some other systems like emotes and ailments use `$USER_HIS$` for token replacement so that could be generalized.

Well, no, that wouldn't be interpolation *per se,* but I get the point. Still, I think the performance hit is worth it to make it less clunky, although you'd open yourself up to the issue of typos like `$USER_HSI$` not getting replaced and being hard to lint for. Could always grep for unsupported tokens with the format `$[A-Z_]+$`, but that could still have false positives, and miss cases like someone leaving off the `$` on the start or the end.

code/modules/projectiles/ammunition.dm
out-of-phaze: It's not quite clear at a glance that it stands for bodypart, and it kind of relies on knowing what the function returns. Even then I'm not really sure what it returns. An object? A define? The next line (after the conditional) implies it's not just bodyparts, it's equipment slots, but it can also be an organ tag??

code/modules/mob/inventory.dm
MistakeNot4892: For this PR, yeah. I think there was some more granular handling for hand slots in the big PR but I can revisit when I get that far.

MistakeNot4892:
Pls merg

MistakeNot4892:
Pls merg

out-of-phaze (approved):
The only comment I have is that the underscore in `BLOCK_HEAD_HAIR` etc. threw me off because `HIDEGLOVES` through `HIDEFACE` don't have an underscore. Would prefer consistency, but that's kind of out of scope for this.

afterthought2 (approved):

PR #2841, opened 2022-11-26
Some event tweaks, addition of update icon event.

MistakeNot4892:
## Description of changes
- Inlines `event_repository.raise_event()`
- Removes some redundant checks/list access from event raising.
- Adds an event raised when an atom updates icon.

Should improve performance generally, and I need the update_icon event for some fork stuff.

## Authorship
Myself.

## Changelog
Nothing player-facing.

### DIFF
--- a/code/__defines/observ.dm
+++ b/code/__defines/observ.dm
@@ -1,0 +1,4 @@
+// This also works, and removes the need for the _REPEAT variant, but the linter hates it:
+//   #define RAISE_EVENT(OBS, args...) (GET_DECL(OBS))?.raise_event(args);
+#define RAISE_EVENT(OBS, args...) var/decl/observ/__event = GET_DECL(OBS); __event?.raise_event(args);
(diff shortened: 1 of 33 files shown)


code/__defines/observ.dm
```diff
@@ -0,0 +1,2 @@
+#define RAISE_EVENT(OBS, args...) var/decl/observ/__event = GET_DECL(OBS); __event?.raise_event(args);
```
afterthought2, line 3: Can probably do `GET_DECL(OBS)?.raise_event(args)`, or even without the `?`, no? I don't like that pattern, but this is one of the few places it would make sense.

afterthought2 (dismissed):

code/__defines/observ.dm
MistakeNot4892: Linter is not a fan unfortunately.

afterthought2 (approved):

PR #2845, opened 2022-12-01
Defensive damage types constants

PsyCommando:
## Description of changes
Replaced literal strings in armor resistance definitions (melee, bullet, laser, bomb, etc) with defined constants (DEF_BULLET, DEF_MELEE, etc)

Spent a long while going through every single usages of those words to replace them with constants, so mass replacing and updating is less of a pain in the ass in the future.


code/__defines/damage_organs.dm
```diff
@@ -33,6 +33,15 @@
 #define FIRE_DAMAGE_MODIFIER 0.0215 // Higher values result in more external fire damage to the skin. (default 0.0215)
 #define  AIR_DAMAGE_MODIFIER 2.025  // More means less damage from hot air scalding lungs, less = more damage. (default 2.025)
 
+//Armor Resistance Types
+#define DEF_MELEE  "melee"   //Blunt and cutting weapons
```
MistakeNot4892: `DEF_` I think is a bit confusing as a macro naming convention - would have gone with something like `ARMOR_` to make it unambiguous. I feel a bit bad for asking you to F&R it all, but the diff noise is significant enough that I think it would be unideal to do it twice.

PsyCommando: Its cool. I don't mind changing it.
Tbh, I was afraid people would think it would be too long to prefix it with armor XD

PsyCommando: Done!

MistakeNot4892:
Needs a squash merge.

MistakeNot4892 (approved):

PR #2846, opened 2022-12-01
Use material alteration for items

PsyCommando:
## Description of changes
Replaced ``applies_material_x`` with ``material_alteration`` and its flags.
This is done prior to the merging of material alteration at the /obj and /obj/item level together. 
This is one of the steps prior to unifying material and damage handling at the /obj level.


MistakeNot4892 (approved):
Honestly thought I already did this. Long overdue.

PR #2847, opened 2022-12-02
Renamed health on hydro tray to plant_health

PsyCommando:
## Description of changes
Hydro trays having a var called "health" for holding the plant health would cause name collision in my obj damage handling PR. So I renamed it to plant_health here instead.

### DIFF
--- a/code/modules/hydroponics/beekeeping/beehive.dm
+++ b/code/modules/hydroponics/beekeeping/beehive.dm
@@ -159,7 +159,7 @@
 	var/trays = 0
 	for(var/obj/machinery/portable_atmospherics/hydroponics/H in view(7, src))
 		if(H.seed && !H.dead)
-			H.health += 0.05 * coef
+			H.plant_health += 0.05 * coef
 			++trays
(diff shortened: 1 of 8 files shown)


MistakeNot4892:
Please target staging with fixes. Dev is going to be sitting there for the next six months.

MistakeNot4892 (approved):

PsyCommando:
That's not a fix. Its something I need for another PR to avoid name collision.

MistakeNot4892:
Misunderstood this as being something from the item damage PR, my mistake.

PsyCommando:
> Misunderstood this as being something from the item damage PR, my mistake.

Well, it's part of it. I renamed the health var on hydro trays so they don't collide with the health var of objects. 
It's just not very useful to me if it goes to staging or stable :P

MistakeNot4892:
If it goes to staging or stable it also goes to dev so I'm not sure what you're suggesting.

PsyCommando:
Just nvm. XD
It's not a fix, so dev is ok?

MistakeNot4892:
Yep.

PR #2849, opened 2022-12-02
Fixed bad use of matter list in errant pisces

PsyCommando:
## Description of changes
That matter list used material names instead of material paths. So I fixed it.


MistakeNot4892:
Stable/staging pls.

PsyCommando:
Seems like the randomly failing random unit test is on staging too.

comma (approved):

PR #2850, opened 2022-12-02
chemdart properly init its reagents

PsyCommando:
## Description of changes
Seems like it was happening too late, so I moved it to initialize_reagents.

### DIFF
--- a/code/modules/projectiles/ammunition/chemdart.dm
+++ b/code/modules/projectiles/ammunition/chemdart.dm
@@ -10,9 +10,9 @@
 
 	muzzle_type = null
 
-/obj/item/projectile/bullet/chemdart/Initialize()
-	. = ..()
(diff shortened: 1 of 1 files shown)


MistakeNot4892 (dismissed):
Can this target stable or staging?

PsyCommando:
I thought we didn't target stable with the freeze?

MistakeNot4892:
Why wouldn't we?

PsyCommando:
I dunno. I just remember that coming up at one point.

MistakeNot4892:
The purpose of staging is to prevent new dev PRs introducing regressions and instabilities while staging is being tested and verified for merge to stable. Anything that can target stable (fixes, etc), should target stable, otherwise they should target dev unless they are a fix for something on staging that isn't on stable.

PsyCommando:
Moved to staging!

MistakeNot4892 (approved):

PR #2851, opened 2022-12-03
Z-Mimic: Season Pass 1

Lohikar:
- Non-OVERWRITE Z-turfs' layering now works when above space parallax (probably).
- Added `ZM_MIMIC_BASETURF`, which makes ZM mimic baseturf instead of below. Has niche uses.
- Added a hook to bin and rebuild ZM's internal z-connection state, for if an existing z stack is mutated.
- Z-turfs copying unlit turfs should now handle shadowing correctly.
- The openturf analyzer has received improvements.
    - Entries now have VV links, where it makes sense.
    - Foreign turfs are shown again, except placed where they're actually layered.
    - Movables now show if they've undergone fixup.

### DIFF
--- a/code/__defines/_tick.dm
+++ b/code/__defines/_tick.dm
@@ -20,6 +20,8 @@
 /// runs stoplag if tick_usage is above the limit
 #define CHECK_TICK ( TICK_CHECK ? stoplag() : 0 )
 
+#define UNTIL(X) while(!(X)) stoplag()
+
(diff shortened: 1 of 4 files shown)


Lohikar:
This needs testing, some of the plane logic may not work on Neb unmodified.

Lohikar:
Looks like Neb uses the same turf plane logic as O7, so it looks like it should work. Nothing obviously broken, at least.

MistakeNot4892 (approved):

PR #2854, opened 2022-12-05
Moving material handling to obj level

PsyCommando:
## Description of changes
This PR contains a number of material handling changes that got split off from the obj damage PR. 
Its mainly stuff related to allowing to grab reliable material info at the /obj level, to be used in computing health, damage resistance, material properties that affect damage, leftover debris after destruction, etc..

Also moved a lot of things that structures had which were also being used in /obj/item and brought them together when possible.
Will also try to integrate material with machines in a sensible way if I can. So everything just works the same way everywhere as much as possible.


MistakeNot4892:
Closing as stale, feel free to reopen when updated.

PR #2855, opened 2022-12-28
Fixes supply program not populating categories.

afterthought2:
## Description of changes

Changes what the `is_category` decl proc does to make its (sole) use in supply code result in the expected behavior.

Currently the supply controller does not register any supply packs existing: since the supply first-layer hierarchy members are not marked as abstract, they are not categories, and are ignored.

This targets dev because while I think stable has a similar issue, the implementation there appears different.

It's possible that the correct fix is to instead mark the first layer as abstract (does that work correctly with the rest of the decl infrastructure?) or to remove this helper entirely.

## Authorship

me

### DIFF
--- a/code/datums/hierarchy.dm
+++ b/code/datums/hierarchy.dm
@@ -17,7 +17,7 @@
 	. = ..()
 
 /decl/hierarchy/proc/is_category()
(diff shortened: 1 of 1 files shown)


comma (approved):

afterthought2:
Depends on #2856, I guess.

PR #2863, opened 2023-01-10
Ambient light fixes.

MistakeNot4892:
## Description of changes
Ports a fix from Lohi's repo for ambient lights, renames STARLIGHT to AMBIENT_LIGHT, and fixes `is_outside()` behaving oddly on max-z open turfs.

Fixes ambient light corruption and turbolifts showing ambient light.

## Authorship
@Lohikar and I.

## Changelog
Nothing player-facing.

### DIFF
--- a/code/controllers/configuration.dm
+++ b/code/controllers/configuration.dm
@@ -203,7 +203,7 @@ var/global/list/gamemode_cache = list()
 	var/dsay_allowed = 1
 	var/aooc_allowed = 1
 
-	var/ambient_light = 0	// Whether space turfs have ambient light or not
+	var/exterior_ambient_light = 0	// The strength of ambient light applied to outside turfs
 
 	var/law_zero = "ERROR ER0RR $R0RRO$!R41.%%!!(%$^^__+ @#F0E4'ALL LAWS OVERRIDDEN#*?&110010"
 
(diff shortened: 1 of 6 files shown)


code/modules/admin/view_variables/vv_set_handlers.dm
```diff
@@ -155,7 +155,7 @@
 			if (istext(value) && !color_regex.Find(value))
 				to_chat(client, "<span class='alert'><pre>ambient_light</pre> must be a 6 digit (<pre>#AABBCC</pre>) hexadecimal color string.</span>")
 				return
-			T.set_ambient_light(value)
+			T.set_ambient_light(COLOR_WHITE, value)
```
Lohikar: This VV hook is for the first param (color).

code/controllers/configuration.dm
```diff
@@ -203,7 +203,7 @@ var/global/list/gamemode_cache = list()
 	var/dsay_allowed = 1
 	var/aooc_allowed = 1
 
-	var/starlight = 0	// Whether space turfs have ambient light or not
+	var/ambient_light = 0	// Whether space turfs have ambient light or not
```
Lohikar: I dunno how I feel about calling this `ambient_light`, I feel like it conflates starlight and the ambient lighting system -- which is not just for starlight and the like.

MistakeNot4892: It's not starlight though, it's just exterior turf ambient lighting. I could name it that, I suppose.

Lohikar: Not attached to 'starlight' specifically, just feels like some differentiation would be good.

code/modules/admin/view_variables/vv_set_handlers.dm
MistakeNot4892: So it is, not sure how I misread that.

Lohikar (dismissed):
Per comments, the VV hook needs to be fixed.

code/controllers/configuration.dm
MistakeNot4892: Changed it to `exterior_ambient_light`.

MistakeNot4892:
VV hook fixed.

comma (approved):

PR #2865, opened 2023-01-10
Reimplemented adaptive darksight.

MistakeNot4892:
## Description of changes
This is a super super barebones implementation of adaptive darksight. Being a Tajaran (or enabling base darksight in the config) will allow you to slowly see better in the dark, up to a point. This just sets alpha on a plane master, an overlay based implementation would be desirable but I am hacking this out to get the functionality in place.

This was a feature on Neb prior to O7lights and it turned out it was removed by default rather than for a functional or technical reason, so I'm re-adding it because several people including me are missing it.

## Authorship
Myself, with reference to Citadel-RP's implementation.

## Changelog
:cl:
add: Tajarans can now see better in the dark, after an adjustment period.
/:cl:

### DIFF
--- a/code/controllers/configuration.dm
+++ b/code/controllers/configuration.dm
@@ -910,9 +910,9 @@ var/global/list/gamemode_cache = list()
 				if("grant_default_darksight")
 					config.grant_default_darksight = TRUE
 				if("default_darksight_range")
-					config.default_darksight_range = text2num(value)
+					config.default_darksight_range = max(text2num(value), 0)
 				if("default_darksight_effectiveness")
-					config.default_darksight_effectiveness = text2num(value)
+					config.default_darksight_effectiveness = clamp(text2num(value), 0, 1)
 				else
 					log_misc("Unknown setting in configuration: '[name]'")
 
--- a/code/controllers/configuration.dm
+++ b/code/controllers/configuration.dm
@@ -253,7 +253,9 @@ var/global/list/gamemode_cache = list()
 	var/no_throttle_localhost
 
 	var/dex_malus_brainloss_threshold = 30 //The threshold of when brainloss begins to affect dexterity.
-	var/grant_basic_darksight = FALSE
+	var/grant_default_darksight = FALSE
+	var/default_darksight_range = 2
+	var/default_darksight_effectiveness = 0.05
 
 	var/static/list/protected_vars = list(
 		"comms_password",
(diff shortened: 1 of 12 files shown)


comma (dismissed):

code/modules/species/species.dm
```diff
@@ -35,6 +35,12 @@ var/global/const/DEFAULT_SPECIES_HEALTH = 200
 	var/flesh_color = "#ffc896"             // Pink.
 	var/blood_oxy = 1
 
+	// Darksight handling
+	var/base_low_light_vision = 0
```
out-of-phaze, line 40: It's never noted what the bounds for these variables are. Since we multiply by `1 - base_low_light_vision` I assume it's from 0.0 to 1.0, but nothing caps it. It might be worth checking that it's within the allowed range in `species/validate()`, since creating a new species is one of the easiest modular projects for anyone working on a Nebula downstream to try (in my opinion), and it'd be better if they got all the information in comments and CI rather than from having to ask someone who already knows.

code/modules/species/species.dm
```diff
@@ -348,6 +354,10 @@ var/global/const/DEFAULT_SPECIES_HEALTH = 200
 
 	. = ..()
 
+	if(config.grant_basic_darksight)
+		darksight_range = max(darksight_range, 2)
```
out-of-phaze: I think the basic darksight strengths (2 and 0.05, specifically) should also be configuration variables, with these as the defaults. (I also think it'd still be worthwhile to keep `grant_basic_darksight` around, since sometimes you just want to be able to toggle something off without clobbering the preferred settings for it.)

out-of-phaze (changes requested):
Regarding the config option: I kind of feel like calling it "basic" darksight might be a bit misleading. It's weaker than species-specific darksight tends to be, sure, but not any simpler than them. It supports all the same features, but might make someone think it's a legacy/deprecated form of darksight rather than just a weak setting for adaptive darksight. If you think that's too pedantic feel free to ignore it, though.

code/modules/species/species.dm
MistakeNot4892: Added config options for these.

code/modules/species/species.dm
MistakeNot4892: Added species validation for these, also renamed the config option to 'default' instead of 'basic' per comment above.

code/controllers/configuration.dm
```diff
@@ -904,7 +907,12 @@ var/global/list/gamemode_cache = list()
 					config.use_loyalty_implants = 1
 				if("dexterity_malus_brainloss_threshold")
 					config.dex_malus_brainloss_threshold = text2num(value)
-
+				if("grant_default_darksight")
+					config.grant_default_darksight = TRUE
+				if("default_darksight_range")
+					config.default_darksight_range = clamp(text2num(value), 0, 1)
```
out-of-phaze: This is 2 by default, but if set in the config, it's clamped between 0 and 1. That seems incorrect. Looks like it should probably just have a minimum of 0.
```suggestion
					config.default_darksight_range = max(text2num(value), 0)
```

out-of-phaze (changes requested):
The comment for `GRANT_DEFAULT_DARKSIGHT` still says `very basic darksight`, but I think given the context it's clear enough.

Only remaining issue is a bad clamp on `default_darksight_range`.

code/controllers/configuration.dm
MistakeNot4892: Yeah this was auto copypaste error.

out-of-phaze (approved):

PR #2878, opened 2023-01-13
Fixes issues with the Taj tail animations.

MistakeNot4892:
Readds missing pixels and fills out south-facing state for Taj tails. This seems to have been lost somewhere when I split the markings PR up.


MistakeNot4892:
Never mind, it went to staging and I missed it.

PR #2884, opened 2023-01-16
Made walls generate a compiled icon and use caching.

MistakeNot4892:
## Description of changes
Walls generate a compiled icon based on their state/colours rather than using n*4 image overlays.

Should improve client performance and hopefully init time. Need to test.

## Authorship
Myself.

## Changelog
Nothing player-facing.


MistakeNot4892:
It's really unclear how this impacts performance. Init testing shows almost no impact except on the number of overlays flushed. I'm not sure if the appearance overhead of setting icon, colour and icon state is just outweighing any benefit of removing the image() creation and overlay setting.

comma (approved):

afterthought2:
I'm not a huge fan. Icon manipulation in dm has always been more trouble than it's worth, with random corruption, asset send issues, lack of flexibility, just generally bad performance, etc. I think you'll also find that on an actual map with ruins loaded, there are a lot more wall icon entries than you might think. I don't buy that this meaningfully improves init time, and am skeptical about actual-round client performance.

Not blocking, but suggest weighing pros/cons more carefully.

EDIT: after talking with author, this is meant to address some client issues with wall overlay stacking. Willing to try  it on that basis.

PR #2888, opened 2023-01-19
Mapload now passes the hash to init_atoms() instead of doing it in the loader

MistakeNot4892:
## Description of changes
- Passes the map hash into `init_atoms()` instead of relying on a global and doing it during the loading process.
- Passes the map hash through the submap landmark and setup proc to allow multiple submaps on the same level without job interference.

Fixes some issues with maploading.

## Authorship
Myself and @out-of-phaze.

## Changelog
Nothing player-facing.

### DIFF
--- a/code/modules/maps/_map_template.dm
+++ b/code/modules/maps/_map_template.dm
@@ -61,7 +61,7 @@
 	tallness = bounds[MAP_MAXZ] - bounds[MAP_MINZ] + 1
 	return TRUE
 
-/datum/map_template/proc/init_atoms(var/list/atoms)
+/datum/map_template/proc/init_atoms(var/list/atoms, var/map_hash)
 	if (SSatoms.atom_init_stage == INITIALIZATION_INSSATOMS)
 		return // let proper initialisation handle it later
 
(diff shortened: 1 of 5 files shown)


MistakeNot4892:
````
[11:00:00] Runtime in shuttle.dm,67: Shuttle "Mule100" could not find its starting location.
````

File: code/modules/maps/_map_template.dm [hunk 1]
```diff
@@ -176,10 +169,9 @@
 		else
 			return FALSE
 
-	global._preloader.current_map_hash = null
-
 	//initialize things that are normally initialized after map load
-	init_atoms(atoms_to_initialise)
+	var/map_hash = modify_tag_vars && "[sequential_id("map_id")]"
```
MistakeNot4892, line 212: This is invalid as the sequential ID is one step up from the previous. Need to pass the hash through metadata or something.

MistakeNot4892: Never mind, two separate entrypoints.

MistakeNot4892:
Will reopen this if I ever work out why it was breaking.

PR #2889, opened 2023-01-20
Fix runtime from invalid explosion_act severity

out-of-phaze:
## Description of changes
Incremented for each atom instead of just adding one to their args.

Fixes a runtime from explosions hitting items with a severity of 4 or above.

### DIFF
--- a/code/game/atoms.dm
+++ b/code/game/atoms.dm
@@ -289,7 +289,7 @@ its easier to just keep the beam vertical.
 		. = (severity <= 3)
 		if(.)
 			for(var/atom/movable/AM in get_contained_external_atoms())
-				AM.explosion_act(severity++)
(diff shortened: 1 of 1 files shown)


MistakeNot4892:
Discord's not letting me send messages but I think this proc should look something like:
````
/atom/proc/explosion_act(var/severity)
    SHOULD_CALL_PARENT(TRUE)
    . = !currently_exploding && severity > 0 && severity <= 3
    if(.)
        currently_exploding = TRUE
        if(severity < 3)
            for(var/atom/movable/AM in get_contained_external_atoms())
                AM.explosion_act(severity+1)
        try_detonate_reagents(severity)
        currently_exploding = FALSE
````
to solve the severity > 3 issue more generally.

MistakeNot4892 (approved):

PR #2891, opened 2023-01-20
Fix closets that start open containing objects

out-of-phaze:
## Description of changes
Closets that start open will now spawn their WillContain items inside their loc rather than inside them.
Also modifies closet onDropInto logic to hopefully prevent similar potential issues where things end up inside an open closet.

This means you won't have closets in an inconsistent state, i.e. "sprite is open, items mapped in are outside the closet, items defined in WillContain are inside the closet".

### DIFF
--- a/code/game/objects/structures/crates_lockers/closets/__closet.dm
+++ b/code/game/objects/structures/crates_lockers/closets/__closet.dm
@@ -50,7 +50,7 @@ var/global/list/closets = list()
 /obj/structure/closet/LateInitialize(mapload, ...)
 	var/list/will_contain = WillContain()
 	if(will_contain)
-		create_objects_in_loc(opened && loc || src, will_contain)
+		create_objects_in_loc(opened ? loc : src, will_contain)
 
 	if(!opened && mapload) // if closed and it's the map loading phase, relevant items at the crate's loc are put in the contents
 		store_contents()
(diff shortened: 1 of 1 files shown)


code/game/objects/structures/crates_lockers/closets/__closet.dm
```diff
@@ -50,7 +50,7 @@ var/global/list/closets = list()
 /obj/structure/closet/LateInitialize(mapload, ...)
 	var/list/will_contain = WillContain()
 	if(will_contain)
-		create_objects_in_loc(src, will_contain)
+		create_objects_in_loc(opened && loc || src, will_contain)
```
MistakeNot4892: For the sake of readability ~~and people like me who can't remember operator prececence~~ this should probably be `(opened && loc) || src`

code/game/objects/structures/crates_lockers/closets/__closet.dm
```diff
@@ -419,7 +419,9 @@ var/global/list/closets = list()
 	open()
 
 /obj/structure/closet/onDropInto(var/atom/movable/AM)
-	return
+	if(opened)
+		return loc
+	return null
```
MistakeNot4892: Couldn't this just be `return opened && loc`?

code/game/objects/structures/crates_lockers/closets/__closet.dm
out-of-phaze: Ideally this would just use onDropInto, but that needs an atom passed. Considered making `create_objects_in_loc` use dropInto but that'd be a bit of a headache as well.

MistakeNot4892 (approved):

PR #2892, opened 2023-01-20
Update carnage landmark to avoid throwing qdeleted items

out-of-phaze:
## Description of changes
Makes the carnage landmark spawn before the for loop instead of inside it, and adds a QDELETED check to the thing being thrown.

Avoids runtimes from the carnage landmark trying to throw something that's been deleted.

### DIFF
--- a/code/modules/maps/helper_landmarks.dm
+++ b/code/modules/maps/helper_landmarks.dm
@@ -32,11 +32,11 @@ INITIALIZE_IMMEDIATE(/obj/abstract/landmark/map_load_mark/non_template)
 	return INITIALIZE_HINT_LATELOAD
 
 /obj/abstract/landmark/carnage_mark/LateInitialize()
-	var/area/A = get_area(src)
-	for(var/atom/movable/AM in A)
-		if(AM && !AM.anchored && AM.simulated && prob(movement_prob))
-			spawn()
-				AM.throw_at_random(FALSE, movement_range, 1)
+	var/area/our_area = get_area(src)
+	spawn()
+		for(var/atom/movable/throw_candidate in our_area)
+			if(!QDELETED(throw_candidate) && !throw_candidate.anchored && throw_candidate.simulated && prob(movement_prob))
+				throw_candidate.throw_at_random(FALSE, movement_range, 1)
 	qdel(src)
 
 //Clears walls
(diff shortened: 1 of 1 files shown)


code/modules/maps/helper_landmarks.dm
```diff
@@ -32,11 +32,11 @@ INITIALIZE_IMMEDIATE(/obj/abstract/landmark/map_load_mark/non_template)
...
 /obj/abstract/landmark/carnage_mark/LateInitialize()
-	var/area/A = get_area(src)
-	for(var/atom/movable/AM in A)
-		if(AM && !AM.anchored && AM.simulated && prob(movement_prob))
-			spawn()
-				AM.throw_at_random(FALSE, movement_range, 1)
+	var/area/our_area = get_area(src)
+	spawn()
+		for(var/atom/movable/throw_candidate in our_area)
+			if(!QDELETED(throw_candidate) && !throw_candidate.anchored && throw_candidate.simulated && prob(movement_prob))
```
MistakeNot4892: Possibly throw_at should include a QDELETED check.

out-of-phaze: It has one. QDELETED also includes a nullcheck so `!throw_candidate.anchored` works properly. I guess I could just make it a nullcheck, but eh.

code/modules/maps/helper_landmarks.dm
```diff
@@ -32,11 +32,11 @@ INITIALIZE_IMMEDIATE(/obj/abstract/landmark/map_load_mark/non_template)
 	return INITIALIZE_HINT_LATELOAD
 
 /obj/abstract/landmark/carnage_mark/LateInitialize()
-	var/area/A = get_area(src)
-	for(var/atom/movable/AM in A)
-		if(AM && !AM.anchored && AM.simulated && prob(movement_prob))
-			spawn()
-				AM.throw_at_random(FALSE, movement_range, 1)
+	var/area/our_area = get_area(src)
+	spawn()
```
MistakeNot4892: Please make this a `set waitfor = FALSE` proc with a sleep, rather than using a zero-wait spawn.

out-of-phaze: I'd have to UNLINT the sleep for that to work, since LateInitialize sets SHOULD_NOT_SLEEP(TRUE).

MistakeNot4892: `set waitfor = FALSE` will stop the linter complaining.
EDIT: posted this before Discord convo, investigating

MistakeNot4892 (approved):

PR #2895, opened 2023-01-20
Fix material processors breaking on (shuttle) move

out-of-phaze:
## Description of changes
Material processors start with `input_turf` and `output_turf` as directions. These are then changed to actual turf references.
However, these aren't updated when they move, which is an issue mainly for shuttles. By using the moved observation we can update it properly on move.

Material processors will now input and output from the right turfs relative to them after moving in a shuttle.

### DIFF
--- a/code/modules/mining/machinery/_material_processing.dm
+++ b/code/modules/mining/machinery/_material_processing.dm
@@ -90,6 +90,7 @@
 /obj/machinery/material_processing/Destroy()
 	input_turf = null
 	output_turf = null
+	events_repository.unregister(/decl/observ/moved, src, src, .proc/on_moved)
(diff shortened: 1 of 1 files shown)


MistakeNot4892:
Per Discord, I think you need to unregister your events in Destroy(). Checked the code and other atoms do it, ex:
````
/datum/extension/deity_be_near/Destroy()
	events_repository.unregister(/decl/observ/moved, holder,src)
	events_repository.unregister(/decl/observ/destroyed, holder, src)
	events_repository.unregister(/decl/observ/item_equipped, holder, src)
	. = ..()
````

MistakeNot4892 (approved):

PR #2896, opened 2023-01-20
Add probabilistic variable setter mapping helper

out-of-phaze:
## Description of changes
Adds a mapping helper, `/obj/abstract/landmark/variable_setter`, that has a percent chance to set a specified variable on a specified type to a specified value during Initialize. Also includes a pre-existing type that's just for closets, since it's the one I used the most.

Well, it's a potential mapper footgun, but that's what we have CI for. Could maybe even override it to always have a probability of 1 during CI so that we catch any nonexistent vars, pending Lummox fixing `nameof` to properly work on variables (and also us using 515...)

This lets us make away mission and submap layouts more varied; a fun combination is the closet opener combined with the carnage landmark, meaning there's a random chance for the closet's contents to have been dumped out and scattered around the room. It's a small touch, but a fun one.


code/modules/maps/helper_landmarks.dm
```diff
@@ -98,4 +98,38 @@
 	events_repository.unregister_global(/decl/observ/shuttle_added, src, .proc/check_shuttle)
 	if(shuttle_datum)
 		events_repository.unregister(/decl/observ/shuttle_moved, shuttle_datum, src, .proc/delete_everything)
-	. = ..()
\ No newline at end of file
+	. = ..()
+
+// Has a percent chance on spawn to set the specified variable on the specified type to the specified value.
+
+/obj/abstract/landmark/variable_setter
```
MistakeNot4892, line 105: Might be worth setting `is_spawnable_type` to FALSE.

MistakeNot4892:
A unit test for this might be worthwhile, though I'm not sure how that could be done. It seems like it would be easy for a var name to change but an away site or ruin not be updated and not hit CI due to never being loaded.

out-of-phaze:
> A unit test for this might be worthwhile, though I'm not sure how that could be done. It seems like it would be easy for a var name to change but an away site or ruin not be updated and not hit CI due to never being loaded.

That's what the away sites testing map is for, isn't it?

MistakeNot4892:
away_sites_testing has been known to not load some maps before, usually stuff that isn't strictly an away site. Admittedly that's more of a problem with our UT model than this PR.

MistakeNot4892 (dismissed):

MistakeNot4892 (approved):

PR #2897, opened 2023-01-20
Improve radiation-bombed exoplanet generation

out-of-phaze:
## Description of changes
So this was doing, like, 350 implicit in world loops during init? Not totally sure how much this will improve performance but it's worth a shot.

Avoids redundant area contents checks (which are implicit in world loops) by caching the results from the first call to `get_area_turfs` and removing turfs as they get used. This takes it from like 350 in-worlds for a 255x255 map down to 1.

### DIFF
--- a/code/modules/overmap/exoplanets/planet_themes/radiation_bombing.dm
+++ b/code/modules/overmap/exoplanets/planet_themes/radiation_bombing.dm
@@ -9,19 +9,20 @@
 /datum/exoplanet_theme/radiation_bombing/get_sensor_data()
 	return "Hotspots of radiation detected."
 
-/datum/exoplanet_theme/radiation_bombing/after_map_generation(obj/effect/overmap/visitable/sector/exoplanet/E)
+/datum/exoplanet_theme/radiation_bombing/after_map_generation(obj/effect/overmap/visitable/sector/exoplanet/our_exoplanet)
 	var/radiation_power = rand(10, 37.5)
-	var/num_craters = round(min(0.5, rand()) * 0.02 * E.maxx * E.maxy)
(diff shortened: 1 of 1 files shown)


code/modules/overmap/exoplanets/planet_themes/radiation_bombing.dm
```diff
@@ -9,19 +9,20 @@
...
-/datum/exoplanet_theme/radiation_bombing/after_map_generation(obj/effect/overmap/visitable/sector/exoplanet/E)
+/datum/exoplanet_theme/radiation_bombing/after_map_generation(obj/effect/overmap/visitable/sector/exoplanet/our_exoplanet)
 	var/radiation_power = rand(10, 37.5)
-	var/num_craters = round(min(0.5, rand()) * 0.02 * E.maxx * E.maxy)
+	var/num_craters = round(min(0.5, rand()) * 0.02 * our_exoplanet.maxx * our_exoplanet.maxy)
+	var/area_turfs = get_area_turfs(our_exoplanet.planetary_area, list(/proc/not_turf_contains_dense_objects))
 	for(var/i = 1 to num_craters)
-		var/turf/simulated/T = pick_area_turf(E.planetary_area, list(/proc/not_turf_contains_dense_objects))
-		if(!T) // ran out of space somehow
+		var/turf/simulated/crater_center = pick_n_take(area_turfs)
```
MistakeNot4892, line 17: Should check `length(area_turfs)` and break early if you run out.

out-of-phaze: We do. `pick_n_take` is safe on empty lists, it returns null, which fails on the next line.

MistakeNot4892: Fair enough, wasn't aware.

comma (approved):

PR #2902, opened 2023-01-21
Make replaced job items go in storage

out-of-phaze:
## Description of changes
Job items replaced by loadout items will now go in storage (or be equipped, or put in hands, or dropped on the floor) instead of vanishing.

Some jobs in the shiptesting modpack I'm working on start with important safety equipment which disappears completely (it actually gets qdeleted) just because my character has a loadout item that takes the same slot.

The alternative to this PR would be to make the loadout item go in storage instead, which I kind of like. Might open it alongside this one.

## Authorship
Me.

## Changelog
:cl:
add: Job items replaced by loadout items now go into storage.
/:cl:

### DIFF
--- a/code/modules/client/preference_setup/loadout/loadout.dm
+++ b/code/modules/client/preference_setup/loadout/loadout.dm
@@ -425,9 +425,13 @@ var/global/list/gear_datums = list()
 	var/obj/item/old_item = wearer.get_equipped_item(slot)
 	if(wearer.equip_to_slot_if_possible(item, slot, del_on_fail = TRUE, force = TRUE, delete_old_item = FALSE))
 		. = item
-		if(old_item && old_item.type != item.type)
-			wearer.u_equip(old_item)
+		if(!old_item)
+			return
+		wearer.u_equip(old_item)
+		if(old_item.type != item.type)
 			place_in_storage_or_drop(wearer, old_item)
+		else
+			qdel(old_item)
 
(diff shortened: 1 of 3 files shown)


MistakeNot4892 (dismissed):

code/modules/client/preference_setup/loadout/loadout.dm
```diff
@@ -417,28 +417,35 @@ var/global/list/gear_datums = list()
...
+/decl/loadout_option/proc/spawn_on_mob(mob/living/carbon/human/wearer, metadata)
+	var/obj/item/item = spawn_and_validate_item(wearer, metadata)
 	if(!item)
 		return
 
-	if(H.equip_to_slot_if_possible(item, slot, del_on_fail = 1, force = 1))
+	var/obj/item/old_item = wearer.get_equipped_item(slot)
+	if(wearer.equip_to_slot_if_possible(item, slot, del_on_fail = TRUE, force = TRUE, delete_old_item = FALSE))
 		. = item
+		if(old_item && old_item.type != item.type)
```
MistakeNot4892: You should probably qdel the old item if the types align.

out-of-phaze: Oh! True. Also, would like to hold off on this being merged until the alternative PR is made as well.

MistakeNot4892:
Per the power of democracy:
image
This will be the one we go with.

MistakeNot4892 (approved):

PR #2904, opened 2023-01-21
Make loadout items targeting full slots go in storage

out-of-phaze:
## Description of changes
Alternative PR to #2902.
Loadout items targeting a full slot (filled due to job equipment) will now go in storage (or be equipped, or put in hands, or dropped on the floor) instead of destroying the job equipment.

Some jobs in the shiptesting modpack I'm working on start with important safety equipment which disappears completely (it actually gets qdeleted) just because my character has a loadout item that takes the same slot.

The alternative to this PR would be to make the job item go in storage instead; see #2902.

In theory this sounded nicer, but I kind of dislike it because it means you'll have to find somewhere to change out of your jumpsuit every round if you select one in your loadout.

## Authorship
Me.

## Changelog
:cl:
add: Loadout items now default to going into storage if they take the same slot as a job item.
/:cl:

### DIFF
--- a/code/controllers/subsystems/jobs.dm
+++ b/code/controllers/subsystems/jobs.dm
@@ -408,7 +408,7 @@ SUBSYSTEM_DEF(jobs)
 	if(!H || !H.client)
 		return
 
-	// Equip custom gear loadout, replacing any job items
+	// Equip custom gear loadout, going into storage if the slot is full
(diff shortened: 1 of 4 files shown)


MistakeNot4892:
Per vote on Discord, we're going with the alt PR.

PR #2905, opened 2023-01-22
Fixes the flood overlay being invisible.

MistakeNot4892:


out-of-phaze (approved):
Have to kind of ask why this is an abstract object and not an effect object; it doesn't delete itself. I guess it's because it's a singleton?

MistakeNot4892:
It could also be an effect, changed it to abstract since it's a visual overlay only rather than something cleanable, or a decal.

comma (approved):

PR #2919, opened 2023-01-23
Fix grab interaction showing up for anchored atoms

out-of-phaze:
## Description of changes
Fixes the grab interaction (with extended alt interactions on) showing up for anchored atoms.

You can't grab anchored atoms anyway. Extended alt interactions are still not that great and I eventually disabled them after a few rounds, but this makes it much easier to use since it doesn't interfere with alt-clicking on catwalks or tables to view their contents. (The examine interaction still does, but I had that removed from alt interactions entirely during the testing.)

### DIFF
--- a/code/game/atoms_movable.dm
+++ b/code/game/atoms_movable.dm
@@ -521,6 +521,9 @@
 	expected_target_type = /atom/movable
 	interaction_flags = INTERACTION_NEEDS_PHYSICAL_INTERACTION | INTERACTION_NEEDS_TURF
 
+/decl/interaction_handler/grab/is_possible(atom/movable/target, mob/user, obj/item/prop)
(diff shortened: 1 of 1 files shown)


code/game/atoms_movable.dm
```diff
@@ -521,6 +521,10 @@
 	expected_target_type = /atom/movable
 	interaction_flags = INTERACTION_NEEDS_PHYSICAL_INTERACTION | INTERACTION_NEEDS_TURF
 
+/decl/interaction_handler/grab/is_possible(atom/movable/target, mob/user, obj/item/prop)
+	if((. = ..()))
```
MistakeNot4892: `return !target.anchored && ..()`?

MistakeNot4892 (approved):

PR #2920, opened 2023-01-23
Make async tests await subsystems properly

out-of-phaze:
## Description of changes
Ensures any awaited subsystems will fire at least once *after* the async test begins, rather than once *overall.*

This should fix the intermittent Desperado CI fails, which were due to weirdness with the timer subsystem falling behind IIRC.

~~watch this fail CI with Desperado issue it's supposed to solve~~


MistakeNot4892 (approved):
Bit uncertain how rigorous this is as a fix (what if timers aren't fully flushed after the next run?) but if you tested it, so be it.

PR #2935, opened 2023-01-28
Makes casings use drop_sound instead of their own system

comma:
## Description of changes
Casings used to have fall_sounds, merged them into drop_sound handling Also drop_sounds now handle lists (picking non list items is fine I tried)

No two parallel systems for drop sounds for object.
Also having support for varied list sounds is nice

## Authorship
Me me me

## Changelog
NA

### DIFF
--- a/code/game/objects/item.dm
+++ b/code/game/objects/item.dm
@@ -102,7 +102,7 @@
 
 /obj/item/proc/dropped_sound_callback()
 	if(!ismob(loc) && drop_sound)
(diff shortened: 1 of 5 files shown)


MistakeNot4892 (approved):

PsyCommando:
But, drop_sound is specifically for when you're holding something and putting it down. 
It's not meant to play when the object is thrown, like casings are?
That means each times you put a casing somewhere it's gonna do the bullet casing hitting the floor noise?

PR #2942, opened 2023-02-01
Microopt for check_state_in_icon().

MistakeNot4892:
## Description of changes
- Adds an alternative faster key to `check_state_in_icon`.
- Skips single icon consideration for clothing, as AFAIK it's all converted to the single icon system already.

Optimization.

## Authorship
@Lohikar and @out-of-phaze helped with the key update.

## Changelog
Nothing player-facing.


MistakeNot4892:
Appears to be completely screwed, nvm.

Issue #2304, opened 2022-03-30
Weapons made with some materials don't use the right damage type


PsyCommando:
#### Description of issue
Depending on the materials, some swords for instance, cause blunt damage instead of edge damage. Its because of some of the materials change that disable the sharp flag on swords based on hardness of the material its made of. The same issue also applies to the energy swords..

#### Steps to reproduce
Try to cut a human's head off with a katana or esword. (Both are considered to be made of steel and get their flags modified)
Notice it gibs the head like a blunt weapon would do.
Try the vibrokatana instead.
Notice how its made of titanium and properly chop the head off?

#### Length of time in which bug has been known to occur

At least a month.

#### Issue bingo

- [X] Issue could be reproduced at least once
- [X] Issue could be reproduced by different players
- [X] Issue could be reproduced in multiple rounds
- [X] Issue happened in a recent (less than 7 days ago) round
- [X] [Couldn't find an existing issue about this](https://github.com/NebulaSS13/Nebula/issues)

MistakeNot4892:
For reference this is specifically handled by `/obj/item/sword/update_force()` and the basic material system has handling for edge/blunt damage as-is - might be worth merging the functions.

comma:
It shouldn't affect e-swords because they're not under `/sword/` path, they're their own item `/obj/item/energy_blade/sword`
Katanas are blunt because steel is not hard enough, going to bump that so it's above cast iron

out-of-phaze:
the esword has sharp and edge both set to false, which implies this is intended? it's a blunt weapon that's just very hot. i think #2953 fixed the actual issues with katanas though, so i'll close this. feel free to open a new one if it's still an issue

PR #2945, opened 2023-02-04
Adds VV option to set item's material

comma:
## Description of changes
Adds an option to VV dropdown menu to set material of the item

Beats defining bunch of debug subtypes for testing material things 

## Authorship
Me

## Changelog
N/A

### DIFF
--- a/code/modules/admin/view_variables/helpers.dm
+++ b/code/modules/admin/view_variables/helpers.dm
@@ -87,6 +87,11 @@
 		<option value='?_src_=vars;emp=\ref[src]'>Trigger EM pulse</option>
 		"}
 
+/obj/item/get_view_variables_options()
+	return ..() + {"
+		<option value='?_src_=vars;setmaterial=\ref[src]'>Set material</option>
+		"}
+
 /turf/get_view_variables_options()
 	return ..() + {"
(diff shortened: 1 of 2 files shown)


code/game/objects/item_materials.dm
```diff
@@ -98,8 +98,10 @@
 		throwforce = round(throwforce)
 		attack_cooldown += material.get_attack_cooldown()
 
-/obj/item/proc/set_material(var/new_material)
+/obj/item/proc/set_material(var/decl/material/new_material)
 	if(new_material)
+		if(!ispath(new_material))
```
MistakeNot4892: This means that the proc is going to fail if you give it a param of the defined type. That feels like a footgun. Maybe `!ispath(new_material) && !istype(new_material)`?

comma: It doesn't handle getting an object instead of a path, since it does GET_DECL on it and gets null

comma: Actually axed this change from this PR, bit too big for a simple debug verb PR imo

comma:
Fixes #2304

comma:
Removed sword stuff, gonna do a separate fix PR with that

MistakeNot4892 (approved):

PR #2946, opened 2023-02-04
Cups now shatter and spill contents when phyiscally destroyed

comma:
## Description of changes
What it says on the tin. They splash contents on their loc, and drop material shards depending on their size

Better than them suddenly popping out of existence when taking too much damage

### DIFF
--- a/code/modules/reagents/reagent_containers/drinkingglass/drinkingglass.dm
+++ b/code/modules/reagents/reagent_containers/drinkingglass/drinkingglass.dm
@@ -227,3 +227,10 @@ var/global/const/DRINK_ICON_NOISY = "noise"
 	. = ..()
 	if(old_temp != temperature)
 		update_icon()
+
+/obj/item/chems/drinks/glass2/physically_destroyed(var/skip_qdel)
+	reagents.splash(loc, reagents.total_volume)
+	if(istype(material))
+		playsound(src, "shatter", 30, 1)
+		material.place_shards(get_turf(src), w_class)
+	return ..()


code/modules/reagents/reagent_containers/drinkingglass/drinkingglass.dm
```diff
@@ -226,4 +226,12 @@ var/global/const/DRINK_ICON_NOISY = "noise"
 	var/old_temp = temperature
 	. = ..()
 	if(old_temp != temperature)
-		update_icon()
\ No newline at end of file
+		update_icon()
+
+/obj/item/chems/drinks/glass2/physically_destroyed(var/skip_qdel)
+	reagents.splash(loc, reagents.total_volume)
+	var/decl/material/M = get_material()
```
MistakeNot4892: Should be able to just use `material` for this.

comma: chaned to that

code/modules/reagents/reagent_containers/drinkingglass/drinkingglass.dm
```diff
@@ -226,4 +226,12 @@ var/global/const/DRINK_ICON_NOISY = "noise"
 	var/old_temp = temperature
 	. = ..()
 	if(old_temp != temperature)
-		update_icon()
\ No newline at end of file
+		update_icon()
+
+/obj/item/chems/drinks/glass2/physically_destroyed(var/skip_qdel)
+	reagents.splash(loc, reagents.total_volume)
+	var/decl/material/M = material
```
MistakeNot4892: Per Discord, you don't need to cast `material`, it's already typed to `/decl/material`.

MistakeNot4892 (approved):

PR #2950, opened 2023-02-05
Makes door leak damage to components when damaged

comma:
## Description of changes
Previously doors would soak all damage up to its health and only then start passing it to components. This made whole thing kinda meaningless because by the time you start damaging copmonents the door is gone, you can just crowbar it open

Now if door is at 75% health or if it's 10 damage in single hit, some damage can leak through. It's scaled to how broken door is, so you're not going to be crunching locks with first few lucky hits. Secure airlocks override that to keep older behaviour because that's more secure

I think you should be able to bash access locks out of doors without completely destroying the airlock
It opens up access to the components hatch which is kinda pointless if machine is dead

## Authorship
Me

### DIFF
--- a/code/game/machinery/doors/_door.dm
+++ b/code/game/machinery/doors/_door.dm
@@ -325,19 +325,35 @@
 	return FALSE
 
 /obj/machinery/door/take_damage(var/damage, damtype=BRUTE)
-	var/initialhealth = src.health
-	src.health = max(0, src.health - damage)
-	if(src.health <= 0 && initialhealth > 0)
+	if(!health)
+		..(damage, damtype)
+		update_icon()
+		return
+
+	var/component_damage = get_damage_leakthrough(damage, BRUTE)
+	damage -= component_damage
+
+	//Part of damage is soaked by our own health
+	var/initialhealth = health
+	health = max(0, health - damage)
+	if(health <= 0 && initialhealth > 0)
 		visible_message(SPAN_WARNING("\The [src] breaks down!"))
-		src.set_broken(TRUE)
-	else if(src.health < src.maxhealth / 4 && initialhealth >= src.maxhealth / 4)
+		set_broken(TRUE)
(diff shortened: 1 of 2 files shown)


code/game/machinery/doors/_door.dm
```diff
@@ -325,20 +325,36 @@
...
+	else if(health < maxhealth * 3/4 && initialhealth >= maxhealth * 3/4)
 		visible_message(SPAN_WARNING("\The [src] shows signs of damage!"))
-	..(max(0, damage - initialhealth), damtype)
+
+	..(component_damage, damtype)
 	update_icon()
 
+//How much damage should be passed to components inside even when door health is non zero
+/obj/machinery/door/proc/get_damage_leakthrough(var/damage, damtype=BRUTE)
+	if(health > 3/4 * maxhealth && damage < 10)
```
MistakeNot4892: `health > (maxhealth * 0.75)` would probably be easier to read.

code/game/machinery/doors/_door.dm
```diff
@@ -325,20 +325,36 @@
...
 /obj/machinery/door/take_damage(var/damage, damtype=BRUTE)
-	var/initialhealth = src.health
-	src.health = max(0, src.health - damage)
-	if(src.health <= 0 && initialhealth > 0)
+	if(!health)
+		..(damage, damtype)
+		update_icon()
+		return
+
+	var/component_damage = get_damage_leakthrough(damage, BRUTE)
```
MistakeNot4892, line 333: Should probably return here if `component_damage` is 0.

comma: Nah it's damaging the door health itself if component damage is 0

code/game/machinery/doors/_door.dm
comma: Done

MistakeNot4892 (approved):

PR #2953, opened 2023-02-07
Sword and steel tweaks

comma:
## Description of changes
Made swords handle setting material better (can handle going from blunt to sharp, not one way)
Buffed steel hardness a bit.
Lowered hardness requirements on a bunch of stack recipies

It fixes katana part of https://github.com/NebulaSS13/Nebula/issues/2304
With new hardness steel is still below plasteel etc, but it's not losing out to bronze
A lot of recipies were locked away before pretty high hardness reqs, so you couldn't make forks of anything weaker titanium. Now it's plastic/wood and up

## Authorship
me

## Changelog
:cl:
bugfix: Steel katanas were woefully underpowered, they are now actually sharp
tweak: Steel is harder now
/:cl:

### DIFF
--- a/code/modules/materials/material_recipes.dm
+++ b/code/modules/materials/material_recipes.dm
@@ -60,7 +60,6 @@
 			. += new/datum/stack_recipe/key(src)
 			. += new/datum/stack_recipe/rod(src)
 
-	if(hardness > MAT_VALUE_RIGID + 10)
 		. += new/datum/stack_recipe/fork(src)
 		. += new/datum/stack_recipe/knife(src)
 		. += new/datum/stack_recipe/bell(src)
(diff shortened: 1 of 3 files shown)


code/modules/materials/material_recipes.dm
```diff
@@ -60,7 +60,6 @@
 			. += new/datum/stack_recipe/key(src)
 			. += new/datum/stack_recipe/rod(src)
 
-	if(hardness > MAT_VALUE_RIGID + 10)
```
MistakeNot4892, line 63: Is this meant to be removed?

comma: Yeah, I changed it to flexible+10 first but then realized it's literally check above, so stuff is already checked for that

MistakeNot4892 (approved):

out-of-phaze (approved):
I was originally not a fan of using initial for resetting sharp/edge/hitsound at the start instead of using the default case of a switch on hardness, but functionally it's the same anyway. I might still change it myself, but that's probably better for a rewrite of swords in general rather than this fix PR.

PR #2955, opened 2023-02-07
Added godmode checks

PsyCommando:
## Description of changes
Added some missing godmode checks.
* Prevent air contaminants from damaging a mob with godmode on.
* Prevent ailments from appearing on the organs of a godmode mob.
* Prevented blood effects from chems on godmode mobs.

### DIFF
--- a/code/modules/ZAS/Contaminants.dm
+++ b/code/modules/ZAS/Contaminants.dm
@@ -72,7 +72,7 @@ var/global/image/contamination_overlay = image('icons/effects/contamination.dmi'
 	if(vsc.contaminant_control.CLOTH_CONTAMINATION) contaminate()
 
 	//Anything else requires them to not be dead.
-	if(stat >= 2 || status_flags & GODMODE)
+	if((stat >= DEAD) || (status_flags & GODMODE))
 		return
 
 	//Burn skin if exposed.
--- a/code/modules/ZAS/Contaminants.dm
+++ b/code/modules/ZAS/Contaminants.dm
@@ -72,7 +72,7 @@ var/global/image/contamination_overlay = image('icons/effects/contamination.dmi'
 	if(vsc.contaminant_control.CLOTH_CONTAMINATION) contaminate()
 
 	//Anything else requires them to not be dead.
-	if(stat >= 2)
+	if(stat >= 2 || status_flags & GODMODE)
 		return
 
 	//Burn skin if exposed.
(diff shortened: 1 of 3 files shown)


code/modules/ZAS/Contaminants.dm
```diff
@@ -72,7 +72,7 @@ var/global/image/contamination_overlay = image('icons/effects/contamination.dmi'
 	if(vsc.contaminant_control.CLOTH_CONTAMINATION) contaminate()
 
 	//Anything else requires them to not be dead.
-	if(stat >= 2)
+	if(stat >= 2 || status_flags & GODMODE)
```
MistakeNot4892: `(stat == DEAD || (status_flags & GODMODE))`

code/modules/organs/organ.dm
```diff
@@ -446,6 +446,8 @@ var/global/list/ailment_reference_cache = list()
 
 /obj/item/organ/proc/get_possible_ailments()
 	. = list()
+	if(owner.status_flags & GODMODE)
+		return .
```
MistakeNot4892, line 450: `return`

Andrew-Fall: `Return .` is clearer I would argue

MistakeNot4892: It's both unnecessary and more expensive.

Andrew-Fall: I swear I saw a post on the byond forums about this being optimized to be the same bytecode

out-of-phaze: That seems to be incorrect. `TICKBENCH` benchmarking appears to show that `return .` is reliably faster than `return` by about 4% to 6%. It's not enough to be worth splitting hairs over, but `return` is definitely not the faster one.

Andrew-Fall: https://www.byond.com/forum/post/2787448
Github actively dying as I'm trying to post :V
This should be fixed according to Lummox in that post

Andrew-Fall: https://www.byond.com/forum/post/2787448

MistakeNot4892: Per discussion this does seem to not matter at all anymore. Personally I am not a fan of the pattern but I'm not going to block over it.

MistakeNot4892 (approved):

code/modules/organs/organ.dm
PsyCommando: Honestly the base idea behind just putting return alone returning something that's not null is kind of confusing and error prone. 
And since it's exactly the same whether you put ``.`` or not, I just put it. It's extremely handy to remind yourself that it's not going to return null, and prevent yourself from doing something stupid. 
True story. X_X

out-of-phaze (approved):

PR #2956, opened 2023-02-07
Make posters individually available in mapping tool

PsyCommando:
## Description of changes
You couldn't place specific posters on the map because they were all datumized and didn't have an actual instance for each datums. 
Added a macro to declare a datum and poster instance. Each with a descriptive name and an a mapper icon.

Makes it way easier to place specific poster instances on the map, or just for spawning. And now you can actually tell what you're placing from the obj icon and name.


MistakeNot4892:
Needs a squashmerge.

code/game/objects/effects/decals/posters/bs12.dm
```diff
@@ -1,290 +1,67 @@
...
+DEFINE_POSTER(bay_55, "bsposter55", "Earth appreciation poster",                       "This is a poster produced by the Generic Space Company, as a part of a series of commemorative posters on the wonders of space. Two of three.")
+DEFINE_POSTER(bay_56, "bsposter56", "Mars appreciation poster",                        "This is a poster produced by the Generic Space Company, as a part of a series of commemorative posters on the wonders of space. Three of three.")
+DEFINE_POSTER(bay_57, "bsposter57", "space carp warning poster",                       "This poster tells of the dangers of space carp infestations.")
+DEFINE_POSTER(bay_58, "bsposter58", "space carp information poster",                   "This poster showcases an old spacer saying on the dangers of migrant space carp.")
+DEFINE_POSTER(bay_59, "bsposter59", "poster - Miss Science 2299",                      "A large piece of space-resistant printed paper. This pin-up poster depicts a woman wearing a corporate labcoat, a bikini, and a sheepish grin. She's shyly posing atop some highly specialized research equipment.")
+
+#undef DEFINE_POSTER
```
out-of-phaze: 👎 I don't like this, personally. The define should be available for maps and modpacks to use.

out-of-phaze (changes requested):
It kind of feels like we should just have a poster spawner type, and define posters in a JSON file or something. Not blocking, though.

Also, did you test this? Posters use `[initial(name)] - [design.name]`, so that might cause problems with the presets.

code/game/objects/effects/decals/posters/bs12.dm
PsyCommando: I could always move it to the define folder and make it global if there's interest for it. 
But good practice dictates not to make macro available outside of where you need them as much as possible. Which is why I didn't make it a global macro to start with.

PsyCommando:
> It kind of feels like we should just have a poster spawner type, and define posters in a JSON file or something. Not blocking, though.
> 
> Also, did you test this? Posters use `[initial(name)] - [design.name]`, so that might cause problems with the presets.

The problem with using json, or a spawner is we don't get to see what the poster we're placing is in the map editor unless its values are known at compile time. Which is the reason why this PR was made in the first place. 

Like, in order to get to see what poster you've picked by using a json file for storing the poster data, you'd still have to declare each instances of each different posters with their name and icon, which already covers most of the data you'd have made the json file to contain in the first place.

Also about the name, I didn't even notice that? Was that a recent change? 
Because I mainly cherry picked this from downstream, and didn't notice it duplicating the name while testing it there... It might be a problem yeah.

out-of-phaze:
> Also about the name, I didn't even notice that? Was that a recent change?

No, it's five years old. I confirmed that it happens:
image

PsyCommando:
> > Also about the name, I didn't even notice that? Was that a recent change?
> 
> No, it's five years old. I confirmed that it happens: image

Oh well, I guess I just never noticed. That's been on my dev branch for several months now too..

code/game/objects/effects/decals/contraband.dm
```diff
@@ -79,20 +79,12 @@
 //############################## THE ACTUAL DECALS ###########################
 
 /obj/structure/sign/poster
-	name = "poster"
```
out-of-phaze, line 82: I'm not quite sure why this was unset.

code/game/objects/effects/decals/contraband.dm
```diff
@@ -101,11 +93,12 @@
 		else
 			poster_type = pick(decls_repository.get_decl_paths_of_subtype(/decl/poster))
 	set_poster(poster_type)
+	set_dir(placement_dir)
 
 /obj/structure/sign/poster/proc/set_poster(var/poster_type)
 	var/decl/poster/design = GET_DECL(poster_type)
-	SetName("[initial(name)] - [design.name]")
-	desc = "[initial(desc)] [design.desc]"
+	SetName("poster - [design.name]")
```
out-of-phaze: These should probably use a `base_name` and `base_desc` var like I mentioned on Discord. Hardcoding it here doesn't sit right with me.

out-of-phaze (dismissed):
The hardcoded names/descriptions should be put into separate vars, I think. Also, I think this needs to be squashmerged.

code/game/objects/effects/decals/contraband.dm
PsyCommando: Why do that when all the implemented posters all prefix their name with "poster", and nothing overrides the base poster class? It's a dead-end atom.

code/game/objects/effects/decals/contraband.dm
PsyCommando: Because that's never used. It used to slap the initial name into the generated name. It doesn't anymore.

code/game/objects/effects/decals/contraband.dm
out-of-phaze: Because if the initial name of the base poster class ever changes, they will become out of sync. It's a redundancy issue and means you have to do twice the work.

Also, it'd be trivial to subtype it for something like flyers that can be taken down and put back up by hand, for example—I was planning on doing something like that myself.

PsyCommando: What will become out of sync? The "poster" string is literally used only one time in that one proc? Unless you mean the descriptions?

Also, if you want me to implement something to help you with your own PR you should just ask. That would clear things up a bit.

code/game/objects/effects/decals/posters/bs12.dm
```diff
@@ -1,290 +1,60 @@
...
-/decl/poster/bay_58
-	icon_state="bsposter58"
-	name = "space carp information poster"
-	desc = "This poster showcases an old spacer saying on the dangers of migrant space carp."
-
-/decl/poster/bay_59
-	icon_state="bsposter59"
-	name = "poster - Miss Science 2299"
-	desc = "A large piece of space-resistant printed paper. This pin-up poster depicts a woman wearing a corporate labcoat, a bikini, and a sheepish grin. She's shyly posing atop some highly specialized research equipment."
+DEFINE_POSTER(bay_1,  "bsposter1",  "Unlucky Space Explorer",                          "This particular one depicts a skeletal form within a space suit.")
```
MistakeNot4892: Some of these are unreadable. I would suggest something like:
````
DEFINE_POSTER(                                                         \
    bay_1,                                                             \
    "bsposter1",                                                       \
    "Unlucky Space Explorer"                                           \
    "This particular one depicts a skeletal form within a space suit." \
)
````

File: code/game/objects/effects/decals/contraband.dm (hunk 3 above)
Comment (MistakeNot4892) [reply code/game/objects/effects/decals/contraband.dm]: It would take about 30 seconds to add:
````
/decl/poster
    var/base_name = "poster"
    var/base_desc = "A large piece of space-resistant paper."
...
    SetName("[design.base_name] - [design.name]")
    desc = "[design.base_desc] [design.desc]"
````
and now the system can be trivially expanded to support placards, signs, brochures, or specific kinds of poster that aren't 'space-resistant paper'.

It's really not worth arguing about and definitely not worth the implication that Haze is trying to force you to do their work for them.

Comment (PsyCommando) [reply code/game/objects/effects/decals/contraband.dm]: > It would take about 30 seconds to add:
> 
> ```
> /decl/poster
>     var/base_name = "poster"
>     var/base_desc = "A large piece of space-resistant paper."
> ...
>     SetName("[design.base_name] - [design.name]")
>     desc = "[design.base_desc] [design.desc]"
> ```
> 
> and now the system can be trivially expanded to support placards, signs, brochures, or specific kinds of poster that aren't 'space-resistant paper'.

But signs and brochures and etc already have their own atoms? That's why I'm saying it's a dead end class. :P
I'm not trying to be a jerk or anything, I just don't like implementing things that will almost certainly never be used?

> It's really not worth arguing about and definitely not worth the implication that Haze is trying to force you to do their work for them.

That's not what I was implying. What I was saying is, instead of trying to find a reason, just ask me if you'd like me to do that change for you? That's reasonable enough I think? 
I wasn't accusing anyone here. If it was something that was actually going to be used, I'd have done it. What I didn't agree with is having to write extra features that would not get used, and eventually end up in one of those "I removed useless stuff" PR.

That's all I was trying to clarify here. Is it being added just because, or if someone is actually going to use this?

If it's that important to you I'll make the change, but I have doubts it'll ever be useful on it's own.

Comment (MistakeNot4892) [reply code/game/objects/effects/decals/contraband.dm]: We don't have placards or brochures or flyers currently. Also ignoring what I said about specific kinds of poster.

It's not a dead end class, making it easily overridable is not the same as having useless feature content that adds maintainability overhead. 

Haze requested a change and told you they have a use case.

File: code/game/objects/effects/decals/posters/bs12.dm (hunk 4 above)
Comment (PsyCommando) [reply code/game/objects/effects/decals/posters/bs12.dm]: You can't do that with macros

Comment (MistakeNot4892) [reply code/game/objects/effects/decals/posters/bs12.dm]: Per Discord, yes you can.

Comment (PsyCommando) [reply code/game/objects/effects/decals/posters/bs12.dm]: I literally tried it doesn't pass compilation. (Also your example is missing a comma)
I haven't seen anything from discord? Can you sum it up in here?

File: code/game/objects/effects/decals/contraband.dm (hunk 3 above)
Comment (PsyCommando) [reply code/game/objects/effects/decals/contraband.dm]: Adding this a bit late, but it should be done now.

File: code/game/objects/effects/decals/posters/bs12.dm (hunk 4 above)
Comment (MistakeNot4892) [reply code/game/objects/effects/decals/posters/bs12.dm]: Pinged you with a link to the snippet. It works fine.

File: code/unit_tests/icon_tests.dm [hunk 5]
```diff
@@ -66,16 +66,16 @@
 	var/contraband_icons = icon_states('icons/obj/contraband.dmi')
 	var/list/invalid_posters = list()
 
-	var/list/all_posters = decls_repository.get_decls_of_subtype(/decl/poster)
-	for(var/poster_type in all_posters)
-		var/decl/poster/P = all_posters[poster_type]
+	var/list/all_posters = decls_repository.get_decls_of_subtype(/decl/poster_design)
```
MistakeNot4892, line 69: Not blocking, but this should be moved onto `/decl/poster_type/validate()` these days.

MistakeNot4892 (approved):
I'm still not a fan of the formatting of the poster defines but I can tidy that myself after merge.

PR #2960, opened 2023-02-08
Wall hitsound fixes and tweaks

comma:
## Description of changes
Fixed it never actually playing to people, because of different zone volume penalty Also now it gets hitsound from material, added a couple

Hitting walls should make sound

## Authorship
me

### DIFF
--- a/code/game/sound.dm
+++ b/code/game/sound.dm
@@ -55,7 +55,8 @@ var/global/const/FALLOFF_SOUNDS = 0.5
 
 	volume *= pressure_factor
 
-	if(istype(T,/turf/simulated) && istype(turf_source,/turf/simulated))
+	//Dense turfs are not in the zones, so they shouldn't be penalized here when source
+	if(!turf_source.density && istype(T,/turf/simulated) && istype(turf_source,/turf/simulated))
 		var/turf/simulated/sim_source = turf_source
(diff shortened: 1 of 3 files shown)


MistakeNot4892 (dismissed):

code/game/sound.dm
```diff
@@ -55,7 +55,8 @@ var/global/const/FALLOFF_SOUNDS = 0.5
 
 	volume *= pressure_factor
 
-	if(istype(T,/turf/simulated) && istype(turf_source,/turf/simulated))
+	//Dense turfs are not in the zones, so they shouldn't be penalized here when source
+	if(!turf_source.density && istype(T,/turf/simulated) && istype(turf_source,/turf/simulated))
```
out-of-phaze:
```suggestion
	if(!turf_source.blocks_air && istype(T,/turf/simulated) && istype(turf_source,/turf/simulated))
```
I'm not 100% sure about this, but it seems better to penalize it based off of whether or not it can participate in a zone (i.e. if it blocks air or not), rather than just using density as a proxy. Haven't tested, could be wrong.

out-of-phaze (changes requested):
A little uncertain about using `density` in that check, as per discussion on Discord.

out-of-phaze (approved):

PR #2963, opened 2023-02-10
Moving internals proc to mob level.

MistakeNot4892:
## Description of changes
Moves internals logic into its own file and condenses it on `/mob/living`.

Splitting this out of the inventory rewrite PR plus working towards destroying the /carbon mobtype.

## Authorship
Myself.

## Changelog
Nothing player-facing.


code/modules/mob/living/internals.dm
```diff
@@ -0,0 +1,136 @@
...
+	return
+
+/mob/living/proc/breathing_hole_covered()
+	var/obj/item/mask = get_equipped_item(slot_wear_mask_str)
+	. = (mask && (mask?.item_flags & ITEM_FLAG_AIRTIGHT))
+
+/mob/living/proc/get_possible_internals_sources()
+	. = list("back" = list(get_equipped_item(slot_back_str), "on"))
+
+/mob/living/proc/set_internals_to_best_available_tank(var/breathes_gas = /decl/material/gas/oxygen, var/list/poison_gas = list(/decl/material/gas/chlorine))
```
out-of-phaze: Having a stub here when it's also redefined later seems redundant and just adds overhead. See line 82.

code/modules/mob/living/internals.dm
```diff
@@ -0,0 +1,136 @@
+/mob/living/proc/set_internals(obj/item/tank/source, source_string)
```
out-of-phaze, line 1: This is overridden at the /mob/living level on line 71.

code/modules/mob/living/internals.dm
```diff
@@ -0,0 +1,136 @@
...
+	return
+
+/mob/living/proc/get_internals()
+	return
+
+// Set internals on or off. Implemented properly on /carbon
+/mob/living/proc/toggle_internals(var/mob/living/user)
```
out-of-phaze, line 15: This is overridden at the /mob/living level on line 39.

code/modules/mob/living/internals.dm
```diff
@@ -0,0 +1,136 @@
...
+		if(gear && (gear.item_flags & ITEM_FLAG_AIRTIGHT))
+			found_mask = TRUE
+			break
+
+	if(!found_mask)
+		to_chat(user, SPAN_WARNING("\The [src] does not have a suitable mask or helmet."))
+		return
+
+	// Find an internal source.
+	for(var/slot in global.internal_source_slots)
```
out-of-phaze: Why not check `get_possible_internals_sources()` here and remove `global.internal_source_slots` entirely? Currently this results in a weird disparity between someone turning on their own internals and someone else enabling it for them.

out-of-phaze (changes requested):
Some stubs exist and are redefined on the same level later in the file, which should be corrected.

Non-blocking suggestions: Feels like the PR title should say `living` or `mob/living` or `living mob` level. Also, maybe consider moving `var/obj/item/tank/internal = null` to the living level as well, so that we can remove the last of the carbon-level overrides in `internals.dm`?

code/modules/mob/living/internals.dm
MistakeNot4892: Did some refactoring around this to make both places use the same proc, and to fold held items into the base proc.

code/modules/mob/living/internals.dm
```diff
@@ -0,0 +1,121 @@
...
+			source_string = source.name
+		to_chat(src, SPAN_NOTICE("You are now running on internals from \the [source_string]."))
+		playsound(src, 'sound/effects/internals.ogg', 50, 0)
+	if(old_internal && !source)
+		to_chat(src, SPAN_NOTICE("You are no longer running on internals."))
+
+/mob/living/proc/get_internals()
+	return
+
+// Set internals on or off. Implemented properly on /carbon
```
out-of-phaze, line 14:
```suggestion
// Set internals on or off.
```
At this point the comment isn't very useful, so you could also just remove it entirely.

out-of-phaze (approved):
One issue with a comment and one future compatibility note, both non-blocking.

Future compatibility note: `parse_zone` is incompatible with modular zones like `BP_L_HAND_UPPER`, `BP_M_HAND`, etc. The way this breaks currently appears to be handled properly, however, in that if the slot name is null it only uses the name of the tank in the message. In the future, it may be wise to replace `parse_zone` with `GET_EXTERNAL_ORGAN(src, zone)?.name` or even `GET_EXTERNAL_ORGAN(src, zone)::name` (so that it's `left upper forelimb` instead of `Unbranded robotic left upper forelimb` or whatever), but for now this should be safe.

PR #2965, opened 2023-02-12
Fix modular prosthetic attachment

out-of-phaze:
## Description of changes
Adds the missing `has_children` entry to organ data.

You're now able to reattach modular prosthetic limbs.

## Authorship
Taken from Polaris, I believe.


MistakeNot4892 (approved):
Missed in the initial rewrite/port.

PR #2967, opened 2023-02-12
Resets organ status when bodytype is set in sync_colour_to_human()

out-of-phaze:
## Description of changes
Resets an organ's status when bodytype is set in `sync_colour_to_human()`, in case any species organ modifications rely on it.

Fixed some bodytype-dependent species organ modifications downstream, which I don't *think* is too uncommon of a pattern. It might be worth looking into moving it onto bodytype entirely, or adding per-bodytype modification stubs upstream.

## Authorship
Me.


MistakeNot4892 (approved):
This makes me a little bit uneasy but it should be fine.

PR #2968, opened 2023-02-12
Fix humans not getting default species settings

out-of-phaze:
## Description of changes
Reshuffles a lot of human/carbon default variables, including setting them to null by default so species vars override them.
Adds `override_existing` as an argument to `set_default_hairstyle()`. Sometimes you want to use it to fix invalid hairstyles, otherwise you want to force the hairstyle instead.
Species/mob defaults are now applied in `/mob/living/carbon/human/proc/setup` rather than `/datum/dna/proc/ResetUIFrom`.

Targeting dev because while I've tested the changes, it's nothing exhaustive.

Fixes blood type not being set properly on non-player mobs, i.e. mapped in or spawned with the spawn verb.
Fixes species defaults not being applied to spawned/mapped non-player mobs.
Properly integrates support for species-specific (facial) hair defaults.

## Authorship
Me, on my Signalis-themed fork/branch.

### DIFF
--- a/code/game/dna/dna2.dm
+++ b/code/game/dna/dna2.dm
@@ -141,15 +141,10 @@ var/global/list/datum/dna/gene/dna_genes[0]
 	SetUIState(DNA_UI_GENDER, character.gender!=MALE, 1)
 
 	// Hair
-	// FIXME:  Species-specific defaults pls
-	if(!character.h_style)
-		character.h_style = /decl/sprite_accessory/hair/bald
 	var/list/hair_types = decls_repository.get_decl_paths_of_subtype(/decl/sprite_accessory/hair)
 	SetUIValueRange(DNA_UI_HAIR_STYLE,  hair_types.Find(character.h_style),  length(hair_types), 1)
 
 	// Facial Hair
-	if(!character.f_style)
-		character.f_style = /decl/sprite_accessory/facial_hair/shaved
 	var/list/beard_types = decls_repository.get_decl_paths_of_subtype(/decl/sprite_accessory/facial_hair)
 	SetUIValueRange(DNA_UI_BEARD_STYLE, beard_types.Find(character.f_style), length(beard_types), 1)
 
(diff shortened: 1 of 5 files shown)


code/game/dna/dna2.dm
```diff
@@ -141,15 +141,10 @@ var/global/list/datum/dna/gene/dna_genes[0]
 	SetUIState(DNA_UI_GENDER, character.gender!=MALE, 1)
 
 	// Hair
-	// FIXME:  Species-specific defaults pls
-	if(!character.h_style)
```
MistakeNot4892, line 145: Should this not be setting it to the character species defaults?

code/modules/mob/living/carbon/human/human_defines.dm
```diff
@@ -1,12 +1,12 @@
 /mob/living/carbon/human
 
-	var/h_style = /decl/sprite_accessory/hair/bald
-	var/f_style = /decl/sprite_accessory/facial_hair/shaved
+	var/h_style
+	var/f_style
 
-	var/hair_colour =        COLOR_BLACK
```
MistakeNot4892, line 6: May be worth adding all these to species validation as the defaults were put in place to avoid stuff like bad icon operations due to null colour.

code/game/dna/dna2.dm
out-of-phaze: No. The character variables are set to the species default in `/setup()`, and then those values are used in this proc to populate the DNA datum.

EDIT: For these in particular, the relevant line is `human.dm:1377`, `species.set_default_hair(src, override_existing = FALSE, defer_update_hair = TRUE)`.

code/modules/mob/living/carbon/human/human_defines.dm
out-of-phaze: Yeah, I ran into that issue earlier, though mostly because I forgot `base_color`.

MistakeNot4892 (approved):

PR #2971, opened 2023-02-12
Readding removed logic from get_equipped_slot_for_item

MistakeNot4892:
Appears to have been missed by my examine rewrite PR.

### DIFF
--- a/code/modules/mob/living/inventory.dm
+++ b/code/modules/mob/living/inventory.dm
@@ -17,7 +17,7 @@
 	var/list/inventory_slots
 
 /mob/living/get_inventory_slots()
-	return inventory_slots
+	return global.all_inventory_slots // inventory_slots
 
 /mob/living/get_inventory_slot_datum(var/slot)
 	return LAZYACCESS(inventory_slots, slot) || LAZYACCESS(held_item_slots, slot)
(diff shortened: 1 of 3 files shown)


out-of-phaze (dismissed):
I noted that I'd prefer this to be an override for `get_inventory_slots()`, but since that would break other changes from the examine PR (if I understood the objection properly), this fix works fine for now.

EDIT: Also, maybe consider unit testing to make sure the various inventory/equip procs work as expected.

code/modules/mob/living/inventory.dm
```diff
@@ -17,7 +17,7 @@
 	var/list/inventory_slots
 
 /mob/living/get_inventory_slots()
-	return inventory_slots
+	return global.all_inventory_slots // inventory_slots
```
out-of-phaze, line 20: As discussed on Discord this won't work since the code that (used to) use `get_inventory_slots()` expected an associative list of slot name -> inventory datum. That said, not really a blocking issue since `get_inventory_slots()` is no longer used.

out-of-phaze (approved):

out-of-phaze:
Merging since it's been more than 24 hours since my first review and this is a pretty critical bugfix.

PR #2972, opened 2023-02-12
Adding QDELETED and null loc checking to mob Bump().

MistakeNot4892:
## Description of changes
Adds QDELETED and null loc checks to `/mob/living/Bump`

Should prevent bumping from moving atoms that are qdeleted on Bump() out of nullspace.

## Authorship
Myself.

## Changelog
Nothing player-facing.


afterthought2 (approved):
Would be better if we got rid of the spawn.

Also not sure that the null loc check is "correct," but really don't care about edge cases abusing this.

PR #2978, opened 2023-02-14
/carbon removal [IDB IGNORE] [MDB IGNORE]

MistakeNot4892:
DNM, opening this for CI and some limited review due to the size.

## Description of changes
Splits `/carbon` logic up between `/human` and `/living` and removes associated code.

Simplifies the type tree and makes behavior available to `/mob/living`.

## Authorship
Myself.

## Changelog
Nothing player-facing.


code/__defines/mobs.dm
```diff
@@ -223,7 +223,7 @@
 #define MOB_PULL_SAME 2
 #define MOB_PULL_LARGER 3
 
-//carbon taste sensitivity defines, used in mob/living/carbon/proc/ingest
+//carbon taste sensitivity defines, used in mob/living/proc/ingest
```
out-of-phaze:
```suggestion
//taste sensitivity defines, used in mob/living/proc/ingest
```

out-of-phaze (changes requested):
This review covers only comments that still mention carbons. In addition to the one I marked inline:
- `code/game/machinery/_machines_base/machinery.dm:113` should remove the mention of carbon lifeforms.
- `code/modules/power/singularity/singularity.dm:402` should remove the mention of carbon mobs. (Notably, I think this was already incorrect.)
- `code/game/objects/items/weapons/handcuffs.dm:51` is incorrect. (The variable should also probably not be named C now.)

MistakeNot4892:
This is going to need heavy rebasing and will be split into multiple PRs anyway so closing to get it off the list.

PR #2984, opened 2023-02-15
Codex entries generated per atom will be kept on the atom.

MistakeNot4892:
## Description of changes
- Atoms will keep a ref to any specific codex entry they generate instead of caching them on the subsystem.
- The subsystem will no longer track atom-specific codex entries and will not run uniqueness checks on them.

Examining two MREs causes the codex to complain.

## Authorship
@out-of-phaze for pointing out the issue, me for the code.

## Changelog
Nothing player-facing.

### DIFF
--- a/code/controllers/subsystems/initialization/codex.dm
+++ b/code/controllers/subsystems/initialization/codex.dm
@@ -9,6 +9,7 @@ SUBSYSTEM_DEF(codex)
 
 	var/list/all_entries =       list()
 	var/list/entries_by_path =   list()
+	var/list/entries_by_ref =    list()
 	var/list/entries_by_string = list()
 	var/list/index_file =        list()
 	var/list/search_cache =      list()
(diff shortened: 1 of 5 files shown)


code/controllers/subsystems/initialization/codex.dm
```diff
@@ -9,6 +9,7 @@ SUBSYSTEM_DEF(codex)
 
 	var/list/all_entries =       list()
 	var/list/entries_by_path =   list()
+	var/list/entries_by_ref =    list()
```
out-of-phaze:
```suggestion
```
Unused var?

out-of-phaze (changes requested):
May be worth gating `SScodex.all_entries -= src` in `/datum/codex_entry/Destroy` (`code/modules/codex/entries/_codex_entry.dm:63`) behind `store_codex_entry`, that's a large list and a lot of atoms with temporary entries getting deleted could impact things like explosion processing.

code/controllers/subsystems/initialization/codex.dm
MistakeNot4892: Yeah added before I decided to do it per-atom.

out-of-phaze (approved):
The Destroy code implies that more than one entry in `index_file` and `search_cache` can be/contain `src`. Assuming that's true, this is fine, although since `cached -= src` can't remove src if it isn't there and that does a search anyway, it might be better to just remove the `src in cached` check on line 74—I'd prefer if someone double-check that before any changes are made, though.

No blocking concerns.

MistakeNot4892:
The only thing that should be destroying codex entries is atom cleanup anyway, so the expanded Destroy() is more for the sake of thoroughness than any practical concern.

PR #2986, opened 2023-02-15
Make design database constructable

out-of-phaze:
## Description of changes
Adds a circuit, circuit design, and construct_state for the design database.

You can't currently create a new design database for, say, a ship or whatever. This fixes that.

## Authorship
Me.

### DIFF
--- a/code/modules/research/design_database.dm
+++ b/code/modules/research/design_database.dm
@@ -14,6 +14,9 @@ var/global/list/default_initial_tech_levels
 	icon_state = "blackbox"
 	density = TRUE
 	anchored = TRUE
+	construct_state = /decl/machine_construction/default/panel_closed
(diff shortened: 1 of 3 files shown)


code/modules/research/design_database.dm
```diff
@@ -14,6 +14,7 @@ var/global/list/default_initial_tech_levels
 	icon_state = "blackbox"
 	density = TRUE
 	anchored = TRUE
+	construct_state = /decl/machine_construction/default/panel_closed
```
afterthought2, line 17: You probably want to also set 
```
	uncreated_component_parts = null
	stat_immune = 0
```
(or, possibly, not include the extra components on the circuit if you prefer). May also be worth checking the power draw vars on it.

afterthought2 (approved):

PR #2987, opened 2023-02-16
[MDB IGNORE] Fix simple pipes with wrong dir on generic tanker map

out-of-phaze:
## Description of changes
This map isn't tested in unit testing because it's defined as an away site. I accidentally included it and it got caught.
Also adds it to the away sites unit testing map.

As an aside, I'm really against this unit test existing, and considering making a precommit hook to parse the map and correct the dirs at commit-time instead.

Better unit testing coverage and unit test compliance on this map.

## Authorship
Me.

### DIFF
--- a/maps/away_sites_testing/away_sites_testing.dm
+++ b/maps/away_sites_testing/away_sites_testing.dm
@@ -5,6 +5,7 @@
 
 	#include "../../mods/content/government/away_sites/icarus/icarus.dm"
 	#include "../../mods/content/government/ruins/ec_old_crash/ec_old_crash.dm"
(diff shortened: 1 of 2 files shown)


afterthought2:
At this point, the pipe dir test is purely stylistic, all systems should handle the "wrong" directions properly. I don't have strong feelings about whether this needs to be unit test enforced or not, but definitely don't feel it merits a precommit hook.

Not sure what's up with map diff bot, but I can't successfully re-trigger it.

MistakeNot4892 (dismissed):
Happy to just forcemerge this over the top of CI.

MistakeNot4892 (approved):

PR #2990, opened 2023-02-17
Fix lighting not multiplying for ghosts

out-of-phaze:
## Description of changes
Moves the lighting planemaster to `/mob` level so that it's always present on mobs.
The lighting planemaster is responsible for properly turning the `BLEND_OVERLAY` lighting overlays (which is necessary because of darksight) into a multiplicative overlay that darkens the map. It was only applied on `/mob/living`, however, meaning `/mob/observer` would see the lighting colors overlaid on the map instead of properly darkening it.

image

## Authorship
Me.

### DIFF
--- a/code/_onclick/hud/hud.dm
+++ b/code/_onclick/hud/hud.dm
@@ -15,9 +15,6 @@
 		hud_used = new hud_type(src)
 	else
 		hud_used = new /datum/hud(src)
-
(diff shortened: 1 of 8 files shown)


MistakeNot4892 (approved):

MistakeNot4892:
Selfmerging for a test update.

PR #2995, opened 2023-02-19
Level data tweaks + lateral z connections

PsyCommando:
## Description of changes
Since I'm trying not to make a mega huge PR anymore, here's the first part of the splitting off of exoplanet gen stuff from exoplanet overmap objects. 

I'm moving the /obj/abstract/level_data stuff into a /datum/level_data instead, since basically, the code would put its z coordinate into a var, then dump the obj into nullspace for it's entire lifetime and never care about it being an /obj ever again.

I've also moved border generation from exoplanet code into level_data, since border generation is useful for lateral z-connections anyways. And added support for creating looping edges or connection edge on any level.

**Left to do:**
 ~~* Test border gen~~
~~* Add lateral connections unit test?~~ (Can't right now exoplanets not in unit test anyways)


code/controllers/subsystems/mapping.dm
```diff
@@ -141,9 +141,10 @@ SUBSYSTEM_DEF(mapping)
 // Z-Level procs after this point.
 /datum/controller/subsystem/mapping/proc/get_gps_level_name(var/z)
 	if(z)
-		var/obj/abstract/level_data/level = levels_by_z[z]
-		if(level?.name)
-			return level.get_gps_level_name()
+		var/datum/level_data/level = levels_by_z[z]
+		. = level?.get_display_name()
```
MistakeNot4892: This should be allowed to runtime - every level should have a z-level data object. Not sure why I `?.`'d it in the first place.

code/game/machinery/cryopod.dm
```diff
@@ -222,8 +222,8 @@
 		if(length(possible_locations))
 			newz = pick(possible_locations)
 	if(!newz)
-		var/obj/abstract/level_data/level = SSmapping.increment_world_z_size(/obj/abstract/level_data/space)
-		newz = level?.my_z
+		var/datum/level_data/level = SSmapping.increment_world_z_size(/datum/level_data/space)
+		newz = level?.level_z
```
MistakeNot4892: Similar to above, this should be left to runtime as we should be able to assume every level has z-data. Also otherwise it's going to proceed with an insane `newz`

code/modules/multiz/level_data.dm
```diff
@@ -1,132 +1,159 @@
...
+	var/border_filler// = /turf/unsimulated/mineral
+	/// If set we will put a looping edge on every unconnected edge of the map. If null, will not loop unconnected edges. If an unconnected edge is facing a connected edge, it will be instead filled with "border_filler" instead, if defined.
+	var/loop_turf_type// = /turf/unsimulated/mimc_edge/transition/loop
+	/// The turf type to use for zlevel lateral connections
+	var/transition_turf_type = /turf/unsimulated/mimic_edge/transition
+
+	// *** Atmos ***
+	/// Temperature of standard exterior atmosphere.
+	var/exterior_atmos_temp = T20C
+	/// Gaxmix datum returned to exterior return_air. Set to assoc list of material to moles to initialize the gas datum.
```
MistakeNot4892: `Gasmix`

code/modules/multiz/level_data.dm
```diff
@@ -1,132 +1,159 @@
...
-		SSmapping.accessible_z_levels[num2text(my_z)] = template.accessibility_weight
-	SSmapping.player_levels |= my_z
-
-/obj/abstract/level_data/Destroy(var/force)
-	if(force)
-		new type(locate(round(world.maxx/2), round(world.maxy/2), my_z))
-		return ..()
-	return QDEL_HINT_LETMELIVE
+/datum/level_data/Destroy(force)
+	//Since this is a datum that lives inside the SSmapping subsystem, I'm not sure if we really need to prevent deletion. It was fine for the obj version of this, but not much point now?
```
MistakeNot4892: It does make me a bit uneasy not to gate it off, but you're right in that there shouldn't be any way to accidentally qdel this (versus obj getting exploded or eaten by the material processor or something).

code/modules/multiz/level_data.dm
```diff
@@ -1,132 +1,159 @@
...
-/obj/abstract/level_data/proc/copy_from(var/obj/abstract/level_data/old_level)
+/datum/level_data/proc/copy_from(var/datum/level_data/old_level)
 	return
 
-/obj/abstract/level_data/proc/initialize_level()
-	var/change_turf = (created_base_turf_type && created_base_turf_type != world.turf)
-	var/change_area = (created_base_area_type && created_base_area_type != world.area)
+///Initialize the turfs on the z-level
+/datum/level_data/proc/initialize_new_level()
+	var/picked_turf = filler_turf? filler_turf : base_turf //Pick the filler_turf for filling if it's set, otherwise use the base_turf
```
MistakeNot4892: `var/picked_turf = filler_turf || base_turf` would be a little bit cleaner

code/modules/multiz/level_data.dm
```diff
@@ -1,132 +1,159 @@
...
+	/// Filled by map gen on init. Indicates where the accessible level area starts past the transition edge.
+	var/tmp/level_inner_y
+	/// Filled by map gen on init. Indicates the width of the accessible area within the transition edges.
+	var/tmp/level_inner_with
+	/// Filled by map gen on init.Indicates the height of the accessible area within the transition edges.
+	var/tmp/level_inner_height
+
+	// *** Lighting ***
 	/// Set to false to leave dark
 	var/take_starlight_ambience = TRUE
```
MistakeNot4892: While you're in the region, could you rename `take_starlight_ambience` to `use_global_exterior_ambience`?

code/modules/overmap/exoplanets/exoplanet_atmosphere.dm
```diff
@@ -10,13 +10,13 @@
...
 	if(habitability_class == HABITABILITY_IDEAL)
-		for(var/obj/abstract/level_data/level_data in zlevels)
+		for(var/datum/level_data/level_data in zlevels)
 			level_data.exterior_atmos_temp = target_temp
 			level_data.exterior_atmosphere = list(
 				/decl/material/gas/oxygen = MOLES_O2STANDARD,
 				/decl/material/gas/nitrogen = MOLES_N2STANDARD
 			)
-			level_data.setup_level_data()
+			level_data.setup_level_data() //#FIXME: That's not very nice! Calling this again, when it should have been called before?
```
MistakeNot4892, line 19: If I remember correctly, this has to be called here due to exoplanet generation happening after `SSmapping` has initialized. If you can move exoplanet gen onto SSmapping, this call won't be necessary.

MistakeNot4892: Ah yep it's also due to the ordering of exoplanet level gen and how convoluted it is - it has to know the temperature and gasmix before it initializes any of the z-level data objects.

code/modules/multiz/level_data.dm
```diff
@@ -1,132 +1,159 @@
...
+				locate(LD.level_max_width, TRANSITIONEDGE, LD.level_z))
+		if(WEST)
+			return block(
+				locate(1,                  LD.level_max_height - TRANSITIONEDGE, LD.level_z),
+				locate(LD.level_max_width, LD.level_max_height,                  LD.level_z))
+
+///Keeps details on how to generate, maintain and access a zlevel
+/datum/level_data
+	///Name displayed to the player to refer to this level in user interfaces and etc. If null, one will be generated.
+	var/display_name
```
MistakeNot4892: Personally I would prefer this to be left as `name` because it means that lists of `/datum/level_data` will format nicely when passed to `input()`

MistakeNot4892:
Generally this looks good so far - I've left some small nitpicks, and I haven't read through the border logic yet, but the principles seem sound.

code/controllers/subsystems/mapping.dm
PsyCommando: Is it possible that gps/overmap tries to get names before exoplanets may have generated? Because, that would be my guess.
But, yeah, I'll change that to runtime if level is null.

code/game/machinery/cryopod.dm
PsyCommando: kk

code/modules/multiz/level_data.dm
PsyCommando: Yeah, I was feeling a bit uneasy as well, but I just couldn't find any reasons to keep gating it. I could always just log whenever those get destroyed/created?

code/modules/multiz/level_data.dm
PsyCommando: kk

code/modules/multiz/level_data.dm
PsyCommando: I'm going to be messing with that anyways when I split planet gen from the overmap marker. 
So, I'll change it in that other PR to avoid extra diff.

code/modules/multiz/level_data.dm
PsyCommando: kk

code/modules/overmap/exoplanets/exoplanet_atmosphere.dm
PsyCommando: Alright. I've been able to move that anyways in the next PR.

PsyCommando:
Moved a lot of stuff from the exoplanet PR that I've done to level data.

code/modules/shuttles/docking_beacon.dm
```diff
@@ -312,11 +312,11 @@
 			break
 		area_turf = get_step(area_turf, dir)
 
-	// Otherwise, use the planetary or world area.
+	// Otherwise, use the level or world area.
 	if(!base_area)
-		var/obj/effect/overmap/visitable/sector/exoplanet/planet = global.overmap_sectors[num2text(z)]
-		if(istype(planet))
-			base_area = ispath(planet.planetary_area) ? planet.planetary_area : planet.planetary_area.type
+		var/datum/level_data/LD = SSmapping.levels_by_z[z]
```
MistakeNot4892, line 317: We're trying to avoid initialisms for variables these days I think, particularly when you're replacing a properly named variable.

code/modules/maps/_map_template.dm
```diff
@@ -1,4 +1,4 @@
-/datum/map_template
+/datum/map_template //#FIXME: Probably should be typed /decl since it's a singleton.
```
MistakeNot4892: Nope, this needs to stay a datum for hotloading/secrets to instantiate them down the track.

code/modules/maps/_map_template.dm
```diff
@@ -135,8 +138,8 @@
 	init_shuttles(shuttle_state, map_hash, initialized_areas_by_type)
 	after_load()
 	for(var/z_index = bounds[MAP_MINZ] to bounds[MAP_MAXZ])
-		var/obj/abstract/level_data/level = SSmapping.levels_by_z[z_index]
-		level.post_template_load(src)
+		var/datum/level_data/level = SSmapping.levels_by_z[z_index]
+		level.after_template_load(src)
```
MistakeNot4892: `post_template_load` literally means `after_template_load`, what's the rationale for changing? Couldn't `before_template_load` be `pre_template_load`?

code/modules/multiz/level_data.dm
```diff
@@ -136,78 +241,307 @@ INITIALIZE_IMMEDIATE(/obj/abstract/level_data)
...
+		if(!length(neighbor.connected_levels))
+			continue
+		neighbor.find_connected_levels(found)
+
+///Returns the instance of the base area for this level
+/datum/level_data/proc/get_base_area_instance()
+	var/area/found = locate(base_area)
+	if(found)
+		return found
+	else if(ispath(base_area))
```
MistakeNot4892: Extraneous `else` here and below.
````
/datum/level_data/proc/get_base_area_instance()
	var/area/found = locate(base_area)
	if(found)
		return found
	if(ispath(base_area))
		return new base_area
	return world.area
````

File: code/modules/multiz/level_data.dm [hunk 13]
```diff
@@ -136,78 +241,307 @@ INITIALIZE_IMMEDIATE(/obj/abstract/level_data)
...
+		return new base_area
+	else
+		return world.area
```
MistakeNot4892: `world.area` isn't an instance.

code/modules/maps/_map_template.dm
```diff
@@ -114,6 +114,9 @@
 	var/list/bounds = list(1.#INF, 1.#INF, 1.#INF, -1.#INF, -1.#INF, -1.#INF)
 	var/list/atoms_to_initialise = list()
 	var/shuttle_state = pre_init_shuttles()
+	for(var/z_index = bounds[MAP_MINZ] to bounds[MAP_MAXZ])
```
MistakeNot4892: `bounds` isn't populated usefully here, you're setting it three lines up.

code/controllers/subsystems/mapping.dm
```diff
@@ -196,3 +245,56 @@ SUBSYSTEM_DEF(mapping)
...
+	if(LD.level_flags & ZLEVEL_PLAYER)
+		player_levels  |= LD.level_z
+	if(LD.level_flags & ZLEVEL_SEALED)
+		sealed_levels  |= LD.level_z
+	return TRUE
+
+/datum/controller/subsystem/mapping/proc/unregister_level_data(var/datum/level_data/LD)
+	if(levels_by_z[LD.level_z] == LD)
+		//Temporarily clear the level z from the list if we're in it.
+		levels_by_z[LD.level_z] = null
```
MistakeNot4892, line 291: `Temporarily`? Won't this cause runtimes until something else populates it?

MistakeNot4892:
Pulled it down and tested it with my Polaris port branch - generally works quite nicely, though the map will need some revision to make room for the transition edges. Some general comments below, not necessarily blockers for the most part:

- The transition turf band is not wide enough to completely hide the map edge from the adaptive view size. It might be worth making it fade out towards the hard map edge.
- Lighting doesn't propagate across map edges. I don't know if this would remotely be feasible to implement, but it would make things look much less strange.
- Transitioning between levels could possibly put you on the transition turf (via direct loc set to avoid weirdness) then move you onto the final turf after a 0-wait. This would give the appearance of a glide. At the moment you chess-mode across the transition.
- Mimic turf isn't always dense, I can walk through the mimiced stone walls. I also don't transition to the other level.
- Ghosts don't transition across levels.

Runtimes from testing. I suspect the latter two are due to the transition edge overlapping a pre-mapped structure and messing with the atoms as a result.
````
[11:04:08] Runtime in projectile.dm, line 468: Cannot read null.z
proc name: pixel move (/obj/item/projectile/proc/pixel_move)
````
````
[11:00:00] Runtime in turf_mimic_edge.dm, line 284: Cannot read null.level_inner_max_y
proc name: shared transition edge bumped (/proc/shared_transition_edge_bumped)
src: null
call stack:
shared transition edge bumped(the world\'s edge (139,7,1) (/turf/unsimulated/mimic_edge/transition), the enhanced power cell (/obj/item/cell/super), null)
the world\'s edge (139,7,1) (/turf/unsimulated/mimic_edge/transition): Entered(the enhanced power cell (/obj/item/cell/super), null)
the enhanced power cell (/obj/item/cell/super): forceMove(the world\'s edge (139,7,1) (/turf/unsimulated/mimic_edge/transition))
the enhanced power cell (/obj/item/cell/super): forceMove(the world\'s edge (139,7,1) (/turf/unsimulated/mimic_edge/transition))
the enhanced power cell (/obj/item/cell/super): forceMove(the world\'s edge (139,7,1) (/turf/unsimulated/mimic_edge/transition))
the enhanced power cell (/obj/item/cell/super): forceMove(the world\'s edge (139,7,1) (/turf/unsimulated/mimic_edge/transition))
the enhanced power cell (/obj/item/cell/super): dropInto(the world\'s edge (139,7,1) (/turf/unsimulated/mimic_edge/transition))
the building materials (/obj/item/stock_parts/building_material): remove material(/obj/item/cell (/obj/item/cell), 1)
the battery backup (rapid) (/obj/item/stock_parts/power/battery/buildable/turbo): on refresh(the robot recharging station (/obj/machinery/recharge_station))
the robot recharging station (/obj/machinery/recharge_station): RefreshParts()
the robot recharging station (/obj/machinery/recharge_station): RefreshParts()
the robot recharging station (/obj/machinery/recharge_station): Initialize(1, 0, 1)
the robot recharging station (/obj/machinery/recharge_station): Initialize(1)
the robot recharging station (/obj/machinery/recharge_station): Initialize(1)
the robot recharging station (/obj/machinery/recharge_station): Initialize(1)
the robot recharging station (/obj/machinery/recharge_station): Initialize(1)
Atoms (/datum/controller/subsystem/atoms): InitAtom(the robot recharging station (/obj/machinery/recharge_station), /list (/list))
Atoms (/datum/controller/subsystem/atoms): InitializeAtoms()
Atoms (/datum/controller/subsystem/atoms): Initialize(346118)
Atoms (/datum/controller/subsystem/atoms): DoInitialize(346118)
````
````
[11:00:00] Runtime in turf_mimic_edge.dm, line 286: Cannot read null.level_inner_min_y
proc name: shared transition edge bumped (/proc/shared_transition_edge_bumped)
src: null
call stack:
shared transition edge bumped(the world\'s edge (129,250,3) (/turf/unsimulated/mimic_edge/transition), APC power cell (/obj/item/cell/apc), null)
the world\'s edge (129,250,3) (/turf/unsimulated/mimic_edge/transition): Entered(APC power cell (/obj/item/cell/apc), null)
APC power cell (/obj/item/cell/apc): forceMove(the world\'s edge (129,250,3) (/turf/unsimulated/mimic_edge/transition))
APC power cell (/obj/item/cell/apc): forceMove(the world\'s edge (129,250,3) (/turf/unsimulated/mimic_edge/transition))
APC power cell (/obj/item/cell/apc): forceMove(the world\'s edge (129,250,3) (/turf/unsimulated/mimic_edge/transition))
APC power cell (/obj/item/cell/apc): forceMove(the world\'s edge (129,250,3) (/turf/unsimulated/mimic_edge/transition))
APC power cell (/obj/item/cell/apc): dropInto(the world\'s edge (129,250,3) (/turf/unsimulated/mimic_edge/transition))
the building materials (/obj/item/stock_parts/building_material): remove material(/obj/item/cell (/obj/item/cell), 1)
the battery backup (/obj/item/stock_parts/power/battery/buildable): on refresh(North Mining Outpost Exterior ... (/obj/machinery/power/apc))
North Mining Outpost Exterior ... (/obj/machinery/power/apc): RefreshParts()
North Mining Outpost Exterior ... (/obj/machinery/power/apc): Initialize(1, 0, 1)
North Mining Outpost Exterior ... (/obj/machinery/power/apc): Initialize(1, null, 1)
North Mining Outpost Exterior ... (/obj/machinery/power/apc): Initialize(1, null, 1)
North Mining Outpost Exterior ... (/obj/machinery/power/apc): Initialize(1, null, 1)
North Mining Outpost Exterior ... (/obj/machinery/power/apc): Initialize(1, null, 1)
North Mining Outpost Exterior ... (/obj/machinery/power/apc): Initialize(1, null, 1)
Atoms (/datum/controller/subsystem/atoms): InitAtom(North Mining Outpost Exterior ... (/obj/machinery/power/apc), /list (/list))
Atoms (/datum/controller/subsystem/atoms): InitializeAtoms()
Atoms (/datum/controller/subsystem/atoms): Initialize(346118)
Atoms (/datum/controller/subsystem/atoms): DoInitialize(346118)
````

**Comment by MistakeNot4892** [comment]
With regards to the projectile runtime, I suspect that this proc should handle transition turfs:
````
/datum/vector_loc/proc/return_turf()
	return turf
````

**Comment by MistakeNot4892** [comment]
https://github.com/MistakeNot4892/Nebula/tree/polaris This is my working branch of Neb + the Southern Cross map + this PR.

**Comment by MistakeNot4892** [comment]
> Runtimes from testing. I suspect the latter two are due to the transition edge overlapping a pre-mapped structure and messing with the atoms as a result.

Can confirm this, after I bumped everything in 7 tiles on each transition side the runtimes stopped.

File: code/modules/shuttles/docking_beacon.dm (hunk 9 above)
Comment (PsyCommando) [reply code/modules/shuttles/docking_beacon.dm:317]: Is that because of that one guy that came in the chat and had a fit about vars being initials?

Personally, I feel initials, when used for a thing you'd just name by it's class name anyways ensure we don't get name collisions with actually important wider scope vars. 
Local, atom, and global vars don't have any naming conventions and can easily clash, so I'd be kinda wary of doing that, unless we get some naming conventions in place as well.

File: code/modules/maps/_map_template.dm (hunk 10 above)
Comment (PsyCommando) [reply code/modules/maps/_map_template.dm]: kk

File: code/modules/maps/_map_template.dm (hunk 11 above)
Comment (PsyCommando) [reply code/modules/maps/_map_template.dm]: There's no rationale, I just matched their names, but didn't realize which name was there first.

File: code/modules/maps/_map_template.dm (hunk 14 above)
Comment (PsyCommando) [reply code/modules/maps/_map_template.dm]: That's true. It's basically leftover copy-pasta. 
I don't think anything even implements before_template_load.

File: code/controllers/subsystems/mapping.dm (hunk 15 above)
Comment (PsyCommando) [reply code/controllers/subsystems/mapping.dm:291]: This only happens when a level_data datum is deleted. Which should only happen if a level_data is replaced at runtime, or we're shutting the game down. So if something hasn't replaced it with something by then and the game isn't shutting down, then it really should crash.

Comment (MistakeNot4892) [reply code/controllers/subsystems/mapping.dm:291]: In that case I think the comment is misleading and should probably be omitted.

**Comment by PsyCommando** [comment]
Well, glad it worked even on a polaris fork XD

>     * The transition turf band is not wide enough to completely hide the map edge from the adaptive view size. It might be worth making it fade out towards the hard map edge.

The transition border issue is applicable to anything relying on ``TRANSITIONEDGE`` really. So space travel also has the problem where changing view range will allow you to see the map edge. Making the edge bigger depending on the max view range would probably work. But you'd have to take into account how much map space is being lost for that, because that could get significant.

>     * Lighting doesn't propagate across map edges. I don't know if this would remotely be feasible to implement, but it would make things look much less strange.

I'd reaally love to have lighting work across the border, but I'm not really sure how lighting works entirely. I did an attempt where I was replacing get_step() in the lighting code with a custom proc that would take into account the border, but then I realized that a lot of things used get_dist(), range, orange, locate, etc, as well, and figuring distances across connected maps can get a bit expensive. Ideally that would have to be done in a separate PR and I'd need some backup on the lighting front. Ideally, I'd love to tap into z-mimic's proxy system for interacting with things across the edge and etc..

>     * Transitioning between levels could possibly put you on the transition turf (via direct loc set to avoid weirdness) then move you onto the final turf after a 0-wait. This would give the appearance of a glide. At the moment you chess-mode across the transition.

Yeah, I was contemplating doing that so the glide would show up. But I don't know if changing the player's loc won't also trigger entered/crossed and etc, along with exited. That could potentially cause problems. If you find a good way to do it let me know!

>     * Mimic turf isn't always dense, I can walk through the mimiced stone walls. I also don't transition to the other level.

I think I know what's going on.. I made mimic turfs non-solid by default because using Bumped() for teleporting stuff also had some side effects. So I'm using Entered, but since it won't teleport you to a solid turf you can walk on the mimic of the wall.. I guess I'll have to look into using bumped again or have the mimic turfs match the density of their target.

>     * Ghosts don't transition across levels.

That's expected behavior currently. Ghost also can't transit in space. I tried to keep the behaviors in-line with each-others.

> Runtimes from testing. I suspect the latter two are due to the transition edge overlapping a pre-mapped structure and messing with the atoms as a result.

Currently, the border turf don't handle cable, connections, pipes, structure/walls smoothing, AO, lighting, mob interactions, throwing, etc.. You're not supposed to map anything onto it.
So the thing runtiming because an APC was on a transition edge makes sense. That will need to be implemented later, because of the amount of work that's going to be needed to abstract distances and adjacency.

File: code/controllers/subsystems/mapping.dm (hunk 15 above)
Comment (PsyCommando) [reply code/controllers/subsystems/mapping.dm:291]: Yeah, I'm changing it.

**Comment by PsyCommando** [comment]
> With regards to the projectile runtime, I suspect that this proc should handle transition turfs:
> 
> ```
> /datum/vector_loc/proc/return_turf()
> 	return turf
> ```

The problem with projectiles that's tied to the other problems right now, is distance handling. Also, you cannot aim at individual things anyways because there are no proxy atom like with z-mimic. Raytrace projectiles will also not work.
(vis_locs might offer some kind of solution I guess)

File: code/modules/shuttles/docking_beacon.dm (hunk 9 above)
Comment (MistakeNot4892) [reply code/modules/shuttles/docking_beacon.dm:317]: I don't know the guy you're referring to. I just don't particularly like the idea of replacing a properly named var with an initialism. Generally if initialisms are used, they should be temporary indices, and they should be in lowercase. The uppercase initialisisms are an old BYOND code thing that isn't something to be emulated.

File: code/modules/shuttles/docking_beacon.dm [hunk 16]
```diff
@@ -312,11 +312,11 @@
...
+		if(istype(LD))
+			base_area = LD.get_base_area_instance()
 		else
 			base_area = world.area
```
MistakeNot4892: This is going to be setting it to a type while the one above is setting it to an instance.

MistakeNot4892:
>I think I know what's going on.. I made mimic turfs non-solid by default because using Bumped() for teleporting stuff also had some side effects. So I'm using Entered, but since it won't teleport you to a solid turf you can walk on the mimic of the wall.. I guess I'll have to look into using bumped again or have the mimic turfs match the density of their target.

One issue with this is going to be trying to interact with a turf on the other side of the gap - mining, for example, you can't dig across levels.

I am starting to think we need a `get_step()` or `get_turf()` that is lateral multiz aware, and we need to work that into spots like pipe/cable connections, projectile propagation and clicking.

MistakeNot4892:
I would be happy to sign off on merging this for further work if you can add some kind of wrapping or qdel for projectiles leaving the level and fix the other `world.area` uses. This is going to need a ton of refinement and there's not much point putting it off until that's all in place. The system generally works fine in my testing - that is good enough for me for now.

code/modules/shuttles/docking_beacon.dm
PsyCommando: You're right.

PsyCommando:
> I am starting to think we need a `get_step()` or `get_turf()` that is lateral multiz aware, and we need to work that into spots like pipe/cable connections, projectile propagation and clicking.

That would be cool. We'd need to split abstracted turf access from physical turf access. As in, accessing a turf virtually means we'd access the turf another turf represents, while physically would access the actual proxy turf. Might be a good idea to ask lohikar for details on how to do proxies properly..

I did test trying to replace get_step and get_dist, but get_dist in particular proved to cause unreported runtimes that completely broke level loading from a quick search and replace. 
Also, I tried modifying face_atom to take into account vis_locs, but I didn't get the logic right I think.

PsyCommando:
> I would be happy to sign off on merging this for further work if you can add some kind of wrapping or qdel for projectiles leaving the level and fix the other `world.area` uses. This is going to need a ton of refinement and there's not much point putting it off until that's all in place. The system generally works fine in my testing - that is good enough for me for now.

How should I implement that projectile check? I'm not sure where to start with that one.

Alright, but I'd like to have the exoplanet stuff ( #2999 ) in before there being further refinements because it's a blocker downstream, and keeping it up to date is going to be a big time drain.

PsyCommando:
Another thing of note, random_map generated maps don't seem to want to respect the borders they were given and will try to replace the transition edges if they have the chance, and that cause all sorts of issues.

MistakeNot4892:
Just had a thought. What happens if a turf being mimicked is changeturf'd?

code/modules/multiz/level_data.dm
```diff
@@ -136,78 +241,302 @@ INITIALIZE_IMMEDIATE(/obj/abstract/level_data)
...
+	if(ispath(base_area))
+		return new base_area
+	return locate(world.area)
+
+////////////////////////////////////////////
+// Level Data Spawner
+////////////////////////////////////////////
+
+/// Mapper helper for spawning a specific level_data datum with the map as it gets loaded
+/obj/abstract/landmark/level_data_spawner
```
MistakeNot4892: Now that I think about it, this shouldn't be `/landmark`. You should really just `..(); return INITIALIZE_HINT_QDEL` in `Initialize()` and be done with it.

MistakeNot4892:
`How should I implement that projectile check?`
Possibly just have projectiles qdel when moving onto a mimic turf.

PsyCommando:
> Just had a thought. What happens if a turf being mimicked is changeturf'd?

Depends on how vis_contents works for turfs? so far it doesn't seem to be a problem, but it's not something that I've tested in depth.

code/modules/multiz/level_data.dm
PsyCommando: Alright, repathed it.

PsyCommando:
> `How should I implement that projectile check?` Possibly just have projectiles qdel when moving onto a mimic turf.

I just made the turfs ignore projectiles. So they can act like they normally do when hitting the edge of the world.

MistakeNot4892 (approved):
Tentative approval with the expectation that this will be refined over time.

PR #2997, opened 2023-02-22
Add limb reagent circulation system

out-of-phaze:
## Description of changes
When you are injected with a syringe, the reagents will first enter the limb, then flow up to their parent limb, before finally entering the mob's reagents once it reaches the heart (one processing cycle after it reaches the external organ containing the heart).

Tourniquets can be used to block the flow out of (but not into) a limb. This means placing a tourniquet on the arm will prevent reagents from it or the hand entering the chest (and therefore the mob's reagents). They also slow bleeding.

Injections on the legs are more difficult than on the arms, and injections on the groin, chest, and head (neck) are even more difficult, only being able to be done by skilled medical professionals. With a tourniquet on the targeted limb, the minimum skill to successfully find a vein is reduced, as well as the minimum skill to cancel rather than syringestab on a failed attempt. It also makes finding a vein faster.

TODO:
- [x] Make injecting with a syringe add to limb reagents
- [x] Make limb reagents flow to parent
- [x] Make external organ with heart flow to mob reagents
- [x] Refactor syringestab and injectMob to properly support /mob/living rather than lying with typehints
- [x] Audit CHEM_INJECT uses and replace with `inject_external_organ` where necessary
- [ ] Research better solutions to the `needs_process()` + `bad_external_organs` problem than just calling `recheck_bad_external_organs()` directly
- [ ] Obtain tourniquet sprites (on-limb, world, inventory, etc)
- [ ] Add phlebotomy tourniquets to prevent spread of reagents
- [ ] Make tourniquets slow (not stop, these are the tourniquets for when you get blood drawn, not for losing an arm) bleeding on limbs
- [ ] Make injecting the groin, chest, and head (neck) more difficult skill-wise than on limbs, maybe even a minimum skill of trained or experienced?
- [ ] Make tourniquets lower the minimum skill for successful injections on a limb and increase the speed of finding a vein.
- [ ] Add a way to remove reagents from a limb (maybe just add it to dialysis?)

Discussed on Discord, I think it'd provide some additional depth to medicine and provide new ways to deal with problems like, say, a poisonous mob bite on a limb.

It could also have some neat interactions with tourniquets in the future, like them making it easier to find a vein for injection, and some reagents having effects that depend on the volume of reagent in the bloodstream rather than just the metabolized amount (drugs, for example) so that using a tourniquet creates a stronger effect.

## Authorship
Me.

## Changelog
:cl:
add: Added phlebotomy tourniquets. Tourniquets slow bleeding, make it easier/faster to find an injection site, and prevent the spread of reagents past the limb they're applied on. They can only be applied on limbs.
add: Reagents injected into a limb now take time to travel up the bloodstream and to the heart before taking effect.
/:cl:


MistakeNot4892:
Closing as stale, feel free to reopen when updated.

PR #2998, opened 2023-03-02
Added alternative neopronoun for gender-neutral nonplural.

MistakeNot4892:
## Description of changes
Adds a new neutral singular pronoun set (ey/eir/em). This is an alternative to they/them.

Adds an example of a non-BYOND pronouns set, expands on options a bit.

## Authorship
Myself.

## Changelog
:cl:
add: Added ey/eir/em singular neutral pronoun.
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/cards_ids_syndicate.dm
+++ b/code/game/objects/items/weapons/cards_ids_syndicate.dm
@@ -64,7 +64,7 @@
 	entries[++entries.len] = list("name" = "Fingerprint Hash",	"value" = fingerprint_hash)
 	entries[++entries.len] = list("name" = "Name", 				"value" = registered_name)
 	entries[++entries.len] = list("name" = "Photo", 			"value" = "Update")
-	entries[++entries.len] = list("name" = "Sex", 				"value" = sex)
+	entries[++entries.len] = list("name" = "Gender", 				"value" = card_gender)
 	entries[++entries.len] = list("name" = "Factory Reset",		"value" = "Use With Care")
 	data["electronic_warfare"] = electronic_warfare
 	data["entries"] = entries
@@ -186,11 +186,11 @@
 				set_id_photo(user)
 				to_chat(user, "<span class='notice'>Photo changed.</span>")
 				. = 1
-			if("Sex")
-				var/new_sex = sanitize(input(user,"What sex would you like to put on this card?","Agent Card Sex", sex) as null|text)
-				if(!isnull(new_sex) && CanUseTopic(user, state))
-					src.sex = new_sex
-					to_chat(user, "<span class='notice'>Sex changed to '[new_sex]'.</span>")
+			if("Gender")
+				var/new_gender = sanitize(input(user,"What gender would you like to put on this card?","Agent Card Gender", card_gender) as null|text)
+				if(!isnull(new_gender) && CanUseTopic(user, state))
+					src.card_gender = new_gender
+					to_chat(user, "<span class='notice'>Gender changed to '[new_gender]'.</span>")
 					. = 1
 			if("Factory Reset")
 				if(alert("This will factory reset the card, including access and owner. Continue?", "Factory Reset", "No", "Yes") == "Yes" && CanUseTopic(user, state))
@@ -210,7 +210,7 @@
 					SetName(initial(name))
 					registered_name = initial(registered_name)
 					unset_registered_user()
-					sex = initial(sex)
+					card_gender = initial(card_gender)
 					military_branch = initial(military_branch)
 					military_rank = initial(military_rank)
 					to_chat(user, "<span class='notice'>All information has been deleted from \the [src].</span>")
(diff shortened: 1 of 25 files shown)


out-of-phaze (dismissed):
Double-checked with some people that we don't need to change is/does/s/es for this pronoun set, and as far as I can tell they're supposed to be grammatically singular, so it checks out.

Would be nice to have a more encyclopedic source to link in the comment, but the existing links on Wikipedia have mostly died out. Until we find a better source with a better overview of the full grammar, this link seems fine.

afterthought2:
Can you check that these are selectable in appearance changer and/or prefs, and that the UI for doing so is reasonable? 

In particular, the different formatting of `name` from the other pronoun sets, the fact that the name does not really distinguish it from `NEUTER` (they are both neuter/signular, after all, the distinguishing feature is that one implies "no" gender, other is gender-agnostic), and `formal_term` being unset all give me pause.

out-of-phaze:
> (they are both neuter/signular, after all, the distinguishing feature is that one implies "no" gender, other is gender-agnostic)

This is incorrect. `NEUTER` is the singular neutral *inanimate* pronoun, whereas this is a singular neutral *personal* pronoun. Inanimate pronouns are generally only used for objects or for people who explicitly want them to be used; it's considered impolite to do so with a stranger. Both the Elverson singular personal neuter pronouns ("ey/em/eir," used in this PR) and the plural neuter "they/them/their" are typically used for either non-gendered individuals who choose to use them, or for individuals of ambiguous gender (e.g. concealed, unknown, or hypothetical).

However, since the Elverson set is singular so grammatically it's simpler to implement. That's why Lambdamoo used the original Spivak pronouns (which is just "e/em/eir", almost identical but it's harder to distinguish "e" from "he" in certain accents), because they didn't require changing any verbs like using the gender-neutral plural would, but the pronouns system exists to make this easier in SS13. Despite not having the technical limitations that led to their implementation, they've gained usage for other reasons now, so it seems worth implementing.

~~I do think maybe changing the name for `/decl/pronouns/neuter` would be fine; something like "inanimate" is descriptive, since English only has the one inanimate pronoun.~~ We can't change it because it's used to map from the built-in BYOND `gender` variable to our custom pronoun sets in `/proc/get_pronouns_by_gender(var/gender)`. It might be good to have a separate user-facing name for it though, since I think `neuter` is a confusing name for what BYOND uses it for.

`formal_term` is inherited from `/decl/pronouns`, though, and is just "other", like with plural and neuter inanimate. It seems to exist purely for bureaucratic reasons like crew records and IDs and is never used to map back from `formal_term` to a pronoun set. If we want to enforce uniqueness for that we should add it in a unit test, but there's no conceptual reason they would need to be unique since it's only a one-way mapping.

Also, the appearance changer just capitalizes the `name` var for everything in `available_pronouns` for the species, which is probably why the `name` var in this PR was already capitalized. It's lowercase for the others for parity with the BYOND gender defines. This kind of indicates that we may just want a `display_name` variable to use instead.

MistakeNot4892:
I've reworked things a bit. Pronouns now display like this in chargen and the appearance changer:

image

afterthought2 (dismissed):

code/game/objects/items/weapons/cards_ids_syndicate.dm
```diff
@@ -186,11 +186,11 @@
...
 				. = 1
-			if("Sex")
-				var/new_sex = sanitize(input(user,"What sex would you like to put on this card?","Agent Card Sex", sex) as null|text)
-				if(!isnull(new_sex) && CanUseTopic(user, state))
-					src.sex = new_sex
-					to_chat(user, "<span class='notice'>Sex changed to '[new_sex]'.</span>")
+			if("Gender")
+				var/new_gender = sanitize(input(user,"What gender would you like to put on this card?","Agent Card Gender", card_gender) as null|text)
+				if(!isnull(new_gender) && CanUseTopic(user, state))
+					src.card_gender = new_gender
```
out-of-phaze, line 192: `Sex: Yes` is dead, long live `Gender: No`

code/modules/pronouns/_pronouns.dm
```diff
@@ -2,6 +2,7 @@
 	var/name = PLURAL
 	var/formal_term = "other"
```
out-of-phaze:
```suggestion
	var/bureaucratic_term = "other"
```

out-of-phaze (changes requested):
Looks good, but the `formal_term` -> `bureaucratic_term` changes haven't been made yet?

out-of-phaze (approved):
I don't see any issues, although I'm kind of confused by the existing Skrell code for `get_gender()`. Why does it check headtail length if the pronouns exist? In fact, that only ever seems to be used in one place, during character creation, and it checks if the associated pronouns are in the usable pronouns list. MALE and FEMALE aren't allowed for Skrell, so that code does nothing as far as I know? Is it just outdated from a bad port?

MistakeNot4892:
Looks like the Skrell get_gender() is a Bay holdover that doesn't fit with the use of headtails to delineate caste.

PR #2999, opened 2023-03-02
Move exoplanet generation

PsyCommando:
Depends on: #2995 

## Description of changes
Splits exoplanet generation, processing, and data from their overmap marker.

* Moved exoplanet gen to a subclass of map_template, since they were a really good fit for abstracting level generation, and categorizing planets and etc..
* Fauna and flora gen/handling are now generated via separate datums. Those can be overridden and the overridden type can be specified in the map_template for allowing better control over how those work for specific planets. 
* Moved day cycle processing to a subsystem.
* Made weather processing handling mostly independent of exoplanets.  

**TODO**
* ~~Generalize planetoid data a bit more so it can be used downstream for planet loaded from a map file.~~
* Fix weather materials handling.
* ~~Fix planet rock colors.~~
* ~~Fix ambient lighting not applying properly.~~
* ~~Allow for lateral z-levels to generate and connect when specified.~~
* ~~Maybe move respawn handling to a "mob ai director" of sort, and make it independent of exoplanets. Maybe a subsystem?~~ (A bit out of scope for this PR.)
* ~~Major clean up and reorganization. I'm still mostly just at the prototype stage.~~
* ~~Add/tweak unit tests for exoplanets.~~ (Unit tests currently skip exoplanets.)
* ~~Fix strata conditions not being respected.~~

Also, I'm looking for some feedback, because some things I feel could be done a lot better, but I'm not quite sure what to focus on. 

With this change, exoplanets can be initialized at the same time as any other templates, since there's no dependencies on when the overmap initialize anymore.
And the overmap marker for exoplanet is now a lot less bloated with junk.


PsyCommando:
@MistakeNot4892 uh, why did you mark it as needing a rebase?

MistakeNot4892:
Because you had a `Level data datum + improv` commit that looked like a duplicate. I thought it needed rebasing since the dep got merged. :P

MistakeNot4892:
Is this waiting on review or is it still WIP?

PsyCommando:
> Is this waiting on review or is it still WIP?

Should be good for review!

code/controllers/subsystems/daycycle.dm
```diff
@@ -0,0 +1,129 @@
+
```
MistakeNot4892: Weird leading linebreaks?

code/controllers/subsystems/weather.dm
```diff
@@ -28,3 +29,41 @@ SUBSYSTEM_DEF(weather)
...
+	var/obj/abstract/weather_system/WS = LAZYACCESS(weather_by_z, topmost_level_data.level_z)
+	if(WS)
+		unregister_weather_system(WS)
+		qdel(WS)
+	//Create the new weather system and let it register itself
+	new /obj/abstract/weather_system(locate(1, 1, topmost_level_data.level_z), topmost_level_data.level_z, initial_state)
+
+///Registers a given weather system obj for getting updates by SSweather.
+/datum/controller/subsystem/weather/proc/register_weather_system(var/obj/abstract/weather_system/WS)
+	if(LAZYACCESS(weather_by_z, WS.z))
```
MistakeNot4892: Shouldn't this be in the loop below?

code/modules/xgm/xgm_gas_mixture.dm
```diff
@@ -519,3 +528,15 @@
 	var/M = get_total_moles()
 	if(M)
 		return get_mass()/M
+
+///Returns a color blended from all materials the gas mixture contains
+/datum/gas_mixture/proc/get_overall_color()
```
MistakeNot4892, line 537: Please cache this a la reagent mix colours, these RGB blend procs are costly.

code/modules/overmap/sectors.dm
```diff
@@ -208,6 +208,9 @@ var/global/list/known_overmap_sectors
 	closeToolTip(usr) //No reason not to, really
 	..()
 
+/obj/effect/overmap/visitable/proc/get_topmost_level_data()
+	return SSmapping.levels_by_z[map_z[1]]
```
MistakeNot4892: Is `map_z` always ordered? Are both of these lists always lists? This feels unsafe.

code/modules/overmap/planetoids/planetoid_skybox.dm
```diff
@@ -0,0 +1,112 @@
...
+/obj/effect/overmap/visitable/sector/planetoid/proc/generate_skybox_image()
+	//Everything is done in a separate proc for each layers, to allow for overriding for a variety of planetoids
+	skybox_image = generate_skybox_planetoid_surface()
+	ASSERT(skybox_image)
+	skybox_image.pixel_x = rand(0,64)
+	skybox_image.pixel_y = rand(128,256)
+	skybox_image.appearance_flags = RESET_COLOR
+
+	skybox_image.underlays += generate_skybox_planetoid_halo()
+	skybox_image.overlays  += generate_skybox_planetoid_clouds()
```
MistakeNot4892:
````
skybox_image.overlays = list(
    etc,
    etc,
    etc
)
````
would likely be a bit faster, though it's admittedly a microopt for this case.

File: code/modules/multiz/turf_mimic_edge.dm [hunk 6]
```diff
@@ -189,12 +201,10 @@
 	return
 
 /turf/exterior/mimic_edge/get_vis_contents_to_add()
-	. = ..()
-	var/turf/NT = mimic_x && mimic_y && mimic_z && locate(mimic_x, mimic_y, mimic_z)
-	if(NT)
-		opacity = NT.opacity
-		//log_debug("[src]([x],[y],[z]) mirroring [NT]([NT.x],[NT.y],[NT.z])")
-		LAZYADD(., NT)
+	. = shared_mimic_edge_get_add_vis_contents(src, get_mimic_turf(), ..())
```
MistakeNot4892, line 204: Heck of a proc name.

code/modules/multiz/turf.dm
```diff
@@ -159,3 +159,322 @@
 
 /turf/exterior/open/cannot_build_cable()
 	return 0
+
+// ////////////////////////////////
+// // Mimic Edges
```
MistakeNot4892: What's with this file? Commented out entirely?

code/modules/multiz/level_data.dm
```diff
@@ -525,13 +523,19 @@ INITIALIZE_IMMEDIATE(/obj/abstract/level_data_spawner)
 
 /datum/level_data/mining_level/asteroid
 	base_turf = /turf/simulated/floor/asteroid
+	level_generators = list(
+		/datum/random_map/automata/cave_system,
+		/datum/random_map/noise/ore
+	)
 
 /datum/level_data/mining_level/generate_level()
 	//#FIXME: This config option is very ambiguous. Most RNG stuff doesn't care about it. Might be worth removing?
```
MistakeNot4892: Main purpose of this config option is to keep init times down for testing, which isn't really that important nowadays. I would prefer to see it generalized to disable all map generation (ie. before build_away_sites() and build_exoplanets()) rather than removed. Possibly rename it to `enable_roundstart_map_generation`.

code/modules/mining/mine_turfs.dm
```diff
@@ -49,6 +49,7 @@
 /turf/simulated/floor/asteroid/is_plating()
 	return !density
 
+//#TODO: This should probably be generalised?
```
MistakeNot4892, line 52: I have a branch somewhere to kill this turf type...

code/modules/materials/geology/_strata.dm
```diff
@@ -5,9 +5,11 @@
 	var/list/ores_rich
 	var/default_strata_candidate = FALSE
 	var/maximum_temperature = INFINITY
+	///The types of material to be used for rock walls for this strata. May have a weight as value for each material types
+	var/list/rock_material_types
```
MistakeNot4892: This is the purpose of `base_materials` on `/decl/strata`.

MistakeNot4892:
Github is struggling with this one. If you could split out stuff like the weather subsystem changes and the SSdaynight cycle addition into their own PRs it might chug less.

code/modules/xenoarcheaology/datums/engravings_flavor.dm
```diff
@@ -0,0 +1,86 @@
+///Contains generated xenorach flavor text for engravings. Was moved out of exoplanet gen.
+/datum/xenoarch_engraving_flavor
+	//#TODO: This stuff could definitely be datumized into some /decl?
```
MistakeNot4892, line 3: From glancing over the code it looks like this could just be made a decl with a couple of changes from instantiating it to GET_DECL().

code/controllers/subsystems/weather.dm
```diff
@@ -28,3 +29,41 @@ SUBSYSTEM_DEF(weather)
...
+		CRASH("Trying to register another weather system on the same z-level([WS.z]) as an existing one!")
+	LAZYDISTINCTADD(weather_systems, WS)
+	LAZYINITLIST(weather_by_z)
```
MistakeNot4892: Treating this like a lazylist, using lazyaccess etc. is potentially problematic since you're using numerical indices; I would just make them straight list access and assume it's a valid list, maybe reindex it on world z expansion like the other lists that use z level as a numerical index.

code/modules/multiz/turf.dm
PsyCommando: I think something went wrong when I merged upstream. All that stuff got moved to code/modules/multiz/turf_mimic_edge.dm. I'll get rid of that comment block.

code/modules/multiz/turf_mimic_edge.dm
PsyCommando: It's really ``shared_mimic_edge_get_additional_vis_contents``, but it was getting a bit long. :P

code/modules/overmap/planetoids/planetoid_skybox.dm
PsyCommando: The main idea was just to split those things in smaller more manageable bits. 
And the skybox is generated exactly once per exoplanets. So idk how much it would matter yeah. 
Is that something you'd want changed?

code/modules/overmap/sectors.dm
PsyCommando: You're right about that. Currently the topmost level is usually the highest z in the sector's zlevels list because of how z stacks works. 

So, in short, easy way to fix this is to do ``max(map_z)`` instead of ``map_z[1]``. But, if we rework z stacks to have non-sequential up/down level connections eventually, that wouldn't work. 
It would be nice to have a standard way of accessing z stacks we could use instead of taking shortcuts like I did. Currently it's mostly done through ``global.z_levels`` and the ``/obj/abstract/map_data`` object, but those are kind of subject to change with z mimic ports.

code/modules/xgm/xgm_gas_mixture.dm
PsyCommando: You mean ``/datum/reagents/proc/get_color()`` right?
I can do that, but /datum/gas_mixture are a lot less well encapsulated than /datum/reagents. A majority of things still change the gas list directly instead of using the procs, and knowing when to clear the cached color is a bit more unreliable.

code/controllers/subsystems/weather.dm
PsyCommando: That's a good point..

code/controllers/subsystems/weather.dm
PsyCommando: You mean adding a call in SSmapping's ``increment_world_z_size()`` proc and get SSweather to rebuild its weather_by_z list?
I could try that.

code/controllers/subsystems/daycycle.dm
PsyCommando: It's probably just when I tried writing a description comment, and just ended up deleting it.

code/modules/materials/geology/_strata.dm
PsyCommando: I'll have to take a deeper look at this. I think there was a reason, but it's been a while. I'll see what I can do later today.

code/modules/mining/mine_turfs.dm
PsyCommando: By killing, do you mean you ported it's features to a more generic type of turf, or you just deleted it and all it's usages?

code/modules/multiz/level_data.dm
PsyCommando: When talking about map generation, do you mean solely like noise maps used to generate terrain features, or like map stitching randomly picked away sites/ruins? 
Because, I assumed the purpose of this var was to prevent random noise terrain gen from happening, seeing as it was preventing that on mining turfs?

code/modules/xenoarcheaology/datums/engravings_flavor.dm
PsyCommando: I could probably leave this as stub, and make another PR to decl-ify that?

code/modules/mining/mine_turfs.dm
MistakeNot4892: Unifying the features with /turf yeah.

code/modules/multiz/level_data.dm
MistakeNot4892: All gen of any kind that could bloat out init time. The sole purpose of the config var is to skip optional overhead.

code/controllers/subsystems/weather.dm
MistakeNot4892: Yeah essentially.

code/modules/xenoarcheaology/datums/engravings_flavor.dm
MistakeNot4892: Sure.

code/modules/maps/template_types/random_exoplanet/random_planet_level_data.dm
```diff
@@ -0,0 +1,143 @@
...
+		return
+	var/area/A = get_base_area_instance()
+	//Make sure we're not going to rename the world's base area
+	if(!istype(A, world.area))
+		global.using_map.area_purity_test_exempt_areas |= A.type //Make sure we add any of those, so unit tests calm down when we rename
+		A.SetName("[location_name]")
+
+///Try to spawn the given amount of ruins onto our level. Returns the template types that were spawned
+/datum/level_data/planetoid/proc/seed_ruins(var/budget = 0, var/list/potentialRuins)
+	//#TODO: Fill in allowed area from a proc or something
```
out-of-phaze, line 91:
```suggestion
	if(!length(potentialRuins))
		log_world("Ruin loader was given no ruins to pick from.")
		return list()
	//#TODO: Fill in allowed area from a proc or something
```
This guards against `pick(candidates_ruins)` runtiming with `pick from empty list` if `potentialRuins` is empty.

out-of-phaze (changes requested):
While testing I noticed some runtimes; these are the fixes to solve them.

atoms_init.dm - add this override to fix the runtimes with grass decals being deleted by ruin loading.
```dm
/atom/movable/EarlyDestroy(force = FALSE)
	loc = null // should NOT use forceMove, in order to avoid events
	return ..()
```

helper_landmarks.dm - move `/obj/abstract/landmark/clear/Initialize()` logic to `/obj/abstract/landmark/clear/LateInitialize()` to avoid a null reference error on `material.type` in `/turf/exterior/wall/proc/dismantle_wall()`.

reader.dm - add `/obj/abstract/landmark/exoplanet_spawn,` to `/dmm_suite/proc/types_to_delete()` in order to fix plants and animals spawning inside ruins. (This is used if `clear_contents = TRUE` and an object of the type exists on one of the turfs placed by the ruin.)

code/modules/maps/template_types/random_exoplanet/random_planet.dm
```diff
@@ -0,0 +1,369 @@
...
+		return lvl_data
+	lvl_data = buid_z_stack(direction_from_root, adjacent_level_data, new_level_data_types, gen_data)
+	if(amount < 2)
+		return lvl_data
+	//If amount if bigger than 2, build another adjacent stack to us
+	lvl_data += build_adjacent_z_stacks(amount - 1, direction_from_root, lvl_data, new_level_data_types, gen_data)
+	return lvl_data
+
+///Create a new z-level stack that's connected to an existing z stack, on the given direction.
+/datum/map_template/planetoid/proc/buid_z_stack(var/direction_from_root, var/list/adjacent_level_data, var/list/new_level_data_types, var/datum/planetoid_data/gen_data)
```
out-of-phaze:
```suggestion
/datum/map_template/planetoid/proc/build_z_stack(var/direction_from_root, var/list/adjacent_level_data, var/list/new_level_data_types, var/datum/planetoid_data/gen_data)
```

PsyCommando:
> While testing I noticed some runtimes; these are the fixes to solve them.
> 
> atoms_init.dm - add this override to fix the runtimes with grass decals being deleted by ruin loading.
> 
> ```dm
> /atom/movable/EarlyDestroy(force = FALSE)
> 	loc = null // should NOT use forceMove, in order to avoid events
> 	return ..()
> ```
> 
> helper_landmarks.dm - move `/obj/abstract/landmark/clear/Initialize()` logic to `/obj/abstract/landmark/clear/LateInitialize()` to avoid a null reference error on `material.type` in `/turf/exterior/wall/proc/dismantle_wall()`.
>
> reader.dm - add `/obj/abstract/landmark/exoplanet_spawn,` to `/dmm_suite/proc/types_to_delete()` in order to fix plants and animals spawning inside ruins. (This is used if `clear_contents = TRUE` and an object of the type exists on one of the turfs placed by the ruin.)

Alright, thanks! I'll try to get around to fixing that this week hopefully.

out-of-phaze:
I was wrong about the pick(candidates_ruins) runtime, it occurs after the last candidate is picked and removed. The check would need to be in the for loop or the for loop condition.

code/modules/multiz/level_data.dm
PsyCommando: But, won't that means we'll skip completely testing level gen during unit tests though? I'm seeing defines that force change the var to skip map gen as long as UNIT_TEST is defined.
Like, is it really worth the shortened testing time if we skip testing something kinda important like that? Picking and placing ruins, generating exoplanet terrain, generating mining data, etc..

EDIT: okay, I think I was wrong about that because the var is force changed after away sites and exoplanets are generated.

code/modules/materials/geology/_strata.dm
PsyCommando: Nah it's actually completely unused. So I'm removing it.

code/controllers/subsystems/weather.dm
PsyCommando: Removed all the lazy stuff.

code/modules/overmap/planetoids/planetoid_skybox.dm
MistakeNot4892: No strong feelings, your call.

code/modules/overmap/planetoids/planetoid_skybox.dm
```diff
@@ -0,0 +1,112 @@
...
+/obj/effect/overmap/visitable/sector/planetoid/proc/has_atmosphere()
+	var/datum/planetoid_data/PD = SSmapping.planetoid_data_by_id[planetoid_id]
+	var/datum/level_data/LD = SSmapping.levels_by_id[PD.surface_level_id]
+	return !isnull(LD.exterior_atmosphere)
+
+/obj/effect/overmap/visitable/sector/planetoid/proc/get_atmosphere_color()
+	var/list/colors = list()
+	//#TODO: Not sure why it's summing up all the z-level atmos color? I really doubt this was ever tested too...
+	for(var/lvl in map_z)
+		var/datum/level_data/level_data = map_z[lvl]
```
out-of-phaze: This runtimes, I tested this fix and it should work.
```suggestion
		var/datum/level_data/level_data = SSmapping.levels_by_z[lvl]
```

out-of-phaze (changes requested):
Will do more testing with the updated version soon.

code/controllers/subsystems/daycycle.dm
PsyCommando: fixed

code/controllers/subsystems/weather.dm
PsyCommando: fixed

code/modules/xgm/xgm_gas_mixture.dm
PsyCommando: done

code/modules/overmap/sectors.dm
PsyCommando: Fixed by trying to grab the map_data for the stack or the max zlevel index in map_z.

code/modules/overmap/planetoids/planetoid_skybox.dm
PsyCommando: I think I'll leave it like that for now.

code/modules/multiz/turf.dm
PsyCommando: Removed the commented out code.

code/modules/multiz/level_data.dm
PsyCommando: Renamed the option and made it apply to all the obvious map gen stuff I could find.

code/modules/maps/template_types/random_exoplanet/random_planet.dm
PsyCommando: Woops, good catch XD

code/modules/overmap/planetoids/planetoid_skybox.dm
PsyCommando: Kk

code/modules/maps/template_types/random_exoplanet/random_planet.dm
PsyCommando: Applied it manually since I had to rename it in a bunch of places.

PsyCommando:
> Github is struggling with this one. If you could split out stuff like the weather subsystem changes and the SSdaynight cycle addition into their own PRs it might chug less.

Hmm, well, I've looked at it, and it doesn't seem like the amount of work needed to port that to the old system really makes sense here? If I make it a stand alone PR, I have to make it compile and test it works, which is at least a 1-2 weeks job. And then I have to re-implement it and fix conflicts so it works with the new changes and test again.

Is there any ways we could reach a compromise on this?

code/modules/maps/template_types/random_exoplanet/fauna_generator.dm
```diff
@@ -0,0 +1,273 @@
...
+
+	for(var/i = 1 to round(repopulate_fauna_threshold - length(live_fauna)))
+		if(prob(90))
+			continue
+		var/turf/simulated/T = pick_area_turf(spawn_area, list(/proc/not_turf_contains_dense_objects))
+		try_spawn_fauna(T)
+
+	time_last_repop = REALTIMEOFDAY
+	//Half the repop threshold each times we empty the respawn list
+	if(LAZYLEN(respawn_queue) <= 0)
```
comma, line 224: Wouldn't this cause threshold to drop p fast even if not a lot of animals are killed? If just 1 animal dies per 5 minutes it'll be depopulated in 20 minutes (right away in casae of megafauna)

code/modules/maps/template_types/random_exoplanet/random_planet_subtemplates.dm
```diff
@@ -0,0 +1,30 @@
...
+			continue
+		if(blacklist & ruin_tags)
+			continue
+		possible_subtemplates += ruin
+
+	if(!length(possible_subtemplates))
+		return //If we don't have any ruins, don't bother
+
+	for(var/datum/level_data/planetoid/LD in created_level_data)
+		if(prob(25))
```
comma: Maybe check that we do have other zlevels here before skipping? Otherwise there's 25% chance of a planet without any ruins placed

code/modules/maps/template_types/random_exoplanet/fauna_generator.dm
PsyCommando: I don't really know, that was the old fauna code I ported into the fauna generator. So I assumed it worked.

code/modules/maps/template_types/random_exoplanet/random_planet_subtemplates.dm
PsyCommando: That's a fair point. But I don't really know what's the expectation on exoplanet ruins. Like what's the minimum?

comma: Non zero at least, I don't see the need for this check at all tbh. Empty planets are boring to visit, we got budget to limit  amount of features, no need for random chance to skip budgeting completely

code/modules/maps/template_types/random_exoplanet/fauna_generator.dm
comma: Well, if that's old code can be figured out later I guess

code/modules/maps/template_types/random_exoplanet/random_planet_subtemplates.dm
PsyCommando: The check is so ruins aren't placed only on the surface level.

comma: It doesn't check if it's on surface or not, there's still 25% to skip the lower zlevel too, or skip entirely if it's 1z planet
There's some sort of budget splitting right after the skip too, does that not make it so ruins aren't placed only on surface?

Also pretty much all ruins are mapped with being surface in mind, I think their distribution should be addressed later with some sort of tag setup, don't need to force them to be underground this hard

out-of-phaze: Every single ruin except the drill site and *maybe*  the radioactive anomaly shrine and the monolith expects to be placed on the surface, having either been built there or crashed on the surface.

PsyCommando: Yeah, I changed it a few hours ago to just put all ruins onto the surface for now.

comma (dismissed):

comma (approved):

out-of-phaze (approved):
Will merge and PR a fix for chlorine exoplanet sprites later, since it's not gamebreaking.

PR #3000, opened 2023-03-06
Added a basic helper for managed instances.

MistakeNot4892:
## Description of changes
Adds a global proc for grabbing a managed shared instance of a type (or id-based type/args instance).

Needed for de-declifying some systems and moving stuff into the secrets system.

## Authorship
Myself.

## Changelog
Nothing player-facing.

### DIFF
--- a/code/controllers/subsystems/managed_instance.dm
+++ b/code/controllers/subsystems/managed_instance.dm
@@ -1,0 +1,47 @@
+SUBSYSTEM_DEF(managed_instances)
+	name = "Managed Instances"
+	flags = SS_NO_FIRE | SS_NO_INIT
+	var/list/_managed_instance_cache = list()
+
+/datum/controller/subsystem/managed_instances/proc/get(var/instance_type, var/cache_id, var/cache_category = "default", var/list/instance_args)
+	if(!cache_id)
+		cache_id = instance_type
+	if(isnull(cache_id)) // No type and no id supplied, no point going further.
+		CRASH("Attempted to retrieve a managed instance with no supplied type or ID.")
+	if(isnum(cache_id)) // The cache is an assoc list of id to instance, so numerical IDs are out.
+		CRASH("Attempted to retrieve a managed instance with numerical ID ([cache_id]).")
+
+	var/category_list = _managed_instance_cache[cache_category]
+	if(category_list)
+		. = category_list[cache_id]
+	else
+		category_list = list()
+		_managed_instance_cache[cache_category] = category_list
+
+	if(!.)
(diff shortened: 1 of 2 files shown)


code/_helpers/managed_instances.dm
```diff
@@ -0,0 +1,27 @@
...
+
+	. = category_list[cache_id]
+	if(!.)
+		var/datum/managed_instance
+		if(length(instance_args))
+			managed_instance = new instance_type(arglist(instance_args))
+		else
+			managed_instance = new instance_type()
+		category_list[cache_id] = managed_instance
+		managed_instance.ManagedInstanceInitialize() // We do this after storing to avoid circular creation loops.
```
afterthought2: I'd send the args into here too, ideally people wouldn't use `New`.

EDIT: alternatively it's possible to _forbid_ args in some sense, which is more singleton-like but might not match the design goals here.

EDIT2: actually, would suggest making the first arg to this (but probably not to `New`) be `cache_id`.

afterthought2:
One thing you could consider is signing up for destroyed events of managed instances and having them crash + clean up the ref. If you do want to have these be destroyable there should be some special API for that, and destroyed events are low overhead.

MistakeNot4892:
Made it a subsystem so I could use event registration/callbacks, added the suggested stack trace and cleanup. Also adjusted when/how args are passed.

code/controllers/subsystems/managed_instance.dm
```diff
@@ -0,0 +1,47 @@
+SUBSYSTEM_DEF(managed_instances)
+	name = "Managed Instances"
+	flags = SS_NO_FIRE | SS_NO_INIT
+	var/list/_managed_instance_cache = list()
+
+/datum/controller/subsystem/managed_instances/proc/get(var/instance_type, var/cache_id, var/cache_category = "default", var/list/instance_args)
+	if(!cache_id)
+		cache_id = instance_type
+	if(isnull(cache_id)) // No type and no id supplied, no point going further.
+		CRASH("Attempted to retrieve a managed instance with no ID.")
```
out-of-phaze:
```suggestion
		CRASH("Attempted to retrieve a managed instance with no type or ID.")
```
Feel like this should be noted, since from just the crash message alone it's unclear that "no ID" means "no type" as well.

MistakeNot4892: Redid the comment, slightly different wording.

afterthought2 (approved):

out-of-phaze (approved):

PR #3001, opened 2023-03-06
Basic stressor system.

MistakeNot4892:
Depends on #3000. Opening for general review, I don't necessarily think this is a good state to merge the PR in. Suggestions for some unobtrusive stressors/relaxors to add would be appreciated.

## Description of changes
Adds a list on /mob/living that can track datum instances representing stressors and relaxors.
Adds functions for modifying this list and retrieving the current stress of the mob.
Adds config options for multiplying passive health recovery, shock recovery/gain and blood regeneration based on current stress.
Adds basic stressors for cooked food, hunger, thirst and use of brute and burn medication.

Adds an indirect buff/debuff system that can be used in the future to reward or punish RP actions, add some value to cooked food, etc.

## Authorship
Myself.

## Changelog
Nothing player-facing currently as no stressors have been implemented.


out-of-phaze:
So based on the discussion in the Discord, I'd like to give some elaboration:
> Adds an indirect buff/debuff system that can be used in the future to reward or punish RP actions, add some value to cooked food, etc.

This is a bit misleading, especially for people who think it's something like TG's mood system; there's a common misconception that the TG mood system was supposed to do something along these lines. Here's the elaboration Loaf gave in the discord:

> it's not mood
it's an overall level of physical comfort and general health
the idea is to give you a very minor boost to recovering after injury if you do stuff that is currently solely RP, like going to the chef rather than going to a vendor
and to very slightly penalize you if you ignore that kind of thing to live off vendor trash and go straight back into combat after being injured
you've been flashbanged (stressor), you're hungry (stressor) and and you're coming down from combat stims (stressor), you're in a poor state in a way that the sim doesn't account for currently, you might be happy or whatever character moodwise but you're still going to recover slower from injury than the guy who has a full nutritious meal in them and is in a bed in medbay
it's very much not a mood system
individual stressors don't do anything specific, they just contribute to your overall stress value
which is then factored into recovery stuff
likewise the guy in medbay is going to recover faster than the same guy if he left surgery and went straight back out into the halls

Based on this, I think the idea is solid and isn't going to fall into the same traps as TG's mood system, although I kind of like the idea of writing a small README for the stressor system with some dos and don'ts to make sure it doesn't happen.

PsyCommando:
Just wondering, but did we ever fix paincrit and painshock? It would go a bit crazy sometimes, like I remember that if you'd dislocate your 4 limbs you could instantly die because pain threshold levels would just cause cardiac arrest at a certain pain level.

Because I noticed this messes with pain shock, but it might making hunting down the problem harder if we add more variables to the calculation?

MistakeNot4892:
You can disable it entirely in config. It hasn't changed the actual logic much.

code/modules/mob/living/carbon/human/life.dm
```diff
@@ -837,70 +837,87 @@
...
-		shock_stage = max(shock_stage - recovery, 0)
+		shock_stage = max(shock_stage - (recovery * (1-stress_modifier)), 0)
+		return
+
+	if(stat != CONSCIOUS)
+		return
+
+	// If we haven't adjusted by at least one full unit, don't run the message logic below.
+	var/next_rounded_shock_stage = round(shock_stage)
+	if(next_rounded_shock_stage == rounded_shock_stage)
```
MistakeNot4892, line 872: Noting to remind myself to check if this breaks anything.

MistakeNot4892: It has indeed broken something.

MistakeNot4892: Actually it's working fine and the pain messages are just low rolls.

code/controllers/subsystems/misc_late.dm
```diff
@@ -6,6 +6,7 @@ SUBSYSTEM_DEF(misc_late)
 	var/list/turbolifts_to_open = list()
 
 /datum/controller/subsystem/misc_late/Initialize()
+
```
out-of-phaze, line 9: very important functional whitespace

out-of-phaze (approved):

PR #3002, opened 2023-03-11
Added missing circuit designs for internet machines

PsyCommando:
## Description of changes
A few things were missing a recipe for their circuitboards. And it seems unit tests didn't pick those up.
Also fixed the internet repeater having no construct state.

## Changelog
:cl:
add: Added missing recipes for internet modem, uplink, uplink computer, and repeater.
/:cl:

### DIFF
--- a/code/modules/overmap/internet/internet_circuitboards.dm
+++ b/code/modules/overmap/internet/internet_circuitboards.dm
@@ -34,3 +34,12 @@
 		/obj/item/stock_parts/keyboard = 1,
 		/obj/item/stock_parts/power/terminal = 1
 	)
+
+/datum/fabricator_recipe/imprinter/circuit/internet_uplink
+	path = /obj/item/stock_parts/circuitboard/internet_uplink
+
+/datum/fabricator_recipe/imprinter/circuit/internet_uplink_computer
+	path = /obj/item/stock_parts/circuitboard/internet_uplink_computer
+
+/datum/fabricator_recipe/imprinter/circuit/internet_repeater
+	path = /obj/item/stock_parts/circuitboard/internet_repeater
--- a/code/modules/overmap/internet/internet_circuitboards.dm
+++ b/code/modules/overmap/internet/internet_circuitboards.dm
@@ -8,11 +8,7 @@
 		/obj/item/stock_parts/micro_laser = 2,
 		/obj/item/stock_parts/smes_coil = 1
 	)
-	additional_spawn_components = list(
-		/obj/item/stock_parts/console_screen = 1,
-		/obj/item/stock_parts/keyboard = 1,
-		/obj/item/stock_parts/power/terminal = 1
-	)
+	additional_spawn_components = null
 
 /obj/item/stock_parts/circuitboard/internet_uplink_computer
 	name = "circuitboard (PLEXUS uplink controller)"
@@ -29,11 +25,7 @@
 		/obj/item/stock_parts/capacitor = 2,
 		/obj/item/stock_parts/micro_laser = 1
 	)
-	additional_spawn_components = list(
-		/obj/item/stock_parts/console_screen = 1,
-		/obj/item/stock_parts/keyboard = 1,
-		/obj/item/stock_parts/power/terminal = 1
-	)
+	additional_spawn_components = null
 
 /datum/fabricator_recipe/imprinter/circuit/internet_uplink
 	path = /obj/item/stock_parts/circuitboard/internet_uplink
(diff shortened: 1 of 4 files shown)


code/modules/overmap/internet/internet_uplink.dm
```diff
@@ -161,6 +162,7 @@ var/global/list/internet_uplinks = list()
 	light_color = COLOR_GREEN
 	idle_power_usage = 250
 	active_power_usage = 500
+	base_type = /obj/machinery/computer/internet_uplink
```
afterthought2: Does this actually work? Won't the circuit not match the construct state if you do this? I'm kind of surprised it passes unit tests (most likely because it's not present on any map).

code/modules/overmap/internet/internet_repeater.dm
```diff
@@ -11,6 +11,7 @@ var/global/list/internet_repeaters = list()
 	use_power = POWER_USE_ACTIVE
 	idle_power_usage = 50
 	active_power_usage = 5000
+	construct_state = /decl/machine_construction/default/panel_closed
```
afterthought2, line 13: The current circuitboard for it gives it a (nonbuildable!) terminal on spawn, which seems clearly wrong (note that it also has the default nonbuildable apc teslalink component as well so you won't notice). Please double-check the components.

code/modules/overmap/internet/internet_circuitboards.dm
```diff
@@ -34,3 +34,12 @@
 		/obj/item/stock_parts/keyboard = 1,
 		/obj/item/stock_parts/power/terminal = 1
 	)
+
+/datum/fabricator_recipe/imprinter/circuit/internet_uplink
+	path = /obj/item/stock_parts/circuitboard/internet_uplink
+
+/datum/fabricator_recipe/imprinter/circuit/internet_uplink_computer
+	path = /obj/item/stock_parts/circuitboard/internet_uplink_computer
```
afterthought2, line 34: This is probably how the computer one should be built, but I think it _also_ fails to set the board_type to computer.

afterthought2 (changes requested):
Unbuildable component variants in `additional_spawn_components` are questionable: they mean that mapped objects will spawn with them, but they will not be dropped when deconstructed, so you have to source your own component when rebuilding (a _built_ object will not spawn them). Better to either place them in `uncreated_component_parts` (will spawn in built objects too) or place the `/buildable` variant in `additional_spawn_components`.

code/modules/overmap/internet/internet_uplink.dm
PsyCommando: I doubt it breaks anything, because that ran through unit tests downstream on a map we had it on.
Base type is just the buildable variant of this machine to fall back to when disassembling it.
It's essentially unneeded here though. I probably forgot to remove it after I got rid of the buildable type for this one.

PsyCommando: > Unbuildable component variants in `additional_spawn_components` are questionable: they mean that mapped objects will spawn with them, but they will not be dropped when deconstructed, so you have to source your own component when rebuilding (a _built_ object will not spawn them). Better to either place them in `uncreated_component_parts` (will spawn in built objects too) or place the `/buildable` variant in `additional_spawn_components`.

Honestly, it's a bit weird to get reviewed over code that isn't even in this PR and that was already reviewed?

But yeah I know. That's something that I've fixed downstream for ages. But apparently didn't make it here. I was mostly just pushing the missing recipes here.

PsyCommando:
> Unbuildable component variants in `additional_spawn_components` are questionable: they mean that mapped objects will spawn with them, but they will not be dropped when deconstructed, so you have to source your own component when rebuilding (a _built_ object will not spawn them). Better to either place them in `uncreated_component_parts` (will spawn in built objects too) or place the `/buildable` variant in `additional_spawn_components`.

Honestly, it's a bit weird to get reviewed over code that isn't even in this PR and that was already reviewed?

But yeah I know. That's something that I've fixed downstream for ages. But apparently didn't make it here. I was mostly just pushing the missing recipes here.

code/modules/overmap/internet/internet_uplink.dm
PsyCommando: It works but it's redundant, because that's not a buildable subtype. So I'll remove that.

code/modules/overmap/internet/internet_circuitboards.dm
PsyCommando: That's true, it doesn't have the right board type. I'll fix that.

PsyCommando: Nah, the default value of the board_type is ``"computer"``. It should be fine.

PsyCommando:
I made changes to the base boards and machines so they act a bit more sensibly.

afterthought2 (approved):
This isn't how I'd have done it, but I don't have issues with it.

PR #3004, opened 2023-03-11
force airlock sensor update during cycling

PsyCommando:
## Description of changes
The sensor would sometimes stop reporting pressure during cycling when having to purge the chamber. Which would cause it to be stuck in the purging phase. Fixed it by forcing updates when the master is cycling, even if the pressure is null.

## Changelog
:cl:
bugfix: Fix airlock controller sometimes getting stuck during purging phase.
/:cl:

### DIFF
--- a/code/game/machinery/doors/airlock_control.dm
+++ b/code/game/machinery/doors/airlock_control.dm
@@ -165,7 +165,9 @@
 		var/datum/gas_mixture/air_sample = return_air()
 		var/new_pressure = round(air_sample.return_pressure(),0.1)
 
-		if(abs(pressure - new_pressure) > 0.001 || pressure == null)
+		//Try preventing the sensor from spamming when there's no differences, and the controller isn't cycling.
+		//Can't rely only on difference, since it's possible for the pressure to stay nearly the same for a few calls, which prevents the machine var from updating
+		if(abs(pressure - new_pressure) > 0.001 || master_cycling)
 			var/decl/public_access/public_variable/airlock_pressure/pressure_var = GET_DECL(/decl/public_access/public_variable/airlock_pressure)
 			pressure_var.write_var(src, new_pressure)
 
--- a/code/game/machinery/doors/airlock_control.dm
+++ b/code/game/machinery/doors/airlock_control.dm
@@ -164,17 +164,13 @@
 	if(!(stat & (NOPOWER | BROKEN)))
 		var/datum/gas_mixture/air_sample = return_air()
 		var/new_pressure = round(air_sample.return_pressure(),0.1)
+		var/decl/public_access/public_variable/airlock_pressure/pressure_var = GET_DECL(/decl/public_access/public_variable/airlock_pressure)
+		pressure_var.write_var(src, new_pressure)
 
-		//Try preventing the sensor from spamming when there's no differences, and the controller isn't cycling.
-		//Can't rely only on difference, since it's possible for the pressure to stay nearly the same for a few calls, which prevents the machine var from updating
-		if(abs(pressure - new_pressure) > 0.001 || master_cycling)
-			var/decl/public_access/public_variable/airlock_pressure/pressure_var = GET_DECL(/decl/public_access/public_variable/airlock_pressure)
-			pressure_var.write_var(src, new_pressure)
-
-			var/new_alert = (pressure < ONE_ATMOSPHERE*0.8)
-			if(new_alert != alert)
-				alert = new_alert
-				update_icon()
+		var/new_alert = (pressure < ONE_ATMOSPHERE*0.8)
+		if(new_alert != alert)
+			alert = new_alert
+			update_icon()
 
 /**Meant to update the icon when the master airlock controller is cycling */
 /obj/machinery/airlock_sensor/proc/set_master_cycling(var/state)
--- a/code/game/machinery/embedded_controller/airlock_program.dm
+++ b/code/game/machinery/embedded_controller/airlock_program.dm
@@ -89,8 +89,7 @@
 	if(!receive_tag) return
 
 	if(receive_tag==tag_chamber_sensor)
-		if(signal.data["pressure"])
-			memory["chamber_sensor_pressure"] = text2num(signal.data["pressure"])
+		memory["chamber_sensor_pressure"] = text2num(signal.data["pressure"])
 
 	else if(receive_tag==tag_exterior_sensor)
 		memory["external_sensor_pressure"] = text2num(signal.data["pressure"])


code/game/machinery/embedded_controller/airlock_program.dm
```diff
@@ -89,8 +89,7 @@
 	if(!receive_tag) return
 
 	if(receive_tag==tag_chamber_sensor)
-		if(signal.data["pressure"])
-			memory["chamber_sensor_pressure"] = text2num(signal.data["pressure"])
+		memory["chamber_sensor_pressure"] = text2num(signal.data["pressure"])
```
afterthought2, line 92: This I think is incorrect: if it sends a signal without this tag at all, this will reset the local pressure tally to `0` for no reason. Check `("pressure" in signal_data)`.

code/game/machinery/doors/airlock_control.dm
```diff
@@ -165,7 +165,9 @@
 		var/datum/gas_mixture/air_sample = return_air()
 		var/new_pressure = round(air_sample.return_pressure(),0.1)
 
-		if(abs(pressure - new_pressure) > 0.001 || pressure == null)
+		//Try preventing the sensor from spamming when there's no differences, and the controller isn't cycling.
+		//Can't rely only on difference, since it's possible for the pressure to stay nearly the same for a few calls, which prevents the machine var from updating
+		if(abs(pressure - new_pressure) > 0.001 || master_cycling)
```
afterthought2: Are you sure that this is needed in light of the other change? `master_cycling` is an icon update var which is logically decoupled from pressure updating. Because the public vars/methods are in principle user-accessible, it's desirable to avoid these kinds of opaque internal dependencies.

I think adding `|| new_pressure == null` would be fine and am surprised it doesn't do that, but is probably inconsequential (will report next tick anyway).

Anything beyond that, I would prefer to modify the logic on the airlock controller instead.

PsyCommando: Yes. master_cycling was never meant to be only for icon updates. I added that to prevent all the airlock machines from going out of sync which they used to do constantly before for a variety of reasons. 

While the master is cycling, you want sensors to prioritize sending pressure updates, even if it's below the threshold. Because otherwise the master controller will never know it actually emptied the chamber, unless it gets really lucky and catches the last pressure update when the chamber is below the threshold. That's the bug this has been fixing downstream for a while. It's even worst when you have an airlock that has to swap atmosphere to an exterior atmosphere.

Also, ``pressure == null`` must be communicated as well, because otherwise the the controller will be sent only the last value before the pressure actually hit 0/null, which could be anything depending on how fast the pumps remove the air from the chamber. We have airlocks with a bunch of high capacity vents downstream and this is what was happening, it was pulling air out too fast for the sensors to update before pressure hit 0, and the controller would get stuck with the last chamber pressure before it hit 0, never completing it's purge.

code/game/machinery/embedded_controller/airlock_program.dm
PsyCommando: All the other sensors for the exterior and interior do not have this check for a reason. 
The airlock program needs to know the current pressure even if it's null. Otherwise if the last value it received was above 0 and was over the target pressure, it will never complete it's cycle.

PsyCommando:
monkeys crashing the unit tests again X_X

code/game/machinery/embedded_controller/airlock_program.dm
afterthought2: Yes...which the check I am suggesting here does not interfere with. `"pressure" = null` is sent, and `null` is recorded if `"pressure"` is one of the keys transmitted. If `"pressure"` is not sent at all, you should not update to `null`.

code/game/machinery/doors/airlock_control.dm
afterthought2: But it _does_ send when `pressure == null` already. I don't follow. Of course, when it sends it, it's not received, because of the bug in the other line you touch, but it is clearly sent.

code/game/machinery/embedded_controller/airlock_program.dm
PsyCommando: I disagree with that. 
If "pressure" isn't sent, that means you don't have a chamber sensor anymore, and it's required. So the pressure being changed to 0/null makes sense. Otherwise it'll stay stuck with the last reading, and that would also be wrong.

code/game/machinery/doors/airlock_control.dm
PsyCommando: 0 != null, so that condition already doesn't work. Null doesn't even gets cast to 0 in that check.
So it would not update when we hit 0, until we hit null, which isn't a given and only happens when there's no moles left of anything, iirc.

But it's true that master_cycling might not be the best replacement either. The condition here was mainly a way to try to lower the cost of having a bunch of airlock sensors sending tons of pressure updates constantly. But it doesn't really do a good job at it.

code/game/machinery/embedded_controller/airlock_program.dm
afterthought2: No, if "pressure" isn't sent someone modified the radio components, which is completely valid. If the sensor is destroyed, then probably being stuck is not unreasonable. In any case, it's up to the airlock to handle destruction of the sensor or whatever gracefully. It can e.g. periodically send an update request, or abort, or just proceed, etc.

code/game/machinery/doors/airlock_control.dm
afterthought2: If it's set to `0` instead of a null literal, then check `pressure` instead of `pressure == null`. You can even check `pressure <= [small number]` if you really want.

PsyCommando: I'll just remove that if entirely.

code/game/machinery/embedded_controller/airlock_program.dm
PsyCommando: I don't know where you're going with this, but you're really losing me here.

The sensor value being set to null is legal. As long as a message tagged with the sensor's tag is sending the message. And that even if it doesn't sends the "pressure" key. 
Try it yourself with a airlock setup with 3 sensors and 2 different atmospheres to change between. You'll see what I mean.

afterthought2: I don't know how to explain myself better. If a signal does not contain the item `list("pressure" = null)` or `list("pressure" = number)`, the airlock controller should not update its pressure tally. The way to check for this is `("pressure" in item)`. 

Currently the check being used is `item["pressure"]`, which ignores the signals `list("pressure" = null)` and `list("pressure" = 0)` and as you point out this leads to bugs. My proposed check will not ignore those signals and hence should resolve the bug you describe. 

You ALSO want, for reasons that escape me, to have a signal of e.g. `list("foo" = 5)` make the controller update the internal pressure tally to `null` (this is what omitting the check I am suggesting does). I dislike this: it is not needed to solve the issue you have described; it leads to intuitive behavior. Note that arguing that this is somehow consistent with how the controller works currently is simply invalid: currently, the controller makes a check _even stronger than_ the one I propose, so it certainly does not update internal pressure tally when `list("foo" = 5)` is sent.

If this is still unclear, it would probably be more useful to discuss on discord or involve another maint.

PsyCommando: Look at this before dismissing this as "simply invalid" pls:
```
	else if(receive_tag==tag_exterior_sensor)
		memory["external_sensor_pressure"] = text2num(signal.data["pressure"])
	else if(receive_tag==tag_interior_sensor)
		memory["internal_sensor_pressure"] = text2num(signal.data["pressure"])
	else if(receive_tag==tag_exterior_door)
		memory["exterior_status"]["state"] = signal.data["door_status"]
		memory["exterior_status"]["lock"] = signal.data["lock_status"]
	else if(receive_tag==tag_interior_door)
		memory["interior_status"]["state"] = signal.data["door_status"]
		memory["interior_status"]["lock"] = signal.data["lock_status"]
```
All other sensors do it exactly like I changed the chamber sensor to for a reason.
Even the doors do it like that. 

That's a legal state of things that's already being handled by the controller. If a matching connected device doesn't send the tags that are expected, you don't have a reading to show. Don't just keep whatever is in memory as there is currently no up to date pressure readings.
It might have made sense if sensors sent anything else, or could just skip sending pressure updates as part of their normal operation. But that's not the case, and would obfuscate there being something interfering with the sensors messages.

I have no ideas what to tell you other than that.

afterthought2: I'm fine with changing all these other places as well, but clearly it's outside the scope of this fix.

I regret not having refactored the airlock controller further (to use a bunch of radio components), which would have already forced it to function like I describe across the board.

PsyCommando:
Well since it's a 2 line fix and it's not going anywhere, I'll let someone else do it upstream. We've got this already downstream.

PR #3005, opened 2023-03-11
Make geothermal generator work

PsyCommando:
## Description of changes
Ported this from downstream. We needed geothermals for our map, and had to fix them. So I thought you guys might want them here too.

* Now uses a terminal in an adjacent turf, so it can actually be installed.
* Now provides a steadier power output.
* Now has everything a machine needs to not crash unit tests, and function.
* Now uses particles for geyser steam jets. Used steam spread before, but that effect caused massive lag and timer problems, while this one is much simpler and faster.

## Changelog
:cl:
add: Added geothermal generator circuit design to imprinter.
add: New, less laggy visual effect for steam geyser spout.
bugfix: Fixed geothermal generators.
/:cl:

### DIFF
--- a/code/modules/power/geothermal/_geothermal.dm
+++ b/code/modules/power/geothermal/_geothermal.dm
@@ -192,17 +192,17 @@ var/global/const/MAX_GEOTHERMAL_PRESSURE =               12000
 		var/consumed_pressure = current_pressure * GEOTHERMAL_PRESSURE_CONSUMED_PER_TICK
 		current_pressure -= consumed_pressure
 		var/remaining_pressure = consumed_pressure
-		consumed_pressure = round(consumed_pressure * efficiency)
+		consumed_pressure = round(consumed_pressure * efficiency, 0.01)
 		remaining_pressure -= consumed_pressure
 
-		last_generated = round(consumed_pressure * GEOTHERMAL_PRESSURE_TO_POWER)
+		last_generated = round(consumed_pressure * GEOTHERMAL_PRESSURE_TO_POWER, 0.01)
 		if(last_generated)
 			generate_power(last_generated)
 		remaining_pressure = round(remaining_pressure * GEOTHERMAL_PRESSURE_LOSS)
 		if(remaining_pressure)
 			addtimer(CALLBACK(src, .proc/propagate_pressure, remaining_pressure), 5)
 	update_icon()
-	if(current_pressure <= 10)
+	if(current_pressure <= 1)
 		return PROCESS_KILL
 
 /obj/machinery/geothermal/proc/propagate_pressure(var/remaining_pressure)
--- a/code/modules/power/geothermal/_geothermal.dm
+++ b/code/modules/power/geothermal/_geothermal.dm
@@ -130,9 +130,7 @@ var/global/const/MAX_GEOTHERMAL_PRESSURE =               12000
 	. = ..()
 	if(get_turf(loc))
 		refresh_neighbors()
-		for(var/turf/T as anything in RANGE_TURFS(loc, 1))
-			for(var/obj/machinery/geothermal/neighbor in T)
-				neighbor.refresh_neighbors()
+		propagate_refresh_neighbors()
 	STOP_PROCESSING_MACHINE(src, MACHINERY_PROCESS_SELF)
 	return INITIALIZE_HINT_LATELOAD
 
@@ -145,9 +143,14 @@ var/global/const/MAX_GEOTHERMAL_PRESSURE =               12000
 	unset_vent()
 	. = ..()
 	if(istype(last_loc))
-		for(var/turf/T as anything in RANGE_TURFS(last_loc, 1))
-			for(var/obj/machinery/geothermal/neighbor in T)
-				neighbor.refresh_neighbors()
+		propagate_refresh_neighbors(last_loc)
+
+///Tell all neighbors of the center atom to call update neighbors
+/obj/machinery/geothermal/proc/propagate_refresh_neighbors(var/atom/center = src)
+	for(var/adir in global.alldirs)
+		var/obj/machinery/geothermal/G = locate(/obj/machinery/geothermal) in get_step(center, adir)
+		if(G?.anchored)
+			G.refresh_neighbors()
 
 /obj/machinery/geothermal/examine(mob/user, distance)
 	. = ..()
--- a/code/modules/power/geothermal/_geothermal.dm
+++ b/code/modules/power/geothermal/_geothermal.dm
@@ -40,6 +40,7 @@ var/global/const/MAX_GEOTHERMAL_PRESSURE =               12000
 	anchored   = TRUE
 	layer      = TURF_LAYER + 0.01
 	level      = 1 //Goes under floor/plating
+	///The particle emitter that will generate the steam column effect for this geyser
 	var/particles/geyser_steam/steamfx
 
 /obj/effect/geyser/Initialize(ml)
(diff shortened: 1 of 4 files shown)


code/modules/power/geothermal/_geothermal.dm
```diff
@@ -1,63 +1,192 @@
...
 		var/matrix/M = matrix()
 		M.Scale(-1, 1)
 		transform = M
 	set_extension(src, /datum/extension/geothermal_vent)
+	steamfx = new //Prepare our FX
+
+/obj/effect/geyser/proc/do_spout()
+	particles = steamfx
+	particles.spawning = initial(particles.spawning)
+	spawn(6 SECONDS)
```
afterthought2: Why not waitfor FALSE?

code/modules/power/geothermal/_geothermal.dm
```diff
@@ -1,63 +1,192 @@
...
+	core_skill                     = SKILL_ENGINES
+	required_interaction_dexterity = DEXTERITY_SIMPLE_MACHINES
+	construct_state                = /decl/machine_construction/default/panel_closed
+	uncreated_component_parts      = list(
+		/obj/item/stock_parts/power/terminal = 1
+	)
+	stock_part_presets             = list(
+		/decl/stock_part_preset/terminal_setup/offset_dir,
+	)
+	base_type = /obj/machinery/geothermal/buildable
```
afterthought2: I think you will accomplish the same effect by setting `uncreated_component_parts = null`, adding the unbuildable terminal to `additional_spawn_components`, and getting rid of this subtype. That also has the advantage that you can use a _buildable_ terminal component, which is more player-friendly.

afterthought2 (commented):
I'll let Loaf comment on the particles if he wants to.

code/modules/power/geothermal/_geothermal.dm
PsyCommando: I actually don't remember why I went with this. 
All I remember is that timers would basically shit themselves and I got mad and used spawn instead.
I guess waitfor could work.

code/modules/power/geothermal/_geothermal.dm
PsyCommando: I think the buildable type was kind of a leftover I forgot about. I just tweaked the base type instead.

code/modules/power/geothermal/_geothermal.dm
```diff
@@ -1,63 +1,198 @@
...
 			T = T.ChangeTurf(/turf/exterior/mud)
 		if(prob(50 - (dist * 10)))
 			new /obj/random/seaweed(T)
 
-/obj/effect/geyser/explosion_act(severity)
-	. = ..()
-	if(!QDELETED(src) && prob(100 - (25 * severity)))
-		physically_destroyed()
+/obj/effect/geyser/underwater/do_spout()
+	//Do some bubbles or something
```
MistakeNot4892: Turfs have a `show_bubbles()` proc that does a bubble animation on the fluid layer.

code/modules/power/geothermal/_geothermal.dm
```diff
@@ -1,63 +1,198 @@
...
+	var/tmp/image/glow
+	var/tmp/list/neighbor_connectors
+	var/tmp/list/neighbor_connectors_glow
+
+//#TODO: Maybe should cache neighbors and put listeners on them?
+
+/obj/machinery/geothermal/Initialize(mapload, d, populate_parts = TRUE)
+	. = ..()
+	if(get_turf(loc))
+		refresh_neighbors()
```
MistakeNot4892, line 134: This is going to run twice since `src` isn't excluded from the RANGE_TURFS logic.

code/modules/power/geothermal/_geothermal.dm
```diff
@@ -1,63 +1,198 @@
...
+// Geyser Object
+//////////////////////////////////////////////////////////////////////
+
+///A prop that periodically emit steam spouts and can have a geothermal generator placed on top to generate power.
 /obj/effect/geyser
-	name = "geothermal vent"
-	desc = "A crack in the ocean floor that occasionally vents gouts of superheated water and steam."
-	icon = 'icons/effects/geyser.dmi'
+	name       = "geothermal vent"
+	desc       = "A vent leading to an underground geothermally heated reservoir, which periodically spew superheated liquid."
```
MistakeNot4892: `spews`

MistakeNot4892:
Would like a screenshot or recording of the particles if possible.

code/modules/power/geothermal/_geothermal.dm
PsyCommando: Oh good idea!

code/modules/power/geothermal/_geothermal.dm
PsyCommando: I'm not sure what you mean?
You mean refresh_neighbors is going to cause src to update itself?

code/modules/power/geothermal/_geothermal.dm
PsyCommando: kk

PsyCommando:
> Would like a screenshot or recording of the particles if possible.

Here's one:
https://user-images.githubusercontent.com/8772299/225117415-e5372f52-3a87-4a74-a559-c3e1bc5a9b0b.mp4

Not sure why, but github cut the beginning of the video...

code/modules/power/geothermal/_geothermal.dm
MistakeNot4892: refresh_neighbors() will be called twice on src in this proc.

PsyCommando: fixed

PsyCommando:
@MistakeNot4892 Did you see the video?

code/modules/power/geothermal/_geothermal.dm
```diff
@@ -1,68 +1,213 @@
...
+		var/datum/extension/geothermal_vent/GV = get_extension(O, /datum/extension/geothermal_vent)
+		if(!GV || QDELETED(GV))
+			continue
+		vent = GV
+		vent.set_my_generator(src)
+		return TRUE
+
+///Disconnect from any geothermal vents we may be connected to
+/obj/machinery/geothermal/proc/unset_vent()
+	if(!vent || QDELETED(vent))
```
MistakeNot4892: `QDELETED` also checks falsey, don't need the `!vent`.

code/modules/power/geothermal/_geothermal.dm
```diff
@@ -1,68 +1,213 @@
...
 
-/obj/effect/geyser/explosion_act(severity)
-	. = ..()
-	if(!QDELETED(src) && prob(100 - (25 * severity)))
-		physically_destroyed()
+/obj/effect/geyser/underwater/do_spout()
+	set waitfor = FALSE
+	var/turf/T = get_turf(src)
+	T.show_bubbles()
+	sleep(6 SECONDS)
```
MistakeNot4892: What is the purpose of this sleep? The proc is non-blocking.

MistakeNot4892:
> @MistakeNot4892 Did you see the video?

I did. Still not a fan of DM particles, but I'm not blocking over it.

code/modules/power/geothermal/_geothermal.dm
PsyCommando: Woops. I dunno why I put that there.

MistakeNot4892 (approved):

PR #3010, opened 2023-03-14
Fixes surgery runtimes from non-player mobs.

MistakeNot4892:
Fixes a couple of runtimes observed from monkeys picking up/using items in surgery contexts.


comma (approved):

out-of-phaze (approved):
I guess this might unintentionally make it so circuits that use items on people couldn't do surgery (unless you used a virtual mob?), but that's not necessarily unwanted, so...

Also, not a huge fan of the comment `Erroneous mob interaction`, can we try to be more descriptive in the future? I'm not sure what it contributes or means here.

PR #3012, opened 2023-03-14
Fix OpenDream compilation and runtime issues

out-of-phaze:
## Description of changes
- Reordered some variable definitions to be defined before their usage.
- Mirror Goon/etc. `INFINITY` macro for better JSON representation (1e69 or whatever can be represented, `1.#inf` can't be)
- Removes an invalid attempt at laterally-overriding a global proc (this isn't something BYOND allows)
- Fixes some invalid text macros (`\a` and `\the` without embedded expressions after)
- Fixes a subsystem priority sort that just flat-out didn't work.
- Tweaks the interface skin to mirror some changes made in https://github.com/tgstation/tgstation/commit/48ab72aa618bb86297147115473d5714e932cdef and after.
- Fixes a runtime due to firemodes not being initialized when a nuclear gun's icon is updated. This fails silently on 514 but runtimes on 515+.
- Temporarily disables some `process_proc` overrides that just point to subprocs of `/datum/proc/Process`. They didn't actually do anything but I'm unsure if they should be removed entirely.
- **CHANGES HOW JSON IS WRITTEN TO BE COMPLIANT WITH THE ACTUAL JSON SPEC.** This means **NO SINGLE QUOTES FOR KEYS.** Use `@'{"x":1,"y":2}'` instead of `"{'x':1,'y':1}"`!

Makes some absurd progress in running the codebase in OpenDream (with a few OD PRs merged). Also fixes a few issues BYOND is silent on, like the nuclear gun icon issue, the JSON thing, the invalid text macros, and the subsystem priority sort. Modernizes `interface/skin.dmf` by removing/moving unnecessary logic.

## Authorship
Me + that one TG commit as reference + Goon for the `INFINITY` macro. I think. Also thanks to the OD people for help.


MistakeNot4892:
Is it possible for us to validate JSON in CI? I guess we can't just regex for `"{` can we.

MistakeNot4892:
Per Discord, the two other PRs have cherrypicked out the relevant parts of this.

PR #3014, opened 2023-03-15
Fixed runtime with quantum mechanic spawn.

MistakeNot4892:


out-of-phaze (approved):
Huh. Think I encountered this when testing #3012 and just assumed it was an engine thing.

PR #3017, opened 2023-03-15
Turbolifts are now handled on SSmapping.

MistakeNot4892:
## Description of changes
Moves turbolift init onto `SSmapping` instead of handling it during `SSatoms` init..

Should avoid weirdness from things like decals overlapping with the elevator area getting deleted and then complaining about it when they try to initialize.

## Authorship
Myself.

## Changelog
Nothing player-facing.

### DIFF
--- a/code/modules/turbolift/turbolift_map.dm
+++ b/code/modules/turbolift/turbolift_map.dm
@@ -1,5 +1,5 @@
 // Map object.
-/obj/turbolift_map_holder
+/obj/abstract/turbolift_spawner
 	name = "turbolift map placeholder"
 	icon = 'icons/obj/turbolift_preview_3x3.dmi'
 	dir = SOUTH         // Direction of the holder determines the placement of the lift control panel and doors.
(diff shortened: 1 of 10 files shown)


code/modules/turbolift/turbolift_map.dm
```diff
@@ -1,5 +1,5 @@
 // Map object.
-/obj/turbolift_map_holder
+/obj/abstract/turbolift_spawner
```
PsyCommando, line 2: Might be worth making those ``/obj/abstract/landmark`` with ``delete_me = TRUE`` so they self-delete?

MistakeNot4892: They're not landmarks and never need to not-self delete. Inheritance for inheritance's sake wouldn't be worthwhile here.

out-of-phaze (approved):
A little worried about the lack of a qdeleted/null check in the SSmapping loop, but if it deletes or harddels prior to SSmapping init, we've got bigger issues.

PR #3027, opened 2023-03-27
Non-human mob inventory rewrite.

MistakeNot4892:
## Description of changes
Moves a bunch of logic down to /mob from /human and makes the various pseudoinventory implementations (nymphs, crows) use the appropriate procs and structures instead.

**TODO:**
- [ ] Convert hattable extension to use the head slot.
- [ ] Convert nymph held items and hat items to use slots.
- [ ] Convert corgis and maintenance drones (maybe).

## Why and what will this PR improve
Splitting it off #2533 to try and get some more forward momentum.

## Authorship
Myself.

## Changelog
Nothing player-facing.

PR #3029, opened 2023-03-28
Auditing attack_hand() and adding dexterity checks.

MistakeNot4892:
## Description of changes
This is going to be a bastard to review. I went through every `attack_hand()` override to add parent calls, truthy returns, and dexterity checks. This is split out from #3027 which will need to make `attack_hand()` callable by `/mob/living/UnarmedAttack()`.

Consistency and generally informative return values for `attack_hand()`. Also should allow for things like implementing unarmed damage against structures down the track.

## Authorship
Myself.

## Changelog
:cl:
tweak: Many interactions with empty hands now have dexterity checks associated.
/:cl:

### DIFF
--- a/code/game/gamemodes/cult/runes.dm
+++ b/code/game/gamemodes/cult/runes.dm
@@ -56,17 +56,19 @@
 		return
 
 /obj/effect/rune/attack_hand(var/mob/user)
+	SHOULD_CALL_PARENT(FALSE)
 	if(!iscultist(user))
 		to_chat(user, "You can't mouth the arcane scratchings without fumbling over them.")
-		return
+		return TRUE
 	if(user.is_silenced())
 		to_chat(user, "You are unable to speak the words of the rune.")
-		return
+		return TRUE
 	var/decl/special_role/cultist/cult = GET_DECL(/decl/special_role/cultist)
 	if(cult.powerless)
 		to_chat(user, "You read the words, but nothing happens.")
 		return fizzle(user)
 	cast(user)
+	return TRUE
 
 /obj/effect/rune/attack_ai(var/mob/user) // Cult borgs!
 	if(Adjacent(user))
@@ -295,11 +297,13 @@
 			to_chat(user, "<span class='danger'>It is about to dissipate.</span>")
 
 /obj/effect/cultwall/attack_hand(var/mob/user)
+	SHOULD_CALL_PARENT(FALSE)
 	if(iscultist(user))
 		user.visible_message("<span class='notice'>\The [user] touches \the [src], and it fades.</span>", "<span class='notice'>You touch \the [src], whispering the old ritual, making it disappear.</span>")
 		qdel(src)
 	else
 		to_chat(user, "<span class='notice'>You touch \the [src]. It feels wet and becomes harder the further you push your arm.</span>")
+	return TRUE
 
 /obj/effect/cultwall/attackby(var/obj/item/I, var/mob/user)
 	if(istype(I, /obj/item/nullrod))
@@ -806,26 +810,27 @@
 		qdel(src)
 
 /obj/effect/rune/tearreality/attack_hand(var/mob/user)
-	..()
-	if(summoning_entity && !iscultist(user))
-		var/input = input(user, "Are you SURE you want to sacrifice yourself?", "DO NOT DO THIS") in list("Yes", "No")
-		if(input != "Yes")
-			return
-		speak_incantation(user, "Uhrast ka'hfa heldsagen ver[pick("'","`")]lot!")
-		to_chat(user, "<span class='warning'>In the last moment of your humble life, you feel an immense pain as fabric of reality mends... with your blood.</span>")
-		for(var/mob/M in global.living_mob_list_)
-			if(iscultist(M))
-				var/decl/pronouns/G = user.get_pronouns()
-				to_chat(M, "You see a vision of \the [user] keeling over dead, his blood glowing blue as it escapes [G.his] body and dissipates into thin air; you hear an otherwordly scream and feel that a great disaster has just been averted.")
-			else
-				to_chat(M, "You see a vision of [name] keeling over dead, his blood glowing blue as it escapes his body and dissipates into thin air; you hear an otherwordly scream and feel very weak for a moment.")
-		log_and_message_admins("mended reality with the greatest sacrifice", user)
-		user.dust()
-		var/decl/special_role/cultist/cult = GET_DECL(/decl/special_role/cultist)
-		cult.powerless = 1
-		qdel(summoning_entity)
-		qdel(src)
-		return
+	SHOULD_CALL_PARENT(FALSE)
+	if(!summoning_entity || iscultist(user))
+		return FALSE
+	var/input = input(user, "Are you SURE you want to sacrifice yourself?", "DO NOT DO THIS") in list("Yes", "No")
+	if(input != "Yes")
+		return TRUE
+	speak_incantation(user, "Uhrast ka'hfa heldsagen ver[pick("'","`")]lot!")
+	to_chat(user, "<span class='warning'>In the last moment of your humble life, you feel an immense pain as fabric of reality mends... with your blood.</span>")
+	for(var/mob/M in global.living_mob_list_)
+		if(iscultist(M))
+			var/decl/pronouns/G = user.get_pronouns()
+			to_chat(M, "You see a vision of \the [user] keeling over dead, his blood glowing blue as it escapes [G.his] body and dissipates into thin air; you hear an otherwordly scream and feel that a great disaster has just been averted.")
+		else
+			to_chat(M, "You see a vision of [name] keeling over dead, his blood glowing blue as it escapes his body and dissipates into thin air; you hear an otherwordly scream and feel very weak for a moment.")
+	log_and_message_admins("mended reality with the greatest sacrifice", user)
+	user.dust()
+	var/decl/special_role/cultist/cult = GET_DECL(/decl/special_role/cultist)
+	cult.powerless = 1
+	qdel(summoning_entity)
+	qdel(src)
+	return TRUE
 
 /obj/effect/rune/tearreality/attackby()
 	if(the_end_comes)
--- a/code/game/gamemodes/cult/runes.dm
+++ b/code/game/gamemodes/cult/runes.dm
@@ -66,13 +66,15 @@
 	var/decl/special_role/cultist/cult = GET_DECL(/decl/special_role/cultist)
 	if(cult.powerless)
 		to_chat(user, "You read the words, but nothing happens.")
-		return fizzle(user)
+		fizzle(user)
+		return TRUE
 	cast(user)
 	return TRUE
 
 /obj/effect/rune/attack_ai(var/mob/user) // Cult borgs!
 	if(Adjacent(user))
-		attack_hand(user)
+		return attack_hand(user)
+	return FALSE
 
 /obj/effect/rune/proc/cast(var/mob/living/user)
 	fizzle(user)
--- a/code/game/gamemodes/cult/runes.dm
+++ b/code/game/gamemodes/cult/runes.dm
@@ -72,9 +72,7 @@
 	return TRUE
 
 /obj/effect/rune/attack_ai(var/mob/user) // Cult borgs!
-	if(Adjacent(user))
-		return attack_hand(user)
-	return FALSE
+	return attack_hand_with_interaction_checks(user)
 
 /obj/effect/rune/proc/cast(var/mob/living/user)
 	fizzle(user)
--- a/code/game/machinery/_machines_base/machine_construction/_construction.dm
+++ b/code/game/machinery/_machines_base/machine_construction/_construction.dm
@@ -81,10 +81,11 @@
 	return MCS_CONTINUE
 
 /decl/machine_construction/proc/attack_hand(mob/user, obj/machinery/machine)
-	if(!validate_state(machine))
-		PRINT_STACK_TRACE("Machine [log_info_line(machine)] violated the state assumptions of the construction state [type]!")
-		machine.attack_hand(user)
+	if(validate_state(machine))
 		return TRUE
+	PRINT_STACK_TRACE("Machine [log_info_line(machine)] violated the state assumptions of the construction state [type]!")
+	machine.attack_hand(user)
+	return TRUE
 
 /decl/machine_construction/proc/attackby(obj/item/I, mob/user, obj/machinery/machine)
 	if(!validate_state(machine))
--- a/code/game/machinery/_machines_base/machine_construction/_construction.dm
+++ b/code/game/machinery/_machines_base/machine_construction/_construction.dm
@@ -81,11 +81,9 @@
 	return MCS_CONTINUE
 
 /decl/machine_construction/proc/attack_hand(mob/user, obj/machinery/machine)
-	if(validate_state(machine))
-		return TRUE
-	PRINT_STACK_TRACE("Machine [log_info_line(machine)] violated the state assumptions of the construction state [type]!")
-	machine.attack_hand(user)
-	return TRUE
+	if(!validate_state(machine))
+		PRINT_STACK_TRACE("Machine [log_info_line(machine)] violated the state assumptions of the construction state [type]!")
+		machine.attack_hand(user)
 
 /decl/machine_construction/proc/attackby(obj/item/I, mob/user, obj/machinery/machine)
 	if(!validate_state(machine))
--- a/code/game/machinery/atmoalter/scrubber.dm
+++ b/code/game/machinery/atmoalter/scrubber.dm
@@ -172,18 +172,14 @@
 	name = "[name] (ID [id])"
 
 /obj/machinery/portable_atmospherics/powered/scrubber/huge/attack_hand(mob/user)
-	if((. = ..()))
-		return
-	to_chat(user, "<span class='notice'>You can't directly interact with this machine. Use the scrubber control console.</span>")
-	return TRUE
+	. = ..() // Buckling, climbers, grab interactions, component attack_hand; unlikely to return true.
+	if(!.)
+		to_chat(user, SPAN_WARNING("You can't directly interact with this machine. Use the scrubber control console."))
+		return TRUE
 
 /obj/machinery/portable_atmospherics/powered/scrubber/huge/on_update_icon()
-	overlays.Cut()
-
-	if((use_power == POWER_USE_ACTIVE) && !(stat & (NOPOWER|BROKEN)))
-		icon_state = "scrubber:1"
-	else
-		icon_state = "scrubber:0"
+	cut_overlays()
+	icon_state = "scrubber:[!!((use_power == POWER_USE_ACTIVE) && !(stat & (NOPOWER|BROKEN)))]"
 
 /obj/machinery/portable_atmospherics/powered/scrubber/huge/attackby(var/obj/item/I, var/mob/user)
 	if(IS_WRENCH(I))
--- a/code/game/machinery/cryopod.dm
+++ b/code/game/machinery/cryopod.dm
@@ -559,11 +559,14 @@
 	var/remains_type = /obj/item/remains/human
 
 /obj/structure/broken_cryo/attack_hand(mob/user)
-	..()
-	if (closed)
-		to_chat(user, SPAN_NOTICE("You tug at the glass but can't open it with your hands alone."))
+	. = ..()
+	if(.)
+		return
+	if(closed)
+		to_chat(user, SPAN_NOTICE("You tug at the glass, but can't open it further without a crowbar."))
 	else
 		to_chat(user, SPAN_NOTICE("The glass is already open."))
+	return TRUE
 
 /obj/structure/broken_cryo/attackby(obj/item/W, mob/user)
 	if (busy)
--- a/code/game/objects/effects/decals/Cleanable/misc.dm
+++ b/code/game/objects/effects/decals/Cleanable/misc.dm
@@ -13,11 +13,13 @@
 	icon_state = "ash"
 
 /obj/effect/decal/cleanable/ash/attack_hand(var/mob/user)
+	SHOULD_CALL_PARENT(FALSE)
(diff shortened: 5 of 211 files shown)


code/game/gamemodes/cult/runes.dm
```diff
@@ -56,17 +56,19 @@
...
-		return
+		return TRUE
 	if(user.is_silenced())
 		to_chat(user, "You are unable to speak the words of the rune.")
-		return
+		return TRUE
 	var/decl/special_role/cultist/cult = GET_DECL(/decl/special_role/cultist)
 	if(cult.powerless)
 		to_chat(user, "You read the words, but nothing happens.")
 		return fizzle(user)
```
out-of-phaze: `fizzle()` has no return value and simply displays a message.
```suggestion
		fizzle(user)
		return TRUE
```

code/game/machinery/cryopod.dm
```diff
@@ -561,7 +561,7 @@
 /obj/structure/broken_cryo/attack_hand(mob/user)
 	..()
 	if (closed)
-		to_chat(user, SPAN_NOTICE("You tug at the glass but can't open it with your hands alone."))
+		to_chat(user, SPAN_NOTICE("You tug at the glass, but can't open it further without a crowbar."))
```
out-of-phaze, line 566: I think the previous message was better. With the tools system, things other than crowbars can work. Maybe 'without more leverage'? 'Without a tool for leverage'?

code/game/objects/effects/decals/Cleanable/misc.dm
```diff
@@ -13,11 +13,13 @@
 	icon_state = "ash"
 
 /obj/effect/decal/cleanable/ash/attack_hand(var/mob/user)
+	SHOULD_CALL_PARENT(FALSE)
 	to_chat(user, "<span class='notice'>[src] sifts through your fingers.</span>")
 	var/turf/simulated/floor/F = get_turf(src)
 	if (istype(F))
 		F.dirt += 4
```
out-of-phaze, line 20: This needs to call `update_dirt()` to actually update the overlay, but `update_dirt()` also adds 0.5 dirt unconditionally, ugh. `/datum/reagents/proc/touch_turf` just modifies the alpha of the dirt overlay directly instead. I guess either of those are valid options?

code/game/objects/items/toys.dm
```diff
@@ -551,16 +551,19 @@
...
+		playsound(src.loc, 'sound/effects/rustle1.ogg', 100, 1)
+		if(user.a_intent == I_HELP)
+			user.visible_message("<span class='notice'><b>\The [user]</b> hugs [src]!</span>","<span class='notice'>You hug [src]!</span>")
+		else if (user.a_intent == I_HURT)
+			user.visible_message("<span class='warning'><b>\The [user]</b> punches [src]!</span>","<span class='warning'>You punch [src]!</span>")
+		else if (user.a_intent == I_GRAB)
+			user.visible_message("<span class='warning'><b>\The [user]</b> attempts to strangle [src]!</span>","<span class='warning'>You attempt to strangle [src]!</span>")
+		else
+			user.visible_message("<span class='notice'><b>\The [user]</b> pokes the [src].</span>","<span class='notice'>You poke the [src].</span>")
+			visible_message("[src] says, \"[phrase]\"")
```
out-of-phaze:
```suggestion
			audible_message("<span class='game say'><span class='name'>\The [src]</span> says, <span class='message'><span class='body'>\"[phrase]\"</span></span>")
```
Since you're already touching it I figure I'd at least suggest it. If you don't feel up to it just mark it as resolved and I'll ignore it in re-review.
Originally I suggested <b></b> on just `\The [src]` but that's technically not equivalent.

code/game/objects/items/weapons/policetape.dm
```diff
@@ -306,7 +306,7 @@ var/global/list/image/hazard_overlays //Cached hazard floor overlays for the bar
 		add_overlay(ovr)
 
 /obj/structure/tape_barricade/attack_hand(mob/user)
-	if (user.a_intent == I_HELP && allowed(user))
+	if (user.a_intent == I_HELP && allowed(user) && user.check_dexterity(DEXTERITY_SIMPLE_MACHINES, TRUE))
 		if(TAPE_BARRICADE_IS_CORNER_NEIGHBORS(neighbors) || (TAPE_BARRICADE_GET_NB_NEIGHBORS(neighbors) > 2))
 			to_chat(user, SPAN_WARNING("You can't lift up the pole. Try lifting the line itself."))
 			return
```
out-of-phaze:
```suggestion
			return TRUE
```

code/game/objects/items/weapons/tech_disks.dm
```diff
@@ -111,7 +111,7 @@
 	var/datum/fabricator_recipe/blueprint
 
 /obj/item/disk/design_disk/attack_hand(mob/user)
-	if(user.a_intent == I_HURT && blueprint)
+	if(user.a_intent == I_HURT && blueprint && user.has_dexterity(DEXTERITY_KEYBOARDS))
```
out-of-phaze:
```suggestion
	if(user.a_intent == I_HURT && blueprint && user.has_dexterity(DEXTERITY_SIMPLE_MACHINES))
```
It's just flipping a switch; see the `to_chat` line.

code/game/machinery/atmoalter/scrubber.dm
```diff
@@ -172,9 +172,8 @@
 	name = "[name] (ID [id])"
 
 /obj/machinery/portable_atmospherics/powered/scrubber/huge/attack_hand(mob/user)
-	if((. = ..()))
-		return
-	to_chat(user, "<span class='notice'>You can't directly interact with this machine. Use the scrubber control console.</span>")
+	SHOULD_CALL_PARENT(FALSE)
```
out-of-phaze: `..()` here is for buckling, climbers, component attack_hand, and physical attack_hand, and should definitely be called. To stop the UI from opening maybe `interface_interact` or `CanUseTopic` should be overridden instead.

code/game/objects/effects/decals/contraband.dm
```diff
@@ -121,22 +121,20 @@
...
-		add_fingerprint(user)
+	if(ruined || !user.check_dexterity(DEXTERITY_GRIP, TRUE))
+		return ..()
+	if(!alert("Do I want to rip the poster from the wall?","You think...","Yes","No") == "Yes")
+		return TRUE
+	if(ruined || !CanPhysicallyInteract(user) || !user.check_dexterity(DEXTERITY_GRIP))
+		return TRUE
+	visible_message("<span class='warning'>\The [user] rips \the [src] in a single, decisive motion!</span>" )
+	playsound(src.loc, 'sound/items/poster_ripped.ogg', 100, 1)
+	ruined = 1
```
out-of-phaze:
```suggestion
	ruined = TRUE
```
Feel free to resolve if you're not up to it, but since you're in the area already, etc.

code/game/objects/items/weapons/grenades/grenade.dm
```diff
@@ -104,4 +104,4 @@
 
 /obj/item/grenade/attack_hand()
 	walk(src, null, null)
```
out-of-phaze:
```suggestion
```
I strongly doubt that this is necessary anymore. Throwing doesn't use walk, right?

code/game/objects/structures/fountain.dm
```diff
@@ -18,26 +18,30 @@
 	set_light(5, 0.5, light_color)
 
 /obj/structure/fountain/attack_hand(var/mob/user)
-	if(user.incapacitated())
-		return
-	if(!CanPhysicallyInteract(user))
```
out-of-phaze, line 23: You don't seem to have readded these two checks anywhere. Is that intentional?

code/game/objects/structures/ironing_board.dm
```diff
@@ -118,21 +118,23 @@
 	..()
 
 /obj/structure/bed/roller/ironingboard/attack_hand(var/mob/user)
+	if(!user.check_dexterity(DEXTERITY_SIMPLE_MACHINES, TRUE))
+		return ..()	//Takes care of unbuckling.
```
out-of-phaze, line 122: Now how do you unbuckle someone from an ironing board if you have full dexterity? Before, `if(density && !holding && !cloth && buckled_mob)` would lead to a parent call; now it doesn't.

code/game/objects/structures/target_stake.dm
```diff
@@ -14,12 +14,13 @@
...
 /obj/structure/target_stake/attack_hand(var/mob/user)
-	. = ..()
-	if (pinned_target && ishuman(user))
-		var/obj/item/target/T = pinned_target
-		to_chat(user, "<span class='notice'>You take [T] out of the stake.</span>")
-		set_target(null)
-		user.put_in_hands(T)
+	if (!pinned_target || !user.check_dexterity(DEXTERITY_GRIP, TRUE))
+		return ..()
+	var/obj/item/target/T = pinned_target
```
out-of-phaze: The variable cast here is entirely pointless.

code/modules/assembly/holder.dm
```diff
@@ -119,28 +119,19 @@
 			var/obj/item/S = special_assembly
 			S.on_found(finder)
 
-
 /obj/item/assembly_holder/Move()
-	..()
```
out-of-phaze, line 124: Are you sure it's correct for the holder to move *after* `holder_movement()` rather than before?

code/modules/mechs/mech_wreckage.dm
```diff
@@ -49,14 +49,15 @@
...
-		var/obj/item/thing = pick(contents)
-		if(istype(thing))
-			thing.forceMove(get_turf(user))
-			user.put_in_hands(thing)
-			to_chat(user, "You retrieve \the [thing] from \the [src].")
-			return
-	return ..()
+	if(!length(contents) || !user.check_dexterity(DEXTERITY_GRIP, TRUE))
+		return ..()
+	var/obj/item/thing = pick(contents)
```
out-of-phaze:
```suggestion
	var/list/contained_atoms = get_contained_external_atoms()
	if(!length(contained_atoms) || !user.check_dexterity(DEXTERITY_GRIP, TRUE))
		return ..()
	var/obj/item/thing = pick(contained_atoms)
```

code/modules/paperwork/paperbin.dm
```diff
@@ -28,20 +28,13 @@
 	. = ..()
 
 /obj/item/paper_bin/attack_hand(mob/user)
-	if(!Adjacent(user))
```
out-of-phaze, line 31: This check is necessary since, iirc, dragging and dropping it onto yourself from any distance will call `attack_hand()`.

code/modules/projectiles/ammunition/boxes.dm
```diff
@@ -62,14 +62,17 @@
 		overlays += I
 
 /obj/item/ammo_magazine/shotholder/attack_hand(mob/user)
-	if((user.a_intent == I_HURT) && (stored_ammo.len))
```
out-of-phaze, line 65: Now won't trying to pick it up off the floor withdraw a casing instead? Please readd the intent check or add some other way to pick it up off the floor.

code/modules/random_map/drop/droppod_doors.dm
```diff
@@ -13,24 +13,34 @@
...
+
+/obj/structure/droppod_door/proc/deploy(var/mob/user, var/delay)
+	set waitfor = FALSE
 
-/obj/structure/droppod_door/proc/deploy()
 	if(deployed)
 		return
 
+	if(user)
+		to_chat(user, SPAN_DANGER("You prime the explosive bolts. Better get clear!"))
```
out-of-phaze, line 34: The bolts can be primed multiple times because `deployed` isn't set until after the delay. `deploying` should probably be set here, since it's already checked and isn't set anywhere else.

code/modules/vehicles/train.dm
```diff
@@ -118,17 +118,14 @@
 		return TRUE
 
 /obj/vehicle/train/attack_hand(mob/user)
-	if(user.stat || user.restrained() || !Adjacent(user))
```
out-of-phaze, line 121: This can be called from across Z-levels on some objects. Also, `UnarmedAttack()` (which calls `attack_hand()`) doesn't check adjacency/proximity before calling this. The adjacency check is required.

maps/away/magshield/magshield.dm
```diff
@@ -70,41 +70,47 @@
 		var/turf/T = get_turf(src)
 		var/area/A = get_area(src)
 		log_game("EMP with size ([heavy_range], [lighter_range]) in area [A.proper_name] ([T.x], [T.y], [T.z])")
-		visible_message("<span class='notice'>\the [src] suddenly activates.</span>", "<span class='notice'>Few lightnings jump between [src]'s rotating hands. You feel everything metal being pulled towards \the [src].</span>")
+		visible_message(
+			SPAN_DANGER("\the [src] suddenly activates!"),
```
out-of-phaze:
```suggestion
			SPAN_DANGER("\The [src] suddenly activates!"),
```

out-of-phaze (changes requested):
A couple concerns. Some of these were written a while ago so I might have missed some things or requested changes I no longer want to see done in this PR, so I'm probably going to go through and resolve a couple of the requested changes.

That said, I only realized about halfway through that you'd removed a number of `Adjacent()` checks, and that `attack_hand` can be called without being adjacent in the first place—which is intentional. Please double-check any changes that remove `Adjacent()` checks that I didn't comment on.

code/game/machinery/cryopod.dm
out-of-phaze: This is an issue that exists elsewhere, I'll probably address it at a later date.

code/game/objects/effects/decals/Cleanable/misc.dm
out-of-phaze: Dirt needs a complete rework. Making a personal note about this later.

code/game/objects/items/weapons/tech_disks.dm
out-of-phaze: Ehhhh.

code/game/machinery/atmoalter/scrubber.dm
out-of-phaze: Resolved already.

code/modules/vehicles/train.dm
MistakeNot4892: UnarmedAttack is called after checking adjacency, line 146 in `click.dm`. The previous call is for items in storage on the mob.

code/game/objects/items/weapons/tech_disks.dm
MistakeNot4892: The minimal dex check is to avoid mice being able to do it.

code/game/objects/items/weapons/grenades/grenade.dm
MistakeNot4892: Nope, is that what this is from? I left it because I wasn't confident I understood why it was there.

code/game/objects/structures/fountain.dm
MistakeNot4892: They're presumed by attack_hand being called, generally.

code/game/objects/structures/target_stake.dm
MistakeNot4892: Removed.

code/modules/assembly/holder.dm
MistakeNot4892: Reordered.

code/game/gamemodes/cult/runes.dm
MistakeNot4892: Applied.

code/game/objects/items/toys.dm
MistakeNot4892: Applied.

code/game/objects/items/weapons/policetape.dm
MistakeNot4892: Applied.

code/game/objects/effects/decals/contraband.dm
MistakeNot4892: Applied.

code/game/objects/items/weapons/grenades/grenade.dm
MistakeNot4892: Killed it regardless.

code/game/objects/structures/ironing_board.dm
MistakeNot4892: Checks `buckled_mob` now.

code/modules/mechs/mech_wreckage.dm
MistakeNot4892: Applied.

code/modules/paperwork/paperbin.dm
MistakeNot4892: Went with `CanPhysicallyInteract`.

code/modules/projectiles/ammunition/boxes.dm
MistakeNot4892: Readded the intent check.

code/modules/random_map/drop/droppod_doors.dm
MistakeNot4892: Applied.

maps/away/magshield/magshield.dm
MistakeNot4892: Applied.

MistakeNot4892:
I've added `attack_hand_with_interaction_checks` and used it in 40-odd places that I'm not sure about,

code/modules/economy/cael/Accounts_DB.dm
```diff
@@ -30,17 +30,12 @@
...
-			return
-		held_card = O
-
-		SSnano.update_uis(src)
-
-	attack_hand(user)
+	if(istype(O, /obj/item/card/id) && !held_card)
+		if(user.try_unequip(O, src))
+			held_card = O
+			ui_interact(user)
```
out-of-phaze:
```suggestion
			SSnano.update_uis(src)
```
This will update it for other users as well.

code/modules/clothing/masks/miscellaneous.dm
```diff
@@ -22,9 +22,10 @@
 
 // Clumsy folks can't take the mask off themselves.
 /obj/item/clothing/mask/muzzle/attack_hand(mob/user)
-	if(user.get_equipped_item(slot_wear_mask_str) == src && !user.check_dexterity(DEXTERITY_GRIP))
-		return 0
-	..()
+	if(user.get_equipped_item(slot_wear_mask_str) != src || user.check_dexterity(DEXTERITY_GRIP))
+		return ..()
+	to_chat(user, SPAN_WARNING("You cannot remove \the [src]."))
```
out-of-phaze:
```suggestion
	to_chat(user, SPAN_WARNING("You cannot remove \the [src] without help."))
```

out-of-phaze (changes requested):

out-of-phaze (approved):
Going to test this locally to catch any bugs, but I think three reviews is enough to say we've gotten most of the obvious issues.

code/game/machinery/_machines_base/machine_construction/_construction.dm
```diff
@@ -81,10 +81,11 @@
 	return MCS_CONTINUE
 
 /decl/machine_construction/proc/attack_hand(mob/user, obj/machinery/machine)
-	if(!validate_state(machine))
-		PRINT_STACK_TRACE("Machine [log_info_line(machine)] violated the state assumptions of the construction state [type]!")
-		machine.attack_hand(user)
+	if(validate_state(machine))
 		return TRUE
```
out-of-phaze:
```suggestion
		return FALSE
```
Otherwise, it stops you from ever interacting with machinery...

out-of-phaze (changes requested):

code/game/machinery/_machines_base/machine_construction/_construction.dm
```diff
@@ -81,11 +81,9 @@
...
 
 /decl/machine_construction/proc/attack_hand(mob/user, obj/machinery/machine)
-	if(validate_state(machine))
-		return TRUE
-	PRINT_STACK_TRACE("Machine [log_info_line(machine)] violated the state assumptions of the construction state [type]!")
-	machine.attack_hand(user)
-	return TRUE
+	if(!validate_state(machine))
+		PRINT_STACK_TRACE("Machine [log_info_line(machine)] violated the state assumptions of the construction state [type]!")
+		machine.attack_hand(user)
```
out-of-phaze, line 86: This is called from component_attack_hand *inside* machine.attack_hand; I don't think this is safe or *was ever safe,* we just didn't have any machines that violated their states.

out-of-phaze (changes requested):
I guess you technically don't have to change this because you didn't break it, but I don't feel good about leaving an issue like this in until another PR fixes it.

out-of-phaze (dismissed):
Will PR a fix for the invalid construct_state.attack_hand stuff later, since it wasn't caused by this PR.

code/modules/multiz/hoist.dm
```diff
@@ -38,18 +38,20 @@
 	var/obj/structure/hoist/source_hoist
 
 /obj/effect/hoist_hook/attack_hand(mob/user)
+	SHOULD_CALL_PARENT(FALSE)
```
out-of-phaze:
```
/obj/effect/hoist_hook/attack_hand(mob/user)
    if (user.incapacitated() || !user.check_dexterity(DEXTERITY_GRIP))
        return ..()

    if(source_hoist && source_hoist.hoistee)
        source_hoist.check_consistency()
        source_hoist.hoistee.forceMove(get_turf(src))
        user.visible_message(SPAN_NOTICE("[user] detaches \the [source_hoist.hoistee] from the hoist clamp."), SPAN_NOTICE("You detach \the [source_hoist.hoistee] from the hoist clamp."), SPAN_NOTICE("You hear something unclamp."))
        source_hoist.release_hoistee()
        return TRUE
```
Per Discord, this is probably how it should work.

out-of-phaze (dismissed):

out-of-phaze (approved):

PR #3030, opened 2023-03-28
pickup() -> on_picked_up()

MistakeNot4892:
- Renames `pickup()` to `on_picked_up()`
- Moved `on_picked_up()` call in item pickup to after `put_in_active_hand()` so it doesn't apply even if you fail to pick the item up.
- Tweaks a couple of `on_picked_up()` overrides.

### DIFF
--- a/code/modules/reagents/reagent_containers/drinks/bottle.dm
+++ b/code/modules/reagents/reagent_containers/drinks/bottle.dm
@@ -215,7 +215,7 @@
 			stop_spin_bottle = FALSE
 
 /obj/item/chems/drinks/bottle/on_picked_up(mob/living/user)
-	animate(src, transform = null, time = 0) //Restore bottle to its original position
+	animate(src, transform = null, time = 0) //Restore bottle to its original position - animate() is needed to interrupt SpinAnimation()
 
 //Keeping this here for now, I'll ask if I should keep it here.
 /obj/item/broken_bottle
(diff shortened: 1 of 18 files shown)


code/modules/reagents/reagent_containers/drinks/bottle.dm
```diff
@@ -214,8 +214,8 @@
 			sleep(sleep_not_stacking) //Not stacking
 			stop_spin_bottle = FALSE
 
-/obj/item/chems/drinks/bottle/pickup(mob/living/user)
-	animate(src, transform = null, time = 0) //Restore bottle to its original position
+/obj/item/chems/drinks/bottle/on_picked_up(mob/living/user)
+	transform = null //Restore bottle to its original position
```
out-of-phaze: I'm pretty sure this was necessary to interrupt `SpinAnimation`. Without using `animate` I think the bottle spin animation will continue in your inventory even after you pick it up. I've yet to test it locally though, so I could be wrong.
```suggestion
	animate(src, transform = null, time = 0) //Restore bottle to its original position and interrupt spin animation
```

out-of-phaze (changes requested):
If someone tests and finds out I'm wrong and the `animate` call is unnecessary, poke me and I'll approve this PR.

out-of-phaze:
image
Confirmed. Hilarious, but please fix it.

out-of-phaze (approved):
Looks good! Big fan of these renames to make stuff less confusing to new contributors who aren't familiar with some of the idiosyncratic names (like `u_equip`).

PR #3031, opened 2023-03-29
Hands slot refactor, supporting framework for inv slots.

MistakeNot4892:
## Description of changes
Splitting some more logic out of #2533. Finally getting the inventory slot datums implemented. This time it's the hands slot refactor and some associated inventory code rewrites that condense it down and move the old style hard references into their own procs for later removal.

This also moves stripping down to `/mob` level.

I've tested this locally and everything appears to be working as expected.

Working towards datum inventories.

## Authorship
Myself.

## Changelog
Nothing player-facing.


MistakeNot4892:
Stripping also tested, seems to be working.

code/modules/mob/mob_grabs.dm
```diff
@@ -8,6 +8,8 @@
 /mob/proc/apply_effect(var/effect = 0,var/effecttype = STUN, var/blocked = 0)
 	return
 // See /mob/living/carbon for this one.
```
out-of-phaze: This comment should be on `/mob/proc/get_organ`, since `has_organ` isn't stubbed.

PsyCommando:
Is this going to affect/alter things that occupy more than one hand slot?
Because, it's been pretty hard to make items that block both slots cleanly. There's the two handed efficiency check currently, but it only works for weapons, and not for say a big crate you have to carry, and want to prevent carrying two of them, one in each hand slots.

MistakeNot4892:
`Is this going to affect/alter things that occupy more than one hand slot?`
This does not change the user experience or behavior at all. We also don't currently have anything that occupies multiple hand slots - it would probably make it somewhat easier to implement structurally but doesn't support it out of the box.

out-of-phaze (changes requested):
Local testing revealed that if a held item gets qdeleted, like inflatable doors for example, that hand is permanently unusable (until you call clear_slot() with VV). Pls fix.

out-of-phaze (dismissed):
With the qdeleted item fix in, it works fine in local testing.

code/modules/mob/mob.dm
```diff
@@ -320,7 +320,70 @@
...
+			var/obj/item/tank/tank = get_equipped_item(slot)
+			if(istype(tank))
+				dat += "<BR><A href='?src=\ref[src];item=internals'>Toggle internals.</A>"
+				break
+
+	// Other incidentals.
+	var/obj/item/clothing/under/suit = get_equipped_item(slot_w_uniform_str)
+	if(istype(suit))
+		dat += "<BR><b>Pockets:</b> <A href='?src=\ref[src];item=pockets'>Empty or Place Item</A>"
+		if(suit.has_sensor == 1)
```
out-of-phaze:
```suggestion
		if(suit.has_sensor == SUIT_HAS_SENSORS)
```
Regression in merge resolution.

out-of-phaze (approved):

PR #3039, opened 2023-03-31
Added very basic sky turfs and parachutes from Polaris.

MistakeNot4892:
## Description of changes
Ports sky icons as a /turf/exterior/open subtype and parachutes as a way to negate fall damage. This is more or less a first draft intended for further refinement, I just need the types defined for some map work.

Needed for supporting some maps, and in the future I think could make exoplants quite a lot more interesting.

## Authorship
Icons and strings. are from Polaris.

## Changelog
Nothing player-facing currently.


code/game/objects/items/weapons/storage/parachute.dm
```diff
@@ -0,0 +1,44 @@
+//Parachutes
+/obj/item/storage/backpack/parachute
+	name = "parachute"
+	desc = "A specially-made backpack, designed to help one survive jumping from incredible heights. It sacrifices some storage space for that added functionality."
+	icon = 'icons/obj/items/storage/backpack/parachute.dmi'
+	var/packed = TRUE
+
+/obj/item/storage/backpack/parachute/Initialize(ml, material_key)
+	max_storage_space = round(max_storage_space * 0.5)
```
out-of-phaze, line 9: I disagree with this as a pattern. It should just have a lower initial storage space.

code/modules/multiz/movement.dm
```diff
@@ -186,11 +186,23 @@
...
+/atom/movable/proc/protected_from_fall_damage(var/turf/landing)
+	return !!(locate(/obj/structure/stairs) in src)
+
+/mob/protected_from_fall_damage(var/turf/landing)
+	. = ..()
+	if(!.)
+		// This is very silly, but it can be refined and made more appropriate as our multiz turf system is expanded.
+		var/obj/item/storage/backpack/parachute/parachute = get_equipped_item(slot_back_str)
+		if(istype(parachute) && parachute.packed)
+			parachute.packed = FALSE
```
out-of-phaze, line 198: Since this will by definition never fire twice in a landing, I think it's probably safe to have a `visible_message` here to signal that the parachute has deployed.

out-of-phaze (approved):
Looks good. I'd like it if it had a visible message on parachute deploy, but that's not blocking.

PR #3041, opened 2023-04-02
Comments most of Atom.dm

Andrew-Fall:
## Description of changes
Adds doc comments to most of atom.dm, moves reveal_blood() to forensics.dm and handle_additional_vomit_reagents() to living.dm. This commit shouldn't change any functionality.

Variables have their expected values/type listed. For example:
```
/// (1|2|3) If it shows up under UV light. 0 doesn't, 1 does, 2 is currently glowing due to UV light.
/// (DICTIONARY) A lazy map. The `key` is a MD5 playername and the `value` is the blood type.
```
When hovered over, this looks like:
image

Procs have a short description of functionality followed by a larger description and nuances. They then list parameters, expected return, raised events, see file, and TODO. For example:
```
/**
	Set the direction of this atom to `new_dir`
	- `new_dir`: The new direction the atom should face.
	- Returns: `TRUE` if the direction has been changed.
	- Events: `dir_set`
*/
```
When hovered over, this looks like:
image

I'm open to style changes to make things clearer :V

Helps future contributors implement and maintain consistent functionality.

## Authorship
Myself

### DIFF
--- a/code/game/objects/items/weapons/tanks/tanks.dm
+++ b/code/game/objects/items/weapons/tanks/tanks.dm
@@ -354,10 +354,10 @@ var/global/list/global/tank_gauge_cache = list()
 	return air_contents
 
 /obj/item/tank/assume_air(datum/gas_mixture/giver)
-	air_contents.merge(giver)
+	. = air_contents.merge(giver)
 	check_status()
 	queue_icon_update()
-	return 1
+	return .
 
 /obj/item/tank/proc/remove_air_volume(volume_to_return)
 	if(!air_contents)
--- a/code/game/objects/items/weapons/tanks/tanks.dm
+++ b/code/game/objects/items/weapons/tanks/tanks.dm
@@ -357,7 +357,6 @@ var/global/list/global/tank_gauge_cache = list()
 	. = air_contents.merge(giver)
 	check_status()
 	queue_icon_update()
-	return .
 
 /obj/item/tank/proc/remove_air_volume(volume_to_return)
 	if(!air_contents)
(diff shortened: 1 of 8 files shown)


code/game/objects/items/weapons/tanks/tanks.dm
```diff
@@ -354,10 +354,10 @@ var/global/list/global/tank_gauge_cache = list()
 	return air_contents
 
 /obj/item/tank/assume_air(datum/gas_mixture/giver)
-	air_contents.merge(giver)
+	. = air_contents.merge(giver)
 	check_status()
 	queue_icon_update()
-	return 1
+	return .
```
MistakeNot4892: This is redundant as the proc ends after this line anyway.

code/game/objects/objs.dm
```diff
@@ -41,7 +41,7 @@
 	if(loc)
 		return loc.assume_air(giver)
 	else
```
MistakeNot4892: Redundant `else` if you feel like cleaning it up while you're in the area.

code/modules/ZAS/Turf.dm
```diff
@@ -285,7 +285,7 @@
 
 /turf/simulated/assume_air(datum/gas_mixture/giver)
 	var/datum/gas_mixture/my_air = return_air()
-	my_air.merge(giver)
+	return my_air.merge(giver)
```
MistakeNot4892: `my_air?` would probably be safer.

Andrew-Fall:
Technically out of scope but not going to make a separate PR for one line changes that don't affect functionality unless you really want me to.

comma (approved):

PR #3045, opened 2023-04-04
Fix APCs resetting twice on in-round construction

out-of-phaze:
## Description of changes
Puts the `reset_area()` call in `apc/Initialize()` behind a `mapload` check, since APCs constructed during the round will trigger an area changed event on creation.

Fixes a runtime from an assert in APC area reset code failing.


afterthought2 (approved):
May be worth checking if anything else is making this mistake; this is due to a change to `movable/Initialize` to actually call `loc.Entered()` (didn't used to back in the day).

PR #3046, opened 2023-04-04
Adds basic shuttle-modifying blueprints

out-of-phaze:
## Description of changes
Adds shuttle blueprints that can add, remove, and rename shuttle areas.
Adds a Destroy hook to remove areas from shuttles on area deletion.
Adds some Destroy hooks to reduce references left on area deletion.
Blueprints now delete areas if the last turf that uses them is removed.

TODO (potentially for future PRs):
- [ ] Check that new areas are contiguous with the rest of the shuttle
- [ ] Rework blueprint UX (allow removing specific turfs from an area instead of deleting the whole thing? allow overwriting areas without deleting first if the turfs don't contain APCs?)
- [ ] Add length of shuttle name to the allowed length of area names (`Mining Vessel Operations Bay` was sadly two characters over in my testing, despite being the name the area had before I deleted it!)

Provides an easy way to modify an existing shuttle without needing a docking port.

## Authorship
Me.

## Changelog
:cl:
add: Added shuttle blueprints that can be used to add, remove, or rename shuttle areas.
/:cl:

### DIFF
--- a/code/modules/shuttles/shuttle.dm
+++ b/code/modules/shuttles/shuttle.dm
@@ -45,18 +45,19 @@
 	if(!isnull(shuttle_area))
 		if(!islist(shuttle_area))
 			shuttle_area = list(shuttle_area)
-		for(var/T in shuttle_area)
-			if(istype(T, /area)) // If the shuttle area is already a type, it does not need to be located.
-				areas += T
+		for(var/area_type in shuttle_area)
+			if(istype(area_type, /area)) // If the shuttle area is already an instance, it does not need to be located.
+				areas += area_type
 				continue
 			var/area/A
 			if(map_hash && islist(SSshuttle.map_hash_to_areas[map_hash]))
-				A = SSshuttle.map_hash_to_areas[map_hash][T] // We try to find the correct area of the given type.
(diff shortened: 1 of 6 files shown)


code/modules/shuttles/shuttle.dm
```diff
@@ -45,18 +45,19 @@
...
-				A = SSshuttle.map_hash_to_areas[map_hash][T] // We try to find the correct area of the given type.
+				A = SSshuttle.map_hash_to_areas[map_hash][area_type] // We try to find the correct area of the given type.
 			else
-				A = locate(T) // But if this is a mainmap shuttle, there is only one anyway so just find it.
+				A = locate(area_type) // But if this is a mainmap shuttle, there is only one anyway so just find it.
 			if(!istype(A))
-				CRASH("Shuttle \"[name]\" couldn't locate area [T].")
+				CRASH("Shuttle \"[name]\" couldn't locate area [area_type].")
 			areas += A
+			events_repository.register(/decl/observ/destroyed, A, src, .proc/remove_shuttle_area)
```
MistakeNot4892, line 61: Won't this avoid registering area instances by continuing up at line 51?

out-of-phaze: uhhhhh. Yes

MistakeNot4892 (approved):

PR #3052, opened 2023-04-12
All turfs now call refresh_vis_contents() during Initialize.

MistakeNot4892:
Should fix issues with turfs keeping their vis contents after changeturf.


out-of-phaze (approved):
This fixes the issue we had the other day where turfs on chlorine planets get a Mexico filter applied permanently until refresh_vis_contents() is called. 👍

PR #3055, opened 2023-04-18
Fixes stairs.

### LINKED ISSUE #3054: Stairs fall damage
#### Description of issue
If you move down the stairs ypu get fall damage

You should move down safely

#### Steps to reproduce
1. Move down on stairs
изображение
2. It hurts
изображение

Client Version: 514
Server Revision: 07ee7f7664a13e4fb209b917fcd084899250cdc0 - dev - 2023-04-12
Game ID: cno-auM0

#### Issue bingo

MistakeNot4892:
Fixes #3054


out-of-phaze (approved):
I think if `landing` is null this probably won't runtime, so it looks good.

PR #3057, opened 2023-04-20
Remove old debug output

PsyCommando:
## Description of changes
When my pr for plural English words got pushed, some debug output got merged accidentally. So I commented it out.


MistakeNot4892 (approved):
Probably should just be removed but not fussed either way.

PR #3059, opened 2023-04-21
Make overmap list indexed

PsyCommando:
## Description of changes
Since we've recently made a bunch of z-level related lists be z indexed instead of using text string keys, I made ``overmap_by_z`` and ``overmap_sectors``  also use the actual z index number instead of having to convert the number to string each times.
It's less confusing, and you don't need to go code reading each times you want to know if you need a number or string for accessing that list.

I also renamed ``overmap_sectors`` to ``overmap_sectors_by_z`` for clarity and to match the other lists indexed by z levels.

**Note** : The ``overmap_by_z`` list probably didn't need that since it will always only have one or two entries. But, turns out, it's called in the overmap turfs's initialize proc. So I'm guessing that it might actually be worth it to skip the cost of string conversion for several hundred turfs.

Slightly improve clarity, makes accessing those lists a tiny bit faster.

### DIFF
--- a/code/game/machinery/hologram.dm
+++ b/code/game/machinery/hologram.dm
@@ -111,7 +111,7 @@ var/global/const/HOLOPAD_MODE = RANGE_BASED
 				var/zlevels_long = list()
 				if(holopadType == HOLOPAD_LONG_RANGE && length(reachable_overmaps))
 					for(var/zlevel in global.overmap_sectors)
-						var/obj/effect/overmap/visitable/O = global.overmap_sectors[num2text(zlevel)]
+						var/obj/effect/overmap/visitable/O = global.overmap_sectors[zlevel]
 						if(!isnull(O) && (O.overmap_id in reachable_overmaps))
 							zlevels_long |= O.map_z
 				for(var/obj/machinery/hologram/holopad/H in SSmachines.machinery)
--- a/code/game/machinery/hologram.dm
+++ b/code/game/machinery/hologram.dm
@@ -110,8 +110,8 @@ var/global/const/HOLOPAD_MODE = RANGE_BASED
 				var/zlevels = SSmapping.get_connected_levels(z)
 				var/zlevels_long = list()
 				if(holopadType == HOLOPAD_LONG_RANGE && length(reachable_overmaps))
-					for(var/zlevel in global.overmap_sectors)
-						var/obj/effect/overmap/visitable/O = global.overmap_sectors[zlevel]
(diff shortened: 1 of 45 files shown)


PsyCommando:
Funny, the unit test failing earlier revealed an old problem with shield generators not clearing their overmap marker registration, because they were doing so after they were moved to nullspace and the proc doing it would just silently fail. But with the indexed list it crashed immediately when it tried to access the "null" index of the list.

code/game/machinery/hologram.dm
```diff
@@ -110,8 +110,8 @@ var/global/const/HOLOPAD_MODE = RANGE_BASED
 				var/zlevels = SSmapping.get_connected_levels(z)
 				var/zlevels_long = list()
 				if(holopadType == HOLOPAD_LONG_RANGE && length(reachable_overmaps))
-					for(var/zlevel in global.overmap_sectors)
-						var/obj/effect/overmap/visitable/O = global.overmap_sectors[num2text(zlevel)]
+					for(var/zlevel in global.overmap_sectors_by_z)
+						var/obj/effect/overmap/visitable/O = global.overmap_sectors_by_z[zlevel]
```
MistakeNot4892, line 114: I don't think this will work, it's a linear list now, not associative. You should be able to do `for(var/obj/effect/overmap/visitable/O in global.overmap_sectors_by_z)` or `for(var/zlevel = 1 to length(global.overmap_sectors_by_z)`.

maps/exodus/exodus_setup.dm
```diff
@@ -13,8 +13,10 @@
 		welcome_text += "Time since last port visit:<br /><b>[rand(60,180)] days</b><br /><hr>"
 		welcome_text += "Scan results show the following points of interest:<br />"
 
-		for(var/zlevel in global.overmap_sectors)
-			var/obj/effect/overmap/visitable/O = global.overmap_sectors[zlevel]
+		for(var/zlevel in global.overmap_sectors_by_z)
```
MistakeNot4892, line 16: Per above I'm not sure this will work.

code/game/machinery/hologram.dm
PsyCommando: True, I'll have to change that.

PsyCommando:
I'm kinda split on this PR. I might close it, since I don't feel it's as relevant as I thought it was.

PsyCommando:
I'll close this. It doesn't really feel that relevant anymore.

PR #3060, opened 2023-04-21
deltheworld prints full exception data when catching one

PsyCommando:
## Description of changes
The "del the world" unit test wouldn't print the line number or filename when it caught an exception, which makes troubleshooting pretty hard. So, I added the missing bits to the failure output.


MistakeNot4892 (approved):
@out-of-phaze

out-of-phaze (approved):
Didn't know exception stringification didn't include that, whoops.

PR #3066, opened 2023-04-24
Adds plastic wall variants and mapped wall preview colors

out-of-phaze:
## Description of changes
Fixes duplicated/conflicting wall color definitions.
Adds mapping preview colors to walls, based on the material colors.
Adds premade plastic walls for mapping. These are useful because they can be removed with just a crowbar instead of a welder.
Also adds a 'facade' type.

Makes it clearer what walls will look like in-game when mapping stuff.
Also, more options for material walls.

### DIFF
--- a/code/game/turfs/simulated/wall_types.dm
+++ b/code/game/turfs/simulated/wall_types.dm
@@ -57,6 +57,11 @@
 	icon_state = "metal"
 	material = /decl/material/solid/metal/iron
 
+/turf/simulated/wall/plastic
+	color = COLOR_EGGSHELL
+	icon_state = "plastic"
+	material = /decl/material/solid/plastic
+
 /turf/simulated/wall/sandstone
 	color = COLOR_GOLD
 	icon_state = "stone"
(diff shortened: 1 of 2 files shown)


code/game/turfs/simulated/wall_types.dm
```diff
@@ -54,9 +53,19 @@
...
 	icon_state = "metal"
 	material = /decl/material/solid/metal/iron
 
+/turf/simulated/wall/plastic
+	color = COLOR_EGGSHELL
+	icon_state = "plastic"
+	material = /decl/material/solid/plastic
+
+// A plastic wall with a plastic girder. Very flimsy but very easy to move or remove with just a crowbar.
+/turf/simulated/wall/plastic/facade
```
MistakeNot4892, line 66: Are these buildable? I would have thought plastic would fail the ~~vibe~~ hardness check for girders.

out-of-phaze: Ah, good question. Truth be told I would like it if we could just build the plastic facades without girders at all, but I don't think that's supported.

MistakeNot4892: I guess it doesn't really matter either way if it's buildable.

MistakeNot4892 (approved):

PR #3072, opened 2023-04-24
Remove germ check from necrotic tissue surgery

out-of-phaze:
## Description of changes
Removes a germ check from necrotic tissue surgery that made it so you could only do the surgery on infected (not necrotic) organs.

Now you can actually do the surgery.


MistakeNot4892 (approved):
I think(?) these germ checks are inverted and are meant to ensure you're not debriding a rotten limb, but not fussed either way.

PR #3076, opened 2023-04-27
Adapted airlock radio fixes.

afterthought2:
## Description of changes

This is an adaptation of the fix in #3004. 

There, @PsyCommando observed that the radio signal check used by airlock controllers, `data["pressure"]`, is incorrect when a `null` or `0` pressure is transmitted, and these are valid values; this leads to some potential bugs with the internal logic.

I then torpedoed that PR by requesting the check you see here, `("pressure" in data)`, be used instead, arguing that having _no check at all_ leads to undesirable behavior if one modifies the now-modifiable radio signalling hardware in the pumps or elsewhere. @PsyCommando noted that there are a bunch of similar places directly below that receive boolean or numerical variables but make no checks at all. This updates all of them.

I remark that #3004 makes an additional change to pump code that I do not reproduce here; it seems unnecessary to me, but that is purely from code inspection.

It may help address intermittent airlock issues with cycling to/from vacuum, and make modifying pump radio keys in round behave more intuitively with airlock controllers.

## Authorship
@PsyCommando, me

### DIFF
--- a/code/game/machinery/embedded_controller/airlock_program.dm
+++ b/code/game/machinery/embedded_controller/airlock_program.dm
@@ -89,24 +89,30 @@
 	if(!receive_tag) return
 
 	if(receive_tag==tag_chamber_sensor)
-		if(signal.data["pressure"])
+		if("pressure" in signal.data)
 			memory["chamber_sensor_pressure"] = text2num(signal.data["pressure"])
 
 	else if(receive_tag==tag_exterior_sensor)
-		memory["external_sensor_pressure"] = text2num(signal.data["pressure"])
+		if("pressure" in signal.data)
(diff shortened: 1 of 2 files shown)


comma (approved):

out-of-phaze (approved):
No clue why it has text2num there; if it was consistently text then the assoc check would work, but it isn't.
```js
		if(IC_FORMAT_NUMBER, IC_FORMAT_INDEX)
			if(istext(command_args))
				return text2num(command_args)
			if(isnum(command_args))
				return command_args
```
Because of this, I'm pretty sure it'll never be text, so the assoc check is invalid.

This could have performance implications, so maybe `!isnull(signal.data["pressure"])` would be better if that's the case, but until that happens, this works.

PR #3081, opened 2023-05-10
Implementing general inventory slots.

MistakeNot4892:
## Description of changes
Implements datum inventory slots. Splitting out of #2533.

- Inventory slot datums have been added to replace the extant slot definitions on species and in legacy code.
- `held_item_slots` -> `_held_item_slots`, is now a linear list of strings only used for checking if something is a held slot.
- `inventory_slots` -> `_inventory_slots`, associative list of slot string to slot datum.

**TODO:**
- [x] Fix QDELETED check preventing item drops.
- [x] Work out what is QDELing inventory slots and causing them to vanish.
- [x] Work out why nothing will equip to hand slots.
- [x] Check why coverage isn't being respected.
- [x] Work out why pockets don't work.
- [ ] Test more.

**Tested:**
- Basic equipping and unequipping.
- Moving items to and from hands.
- Moving items to and from storage.
- Checking robots still work as expected.
- Stripping and reequipping.

Flexible modular inventory code.

## Authorship
Myself.

## Changelog
In theory, nothing player-facing.


code/modules/mob/inventory.dm
```diff
@@ -247,29 +204,13 @@
...
-		return TRUE
-	if(W == _wear_mask)
-		_wear_mask = null
-		update_inv_wear_mask(0)
-		return TRUE
-	return FALSE
-
 /mob/proc/unequip(obj/W)
 	SHOULD_CALL_PARENT(TRUE)
 	if(!istype(W))
```
out-of-phaze, line 201: This is where the QDELETED check was that broke item dropping. It looks to me like this should work fine now.

code/datums/inventory_slots/slots/slot_ears.dm
```diff
@@ -0,0 +1,35 @@
...
+	if(.)
+		for(var/slot in global.ear_slots)
+			if(slot == slot_id)
+				continue
+			var/datum/inventory_slot/inv_slot = user.get_inventory_slot_datum(slot)
+			if(inv_slot?.get_equipped_item() == prop)
+				inv_slot.clear_slot()
+
+/datum/inventory_slot/ear/can_equip_to_slot(var/mob/user, var/obj/item/prop, var/slot, var/disable_warning, var/force)
+	return ..() && prop.w_class <= ITEM_SIZE_TINY
```
out-of-phaze: This is why CI is failing. `/obj/item/clothing/head/hairflower`, used in the test, is not `ITEM_SIZE_TINY` and therefore can't be equipped now.
```
			if((w_class > ITEM_SIZE_TINY) && !(slot_flags & SLOT_EARS))
				return FALSE
```
It seems you've inverted the current check, which is intended to allow tiny items to be hidden in the ear slot even without the SLOT_EARS flag.
```suggestion
	return ..() || prop.w_class > ITEM_SIZE_TINY
```
I checked and couldn't find any other cases where this logic was inverted; pockets use something similar and it works fine.

out-of-phaze (changes requested):
Preliminary review to address CI failure/testing regression from an inverted check.

code/datums/inventory_slots/slots/slot_ears.dm
MistakeNot4892: Good eye for the issue, suggestion is not quite right - fix was to move the slot checking to a proc and `..() || prop.w_class <= ITEM_SIZE_TINY`

out-of-phaze: Gah, don't know how I messed the logic up that bad.

out-of-phaze (approved):
Really impressed with the expanded unit test coverage for this, will hopefully make future changes to this system less error-prone.

PR #3087, opened 2023-05-28
Turret Rewrite

NataKilar:
## Description of changes
Updates and finishes @Neerti's turret rewrite in https://github.com/NebulaSS13/Nebula/pull/1424

Turret Changes
* Turrets now use state machines for idling, tracking, engaging and reloading
* Turrets no longer search for targets in their Process proc, instead using the proximity trigger datums to locate targets and update their target angle
* Turrets no longer lock onto their target angle instantly, instead using 2 decisecond loopable timers to update their angle.
* Turrets have a certain angular range determined by their ``traverse`` variable
* Turrets can now also load and fire ballistic weapons using the ammo box stock part. Currently this is restricted to magazine loaded weapons since shotguns, rifles etc. would require more extensive refactoring
* Buildable sentry turrets are machine subtypes. They are tied to computer networks, and controllable remotely via a computer program. Targeting settings are tied to crew records on the network.
* Left the old turret type (``/obj/machinery/porta_turret``) in, since the behavior is not 1:1. The turret frame recipe has been removed.

Other changes
* Added a generic hostility datum for determining valid targets
* Added a proximity trigger datum for an angular range
* Proximity triggers now update when opaque movables are moved/destroyed/updated
* Projectile weapons can now handle non-living firers
* Fixes shielding stock parts being able to absorb damage infinitely.

I can also split this PR into the gun firing changes and turret additions if requested.

Gives turrets more counter play and variability, since they no longer track targets instantly. Remote control of turrets means they can be used as weapons installations on shuttles. Tying turrets to the network means targeting based off crew records is no longer universal.

Reworked turrets are more performant, since they no longer look for targets every Process tick. The timer based tracking had a negligible performance impact in my testing. Still, it's worth considering adding a separate timer subsystem for turrets, or throttling tracking another way if necessary.

## Authorship
The majority of the code work was done by Neerti, particularly the turret tracking, state machine, and hostility logic. I updated the code to dev, reworked turret charging / ballistic reloading, changed up some of the angle math, and added the network functionality. Changes to the proximity trigger datum were done by myself.

## Changelog
:cl:
add: Reworked turrets into sentry turrets, with improved target tracking
fix: Shielding stock parts can no longer absorb an infinite amount of damage
/:cl:


nano/templates/network_turret.tmpl
```diff
@@ -0,0 +1,74 @@
+<div style="min-width: 480px">
+	<div class='item'>
```
comma: This and next template seem to be pretty identical, as far as I can tell only difference is the network settings part.

Maybe have it be one template but with network settings gated behind if check with network turrets setting that flag in ui data?

code/datums/hostility/hostility.dm
```diff
@@ -0,0 +1,72 @@
+// A small object that is useful for recycling hostility logic in things such as mobs or turrets.
+
+/datum/hostility
+	var/datum/holder = null
```
MistakeNot4892: Per Discord, I'm fairly sure you could pass holder in as an argument and make `/datum/hostility` an immutable singleton, which would make it scale better if it ever gets expanded to mob AI down the track as the comment implies.

code/unit_tests/proximity_tests.dm
```diff
@@ -129,7 +133,7 @@
 
 /obj/proximity_listener/proc/SetTrigger(trigger_type, listener_flags)
 	QDEL_NULL(trigger)
-	trigger = new trigger_type(src, /obj/proximity_listener/proc/OnTurfEntered, /obj/proximity_listener/proc/OnTurfsChanged, 7, listener_flags)
+	trigger = new trigger_type(src, /obj/proximity_listener/proc/OnTurfEntered, /obj/proximity_listener/proc/OnTurfsChanged, 7, listener_flags, null, 90, 270)
```
MistakeNot4892, line 136: Aren't these new arguments meant to define an angle? What's the upshot of passing these hardcoded values in here?

code/modules/projectiles/guns/projectile/bolt_action.dm
```diff
@@ -24,11 +24,13 @@
...
 
-/obj/item/gun/projectile/bolt_action/handle_post_fire(mob/user, atom/target, var/pointblank=0, var/reflex=0)
+/obj/item/gun/projectile/bolt_action/handle_post_fire(atom/movable/firer, atom/target, var/pointblank=0, var/reflex=0)
 	..()
-	if(user && user.skill_check(SKILL_WEAPONS, SKILL_PROF))
-		to_chat(user, "<span class='notice'>You work the bolt open with a reflexive motion, ejecting [chambered]!</span>")
-		unload_shell()
+	if(ismob(firer))
+		var/mob/user = firer
+		if(user && user.skill_check(SKILL_WEAPONS, SKILL_PROF))
```
MistakeNot4892: Redundant `user` check after the `ismob()` above.

code/modules/projectiles/guns/projectile.dm
```diff
@@ -231,7 +233,7 @@
 
 /obj/item/gun/projectile/afterattack(atom/A, mob/living/user)
 	..()
-	if(auto_eject && ammo_magazine && ammo_magazine.stored_ammo && !ammo_magazine.stored_ammo.len)
+	if(auto_eject && isliving(user) && ammo_magazine && ammo_magazine.stored_ammo && !ammo_magazine.stored_ammo.len)
```
MistakeNot4892: Why is the magazine removal being gated by the user being living?

code/modules/projectiles/guns/launcher/rocket.dm
```diff
@@ -43,6 +43,6 @@
 		return M
 	return null
 
-/obj/item/gun/launcher/rocket/handle_post_fire(mob/user, atom/target)
+/obj/item/gun/launcher/rocket/handle_post_fire(atom/movable/firer, atom/target)
 	log_and_message_admins("fired a rocket from a rocket launcher ([src.name]) at [target].")
```
MistakeNot4892: I think this uses `usr` for the source key/name, so it might display strangely with a non-mob.

code/modules/projectiles/guns/launcher/crossbow.dm
```diff
@@ -66,13 +66,13 @@
 /obj/item/gun/launcher/crossbow/update_release_force()
 	release_force = tension*release_speed
 
-/obj/item/gun/launcher/crossbow/consume_next_projectile(mob/user=null)
-	if(tension <= 0)
-		to_chat(user, "<span class='warning'>\The [src] is not drawn back!</span>")
+/obj/item/gun/launcher/crossbow/consume_next_projectile(atom/movable/firer)
+	if(tension <= 0 && isliving(firer))
+		to_chat(firer, "<span class='warning'>\The [src] is not drawn back!</span>")
```
MistakeNot4892: `SPAN_WARNING` would be appreciated while you're in the area.

code/modules/projectiles/guns/energy/laser.dm
```diff
@@ -45,12 +45,12 @@
...
 	return 1
 
-/obj/item/gun/energy/laser/practice/handle_post_fire(mob/user, atom/target, var/pointblank=0, var/reflex=0)
+/obj/item/gun/energy/laser/practice/handle_post_fire(atom/movable/firer, atom/target, var/pointblank=0, var/reflex=0)
 	..()
 	if(hacked())
 		max_shots--
 		if(!max_shots) //uh hoh gig is up
-			to_chat(user, "<span class='danger'>\The [src] sizzles in your hands, acrid smoke rising from the firing end!</span>")
+			to_chat(firer, "<span class='danger'>\The [src] sizzles in your hands, acrid smoke rising from the firing end!</span>")
```
MistakeNot4892: `SPAN_DANGER` would be nice while you're editing the line.

code/modules/projectiles/gun.dm
```diff
@@ -467,23 +489,23 @@
 
 	//shooting while in shock
 	var/shock_dispersion = 0
-	if(istype(user, /mob/living/carbon/human))
-		var/mob/living/carbon/human/mob = user
+	if(istype(firer, /mob/living/carbon/human))
```
MistakeNot4892, line 492: `ishuman`

MistakeNot4892 (changes requested):
Incomplete review.

code/unit_tests/proximity_tests.dm
NataKilar: The passed angles only affect the angle restricted version of the proximity detector. The hardcoded values just define half of a circular range, which allows the unit test to make sure the angle restricted proximity detector is still working.

code/datums/hostility/hostility.dm
NataKilar: Done. Left securitron hostility without reference to a holder's variables because it's currently unused and I don't feel like refactoring secbot.dm at the moment.

nano/templates/network_turret.tmpl
NataKilar: Seems reasonable, I merged the templates

code/modules/projectiles/guns/projectile/bolt_action.dm
NataKilar: Fixed.

code/modules/projectiles/guns/projectile.dm
NataKilar: Not really sure, this was a direct port. Removed.

code/modules/projectiles/guns/launcher/rocket.dm
NataKilar: Added firer as an argument.

code/modules/projectiles/guns/launcher/crossbow.dm
NataKilar: Fixed (along with others in the file)

code/modules/projectiles/guns/energy/laser.dm
NataKilar: Ditto.

code/modules/projectiles/gun.dm
NataKilar: Done

MistakeNot4892 (approved):
My concerns have been addressed.

PR #3088, opened 2023-05-29
Terminal script runner

comma:
Adds a button to file browser to run text file as a script in terminal Just feeds it line by line to a terminal, no loops or logic

## Description of changes
Adds RUN button to file viewer that starts a terminal and feeds contents of file to it line by line
Effectively very dumb scripting engine, no loops or conditions or variables, but still handy if you want to open and bolt 4 doors in timely manner.

Fun things with network commands like door manipulations would be way more useful if done faster instead of typing commands manually every time.

## Authorship

Me

## Changelog
:cl:
add: Terminal scripts. Write a file, put a terminal command on each line, and press RUN when viewing it in file editor.
/:cl:

### DIFF
--- a/code/modules/modular_computers/file_system/programs/generic/file_manager.dm
+++ b/code/modules/modular_computers/file_system/programs/generic/file_manager.dm
@@ -14,7 +14,7 @@
 	var/error
 	var/list/disks = list() // List of list(/datum/file_storage, weakref(directory file)).
 	var/current_index
-	
+
 	var/datum/file_transfer/current_transfer	//ongoing file transfer between file_storage datums
 
 /datum/computer_file/program/filemanager/on_shutdown()
@@ -70,7 +70,7 @@
 		if(!avail_disk.hidden && avail_disk.check_access(accesses))
 			computer.unmount_storage(root_name)
 			return TOPIC_REFRESH
-		
+
 		to_chat(user, SPAN_WARNING("You lack permission to unmount this network drive!"))
 		return TOPIC_HANDLED
 
@@ -81,15 +81,15 @@
 			return TOPIC_REFRESH
 		current_index = disk_index
 		return TOPIC_REFRESH
-	
+
 	if(!current_index || current_index > disks.len)
 		current_index = 0
 		return TOPIC_REFRESH
-	
+
 	var/datum/file_storage/current_disk = disks[current_index]?[1]
 	if(!current_disk)
 		return TOPIC_HANDLED
-	
+
 	if(href_list["PRG_exit_disk"])
 		if(!current_index)
 			return TOPIC_HANDLED
@@ -132,7 +132,7 @@
 			else
 				to_chat(user, SPAN_WARNING("You do not have permission to read this file."))
 				return TOPIC_HANDLED
-			
+
 		if(istype(F, /datum/computer_file/data))
 			if(F.get_file_perms(accesses, user) & OS_READ_ACCESS)
 				open_file = href_list["PRG_openfile"]
@@ -165,7 +165,7 @@
 				else
 					to_chat(user, SPAN_WARNING("Unable to create new file. The hard drive may be non-functional."))
 					return TOPIC_REFRESH
-	
+
 	if(href_list["PRG_newdir"])
 		. = TOPIC_REFRESH
 		var/newname = sanitize_for_file(input(usr, "Enter directory name or leave blank to cancel:", "New directory"))
@@ -183,7 +183,7 @@
 				else
 					to_chat(user, SPAN_WARNING("Unable to create new directory. The hard drive may be non-functional."))
 					return TOPIC_REFRESH
-	
+
 	if(href_list["PRG_deletefile"])
 		var/datum/computer_file/del_file = current_disk.get_file(href_list["PRG_deletefile"], current_directory, accesses, user)
 		var/deleted = current_disk.delete_file(del_file, accesses, user)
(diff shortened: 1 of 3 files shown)


code/modules/modular_computers/file_system/programs/generic/file_manager.dm
```diff
@@ -325,6 +325,20 @@
...
+		. = TOPIC_HANDLED
+		if(!open_file)
+			return
+		var/datum/computer_file/program/P = current_disk.get_file(open_file, current_directory)
+		if(istype(P))
+			computer.run_program(P.filename)
+			return TOPIC_HANDLED
+		var/datum/computer_file/data/F = current_disk.get_file(open_file, current_directory)
+		if(!F || !istype(F))
+			return
```
MistakeNot4892, line 338: What does a null return in OnTopic do?

code/modules/modular_computers/os/_os.dm
```diff
@@ -309,4 +309,20 @@
...
+	var/datum/terminal/T = has_terminal(user)
+	if(!istype(T))
+		return  TOPIC_HANDLED
+
+	T.show_terminal(user)
+	T.append_to_history(">Running '[script.filename].[script.filetype]'")
+	var/list/lines = splittext(script.stored_data, "\[br\]")
+	for(var/line in lines)
+		var/output = T.parse(line, user)
+		if(QDELETED(T)) // Check for exit.
```
MistakeNot4892, line 325: Should this be checked before `T.parse()` above or is `parse()` what qdels T?

nano/templates/file_manager.tmpl
```diff
@@ -23,6 +23,7 @@
 		{{:helper.link('CLOSE', null, { "PRG_closefile" : 1 })}}
 		{{:helper.link('EDIT', null, { "PRG_edit" : 1 })}}
 		{{:helper.link('PRINT', null, { "PRG_printfile" : 1 })}}
+		{{:helper.link('RUN', null, { "PRG_runfile" : 1 })}}
```
MistakeNot4892, line 26: Idle thought: force them to `chmod a+x` before they can run the file.

code/modules/modular_computers/os/_os.dm
comma: Parse does, exit command specifically

code/modules/modular_computers/file_system/programs/generic/file_manager.dm
comma: Nothing, it's TOPIC_NOACTION, mostly just checked by overrides in subtypes (so nothing at all in this case)

nano/templates/file_manager.tmpl
comma: Maybe once permissions settings are more exposed, currently it's just done via some arcane terminal fuckery

MistakeNot4892 (approved):

code/modules/modular_computers/file_system/programs/generic/file_manager.dm
out-of-phaze: It's actually `TOPIC_HANDLED` here because of `. = TOPIC_HANDLED`. I'm pretty sure it's still fine though.

code/modules/modular_computers/file_system/programs/generic/file_manager.dm
```diff
@@ -325,6 +325,20 @@
...
 
+	if(href_list["PRG_runfile"])
+		. = TOPIC_HANDLED
+		if(!open_file)
+			return
+		var/datum/computer_file/program/P = current_disk.get_file(open_file, current_directory)
+		if(istype(P))
+			computer.run_program(P.filename)
+			return TOPIC_HANDLED
+		var/datum/computer_file/data/F = current_disk.get_file(open_file, current_directory)
```
out-of-phaze, line 336: These point to the same file, we're just checking a different type.
```suggestion
		var/datum/computer_file/data/F = P
```
Another alternative could be:
```dm
		var/datum/computer_file/runnable_file = current_disk.get_file(open_file, current_directory)
		if(istype(runnable_file, /datum/computer_file/program))
			computer.run_program(runnable_file.filename)
			return TOPIC_HANDLED
		if(!runnable_file || !istype(runnable_file, /datum/computer_file/data))
			return TOPIC_NOACTION
		computer.run_script(usr, runnable_file)
		return TOPIC_HANDLED
```
Since we're not using any methods defined on those subtypes, we can avoid the cast altogether and just check the type explicitly.

code/modules/modular_computers/os/_os.dm
```diff
@@ -309,4 +309,20 @@
...
+
+/datum/extension/interactive/os/proc/run_script(mob/user, var/datum/computer_file/data/script)
+	open_terminal(user)
+	var/datum/terminal/T = has_terminal(user)
+	if(!istype(T))
+		return  TOPIC_HANDLED
+
+	T.show_terminal(user)
+	T.append_to_history(">Running '[script.filename].[script.filetype]'")
+	var/list/lines = splittext(script.stored_data, "\[br\]")
```
out-of-phaze, line 322: Are newlines in the script automatically converted to pencode, or do you have to explicitly write `[br]` to separate lines in scripts?

out-of-phaze: Looks like it may be? When you input it, it replaces newlines with `[br]` tags, and then when you edit it, it replaces the [br] tags with newlines. Good enough for me.

code/modules/modular_computers/file_system/programs/generic/file_manager.dm
out-of-phaze: Actually, I'll PR this myself with some of my other OS/filesystem changes, probably.

out-of-phaze (approved):
Had a few thoughts about extending this, but nothing blocking. I'll throw some related stuff into my bundled OS/filesystem fixes PR after this is merged.

PR #3089, opened 2023-06-01
Planet gen second pass

PsyCommando:
## Description of changes
Reworked planetoid stuff so it can allow for loading mapped planetoid from template instead of only randomly generating planetoid templates.

* Moved a big chunk of planetoid generation and data setup to planetoid_data, to allow defining a lot of the planetoid data at compile time with minimal work involved.
* Separated the ``/datum/map_template/planetoid`` type into ``/datum/map_template/planetoid`` and ``/datum/map_template/planetoid/random``. The base planetoid one allows loading from map files, while the random subtype is for generating planetoid at runtime.
* Fixed level_data bound calculation issues when dealing with templates loaded at another origin than 1,1.
* Added some mapper icons to the planet landmarks and level related landmarks.


PsyCommando:
Just leaving this here. But, the random test planet I slapped into the tradeship map isn't meant to get merged in that spot. I just don't know where to put it tbh..

code/modules/overmap/stellar_bodies/star_skybox.dm
```diff
@@ -0,0 +1 @@
+/obj/effect/overmap/star/get_skybox_representation()
```
comma: does this really need a separate file?

out-of-phaze:
Please add all of the changes (the stellar object/star stuff, for example) to the PR body.

PsyCommando:
> Please add all of the changes (the stellar object/star stuff, for example) to the PR body.

The star stuff shouldn't be in anymore? I forced pushed it out. It was a leftover from a cherry pick.

out-of-phaze:
> > Please add all of the changes (the stellar object/star stuff, for example) to the PR body.
> 
> The star stuff shouldn't be in anymore? I forced pushed it out. It was a leftover from a cherry pick.

It was in when I left the comment.

PsyCommando:
**Update:**
The problem with unit tests runtiming is due to an existing bug. Normally the template code silently skip over loading templates that are too big for where they're being placed, but I added some CRASH to actually make it warn us something went wrong and the template was not loaded.

Currently, it fails on nexus and the example map. That's because some of the antag sites are pretty big. Like the heist site is 70x70. And nexus' world maxx and maxy are set to 129x129 during loading. The unit test code force all tempaltes to be loaded with their lower left corner at ``world.maxx / 2`` and ``world.maxy / 2`` which means the template doesn't have enough room to be loaded fully and trips the sanity checks. 

I'll try fixing that in a separate PR but I'd welcome some suggestions on what to do, because there's a number of ways I could handle this. But some might disrupt the map more than others (aka enlarging the map).

PsyCommando:
Okay so I fixed the template issues crashing the unit tests in #3105.

MistakeNot4892:
Centering logic still seems iffy:
````
 ## UNIT_TEST ##: Loading template 'Mercenary Base' (/datum/map_template/ruin/antag_spawn/mercenary) at space (/turf/space) ([0x100fdb5]) (50,50,7) (/area/space)
[23:00:00] Runtime in _map_template.dm,218: Couldn't fit the entire template 'Mercenary Base' (size: 100x100) between lower left corner (50, 50)(WORLD CENTER) and upper right corner (150, 150) in current world size (100x100x7).
  proc name: load (/datum/map_template/proc/load)
  src: Mercenary Base (/datum/map_template/ruin/antag_spawn/mercenary)
  call stack:
  Mercenary Base (/datum/map_template/ruin/antag_spawn/mercenary): load(space (50,50,7) (/turf/space), 1)
  Unit Tests (/datum/controller/subsystem/unit_tests): load template(Mercenary Base (/datum/map_template/ruin/antag_spawn/mercenary))
  Unit Tests (/datum/controller/subsystem/unit_tests): load map templates()
  Unit Tests (/datum/controller/subsystem/unit_tests): Initialize(857251)
  Unit Tests (/datum/controller/subsystem/unit_tests): DoInitialize(857251)
  ````

**Comment by MistakeNot4892** [comment]
>The unit test code force all tempaltes to be loaded with their lower left corner at world.maxx / 2 and world.maxy / 2 which means the template doesn't have enough room to be loaded fully and trips the sanity checks.

This feels like possibly it predates the centering logic or the author wasn't aware of it.  Likewise a bit perplexed at the nexus testing forcing the map size to be set.

**Comment by PsyCommando** [comment]
Yeah, I'm thinking I just messed up when I resolved the conflict. I'll have to go in and check the math and conditionals.

**Comment by PsyCommando** [comment]
Seems like it works now.

**Comment by comma** [comment]
I don't think Neutralia should be commited, maybe to a separate folder as discussed in discord.

**Comment by PsyCommando** [comment]
> I don't think Neutralia should be commited, maybe to a separate folder as discussed in discord.

Yeah, that's what I was working on

File: code/unit_tests/~unit_test_subsystems.dm [hunk 2]
```diff
@@ -37,12 +37,21 @@ SUBSYSTEM_DEF(unit_tests)
 	log_unit_test("[queue.len] unit tests loaded.")
 	. = ..()
 
+///Returns whether a template should be loaded for all unit tests or it's tested in a specific unit test on its own.
+/datum/controller/subsystem/unit_tests/proc/is_tested_separately(var/datum/map_template/map_template)
+	for(var/cat in map_template.template_categories)
```
MistakeNot4892, line 48: I would suggest something like:
````
/datum/controller/subsystem/unit_tests/proc/is_tested_separately(var/datum/map_template/map_template)
	var/static/list/separately_tested_categories = list(
		MAP_TEMPLATE_CATEGORY_AWAYSITE,
		MAP_TEMPLATE_CATEGORY_PLANET,
		MAP_TEMPLATE_CATEGORY_EXOPLANET
	)
	for(var/cat in map_template.template_categories)
		if(cat in separately_tested_categories)
			return TRUE
	return FALSE
	````

File: code/unit_tests/~unit_test_subsystems.dm [hunk 3]
```diff
@@ -37,12 +37,21 @@ SUBSYSTEM_DEF(unit_tests)
...
+	return FALSE
+
 /datum/controller/subsystem/unit_tests/proc/load_map_templates()
 	for(var/map_template_name in SSmapping.map_templates)
 		var/datum/map_template/map_template = SSmapping.get_template(map_template_name)
 		// Away sites are supposed to be tested separately in the Away Site environment
-		if(MAP_TEMPLATE_CATEGORY_AWAYSITE in map_template.template_categories)
-			report_progress("Skipping template '[map_template]' ([map_template.type]): Is an Away Site")
+		if(is_tested_separately(map_template))
+			report_progress("Skipping template '[map_template]' ([map_template.type]): Is tested separately")
```
MistakeNot4892: Would be nice if either this casing was made Like The Others, or the others got corrected.

code/modules/xgm/xgm_gas_mixture.dm
```diff
@@ -19,10 +19,17 @@
 	///The last cached color of the gas mixture
 	var/tmp/cached_mix_color
 
-/datum/gas_mixture/New(_volume = CELL_VOLUME, _temperature = 0, _group_multiplier = 1)
-	volume = _volume
-	temperature = _temperature
-	group_multiplier = _group_multiplier
+/datum/gas_mixture/New(_volume, _temperature, _group_multiplier)
```
MistakeNot4892, line 22: What is the purpose of adding the null gates and removing the default arguments?

code/modules/xgm/xgm_gas_mixture.dm
```diff
@@ -19,10 +19,17 @@
...
+	if(!isnull(_volume))
+		volume = _volume
+	if(!isnull(_temperature))
+		temperature = _temperature
+	if(!isnull(_group_multiplier))
+		group_multiplier = _group_multiplier
+
+	//Since we may have values defined on creation, update everything.
+	if(volume)
```
MistakeNot4892: Not sure I'm a a fan of universally calling update_values on newly created gas mixtures. This is going to add a fair bit of proc overhead.

code/modules/materials/_materials.dm
```diff
@@ -428,7 +428,7 @@ INITIALIZE_IMMEDIATE(/obj/effect/gas_overlay)
 
 //Clausius–Clapeyron relation
 /decl/material/proc/get_boiling_temp(var/pressure = ONE_ATMOSPHERE)
-	return (1 / (1/max(boiling_point, TCMB)) - ((R_IDEAL_GAS_EQUATION * log(pressure / ONE_ATMOSPHERE)) / (latent_heat * molar_mass)))
+	return (1 / (1/max(boiling_point, TCMB)) - ((R_IDEAL_GAS_EQUATION * ((pressure > 0)? log(pressure / ONE_ATMOSPHERE) : 0)) / (latent_heat * molar_mass)))
```
MistakeNot4892: Why is this being called with null/negative pressure?

code/modules/maps/_map_template.dm
```diff
@@ -112,9 +126,33 @@
 	if(destroyed_area.saved_map_hash)
 		SSshuttle.map_hash_to_areas[destroyed_area.saved_map_hash] -= destroyed_area
 
-/datum/map_template/proc/load_new_z(no_changeturf = TRUE, centered=TRUE)
-	var/x = max(round((world.maxx - width)/2), 1)
-	var/y = max(round((world.maxy - height)/2), 1)
+///Handle loading a single map path its bottom left corner starting at x,y,z.
+/// Returns a /datum/map_load_metadata if loading was successful.
+/// Meant to be overridden for handling extra per-map file processing.
+/datum/map_template/proc/load_a_path(var/mappath, var/x, var/y, var/z, var/list/bounds, var/list/initialized_areas_by_type, var/no_changeturf, var/cropMap)
```
MistakeNot4892: Not sure I like this proc name. Could be something like `load_single_level`?

maps/~mapsystem/maps.dm
```diff
@@ -252,37 +252,47 @@ var/global/const/MAP_HAS_RANK = 2		//Rank system, also togglable
 
 /datum/map/proc/build_exoplanets()
 #ifdef UNIT_TEST
-	report_progress("Unit testing, so not loading exoplanets.")
+	report_progress("Unit testing, so not loading planets.")
```
MistakeNot4892, line 255: Should probably rename the proc to `build_planets` if you're renaming the strings here.

PsyCommando:
Btw, this should be good to review now. Planet unit test is setup and working, And I'm pretty sure all issues have been fixed now.

PsyCommando:
Oh crap, github didn't refresh for me and I didn't see all those comments....

code/unit_tests/~unit_test_subsystems.dm
PsyCommando: Alright, though, might make sense to have that list defined in the unit test datum instead? So it's a bit more visible and can be re-used if someone needs that somewhere else?

code/unit_tests/~unit_test_subsystems.dm
PsyCommando: You mean the report_progress strings? I can do that.

code/modules/xgm/xgm_gas_mixture.dm
PsyCommando: This is so you can define a gas mixture at compile time. Also, using the args as default values was redundant because the values defined in the datum are already set to those.

code/modules/xgm/xgm_gas_mixture.dm
PsyCommando: ``update_value()`` is fairly light however, and doesn't take any args. The datum not being in a complete/usable state after construction has been kinda problematic. 

If the gas mixture had some kind of lazy access implemented you could probably bypass having to do this. But as it is, this isn't the most optimized datum we got.

code/modules/materials/_materials.dm
PsyCommando: it's called with a pressure of 0. Which is legal and to be expected. Except ``log(0)`` is undefined and crashes the proc.

code/modules/maps/_map_template.dm
PsyCommando: Thing is, that's not even accurate. A single map file could have more than one z-level in it technically. Disregarding our local standard of mapping.

maps/~mapsystem/maps.dm
PsyCommando: Hmm... That is true. 
Actually I probably should call it planetoid, because technically planetoids could be asteroid and anything else.. -_-

code/unit_tests/~unit_test_subsystems.dm
PsyCommando: Done!

code/unit_tests/~unit_test_subsystems.dm
PsyCommando: Done!

maps/~mapsystem/maps.dm
PsyCommando: Fixed!

code/modules/xgm/xgm_gas_mixture.dm
MistakeNot4892: In what way has it been problematic? Not rhetorical, genuinely asking.

code/modules/xgm/xgm_gas_mixture.dm
MistakeNot4892: This sort of relates to the update_values thing, in the past generating gas mixtures at runtime was usually done by creating an empty one and then using `adjust_gas` or `adjust_gas_multi` - see `initial_air` etc. I'm not sure overriding New() and calling update_values like this is a better approach than removing the New() override and expecting all datums to be customized in place, since that would ensure consistency without overhead on pre-existing uses.

code/modules/materials/_materials.dm
MistakeNot4892: Gotcha. Might as well move the entire calc into the ternary though. Or something like:
````
. = 1 / (1/max(boiling_point, TCMB))
if(pressure > 0)
    . -= ((R_IDEAL_GAS_EQUATION * ((log(pressure / ONE_ATMOSPHERE))) / (latent_heat * molar_mass)))
````
since unless I'm misreading it, all that math is pointless if the initial value is zero (n*0 and n/0 being 0).

File: code/modules/maps/_map_template.dm (hunk 7 above)
Comment (MistakeNot4892) [reply code/modules/maps/_map_template.dm]: `load_single_path` then? Something about `load_a_path` feels untidy to me. Sorry for the nitpick.

**Comment by MistakeNot4892** [comment]
There's a few unanswered comments still.

File: code/modules/maps/_map_template.dm (hunk 7 above)
Comment (PsyCommando) [reply code/modules/maps/_map_template.dm]: Alright, deal, I'll change it to load single path then.

File: code/modules/materials/_materials.dm (hunk 6 above)
Comment (PsyCommando) [reply code/modules/materials/_materials.dm]: Well, the thing is the first equation you got there is wrong. Because, let's take water for example :
```
B = 1 / 1 / 100
B = 1 / 0.01
B = 100
```
Sooo, it would mean the boiling point at one atmosphere is the boiling point at 0 atmosphere, which is wrong?
I wonder if there isn't a problem with the whole equation itself?

Comment (PsyCommando) [reply code/modules/materials/_materials.dm]: @NataKilar Gotta ask the resident physics man.

File: code/modules/xgm/xgm_gas_mixture.dm (hunk 5 above)
Comment (PsyCommando) [reply code/modules/xgm/xgm_gas_mixture.dm]: Well, mainly because you need a structure around the datum that ensures update_value gets called on init, otherwise its data is invalid.
Like it doesn't guarantees it's own state is valid after init. But a proc that gets the datum doesn't know that. And it shouldn't have to know that. This is the kind of gotchas that creates obscure bugs that are hard to track down.

So, the problem we have is you can't just define a gas mix subtype with the gases and everything you want defined, you have to make a separate list to hold the gas values and feed those to the mixture at runtime. 

At the end of the day, if you skip updating values on new, it means you'll need to do several proc calls and update_values calls later on.

**Comment by PsyCommando** [comment]
> There's a few unanswered comments still.

Answered the last comments!

File: code/modules/xgm/xgm_gas_mixture.dm (hunk 5 above)
Comment (PsyCommando) [reply code/modules/xgm/xgm_gas_mixture.dm]: So, I realized that we really didn't need to call update_values() when the gas list is empty. So I added a condition for that. It should basically cut all the extra overhead where it's not needed at all.

File: code/modules/materials/_materials.dm (hunk 6 above)
Comment (NataKilar) [reply code/modules/materials/_materials.dm]: I think there may be an order of operations and units error, but it otherwise seems correct. I would say:
```
if(pressure == 0)
    return melting_point

p_ratio = log(ONE_ATMOSPHERE / pressure)
molar_heat = latent_heat * molar_mass * 1000 // kJ -> J

. = 1/((R_IDEAL_GAS_EQUATION/molar_heat)*p_ratio + 1/boiling_point)

return max(melting_point, .)
```

At some point between atmospheric pressure and 0 pressure the assumptions of the relation are probably violated somehow, but it's a fine approximation. It doesn't make sense to return a boiling point lower than the melting point.

image

PsyCommando: The weird thing is looking at some of the values they get from trying to boil water in vacuum, it doesn't add up:
image
They get -68c as boiling point in almost complete vaccum : https://www.engineeringtoolbox.com/water-evacuation-pressure-temperature-d_1686.html

out-of-phaze: What do you mean that doesn't add up? I'm pretty sure it does.

PsyCommando: I meant, the whole thing about the boiling temp not going lower than the melting temp.

PsyCommando: But, like nata said on discord, another big problem is we don't calculate the melting point based on temperature and pressure, which will cause problems. That probably should get handled in another PR though.

code/modules/xgm/xgm_gas_mixture.dm
PsyCommando: The New() override with those parameters was already there though, I just modified it? 
Also, I added a condition so that, update_values is only called if the gas list has any entries at all. Which should take care of most of any potential extra overhead you're afraid of.

Also, one advantage to consider is that if we can define gas_mixture subtypes with gases, volume, temp set at compilation we save on needing more runtime prep and initial gas lists instances.

MistakeNot4892 (approved):

PR #3090, opened 2023-06-02
Added exterior turf ramps.

MistakeNot4892:
## Description of changes
Allows ramps to be cut with pickaxes using alt-click (this is a placeholder, input on better methods appreciated).
Opening for review mainly, the code is a bit of an atrocity currently.

Allows digging your way out of holes.

## TODO
- [X] Basic implementation.
- [x] Verify that icon blending works and doesn't look awful.
- [x] Verify that directional logic works.
- [x] Verify that digging ramps works.
 
## Authorship
Myself for the terrible code.
Pawn for the amazing ramp icons.

## Changelog
:cl:
add: You can now alt-click an exterior wall using a drill to dig a ramp.
/:cl:


MistakeNot4892:
Currently very busted, will revisit over the weekend.

PsyCommando:
Hmm, why not make a stand-alone slope turf? 
It feels a bit tacked on to implement that into exterior walls directly. And we still use type lookup a lot for telling what kind of turfs you're up against, and expect they always act the same way.
Plus, changeturf handles a lot of the updates you'd probably want triggered when digging a ramp.

Also, it might be better to have a "ramp mode" toggle on the pickaxe via interactions. It would handle having more than one use mode better in the long term imo than having to pick a digging mode each times you'd hit a turf while pressing alt.

It might also be worthwhile to add a proc to tell above turfs that they're being dug up, and they should act accordingly. So you can have digging effects showing up on the upper level, and custom reactions from the upper level turf.
It would probably find some use when building ladders and stairs in-game too, since they both need to often remove a turf on the upper level.

MistakeNot4892:
ChangeTurf is generally to be avoided, particularly on turfs that will see a lot of use in mapgen, like this one. Discriminating based on type is a pretty terrible pattern and in particular with turfs it's one I'm working on killing. It should still behave like a regular exterior turf, just with a different icon and bump behavior.

PsyCommando:
> ChangeTurf is generally to be avoided, particularly on turfs that will see a lot of use in mapgen, like this one. Discriminating based on type is a pretty terrible pattern and in particular with turfs it's one I'm working on killing. It should still behave like a regular exterior turf, just with a different icon and bump behavior.

Discriminating by type is currently ubiquitous when dealing with turfs. Map gen uses it heavily, and a lot of other things also do. I don't think it really makes sense currently to just create one token turf that expects everything to behave differently when dealing with it? Like, there's probably some groundwork to do before switching over to something less reliant on typing? At least that's just my opinion here.

I'm not sure about what you're saying about the mapgen stuff. The big problem currently with mapgen is that it's extremely wasteful. Ideally all random_map should share a common map buffer, and only after all the generators did their job should it be applied to the actual map. That change would seriously overshadow any perf losses due to using changeturf on a few ramp turfs imo.

About having a separate type for the turf. Mappers will end up having to make a separate turf type anyways to have all the specific vars defined in a practical way to be used on a map, and display a relevant icon in the mapping tool. So at the end of the day it's kind of irrelevant if it's implemented in the exterior turf itself. It just means you have now to also maintain the custom exterior turf transformation code stuff at the same time as the usual changeturf stuff. 

I'm just sharing some thoughts because I've been neck deep in turf and map gen code for the last 2 months.

MistakeNot4892:
>Discriminating by type is currently ubiquitous when dealing with turfs.

I am actively working towards limiting or removing this pattern.

Mapping subtypes aren't relevant as they still avoid ChangeTurf. ChangeTurf is one of the most expensive operations in the game and has implications beyond the map generator. 

Additionally the slope var and corresponding sprite used are the only differences between a wall and a slope, there's no compelling reason to separate it out and then factor in separate smoothing, material retrieval, deconstructability, etc. A ramp shouldn't be distinct from its endpoint structurally. It doesn't make sense to be able to build a floor or dig a hole under a ramp and that would also need to be handled.

I don't follow what you mean by custom exterior transform code. This is a draft PR, slopes will be implemented more generally by the time it's mergable. Even if I defined it on a subtype it would still need to be maintained.

MistakeNot4892:
This really is a pretty terrible implementation. If someone does corner and inset states in the future it can likely be made less bad.

At the moment you can dig a ramp by alt-clicking an exterior wall turf using a drill. The ramp will take the direction you are standing as the slope (so digging a ramp while facing west will create an west-upward, east-downward-sloping ramp). Digging a ramp will drop any ore etc. the turf is holding. You can then move up ramps by walking in the direction of the wall.

If the wall is removed (or turned into another ramp), ramps will simply cease to exist as though drilled out.

There's an abstract effect called /obj/abstract/ramp_sculptor that can be used to map ramps or spawn them at runtime. Currently none of the maps etc. use ramps.

out-of-phaze (approved):

PR #3092, opened 2023-06-02
Dev update from stable

MistakeNot4892:


code/modules/paperwork/paper.dm
```diff
@@ -51,6 +51,17 @@
...
+	P.fields             = fields
+	P.last_modified_ckey = last_modified_ckey
+	P.rigged             = rigged
+	P.is_crumpled        = is_crumpled
+	P.stamp_text         = stamp_text
+	P.applied_stamps     = LAZYLEN(applied_stamps)? applied_stamps.Copy() : null
+	P.metadata           = LAZYLEN(metadata)?       metadata.Copy()       : null
+
 /obj/item/paper/GetCloneArgs()
 	return list(null, material, info, name)
```
out-of-phaze: As far as I know, this Clone definition is redundant and folded into the datum-level cloning procs.
```suggestion
/obj/item/paper/GetCloneArgs()
	return list(null, material?.type, info, name)
```

out-of-phaze (changes requested):
Paper conflict wasn't resolved properly :(

PR #3095, opened 2023-06-04
Fix several filesystem and directory issues

out-of-phaze:
## Description of changes
- Fixes inaccurate error reporting for removable drive slots.
- Clarifies error reporting when cloning a file.
- Fixes certain procs (store_file, try_store_file, etc.) not actually passing access to check
- Fixes a missing argument causing cloning to not pass permissions properly
- Fixes a softlock when the media disk is selected while a removable disk is not present.

Some fixes for filesystem stuff, cherrypicked from downstream. Targets dev because there were merge conflicts on both stable and staging that I wasn't confident I could resolve properly.

### DIFF
--- a/code/modules/modular_computers/file_system/directory.dm
+++ b/code/modules/modular_computers/file_system/directory.dm
@@ -15,7 +15,7 @@
 		if(held_file && hard_drive)
 			hard_drive.remove_file(held_file, forced = TRUE)
 
-	QDEL_NULL(temp_file_refs)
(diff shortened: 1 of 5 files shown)


MistakeNot4892 (approved):
Seems fine to me, would appreciate Nata or comma's thoughts.

MistakeNot4892:
Going ahead and merging this since it's been 2 days with no comments.

PR #3105, opened 2023-06-05
Fix map_templates silent failure

PsyCommando:
## Description of changes
Added some CRASHes to map_template loading so it doesn't just silently fail when it cannot load a map_template.
Also fixes map_template loading failures currently happening during unit tests.

### DIFF
--- a/code/modules/maps/_map_template.dm
+++ b/code/modules/maps/_map_template.dm
@@ -151,7 +151,7 @@
 	log_game("Z-level [name] loaded at [x],[y],[world.maxz]")
 	loaded++
 
-	return locate(world.maxx/2, world.maxy/2, world.maxz)
+	return WORLD_CENTER_TURF(world.maxz)
 
 /datum/map_template/proc/load(turf/T, centered=FALSE)
 	if(centered)
--- a/code/modules/maps/_map_template.dm
+++ b/code/modules/maps/_map_template.dm
@@ -155,11 +155,11 @@
 
 /datum/map_template/proc/load(turf/T, centered=FALSE)
 	if(centered)
-		T = locate(T.x - round(width/2) , T.y - round(height/2) , T.z)
+		T = locate(T.x - (round(width/2) - 1), T.y - (round(height/2) - 1), T.z)
 	if(!T)
-		CRASH("Can't load '[src]' on a null turf! Current world size ([world.maxx], [world.maxy]).")
-	if(((T.x + width) > world.maxx) || ((T.y + height) > world.maxy))
-		CRASH("Couldn't fit the entire template '[src]' (size: [width]x[height]) between lower left corner ([T.x], [T.y])[centered?"(WORLD CENTER)":""] and upper right corner ([T.x + width], [T.y + height]) in current world size ([world.maxx], [world.maxy]).")
+		CRASH("Can't load '[src]' (size: [width]x[height]) onto a null turf! Current world size ([WORLD_SIZE_TO_STRING]).")
+	if(!IS_WITHIN_WORLD((T.x + (width - 1)), (T.y + (height - 1))))
+		CRASH("Couldn't fit the entire template '[src]' (size: [width]x[height]) between lower left corner ([T.x], [T.y])[centered?"(WORLD CENTER)":""] and upper right corner ([T.x + width], [T.y + height]) in current world size ([WORLD_SIZE_TO_STRING]).")
 
 	var/list/atoms_to_initialise = list()
 	var/shuttle_state = pre_init_shuttles()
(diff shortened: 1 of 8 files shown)


code/modules/maps/_map_template.dm
```diff
@@ -133,7 +133,9 @@
 			bounds = extend_bounds_if_needed(bounds, M.bounds)
 			atoms_to_initialise += M.atoms_to_initialise
 		else
-			return FALSE
```
MistakeNot4892, line 136: Erroneous removal?

MistakeNot4892: I'm dumb, was reading CRASH as PRINT_STACK_TRACE

MistakeNot4892:
````
 ## UNIT_TEST ##: Loading template 'Heist Base' (/datum/map_template/ruin/antag_spawn/heist) at space (/turf/space) ([0x101e705]) (64,64,8) (/area/space)
[23:00:00] Runtime in _map_template.dm,160: Couldn't fit the entire template 'Heist Base' (size: 70x70) between lower left corner (64, 64) and upper right corner (134, 134) in current world size (129, 129).
  proc name: load (/datum/map_template/proc/load)
  src: Heist Base (/datum/map_template/ruin/antag_spawn/heist)
  call stack:
  Heist Base (/datum/map_template/ruin/antag_spawn/heist): load(space (64,64,8) (/turf/space), 0)
  Unit Tests (/datum/controller/subsystem/unit_tests): load template(Heist Base (/datum/map_template/ruin/antag_spawn/heist))
  Unit Tests (/datum/controller/subsystem/unit_tests): load map templates()
  Unit Tests (/datum/controller/subsystem/unit_tests): Initialize(94709)
  Unit Tests (/datum/controller/subsystem/unit_tests): DoInitialize(94709)
SSzcopy: Z-Level maximums: [2,2,3,4,5,6,7,8,9]
## UNIT_TEST ##: Loading template 'Mercenary Base' (/datum/map_template/ruin/antag_spawn/mercenary) at space (/turf/space) ([0x1022806]) (64,64,9) (/area/space)
````
````
[23:00:00] Runtime in _map_template.dm,160: Couldn't fit the entire template 'Mercenary Base' (size: 100x100) between lower left corner (64, 64) and upper right corner (164, 164) in current world size (129, 129).
  proc name: load (/datum/map_template/proc/load)
  src: Mercenary Base (/datum/map_template/ruin/antag_spawn/mercenary)
  call stack:
  Mercenary Base (/datum/map_template/ruin/antag_spawn/mercenary): load(space (64,64,9) (/turf/space), 0)
  Unit Tests (/datum/controller/subsystem/unit_tests): load template(Mercenary Base (/datum/map_template/ruin/antag_spawn/mercenary))
  Unit Tests (/datum/controller/subsystem/unit_tests): load map templates()
  Unit Tests (/datum/controller/subsystem/unit_tests): Initialize(94709)
  Unit Tests (/datum/controller/subsystem/unit_tests): DoInitialize(94709)
  ````

File: code/modules/maps/_map_template.dm [hunk 2]
```diff
@@ -154,12 +156,8 @@
 /datum/map_template/proc/load(turf/T, centered=FALSE)
 	if(centered)
 		T = locate(T.x - round(width/2) , T.y - round(height/2) , T.z)
-	if(!T)
-		return
-	if(T.x+width > world.maxx)
-		return
-	if(T.y+height > world.maxy)
-		return
+	if(!T || ((T.x + width) > world.maxx) || ((T.y + height) > world.maxy))
```
MistakeNot4892: Null `T` is causing the CRASH messages to fail on a null reference.

PsyCommando: Yeah, but that's legit. It would silently return on null T anyways before.

PsyCommando: I see what you mean now. The crash string prints out the turf's coordinates without nullchecking them..

PsyCommando:
So, more problems. 
The code for the ``centered == TRUE`` case seems to just not have been thought out to deal with templates fitting exactly within the world's bounds. The 100x100 template won't fit in the map because it adds the template width to the coordate of the turf, 1,1. That means it thinks its top right corner is at 101,101 instead of being at 100,100. 
Going to take a look later tomorrow.

MistakeNot4892:
Might just be a case of flooring rather than ceiling when finding the lower left corner?

PsyCommando:
> Might just be a case of flooring rather than ceiling when finding the lower left corner?

Nah, pretty sure it's just because the checks and code to grab the world center don't subtract 1 from the width/height.

PsyCommando:
Hmm.. For some reasons my commits aren't showing up and it's not running unit tests...

PsyCommando:
Oh seems to have worked now.

MistakeNot4892 (approved):

PR #3106, opened 2023-06-08
Gasmix consistency tweaks

MistakeNot4892:
## Description of changes
Should hopefully guarantee that return_air always returns a gas datum, to try and get ZAS to cooperate.
I'm not actually sure if this should be merged or not.

ZAS seems to expect to never have to deal with null gas datums.

## Authorship
Myself.

## Changelog
Nothing player-facing.

### DIFF
--- a/code/controllers/communications.dm
+++ b/code/controllers/communications.dm
@@ -234,7 +234,7 @@ var/global/datum/controller/radio/radio_controller
 	var/encryption
 	var/frequency = 0
 
-/datum/signal/proc/copy_from(datum/signal/model)
+/datum/signal/proc/copy_from_model(datum/signal/model)
 	source = model.source
 	data = model.data
 	encryption = model.encryption
--- a/code/game/atoms.dm
+++ b/code/game/atoms.dm
@@ -88,7 +88,7 @@
 */
 /atom/proc/return_air()
 	RETURN_TYPE(/datum/gas_mixture)
-	return loc?.return_air()
+	return loc?.return_air() || new /datum/gas_mixture
 
 /**
 	Get the flags that should be added to the `users` sight var.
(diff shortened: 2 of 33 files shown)


PsyCommando:
I don't know if we really want to do this? It means a ton of empty gas_mixture datums are going to get created and discarded immediately. 
And some pretty much established assumptions on return_air and gas_mixture usage in general are going to be violated and probably cause problems. Also will cause issues in forks.

MistakeNot4892:
`return_air` is not super hot at least. Datum overhead shouldn't be significant and they don't keep references to anything so deletion should be quick.

I really don't see how this will cause issues for forks.

out-of-phaze:
> ZAS seems to expect to never have to deal with null gas datums.

This is correct, but I believe the only restriction is that *turf* `return_air()` never returns null. Making, for example, objects in nullspace return an empty gasmix is unnecessary because they will never be a zone component. As far as I know the only reason turf return_air can return null is this:

```dm
		var/datum/gas_mixture/gas = level?.get_exterior_atmosphere()
		if(!gas)
			return
```

Making *that* return an empty gasmix would probably be fine—*or* you could make `get_exterior_atmosphere()` return an empty gasmix if `exterior_atmosphere` is unset, making the nullcheck unnecessary. Level should never, ever be null:
```dm
	for(var/z = 1 to world.maxz)
		var/datum/level_data/level = levels_by_z[z]
		if(!istype(level))
			level = new /datum/level_data/space(z)
			PRINT_STACK_TRACE("Missing z-level data object for z[num2text(z)]!")
		level.setup_level_data()
```
As far as I can tell any other changes in this PR are irrelevant to the issue it tries to solve.

PR #3108, opened 2023-06-09
Fix infinite recursion in get_all_connected_level_z

out-of-phaze:
## Description of changes
Fixes infinite recursion in `get_all_connected_level_z` due to missing the deduplication argument. This fixes issues with level_data not being initialized properly leading to null exterior atmosphere.

Alternative to #3106. The runtime that PR is attempting to solve only happened because of this runtime leading to level_data setup procs not getting called; see the FIXME comment about exterior_atmosphere somehow being null.


comma (approved):

MistakeNot4892:
Nice.

PR #3112, opened 2023-06-11
Converting dexterity to a bitflag system.

MistakeNot4892:
## Description of changes
- Cleans up some dexterity checking infrastructure.
- Switches dexterity to use bitflags instead of a linear value comparison.
- Adds a check to shower curtains so mice can't close or open them.
 
Makes it much easier to handle mob inventory and equipment without having to juggle linear dexterity values.

## Authorship
Myself.

## Changelog

### DIFF
--- a/code/modules/mob/living/carbon/human/human.dm
+++ b/code/modules/mob/living/carbon/human/human.dm
@@ -1197,31 +1197,26 @@
 
 				new /obj/effect/temp_visual/bloodsplatter(loc, hit_dir, species.get_blood_color(src))
 
-/mob/living/carbon/human/has_dexterity(var/dex_level)
-	. = check_dexterity(dex_level, silent = TRUE)
-
-/mob/living/carbon/human/check_dexterity(var/dex_level = DEXTERITY_FULL, var/silent, var/force_active_hand)
-	if(isnull(force_active_hand))
-		force_active_hand = get_active_held_item_slot()
-	var/obj/item/organ/external/active_hand = GET_EXTERNAL_ORGAN(src, force_active_hand)
-	var/dex_malus = 0
-	if(getBrainLoss() && getBrainLoss() > config.dex_malus_brainloss_threshold) ///brainloss shouldn't instantly cripple you, so the effects only start once past the threshold and escalate from there.
-		dex_malus = round(clamp(round(getBrainLoss()-config.dex_malus_brainloss_threshold)/10, DEXTERITY_NONE, DEXTERITY_FULL))
+/mob/living/carbon/human/get_dexterity(var/silent = FALSE)
+	var/check_slot = get_active_held_item_slot()
+	var/obj/item/organ/external/active_hand = check_slot && GET_EXTERNAL_ORGAN(src, check_slot)
 	if(!active_hand)
 		if(!silent)
-			to_chat(src, SPAN_WARNING("Your hand is missing!"))
-		return FALSE
+			to_chat(src, SPAN_WARNING("Your [check_slot ? parse_zone(check_slot) : "hand"] is missing!"))
+		return DEXTERITY_NONE
 	if(!active_hand.is_usable())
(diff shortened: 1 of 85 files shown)


code/modules/mob/mob.dm
```diff
@@ -1062,11 +1062,14 @@
...
-/mob/proc/check_dexterity(var/dex_level, var/silent)
-	. = has_dexterity(dex_level)
+/mob/proc/get_dexterity(var/silent = FALSE)
+	var/decl/species/my_species = get_species()
+	if(my_species)
+		return my_species.get_manual_dexterity()
+	return DEXTERITY_BASE
+
+/mob/proc/check_dexterity(var/dex_level = DEXTERITY_FULL, var/silent = FALSE)
+	. = !!(get_dexterity(silent) & dex_level)
```
out-of-phaze: Are we sure this doesn't check if it has *any* of the bitflags rather than all? 10 & 11 -> 10, !!10 -> TRUE. Should probably be `(get_dexterity(silent) & dex_level) == dex_level`?

code/modules/mob/living/carbon/human/human.dm
out-of-phaze, line 1: Non-blocking but I would like it if some of the fixes in #3187 made it in here, particularly checking `check_slot.requires_organ_tag` in human `get_dexterity()`.

mods/mobs/dionaea/mob/_nymph.dm
```diff
@@ -85,6 +85,3 @@
 	var/datum/extension/hattable/hattable = get_extension(src, /datum/extension/hattable)
 	if(hattable?.hat)
 		to_chat(user, SPAN_NOTICE("It is wearing [html_icon(hattable.hat)] \a [hattable.hat]."))
-
-/mob/living/carbon/alien/diona/has_dexterity()
```
out-of-phaze, line 89: Without this, won't it still have `DEXTERITY_BASE`, since `carbon/alien` doesn't support species?

mods/species/ascent/mobs/nymph/_nymph.dm
```diff
@@ -58,9 +58,6 @@
 	if(holding_item)
 		to_chat(user, SPAN_NOTICE("It is holding \icon[holding_item] \a [holding_item]."))
 
-/mob/living/carbon/alien/ascent_nymph/has_dexterity()
```
out-of-phaze, line 61: See above comment about the same removal on diona nymphs.

out-of-phaze (changes requested):

code/modules/mob/mob.dm
MistakeNot4892: Valid, I didn't consider compound dex checks.

out-of-phaze: The default check appears to be `DEXTERITY_FULL`, is that supposed to be "has any dexterity" rather than "has every dexterity"?

MistakeNot4892:
This is ready for review.

code/modules/mob/mob.dm
MistakeNot4892: it's meant to be 'has every needed dex level' but the reversal from threshold to flag confused me.

code/modules/mob/mob.dm
```diff
@@ -1063,11 +1063,14 @@
...
-/mob/proc/check_dexterity(var/dex_level, var/silent)
-	. = has_dexterity(dex_level)
+/mob/proc/get_dexterity(var/silent = FALSE)
+	var/decl/species/my_species = get_species()
+	if(my_species)
+		return my_species.get_manual_dexterity()
+	return DEXTERITY_BASE
+
+/mob/proc/check_dexterity(var/dex_level = DEXTERITY_FULL, var/silent = FALSE)
+	. = (get_dexterity(silent) & dex_level) == dex_level
```
out-of-phaze, line 1073: Might be worth adding a comment here down the line, since this is a relatively rare check.

out-of-phaze (approved):

PR #3119, opened 2023-06-16
Generalizing stasis and life_tick to /mob/living.

MistakeNot4892:
## Description of changes
- Moves stasis procs and vars to /mob/living and renames them.
- Moves life_tick incrementing and checking to /mob/living.

Working towards killing /carbon. Also allows for the condensing/collapsing of Life() down the track.

## Authorship
Myself.

## Changelog
Nothing player-facing.

### DIFF
--- a/code/game/machinery/cryopod.dm
+++ b/code/game/machinery/cryopod.dm
@@ -154,7 +154,7 @@
 	var/allow_occupant_types = list(/mob/living/carbon/human)
 	var/disallow_occupant_types = list()
 
-	var/mob/occupant = null       // Person waiting to be despawned.
+	var/mob/living/occupant       // Person waiting to be despawned.
 	var/time_till_despawn = 9000  // Down to 15 minutes //30 minutes-ish is too long
 	var/time_entered = 0          // Used to keep track of the safe period.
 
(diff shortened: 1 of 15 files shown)


code/game/machinery/cryopod.dm
```diff
@@ -291,9 +291,9 @@
 //Lifted from Unity stasis.dm and refactored.
 /obj/machinery/cryopod/Process()
 	if(occupant)
-		if(applies_stasis && iscarbon(occupant) && (world.time > time_entered + 20 SECONDS))
-			var/mob/living/carbon/C = occupant
-			C.SetStasis(2)
+		if(applies_stasis && isliving(occupant) && (world.time > time_entered + 20 SECONDS))
```
out-of-phaze: Should we just repath `occupant` to `var/mob/living/occupant` instead of `var/mob/occupant`? It doesn't seem like it'd ever be sane for an observer, new_player, or virtual mob to be a cryopod occupant.

out-of-phaze (dismissed):

code/game/machinery/cryopod.dm
```diff
@@ -291,9 +291,9 @@
 //Lifted from Unity stasis.dm and refactored.
 /obj/machinery/cryopod/Process()
 	if(occupant)
-		if(applies_stasis && iscarbon(occupant) && (world.time > time_entered + 20 SECONDS))
-			var/mob/living/carbon/C = occupant
-			C.SetStasis(2)
+		if(applies_stasis && (world.time > time_entered + 20 SECONDS))
+			var/mob/living/patient = occupant
```
comma: do you need to cast here if occupant is already /living? Maybe a typecheck if nothing prevents non living form entering

MistakeNot4892: Oh yeah this is from before occupant got changed to living.

out-of-phaze: > do you need to cast here if occupant is already /living? Maybe a typecheck if nothing prevents non living form entering

That might be more appropriate in set_occupant, and then we can just always assume occupant is either living or null.

MistakeNot4892: A nonliving mob can't really interact with it anyway.

out-of-phaze (approved):

PR #3120, opened 2023-06-16
Moving temperature thresholds into a helper proc.

MistakeNot4892:
## Description of changes
Moves temperature thresholds onto a proc at /mob/living instead of doing it all in place with species member vars.

Easier to get temperature thresholds, cleaner code, support for eventually killing /carbon.

## Authorship
Myself.

## Changelog
Nothing player-facing.

### DIFF
--- a/code/game/gamemodes/cult/ghosts.dm
+++ b/code/game/gamemodes/cult/ghosts.dm
@@ -246,8 +246,9 @@
 		return
 
 	to_chat(choice, "<span class='danger'>You feel as if something cold passed through you!</span>")
-	if(choice.bodytemperature >= choice.species.cold_level_1 + 1)
-		choice.bodytemperature = max(choice.species.cold_level_1 + 1, choice.bodytemperature - 30)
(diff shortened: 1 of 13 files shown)


PsyCommando:
Just mentioning, but #3089 relies on species temps to setup a habitable atmos, so there's gonna be conflicts. Idk if you want to handle them? Because, that could keep 3089 in the PR list for a while before I get around to fixing the conflicts and running manual tests again.

MistakeNot4892:
The conflicts should just be swapping a var for a proc call so I don't mind which end they get resolved on.

comma (dismissed):

out-of-phaze (approved):
Not looking forward to pulling this into the bodytypes branch but at least it seems mostly straightforward. I figure in the future we'd remove the isSynthetic section and check bodytype instead of species and that's it?

Anyway, looks like it handles all of the issues I thought of in review—mostly I double-checked that the speciesless defaults in living/stasis.dm were still handled properly, and they should be. Should also make temp threshold stuff far more consistent and allow it to be modified per-mob by other stuff down the line.

PR #3123, opened 2023-06-16
Sign installation + data handling tweaks

PsyCommando:
## Description of changes
Sign and diploma stuff:
* Repathed some signs so their type path matches the base icons and features they share better.
* Signs item now holds the original sign structure type instead of creating a look-alike sign when placed.
* Separated sign icons into their own icon files to make diffs easier in the future.
* Improved how laminated diplomas work, so now they keep their diploma data even after being uninstalled from the wall.
* Simplified the process of placing a sign on a wall. So it works like everything else. No more prompt asking for a direction, it places it depending on what wall you're clicking on.
* Placing and removing a sign now also carries over labels, forensics evidences, and scent.
* The sign item type that's dropped on sign structure uninstall is now the ``part_type`` on the structure. This makes better use of existing code to do the same thing. It also allow signs that don't work like the others (diplomas) to properly move over their data in their item form with the proper item subtype.

Pen stuff:
* It's a lot easier to keep the fonts used for each types of pen when they're stored as tool properties, so did that. The font types also now are defines. (Used by diploma signing, so they have the correct font depending on the pen used to sign it.)

I've had to make some changes to extensions copying so they can be carried over to and from the sign item and structure. (Labels and stuff)
* Since extensions cannot be cloned using the Clone() proc, I've made a work around where extensions have a proc called ``copy_from`` which copies data to an existing extension, solving the problem of extensions always needing to have a holder on creation, which prevent us from using Clone().
* The copy_from proc uses PopulateClone to actually populate the copied data, so this can be easily refactored in the future to use the base Clone system if we want.
* Added a copy_extension proc to copy an extension of the given base type over to another atom. Useful for things like labels and scent and etc..

## Changelog
:cl:
tweak:  Signs are easier to place now, and do a better job keeping their info when installed and removed repeatedly.
/:cl:


code/modules/tools/tool_archetype_definition_pen.dm
```diff
@@ -9,6 +9,7 @@
 		TOOL_PROP_USES            = -1, \
 		TOOL_PROP_PEN_SIG         = null,\
 		TOOL_PROP_PEN_SHADE_COLOR = "black",\
+		TOOL_PROP_PEN_FONT        = PEN_FONT_DEFAULT,\
```
MistakeNot4892: The escapes in this list are unnecessary.

code/datums/extensions/extensions.dm
```diff
@@ -15,6 +15,24 @@
 	holder = null
 	. = ..()
 
+///Extensions can't be cloned from a Clone() call, because they need a holder on New, and cannot be cloned onto the same object.
+/// Use copy_from instead!
+/datum/extension/CanClone()
```
MistakeNot4892, line 20: Might be worth overriding Clone() and print a stack trace or a warning so people are aware that extensions have a separate proc.

MistakeNot4892:
Not expecting you to fix this, but noting a lot of the signs in the maps have state and string overrides that will be lost when they are deconstructed into their item form.

PsyCommando:
> Not expecting you to fix this, but noting a lot of the signs in the maps have state and string overrides that will be lost when they are deconstructed into their item form.

That's true for a few things when spawning a sign item with a sign type set. Probably could be fixed by using the atom repository I guess?

code/modules/tools/tool_archetype_definition_pen.dm
MistakeNot4892: Tidying these before merge would be appreciated.

PsyCommando: Alright, removed them.

MistakeNot4892 (approved):

PR #3129, opened 2023-06-16
Use get_contained_external_atoms instead of bespoke checks

out-of-phaze:
## Description of changes
Replaces a bunch of `(contents - foo)` checks with `get_contained_external_atoms()` with possible, including outright replacing some loops with `dump_contents()`.

Better code quality! Also necessary for a future PR.

### DIFF
--- a/code/game/machinery/cryopod.dm
+++ b/code/game/machinery/cryopod.dm
@@ -461,7 +461,7 @@
 	icon_state = base_icon_state
 
 	//Eject any items that aren't meant to be in the pod.
-	var/list/items = contents - component_parts
+	var/list/items = get_contained_external_atoms()
 	if(occupant) items -= occupant
 
 	for(var/obj/item/W in items)
--- a/code/game/machinery/kitchen/gibber.dm
+++ b/code/game/machinery/kitchen/gibber.dm
@@ -143,7 +143,7 @@
 /obj/machinery/gibber/proc/go_out()
 	if(operating || !src.occupant)
 		return
-	for(var/obj/O in (contents - component_parts))
+	for(var/obj/O in get_contained_external_atoms())
 		O.dropInto(loc)
 	if (src.occupant.client)
 		src.occupant.client.eye = src.occupant.client.mob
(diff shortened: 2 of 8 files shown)


code/game/machinery/cryopod.dm
```diff
@@ -461,7 +461,7 @@
 	icon_state = base_icon_state
 
 	//Eject any items that aren't meant to be in the pod.
-	var/list/items = contents - component_parts
+	var/list/items = get_contained_external_atoms()
 	if(occupant) items -= occupant
```
MistakeNot4892, line 465: Should the occupant removal be part of get_contained_external_atoms()?

code/game/machinery/kitchen/gibber.dm
```diff
@@ -143,7 +143,7 @@
 /obj/machinery/gibber/proc/go_out()
 	if(operating || !src.occupant)
 		return
-	for(var/obj/O in (contents - component_parts))
+	for(var/obj/O in get_contained_external_atoms())
 		O.dropInto(loc)
```
MistakeNot4892, line 147: Seems like this could just be dump_contents().

code/game/machinery/kitchen/gibber.dm
```diff
@@ -211,7 +211,7 @@
 	qdel(occupant)
 
 	playsound(loc, 'sound/effects/splat.ogg', 50, 1)
-	for (var/obj/thing in (contents - component_parts))
+	for (var/obj/thing in get_contained_external_atoms())
```
MistakeNot4892, line 214: `/atom/movable` and `as anything` would be valid here.

code/modules/recycling/disposal.dm
```diff
@@ -295,7 +295,7 @@ var/global/list/diversion_junctions = list()
 
 // eject the contents of the disposal unit
 /obj/machinery/disposal/proc/eject()
-	for(var/atom/movable/AM in (contents - component_parts))
+	for(var/atom/movable/AM in get_contained_external_atoms())
```
MistakeNot4892, line 298: `as anything` would be valid here.

code/game/machinery/kitchen/gibber.dm
out-of-phaze: Nope. Objects only.

code/game/machinery/kitchen/gibber.dm
out-of-phaze: No, because mobs can be inside the gibber...? See `occupant`. That's the whole point of the /obj check here and above.

MistakeNot4892: The occupant is qdeleted immediately prior to this line, there should be no other mobs in there. Held mobs would have dropped as /obj/item/holder.

MistakeNot4892 (approved):
Questions aren't blockers.

PR #3138, opened 2023-06-16
Test random stratum validity against level_data

out-of-phaze:
## Description of changes
Checks random strata against level_data atmosphere.

No more permafrost on non-frozen levels.

### DIFF
--- a/code/modules/multiz/level_data.dm
+++ b/code/modules/multiz/level_data.dm
@@ -252,7 +252,15 @@
 /datum/level_data/proc/setup_strata()
 	//If no strata, pick a random one
 	if(isnull(strata))
-		strata = pick(decls_repository.get_decl_paths_of_subtype(/decl/strata))
+		var/list/all_strata      = decls_repository.get_decls_of_subtype(/decl/strata)
+		var/list/possible_strata = list()
+
+		for(var/stype in all_strata)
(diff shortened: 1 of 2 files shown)


code/modules/multiz/level_data.dm
```diff
@@ -252,7 +252,15 @@
...
-		strata = pick(decls_repository.get_decl_paths_of_subtype(/decl/strata))
+		var/list/all_strata      = decls_repository.get_decls_of_subtype(/decl/strata)
+		var/list/possible_strata = list()
+
+		for(var/stype in all_strata)
+			var/decl/strata/strata = all_strata[stype]
+			if(strata.is_valid_level_stratum(src))
+				possible_strata += stype
+
+		strata = DEFAULTPICK(possible_strata, GET_DECL(/decl/strata/sedimentary))
```
MistakeNot4892, line 263: This won't work, `strata` is meant to be a type at this point.

out-of-phaze: Iiiiiiiiiiiiii think it works just fine. `ispath(strata)` is checked below. Inefficient? Sure, I'll fix it, but like, it works.

MistakeNot4892 (approved):
Comment isn't a blocker.

PR #3140, opened 2023-06-16
Port darker dirt sprite from Polaris

out-of-phaze:
## Description of changes
Makes dirt darker using a sprite I lifted from Virgo, which was *probably* from Polaris?

image
Left is old, right is new. A lot more pleasant, IMO, and it reads more as plain dirt rather than like, red clay dirt.

## Authorship
Polaris? I think. IDK any more specific details.

## Changelog
:cl:
imageadd: Dirt turfs look darker now, courtesy of some Polaris sprites.
/:cl:


MistakeNot4892 (approved):

MistakeNot4892:
Similar to the other, no reason to freeze this.

PR #3143, opened 2023-06-16
Change autojoin to try connecting before queueing

out-of-phaze:
## Description of changes
Replaces `queue_connection()` with `try_connect()` in the network device extension's auto-join feature.

This required moving it to post_construction so that `get_extension(src, /datum/extension/network_device)` in fabricators' post-network-connection would work properly.

When a modular computer booted for the first time, autorun programs that require the network would fail to boot if it was within 2 seconds of initializing the computer (this probably also happened if you replaced the network card with a newly made one?) because the connection queue hadn't resolved yet. Now, it just tries to connect directly before going into the queue.

### DIFF
--- a/code/modules/modular_computers/networking/device_types/_network_device.dm
+++ b/code/modules/modular_computers/networking/device_types/_network_device.dm
@@ -22,6 +22,9 @@
 	var/list/device_variables
 	var/list/device_methods
 
+	/// Tracking var for autojoin, to resolve an ordering issue in device creation/connection.
+	VAR_PRIVATE/_autojoin
+
 /datum/extension/network_device/New(datum/holder, n_id, n_key, r_type, autojoin = TRUE)
 	..()
 	network_id = n_id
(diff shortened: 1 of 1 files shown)


code/modules/modular_computers/networking/device_types/_network_device.dm
```diff
@@ -44,6 +46,12 @@
 	if(has_commands)
 		reload_commands()
 
+// Must be done here so that our holder's get_extension calls work.
+/datum/extension/network_device/post_construction()
+	. = ..()
+	if(_autojoin)
+		SSnetworking.try_connect(src)
```
MistakeNot4892, line 53: Should this be queuing if autojoin is unset?

out-of-phaze: No. It shouldn't try to join automatically at all if autojoin is unset, whether it's via the queue or via a direct connection attempt.

MistakeNot4892 (approved):

PR #3144, opened 2023-06-16
Adjust pool turfs to more sensible depths

out-of-phaze:
## Description of changes
Lowers the depth of pool turfs to be around half of what it was before, adds a deeper variant to compensate.

Pools were kind of hilariously deep, you'd instantly drown when entering them. This changes them to sane depths; I might even add a shallow variant at some point in the future.


MistakeNot4892 (approved):

MistakeNot4892:
No reason to freeze this, it's tiny. We're still waiting on the planet PR anyway.

PR #3146, opened 2023-06-16
Add ZM_PARTITION_STACK for weather/outsideness/sky

out-of-phaze:
## Description of changes
Adds a turf z_flag that terminates outsideness z-stack traversal when the turf is reached.

For cases where you have a surface level, a sky level, and a space level all connected in a z-stack, surface weather should only consider being blocked up to the sky level; something in space is not going to stop the weather.

### DIFF
--- a/code/game/turfs/exterior/exterior_sky.dm
+++ b/code/game/turfs/exterior/exterior_sky.dm
@@ -2,7 +2,11 @@
 	name = "sky"
 	desc = "Hope you don't have a fear of heights..."
 	icon = 'icons/turf/exterior/sky_static.dmi'
-	z_flags = 0
+	z_flags = ZM_PARTITION_STACK
+
+// No matter what, the sky will never be 'inside'.
+/turf/exterior/open/sky/is_outside()
+	return TRUE // you can't take the sky from meee
 
 /turf/exterior/open/sky/north
 	dir = NORTH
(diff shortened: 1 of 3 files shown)


code/game/turfs/exterior/exterior_sky.dm
```diff
@@ -2,7 +2,11 @@
 	name = "sky"
 	desc = "Hope you don't have a fear of heights..."
 	icon = 'icons/turf/exterior/sky_static.dmi'
-	z_flags = 0
+	z_flags = ZM_PARTITION_STACK
+
+// No matter what, the sky will never be 'inside'.
+/turf/exterior/open/sky/is_outside()
```
MistakeNot4892, line 8: Conceptually I'm not sure I agree with this. I'm imagining an airship or something with a base level of sky turfs and a ceiling built above it.

out-of-phaze: It still wouldn't hold an atmosphere then, would it? I suppose we wouldn't want it to rain with a ceiling though.

MistakeNot4892: It would hold atmosphere in the sense of participating in ZAS. It'd leak out the sides/downwards.

MistakeNot4892:
Per Discord, going to close this for now as it's more appropriate downstream.

PR #3150, opened 2023-06-17
Converting last_special and simple_animal attack cooldowns to use helper procs.

MistakeNot4892:
## Description of changes
- `last_special` is now handled with helper procs.
- Simple animal special attacks now use helper procs.

Leaving this WIP until I can test, probably tomorrow.

Unified code, less naked var use. Supports HUD rewrite.

## Authorship
Myself.

## Changelog
Nothing player-facing.

### DIFF
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -1105,3 +1105,11 @@ default behaviour is:
 /mob/living/get_speech_bubble_state_modifier()
 	return isSynthetic() ? "synth" : ..()
 
+/mob/living/proc/is_on_special_ability_cooldown()
+	return world.time < next_special_ability
+
+/mob/living/proc/set_special_ability_cooldown(var/amt)
+	next_special_ability = max(next_special_ability, world.time+amt)
+
+/mob/living/proc/get_seconds_until_next_special_ability_string()
+	return ticks2readable(next_special_ability - world.time)
--- a/code/game/gamemodes/godmode/god_altar.dm
+++ b/code/game/gamemodes/godmode/god_altar.dm
@@ -82,10 +82,10 @@
 /obj/structure/deity/altar/OnTopic(var/user, var/list/href_list)
 	if(href_list["resist"])
 		var/mob/living/M = locate(href_list["resist"])
-		if(!istype(M) || target != M || M.stat || M.last_special > world.time)
+		if(!istype(M) || target != M || M.stat || M.is_on_special_ability_cooldown())
 			return TOPIC_HANDLED
 
-		M.last_special = world.time + 10 SECONDS
+		M.set_special_ability_cooldown(10 SECONDS)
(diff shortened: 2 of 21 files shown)


MistakeNot4892:
Tested humans, antlions and spiders. All seems working.

code/modules/mob/living/living.dm
```diff
@@ -1105,3 +1105,11 @@ default behaviour is:
 /mob/living/get_speech_bubble_state_modifier()
 	return isSynthetic() ? "synth" : ..()
 
+/mob/living/is_on_special_ability_cooldown()
+	return world.time < next_special_ability
+
+/mob/living/set_special_ability_cooldown(var/amt)
+	next_special_ability = max(next_special_ability, world.time+amt)
+
+/mob/living/get_time_until_next_special_ability()
```
out-of-phaze:
```suggestion
/mob/living/get_seconds_until_next_special_ability()
```
`get_time` is a bit unclear since as far as I can tell this is returning seconds, not deciseconds? Wouldn't want someone to use it expecting deciseconds. Non-blocking.

MistakeNot4892: I might rename it seconds and move the `second\s` part in here. That would let it potentially handle minutes etc. in the future.

code/modules/mob/living/living.dm
```diff
@@ -1105,3 +1105,11 @@ default behaviour is:
...
 	return isSynthetic() ? "synth" : ..()
 
+/mob/living/is_on_special_ability_cooldown()
+	return world.time < next_special_ability
+
+/mob/living/set_special_ability_cooldown(var/amt)
+	next_special_ability = max(next_special_ability, world.time+amt)
+
+/mob/living/proc/get_seconds_until_next_special_ability_string()
+	return ticks2readable((next_special_ability - world.time)*0.1)
```
out-of-phaze:
```suggestion
	return ticks2readable(next_special_ability - world.time)
```
I believe ticks2readable takes deciseconds? It uses the `HOUR`, `MINUTE`, and `SECOND` macros at least.

code/modules/mob/mob.dm
```diff
@@ -1320,3 +1320,12 @@
...
 /mob/proc/get_target_zone()
 	return zone_sel?.selecting
+
+/mob/proc/is_on_special_ability_cooldown()
+	return FALSE
+
+/mob/proc/set_special_ability_cooldown(var/amt)
+	return
+
+/mob/proc/get_time_until_next_special_ability()
```
out-of-phaze: I think this is a holdover from before you renamed it to `get_seconds_until_next_special_ability_string`. Since that's defined on `/mob/living` this override can probably be deleted?

out-of-phaze (changes requested):

out-of-phaze (approved):

PR #3157, opened 2023-06-22
Adding guards to simple_animal attackby calls.

MistakeNot4892:
Prevents some runtimes observed in testing.


out-of-phaze (approved):
~~I'll choose not to be silly and ask you to change that `//Time for the hurt to begin!` comment to `//Maybe time for the hurt to possibly begin!`.~~

PR #3158, opened 2023-06-25
Standardizes stack merge proc argument, fixes runtime.

MistakeNot4892:


code/game/objects/items/stacks/stack.dm
```diff
@@ -355,16 +355,14 @@
...
-		return
 	var/list/stacks = list()
 	if(check_hands && user)
 		for(var/obj/item/stack/item in user.get_held_items())
 			stacks |= item
 	for (var/obj/item/stack/item in user?.loc)
 		stacks |= item
 	for (var/obj/item/stack/item in stacks)
-		if (item==src)
+		if(item == src || !(can_merge_stacks(item) || item.can_merge_stacks(src)))
```
MistakeNot4892, line 365: This pattern is a bit ??? but the other merge check uses it and I figured I'd be consistent.

out-of-phaze (approved):

code/modules/power/cable.dm
```diff
@@ -621,16 +621,16 @@ By design, d1 is the smallest direction and d2 is the highest
 // Items usable on a cable coil :
 //   - Wirecutters : cut them duh !
 //   - Cable coil : merge cables
-/obj/item/stack/cable_coil/can_merge(var/obj/item/stack/cable_coil/C)
-	return color == C.color
+/obj/item/stack/cable_coil/can_merge_stacks(var/obj/item/stack/other)
+	return !other || (istype(other) && other.color == color)
```
afterthought2, line 625: I don't quite get `!other`; is it correct/important that we can merge with `null`?

MistakeNot4892: Probably not, it was an initial fix for can_merge() being called with no argument causing a runtime here.

PR #3159, opened 2023-06-25
Apply stopgap fix for load_new_z

out-of-phaze:
## Description of changes
As discussed on Discord, lighting overlays were not being created on turfs loaded in `load_new_z` because `extend_bounds_if_needed` was replacing the bounds list instead of mutating it, meaning load_new_z never received the updated bounds and thus never initialized the turfs.

Lighting works again on maps loaded onto a new z-level.


afterthought2 (approved):

MistakeNot4892:
Per Discord, DNM pending the more complete fix.

out-of-phaze:
I've decided this one is good enough—the rework can go to dev once I've figured out the issues with it.

PR #3160, opened 2023-06-26
Simplify new z maploading and fix lighting sanity checks

out-of-phaze:
## Description of changes
Restores some missing sanity checks in lighting init code.
Removes code that was violating those sanity checks. As far as I can tell, calling `SSlighting.InitZLev` like that is invalid and leads to the sanity check failing.
`load_new_z` now just calls `load` directly.

Supersedes #3159, fixes the same issue, improves code quality.

### DIFF
--- a/code/controllers/subsystems/lighting.dm
+++ b/code/controllers/subsystems/lighting.dm
@@ -79,9 +79,12 @@ SUBSYSTEM_DEF(lighting)
 /datum/controller/subsystem/lighting/proc/InitializeZlev(zlev)
 	for (var/thing in Z_ALL_TURFS(zlev))
 		var/turf/T = thing
-		if (TURF_IS_DYNAMICALLY_LIT_UNSAFE(T) && !T.lighting_overlay)	// Can't assume that one hasn't already been created on bay/neb.
-			new /atom/movable/lighting_overlay(T)
-			. += 1
+		if (TURF_IS_DYNAMICALLY_LIT_UNSAFE(T))
+			if (T.lighting_overlay)
+				log_ss(name, "Found unexpected lighting overlay at [T.x],[T.y],[T.z]")
+			else
+				new /atom/movable/lighting_overlay(T)
+				. += 1
 			if (T.ambient_light)
 				T.generate_missing_corners()	// Forcibly generate corners.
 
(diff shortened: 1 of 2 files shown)


code/controllers/subsystems/lighting.dm
```diff
@@ -79,9 +79,12 @@ SUBSYSTEM_DEF(lighting)
 /datum/controller/subsystem/lighting/proc/InitializeZlev(zlev)
 	for (var/thing in Z_ALL_TURFS(zlev))
 		var/turf/T = thing
-		if (TURF_IS_DYNAMICALLY_LIT_UNSAFE(T) && !T.lighting_overlay)	// Can't assume that one hasn't already been created on bay/neb.
-			new /atom/movable/lighting_overlay(T)
-			. += 1
+		if (TURF_IS_DYNAMICALLY_LIT_UNSAFE(T))
+			if (T.lighting_overlay)
+				log_ss(name, "Found unexpected lighting overlay at [T.x],[T.y],[T.z]")
```
MistakeNot4892, line 84: Should this also fail CI?

code/modules/maps/_map_template.dm
```diff
@@ -156,37 +156,9 @@
...
-	init_shuttles(shuttle_state, map_hash, initialized_areas_by_type)
-	after_load()
-	for(var/z_index = bounds[MAP_MINZ] to bounds[MAP_MAXZ])
-		var/datum/level_data/level = SSmapping.levels_by_z[z_index]
-		level.after_template_load(src)
-		if(SSlighting.initialized)
-			SSlighting.InitializeZlev(z_index)
-	log_game("Z-level [name] loaded at [x],[y],[world.maxz]")
-	loaded++
+	var/z = world.maxz + 1
```
MistakeNot4892, line 159: Is there a reason to store it here rather than incrementing world Z size and using `world.maxz` directly?

PsyCommando (commented):
Just a comment on line 152:
`/datum/map_template/proc/load_new_z(no_changeturf = TRUE, centered = TRUE)`

Maybe removing the ``no_changeturf`` arg be desirable here? Since it doesn't do anything anymore?

out-of-phaze:
Yeah, this isn't happening in a reasonable amount of time. Stopgap PR works. Will redo this later and target dev.

PR #3161, opened 2023-06-27
Simplify new z maploading and fix lighting sanity checks

out-of-phaze:
## Description of changes
Restores some missing sanity checks in lighting init code.
Removes code that was violating those sanity checks. As far as I can tell, calling `SSlighting.InitZLev` like that is invalid and leads to the sanity check failing, and some code in ChangeTurf led to double initializations.
`load_new_z` now just calls `load` directly.

Reduces code duplication in `load_new_z`, makes lighting init more sane (and hopefully faster, since we're not wasting any update passes).

### DIFF
--- a/code/controllers/subsystems/lighting.dm
+++ b/code/controllers/subsystems/lighting.dm
@@ -79,9 +79,12 @@ SUBSYSTEM_DEF(lighting)
 /datum/controller/subsystem/lighting/proc/InitializeZlev(zlev)
 	for (var/thing in Z_ALL_TURFS(zlev))
 		var/turf/T = thing
-		if (TURF_IS_DYNAMICALLY_LIT_UNSAFE(T) && !T.lighting_overlay)	// Can't assume that one hasn't already been created on bay/neb.
-			new /atom/movable/lighting_overlay(T)
-			. += 1
+		if (TURF_IS_DYNAMICALLY_LIT_UNSAFE(T))
+			if (T.lighting_overlay)
+				log_ss(name, "Found unexpected lighting overlay at [T.x],[T.y],[T.z]")
+			else
+				new /atom/movable/lighting_overlay(T)
+				. += 1
 			if (T.ambient_light)
 				T.generate_missing_corners()	// Forcibly generate corners.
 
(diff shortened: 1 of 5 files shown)


PsyCommando:
After thinking a bit, I remembered a use case for why load and load_new_z are separate.
If a map file has more than one z-level, load will just load all of them onto the same z-level. While load_new_z will load all of them onto a new z level.

It's not a use case we use on neb, but it's legal. So maybe having some kind of way to deal with that, or maybe an error message would be nice when trying to load a mapfile with more than one zlevel?

out-of-phaze:
> After thinking a bit, I remembered a use case for why load and load_new_z are separate. If a map file has more than one z-level, load will just load all of them onto the same z-level. While load_new_z will load all of them onto a new z level.
> 
> It's not a use case we use on neb, but it's legal. So maybe having some kind of way to deal with that, or maybe an error message would be nice when trying to load a mapfile with more than one zlevel?

Why does load do that? Shouldn't it respect tallness? I thought _load_impl already respected it, even.

PsyCommando:
> > After thinking a bit, I remembered a use case for why load and load_new_z are separate. If a map file has more than one z-level, load will just load all of them onto the same z-level. While load_new_z will load all of them onto a new z level.
> > It's not a use case we use on neb, but it's legal. So maybe having some kind of way to deal with that, or maybe an error message would be nice when trying to load a mapfile with more than one zlevel?
> 
> Why does load do that? Shouldn't it respect tallness? I thought _load_impl already respected it, even.

Because, load literally constrains loading to a specific x,y,z coordinate.
Look at the implementation of load_map_impl:

``(code\modules\maps\reader.dm:60``
```dm
/dmm_suite/proc/load_map_impl(dmm_file, x_offset, y_offset, z_offset, cropMap, measureOnly, no_changeturf, clear_contents, x_lower = -INFINITY, x_upper = INFINITY, y_lower = -INFINITY, y_upper = INFINITY, initialized_areas_by_type, level_data_type = /datum/level_data/space)
	var/tfile = dmm_file//the map file we're creating
	if(isfile(tfile))
		tfile = safe_file2text(tfile, FALSE)

	if(!x_offset)
		x_offset = 1
	if(!y_offset)
		y_offset = 1
	if(!z_offset)
		z_offset = world.maxz + 1
```

This last line is what does it. If you pass a z offset, it'll override using the current highest z offset and adding one. And in the old load_new_z code, it would manually pass world.maxz + 1.

PR #3164, opened 2023-06-28
Moving text loading from directories into a proc.

MistakeNot4892:
## Description of changes
Generalizes the method custom items use to load their json from the file system.

Allows other subsystems to use it.

## Authorship
Myself.

## Changelog
Nothing player-facing.

### DIFF
--- a/code/_helpers/lists.dm
+++ b/code/_helpers/lists.dm
@@ -800,6 +800,70 @@ var/global/list/json_cache = list()
 			log_error("Exception during JSON decoding ([json_to_decode]): [e]")
 	return list()
 
+/proc/load_text_from_directory(var/directory, var/expected_extension = ".txt", var/recursive = TRUE)
+
+	if(!directory)
+		return list(
+			"files"      = list(),
+			"item_count" = 0,
+			"dir_count"  = 0
+		)
+
+	if(copytext(directory, -1) != "/")
+		directory += "/"
+
+	var/list/walked_directories = list()
+	var/list/loaded_files = list()
+	var/dir_count  = 0
+	var/item_count = 0
+
+	// Use a while loop so we can recurse over subdirectories.
+	var/list/directories_to_check = list(directory)
+	// Cache the extension length so we can save some time when checking for matches.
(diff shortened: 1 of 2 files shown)


code/_helpers/lists.dm
```diff
@@ -800,6 +800,60 @@ var/global/list/json_cache = list()
...
+	var/expected_extension_length = -(length(expected_extension))
+	while(length(directories_to_check))
+
+		// Check each directory only once to avoid wasted effort.
+		var/checkdir = directories_to_check[1]
+		directories_to_check -= checkdir
+		walked_directories += checkdir
+
+		// Skip an examples directory if provided.
+		if(checkdir == "[directory]examples/")
```
out-of-phaze, line 833: Noting that it should probably be enforced or checked somewhere that `directory` ends in a `/`, or we should make path handling more robust in general.

code/controllers/subsystems/initialization/customitems.dm
```diff
@@ -6,42 +6,18 @@ SUBSYSTEM_DEF(customitems)
 	var/list/custom_icons_by_ckey = list()
 
 /datum/controller/subsystem/customitems/proc/get_json_paths_from_directory(var/category, var/directory)
-
 	. = list()
```
out-of-phaze: Is this used? It used to return the loaded JSON, but now that's done elsewhere. Only noticed it because I realized the old version didn't have an explicit value returned. May be better to change the returns on line 12 and 16 to `return list()` if returning an empty list on failure is desirable—but no current code seems to rely on it returning an empty list rather than a null.

code/_helpers/lists.dm
MistakeNot4892: Fails on null directory and appends `/` if missing now.

code/controllers/subsystems/initialization/customitems.dm
MistakeNot4892: Removed.

out-of-phaze (approved):

PR #3165, opened 2023-06-28
Character comments system.

MistakeNot4892:
Opening for diff and feedback. I expect this to change a lot over the course of implementation.

## Description of changes
- [X] Adds a subsystem (SScomments) for serializing and deserializing character IC/OOC info and comments. Needs to be a subsystem so you can search and update info regardless of being on a given character inround.
- [x] Adds a pref panel for opting in or out of the comments system and for setting your IC and OOC info.
- [x] Adds a way to view comments for your character, and to add comments to other characters.
- [x] Adds an admin verb to dump out the comments list in a spreadsheet a la the Polaris shipping manifest.

- Formalizes IC/OOC info and adds a way to set and display it without abusing flavour text.
- Adds a system that encourages character roleplay, provides character feedback, 

TODO:
- [x] Finish implementation.
- [x] Test.
- ~~Modpack the system.~~

Screenshots:
image
image

## Authorship
Myself.

## Changelog
:cl:
add: Added a character IC/OOC panel to preferences.
add: Added a character comment tracking system.
/:cl:


MistakeNot4892:
Depends on #3164

MistakeNot4892:
Depends on #3168.

MistakeNot4892:
I'm not sure if I can cleanly modpack this as it fiddles with some core pref code. Otherwise it's ready to review.

MistakeNot4892:
Ready for review.

code/modules/client/preference_setup/global/05_settings.dm
```diff
@@ -24,6 +24,8 @@
 	if(!istype(pref.slot_names))
 		pref.slot_names = list()
 
+	for(var/slot_name in pref.slot_names)
```
MistakeNot4892: Wot.

code/controllers/subsystems/initialization/character_info.dm
```diff
@@ -0,0 +1,174 @@
...
+			continue
+
+		// Remove any pre-created records (early joiners, etc.)
+		if(_comment_holders_by_id[comments.id])
+			var/old_record = _comment_holders_by_id[comments.id]
+			_comment_holders_by_id -= comments.id
+			qdel(old_record)
+
+		comments.file_location = loaded_file
+		_comment_holders_by_id[trim(lowertext(comments.id))] = comments
```
out-of-phaze: Either this should be changed to just `[comments.id]` or lines 28, 29, and 30 should also use trim and lowertext.

out-of-phaze: load_record and get_record also use trim and lowertext so it's likely these need to be changed. You could maybe even just make it use get_record to standardize it:
```dm
		// Remove any pre-created records (early joiners, etc.)
		var/datum/character_information/old_record = get_record(comments.id, TRUE)
		if(istype(old_record))
			_comment_holders_by_id -= old_record.id
			qdel(old_record)
```

code/controllers/subsystems/initialization/character_info.dm
```diff
@@ -0,0 +1,174 @@
...
+/datum/controller/subsystem/character_info/Topic(href, href_list)
+	. = ..()
+	if(!. && href_list["download_manifest"] && usr && check_rights(R_ADMIN))
+		try
+			var/dump_file_name = "[comments_source]dump_character_manifest.html"
+			if(fexists(dump_file_name))
+				fdel(dump_file_name)
+			text2file(JOINTEXT(SScharacter_info.get_character_manifest_html(apply_striping = FALSE)), dump_file_name)
+			if(fexists(dump_file_name))
+				usr << ftp(dump_file_name, "dump_manifest.html")
```
out-of-phaze:
```suggestion
				direct_output(usr, ftp(dump_file_name, "dump_manifest.html"))
```
Or create a macro specifically for `ftp`? That'd mean you wouldn't have to add a `<<` to the count in check-paths.

code/modules/client/preference_setup/global/05_settings.dm
out-of-phaze: Merge artifact?

code/controllers/subsystems/initialization/character_info.dm
```diff
@@ -0,0 +1,177 @@
...
+	var/list/loaded_files = loaded_json_manifest["files"]
+	var/loaded_comments = 0
+	for(var/loaded_file in loaded_files)
+		var/datum/character_information/comments = load_record(loaded_files[loaded_file])
+		if(!istext(comments.id))
+			PRINT_STACK_TRACE("Deserialized character information had invalid id '[comments.id || "NULL"]'!")
+			continue
+
+		// Remove any pre-created records (early joiners, etc.)
+		if(_comment_holders_by_id[comments.id])
```
out-of-phaze: This still doesn't seem to use `trim(lowertext())` or `get_record_id()`.

MistakeNot4892: It shouldn't need to. Comments id should be passed through the proc already at this point.

out-of-phaze: Then why is it done in line 34?

MistakeNot4892: Missed in review - should be consistent now.
Record ID is only set once, in the subsystem creation proc, and it's trimmed/lowertexted there now, so it shouldn't be needed when referring to `comments.record_id` anymore.

out-of-phaze (dismissed):

MistakeNot4892:
Need to add some forgotten features to this:
- [x] Ability to prune or skip loading records for character who haven't been seen for a certain amount of time.
- [X] Ability to set secondary mood to change border colour.

MistakeNot4892:
Ended up just hiding comments older than a particular time set in config. Very iffy on some of the implementation in this PR, review would be appreciated.

MistakeNot4892:
Going to reopen this one after seeing how the system works on Polaris for awhile.

MistakeNot4892:
System never made it into Polaris due to differences in how stuff was set up in prefs there.

MistakeNot4892:
I'm iffy on some of the internals, particularly around when and where character records are loaded and how they're stored, so any input on that would be appreciated. Otherwise this is generally ready for review.

out-of-phaze (approved):

PR #3172, opened 2023-06-30
Remove shuttle console damage invulnerabilities

out-of-phaze:
## Description of changes
Shuttle consoles are no longer immune to damage, because they can be rebuilt.

I dropped a giant bomb on a ship and the only thing left was the shuttle console, which was pretty jarring. Also, you could just take it apart and then the circuitboard (with the data) would still be able to be destroyed, so the original rationale no longer really applies either.

### DIFF
--- a/code/modules/shuttles/shuttle_console.dm
+++ b/code/modules/shuttles/shuttle_console.dm
@@ -117,13 +117,3 @@
 		hacked = 1
 		to_chat(user, "You short out the console's ID checking system. It's now available to everyone!")
 		return 1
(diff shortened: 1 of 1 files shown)


MistakeNot4892:
Noting that multishuttles can't be rebuilt currently, we need a `/obj/item/stock_parts/circuitboard/shuttle_console/multi` for them, but the shuttle tag setting interaction covers my main concern with this.

MistakeNot4892 (approved):

PR #3177, opened 2023-07-01
Moving aquatic simple_animal level down to base simple_animal.

MistakeNot4892:
## Description of changes
Merges aquatic logic at base simple_animal level.

Cleaner code.

## Authorship
Myself.

## Changelog
Nothing player-facing.


code/modules/mob/living/simple_animal/simple_animal.dm
```diff
@@ -109,25 +117,36 @@
...
+		global.simplemob_icon_bitflag_cache[type] = mob_icon_state_flags
 
 /mob/living/simple_animal/on_update_icon()
 
 	..()
 
 	icon_state = ICON_STATE_WORLD
-	if(stat == DEAD && (mob_icon_state_flags & MOB_ICON_HAS_DEAD_STATE))
+	if(is_aquatic && stat != DEAD && HAS_STATUS(src, STAT_PARA) && (mob_icon_state_flags & MOB_ICON_HAS_DYING_STATE))
+		icon_state += "-dying"
```
afterthought2: Why gate this behind aquatic, as opposed to just letting the `HAS_STATUS` figure it out?

MistakeNot4892: Valid point I guess. I guess mainly because STAT_PARA doesn't imply dying for most mobs. But that said, most mobs won't have a `world-dying` state.

MistakeNot4892: I might rename it to `-paralyzed` and remove the aquatic check.

MistakeNot4892: Done.

afterthought2 (dismissed):

code/modules/mob/living/simple_animal/aquatic/aquatic_carp.dm
```diff
@@ -6,13 +6,6 @@
 	maxHealth = 20
 	health = 20
 
-	meat_type = /obj/item/chems/food/fish/carp
```
out-of-phaze, line 9: Unintentional change? `retaliate/aquatic` only has normal fish meat, not carp meat.

code/modules/mob/living/simple_animal/simple_animal.dm
```diff
@@ -160,6 +179,11 @@
 	. = ..()
 
 /mob/living/simple_animal/Life()
+	if(is_aquatic && !submerged() && stat != DEAD)
+		walk(src, 0)
+		if(PENDING_STATUS(src, STAT_PARA))
```
out-of-phaze: Doesn't this mean it won't get paralyzed if it isn't already?

code/modules/mob/living/simple_animal/simple_animal.dm
```diff
@@ -245,30 +269,29 @@
 
 /mob/living/simple_animal/proc/handle_atmos(var/atmos_suitable = 1)
 	//Atmos
-
 	if(!loc)
 		return
 
+	// don't bother checking it twice if we got a supplied 0 val.
```
out-of-phaze: Sort of a nitpick but this comment is about checking `atmos_suitable`, not `environment`, isn't it? Shouldn't it be next to that line?

out-of-phaze (changes requested):

code/modules/mob/living/simple_animal/simple_animal.dm
MistakeNot4892: Yeah that's missing a `!`. Should probably be HAS_STATUS I guess though that puts suffocation on a 2-second delay which might not be ideal.

out-of-phaze: But it already uses `SET_STATUS_MAX` anyway, why are we checking that to begin with...?

code/modules/mob/living/simple_animal/simple_animal.dm
```diff
@@ -160,6 +179,11 @@
 	. = ..()
 
 /mob/living/simple_animal/Life()
+	if(is_aquatic && !submerged() && stat != DEAD)
+		walk(src, 0)
+		if(!HAS_STATUS(src, STAT_PARA)) // gated to avoid redundant update_icon() calls.
```
out-of-phaze, line 184: I'll note that this probably means you'll get a free tick or two without paralysis every 3 or so ticks, but maybe that's intended?

out-of-phaze (approved):

PR #3179, opened 2023-07-01
Fixes dropping inventory on species change

NataKilar:
## Description of changes
Reworks the ``mob_can_equip`` proc to accommodate checking for equipment already on the mob by adding the argument ``ignore_equipped``. This previously caused items to be deleted when changing species (and also caused an issue when deserializing downstream).

Adjusts the ``set_inventory_slots`` proc to drop items in ``old_slots`` into the world if no equivalent slot exists in ``new_slots``, rather then keeping them in mob contents. Currently items in hands will still be dropped on species transform, presumably due to the way organs are changed, but I think that's out of scope to fix for this PR.

Fixes bugs with species transformation

## Authorship
Myself

### DIFF
--- a/code/datums/inventory_slots/_inventory_slot.dm
+++ b/code/datums/inventory_slots/_inventory_slot.dm
@@ -133,8 +133,8 @@
 		return FALSE
 	return TRUE
 
-/datum/inventory_slot/proc/can_equip_to_slot(var/mob/user, var/obj/item/prop, var/disable_warning)
-	return (!_holding && prop && slot_id && prop_can_fit_in_slot(prop))
+/datum/inventory_slot/proc/can_equip_to_slot(var/mob/user, var/obj/item/prop, var/disable_warning, var/ignore_equipped)
+	return ((!_holding || ignore_equipped && _holding == prop) && prop && slot_id && prop_can_fit_in_slot(prop))
 
 /datum/inventory_slot/proc/prop_can_fit_in_slot(var/obj/item/prop)
 	return (isnull(requires_slot_flags) || (requires_slot_flags & prop.slot_flags))
--- a/code/datums/inventory_slots/_inventory_slot.dm
+++ b/code/datums/inventory_slots/_inventory_slot.dm
@@ -134,7 +134,7 @@
 	return TRUE
 
 /datum/inventory_slot/proc/can_equip_to_slot(var/mob/user, var/obj/item/prop, var/disable_warning, var/ignore_equipped)
-	return ((!_holding || ignore_equipped && _holding == prop) && prop && slot_id && prop_can_fit_in_slot(prop))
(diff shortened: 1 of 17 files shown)


code/datums/inventory_slots/_inventory_slot.dm
```diff
@@ -133,8 +133,8 @@
 		return FALSE
 	return TRUE
 
-/datum/inventory_slot/proc/can_equip_to_slot(var/mob/user, var/obj/item/prop, var/disable_warning)
-	return (!_holding && prop && slot_id && prop_can_fit_in_slot(prop))
+/datum/inventory_slot/proc/can_equip_to_slot(var/mob/user, var/obj/item/prop, var/disable_warning, var/ignore_equipped)
+	return ((!_holding || ignore_equipped && _holding == prop) && prop && slot_id && prop_can_fit_in_slot(prop))
```
MistakeNot4892: `(!_holding || (ignore_equipped && _holding == prop))` would probably be clearer.

code/modules/mob/living/inventory.dm
```diff
@@ -122,17 +122,30 @@
 		for(var/slot in get_held_item_slots())
 			if(slot in old_slots)
 				LAZYSET(_inventory_slots, slot, old_slots[slot])
+				old_slots -= slot
```
MistakeNot4892, line 125: Isn't this going to prevent the comparisons in the upcoming for loop from working correctly?

PsyCommando:
I thought we already had something to prevent this downstream?

NataKilar:
> I thought we already had something to prevent this downstream?

Not post slot rework. I'm sure we could do a workaround, but this is cleaner and fixes a upstream bug

code/modules/mob/living/inventory.dm
NataKilar: It might, yeah. I'm not clear on the intended behavior - do we want to keep the held item slots by default and only replace them if there are new held item slots passed in ``new_slots`` (in which case the current behavior should work ok), or do we always want to keep the held item slots the same regardless of what is passed in ``new_slots``?

code/datums/inventory_slots/_inventory_slot.dm
NataKilar: Fixed

code/modules/mob/living/inventory.dm
MistakeNot4892: Looks like the held slot filtering is already done before `set_inventory_slots` is called. Not sure why it's being done twice  but you're right, this change is fine.

MistakeNot4892 (approved):

out-of-phaze:
This seems to have introduced a bug when loading in with loadout items selected on dev.

PR #3183, opened 2023-07-02
Simplifies status checking a bit, also allows same-tick checking to work.

MistakeNot4892:
## Description of changes
Condenses PENDING_STATUS into HAS_STATUS.

Checking status on the tick that the status is adjusted will now work.

## Authorship
Myself.

## Changelog
Nothing player-facing.


MistakeNot4892:
Realized this has some unaddressed complexity and probably isn't desirable as it messes with the existing status update logic.

PR #3185, opened 2023-07-03
Fix carbon paper copy issue

PsyCommando:
## Description of changes
* The carbon paper would just create 2 copies of itself because it called Clone() on itself expecting to get a regular sheet of paper...
* Fixed path for the get_alt_interactions(mob/user) proc for carbon paper.

## Authorship
Thanks to @out-of-phaze for pointing that mistake I made.

## Changelog
:cl:
bugfix: fixed carbon paper creating a copy of itself when trying to split the original and copy.
bugfix: fixed the paint spray having an interaction to remove a carbon copy from it for reasons.
/:cl:

### DIFF
--- a/code/modules/paperwork/carbonpaper.dm
+++ b/code/modules/paperwork/carbonpaper.dm
@@ -11,8 +11,12 @@
 		add_overlay("paper_stack_words")
 
 /obj/item/paper/carbon/proc/remove_copy(var/mob/user)
-	var/obj/item/paper/original = Clone()
-	var/obj/item/paper/copy     = Clone()
+	//Make a new paper that copies our contents
+	var/obj/item/paper/original = new
+	original.PopulateClone(src)
+	original.updateinfolinks()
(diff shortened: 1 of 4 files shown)


code/modules/paperwork/carbonpaper.dm
```diff
@@ -11,23 +11,47 @@
...
-	copy.set_content("<font color = #101010>[copycontents]</font>", "Copy - [original.name]")
+	copy.set_content("<font color = #101010>[copycontents]</font>", (original.name != initial(original.name))? "Copy - [original.name]" : null)
+	original.update_icon()
 
 	qdel(src)
 	user.put_in_active_hand(original)
 	user.put_in_hands(copy)
 	return copy
 
-/obj/item/paint_sprayer/get_alt_interactions(mob/user)
```
MistakeNot4892, line 30: Copypaste my foe...

MistakeNot4892 (approved):
Needs a squash merge.

PR #3190, opened 2023-07-04
Removing some reliance on /datum/dna.

MistakeNot4892:
## Description of changes
- adds a `/mob/proc/get_unique_enzymes()` helper and uses it where possible..
- adds a `/mob/proc/get_blood_type()` helper and uses it where possible. Also renames `b_type` in a bunch of spots for clarity.
- made fingerprints use `get_full_print()` where possible.
- converts pAIs and gibs to not rely directly on `/datum/dna`.

Less deps on DNA, easier to remove/rewrite it.

## Authorship
Myself.

## Changelog
Nothing player-facing.

### DIFF
--- a/code/game/dna/dna2.dm
+++ b/code/game/dna/dna2.dm
@@ -59,9 +59,9 @@ var/global/list/assigned_blocks[DNA_SE_LENGTH]
 /datum/dna
 	// READ-ONLY, GETS OVERWRITTEN
 	// DO NOT FUCK WITH THESE OR BYOND WILL EAT YOUR FACE
-	var/uni_identity="" // Encoded UI
-	var/struc_enzymes="" // Encoded SE
-	var/unique_enzymes="" // MD5 of player genetic marker value
+	var/uni_identity=""   // Encoded UI
+	var/struc_enzymes=""  // Encoded SE
+	var/unique_enzymes="" // MD5 of mob genetic marker value
 
 	var/fingerprint
 
--- a/code/game/dna/dna2.dm
+++ b/code/game/dna/dna2.dm
@@ -151,7 +151,7 @@ var/global/list/assigned_blocks[DNA_SE_LENGTH]
 		if(LAZYLEN(E.markings))
 			body_markings[E.organ_tag] = E.markings.Copy()
 
-	b_type = character.b_type
+	b_type = character.blood_type
 
 	UpdateUI()
 
--- a/code/game/dna/dna2.dm
+++ b/code/game/dna/dna2.dm
@@ -61,7 +61,7 @@ var/global/list/assigned_blocks[DNA_SE_LENGTH]
 	// DO NOT FUCK WITH THESE OR BYOND WILL EAT YOUR FACE
 	var/uni_identity="" // Encoded UI
 	var/struc_enzymes="" // Encoded SE
-	var/unique_enzymes="" // MD5 of player name
+	var/unique_enzymes="" // MD5 of player genetic marker value
 
 	var/fingerprint
 
@@ -141,7 +141,8 @@ var/global/list/assigned_blocks[DNA_SE_LENGTH]
 
 	SetUIState(DNA_UI_GENDER, character.gender!=MALE, 1)
 
-	fingerprint = md5(character.unique_mob_number)
+	fingerprint    = character.get_full_print(ignore_blockers = TRUE)
+	unique_enzymes = character.get_unique_enzymes()
 
 	// Hair
 	var/list/hair_types = decls_repository.get_decl_paths_of_subtype(/decl/sprite_accessory/hair)
@@ -156,7 +157,7 @@ var/global/list/assigned_blocks[DNA_SE_LENGTH]
 		if(LAZYLEN(E.markings))
 			body_markings[E.organ_tag] = E.markings.Copy()
 
-	b_type = character.blood_type
+	b_type = character.get_blood_type()
 
 	UpdateUI()
 
@@ -356,7 +357,7 @@ var/global/list/assigned_blocks[DNA_SE_LENGTH]
 			ResetSE()
 
 		if(length(unique_enzymes) != 32)
-			unique_enzymes = md5(character.real_name)
+			unique_enzymes = md5(num2text(character.original_genetic_seed))
 	else
 		if(!species)
 			species = global.using_map.default_species
@@ -370,5 +371,4 @@ var/global/list/assigned_blocks[DNA_SE_LENGTH]
 /datum/dna/proc/ready_dna(mob/living/carbon/human/character)
 	ResetUIFrom(character)
 	ResetSE()
-	unique_enzymes = md5(character.real_name)
-	global.reg_dna[unique_enzymes] = character.real_name
+	unique_enzymes = character.get_unique_enzymes()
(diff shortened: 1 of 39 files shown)


out-of-phaze:
Before review: noting that it also makes it harder to metagame fingerprints/DNA in forensics, I think, because they're now generated using sequential ID rather than mob name. A good change, but undocumented in the PR body.

code/game/dna/dna2.dm
```diff
@@ -61,7 +61,9 @@ var/global/list/assigned_blocks[DNA_SE_LENGTH]
 	// DO NOT FUCK WITH THESE OR BYOND WILL EAT YOUR FACE
 	var/uni_identity="" // Encoded UI
 	var/struc_enzymes="" // Encoded SE
-	var/unique_enzymes="" // MD5 of player name
+	var/unique_enzymes="" // MD5 of player genetic marker value
```
out-of-phaze:
```suggestion
	var/unique_enzymes="" // MD5 of character genetic marker value
```
Always bugged me how older code doesn't maintain a proper distinction between player and character. Not blocking, but I'd prefer it changed if you're doing any other changes.

code/game/dna/dna2_helpers.dm
```diff
@@ -133,62 +133,57 @@
...
+	dna.check_integrity()
+
+	fingerprint        = dna.fingerprint
+	unique_enzymes     = dna.unique_enzymes
+	hair_colour        = rgb(dna.GetUIValueRange(DNA_UI_HAIR_R,255),  dna.GetUIValueRange(DNA_UI_HAIR_G,255),  dna.GetUIValueRange(DNA_UI_HAIR_B,255))
+	facial_hair_colour = rgb(dna.GetUIValueRange(DNA_UI_BEARD_R,255), dna.GetUIValueRange(DNA_UI_BEARD_G,255), dna.GetUIValueRange(DNA_UI_BEARD_B,255))
+	skin_colour        = rgb(dna.GetUIValueRange(DNA_UI_SKIN_R,255),  dna.GetUIValueRange(DNA_UI_SKIN_G,255),  dna.GetUIValueRange(DNA_UI_SKIN_B,255))
+	eye_colour         = rgb(dna.GetUIValueRange(DNA_UI_EYES_R,255),  dna.GetUIValueRange(DNA_UI_EYES_G,255),  dna.GetUIValueRange(DNA_UI_EYES_B,255))
+	skin_tone          = 35 - dna.GetUIValueRange(DNA_UI_SKIN_TONE, 220) // Value can be negative.
+	update_eyes()
```
out-of-phaze: `update_eyes()` is already called later, on line 184.

code/game/dna/dna2_helpers.dm
```diff
@@ -133,62 +133,57 @@
...
+		h_style = hair_subtypes[hair]
+
+	//Facial Hair
+	var/list/beard_subtypes = decls_repository.get_decl_paths_of_subtype(/decl/sprite_accessory/facial_hair)
+	var/beard = dna.GetUIValueRange(DNA_UI_BEARD_STYLE, length(beard_subtypes))
+	if((0 < beard) && (beard <= length(beard_subtypes)))
+		f_style = beard_subtypes[beard]
+
+	force_update_limbs()
+	update_body()
```
out-of-phaze:
```suggestion
```
`force_update_limbs()` calls `update_body(0)` already.

code/game/dna/dna2_helpers.dm
```diff
@@ -133,62 +133,57 @@
...
+	update_eyes()
+	update_hair()
```
out-of-phaze:
```suggestion
	update_hair(update_icons = FALSE)
	update_eyes()
```
`update_hair()` should skip the `update_icon()` call because more updates are done after.
`update_eyes()` should go last, because it (for some reason) calls `refresh_visible_overlays()`. (I wouldn't skip `update_hair()` entirely, because it only does so if you have eyes, and not updating hair because you're missing eyes is a bit weird.)

code/modules/client/preference_setup/general/02_body.dm
out-of-phaze, line 1: Mid-line indentation on some of the changed lines here seems to be messed up? Non-blocking.

code/modules/organs/organ.dm
```diff
@@ -121,10 +121,7 @@
 /obj/item/organ/proc/set_dna(var/datum/dna/new_dna)
 	QDEL_NULL(dna)
 	dna = new_dna.Clone()
-	if(!blood_DNA)
-		blood_DNA = list()
-	blood_DNA.Cut()
-	blood_DNA[dna.unique_enzymes] = dna.b_type
+	LAZYSET(blood_DNA, dna.unique_enzymes, dna.b_type)
```
out-of-phaze: The `Cut()` is now missing, meaning old blood DNA will persist, won't it? Should just do `blood_DNA = list((dna.unique_enzymes) = dna.b_type)` or something, no?

code/modules/organs/blood.dm
```diff
@@ -203,19 +203,30 @@
...
-	data["dose_chem"] = chem_doses ? chem_doses.Copy() : list()
-	data["blood_color"] = species.get_blood_color(src)
+	data["trace_chem"]  = temp_chem
+	data["dose_chem"]   = chem_doses ? chem_doses.Copy() : list()
+
+	var/decl/species/my_species = get_species()
+	if(my_species)
+		data["has_oxy"]     = my_species.blood_oxy
+		data["blood_color"] = my_species.get_blood_color(src)
+	else if(isSynthetic())
```
out-of-phaze, line 224: What's with this new use of `isSynthetic()`? Is `species.blood_oxy` not good enough? Note that there *are no* FBPs; blood color, oxygenation, etc. are already set on species like utility frames. This just makes it so you can't change the blood color of synthetic species at all.

EDIT: I just noticed this is only if species is false, which... is a bit odd, but I guess there's no problem with it. It won't clobber species data, at least.

out-of-phaze (changes requested):
Conceptual issue: if a character leaves from cryo and rejoins, they are going to have a different fingerprint/unique enzymes, which would make it so you would lose, at the very least, any forensic info and your owned pAI.

code/game/dna/dna2.dm
MistakeNot4892: What would you like changed, sorry?

out-of-phaze: See the suggested change in the comment. It's a nitpick so I'm not too pressed about it, but if you make additional changes to the current state of the PR I'd like it to be done.

MistakeNot4892:
This is ready for re-review.

out-of-phaze (dismissed):

code/modules/mob/living/say.dm
```diff
@@ -78,7 +78,7 @@
 			var/obj/item/radio/radio = thing.get_radio(message_mode)
 			if(istype(radio))
 				LAZYDISTINCTADD(., radio)
-
+me
```
out-of-phaze:
```suggestion
```

out-of-phaze (changes requested):

code/modules/mob/living/say.dm
MistakeNot4892: me irl

out-of-phaze (approved):

PR #3193, opened 2023-07-04
Migrate some computer part behavior upwards

out-of-phaze:
## Description of changes
- Moves `external_slot` on computer parts to `PART_FLAG_INTERACT_CLOSED` on stock parts.
- Replaces `enabled` on computer parts with `PART_STAT_ACTIVE`.
- Replaces `hardware_size` with `w_class` since they were already synonymous.
- Adds supporting code for installing stock parts into assemblies (currently no non-computer parts are allowed anyway).
- Fixes duplicate tesla relays on modular consoles.

Lays some of the necessary groundwork for combining stock parts and computer parts.

## Future issues:
- It's unclear how stock parts should be displayed and interacted with in the computer configuration menu's hardware list.
- The usage and behavior of `PART_STAT_ACTIVE` is not standardized, leading to odd results when the user is made able to toggle it at will in the configuration menu.
- `power_usage` should probably be moved to stock parts so that part selection can affect power usage without bespoke per-machine code.
- stock part `on_install` always expects a machine—may need to be split into `on_install`, `on_install_machine`, and `on_install_assembly`?
- `critical` should probably be replaced with something else—maybe an 'in use' stat flag that prevents it from being disabled while active, so if network boot is added, you should be able to disable a local hard drive if it's not the OS drive?
- `critical_parts` should be replaced with things like OS drive checks, power supply checks, etc.

### DIFF
--- a/code/modules/modular_computers/os/ui.dm
+++ b/code/modules/modular_computers/os/ui.dm
@@ -74,14 +74,14 @@
 		return TOPIC_REFRESH
 	if(href_list["PC_enable_component"] )
 		var/obj/item/stock_parts/computer/H = locate(href_list["PC_enable_component"]) in holder
-		if(H && istype(H) && !H.enabled)
-			H.enabled = 1
+		if(H && istype(H) && !(H.status & PART_STAT_ACTIVE))
+			H.set_status(holder, PART_STAT_ACTIVE)
 			H.on_enable(src)
 		return TOPIC_REFRESH
 	if(href_list["PC_disable_component"] )
 		var/obj/item/stock_parts/computer/H = locate(href_list["PC_disable_component"]) in holder
-		if(H && istype(H) && H.enabled)
-			H.enabled = 0
+		if(H && istype(H) && (H.status & PART_STAT_ACTIVE))
+			H.unset_status(holder, PART_STAT_ACTIVE)
 			H.on_disable()
 		return TOPIC_REFRESH
 	if( href_list["PC_enable_update"] )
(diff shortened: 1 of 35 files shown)


code/modules/modular_computers/os/ui.dm
```diff
@@ -74,14 +74,14 @@
...
+		if(H && istype(H) && !(H.status & PART_STAT_ACTIVE))
+			H.set_status(holder, PART_STAT_ACTIVE)
 			H.on_enable(src)
 		return TOPIC_REFRESH
 	if(href_list["PC_disable_component"] )
 		var/obj/item/stock_parts/computer/H = locate(href_list["PC_disable_component"]) in holder
-		if(H && istype(H) && H.enabled)
-			H.enabled = 0
+		if(H && istype(H) && (H.status & PART_STAT_ACTIVE))
+			H.unset_status(holder, PART_STAT_ACTIVE)
```
afterthought2, line 84: I believe the overall structure of the code is such that you aren't allowed to call this externally on parts: the part calls this internally to trigger events on the machine, but is updating its own status alone. This relates to one of the points in your PR header, that `PART_STAT_ACTIVE` is used inconsistently. For power components, `ACTIVE` is not really "turned on," it's more like "I'm the one currently providing power," and to set/unset it you are supposed to use `not_needed`/`can_provide_power`.

I'd suggest something like a stock part proc for "activated/deactivated by computer." and a getter for "am I on." Then parts can do whatever internally, possibly using a separate flag. It might even be that toggling the `INSTALLED` flag is closer to what you want. Obviously this is annoying to implement until the part types are combined.

I guess it's also worth considering: what purpose does manually activating/deactivating parts without removal in a computer serve? With machines, my conclusion was that it served no purpose, so they just don't have this concept. Of course computers are likely different.

MistakeNot4892:
What's the status on this one?

out-of-phaze:
> What's the status on this one?

Needs to be redone.

PR #3196, opened 2023-07-07
Improve and fix various aspects of the codex

### LINKED ISSUE #2937: Atmospherics codex entries lack a name field
#### Description of issue
Title. This causes them to not show up in-game as they are never generated due to not having a name.

#### Steps to reproduce
Compile.

code\modules\codex\entries\atmospherics.dm

🤷 

dev as of this issue

out-of-phaze:
## Description of changes
- Replaces a `centry.name` check with a `TYPE_IS_ABSTRACT(centry) || !initial(centry.store_codex_entry) || initial(centry.skip_hardcoded_generation)` check, so that codex entries without a predefined name will show up.
- Adds static typing hints to some of the codex lists.
- Adds a variable that makes `associated_paths` a list of root paths to expand via `typesof()` rather than a direct list of paths, and sets it on a handful of types.
- Renames some codex entries to avoid associated string overlaps.
- Makes the associated string overlap test less strict (it is fine for them to have overlaps if they point to the same entry).
- Adds some basic deduplication handling for the `entries_by_path`/`entries_by_string` lists.
- Downgrades some duplicate entry stacktrace warnings to deduplication debug logging.
- Fixes the atmos pipe entry in the codex being named `"machinery (atmospherics)"` by adding the name `"pipe"` to the base pipe type.
- Removes some no-longer-necessary `get_codex_value()` overrides; many of them didn't work anyway because they were hardcoded strings.

While I was at it I noticed a few things:
- Xenoarch finds have their prefix applied twice, probably due to merge skew from two separate fixes.
- Fixes some comment typos I found.
- Uses `\A` and `\a` where necessary so that things say `an object` instead of `a object`.

Current issues:
- The codex deduplication will spam `log_debug` warnings that don't really serve much of a purpose, because overwriting general types with more specific ones is okay in almost all instances. I'd like a better solution but if that's not possible I'll probably just remove the log lines entirely.

Closes #2937.
This allows you to access relevant codex entries from many more objects than before, and makes more codex entries show up to begin with.


MistakeNot4892 (approved):
Bit perplexed by the random trailing comma but otherwise fine.

PR #3197, opened 2023-07-09
Add define-gated profiling code and implement a few optimizations found with it

out-of-phaze:
## Description of changes
The profiling code is set up to support BYOND-Tracy and TG's (relatively old/outdated at this point?) profiler macros.

Aside from that, the optimizations:
- Fixes noisemap smoothing passes being wasted.
- Replaces the naive blur/smoothing implementation with one found at http://amritamaz.net/blog/understanding-box-blur
- Fixes the adjacent area registration update proc being called on the same turf five times, instead of on the adjacent ones.
- Makes it so that the buried_resources extension isn't initialized as soon as it's created.
- Fixes the ore noisemap being placed at 32, 32 instead of 8, 8.
- Merges `map_generators` into `level_generators` to fix an ordering issue with ore noisemaps being overwritten by exoplanet terrain noisemaps.
- Moves the include for `planet_testing_define.dm` out of the map and into the DME, as is standard. (TODO: document this.)
- Optimize `smooth_single_tiles` in noisemap and make it check path instead of value. Doesn't currently work but I'm not sure it did before.

Profiling support code just makes my life easier, instead of reapplying the same stash I can just have a local copy of the byond-tracy dll and then enable it with the define.
In my testing it was roughly 3-4 seconds added to init per random exoplanet z-level, with ~8 or so taking 32 seconds; doubling their height made init take ~60 seconds. 3-4 seconds per 255x255 z-level seems pretty good to me!

## Changelog
:cl:
bugfix: Some exoplanet terrain should be generated more smoothly, as intended.
/:cl:

### DIFF
--- a/code/modules/maps/template_types/random_exoplanet/planet_types/chlorine.dm
+++ b/code/modules/maps/template_types/random_exoplanet/planet_types/chlorine.dm
@@ -23,6 +23,7 @@
 	exterior_atmos_temp            = null
 	level_generators               = list(
 		/datum/random_map/noise/exoplanet/chlorine,
+		/datum/random_map/noise/ore/poor,
 	)
 
 ////////////////////////////////////////////////////////////////////////////
@@ -91,9 +92,6 @@
 	prefered_level_data_per_z  = list(
 		/datum/level_data/planetoid/exoplanet/chlorine,
 		/datum/level_data/planetoid/exoplanet/underground
-	)
-	map_generators = list(
-		/datum/random_map/noise/ore/poor,
 	)
 
 ////////////////////////////////////////////////////////////////////////////
--- a/.gitignore
+++ b/.gitignore
@@ -45,6 +45,7 @@ tags
 # ignore built libs
 lib/*.dll
 lib/*.so
+/prof.dll
 
 # python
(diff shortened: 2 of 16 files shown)


code/modules/maps/template_types/random_exoplanet/planet_types/chlorine.dm
```diff
@@ -23,6 +23,7 @@
 	exterior_atmos_temp            = null
 	level_generators               = list(
 		/datum/random_map/noise/exoplanet/chlorine,
+		/datum/random_map/noise/ore/poor,
```
comma, line 26: here and everywhere, wouldn't trailing comma result in a null entry in the lists?

MistakeNot4892: Hasn't been an issue since a few versions ago afaik.

out-of-phaze: It had a trailing comma before. It just reduces line diff noise, I figured since it already had one I would keep it that way.

comma (approved):

PR #3201, opened 2023-07-11
Add modcomp diskreader component

out-of-phaze:
## Description of changes
Adds a temporary computer part variant of the disk reader machine component. Works a lot like the hard drive slot component. Doesn't support directories or non-data files.

TODO: Merge this into the main disk reader component after #3193 and some later changes are in.

It's good to have a way to interact with these files. (Notably, it handles access checks, so if this makes something like survey points able to be hacked, we can either do that or add an integrity check (like having the file uid as metadata) to make sure it hasn't been tampered with.

## Authorship
Me.

## Changelog
:cl:
add: Added a computer disk reader component that allows you to read and write disk contents.
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/tech_disks.dm
+++ b/code/game/objects/items/weapons/tech_disks.dm
@@ -50,7 +50,8 @@
 	if(!F || (F.read_only && !force))
 		return FALSE
 	free_blocks = clamp(round(free_blocks + F.block_size), 0, block_capacity)
-	qdel(LAZYACCESS(stored_files, name))
+	// do not qdel; should be GC'd once it has no references anyway
+	F.holder = null
 	LAZYREMOVE(stored_files, name)
 	return TRUE
 
--- a/code/modules/modular_computers/hardware/disk_slot.dm
+++ b/code/modules/modular_computers/hardware/disk_slot.dm
@@ -1,0 +1,108 @@
+/obj/item/stock_parts/computer/data_disk_drive
+	name = "data disk slot"
+	desc = "Slot that allows this computer to accept data disks."
+	power_usage = 10 //W
+	critical = 0
+	icon_state = "cardreader"
+	hardware_size = 1
+	origin_tech = "{'programming':2}"
+	usage_flags = PROGRAM_ALL
+	external_slot = TRUE
+	material = /decl/material/solid/metal/steel
+
+	var/obj/item/disk/stored_disk = null
+	var/mount_name
+
+/obj/item/stock_parts/computer/data_disk_drive/diagnostics()
+	. = ..()
+	. += "[name] status: [stored_disk ? "Disk Inserted" : "Disk Not Present"]\n"
+
+/obj/item/stock_parts/computer/data_disk_drive/proc/verb_eject_disk()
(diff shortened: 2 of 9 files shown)


code/game/objects/items/weapons/tech_disks.dm
```diff
@@ -50,7 +50,8 @@
 	if(!F || (F.read_only && !force))
 		return FALSE
 	free_blocks = clamp(round(free_blocks + F.block_size), 0, block_capacity)
-	qdel(LAZYACCESS(stored_files, name))
+	// do not qdel; should be GC'd once it has no references anyway
+	F.holder = null
```
MistakeNot4892, line 54: What's the distinction between qdel and nulling holder here? Why don't we null holder in Destroy() and use qdel?

code/modules/modular_computers/hardware/disk_slot.dm
```diff
@@ -0,0 +1,108 @@
...
+	. = ..()
+	. += "[name] status: [stored_disk ? "Disk Inserted" : "Disk Not Present"]\n"
+
+/obj/item/stock_parts/computer/data_disk_drive/proc/verb_eject_disk()
+	set name = "Remove Disk"
+	set category = "Object"
+	set src in view(1)
+
+	if(!CanPhysicallyInteract(usr))
+		to_chat(usr, "<span class='warning'>You can't reach it.</span>")
```
MistakeNot4892: `SPAN_WARNING`

code/modules/modular_computers/hardware/disk_slot.dm
```diff
@@ -0,0 +1,108 @@
...
+		return
+
+	var/obj/item/stock_parts/computer/data_disk_drive/device = src
+	if (!istype(device))
+		device = locate() in src
```
MistakeNot4892, line 31: Probably need to nullcheck this again after locate().

code/modules/modular_computers/hardware/disk_slot.dm
```diff
@@ -0,0 +1,108 @@
...
+		return FALSE
+
+	if(user)
+		to_chat(user, "You remove [stored_disk] from [src].")
+		user.put_in_hands(stored_disk)
+	else
+		dropInto(loc)
+	stored_disk = null
+
+	loc.verbs -= /obj/item/stock_parts/computer/data_disk_drive/proc/verb_eject_disk
```
MistakeNot4892, line 54: Should this be handled by Moved or something on the disk?

code/game/objects/items/weapons/tech_disks.dm
out-of-phaze: That made file transfers break because they check QDELETED on the transferring file.

code/modules/modular_computers/hardware/disk_slot.dm
out-of-phaze: This and the above code were copied directly from the portable hard drive slot. I can still update it if you'd like, but my goal was to get it to parity with that.

MistakeNot4892 (approved):

PR #3203, opened 2023-07-13
Fix job items replaced by loadout being deleted

out-of-phaze:
## Description of changes
Removes some duplicate item deletion code that ran in `mob_can_equip`, which should ideally be pure even with the force var set.

Fixes a runtime on dev when a character is loaded/spawns in with loadout items selected.

### DIFF
--- a/code/game/objects/item.dm
+++ b/code/game/objects/item.dm
@@ -475,15 +475,8 @@
 	if(!inv_slot)
 		return FALSE
 
-	var/already_equipped = inv_slot.get_equipped_item()
-	if(!ignore_equipped || already_equipped != src)
-		if(already_equipped)
-			if(!force)
-				return FALSE
-			inv_slot.clear_slot()
-			qdel(already_equipped)
-
-		if(!inv_slot.is_accessible(M, src, disable_warning))
+	if(!force)
+		if(!ignore_equipped && !inv_slot.is_accessible(M, src, disable_warning))
 			return FALSE
 
 	if(!inv_slot.can_equip_to_slot(M, src, disable_warning, ignore_equipped))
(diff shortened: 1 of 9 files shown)


code/game/objects/item.dm
```diff
@@ -475,19 +475,12 @@
...
-	if(!ignore_equipped || already_equipped != src)
-		if(already_equipped)
-			if(!force)
-				return FALSE
-			inv_slot.clear_slot()
-			qdel(already_equipped)
-
-		if(!inv_slot.is_accessible(M, src, disable_warning))
+	if(!force)
+		if(!ignore_equipped && !inv_slot.is_accessible(M, src, disable_warning))
```
NataKilar, line 479: As mentioned on Discord, my implementation for ``ignore_equipped`` was only intended to ignore blocking from the src item if it was equipped, not blocking from all items equipped. This was why the if statement had ``||`` rather than ``&&``. It's probably okay to change this, but the if statements in the overrides on gloves and magboots should be changed as well.

code/game/objects/item.dm
```diff
@@ -475,19 +475,12 @@
...
 			return FALSE
 
-	if(!inv_slot.can_equip_to_slot(M, src, disable_warning, ignore_equipped))
-		return FALSE
+		if(!inv_slot.can_equip_to_slot(M, src, disable_warning, ignore_equipped))
```
NataKilar: Worth noting that ignoring the ``can_equip_to_slot`` check if ``force`` is true is a change in behavior. It'll allow you to do things like equip belts and IDs without a jumpsuit, rather than just ignoring if an item is already equipped in the slot.

out-of-phaze: I feel like things that want to ignore the equipped item/covering check but not the can_equip_to_slot check shouldn't be using force, honestly.

MistakeNot4892: can_equip_to_slot is also responsible for checking for things like missing organs and whether or not the item has the appropriate slot flags, this probably is not a wise change.

MistakeNot4892 (approved):

PR #3208, opened 2023-07-16
Staging runtime fixes.

MistakeNot4892:
Fixes several runtimes reported on staging.

### DIFF
--- a/code/modules/mob/living/carbon/taste.dm
+++ b/code/modules/mob/living/carbon/taste.dm
@@ -17,13 +17,17 @@
 catalogue the 'taste strength' of each one
 calculate text size per text.
 */
-/datum/reagents/proc/generate_taste_message(mob/living/carbon/taster = null, datum/reagents/source_holder = null)
-	var/minimum_percent = 15
-	if(ishuman(taster))
-		var/mob/living/carbon/human/H = taster
+/datum/reagents/proc/generate_taste_message(mob/living/taster, datum/reagents/source_holder)
+
+	var/minimum_percent
+	var/decl/species/my_species = istype(taster) && taster?.get_species()
+	if(my_species && !taster.isSynthetic())
 		// the taste bonus makes it so that sipping a drink lets you taste it better
-		var/taste_bonus = clamp(1 + (1/total_volume), 1, 2) // 1u makes it 2x as strong, 2u makes it 1.5x, 5u makes it 1.2x, etc
-		minimum_percent = round(15/ (H.isSynthetic() ? TASTE_DULL : H.species.taste_sensitivity * taste_bonus))
+		// 1u makes it 2x as strong, 2u makes it 1.5x, 5u makes it 1.2x, etc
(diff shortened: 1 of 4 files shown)


code/modules/mob/living/carbon/taste.dm
out-of-phaze, line 1: Minor functional difference here: beforehand, minimum_percent for a nonhuman mob would be 15. Now, it's 30. The taste sensitivity for nonhuman mobs has functionally been halved.

code/modules/projectiles/guns/energy/nuclear.dm
```diff
@@ -54,12 +54,10 @@
...
 	. = ..()
-	var/list/new_overlays = list()
-	var/reactor_icon = fail_counter ? "danger" : "clean"
-	new_overlays += mutable_appearance(icon, "[get_world_inventory_state()]_[reactor_icon]")
-	var/datum/firemode/current_mode = firemodes[sel_mode]
-	new_overlays += mutable_appearance(icon, "[get_world_inventory_state()]_[current_mode.name]")
-	overlays += new_overlays
+	add_overlay("[get_world_inventory_state()]_[fail_counter ? "danger" : "clean"]")
+	var/datum/firemode/current_mode = LAZYACCESS(firemodes, sel_mode)
+	if(current_mode)
```
out-of-phaze, line 59: I believe as discussed in several issues here, the issue is update_icon() for nuclear guns getting called prior to `/obj/item/gun/Initialize()` initializing the firemode.
image
The root cause here is that guns can update their icon prior to firemode init, which should probably be addressed instead of/in addition to fixing it here. (I'd just move the firemode init prior to the parent call.)

out-of-phaze: Fix is in https://github.com/NebulaSS13/Nebula/pull/3209. This commit and the one after it are probably no longer needed, and I don't even think this one would have fixed it anyway since `current_mode` isn't checked to see if it's a firemode datum instead of a list.

out-of-phaze (changes requested):
Some of the `?.` checks in `Sleeper.dm` aren't really necessary but I don't care much about them. The first two commits should probably be dropped in favor of #3209, since I don't think they would've done much to help the issues to begin with.

code/modules/mob/living/carbon/taste.dm
MistakeNot4892: I don't understand where 30 is coming from. It's initialized to 15 then overwritten in the previous code.

code/modules/projectiles/guns/energy/nuclear.dm
MistakeNot4892: The main issue this commit is fixing is adv egun not using SSoverlays calls. It's unrelated to the firemode changes, I was just in the region.

out-of-phaze: The crash check commit is still unnecessary, since we know what caused the issue and have a fix for it—I believe I pointed the issue out back in March when my OpenDream stuff found it.

MistakeNot4892:
This has been split up into other PRs.

Issue #2747, opened 2022-10-20
Code issues: Uses of initial on lists


PsyCommando:
#### Description of issue
Here are a bunch of initial() call on lists which return garbage non-null values because lists are apparently unsupported by initial.

code\controllers\subsystems\initialization\codex.dm:30
```
if(initial(centry.name) || initial(centry.associated_paths) || initial(centry.associated_strings))
```

code\modules\clothing\spacesuits\rig\rig_wiring.dm:32
```dm
rig.req_access = initial(rig.req_access)
```

code\modules\shuttles\shuttle_emergency.dm:118
```dm
authorized = initial(authorized)
```

code\modules\admin\verbs\massmodvar.dm:116
```dm
O.vars[variable] = initial(O.vars[variable])
```
It doesn't check if the initial variable value is a list, object instance, or anything that can't be used with initial.

code\modules\admin\verbs\modifyvariables.dm:228
```dm
new_var = initial(variable)
```
Same as above.

code\modules\admin\view_variables\helpers.dm:107
```dm
/datum/proc/get_initial_variable_value(varname)
	return initial(vars[varname])
```
Same as above.

MistakeNot4892:
VV ones are a bit w/e, no clean way to identify and warn against those.

Kapu1178:
The technical explanation of why lists aren't supported by initial, is you need to remember lists are an object type in byond.

`initial()` returns `null` for all variables whose defined value is an object instance. This is because it's compile time value is actually `null`, and the hidden internal proc called `init()` creates the instances.

PsyCommando:
> The technical explanation of why lists aren't supported by initial, is you need to remember lists are an object type in byond.
> 
> `initial()` returns `null` for all variables whose defined value is an object instance. This is because it's compile time value is actually `null`, and the hidden internal proc called `init()` creates the instances.

It  doesn't return null actually. It returns a garbage value that doesn't cast to null.

MistakeNot4892:
From what I can tell the fixable entries in this issue have been fixed. The VV handlers are not something easily updated to handle list values without a broader feature PR.

Specific individual initial() on lists should probably be given their own issues. I'm going to close this one for now as I don't think it serves much of a purpose being open indefinitely with the main body of it resolved.

PR #3211, opened 2023-07-17
Cherrypick select fixes from OpenDream compatibility PR

out-of-phaze:
## Description of changes
Several of the changes from #3012, without any reordering or JSON stuff. See commit names for information.

A lot of these are issues even in BYOND, so fixing them would be good even while my OpenDream stuff is on hold.

### DIFF
--- a/code/modules/admin/verbs/modifyvariables.dm
+++ b/code/modules/admin/verbs/modifyvariables.dm
@@ -224,13 +224,6 @@
 		if("list")
 			mod_list(variable, O, original_name, objectvar)
 
-		if("restore to default")
-			new_var = initial(variable)
-			if(assoc)
-				L[assoc_key] = new_var
-			else
-				L[L.Find(variable)] = new_var
-
 		if("edit referenced object")
 			modify_variables(variable)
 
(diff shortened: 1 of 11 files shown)


code/modules/admin/verbs/modifyvariables.dm
```diff
@@ -224,13 +224,6 @@
 		if("list")
 			mod_list(variable, O, original_name, objectvar)
 
-		if("restore to default")
```
MistakeNot4892, line 227: I don't see this mentioned, what's the go?

mods/content/psionics/datum/armour.dm
```diff
@@ -1,6 +1,7 @@
+/*
 /get_armor_key(damage_type, damage_flags)
```
MistakeNot4892: This isn't really a fix if it's just commenting it out. Should probably be deleted, or move get_armor_key onto a subsystem.

code/modules/admin/verbs/modifyvariables.dm
out-of-phaze: This is the VV dropdown for modifying lists (`mod_list()`). You can't do initial() on list contents, so this option just leads to confusion and errors. #2747

MistakeNot4892 (approved):

PR #3212, opened 2023-07-17
Added a divide by zero check and tidied code in taste calc.

MistakeNot4892:
- Generalized the taste logic to allow for non-carbon mobs tasting food.
- Added a divide by zero prevention max().


out-of-phaze (approved):
Addressed my concerns from the previous PR and cleans it up nicely. Maybe the only thing I'd change is putting the `my_species` definition before the if/else chain, then doing `if (taster.isSynthetic()) else if (my_species) else` all at the same indentation level, but really it's stylistic.

PR #3232, opened 2023-07-18
Reworking pulse procs.

MistakeNot4892:
## Description of changes
- Renamed get_pulse() to get_pulse_as_string().
- Renamed pulse() to get_pulse().
- Moved pulse procs down to /mob/living.
- May also fix issues with mobs having pulses after death by explicitly checking stat.

Makes it easier to generalize code.

## Authorship
Myself.

## Changelog
Nothing player-facing.

### DIFF
--- a/code/game/objects/items/devices/scanners/health.dm
+++ b/code/game/objects/items/devices/scanners/health.dm
@@ -117,13 +117,13 @@
 		if(H.status_flags & FAKEDEATH)
 			pulse_result = 0
 		else
-			pulse_result = H.get_pulse(GETPULSE_TOOL)
+			pulse_result = H.get_pulse_as_string(GETPULSE_TOOL)
 		pulse_result = "[pulse_result]bpm"
-		if(H.pulse() == PULSE_NONE)
+		if(H.get_pulse() == PULSE_NONE)
 			pulse_result = "<span class='scan_danger'>[pulse_result]</span>"
-		else if(H.pulse() < PULSE_NORM)
+		else if(H.get_pulse() < PULSE_NORM)
 			pulse_result = "<span class='scan_notice'>[pulse_result]</span>"
-		else if(H.pulse() > PULSE_NORM)
+		else if(H.get_pulse() > PULSE_NORM)
 			pulse_result = "<span class='scan_warning'>[pulse_result]</span>"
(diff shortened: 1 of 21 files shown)


code/modules/assembly/assembly.dm
```diff
@@ -40,7 +40,7 @@
 /obj/item/assembly/proc/pulsed(var/radio = 0)						//Called when another assembly acts on this one, var/radio will determine where it came from for wire calcs
 	return
 
-/obj/item/assembly/proc/pulse(var/radio = 0)						//Called when this device attempts to act on another device, var/radio determines if it was sent via radio or direct
+/obj/item/assembly/proc/get_pulse(var/radio = 0)						//Called when this device attempts to act on another device, var/radio determines if it was sent via radio or direct
```
comma: looks like mass rename artefact, in rest of assembly code too

comma (changes requested):

code/modules/assembly/assembly.dm
MistakeNot4892: Oh god dammit.

comma (approved):

code/game/objects/items/devices/scanners/health.dm
```diff
@@ -117,13 +117,13 @@
 		if(H.status_flags & FAKEDEATH)
 			pulse_result = 0
 		else
-			pulse_result = H.get_pulse(GETPULSE_TOOL)
+			pulse_result = H.get_pulse_as_string(GETPULSE_TOOL)
```
out-of-phaze, line 120: See line 121; `"extremely weak and fast, patient's artery feels like a threadbpm"` is invalid, `pulse_result` should maybe use `get_pulse_as_number()` instead.
```suggestion
			pulse_result = H.get_pulse_as_number()
```
Or modify/remove line 121 instead, since it *should* always be overwritten anyway.

out-of-phaze: Nevermind, `GETPULSE_TOOL` is the more accurate one (terrible name) so it'll output `>250bpm` which is valid. I still don't like that but it's not enough to block over.

out-of-phaze (approved):

PR #3236, opened 2023-07-18
Nutrition and hydration adjustments.

MistakeNot4892:
## Description of changes
- Adds procs and vars for handling nutrition and hydration at /mob/living.
- Opts out of this for simple mobs and silicons for the time being.

Allows for unifying a bunch of shared logic.

## Authorship
Myself.

## Changelog
Nothing player-facing.

### DIFF
--- a/mods/content/xenobiology/slime/_slime.dm
+++ b/mods/content/xenobiology/slime/_slime.dm
@@ -22,13 +22,13 @@
 	bone_amount = 0
 	ai = /datum/ai/slime
 	hud_type = /datum/hud/slime
+	nutrition = 800
 
 	var/is_adult = FALSE
 	var/mutation_chance = 30 // Chance of mutating, should be between 25 and 35
 	var/powerlevel = 0 // 0-10 controls how much electricity they are generating
 	var/amount_grown = 0 // controls how long the slime has been overfed, if 10, grows or reproduces
 	var/weakref/feeding_on
-	var/nutrition = 800
 	var/toxloss = 0
 	var/hurt_temperature = T0C-50 // slime keeps taking damage when its bodytemperature is below this
 	var/die_temperature = 50 // slime dies instantly when its bodytemperature is below this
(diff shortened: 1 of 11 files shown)


mods/content/xenobiology/slime/life.dm
```diff
@@ -80,9 +80,15 @@
...
+/mob/living/slime/get_hunger_factor()
+	return (0.1 + 0.05 * is_adult)
+
+/mob/living/slime/get_thirst_factor()
+	return 0
+
 /mob/living/slime/handle_nutrition_and_hydration()
 
-	adjust_nutrition(-(0.1 + 0.05 * is_adult))
+	..()
```
out-of-phaze: Why was this moved prior to digesting mobs in contents?

mods/content/xenobiology/slime/_slime.dm
```diff
@@ -292,6 +292,16 @@
 /mob/living/slime/check_has_mouth()
 	return 0
 
+/mob/living/slime/set_nutrition(amt)
+	..()
+	if(get_hydration() < amt)
+		set_hydration(amt)
+
+/mob/living/slime/set_hydration(amt)
```
out-of-phaze: Why not just set the hydration rate to zero and opt out of hydration altogether? Doing `nutrition = max(nutrition, hydration); hydration = max(hydration, nutrition)` is a bit confusing and I'm not sure what it's meant to solve.

out-of-phaze (commented):
A few clarifying questions, then I'll re-review.

mods/content/xenobiology/slime/_slime.dm
MistakeNot4892: It chains the values to each other as I don't think slimes have a way to get hydration (since water kills them) but the bodytemp code in the other PR will subtract hydration to stabilize bodytype. Chaining them to each other seems like a better solution than always returning max

mods/content/xenobiology/slime/life.dm
MistakeNot4892: Didn't seem important, I'll move it beneath.

mods/content/xenobiology/slime/_slime.dm
out-of-phaze: Is this chaining, though? I feel like this could lead to weird edge cases given the conditional.

MistakeNot4892: Replaced with just returning `get_nutrition()` for `get_hydration()`.

out-of-phaze (approved):

PR #3241, opened 2023-07-20
Job descriptions are now displayed on spawn

ghost:
## Description of changes
The `description` gets shot at the player after the `You are the...` message on spawn. It's previously only been available in the job's `?` tab. Also adds description to the example map's Tourist job just for testing.

I feel like this either was implemented previously or on another server, but either way it allows for basic tips about the job to be easier forwarded to players. I'm also using this downstream.

## Authorship
10sc

## Changelog
:cl: 10sc
add: Job descriptions are now displayed on spawn.
/:cl:


MistakeNot4892 (approved):

LowFP:
This is a cool change

PR #3244, opened 2023-07-22
Fixing/implementing mouth grippers.

MistakeNot4892:
## Description of changes
This probably should target dev, but I want it for Scav and it does fix some existing unfinished/broken code.
- Moves the upper hand defines out of the Ascent modpack into general code.
- Sorts hand slots by a new priority var for nicer display.
- Adds a UI icon for mouth slot.
- Adds per-gripper dexterity (will need to be resolved with #3112 when that goes in)
- Adds some inventory slot organ checking to disarm and dexterity code.

Better inventory slot and organ integration, allows extra hand slots to be used more easily.

## Authorship
Myself.

## Changelog


code/datums/inventory_slots/inventory_gripper.dm
```diff
@@ -19,18 +24,31 @@
...
-	requires_organ_tag = null
+	requires_organ_tag = BP_HEAD
 	can_use_held_item = FALSE
+	overlay_slot = BP_MOUTH
+	ui_label = "M"
+	hand_sort_priority = 3
+
+/datum/inventory_slot/gripper/mouth/get_examined_string(mob/owner, mob/user, distance, hideflags, decl/pronouns/pronouns)
+	if(_holding)
+		return "[pronouns.He] [pronouns.is] holding [_holding.get_examine_line()] in [pronouns.his] mouth."
```
out-of-phaze: Is this override necessary? `E?.name || lowertext(slot_name)` should work fine even with no 'mouth' organ, and `lowertext(slot_name)` would certainly work without hardcoding `mouth`.

code/modules/mob/living/say.dm
```diff
@@ -164,9 +164,11 @@
 		speaking.broadcast(src,trim(message))
 		return 1
 
-	if((is_muzzled()) && !(speaking && (speaking.flags & LANG_FLAG_SIGNLANG)))
-		to_chat(src, "<span class='danger'>You're muzzled and cannot speak!</span>")
-		return
+	if(!speaking || !(speaking.flags & LANG_FLAG_SIGNLANG))
+		var/obj/item/muzzled = is_muzzled()
```
out-of-phaze: I don't know if I'm okay with an `is_\w+` proc returning a datum rather than `FALSE`/`TRUE`. I'd probably use something like `get_item_blocking_mouth()` or something, but it's a bit wordy. Up to you though.

code/modules/mob/say.dm
```diff
@@ -109,8 +109,13 @@ var/global/list/special_channel_keys = list(
...
 
 /mob/proc/is_muzzled()
-	var/obj/item/mask = get_equipped_item(slot_wear_mask_str)
-	return istype(mask, /obj/item/clothing/mask/muzzle) || istype(mask, /obj/item/clothing/sealant)
+	// Can't talk with something in your mouth.
+	var/datum/inventory_slot/mouth_slot = get_inventory_slot_datum(BP_MOUTH)
+	. = mouth_slot?.get_equipped_item()
+	if(!.)
+		var/obj/item/mask = get_equipped_item(slot_wear_mask_str)
+		if(istype(mask, /obj/item/clothing/mask/muzzle) || istype(mask, /obj/item/clothing/sealant))
```
out-of-phaze, line 120: Out of scope, but this should probably be generalized somehow...? I'm not sure 100% how that'd be done in a way that could be reused elsewhere—add vars or flags to mark clothing as tightly-fit, and to mark them as blocking air, then if both are true *and* it covers the mouth you can't speak through it? Way off in the weeds here, though.

code/datums/inventory_slots/inventory_gripper.dm
MistakeNot4892: It's necessary due to the grammar, the base proc uses 'wearing x on their y', not 'in'.

code/modules/mob/living/say.dm
MistakeNot4892: Eh that also works yeah.

code/modules/mob/say.dm
MistakeNot4892: Could possibly just be an item flag, ITEM_FLAG_BLOCKS_SPEECH or such.

code/datums/inventory_slots/inventory_gripper.dm
out-of-phaze: For grippers:
`return "[pronouns.He] [pronouns.is] holding [_holding.get_examine_line()] in [pronouns.his] [E?.name || lowertext(slot_name)]."`

MistakeNot4892: Per Discord I misremembered the specific issue, it was due to my original code producing 'he is carrying the spare ID in his head'. Shouldn't be needed now, snipping it.

code/modules/mob/say.dm
MistakeNot4892: Noting as a side-thought that we probably need some kind of species or bodytype support for 'speaks from this organ' for species like dionaea (though they specifically aren't on Neb as humanoids, so)

out-of-phaze (approved):

PR #3149, opened 2023-06-17
Unify bodytype/prosthetics, move species vars onto bodytype, remove carbon/var/bodytype

out-of-phaze:
## Description of changes
Spiritual successor to #2610, with a small subset of those changes.
Makes prosthetics manufacturers a subtype of bodytype.
Removes `model` on external organs, replaces it with bodytype.
Moves `slowdown` and `appearance_flags` and the various `base_color` vars from species onto bodytype.
Fixes Vox prosthetics aspect showing up on all species.
Fixes the layout of the aspect selection menu.
Moves organs and limbs lists/variables onto bodytypes.
Moves default hairstyles, default facial hair styles, and base markings onto bodytypes.
Moves `SPECIES_FLAG_NO_SCAN` and `SPECIES_FLAG_NO_PAIN` onto bodytype.
Fixes the old Deity species since removing them would be too complicated without removing all of Deity.
Adds material and matter vars to bodytypes.
Adds `apply_bodytype_organ_modifications`.
Removes `robotize`, folds it into `set_bodytype` and `apply_bodytype_organ_modifications`.
Removes `apply_species_organ_modifications`, since its functionality was either redundant with new changes or could be moved onto `apply_bodytype_organ_modifications`.
Removes the bodytype var on carbon mobs. `get_bodytype()` should be used instead and will return the bodytype of the root organ of the mob, currently always the torso. In the future this could default to whatever organ in the bodytype is parentless (also, should probably lint that bodytypes always have exactly one parentless organ.)

TODO in a future PR:
- The rest of the changes in #2610.

## Why and what will this PR improve
The start of my quest to abolish `IsSynthetic()`, and part of the 'bag of organs' project.

PR #3246, opened 2023-07-22
Moving eye icons and vars onto bodytype.

MistakeNot4892:
## Description of changes
- Moves a bunch of eye vars onto bodytype.
- Splits BODY_FLAG_NO_DNA into a couple of other flags for defib and stasis purposes.
- Allows vox to be defibbed and put in stasis.

Removes need for weird unwieldly eye subtypes.

## Authorship
Myself.

## Changelog
Nothing user-facing.

### DIFF
--- a/code/modules/organs/prosthetics/prosthetics_manufacturer.dm
+++ b/code/modules/organs/prosthetics/prosthetics_manufacturer.dm
@@ -7,6 +7,8 @@
 	is_robotic = TRUE
 	body_flags = BODY_FLAG_NO_DNA | BODY_FLAG_NO_DEFIB | BODY_FLAG_NO_STASIS | BODY_FLAG_NO_PAIN | BODY_FLAG_NO_EAT
 	material = /decl/material/solid/metal/steel
+	eye_flash_mod = 1
+	eye_darksight_range = 2
 	/// Determines which bodyparts can use this limb.
 	var/list/applies_to_part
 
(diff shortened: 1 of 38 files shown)


out-of-phaze (commented):
~~This might need to wait until after #3149, because currently bodytype is never even set on internal organs, IIRC? It's always the human bodytype no matter what.~~

Nevermind, I was thinking of `model`. Still, this redoes a couple changes I do in that PR and the inevitable conflicts have me concerned...

code/modules/organs/prosthetics/prosthetics_manufacturer.dm
```diff
@@ -5,7 +5,7 @@
 	limb_tech = "{'engineering':1,'materials':1,'magnets':1}"
 	modifier_string = "robotic"
 	is_robotic = TRUE
-	body_flags = BODY_FLAG_NO_DNA | BODY_FLAG_NO_PAIN | BODY_FLAG_NO_EAT
+	body_flags = BODY_FLAG_NO_DNA | BODY_FLAG_NO_DEFIB | BODY_FLAG_NO_STASIS | BODY_FLAG_NO_PAIN | BODY_FLAG_NO_EAT
 	material = /decl/material/solid/metal/steel
 	/// Determines which bodyparts can use this limb.
```
out-of-phaze, line 12: Prosthetic eyes had this set in eyes.dm, it should probably be moved to here.
```dm
		flash_mod = 1
		darksight_range = 2
```
Given that prosthetic eyes would overwrite species darksight (and maybe rightfully so?), I feel like maybe we should include this still as some sort of compensation.

code/modules/species/outsider/random.dm
```diff
@@ -15,6 +15,11 @@
 	base_color  = RANDOM_RGB
 	. = ..()
 
+/decl/bodytype/alium/Initialize()
```
out-of-phaze: `/decl/bodytype/alium/Initialize()` already exists directly above this, please combine these overrides.

out-of-phaze (changes requested):

out-of-phaze (approved):

PR #3249, opened 2023-07-22
Removing outdated verbiage from paramount coercion and borer antags.

MistakeNot4892:
## Description of changes
- Replaces 'slave' and 'thrall' with 'beguiled' or 'host'.
- Rewrites the coercion messages to be more appropriate to the new concept.
- Adds a confirmation prompt and punishment for refusal to paramount coercion.

Had a lengthy discussion about this with some folks on Polaris who were uncomfortable with it conceptually and in terms of the language used, as well as the issues with conversion modes in terms of agency/consent/not really being fun if you didn't agree to them. 

Additionally from a purely pragmatic perspective if you convert someone who doesn't want to be converted they'll take the first opportunity to leave, if they don't ghost immediately.

The language in question is not something I think needs justification, it hasn't been appropriate to casually use 'slave' for years. Thrall is a bit less problematic in some senses but is close enough that I just trimmed it out as well.

## Changelog
:cl:
tweak: Paramount Coercion is now 'Beguile', and does not have a channel or a risk of backblast. It also prompts the target to join willingly, knocking them unconscious and dealing brain damage so you can escape if they refuse.
/:cl:

### DIFF
--- a/code/game/antagonist/antagonist_add.dm
+++ b/code/game/antagonist/antagonist_add.dm
@@ -48,13 +48,13 @@
 
 /decl/special_role/proc/add_antagonist_mind(var/datum/mind/player, var/ignore_role, var/nonstandard_role_type, var/nonstandard_role_msg)
 	if(!istype(player))
-		return 0
+		return FALSE
 	if(!player.current)
-		return 0
+		return FALSE
 	if(player in current_antagonists)
-		return 0
+		return FALSE
 	if(!can_become_antag(player, ignore_role))
-		return 0
+		return FALSE
 	current_antagonists |= player
 
 	if(faction_verb)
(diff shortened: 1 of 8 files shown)


out-of-phaze:
I'm not opposed to improving the language used in the game—"slave-mind" and "mindslave" etc. are pretty egregious, I think—but I disagree with the choice to replace the word thrall, given it has established connotations in the context of vampires, and if you're not a history nerd you'd have no clue it's etymologically related to a Norse word for slave/servant.

For borers, I do think changing it to "host" makes more sense, at least, especially since symbiotes (sometimes) exist.

As for the changes to the beguile ability, I don't really like the idea of a non-coercive coercion-tree ability, but forced conversion antags just generally don't work so I think it's acceptable.

MistakeNot4892:
It's not non-coercive just because it's not violent, though.

I was unaware of thrall's meaning and associated it with vampires too but them's the breaks. If we ever want to add literal vampires we could discuss it again, but there isn't a real need to use it here.

out-of-phaze (approved):
So I was under the impression we *had* vampires. My bad. As per the discussion on Discord I think this is probably fine, although a little fantasy-y.

PR #3250, opened 2023-07-23
Porting renderers and emissive blockers from Bay.

MistakeNot4892:
WIP, DNM, proper PR body with credit to PRs to come.


code/game/atoms_init.dm
```diff
@@ -107,6 +108,13 @@
 	if (!ml && loc)
 		loc.Entered(src, null)
 
+	// Handle emissive blockers here to avoid an override in atoms_movable.dm.
+	var/emissive_block = update_emissive_blocker()
+	if(emissive_block)
+		overlays += emissive_block
```
MistakeNot4892: I think this probably needs to be a priority overlay instead?
@Lohikar would appreciate your input into the `managed_overlays` stuff.

Lohikar: Hard no on managed overlays. This should be done with priority overlays.

Lohikar: This is also going to require every movable in the game be mangled, mind.

MistakeNot4892:
This is beyond me currently, will reopen someday down the track if I manage to get my head around render targets and filters.

PR #3251, opened 2023-07-23
Converting mouse_opacity values to defines.

MistakeNot4892:
## Description of changes
Changes out magic numbers for defines in mouse opacity setting.

More obvious values, cleaner code.

## Authorship
CrimsonShrike on Bay I assume, this has come along as part of the work towards porting emissives.

## Changelog
Nothing player-facing.

### DIFF
--- a/code/__defines/misc.dm
+++ b/code/__defines/misc.dm
@@ -300,3 +300,10 @@
 
 //Damage stuff
 #define ITEM_HEALTH_NO_DAMAGE -1
+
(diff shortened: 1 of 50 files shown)


out-of-phaze (dismissed):
It might(?) be worth also adding grep checks to CI for `mouse_opacity\s*=\s*[0-2]` to at least catch any trivial cases where the defines should be used instead.

out-of-phaze (approved):

PR #3252, opened 2023-07-23
Moved PIXEL_SCALE into a DEFAULT_APPEARANCE_FLAGS var.

MistakeNot4892:
## Description of changes
- Adds `DEFAULT_APPEARANCE_FLAGS` currently containing only `PIXEL_SCALE`.
- Sets this by default on several types (atom, atom/movable, image, mutable_appearance).
- Removes direct setting in favour of |= in cases that previously included PIXEL_SCALE.

Part of the work towards porting Bay emissives, also generally cleaner code.

## Authorship
Unsure, possibly CrimsonShrike, this is part of the port chain.

## Changelog
Nothing player-facing ideally.


code/game/objects/structures/rubble.dm
```diff
@@ -35,7 +35,6 @@
 				I.color = initial(A.color)
 			if(!lootleft)
 				I.color = "#54362e"
-		I.appearance_flags = PIXEL_SCALE
```
out-of-phaze, line 38: This line served to remove `TILE_BOUND` and should remain here.
```dm
		I.appearance_flags &= ~TILE_BOUND
```

out-of-phaze (changes requested):
Comment: in the future (515+) I think a number of these could be replaced with type-relative definitions, like:
```
    appearance_flags = parent_type::appearance_flags | KEEP_TOGETHER & ~TILE_BOUND
```
which prevents having to re-specify all the parent's flags.

code/game/objects/structures/rubble.dm
MistakeNot4892: `I` is a fresh image instance and never had `TILE_BOUND` set.

out-of-phaze (dismissed):

out-of-phaze (approved):

PR #3256, opened 2023-07-24
Max health and updatehealth() rewrite.

MistakeNot4892:
## Description of changes
- `updatehealth` is now `update_health`.
- `update_health` removed from a bunch of locations and folded into a `do_update_health` arg to appropriate damage setters.
- Other redundant `update_health` calls removed.
- `maxHealth` on blobs is now `blob_max_health`.
- `maxHealth` on mobs is now `mob_default_max_health`.
- Added a `get_max_health()` getter for retrieving max health. 
- Condenses a lot of death checking into `updatehealth()` and new procs (`get_total_life_damage()` and `should_be_dead()`) and cleans up death() calls outside of `updatehealth()`.

Centralizes health updates, less unnecessary/redundant calls, will allow for aspects etc to modify maxhealth down the track.

## Authorship
Myself.

## Changelog
Nothing player-facing.


MistakeNot4892:
Clobbered a ton of my own changes to this, rip.

PsyCommando:
Those are mob changes right? Because I got the same thing going for /obj?

MistakeNot4892:
Yeah this is specifically for mobs. It should make it much easier to unify down the track.

MistakeNot4892:
This spiralled a bit out of control, but should be good to review now.

MistakeNot4892:
These CI fails are baffling. Wondering if I have an infinite loop in my code somewhere.

MistakeNot4892:
Honestly I am expecting this to cause a lot of bugs, but I don't think we're going to isolate them in review due to how labyrinthine these systems are. Fixing them piecemeal as they come up seems like a more effective approach to me.

code/modules/mob/living/carbon/brain/brain.dm
```diff
@@ -9,6 +9,9 @@
 	icon_state = "brain1"
 	mob_sort_value = 7
 
+/mob/living/carbon/brain/can_emote()
+	return stat == CONSCIOUS && istype(loc, /obj/item/mmi)
```
out-of-phaze: Will this cause issues for brainmobs in posibrains, or are those still a subtype of MMI at this point?

code/modules/mob/living/carbon/brain/life.dm
```diff
@@ -72,77 +63,73 @@
...
+			if(2 to 9)//Low level of EMP damage, has few effects(handled elsewhere)
+				if(!alert)
+					emote("notice")
+					to_chat(src, "<span class='warning'>System reboot nearly complete.</span>")
+					alert = 1
+				if(prob(25))
 					emp_damage -= 1
-
+			if(1)
+				alert = 0
```
out-of-phaze:
```suggestion
				alert = FALSE
```

code/modules/mob/living/carbon/brain/life.dm
```diff
@@ -72,77 +63,73 @@
...
+			if(31 to INFINITY)
+				emp_damage = 30//Let's not overdo it
+			if(21 to 30)//High level of EMP damage, unable to see, hear, or speak
+				set_status(STAT_BLIND, 1)
+				SET_STATUS_MAX(src, STAT_DEAF, 1)
+				set_status(STAT_SILENCE, 1)
+				if(!alert)//Sounds an alarm, but only once per 'level'
+					emote("alarm")
+					to_chat(src, "<span class='warning'>Major electrical distruption detected: System rebooting.</span>")
+					alert = 1
```
out-of-phaze:
```suggestion
					alert = TRUE
```
Same goes for other uses of `alert = 1` in here.

code/modules/mob/living/life.dm
```diff
@@ -183,9 +183,6 @@
 			if(LAZYACCESS(chem_doses, T) <= 0)
 				LAZYREMOVE(chem_doses, T)
 
-	if(apply_chemical_effects())
-		updatehealth()
```
out-of-phaze, line 187: This... is necessary to make sure that `CE_REGEN_BURN` and `CE_REGEN_BRUTE` cause fewer repeated update_health calls, no? Now `heal_organ_damage()` calls it directly, meaning other overrides will build onto that—for example, downstream I have a couple extra healing chem effects on top of these two. The check is to make sure update_health() is delayed until `apply_chemical_effects()` is over.

out-of-phaze (changes requested):
Making health not update by default after damage feels like a major footgun waiting to happen for new contributors... I had to deal with issues related to that on Sunset in chem code. Is it really necessary?

EDIT: As discussed on Discord, it should probably be changed to update by default.

out-of-phaze (approved):

PR #3271, opened 2023-08-12
Holomap fixes

MistakeNot4892:
- Holomaps don't runtime error on qdel or reinit.
- Holomap cursors work first go instead of only after a level change.
Thanks @out-of-phaze for identifying the cursor issue and a fix.

### DIFF
--- a/code/controllers/subsystems/holomap.dm
+++ b/code/controllers/subsystems/holomap.dm
@@ -39,16 +39,11 @@ SUBSYSTEM_DEF(minimap)
 			single.Blend(A.holomap_color, ICON_MULTIPLY)
 		combinedareas.Blend(single, ICON_OVERLAY)
 
-	var/icon/map_base = icon(data.holomap_base)
-
-	// Generate the full sized map by blending the base and areas onto the backdrop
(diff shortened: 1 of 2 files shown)


code/controllers/subsystems/holomap.dm
```diff
@@ -41,15 +39,12 @@ SUBSYSTEM_DEF(minimap)
 			single.Blend(A.holomap_color, ICON_MULTIPLY)
 		combinedareas.Blend(single, ICON_OVERLAY)
 
-	data.holomap_areas_combined = combinedareas
-
 	var/icon/map_base = icon(data.holomap_base)
 
 	// Generate the full sized map by blending the base and areas onto the backdrop
 	var/icon/big_map = icon(HOLOMAP_ICON, "stationmap")
```
out-of-phaze: This variable is now unused also, which means we're wasting two 480x480 blend ops.

out-of-phaze (changes requested):

MistakeNot4892:
Rebased onto dev rip.

out-of-phaze (approved):

PR #3278, opened 2023-08-16
misc changes

wd440:
legitimate salvage (for now)
tidies up guncode, adds gun boxes, designs, etc
misc changes in general
also touches prices a bit
also needs a merge (by merge I mean merge to this branch cuz its outdaaated)
but i'm opening it anyway (feel free to close)
also feel free to take out any changes etc etc whatever if I'm going to return to this its going to be a long time
im leaving this there to come back later and feel "complete" right now


MistakeNot4892:
I appreciate your work ethic, but PRs need to be specific and realistic. Commits should also be named appropriately and ideally should encapsulate a single change or set of changes. I can't see this PR fitting those criteria. If you would like to PR the individual gun changes, I would suggest breaking them off into their own commits and PRing them separately, with an informative name/body. 

With the above in mind, I'm going to close this one.

PR #3281, opened 2023-08-17
Cleaning up some old emote code.

MistakeNot4892:
## Description of changes
Removes `use_me`, merges a few things into `can_emote()` including the failure message.

Nicer, more consistent code.

## Authorship
Myself.

## Changelog
Nothing player-facing.

### DIFF
--- a/code/modules/emotes/emote_mob.dm
+++ b/code/modules/emotes/emote_mob.dm
@@ -1,8 +1,17 @@
 /mob/proc/can_emote(var/emote_type)
+	. = check_mob_can_emote(emote_type)
+	if(!.)
+		to_chat(src, SPAN_WARNING("You cannot currently [emote_type == AUDIBLE_MESSAGE ? "audibly" : "visually"] emote!"))
+
+/mob/proc/check_mob_can_emote(var/emote_type)
+	SHOULD_CALL_PARENT(TRUE)
 	return (stat == CONSCIOUS)
 
-/mob/living/can_emote(var/emote_type)
-	return (..() && !(HAS_STATUS(src, STAT_SILENCE) && emote_type == AUDIBLE_MESSAGE))
+/mob/living/check_mob_can_emote(var/emote_type)
+	return ..() && !(HAS_STATUS(src, STAT_SILENCE) && emote_type == AUDIBLE_MESSAGE)
+
+/mob/living/carbon/brain/check_mob_can_emote(var/emote_type)
+	return ..() && istype(loc, /obj/item/mmi)
 
 /mob/proc/emote(var/act, var/m_type, var/message)
 	set waitfor = FALSE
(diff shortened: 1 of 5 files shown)


comma (approved):

code/modules/emotes/emote_mob.dm
```diff
@@ -1,8 +1,17 @@
...
+	SHOULD_CALL_PARENT(TRUE)
 	return (stat == CONSCIOUS)
 
-/mob/living/can_emote(var/emote_type)
-	return (..() && !(HAS_STATUS(src, STAT_SILENCE) && emote_type == AUDIBLE_MESSAGE))
+/mob/living/check_mob_can_emote(var/emote_type)
+	return ..() && !(HAS_STATUS(src, STAT_SILENCE) && emote_type == AUDIBLE_MESSAGE)
+
+/mob/living/carbon/brain/check_mob_can_emote(var/emote_type)
+	return ..() && istype(loc, /obj/item/mmi)
```
out-of-phaze, line 14: I kind of wonder if a better approach would be to just check isSynthetic on the synth emotes. That way, it'd work for posibrains *and* MMIs. The only issue is that currently isSynthetic only returns true for brains in a posibrain or digital MMI, but since isSynthetic is supposed to return true for FBPs, saying an MMI is synthetic seems fine.

out-of-phaze (changes requested):
I'm not a fan of the bespoke MMI loc check especially since it misses the case of posibrain organs, I think, which haven't been an MMI subtype for a long time. Either way I think we should use `isSynthetic` instead, but I'm undecided on whether or not the check should be moved to the beep-and-boop synth emotes (in which case we'd also have to make it handle humantype synths, silicons, etc.) or kept where it is on `/mob/living/carbon/brain/check_mob_can_emote`.

The rest of the changes in this PR seem fine.

code/modules/emotes/emote_mob.dm
MistakeNot4892: The point of this is to prevent a bare brain outside an MMI emoting, I'm not sure where the isSynthetic stuff comes into it?

MistakeNot4892:
> I'm not a fan of the bespoke MMI loc check especially since it misses the case of posibrain organs, I think, which haven't been an MMI subtype for a long time. Either way I think we should use `isSynthetic` instead, but I'm undecided on whether or not the check should be moved to the beep-and-boop synth emotes (in which case we'd also have to make it handle humantype synths, silicons, etc.) or kept where it is on `/mob/living/carbon/brain/check_mob_can_emote`.
> 
> The rest of the changes in this PR seem fine.

Posibrains don't use /mob/living/carbon/brain and have their own emote handling.

out-of-phaze:
> Posibrains don't use /mob/living/carbon/brain and have their own emote handling.

As discussed on Discord, nope. But we can rework this later, this is just cleanup.

out-of-phaze (approved):

PR #3285, opened 2023-08-20
Update Byond version SECURITY.md

AliceDTRH:
## Description of changes

I have updated the Byond version in SECURITY.md

The SECURITY.md says 
> If this version number is at odds with `BYOND_MAJOR`.`BYOND_MINOR` as defined in [the test workflow](https://github.com/NebulaSS13/Nebula/blob/dev/.github/workflows/test.yml#L11) in the .github workflow folder, the workflow configuration should be considered authoritative and this document should be noted as out of date.

I noticed they were at odds, so I made this PR.

## Authorship

@MistakeNot4892 and @afterthought2 seem to have been the ones that wrote the last version.

### DIFF
--- a/docs/SECURITY.md
+++ b/docs/SECURITY.md
@@ -2,7 +2,7 @@
 
 ## Supported Versions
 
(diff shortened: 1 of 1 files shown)


MistakeNot4892:
Thanks for catching that.

MistakeNot4892 (approved):

PR #3289, opened 2023-08-23
Converting human icon code to exist on the inventory slots instead.

MistakeNot4892:
## Description of changes
- Inventory slot `update_overlay` now becomes `update_mob_equipment_overlay`.
- `update_mob_equipment_overlay` can handle a null prop being passed in.
- `update_mob_equipment_overlay` is called by `refresh_visible_overlays` at `/mob/living` level.
- The various `update_inv_foo` procs have been rolled into their respective inventory slots which are now updated with `update_equipment_overlay`.
- `set/get_current_mob_over/underlay` added for the human `over/underlays_standing` updating.
- `overlays_standing` is now `mob_overlays`, `underlays_standing` is now `mob_underlays`. Direct use of these lists has been replaced with getters/setters pending a better layering solution.

Tested on local but I expect I will have missed `src` or not set slots in some places.

Centralizes inventory updates in a way that allows for many mobs to benefit from the logic, not just humans.
Allows me to do some cheating with inventory slots when replacing the hattable extension in the future.

## Authorship
Myself.

## Changelog
Nothing player-facing.

### DIFF
--- a/code/game/objects/item.dm
+++ b/code/game/objects/item.dm
@@ -183,7 +183,7 @@
 /obj/item/proc/update_held_icon()
 	if(ismob(src.loc))
 		var/mob/M = src.loc
-		M.update_inv_hands()
+		M.update_inhand_overlays()
 
 /obj/item/proc/is_held_twohanded(mob/living/M)
 	if(!M)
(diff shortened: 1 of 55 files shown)


code/game/objects/item.dm
```diff
@@ -806,9 +806,25 @@ modules/mob/living/carbon/human/life.dm if you die, you will be zoomed out.
 	update_icon()
 	update_clothing_icon()
 
+// Used to call appropriate slot updates in update_clothing_icon()
+/obj/item/proc/get_associated_equipment_slots()
+	return
```
out-of-phaze:
```suggestion
	if(item_flags & ITEM_FLAG_IS_BELT)
		return slot_belt_str
	return null
```
Maybe? Unsure if necessary.

out-of-phaze: If it is, this should probably be included in a parent call in all the others, since, for example, clipping a mask to your belt would make sense.

out-of-phaze (changes requested):

code/game/objects/item.dm
```diff
@@ -806,9 +806,27 @@ modules/mob/living/carbon/human/life.dm if you die, you will be zoomed out.
...
+	if(!equip_slots)
+		return FALSE
+	var/mob/wearer = loc
+	if(!istype(wearer))
+		return FALSE
+	if(islist(equip_slots))
+		for(var/slot in equip_slots)
+			wearer.update_equipment_overlay(slot, FALSE)
+		wearer.update_icon()
+	else
```
out-of-phaze: As far as I know, this `else` case can never be triggered now. Either `get_associated_equipment_slots` returns null or a list. Null is checked earlier.

out-of-phaze (changes requested):

code/game/objects/item.dm
MistakeNot4892: Good point.

MistakeNot4892: Removed the checking.

out-of-phaze (dismissed):

out-of-phaze (approved):

PR #3191, opened 2023-07-04
Sprite accessories rewrite.

MistakeNot4892:
## Description of changes
- Generalizes the marking system to include hair, lips and facial hair alongside markings in the organ marking list.
- Changes unathi to have frills and horns instead of hair and facial hair.

**TODO**:
- [X] Update the prefs panel.
- [x] Get saving and loading to work.
- [x] Make sure grandfathered loading works.
- ~~Consider converting eyes to markings.~~

## Why and what will this PR improve
Cleans up some old crusty code and potentially makes it easier to add new categories of sprite accessory (like wings or tails) in the future.

image

## Authorship
Myself.

## Changelog
Hopefully nothing player-facing.

PR #3293, opened 2023-08-27
Vox bodytype updates.

MistakeNot4892:
## Description of changes
- Finishes implementing servitor voxform (alt appearance).
- Adds alchemist voxform (aesthetic variation only).
- Adds a Set Bodytype helper to VV.
- Adds bodytype category checking to sprite accessories and renames ambiguous body flag checking.
- Adds basic icon validation to sprite accessories.
- Sets UIDs on the sprite accessories decls and enforces them.

- Pulls UID changes out of #3191.
- Somewhat more rigorous accessory testing, more debugging options for bodytypes.
- Vox servitor icons aren't just sitting in the codebase unused.

## Authorship
Myself.

## Changelog
:cl:
add: Vox now have two additional bodytypes.
/:cl:


code/modules/mob/new_player/preferences_setup.dm
```diff
@@ -1,13 +1,24 @@
-//The mob should have a gender you want before running this proc. Will run fine without H
+/datum/preferences/proc/random_hair_style()
+	var/decl/species/mob_species = get_species_by_key(species || global.using_map.default_species)
```
out-of-phaze:
```suggestion
	var/decl/species/mob_species = get_species_decl()
```

code/modules/mob/new_player/preferences_setup.dm
```diff
@@ -1,13 +1,24 @@
...
+	var/decl/bodytype/mob_bodytype = mob_species?.get_bodytype_by_name(bodytype) || mob_species?.default_bodytype || GET_DECL(global.using_map.default_bodytype)
```
out-of-phaze:
```suggestion
	var/decl/bodytype/mob_bodytype = get_bodytype_decl()
```
`using_map.default_bodytype` is deprecated as far as I know. IMO it should be removed and all uses of it replaced with the default species' default bodytype.

code/modules/mob/new_player/preferences_setup.dm
```diff
@@ -1,13 +1,24 @@
...
+	var/list/valid_styles = mob_species?.get_hair_style_types(mob_bodytype)
+	return length(valid_styles) ? pick(valid_styles) : /decl/sprite_accessory/hair/bald
+
+/datum/preferences/proc/random_facial_hair_style()
+	var/decl/species/mob_species = get_species_by_key(species || global.using_map.default_species)
```
out-of-phaze:
```suggestion
	var/decl/species/mob_species = get_species_decl()
```

code/modules/mob/new_player/preferences_setup.dm
```diff
@@ -1,13 +1,24 @@
...
+	var/decl/bodytype/mob_bodytype = mob_species?.get_bodytype_by_name(bodytype) || mob_species?.default_bodytype || GET_DECL(global.using_map.default_bodytype)
```
out-of-phaze:
```suggestion
	var/decl/bodytype/mob_bodytype = get_bodytype_decl()
```

out-of-phaze (changes requested):
First pass caught some preferences stuff that should probably be simplified using helpers. I haven't gone through and replaced existing versions of this, but new/modified ones should definitely use the helpers instead of duplicating the code.

code/modules/mob/new_player/preferences_setup.dm
MistakeNot4892: It doesn't actually _have_ any other uses lol.

code/unit_tests/icon_tests.dm
```diff
@@ -19,7 +19,7 @@
 		for(var/sprite_accessory_type in subtypesof(sprite_accessory_main_type))
 			var/failed = FALSE
 			var/decl/sprite_accessory/sat = sprite_accessory_type
-			var/sat_name = initial(sat.name)
+			var/sat_name = "[initial(sat.name)] ([initial(sat.uid)])"
```
out-of-phaze: It seems like this part of the test is checking if there are any sprite accessories with the same UID and name. Isn't UID checking already done elsewhere? I don't see the purpose of this check if we're grouping by `name (uid)`—grouping by name at least makes sense for display purposes.

out-of-phaze (dismissed):
I still kind of think `check_gender` should be obliterated, because of the issues with it we discussed on Discord, but I won't block over it. Same with the unit test concern I raised, though if you want to fix that one before merge that's fine too.

code/unit_tests/icon_tests.dm
MistakeNot4892: Looks like this whole test is obviated by `validate()`.

out-of-phaze (approved):

PR #3295, opened 2023-08-30
Adding some tweaks to useme brain PR that weren't included.

MistakeNot4892:


comma (approved):

code/modules/mob/living/carbon/brain/brain.dm
```diff
@@ -21,13 +21,13 @@
 	. = ..()
 
 /mob/living/carbon/brain/say_understands(mob/speaker, decl/language/speaking)
-	return (issilicon(speaker) && istype(container, /obj/item/mmi)) || ishuman(speaker) || ..()
+	return (issilicon(speaker) && (istype(container, /obj/item/mmi) || istype(loc, /obj/item/organ/internal/posibrain))) || ishuman(speaker) || ..()
```
out-of-phaze, line 24: I was going to complain about this not using container, but it turns out that `mmi_holder` sets `container` but `posibrain` doesn't. Works for me.

out-of-phaze (approved):

PR #3301, opened 2023-09-04
Fixes ID card object not being initialized due to mind.initial_account being set before the account is initialized

LenSkozzy:
## Description of changes

Modified the order of functions called in equip_rank in code/controllers/subsystems/jobs.dm, and moved some code that got called in /decl/hierarchy/outfit/proc/equip in /code/datums/outfits/outfit.dm to equip_id later in that file in order to avoid conflicts from the order change.

Although it does function properly as far as I've been able to tell, there's no way this PR is at the moment considered a "clean" fix; please let me know what needs to be changed in order to be accepted.

This PR will allow functions that attempt to use the ID object to function properly (i.e. using ATMs via your ID card instead of your account number, obtaining cash from folding@home, etc).

## Changelog
:cl:
bugfix: Fixed ID object initialization
/:cl:

### DIFF
--- a/code/datums/outfits/outfit.dm
+++ b/code/datums/outfits/outfit.dm
@@ -77,14 +77,6 @@ var/global/list/outfits_decls_by_type_
 /decl/hierarchy/outfit/proc/equip(mob/living/carbon/human/H, var/rank, var/assignment, var/equip_adjustments)
 	equip_base(H, equip_adjustments)
 
-	rank = id_pda_assignment || rank
-	assignment = id_pda_assignment || assignment || rank
-	var/obj/item/card/id/W = equip_id(H, rank, assignment, equip_adjustments)
-	if(W)
-		rank = W.rank
-		assignment = W.assignment
-	equip_pda(H, rank, assignment, equip_adjustments)
-
 	for(var/path in backpack_contents)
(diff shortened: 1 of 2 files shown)


code/datums/outfits/outfit.dm
```diff
@@ -191,10 +181,13 @@ var/global/list/outfits_decls_by_type_
 		W.rank = rank
 	if(assignment)
 		W.assignment = assignment
+	H.update_icon()
```
MistakeNot4892, line 184: I think this is probably redundant with the various other equip procs already causing icon updates. I'm not sure why it was present in previous code.

MistakeNot4892:
Assuming this doesn't mess with any of the other equip procs via ordering, which it looks like it doesn't, the fix looks looks fine. Not sure what you mean about a "cleaner" fix - this looks pretty clean. Only thing I'd request would be squashing it, and that can be done on our end anyway.

LenSkozzy:
Your changes cleaned up most of what I was worried about anyhow. Dunno what squashing entails, but either way, if you can do it on your end, sounds good to me. Thanks for the fixes!

MistakeNot4892 (approved):
update_icon() may be redundant but w/e

PR #3311, opened 2023-09-10
Fixing ID granting on staging.

MistakeNot4892:


out-of-phaze (approved):
This does make me worry about what would happen if, say, you equipped a job's outfit via the VV dropdown menu (I think it would overwrite, or at least add to, the existing ID), but it's more important to fix IDs in general rather than worry about adminbus.

MistakeNot4892:
I don't think the base outfit equip proc calls equip_id(). equip_rank() is the entrypoint that does that.

PR #3316, opened 2023-09-12
Breaking up some of lefthand.dmi and righthand.dmi.

MistakeNot4892:
## Description of changes
Deletes several unused states from lefthand/righthand.dmi.
Reworks various items to use the single icon system instead.

Someday I want to destroy righthand.dmi and lefthand.dmi and this is needed for that.

## Authorship
Myself.

## Changelog
Nothing player-facing.


code/modules/projectiles/guns/launcher/syringe_gun.dm
```diff
@@ -17,8 +17,10 @@
 	. = ..()
 	underlays.Cut()
 	if(syringe)
-		underlays += image(syringe.icon, src, syringe.icon_state)
-		underlays += syringe.filling
+		var/mutable_appearance/MA = syringe.appearance
```
out-of-phaze, line 20: Are pixel_x and pixel_y cleared? Would they even apply anyway?

code/modules/reagents/reagent_containers/hypospray.dm
```diff
@@ -2,12 +2,12 @@
...
-/obj/item/chems/hypospray //obsolete, use hypospray/vial for the actual hypospray item
+/obj/item/chems/hypospray
 	name = "hypospray"
 	desc = "A sterile, air-needle autoinjector for rapid administration of drugs to patients."
-	icon = 'icons/obj/syringe.dmi'
-	item_state = "hypo"
-	icon_state = "hypo"
+	icon = 'icons/obj/hypospray.dmi'
+	icon_state = ICON_STATE_WORLD
+	abstract_type = /obj/item/chems/hypospray
```
out-of-phaze, line 10: Isn't it convention for `abstract_type` to come first, i.e. before `name =`? That makes it clearer that it's abstract.

out-of-phaze (commented):
Did you make sure to test that syringe filling overlays still work? Without IDB, it's hard to check that without checking out the branch and running it, which is kind of a pain.

code/modules/reagents/reagent_containers/hypospray.dm
MistakeNot4892: I wasn't aware we had a convention for that

code/modules/projectiles/guns/launcher/syringe_gun.dm
MistakeNot4892: Fixed.

code/modules/reagents/reagent_containers/hypospray.dm
MistakeNot4892: Per Discord, leaving it where it is, but I readded a comment.

code/modules/projectiles/guns/launcher/syringe_gun.dm
out-of-phaze:
```dm
	if(randpixel)
		pixel_z = randpixel //an idea borrowed from some of the older pixel_y randomizations. Intended to make items appear to drop at a character
```
Looks like pixel_z should also be cleared. :(
EDIT: Maybe pixel_w too, at this point.

out-of-phaze (changes requested):

out-of-phaze (approved):

PR #3318, opened 2023-09-13
Fix overstripe corner states

out-of-phaze:
## Description of changes
Adds missing corner states to the overstripe detail overlay used for colored stripe floor decals.

Before, colored corner stripe floor decals would only be half-colored.


out-of-phaze:
IDB2 is busted, have this instead.
image

MistakeNot4892 (approved):

PR #3324, opened 2023-09-13
Fix immediately leaving recharge stations

out-of-phaze:
## Description of changes
Adds a 1-tick timer when entering recharge stations via Bumped() so that the keypress can be unregistered and the mob won't immediately exit.
This is a side-effect of porting SSInput, I think, so Bay shouldn't have this issue.

Prevents mobs from immediately exiting recharge stations the same tick they enter them.


MistakeNot4892 (approved):
Might be worth making this unique/override to avoid holding the key creating 900 of them.

PR #3330, opened 2023-09-13
Add support for compiled-in map template load landmarks

out-of-phaze:
## Description of changes
/obj/abstract/map_data now uses INITIALIZE_IMMEDIATE, since there's no reason for it not to.
Map template load landmarks will work when compiled in.
Map template load landmarks can disable loading their templates centered on their loc (centering is on by default).

The map load landmark doesn't use INITIALIZE_IMMEDIATE because running it in Initialize will break things.

Adds support for compiled-in map template load landmarks.
Adds support for map template load landmarks that aren't centered.
Makes /obj/abstract/map_data use INITIALIZE_IMMEDIATE to free up a `/New` override so I don't have to change the number of those in CI.

### DIFF
--- a/code/modules/multiz/map_data.dm
+++ b/code/modules/multiz/map_data.dm
@@ -19,8 +19,8 @@
 
 // If the height is more than 1, we mark all contained levels as connected.
 // This is in New because it is an auxiliary effect specifically needed pre-init.
-/obj/abstract/map_data/New(turf/loc, _height)
-	..()
+INITIALIZE_IMMEDIATE(/obj/abstract/map_data)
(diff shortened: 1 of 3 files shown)


code/modules/multiz/map_data.dm
```diff
@@ -19,8 +19,8 @@
 
 // If the height is more than 1, we mark all contained levels as connected.
 // This is in New because it is an auxiliary effect specifically needed pre-init.
```
MistakeNot4892, line 21: Comment is slightly out of date with the New override removed.

MistakeNot4892 (approved):

PR #3333, opened 2023-09-13
Fix terminal_setup adding terminals to unanchored machines 

out-of-phaze:
## Description of changes
The terminal_setup stock part preset no longer applies if the machine is unanchored.

This makes it so you can safely drag unanchored shield generators from a storage room to wherever they need to be deployed, without leaving a ghost terminal behind or causing runtimes.


MistakeNot4892 (approved):
Does anchoring/unanchoring destroy/remove/recreate terminals?

out-of-phaze:
> Does anchoring/unanchoring destroy/remove/recreate terminals?

I think construct states stop you from unanchoring it if it still has a terminal?

PR #3336, opened 2023-09-13
Move identity and name procs up to mob-level

out-of-phaze:
## Description of changes
Adds `exceptions` keyword argument to all overrides of `GetIdCards()` on `/mob`.
Adds return type annotation to `/mob/GetIdCard()`.
Moves various identity-related procs up to mob-level and generalizes/cleans up the associated logic.
Removes a redundant helper proc, `/proc/FindNameFromID()`, that was only used in one spot. Replaces it with the almost identical proc `/mob/proc/get_id_name()`.

Cleanup and generalization of some pretty old code. Removes humantype/carbontype bias in code which should allow for carbon removal and non-humantype playable mobs in the future.


code/modules/mob/mob.dm
```diff
@@ -1359,4 +1359,71 @@
...
 /mob/GetIdCard(exceptions = null)
+	RETURN_TYPE(/obj/item/card/id)
 	return LAZYACCESS(GetIdCards(exceptions), 1)
+
+// Get rank from ID, ID inside PDA, PDA, ID in wallet, etc.
+/mob/proc/get_authentification_rank(if_no_id = "No id", if_no_job = "No job")
+	var/obj/item/card/id/id = GetIdCard()
+	if(istype(id))
+		return id.rank ? id.rank : if_no_job
+	else
```
MistakeNot4892: Extraneous `else` that you seem to have trimmed off the other procs.

code/modules/mob/mob.dm
```diff
@@ -1359,4 +1359,71 @@
...
+	else if(get_rig()?.visible_name)
+		return get_rig()?.visible_name
+	return "Unknown"
+
+/mob/proc/identity_is_visible()
+	if(!real_name)
+		return FALSE
+	var/obj/item/clothing/mask/mask = get_equipped_item(slot_wear_mask_str)
+	var/obj/item/head = get_equipped_item(slot_head_str)
+	if(mask?.flags_inv & HIDEFACE || head?.flags_inv & HIDEFACE)
```
MistakeNot4892: `(mask?.flags_inv & HIDEFACE)` and `(head?.flags_inv & HIDEFACE)`

code/modules/mob/mob.dm
```diff
@@ -1359,4 +1359,71 @@
...
+		return FALSE
+	var/decl/bodytype/our_bodytype = get_bodytype()
+	if(our_bodytype?.has_limbs[BP_HEAD])
```
MistakeNot4892: Maybe we need a `should_have_limb()`. Noting as a side comment that `should_have_organ()` needs to be moved to /mob/living too.

MistakeNot4892 (changes requested):
Paren addition only, other code looks fine.

MistakeNot4892 (commented):
Looks good other than the conflict and CI stall.

code/modules/mob/examine.dm
```diff
@@ -40,11 +40,14 @@
 /mob/proc/show_other_examine_strings(mob/user, distance, infix, suffix, hideflags, decl/pronouns/pronouns)
 	return
 
-/mob/examine(mob/user, distance, infix, suffix)
+/mob/examine(mob/living/user, distance, infix, suffix)
```
MistakeNot4892: This makes the proc signature at odds with everywhere else, and ghosts will be unable to examine anyone now. Really should just gate the tail check imo.

MistakeNot4892 (approved):

PR #3338, opened 2023-09-13
Fix duplicate table removal and logging.

out-of-phaze:
## Description of changes
Fixes a runtime in the code for removing duplicate tables, due to update_icon() being called on a deleted table.
Replaces the bespoke table duplicate removal code with what's already used for lattices and catwalks.

Fixes a runtime and improves code quality.

### DIFF
--- a/code/game/objects/structures/tables.dm
+++ b/code/game/objects/structures/tables.dm
@@ -91,6 +91,8 @@
 	. = ..()
 	if(istype(oldloc))
 		for(var/obj/structure/table/table in range(oldloc, 1))
+			if(QDELETED(table))
+				continue
 			table.update_connections(FALSE)
 			table.update_icon()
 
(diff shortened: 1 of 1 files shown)


code/game/objects/structures/tables.dm
```diff
@@ -92,10 +90,11 @@
 	additional_reinf_material = null
 	. = ..()
 	if(istype(oldloc))
-		for(var/turf/turf as anything in RANGE_TURFS(oldloc, 1))
-			for(var/obj/structure/table/table in turf)
-				table.update_connections(FALSE)
-				table.update_icon()
+		for(var/obj/structure/table/table in range(oldloc, 1))
```
MistakeNot4892, line 93: I thought the runtime was coming from a destroyed table trying to call update_icon() on an uninitialized neighbor, rather than a qdeleted neighbor.

MistakeNot4892 (changes requested):
Code looks good otherwise, just curious about that qdeleted vs uninitialized issue.

code/game/objects/structures/tables.dm
out-of-phaze: iirc it was, then I changed it to use the macro and it was trying to call it on a qdeleted neighbor instead, I think?

MistakeNot4892 (approved):

PR #3341, opened 2023-09-13
Modernize unsim air comparisons and allow Test_ZAS_Connection for nonsim turfs

out-of-phaze:
## Description of changes
Modernizes unsim air comparisons to actually compare the airmixes rather than just the initial gas vars.
Removes unnecessary typechecks in Test_ZAS_Connection so that nonsim turfs participating in zones can be checked.

Fix Test_ZAS_Connection for nonsim turfs participating in zones.
Improves unsim air comparisons to be more accurate.

### DIFF
--- a/code/modules/ZAS/Diagnostic.dm
+++ b/code/modules/ZAS/Diagnostic.dm
@@ -13,7 +13,7 @@
 		for(var/g in mix.gas)
 			to_chat(mob, "ZONE GASES: [g]: [mix.gas[g]]\n")
 
-/client/proc/Test_ZAS_Connection(var/turf/simulated/T)
+/client/proc/Test_ZAS_Connection(var/turf/T)
 	set category = "Debug"
 	if(!istype(T))
 		return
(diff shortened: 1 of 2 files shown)


code/modules/ZAS/Diagnostic.dm
```diff
@@ -40,7 +40,7 @@
 			to_chat(mob, "No air passage :x")
 		return
 
-	var/turf/simulated/other_turf = get_step(T, direction_list[direction])
+	var/turf/other_turf = get_step(T, direction_list[direction])
 	if(!istype(other_turf))
```
MistakeNot4892, line 44: This should probably be `!other_turf?.zone` or something along those lines if the expectation is that it's participating in ZAS.

MistakeNot4892 (changes requested):
Not sure if my comment is the case or not.

code/modules/ZAS/Diagnostic.dm
out-of-phaze: it isn't, I used this to check why nonsim turfs weren't forming unsim edges on Tether

MistakeNot4892 (approved):

PR #3346, opened 2023-09-13
Port floor lamps from Polaris

out-of-phaze:
## Description of changes
Ports floor lamps from Polaris, with substantial rewriting for better code quality.

Polaris parity, plus they look neat.

## Authorship
Polaris/Virgo for the original sprites, some of the var values, and the initial code. @PlayerDeer for the updated sprites. Me for the modifications/rewriting/improvements.


code/modules/power/floorlamp.dm
```diff
@@ -0,0 +1,98 @@
...
+		lampshade.dropInto(loc)
+		lampshade = null
+		update_icon(0)
+		return TRUE
+	return ..()
+
+/obj/machinery/light/flamp/expected_to_be_on()
+	// Lamps with lampshades are controlled by a switch on the lamp instead of a lightswitch.
+	if(lampshade)
+		return switch_state && !(stat & NOPOWER)
```
MistakeNot4892, line 38: `operable()` instead of `!(stat & NOPOWER)` possibly?

code/modules/power/floorlamp.dm
```diff
@@ -0,0 +1,98 @@
...
+	. = ..()
+
+/obj/machinery/light/flamp/attackby(obj/item/held_item, mob/user)
+	if(!lampshade)
+		if(istype(held_item, /obj/item/lampshade))
+			lampshade = held_item
+			user.drop_from_inventory(held_item, src)
+			update_icon(0)
+			return TRUE
+	else if(IS_SCREWDRIVER(held_item))
```
MistakeNot4892: Not blocking, but it would be good if new inclusions could use `do_tool_interaction` instead of the typecheck pattern. At some point we're going to need to go back and change all of these to use it so the tool quality system is actually respected.

code/modules/power/floorlamp.dm
```diff
@@ -0,0 +1,98 @@
...
+	name = "large light bulb"
+	b_range = 6
+	b_power = 1
+	lighting_modes = list(
+		LIGHTMODE_EMERGENCY = list(l_range = 6, l_power = 0.45, l_color = LIGHT_COLOR_EMERGENCY),
+	)
+
+/obj/item/light/bulb/large/Initialize(mapload, obj/machinery/light/fixture)
+	. = ..()
+	set_scale(1.2, 1.2) // i hate this
```
MistakeNot4892, line 64: I hate this also, why not just a bigger icon?

MistakeNot4892 (changes requested):
Only the `operable()` comment is potentially blocking.

code/modules/power/floorlamp.dm
out-of-phaze: no sprites for it + on polaris/virgo there's just no visual difference at all, so even a scale is better than nothing imo

code/modules/power/floorlamp.dm
out-of-phaze: c.f. the code for normal lights
```dm
/obj/machinery/light/proc/expected_to_be_on()
	var/area/A = get_area(src)
	return A?.lightswitch && !(stat & NOPOWER)
```

PlayerDeer:
Dang my sprites are sick 👍 good PR

MistakeNot4892 (approved):

PR #3351, opened 2023-09-15
Material Extractor Rework

NataKilar:
## Description of changes
Depends on #3349

* Reworks the material extractor. It's now much simpler, depending on the atmos connection extension so that it may descend from the mineral processing subtype
* Adjusts some values of hydrates/ices for balancing purposes, pending proper state changes
* Adds some safety checks to ``transfer_type`` procs on reagent holders to prevent transferring reagents to containers which are full
* The atom flag ``ATOM_FLAG_NO_REACT`` now only limits chemical reactions, not phase changes and dissolving (for which other flags exist). Disable alloying on the smelter will now only disable alloying, not smelting of the materials.
The material extractor is currently non-functional, and uses a placeholder sprite. This rework should resolve both issues. Various code improvements and flexibility for chemistry.
## Authorship
Sprites and code by myself
## Changelog
:cl:
code: Reworked the material extractor
/:cl:


code/modules/materials/_materials.dm
```diff
@@ -441,6 +445,17 @@ INITIALIZE_IMMEDIATE(/obj/effect/gas_overlay)
 		return MAT_PHASE_LIQUID
 	return MAT_PHASE_SOLID
 
+// Returns the number of mols of material for the amount of solid or liquid units passed.
+/decl/material/proc/get_mols_from_units(units, phase)
```
MistakeNot4892, line 449: Should this handle gas as well?

code/modules/materials/definitions/solids/materials_solid_ice.dm
```diff
@@ -101,10 +101,10 @@
 	codex_name = null
 	uid = "solid_ice_hydrogen"
 	heating_products = list(
-		/decl/material/gas/hydrogen = 0.2,
-		/decl/material/liquid/water = 0.65,
-		/decl/material/gas/hydrogen/deuterium = 0.1,
-		/decl/material/gas/hydrogen/tritium = 0.05
+		/decl/material/gas/hydrogen = 0.05,
```
MistakeNot4892, line 104: The numerical tweaks seem pretty severe, what are the balancing issues you're fixing here?

code/modules/materials/_materials.dm
NataKilar: Not really, since gases have variable density, you can't tell how many moles you have from volume alone. Here we're just assuming the gases in the ice are in a liquid state, which doesn't make much sense but its the simplest thing to do right now.

code/modules/materials/definitions/solids/materials_solid_ice.dm
NataKilar: This is due to the unit to mol conversion changes.  I adjusted the values until I got something that seemed to give balanced results in game, but I'm happy to adjust them as requested.

code/modules/materials/_materials.dm
MistakeNot4892: Per Discord, I am concerned about people seeing the proc and assuming it works for any input phase. A warning if you give it a gas phase would probably be sufficient.

NataKilar: Warning added.

code/modules/mining/machinery/material_extractor.dm
```diff
@@ -1,409 +1,280 @@
...
+				continue
+			if(!can_eat(eating))
+				if(output_turf)
+					eating.dropInto(output_turf)
+				continue
+			eaten++
+			if(eating.reagents?.total_volume)
+				eating.reagents.trans_to_obj(src, eating.reagents.total_volume)
+			for(var/mtype in eating.matter)
+				reagents.add_reagent(mtype, FLOOR(eating.matter[mtype] * REAGENT_UNITS_PER_MATERIAL_UNIT))
```
PsyCommando, line 91: One of the problem with the old extractor was that things inside the ``matter`` and ``reagents`` var were never intended to be mixed together. The smelter only mixes things together in alloy mode, and doesn't care about the contents of the ``reagents`` holder.

You also removed the noreact flag, so those things are gonna react immediately when inserted, so you can't even filter them out at the output since they'll have reacted by then. 
Remember that the point of this is to extract specific chemicals and gases, not make a huge chemical reactor of doom :P

I suggest setting the noreact flag. Or at least some way of filtering it before it react. Gases should be mostly okay, since gas filters exist and there's no hypergolic gases/fluids in the game, or gas reactions.

code/modules/mining/machinery/material_extractor.dm
```diff
@@ -1,409 +1,280 @@
...
+			qdel(eating)
+			if(eaten >= MAX_INTAKE_ORE_PER_TICK)
+				break
+
+/obj/machinery/material_processing/extractor/on_reagent_change()
+	. = ..()
+
+	if(!reagents)
```
PsyCommando, line 99: Is this ever going to happen?

code/modules/mining/machinery/material_extractor.dm
```diff
@@ -1,409 +1,280 @@
...
+				var/sheets = FLOOR((removing / REAGENT_UNITS_PER_MATERIAL_UNIT) / SHEET_MATERIAL_AMOUNT)
+				if(sheets > 0) // If we can't process any sheets at all, leave it for manual processing.
+					adjusted_reagents = TRUE
+					SSmaterials.create_object(mtype, output_turf, sheets)
+					reagents.remove_reagent(mtype, removing)
+
+	return adjusted_reagents
+
+/obj/machinery/material_processing/extractor/attackby(obj/item/I, mob/user)
+	if(IS_WRENCH(I) && !panel_open)
```
PsyCommando, line 156: Doesn't this prevents removing components or something else from the base machinery code?

code/modules/mining/machinery/material_extractor.dm
```diff
@@ -1,409 +1,280 @@
...
+			to_chat(user, SPAN_NOTICE("You disconnect \the [src] from the port."))
+			return
+		else
+			var/obj/machinery/atmospherics/portables_connector/possible_port = locate(/obj/machinery/atmospherics/portables_connector) in loc
+			if(possible_port)
+				if(connection.connect(possible_port))
+					to_chat(user, SPAN_NOTICE("You connect \the [src] to the port."))
+					return
+				else
+					to_chat(user, SPAN_NOTICE("\The [src] failed to connect to the port."))
```
PsyCommando: Might wanna use something like span warning so people notice it failed.

code/modules/mining/machinery/material_extractor.dm
```diff
@@ -1,409 +1,280 @@
...
-	else
-		icon_state = initial_state
+		if(!CanInteract(user, global.default_topic_state))
+			return TOPIC_HANDLED
+
+		dispense_amount = clamp(amount, 0, 120)
+		return TOPIC_REFRESH
+
+	if(href_list["dispense"])
+		var/reagent_index = text2num(href_list["dispense"])
```
PsyCommando, line 204: I like that solution for filtering the reagents that get outputed. But, see above comment about my concerns on making a big reagents soup from them.

code/modules/mining/machinery/material_extractor.dm
```diff
@@ -1,409 +1,280 @@
...
-	if(!reagents)
-		create_reagents(GAS_EXTRACTOR_REAGENTS_TANK)
-	if(!input_buffer)
-		input_buffer = new(src)
-		input_buffer.create_reagents(GAS_EXTRACTOR_REAGENTS_INPUT_TANK) //Did this here because reimplementing that in the new() proc failed a test for some reasons
+	if(!gas_contents)
+		gas_contents = new(800)
+	set_extension(src, /datum/extension/atmospherics_connection, FALSE, gas_contents)
+
+	create_reagents(INFINITY)
```
PsyCommando: I wouldn't do that. Like maybe set it to double the processing limit, but setting it to infinity could cause some unforseen consequences. Especially since in dm all numbers are basically floats. and the number for infinity could add a huge amount of inaccuracy whenever the max capacity is used for a calculation.

code/modules/mining/machinery/material_extractor.dm
```diff
@@ -1,409 +1,280 @@
...
+			return TOPIC_REFRESH
+
+		if(!output_container || !output_container.reagents)
+			return TOPIC_HANDLED
+
+		reagents.trans_type_to(output_container, mtype, dispense_amount)
+		return TOPIC_REFRESH
+
+	if(href_list["eject"])
+		if(!output_container)
```
PsyCommando: Please put ejecting into it's own proc. So it can be re-used when dismantling and such.

code/modules/mining/machinery/material_extractor.dm
```diff
@@ -1,409 +1,280 @@
...
-		to_chat(user, output_desc)
+	var/datum/extension/atmospherics_connection/connection = get_extension(src, /datum/extension/atmospherics_connection)
+	if(connection.connected_port)
+		to_chat(user, SPAN_NOTICE("It is connected to \the [connection.connected_port]."))
 	else
-		to_chat(user, SPAN_NOTICE("It has nothing to pour reagents into."))
+		to_chat(user, SPAN_NOTICE("It may be connected to an atmospherics connector port with a wrench."))
+
+	if(atom_flags & ATOM_FLAG_OPEN_CONTAINER)
+		to_chat(user, SPAN_NOTICE("Its open for refilling."))
```
PsyCommando: Why would you need to pour anything into that? You probably don't want to let people fill up an infinite sized reagents holder?

code/modules/mining/machinery/material_extractor.dm
```diff
@@ -1,409 +1,280 @@
...
-/obj/machinery/atmospherics/unary/material/extractor/ProcessAtomTemperature()
-	if(operable() && use_power)
-		//We process temp for the input reagent_holder too
-		input_buffer.temperature = GAS_EXTRACTOR_OPERATING_TEMP
-		temperature = GAS_EXTRACTOR_OPERATING_TEMP
-		return TRUE
-	return ..()
 
-/obj/machinery/atmospherics/unary/material/extractor/examine(var/mob/user)
+/obj/machinery/material_processing/extractor/examine(mob/user)
```
PsyCommando, line 59: You might wanna leave the examine line for seeing if the output container is in the machine. It's pretty handy. Especially once we get a real sprite for this, which may not make it as obvious as the debug sprite.

code/modules/mining/machinery/material_extractor.dm
NataKilar: Setting the noreact flag will disable heating products as well, so that's a no-go as it currently is. I can change that, but I don't actually think there's a problem. You need to be careful with what you put into the extractor so that you don't get unintended reactions. If specific ices have heating products that immediately react, then that's a problem with the ices' products, and not the machinery in my opinion.

code/modules/mining/machinery/material_extractor.dm
NataKilar: Seems unlikely, although the check is on the smelter too.

code/modules/mining/machinery/material_extractor.dm
NataKilar: No, it requires the panel to be closed.

code/modules/mining/machinery/material_extractor.dm
NataKilar: Fair, I'll get rid of that option.

code/modules/mining/machinery/material_extractor.dm
NataKilar: I'll add the examine line back, but note that I did add an permanentish sprite for the machine.

code/modules/mining/machinery/material_extractor.dm
PsyCommando: > Setting the noreact flag will disable heating products as well, so that's a no-go as it currently is. I can change that, but I don't actually think there's a problem. You need to be careful with what you put into the extractor so that you don't get unintended reactions. If specific ices have heating products that immediately react, then that's a problem with the ices' products, and not the machinery in my opinion.

Yes, but that was the main issue with the previous iteration. There's always going to be a possibility for reactions currently no matter what you put in. 
Especially since people may add more reactions in the future. So you'd have to constantly change things each times a recipe might happen with the ices you allow.. Which is a lot of work, and essentially repeating the mistakes I made.

Also, consider that, if you can't actually extract those things separately from ices, this machine is essentially useless for reagents. You can just use a reagent grinder since it'll do the same thing.

code/modules/mining/machinery/material_extractor.dm
PsyCommando: You're right. I read that a bit quickly, and thought it was checking for the panel being open.

code/modules/mining/machinery/material_extractor.dm
NataKilar: Done

code/modules/mining/machinery/material_extractor.dm
NataKilar: Done, set the reagent volume to a finite amount.

code/modules/mining/machinery/material_extractor.dm
NataKilar: Done.

code/modules/mining/machinery/material_extractor.dm
NataKilar: Fair enough. I've split up the atom flags and set the extractor to only perform phase changes, not chemical reactions or dissolving.

MistakeNot4892 (approved):
I don't know how I feel about it being magic and preventing reacting or dissolving, but it's niche enough that I don't have a strong opinion on it.

PR #3360, opened 2023-09-21
Move species trait logic to living-level

out-of-phaze:
## Description of changes
Moves species trait logic from /living/carbon/human to /living.
Nulls (canonizes) the mob's trait list when it matches the mob's species traits exactly.
Adds `RemoveExtrinsicTrait` to reset a mob trait back to its species trait level.

Generalizes species trait logic, adds additional helpers.

### DIFF
--- a/code/datums/traits/traits.dm
+++ b/code/datums/traits/traits.dm
@@ -16,45 +16,62 @@
 /mob/living/proc/GetTraits()
 	SHOULD_NOT_SLEEP(TRUE)
 	RETURN_TYPE(/list)
-	return traits
-
-/mob/living/carbon/human/GetTraits()
-	if(traits)
-		return traits
-	return species.traits
+	var/decl/species/our_species = get_species()
+	return traits || our_species?.traits
 
 /mob/living/proc/SetTrait(trait_type, trait_level)
 	SHOULD_NOT_SLEEP(TRUE)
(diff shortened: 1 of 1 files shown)


code/datums/traits/traits.dm
```diff
@@ -16,45 +16,62 @@
...
+	if(!our_species) // Doesn't apply without a species.
+		return FALSE
+	var/list/missing_traits = traits ^ our_species?.traits
+	var/list/matched_traits = traits & our_species?.traits
+	if(LAZYLEN(missing_traits))
+		return FALSE
+	for(var/trait in matched_traits) // inside this loop we know our_species exists and has traits
+		if(traits[trait] != our_species.traits[trait])
+			return FALSE
+	traits = LAZYLEN(our_species.traits) ? our_species.traits.Copy() : null // if it has traits, use them, otherwise null
```
MistakeNot4892: Won't this mean that mobs that previously would have had null traits and hence return the species traits will now always copy their species traits? I feel like nulling `traits` here would do the same job unless I'm totally misunderstanding the idea.

out-of-phaze: oh yeah no it should just be nulled here no matter what i think?

MistakeNot4892 (approved):

PR #3365, opened 2023-09-22
Z-Mimic: Synchronize with upstream

Lohikar:
Updates Z-Mimic to the latest (in-dev) version. This should fix several edge case issues, as well as close the discovery loophole without ruining performance like 3-phase ZM. Support is also added for handling of long/wide atoms. ZM performance logging code is present, but define gated off for now.

This is currently not well tested, though the diff changes less than it appears to.


MistakeNot4892 (dismissed):

Lohikar:
DNM: OVERRIDE needs some changes.

code/__defines/zmimic.dm
```diff
@@ -1,7 +1,28 @@
+#define ZM_DESTRUCTION_TIMER(TARGET) addtimer(CALLBACK(src, TYPE_PROC_REF(/datum, qdel_self)), 10 SECONDS, TIMER_STOPPABLE)
```
MistakeNot4892: Suspect this should be `TARGET` not `src`.

Lohikar:
This should be good to go -- I'm not sure about the utility of some of those flags, but they should all work.

MistakeNot4892 (approved):

out-of-phaze (approved):

PR #3383, opened 2023-09-25
Multi-item recipes fix #3382

### LINKED ISSUE #3382: Recipes incorrectly handling multiple items
#### Description of issue

Judging by the source code, the handling of ingredients that are required in quantity other than 1x, is broken.
An example of that was reported by @sirleatherflute in cooking `milosoup`.

Instead of actually producing milosoup, recipe as described in codex produces a `burned mess`.

#### Steps to reproduce

- Put 10u of water, 2x tofu and 2x soydope into microwave
- Cook

`code/datums/recipe.dm`, actually.
Going to make a simple pull request for fixing that after some testing.

No idea.

Tested on local compiles of ScavStation:`dev` and Nebula:`staging`, as well as (with some effort expended) on Nebula testing server:
```
Client Version: 514
Server Revision: 88ae4d3c8e33b4f51fbef970a1890596fd020a10 - staging - 2023-09-24
Game ID: cpX-acgh
```

#### Issue bingo

cyberillithid:
## Description of changes
Since `/decl/recipe` uses a sometimes-numerical value to the component items (i.e., a few microwave food cooking recipes), the existing code was modified:
- the recipe deciding algorithm at `check_items` needs to iterate over all items (and not to break after finding a first example of required item);
- the procedure actually finalizing cooking process (`make_food`) should similarly consume and move all of the used-up items.

This way, the recipes should work the way they are written down (in code and in codex)

## Authorship
Myself.

## Changelog
:cl:
bugfix: fixed multi-item (food) recipe
/:cl:

## Postscriptum
Looking around, I guess that general `/decl/recipe/proc/make` may also need similar edits, but I am unsure.

### DIFF
--- a/code/datums/recipe.dm
+++ b/code/datums/recipe.dm
@@ -95,7 +95,7 @@
 					needed_items -= itype
 			else
 				needed_items -= itype
-			break
+			// break
 		if(!length(container_contents))
 			break
 	return !length(needed_items)
@@ -128,10 +128,15 @@
 	//Find items we need
 	if (LAZYLEN(items))
 		for (var/i in items)
-			var/obj/item/I = locate(i) in container_contents
-			if (I && I.reagents)
-				I.reagents.trans_to_holder(buffer,I.reagents.total_volume)
-				qdel(I)
+			var/cnt = 1
+			if (isnum(items[i]))
+				cnt = items[i]
+			for (cnt, cnt > 0, cnt--)
+				var/obj/item/I = locate(i) in container_contents
+				if (I && I.reagents)
+					container_contents -= I
+					I.reagents.trans_to_holder(buffer,I.reagents.total_volume)
+					qdel(I)
 
 	//Find fruits
 	if (LAZYLEN(fruit))
--- a/code/datums/recipe.dm
+++ b/code/datums/recipe.dm
@@ -93,8 +93,10 @@
 				--needed_items[itype]
 				if(needed_items[itype] <= 0)
 					needed_items -= itype
+					break
 			else
 				needed_items -= itype
+				break
 			// break
 		if(!length(container_contents))
 			break


code/datums/recipe.dm
```diff
@@ -162,7 +167,7 @@
 	If, as in the most common case, there is only a single result, then it will just be a reference to
 	the single-result's reagents
 	*/
-	var/datum/reagents/holder = new/datum/reagents(INFINITY, global.temp_reagents_holder)
+ 	var/datum/reagents/holder = new/datum/reagents(INFINITY, global.temp_reagents_holder)
```
MistakeNot4892: Looks like an extra space - DM is very finnicky about whitespace, this will cause all kinds of fun issues.

cyberillithid: Oops. It was accidental and apparently after I ran local tests. Yeah, will fix that now.

code/datums/recipe.dm
```diff
@@ -95,7 +95,7 @@
 					needed_items -= itype
 			else
 				needed_items -= itype
```
out-of-phaze: The break is necessary to avoid unnecessarily using stuff up in container_contents.
```suggestion
					needed_items -= itype
					break
			else
				needed_items -= itype
				break
```

out-of-phaze (changes requested):

code/datums/recipe.dm
```diff
@@ -93,9 +93,11 @@
 				--needed_items[itype]
 				if(needed_items[itype] <= 0)
 					needed_items -= itype
+					break
 			else
 				needed_items -= itype
-			break
+				break
+			// break
```
out-of-phaze, line 100:
```suggestion
```
Probably shouldn't leave this commented out here, but it's no big deal.

out-of-phaze (approved):

PR #3384, opened 2023-09-25
Tweaks syringe stab to a var.

Sparrowbird611:
## Description of changes
Adds a var, can_stab, to syringes, which determines whether they can or cannot stab.

Functionally it's not really different from how things are right now, it just makes it easier to allow a syringe to stab or not. I'm hoping to use this for a downstream.

## Authorship
Sparrowbird611

### DIFF
--- a/code/modules/reagents/reagent_containers/syringes.dm
+++ b/code/modules/reagents/reagent_containers/syringes.dm
@@ -74,10 +74,12 @@
 
 	if(!target.reagents)
 		return
-	if(!can_stab)
-		to_chat(user, SPAN_NOTICE("This syringe is too big to stab someone with it."))
-	else if((user.a_intent == I_HURT) && ismob(target))
-		syringestab(target, user)
+
+	if((user.a_intent == I_HURT) && ismob(target))
+		if(can_stab)
+			syringestab(target, user)
+		else
+			to_chat(user, SPAN_WARNING("This syringe is too big to stab someone with it."))
 		return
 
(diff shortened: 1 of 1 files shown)


code/modules/reagents/reagent_containers/syringes.dm
```diff
@@ -73,8 +74,9 @@
 
 	if(!target.reagents)
 		return
-
-	if((user.a_intent == I_HURT) && ismob(target))
+	if(!can_stab)
```
MistakeNot4892: I think what you want here is something more like
````
	if((user.a_intent == I_HURT) && ismob(target))
		if(can_stab)
			syringestab(target, user)
		else
			to_chat(user, SPAN_NOTICE("This syringe is too big to stab someone with it."))
		return
````
otherwise you don't return after printing the warning message, and every time the syringe is used it'll print the warning.

MistakeNot4892: SPAN_WARNING is also generally used when telling users why they can't do a thing, rather than SPAN_NOTICE.

MistakeNot4892 (changes requested):
See comments.

code/modules/reagents/reagent_containers/syringes.dm
Sparrowbird611: Done, thank you.

MistakeNot4892 (approved):

Issue #3373, opened 2023-09-24
Kharmaan nymph issues.


MistakeNot4892:
#### Description of issue
- kharmaan nymphs have a proper noun name
- ~~molting seems to go off nutrition rather than eating crystalline stuff~~ ate jailbreaker verte bottle, counted as crystal
- ~~randomly died as a carmoney nymph after 3-4 molts~~ drank jailbreaker verte

#### Client version, Server revision & Game ID
Client Version: 514
Server Revision: 88ae4d3c8e33b4f51fbef970a1890596fd020a10 - staging - 2023-09-24
Game ID: cpX-acgh
Current map: Tradeship Ocelot Alpha

PR #3386, opened 2023-09-28
Tweaking nymph name and stat entry.

MistakeNot4892:
Fixes nymph name, everything else in #3373 was user error (glass gives crystal, stuff in glass bottles caused damage via poison).

### DIFF
--- a/mods/species/ascent/mobs/nymph/nymph_life.dm
+++ b/mods/species/ascent/mobs/nymph/nymph_life.dm
@@ -3,7 +3,7 @@
 	. = ..()
 	if(stat == DEAD)
 		return
-	
+
 	// Generate some crystals over time.
 	if(nutrition >= 300 && crystal_reserve < ANYMPH_MAX_CRYSTALS)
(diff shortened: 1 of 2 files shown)


comma (approved):

mods/species/ascent/mobs/nymph/nymph_life.dm
```diff
@@ -42,6 +42,12 @@
 				if(150 to 250)					nymph_hud.drink.icon_state = "hydration3"
 				else							nymph_hud.drink.icon_state = "hydration4"
 
+/mob/living/carbon/alien/ascent_nymph/Stat()
+	. = ..()
+	if(client && statpanel("Status"))
```
out-of-phaze, line 47: I'm not sure if the client check is even necessary outside of setting `statobj`, which I don't think we do. The DM ref says `Stat()` is called periodically by the *client,* making me assume a client is a given if Stat() is running.

out-of-phaze (approved):

PR #3388, opened 2023-09-29
SSoverlays functions in credstick and pAI card

cyberillithid:
## Description of changes
Replaced old approach of direct overlay list manipulation with SSoverlays functions for credstick and pAI card.

Fixes https://github.com/ScavStation/ScavStation/issues/781 (with credstick inventory icon) as well as a similar problem on pAI device.

## Authorship
Myself.

## Changelog
:cl:
bugfix: credstick now display value in credits, not in thousands of credits
bugfix: pAIs emotes are shown on the pAI device inventory icon
refactor: logic for Nixie-tube-like 3-digit displays
/:cl:


code/modules/economy/worth_cash.dm
```diff
@@ -284,17 +284,17 @@
 		return
 
 	if(loaded_worth > 999999)
-		overlays += image(icon, "9__")
-		overlays += image(icon, "_9_")
-		overlays += image(icon, "__9")
+		add_overlay(image(icon, "9__"))
```
MistakeNot4892: One note for `add_overlay` (and overlays in general) - if the state is in the atom icon like it is here, you can just use the state name (`"9__"` in this case) and save the image() overhead.

code/modules/economy/worth_cash.dm
```diff
@@ -284,18 +284,36 @@
 	if(lock.locked)
 		return
 
-	if(loaded_worth > 999999)
-		add_overlay("9__")
-		add_overlay("_9_")
-		add_overlay("__9")
-		return
+	var/decl/currency/cur = GET_DECL(currency)
+	var/digits = loaded_worth / cur.absolute_value  // In currency
```
MistakeNot4892: Probably want a FLOOR() around the value here.

cyberillithid: On the contrary, I think that all `FLOOR`s in here are redundant and can (or should?) be safely removed: `%` always returns an integer.
But an edge use case I forgot to consider is negatives.

MistakeNot4892 (approved):

comma (approved):

PR #3394, opened 2023-09-30
Overhead text can now be offset for large atoms.

MistakeNot4892:
image


comma (approved):

out-of-phaze (approved):
Would like to revisit this at some point and see if we'd rather just take the parent's pixel_x/pixel_y (or z?) but for now this is fine.

PR #3395, opened 2023-09-30
Posters set dir in Initialize.

MistakeNot4892:
Fixes https://github.com/ScavStation/ScavStation/issues/782


comma (approved):

out-of-phaze (approved):
Looks like it never actually set the dir before—it just did bespoke offsetting stuff based on the placement dir. This is a better solution, though.

out-of-phaze (approved):

PR #3397, opened 2023-09-30
Upstreams fixes from Lighthouse and Shiptesting modpack branch

out-of-phaze:
## Description of changes
- Fixes the auto-generation of hydrate UIDs, which just outright did not work (that's why the hydrate UIDs are manually set).
- Fixes/expands the disposals connection unit test
  - Adds exceptions to disposals connection map unit tests, since we don't have down-facing disposals outlets.
  - Stops unit test disposalholders from merging if they stall, so it's clearer that they've stalled.
  - Doubles the step limit for disposalholders, since some maps I was testing were too large (!!!) for the current one.
- Prevents the vault cash freezer closet from overfilling by correcting the w_class calculation.
- Adds more diagnostics to the air alarm connection unit test, to prevent more Desperado-esque frustration.
- Disables transmitter latency for unit testing, which should prevent the Desperado issue from ever happening again even with a boatload of timers.
- Makes camera tag autogeneration use sequential_id instead of a bespoke system.
- Allows holodeck control circuitboards to create subtypes of holodeck control computers, useful for maps with multiple holodecks.
- Makes heat exchange pipes' set_dir override return parent. Also makes it use the general initialize_directions helper directly.
- Fixes the slime scanner and slime grappling messages.
- Fixes edge mimic turfs doubling their air/gas overlay.
- Fixes not being able to wear PDAs (and other items) on your belt slot.

Fixes a bunch of issues I caught/solved downstream.

### DIFF
--- a/mods/content/xenobiology/slime/_slime.dm
+++ b/mods/content/xenobiology/slime/_slime.dm
@@ -213,9 +213,9 @@
 			adjust_friendship(user, rand(2,3))
 		return TRUE
 
-	if(feeding_on)
-		var/prey = feeding_on
-		if(feeding_on == user)
+	var/prey = feeding_on?.resolve()
+	if(prey)
+		if(prey == user)
 			if(prob(60))
 				visible_message(SPAN_DANGER("\The [user] fails to escape \the [src]!"))
 			else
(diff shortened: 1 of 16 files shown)


comma (dismissed):

mods/content/xenobiology/slime/_slime.dm
```diff
@@ -213,22 +213,22 @@
 			adjust_friendship(user, rand(2,3))
 		return TRUE
 
-	if(feeding_on)
-		var/prey = feeding_on
-		if(feeding_on == user)
+	var/prey = feeding_on.resolve()
```
MistakeNot4892: `feeding_on?.resolve()` would avoid runtimes here.

MistakeNot4892 (changes requested):
Otherwise looks good.

mods/content/xenobiology/slime/_slime.dm
out-of-phaze: for some reason i thought feeding_on could never be null here, but maybe that was before i removed the `if(feeding_on)` check

MistakeNot4892 (approved):

PR #3398, opened 2023-09-30
Modpack config handling

PsyCommando:
## Description of changes
Ported this from a PR I was working on downstream to get configs that's persistence specific.
Added 3 hook procs stubs in /datum/configuration which get called in the huge switches for loading config options from file. Those can be overridden in modpacks to handle any extra config options specific to a modpack.

Adding modpack specific config options normally means you have to modify the core files to get them parsed. But this way you can just define all your new config options in your modpack along with the code parsing them.

### DIFF
--- a/code/controllers/configuration.dm
+++ b/code/controllers/configuration.dm
@@ -838,7 +838,9 @@ var/global/list/gamemode_cache = list()
 					config.show_typing_indicator_for_whispers = TRUE
 
 				else
-					log_misc("Unknown setting in configuration: '[name]'")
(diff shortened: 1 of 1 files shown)


MistakeNot4892 (approved):
Could be `else if` and lose an indentation level but that's a nitpick.
I think we need to declize config at some point but this is fine as an interim method imo.

PsyCommando:
> Could be `else if` and lose an indentation level but that's a nitpick. I think we need to declize config at some point but this is fine as an interim method imo.

It can't, it's a switch case.
But yeah, that would be nice to do something like that.

PR #3400, opened 2023-10-01
Converting config to a decl system.

MistakeNot4892:
## Description of changes
Another fey mood.
- Moves configuration onto a `/decl/config` type handled by a new configuration subsystem.
- Removes `/datum/configuration`.

No more bespoke datum for config. Subsystem has VV stuff, status reporting, so on.
No more horrifying switch statement for config options with individual var validation and conversion jammed into one giant 800 line block.
No more lack of clarity/sync between default config in the folder and the actual defaults in code.
Centralized value validation, update procs when values change, bespoke logic like parsing lists, etc.

## TODO
- [X] Coarse mass conversion of the old config system.
- [x] Standardize/account for conversion stuff (`MINUTES` `SECONDS` etc).
- [x] Implement protected config for stuff like comms password.
- [x] Implement the bespoke list parsing stuff for event probabilities and selectable modes.
- [x] Make sure the new ordering actually works and config is loaded before anything starts interrogating it.
- [x] Write a verb to dump out the current defaults to replace the `config/example` files.
- ~~Performance profiling and optimization of spots where the proc and decl overhead happens in hot code.~~
- ~~Merge all the configuration toggle verbs into one general verb or Config stat panel.~~
- ~~Testing in general.~~

## Authorship
Myself.

## Changelog
:cl:
tweak: Configuration has been rewritten to a new format. Servers with configuration changes from defaults will need to run the server to generate the new config file, then mirror their configuration changes in the new config system. Check the PR body for details.
/:cl:


code/controllers/subsystems/configuration.dm
```diff
@@ -0,0 +1,151 @@
+SUBSYSTEM_DEF(configuration)
```
PsyCommando, line 1: Not sure if this should be a subsystem? Because config is actually loaded before master inits inside world procs. Things like fps, ticklag, sql sever login and more.

code/datums/config/_config.dm
```diff
@@ -0,0 +1,126 @@
+#define CONFIG_FLAG_BOOL               BITFLAG(0)
+#define CONFIG_FLAG_NUM                BITFLAG(1)
+#define CONFIG_FLAG_TEXT               BITFLAG(2)
+#define CONFIG_FLAG_LIST               BITFLAG(3)
+#define CONFIG_FLAG_HAS_VALUE          BITFLAG(4)
```
PsyCommando: Just adding this here, but there's another type of config option that could be handled that's relatively common. Enums. There's a at least one options that accept a string from a list of possible strings which gets turned into a numeric value on parsing "objectives_disabled". It has 3 possible values.
Also, that kind of field is useful in general.

code/_helpers/game.dm
```diff
@@ -449,7 +449,7 @@
 /proc/SecondsToTicks(var/seconds)
 	return seconds * 10
 
-/proc/round_is_spooky(var/spookiness_threshold = config.cult_ghostwriter_req_cultists)
+/proc/round_is_spooky(var/spookiness_threshold = get_config_value(/decl/config/num/cult_ghostwriter_req_cultists))
```
PsyCommando, line 444: Shouldn't this config option be renamed to something a bit more meaningful, since it impacts more things than just ghost blood writing?

code/controllers/subsystems/configuration.dm
```diff
@@ -0,0 +1,151 @@
...
+			if ("database")
+				sqldb = value
+			if ("login")
+				sqllogin = value
+			if ("password")
+				sqlpass = value
+			else
+				log_misc("Unknown setting in configuration: '[name]'")
+
+/datum/admins/proc/dump_configuration()
```
PsyCommando, line 136: Seems like you got the same idea! Dumping options and writing the comments at the same time from code means the comments and value are more likely to match XD

code/datums/config/_config.dm
```diff
@@ -0,0 +1,126 @@
...
+	set_config_value(config_decl, !config_option.value)
+	return config_option.value
+
+/decl/config
+	decl_flags = DECL_FLAG_MANDATORY_UID
+	var/config_description
+	var/value
+	var/default_value
+	var/config_flags = CONFIG_FLAG_HAS_VALUE
+	var/dump_to_file = "config/configuration.txt"
```
PsyCommando: Here's a suggestion:
What about maybe dividing configuration options by categories instead of putting a filename directly into every single one of them? And then the config handler would sort out what categories goes into what config file?

code/datums/config/config_list.dm
```diff
@@ -0,0 +1,138 @@
...
+/decl/config/lists/mode_names
+	uid = "mode_names"
+
+// allowed modes
+/decl/config/lists/mode_allowed
+	uid = "modes"
+
+// votable modes
+/decl/config/lists/mode_votable
+	uid = "votable_modes"
```
PsyCommando: As penny mentioned on discord, ``mode_names``, ``mode_allowed``, ``mode_votable`` aren't really config options, they're just populated and stored in the config options for whatever reasons even if you can't change them via config. There's a few things like that.

code/datums/config/config_list.dm
```diff
@@ -0,0 +1,138 @@
...
+
+// relative probability of each mode
+/decl/config/lists/mode_probabilities
+	uid = "probabilities"
```
PsyCommando: Maybe this one config entry would need some rethinking? Because currently it's initialized and filled on game start, but it also lets you edit keys in it from the config file. And it worked in a much different way than any other options, since if you have that config key each times with a different game mode name, it'll set the probability for a different gamemode So unlike every single other config options having this one duplicated doesn't override it's last value, so checking for duplicate entries in the config just got a bit more difficult and less straightforwards.

code/datums/config/config_num.dm
```diff
@@ -0,0 +1,454 @@
+/decl/config/num
+	abstract_type = /decl/config/num
+	config_flags = CONFIG_FLAG_NUM | CONFIG_FLAG_HAS_VALUE
+	default_value = 0
+	var/min_value = -(INFINITY)
+	var/max_value = INFINITY
+
+/decl/config/num/sanitize_value()
+	value = clamp(value, min_value, max_value)
```
PsyCommando, line 11: I'd add rounding here too. Since sometimes you want strictly integers or allow decimals. Like time values have to be integers in most case. Same with things that can't be less than one, like the amount of mushroom mobs, or the amount of character slots and etc..

code/datums/config/config_num.dm
```diff
@@ -0,0 +1,454 @@
+/decl/config/num
+	abstract_type = /decl/config/num
+	config_flags = CONFIG_FLAG_NUM | CONFIG_FLAG_HAS_VALUE
+	default_value = 0
+	var/min_value = -(INFINITY)
```
PsyCommando, line 5: From having experimented on my side, it seems most config that want a number will want an unsigned number. With a few exceptions. Some also just accept -1 as "disabled" value.

code/datums/config/config_num.dm
```diff
@@ -0,0 +1,454 @@
...
+	default_value = TICK_LIMIT_MC_INIT_DEFAULT
+
+//if objectives are disabled or not
+/decl/config/num/objectives_disabled
+	uid = "objectives_disabled"
+	default_value = CONFIG_OBJECTIVE_NONE
+	config_description = "If objectives are disabled."
+
+/decl/config/num/objectives_disabled/set_value(new_value)
+	switch(new_value)
```
PsyCommando: Like I mentioned earlier maybe making an enum subtype for this kind of value be worthwhile?
Also, it's currently a bit ambiguous cause it's labelled as a /decl/num but it parses a string. It might be a good idea to explain somewhere that the default value and the /decl/config subtype name is based on the stored value and not the one inside the config file?

code/controllers/subsystems/configuration.dm
MistakeNot4892: Subsystems are created during world/New, the load is called directly from the same place old load was, so there's no ordering change. Making it a subsystem means getting an entry on the MC, not needing bespoke VV stuff, etc.

code/_helpers/game.dm
MistakeNot4892: I haven't done any name revision yet, more interested in getting it 1:1 with the prior config system first.

code/controllers/subsystems/configuration.dm
MistakeNot4892: This is actually done on first run if it can't find a config file, replacing the example folder file. Proc is just for debugging.

code/datums/config/_config.dm
MistakeNot4892: Categories are on the todo, though I didn't add them to the PR list yet.

code/datums/config/config_list.dm
MistakeNot4892: Yep, I need to go over modes and events config and make them their own systems prolly.

code/datums/config/config_num.dm
MistakeNot4892: Rounding could be a flag, sure. It can't be universal as a lot of numeric config is not integer based.

code/datums/config/config_num.dm
MistakeNot4892: The idea for these values is effectively just assuming any config that needs max/min sanitizing will set them to something sane.

code/datums/config/_config.dm
MistakeNot4892: Isn't the objectives thing the only enum-like?

code/datums/config/config_list.dm
MistakeNot4892: Mode stuff should probably be on the map datum rather than handled in config.

code/controllers/subsystems/configuration.dm
PsyCommando: Sounds good! 
The vv stuff I thought was to prevent changing settings that would have no effects at runtime?

code/datums/config/_config.dm
PsyCommando: It currently is, upstream. But downstream I got at least another one like that. And I could see more use cases for that.

code/_helpers/game.dm
PsyCommando: Yeah, that makes sense.

code/controllers/subsystems/configuration.dm
PsyCommando: Alright. I liked the idea of having everything at the same place and just having a git action dump it or something. But, it's not like imperative tbh.

code/datums/config/_config.dm
PsyCommando: Sounds good!

code/datums/config/config_list.dm
PsyCommando: Yeah, that would be best imo.

code/datums/config/config_num.dm
PsyCommando: What I did was not make a flag, but a rounding value like this:
```
/decl/config
   var/rounding = 0.01

/decl/config/proc/parse(val)
   return round(clamp(val, minimum, maximum), rounding)

/decl/config/someinteger
   rounding = null
```

So you can just set the rounding value to something else if you need/expect more precision and also how much precision.
And after having gone through a lot of the numeric config entry, most of them seems to be about strictly integers? Or at least they wouldn't produce the expected result if they had a decimal part.

code/datums/config/config_num.dm
PsyCommando: That's fair. Just remarked it was a more common occurrence than accepting negatives.

code/controllers/subsystems/configuration.dm
MistakeNot4892: Well, ideally with the update_post_value_change() proc in this PR we can do hot reloading of config.

code/datums/config/_config.dm
MistakeNot4892: Enum now has a type, it's more like a map though.

code/datums/config/_config.dm
MistakeNot4892: Done!

code/datums/config/config_list.dm
MistakeNot4892: I ended up just rewriting how mode probabilities etc. were handled, a more thorough review can happen after this PR.

code/datums/config/config_list.dm
MistakeNot4892: Per above, have left it in config for now.

code/datums/config/config_num.dm
MistakeNot4892: Done.

MistakeNot4892:
Depends on #3410

code/datums/config/config_list.dm
```diff
@@ -0,0 +1,25 @@
...
+/decl/config/lists/handle_value_conversion(var/new_value)
+	return json_decode(new_value)
+
+/decl/config/lists/compare_values(var/value_one, var/value_two)
+	if(!islist(value_one) || !islist(value_two))
+		return ..(value_one, value_two)
+	if(length(value_one) != length(value_two))
+		return FALSE
+	for(var/i = 1 to length(value_one))
+		if(value_one[i] != value_two[i])
```
out-of-phaze:
```suggestion
		if(!same_entries(value_one[i], value_two[i]) && value_one[i] != value_two[i])
```
We need to handle nested lists, apparently. Should probably make a `deep_same_entries()` helper in `code/_helpers/lists.dm` that handles this, and then have `/decl/config/lists/compare_values(one, two)` just call that.

out-of-phaze (changes requested):

out-of-phaze (dismissed):
Been long enough, config validation is passing and I didn't find any deep issues the last time I checked. This will be something to iron out during staging regardless, I think.

MistakeNot4892:
Leaving the testing and profiling for dev staging.

code/datums/config/config_num.dm
```diff
@@ -0,0 +1,15 @@
...
+	default_value = 0
+	var/min_value = -(INFINITY)
+	var/max_value = INFINITY
+	var/rounding
+
+/decl/config/num/sanitize_value()
+	..()
+	value = clamp(value, min_value, max_value)
+	if(isnull(rounding))
+		value = round(value)
```
out-of-phaze: I think `rounding = null` should skip rounding entirely. This seems to have introduced a number of CI fails.

out-of-phaze (changes requested):

code/datums/config/config_num.dm
out-of-phaze, line 1: Using an epsilon-compare should fix the CI issues, and thankfully we even have `rounding` to provide a precision level.
```dm
/decl/config/num/compare_values(value_one, value_two)
    if(!isnum(value_one) || !isnum(value_two) || !rounding)
        return ..(value_one, value_two)
    return abs(value_one - value_two) < rounding // this could also be rounding / 2, for greater accuracy
```

out-of-phaze (dismissed):

code/datums/config/config_num.dm
MistakeNot4892: Done.

code/datums/config/config_num.dm
MistakeNot4892: Done.

out-of-phaze (dismissed):

out-of-phaze (approved):

PR #3402, opened 2023-10-02
Fixes to weather system.

MistakeNot4892:
- De-lazied `weather_systems` as the subsystem expects to copy it and was causing a runtime.
- Replaced a helper with direct list access to `weather_by_z` as the list is reindexed by SSmapping and should always be long enough.
- Fixed debug weather creation verb.
- Fixed edge case where lateral z-levels were messing with `init_weather()`.


comma (approved):

out-of-phaze (approved):
This might have conflicts with dev eventually since IIRC I did some weather fixes on there too at one point?

PR #3405, opened 2023-10-02
Exterior turfs and lighting blend across lateral z-transitions.

MistakeNot4892:
## Description of changes
- Adds new helper procs for getting a step while also resolving to an actual turf, in the case of mimicked turfs.
- Plugs these procs into corner creation and exterior turf icon updates.

Would very much like @Lohikar's input on this one as I am fairly sure this will break things I don't understand. There are a couple of [visual artefacts that may be related](https://github.com/NebulaSS13/Nebula/assets/2468979/b1ff2e66-0d7a-48ac-97ea-37f7e763fdc7).

[Before.](https://cdn.discordapp.com/attachments/697198052183769089/1158291884721062019/image.png?ex=651bb6e1&is=651a6561&hm=15578e3db248aabe1f12842b58f1e941a4225a89831f527ed3742250d395398e&)
[After.](https://cdn.discordapp.com/attachments/697198052183769089/1158309812753420328/image.png?ex=651bc793&is=651a7613&hm=a83a7a2888d2200ae1a2a967d6deba65d05b4f632aefbef9d6141d7ab3bfcf70&)

## Authorship
Myself.

## Changelog
:cl:
tweak: Exterior turfs and lighting blend across z-level transitions now.
/:cl:

### DIFF
--- a/code/modules/lighting/lighting_corner.dm
+++ b/code/modules/lighting/lighting_corner.dm
@@ -66,18 +66,18 @@ var/global/list/REVERSE_LIGHTING_CORNER_DIAGONAL = list(0, 0, 0, 0, 3, 4, 0, 0, 
 
 	var/has_ambience = FALSE
 
-	t1 = new_turf
-	z = new_turf.z
+	t1 = new_turf.resolve_to_actual_turf()
+	z = t1.z
 	t1i = oi
 
-	if (new_turf.ambient_light)
+	if (t1.ambient_light)
 		has_ambience = TRUE
 
 	var/vertical   = diagonal & ~(diagonal - 1) // The horizontal directions (4 and 8) are bigger than the vertical ones (1 and 2), so we can reliably say the lsb is the horizontal direction.
 	var/horizontal = diagonal & ~vertical       // Now that we know the horizontal one we can get the vertical one.
 
-	x = new_turf.x + (horizontal == EAST  ? 0.5 : -0.5)
-	y = new_turf.y + (vertical   == NORTH ? 0.5 : -0.5)
+	x = t1.x + (horizontal == EAST  ? 0.5 : -0.5)
+	y = t1.y + (vertical   == NORTH ? 0.5 : -0.5)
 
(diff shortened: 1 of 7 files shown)


code/modules/lighting/lighting_corner.dm
```diff
@@ -66,7 +66,7 @@ var/global/list/REVERSE_LIGHTING_CORNER_DIAGONAL = list(0, 0, 0, 0, 3, 4, 0, 0,
 
 	var/has_ambience = FALSE
 
-	t1 = new_turf
+	t1 = new_turf?.resolve_to_actual_turf()
 	z = new_turf.z
```
MistakeNot4892: This should probably be `t1.z`

PsyCommando:
That's some pretty cool stuff! 

Any plans to make interacting with movables in the transition edge's vis contents work properly? And like things that connect with their neighbors work too? :P

MistakeNot4892:
I need to have a look at Adjacent() and see how cleanly the helper proc will integrate. I think it's potentially a good idea but it'll need care taken for stuff like targeting the right turf.

out-of-phaze (approved):
Looks good in terms of code. My only concern is the performance, but this shouldn't add many extra exterior turf `update_icon()` calls, just make the ones we already have slightly more expensive.

MistakeNot4892:
Regarding performance:
````
exodus ticked
ambient exterior light set to 0.5
num_exoplanets = 5
force_exoplanet_type = /datum/map_template/planetoid/random/exoplanet/volcanic

with mimic changes:
113s
116s
115s
103s
113s

without mimic changes:
100s
94s
96s
101s
106s
````
I tried a method that might potentially be cheaper, keeping a reference on the turf and using `?.` to reference it, but it came out universally more expensive.

Personally I think the init time increase, while bigger than I expected, is acceptable given the functionality.

out-of-phaze:
Yeah, why not, let's do it. We wouldn't be okay with shaving off 15 seconds by breaking borders, so why would we be okay with not fixing it to avoid a 15 second increase?

PR #3414, opened 2023-10-05
Fixing bodyscanner display connection logic

cyberillithid:
## Description of changes
For some reason bodyscanner used `tag` for connecting to displays instead of `id_tag` used on a map downstream. Fixed that and added calls for icon updates to redraw the display overlays.

Now the bodyscanner console would be able to connect to bodyscanner display, which in turn will light up according to its current state.

## Authorship
Myself.

## Changelog
:cl:
bugfix: fixed bodyscanner display connection
/:cl:

### DIFF
--- a/code/game/machinery/bodyscanner_console.dm
+++ b/code/game/machinery/bodyscanner_console.dm
@@ -9,7 +9,6 @@
 	construct_state = /decl/machine_construction/default/panel_closed
 	uncreated_component_parts = null
 	stat_immune = 0
-	var/list/display_tags = list()
 	var/list/connected_displays = list()
 	var/list/data = list()
 
@@ -36,7 +35,7 @@
 
 /obj/machinery/body_scanconsole/proc/FindDisplays()
 	for(var/obj/machinery/body_scan_display/D in SSmachines.machinery)
-		if(D.tag in display_tags)
+		if(D.id_tag == connected.id_tag)
 			connected_displays += D
 			events_repository.register(/decl/observ/destroyed, D, src, .proc/remove_display)
 	return !!connected_displays.len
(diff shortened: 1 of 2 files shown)


code/game/machinery/bodyscanner_console.dm
```diff
@@ -36,7 +35,7 @@
 
 /obj/machinery/body_scanconsole/proc/FindDisplays()
 	for(var/obj/machinery/body_scan_display/D in SSmachines.machinery)
-		if(D.tag in display_tags)
+		if(D.id_tag == connected.id_tag)
```
out-of-phaze, line 38:
```suggestion
		if(D.id_tag in display_tags)
```
I believe the intent of this was so that you could have a many-to-one connection? I don't fully understand the reason for removing this.

comma: This list doesn't seem to get populated anywhere. Multiple displays should still connect if they have same tag as the connected scanner if you mean that.

cyberillithid: As far as I get it, previous implementation attempted to use default `/datum/var/tag` property (acc. to DM reference, to be used with `locate()`) on the displays themselves, which wasn't set and thus not working neither on ministation nor downstream. The ministation2 map sets `id_tag` property (as defined in `/obj/machinery` IIRC) instead, which ostensibly fits the usecase better: setting it to the same value on a bodyscanner and multiple displays would likewise achieve the many-to-one connection.

out-of-phaze: > This list doesn't seem to get populated anywhere. Multiple displays should still connect if they have same tag as the connected scanner if you mean that.

It's supposed to be set by mappers.

> setting it to the same value on a bodyscanner and multiple displays would likewise achieve the many-to-one connection.

I suppose that *is* true. We can probably just revisit it if id_tag on body scanner consoles gets any other uses that would cause problems if shared by multiple consoles.

out-of-phaze (approved):
Approving, though making a note that the `display_tags` list might become necessary in the future if we add more wi-fi/radio/signal functionality to body scanners based on id_tag.

PR #3379, opened 2023-09-24
Crossed()/Uncrossed() consistency pass.

MistakeNot4892:
## Description of changes
- Standardizes argument to `Crossed()`/`Uncrossed()`.
- Generally cleans up overrides.
- Fixes a few places where the crosser is insufficiently checked.
- Fixes #3378,
- Broke some vehicle and mulebot logic out into an /atom/movable proc (`crossed_mob()`).

## Why and what will this PR improve
Fixes a number of bugs, general code quality.

## Authorship
Myself.

## Changelog
Nothing player-facing.

PR #3417, opened 2023-10-08
Merging r3 staging to stable. [IDB IGNORE] [MDB IGNORE]

MistakeNot4892:
Opening for 24h to get any comments/objections.


out-of-phaze:
I would at least like #3379 and #3420 in first, to avoid having to retarget them to staging.

PR #3420, opened 2023-10-09
Disable belt uniform check for species without it

out-of-phaze:
## Description of changes
If a uniform slot doesn't exist, can_equip_to_slot now returns true for the belt slot.

Species without a uniform slot can now equip belt slot items even without `ITEM_FLAG_IS_BELT` without a uniform on.
I'm not actually sure if this is a good thing or not. Maybe it should check a variable on bodytype/species/something instead of just unilaterally returning true?


MistakeNot4892 (approved):
This is fine, I just broke it when I implemented it from the old HUD slot system. Would you mind looking at the ID slot as well? I believe it has the same check.

out-of-phaze:
> This is fine, I just broke it when I implemented it from the old HUD slot system. Would you mind looking at the ID slot as well? I believe it has the same check.

IDs already do effectively the same thing, returns false if `if(check_slot && !check_slot.get_equipped_item())`.

PR #3423, opened 2023-10-10
Replacing invisibility magic numbers with defines/helper where possible.

MistakeNot4892:
## Description of changes
Sneaking in this before freezing stuff for staging.
Replaces bare numbers and direct var setting of invisibility with the defines and helper.
Fixes flood overlays being invisible.

More consistent values, and use of procs allows overrides.

## Authorship
Myself. @out-of-phaze spotted the issue with flood overlays.

## Changelog
Nothing player-facing.


MistakeNot4892:
Moved the flood fixes into #3424 so I could target stable.

comma (approved):

out-of-phaze (approved):

PR #3429, opened 2023-10-11
Spawning tweaks.

MistakeNot4892:
## Description of changes
- Various global latejoin lists are replaced with spawnpoint decls.
- Observer spawnpoints are now a spawnpoint decl.
- Name-based backup decl finding if uid fails has been generalized in a proc on decls_repository.
- Spawnpoint selection in prefs is now non-modal.

Nicer code, easier to override/modify downstream or in a modpack.

## Authorship
Myself.

## Changelog
Nothing player-facing.


comma (dismissed):

code/controllers/subsystems/jobs.dm
```diff
@@ -505,7 +505,7 @@ SUBSYSTEM_DEF(jobs)
 			H.forceMove(S.loc)
 		else
 			var/decl/spawnpoint/spawnpoint = job.get_spawnpoint(H.client)
-			H.forceMove(pick(spawnpoint.turfs))
+			H.forceMove(length(spawnpoint.spawn_turfs) ? pick(spawnpoint.spawn_turfs) : locate(1,1,1))
```
Andrew-Fall: DEFAULTPICK could be used instead

code/modules/mob/observer/ghost/ghost.dm
```diff
@@ -61,11 +61,15 @@ var/global/list/image/ghost_sightless_images = list() //this is a list of images
...
-		else
-			T = locate(1, 1, 1)
+
+		var/list/spawn_locs = list()
+		var/list/all_spawns = decls_repository.get_decls_of_subtype(/decl/spawnpoint)
+		for(var/spawn_type in all_spawns)
+			var/decl/spawnpoint/spawn_data = all_spawns[spawn_type]
+			if(spawn_data.ghost_can_spawn && length(spawn_data.spawn_turfs))
+				spawn_locs |= spawn_data.spawn_turfs
+		T = length(spawn_locs) ? pick(spawn_locs) : get_respawn_loc()
```
Andrew-Fall: DEFAULTPICK

Andrew-Fall: Also I feel as though it would be better to use get_respawn_loc() and have that proc do checks on either flags or using another proc that takes the mob as a parameter.

code/modules/mob/new_player/new_player.dm
```diff
@@ -360,7 +363,7 @@ INITIALIZE_IMMEDIATE(/mob/new_player)
 		if(!job)
 			job = SSjobs.get_by_title(global.using_map.default_job_title)
 		var/decl/spawnpoint/spawnpoint = job.get_spawnpoint(client, client.prefs.ranks[job.title])
-		spawn_turf = pick(spawnpoint.turfs)
+		spawn_turf = length(spawnpoint.spawn_turfs) ? pick(spawnpoint.spawn_turfs) : get_respawn_loc()
```
Andrew-Fall: DEFAULTPICK

code/datums/repositories/decls.dm
```diff
@@ -40,6 +40,19 @@ var/global/repository/decls/decls_repository = new
 	RETURN_TYPE(/decl)
 	. = get_decl(fetched_decl_ids[decl_id], validate_decl_type)
 
+/repository/decls/proc/get_decl_by_id_or_var(var/decl_id, var/decl_prototype, var/check_var = "name")
```
Andrew-Fall, line 55: Given the code this is replacing, I feel like this should have a comment saying it's for grandfathered in items and shouldn't be used for new code without good reason.

code/controllers/subsystems/jobs.dm
out-of-phaze:
```suggestion
			H.forceMove(DEFAULTPICK(spawnpoint.spawn_turfs, locate(1,1,1)))
```
I kind of feel like we should log if it doesn't have turfs, though. Is that done elsewhere?

out-of-phaze (changes requested):
Seconding the requests for changes from Andrew. I'm also not really sure what the purpose of `get_respawn_loc()` is. Is it just... *any* turf you could spawn on? The name doesn't make it clear and there's no comments, either.

code/controllers/subsystems/jobs.dm
MistakeNot4892: There's a unit test to validate that all spawnpoints that are either in the allowed list or forced by jobs have turfs to spawn on.

code/datums/repositories/decls.dm
MistakeNot4892: Comments added.

MistakeNot4892:
Okay, this has spiraled a bit.
- Replaced `spawn_turfs` being used directly with a proc so I can override it on my fork and check prefs/culture stuff.
- Added a flag system instead of the bools and made the random spawn point proc take a flag to filter results.
- Moved the gateway onto Exodus since that's the only map that uses it.
- Made the default allowed spawns just Arrivals.

code/unit_tests/map_tests.dm
```diff
@@ -299,19 +299,70 @@
...
 
 /datum/unit_test/correct_allowed_spawn_test/start_test()
+
 	var/list/failed = list()
-	for(var/decl/spawnpoint/spawnpoint as anything in global.using_map.allowed_spawns)
-		if(!length(spawnpoint.turfs))
+	var/list/check_spawn_flags = list(
+		"SPAWN_FLAG_PRISONERS_CAN_SPAWN" = SPAWN_FLAG_PRISONERS_CAN_SPAWN,
+		"SPAWN_FLAG_JOBS_CAN_JOIN"       = SPAWN_FLAG_JOBS_CAN_SPAWN,
+		"SPAWN_FLAG_PERSISTENCE_CAN_USE" = SPAWN_FLAG_PERSISTENCE_CAN_SPAWN
```
out-of-phaze:
```suggestion
		"SPAWN_FLAG_JOBS_CAN_SPAWN"       = SPAWN_FLAG_JOBS_CAN_SPAWN,
		"SPAWN_FLAG_PERSISTENCE_CAN_SPAWN" = SPAWN_FLAG_PERSISTENCE_CAN_SPAWN
```
I'm not sure why these don't match?

code/modules/client/preferences_spawnpoints.dm
```diff
@@ -1,14 +1,50 @@
...
+	abstract_type = /decl/spawnpoint
+	decl_flags = DECL_FLAG_MANDATORY_UID
+	/// Name displayed in preference setup.
+	var/name
+	/// Message to display on the arrivals computer. If null, no message will be sent.
+	var/spawn_announcement
+	/// Determines validity for get_random_spawn_turf()
+	var/spawn_flags = (SPAWN_FLAG_GHOSTS_CAN_SPAWN | SPAWN_FLAG_JOBS_CAN_SPAWN | SPAWN_FLAG_PRISONERS_CAN_SPAWN | SPAWN_FLAG_PERSISTENCE_CAN_SPAWN)
+	/// List of turfs to spawn on. Retrieved via get_spawn_turfs().
+	var/list/_spawn_turfs
```
out-of-phaze: I think this should be `VAR_PRIVATE/list/_spawn_turfs` or `VAR_PROTECTED/list/_spawn_turfs`. Unless you have a use-case downstream/in a modpack for things other than `/decl/spawnpoint` and its subtypes modifying `_spawn_turfs` directly rather than through helpers.

out-of-phaze (changes requested):

code/unit_tests/map_tests.dm
MistakeNot4892: Oh, I renamed the flags at some point. The string is just for display purposes but I'll amend it.

out-of-phaze (approved):

PR #3432, opened 2023-10-13
Unacidable var replaced with material checks.

MistakeNot4892:
## Description of changes
- Removes `unacidable`.
- Sets materials on some things that didn't have it set already.
- Moves mob response to acid into a proc that can be overridden.
- Makes unacidable checks use the material of the object and solvent levels in a `solvent_can_melt()` helper.
- Overrides `solvent_can_melt()` to return false in a few specific spots that I wasn't sure about.

Tying systems together, removing ancient logic.

## Authorship
Myself.

## Changelog
:cl:
tweak: Acid resistance has been tweaked; report things melting or not melting unexpectedly on the bug tracker.
/:cl:

### DIFF
--- a/code/game/objects/objs.dm
+++ b/code/game/objects/objs.dm
@@ -325,5 +325,7 @@
 
 // TODO: maybe iterate the entire matter list or do some partial damage handling
 /obj/proc/solvent_can_melt(var/solvent_power = MAT_SOLVENT_STRONG)
+	if(!simulated)
+		return FALSE
 	var/decl/material/mat = get_material()
 	return !mat || mat.dissolves_in <= solvent_power
--- a/code/game/objects/objs.dm
+++ b/code/game/objects/objs.dm
@@ -8,7 +8,6 @@
 	var/list/req_access
 	var/list/matter //Used to store information about the contents of the object.
 	var/w_class // Size of the object.
-	var/unacidable = 0 //universal "unacidabliness" var, here so you can use it in any obj.
 	var/throwforce = 1
 	var/sharp = 0		// whether this object cuts
 	var/edge = 0		// whether this object is more likely to dismember
@@ -323,3 +322,8 @@
 	..()
 	if(!QDELETED(src) && fluids?.total_volume)
 		fluids.touch_obj(src)
+
+// TODO: maybe iterate the entire matter list or do some partial damage handling
+/obj/proc/solvent_can_melt(var/solvent_power = MAT_SOLVENT_STRONG)
+	var/decl/material/mat = get_material()
+	return !mat || mat.dissolves_in <= solvent_power
--- a/code/modules/materials/_materials.dm
+++ b/code/modules/materials/_materials.dm
@@ -741,33 +741,8 @@ INITIALIZE_IMMEDIATE(/obj/effect/gas_overlay)
 				return
 		M.clean_blood()
 
-	if(solvent_power > MAT_SOLVENT_NONE && removed >= solvent_melt_dose && M.acid_act(min(removed * solvent_power * ((removed < solvent_melt_dose) ? 0.1 : 0.2), solvent_max_damage), solvent_melt_dose, solvent_power))
+	if(solvent_power > MAT_SOLVENT_NONE && removed >= solvent_melt_dose && M.solvent_act(min(removed * solvent_power * ((removed < solvent_melt_dose) ? 0.1 : 0.2), solvent_max_damage), solvent_melt_dose, solvent_power))
 		holder.remove_reagent(type, REAGENT_VOLUME(holder, type))
-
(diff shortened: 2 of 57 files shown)


code/modules/materials/_materials.dm
```diff
@@ -745,34 +741,33 @@ INITIALIZE_IMMEDIATE(/obj/effect/gas_overlay)
...
-						screamed = TRUE
-						H.emote("scream")
-					affecting.status |= ORGAN_DISFIGURED
-
-		if(!M.unacidable)
-			M.take_organ_damage(0, min(removed * solvent_power * ((removed < solvent_melt_dose) ? 0.1 : 0.2), solvent_max_damage), override_droplimb = DISMEMBER_METHOD_ACID)
+	if(solvent_power > MAT_SOLVENT_NONE && removed >= solvent_melt_dose && M.acid_act(min(removed * solvent_power * ((removed < solvent_melt_dose) ? 0.1 : 0.2), solvent_max_damage), solvent_melt_dose, solvent_power))
+		holder.remove_reagent(type, REAGENT_VOLUME(holder, type))
+
+/mob/living/proc/acid_act(var/severity, var/amount_per_item, var/solvent_power)
```
comma: shouldn't this be somewhere in mob code with similar damage procs, kinda weird having it in material module

code/modules/reagents/reagent_containers/spray.dm
```diff
@@ -22,7 +21,9 @@
 	var/tmp/particle_move_delay       = 10                          ///lower is faster
 	var/tmp/sound_spray               = 'sound/effects/spray2.ogg'  ///Sound played when spraying
 	var/safety                        = FALSE                       ///Whether the safety is on
-	
+
+/obj/item/chems/spray/solvent_can_melt(var/solvent_power = MAT_SOLVENT_STRONG)
+	return FALSE // maybe reconsider this
```
out-of-phaze, line 26: Probably not necessary unless we make the contents inside chem objects able to melt them.

code/game/objects/objs.dm
```diff
@@ -323,3 +322,8 @@
 	..()
 	if(!QDELETED(src) && fluids?.total_volume)
 		fluids.touch_obj(src)
+
+// TODO: maybe iterate the entire matter list or do some partial damage handling
+/obj/proc/solvent_can_melt(var/solvent_power = MAT_SOLVENT_STRONG)
```
out-of-phaze, line 327: IMO, return false here if `simulated` is false.

code/modules/mechs/equipment/utility.dm
```diff
@@ -445,15 +445,11 @@
...
-			to_chat(user, SPAN_WARNING("\The [target] can't be drilled away."))
-			return
-		else
-			to_chat(tmob, FONT_HUGE(SPAN_DANGER("You're about to get drilled - dodge!")))
+		to_chat(tmob, FONT_HUGE(SPAN_DANGER("You're about to get drilled - dodge!")))
 
 	else if (isobj(target))
 		var/obj/tobj = target
-		if (tobj.unacidable)
+		if (!tobj.solvent_can_melt()) // why the hell does this use an acid check
```
out-of-phaze: This should check `drill_head.material.hardness`, probably. (Later, the proc also uses `drill_head.material.brute_armor` which is an unpleasant abuse of variables...)

out-of-phaze (changes requested):

code/modules/materials/_materials.dm
out-of-phaze: Personally I think it's fine to keep acid-related code in one section? We already do something similar with multiz code.

code/modules/reagents/reagent_containers/spray.dm
MistakeNot4892: I am 90% sure they do currently. May be thinking of Polaris though. I have memories of pouring acid into something to dissolve a material and melting the container.

code/modules/materials/_materials.dm
MistakeNot4892: I left it in there by mistake, meant to move it to living code.

code/modules/reagents/reagent_containers/spray.dm
out-of-phaze: `standard_pour_into` calls `trans_to` which calls `touch_obj`, yeah. the other thing to consider would be just, do we have beakers of acid-safe materials? there's glass which is probably fine. could unit test it maybe, to make sure all prepopulated reagent containers can hold their initial reagents without melting

MistakeNot4892: Glass is currently immune to acid so beakers etc should be fine.

out-of-phaze (approved):

PR #3434, opened 2023-10-14
Tweaking fuel_value on materials.

MistakeNot4892:
## Description of changes
- `fuel_value` is now `accelerant_value` and uses a set of defines instead of largely magic numbers.
- Various accelerant values have been tweaked.
- Burn products are now added to the environment using a material proc.
- `/obj/item/melt()` has been moved to `/obj/melt()` so structures and such can use it.
- CI checks have been added to materials for accelerant value vs ignition point.

Needed to help make the fire sim a bit more consistent.

## Authorship
Myself.

## Changelog
:cl:
tweak: Some accelerant values have been tweaked; report fuels burning too vigorously to the issue tracker.
/:cl:


code/modules/materials/_materials.dm
```diff
@@ -834,4 +838,9 @@ INITIALIZE_IMMEDIATE(/obj/effect/gas_overlay)
 			.[interaction] = interacted_units
 			total_interacted_units -= interacted_units
 		if(total_interacted_units <= 0)
-			return
\ No newline at end of file
+			return
+
+/decl/material/proc/add_burn_product(var/datum/gas_mixture/environment, var/amount)
```
out-of-phaze: I feel like `add_burn_product` should take `loc` instead of `environment`, so that (as mentioned by the TODO comment about it) it can handle solid products as well?

code/modules/power/port_gen.dm
```diff
@@ -177,11 +177,9 @@
 	..()
 
 /obj/machinery/port_gen/pacman/proc/process_exhaust()
-	var/decl/material/mat = GET_DECL(sheet_material)
-	if(mat && mat.burn_product)
-		var/datum/gas_mixture/environment = loc.return_air()
-		if(environment)
-			environment.adjust_gas(mat.burn_product, 0.05*power_output)
+	var/decl/material/mat = sheet_material && GET_DECL(sheet_material)
```
out-of-phaze: Is this any different than calling `GET_DECL` directly? Doesn't it have a ternary that checks `ispath(foo)`? `foo && ispath(foo)` will always equal `ispath(foo)`, so this check seems redundant.

out-of-phaze (changes requested):

code/modules/power/port_gen.dm
MistakeNot4892: GET_DECL runtimes on a null path.

MistakeNot4892: Per Discord, apparently not.

code/modules/materials/_materials.dm
MistakeNot4892: Couldn't make this universal due to the zone use of it but made it accept both.

out-of-phaze (dismissed):

out-of-phaze (approved):

PR #3442, opened 2023-10-14
Adds dragging items around tables and racks.

MistakeNot4892:
## Description of changes
- Mousedrop handlers now take params.
- You can drag and drop items on tables to scoot them around (like sliding a drink down the bar)

It's a fun little interaction.

https://github.com/NebulaSS13/Nebula/assets/2468979/f347430a-ad5f-4cfd-8b7e-579a36a2d4e0

## Authorship
The concept is from Polaris but the code is entirely mine.

## Changelog
:cl:
add: You can now slide items around on tables with drag and drop.
/:cl:


code/modules/mob/observer/ghost/ghost.dm
```diff
@@ -421,8 +421,9 @@ This is the proc mobs get to turn into a ghost. Forked from ghostize due to comp
 	show_browser(src, dat, "window=manifest;size=370x420;can_close=1")
 
 //This is called when a ghost is drag clicked to something.
-/mob/observer/ghost/MouseDrop(atom/over)
+/mob/observer/ghost/MouseDrop(over_object, src_location, over_location, src_control, over_control, params)
 	SHOULD_CALL_PARENT(FALSE)
+	var/atom/over = over_object
```
out-of-phaze, line 423: I don't understand why you can't just have the first argument here be `atom/over_object`?

code/game/objects/item.dm
```diff
@@ -253,27 +253,59 @@
...
+		var/obj/structure/struct = over
+		if(struct.structure_flags & STRUCTURE_FLAG_SURFACE)
+			if(user == loc && !user.try_unequip(src, get_turf(user)))
+				return TRUE
+			if(!isturf(loc))
+				return TRUE
+			// Clear our glide so we don't do an animation when dropped into the target turf.
+			var/list/click_data = params2list(params)
+			do_visual_slide(get_turf(src), pixel_x, pixel_y, get_turf(over), text2num(click_data["icon-x"])-1, text2num(click_data["icon-y"])-1)
+			// Reset our glide_size now that movement has completed.
```
out-of-phaze:
```suggestion
			var/list/click_data = params2list(params)
			do_visual_slide(get_turf(src), pixel_x, pixel_y, get_turf(over), text2num(click_data["icon-x"])-1, text2num(click_data["icon-y"])-1)
```
These comments should probably be moved into `do_visual_slide()`.

code/game/objects/item.dm
```diff
@@ -253,27 +253,59 @@
 
 	return ..(user, distance, "", desc_comp)
 
+/obj/item/proc/do_visual_slide(var/turf/from, var/from_pixel_x, var/from_pixel_y, var/turf/target, var/target_pixel_x, var/target_pixel_y)
```
out-of-phaze: I don't know if this proc is more appropriate here or in `code/_helpers/animations.dm`. It seems the only reason it has to be on `/obj/item` to begin with, rather than a more general proc, is that it uses `center_of_mass`?

out-of-phaze (dismissed):

code/modules/mob/observer/ghost/ghost.dm
MistakeNot4892: I was making the procs align with the DM doc, which doesn't seem to type them. I can edit it if you really want me to.

code/game/objects/item.dm
MistakeNot4892: More or less yeah. I did consider generalizing it but I'm not sure how much use it would see.

code/modules/mob/observer/ghost/ghost.dm
out-of-phaze: I mean, it's fine to have types in proc arguments that differ from their corresponding parent proc argument. I just wasn't sure why it was done this way, but you explained.

MistakeNot4892:
Wrestling to get this to disable atom gliding when moving between adjacent tiles.

MistakeNot4892:
Working now.
https://cdn.discordapp.com/attachments/697198052183769089/1164735948769198150/dreamseeker_DVHdIVo99L.mp4?ex=65444be1&is=6531d6e1&hm=3a571dcf2ccbf909dd42bf9e05394df11475b38b960bc20f46acc519369bee7a&

code/_helpers/animations.dm
```diff
@@ -98,4 +98,23 @@
...
+
+// This proc is used to move an atom to a target loc and then interpolite to give the illusion of sliding from start to end.
+/proc/do_visual_slide(var/atom/movable/sliding, var/turf/from, var/from_pixel_x, var/from_pixel_y, var/turf/target, var/target_pixel_x, var/target_pixel_y, var/center_of_mass)
+	set waitfor = FALSE
+	var/start_pixel_x = sliding.pixel_x - ((target.x-from.x) * world.icon_size)
+	var/start_pixel_y = sliding.pixel_y - ((target.y-from.y) * world.icon_size)
+	// Clear our glide so we don't do an animation when dropped into the target turf.
+	var/old_animate_movement = sliding.animate_movement
+	sliding.animate_movement = NO_STEPS
+	sleep(2) // Due to BYOND being byond, animate_movement has to be set for at least 2 ticks before gliding will be disabled.
```
out-of-phaze:
```suggestion
	sleep(2 * world.tick_lag) // Due to BYOND being byond, animate_movement has to be set for at least 2 ticks before gliding will be disabled.
```
I think Kapu said something about how this is more accurate? Since you're pausing for 2 deciseconds, rather than 2 ticks, which is more than is necessary.

out-of-phaze (changes requested):

out-of-phaze (dismissed):

out-of-phaze (dismissed):

out-of-phaze (approved):

PR #3444, opened 2023-10-14
Adds maximum temp values to rock strata

comma:
## Description of changes
They were previously set to default infinite value

Added values from cursory google search

It resulted in volcanic caves etc generating with sandstone rocks and such, immersion shattered

## Authorship

Me

## Changelog
Doubt anyone notices

### DIFF
--- a/code/modules/materials/definitions/solids/materials_solid_stone.dm
+++ b/code/modules/materials/definitions/solids/materials_solid_stone.dm
@@ -32,6 +32,7 @@
 	uid = "solid_sandstone"
 	lore_text = "A clastic sedimentary rock. The cost of boosting it to orbit is almost universally much higher than the actual value of the material."
 	value = 1.5
+	melting_point = T0C + 600
 
 /decl/material/solid/stone/granite
 	name                   = "granite"
@@ -66,6 +67,7 @@
 	wall_support_value = MAT_VALUE_VERY_HEAVY
 	hardness = MAT_VALUE_HARD
 	reflectiveness = MAT_VALUE_SHINY
+	melting_point  = T0C + 1200
 	brute_armor = 3
 	integrity = 201 //hack to stop kitchen benches being flippable, todo: refactor into weight system
 	construction_difficulty = MAT_VALUE_HARD_DIY


code/modules/materials/geology/strata_metamorphic.dm
```diff
@@ -2,6 +2,7 @@
 	name = "metamorphic rock"
 	base_materials = list(/decl/material/solid/stone/marble)
 	default_strata_candidate = TRUE
+	maximum_temperature = T0C + 600
```
MistakeNot4892: Shouldn't this just use the melting point of the base materials?

MistakeNot4892: Per Discord, strata use the melting point of their base materials but it's not set properly on marble/etc.

comma: Done

code/modules/materials/definitions/solids/materials_solid_stone.dm
```diff
@@ -66,6 +67,7 @@
 	wall_support_value = MAT_VALUE_VERY_HEAVY
 	hardness = MAT_VALUE_HARD
 	reflectiveness = MAT_VALUE_SHINY
+	melting_point  = T0C + 600
```
out-of-phaze: > The melting temperature of dry granite at ambient pressure is 1215–1260 °C (2219–2300 °F); it is strongly reduced in the presence of water, down to 650 °C at a few hundred megapascals of pressure.

Are we sure 600C is high enough?

MistakeNot4892 (dismissed):

code/modules/materials/definitions/solids/materials_solid_stone.dm
```diff
@@ -32,6 +32,7 @@
 	uid = "solid_sandstone"
 	lore_text = "A clastic sedimentary rock. The cost of boosting it to orbit is almost universally much higher than the actual value of the material."
 	value = 1.5
+	melting_point = T0C + 400
```
out-of-phaze: It seems that silicon dioxide (quartz, the main component of sandstone) melts at 1,713C? Which is a bit high. It looks like heating it to 600C might cause it to start fracturing so we could probably use that as a limit.

out-of-phaze (changes requested):

code/modules/materials/definitions/solids/materials_solid_stone.dm
comma: Made it 1200

code/modules/materials/definitions/solids/materials_solid_stone.dm
comma: Made it 600

out-of-phaze (approved):

PR #3446, opened 2023-10-16
Screen object unification.

MistakeNot4892:
## Description of changes
- Unifies a lot of /obj/screen functionality on the base object.
- Centralizes the code in its own folder.
- Does a lot of cleanup and refactoring in the process of updating screen objects to use handle_click() etc.

Screen object code is a mess and this cleans it up significantly. Will help a lot with the eventual HUD datum rewrite I have in the pipeline.

## Testing
- [x] Intent selector
- [x] Zone selector
- [x] Inventory
- [x] Hands
- [x] Make sure gear highlights didn't get clobbered
- [x] Exosuits
- [x] Constructs
- [x] Robots
- [x] pAIs
- [x] AIs
- [x] Spells
- [x] Ability master
- [x] Action buttons
- [x] Gun fire mode selector
- [x] Credits
- [x] Global HUDs (NVG, thermals)
- [x] Pain/welding overlays
- [x] Radials
- [x] Click catchers
- [x] Cataloguer overlay
- [x] Holomaps
- [x] Storage
- [x] Probably other stuff I will remember later

## Authorship
Myself.

## Changelog
Nothing player-facing.


MistakeNot4892:
Tested and ready to go.

code/_onclick/hud/deity.dm
```diff
@@ -2,49 +2,9 @@
 	hud_type = /datum/hud/deity
 
 /datum/hud/deity/FinalizeInstantiation()
-	var/obj/screen/intent/deity/D = new()
-	adding += D
-	action_intent = D
+	action_intent =  new /obj/screen/intent/deity(null, mymob)
+	src.adding = list(action_intent)
+	src.other = list()
```
out-of-phaze, line 7: Is there a reason to use `src` for these?

code/_onclick/hud/screen/screen_intent.dm
```diff
@@ -0,0 +1,69 @@
...
+			intent = I_HELP
+	else if(icon_y <= world.icon_size/2)
+		intent = I_GRAB
+	update_icon()
+	user.a_intent = intent
+
+/obj/screen/intent/on_update_icon()
+	icon_state = "intent_[intent]"
+
+/obj/screen/intent/deity
```
out-of-phaze, line 26: I don't know how to feel about this being separated from other deity code. I thought we were going to move gamemode content to modpacks?

code/_onclick/hud/screen/screen_intent.dm
```diff
@@ -0,0 +1,69 @@
...
+			S.maptext = "<span style='font-size:7pt;font-family:Impact'><font color='#3C3612'>[P.name]</font></span>"
+		else
+			S.maptext = null
+
+/obj/screen/intent/deity/handle_click(mob/user, params)
+	..()
+	update_text()
+
+/obj/screen/deity_marker
+	name = "" //Don't want them to be able to actually right click it.
```
out-of-phaze, line 65: This might just make its name default to `deity_marker`, the last part of the path, I'm unsure. I know it does this for a `null`/unset name...

PsyCommando:
I just have a slight comment/request. 
The action bar used to react to the incapacitation state of the character, so when they'd go unconscious or something the buttons would be grayed out and translucent.  (It also used to get stuck like that cause state updates didn't cause the ui to update consistently. But that's something else XD )

Maybe that feature could be built-in the base screen object? So you could just toggle it on the elements you want to go disabled when they can't be used?

MistakeNot4892:
Sorry, that's out of scope for this.
I am generalizing it a bit in the HUD rewrite though.

PsyCommando:
> Sorry, that's out of scope for this. I am generalizing it a bit in the HUD rewrite though.

That's all good!

out-of-phaze (changes requested):
Partial review, hasn't made any progress in the past few weeks so I'll probably give the rest of the PR a cursory look-over and then approve.

code/_onclick/hud/screen/screen_intent.dm
MistakeNot4892: Sure, but until we do, I wanted to keep the screen stuff all in the same spot.

code/_onclick/hud/screen/screen_intent.dm
MistakeNot4892: This is the existing code. I don't know offhand if this works or if you need to do `verbs.Cut()` as well.

out-of-phaze (approved):

PR #3451, opened 2023-10-17
Cleaning up ID and PDA equipping.

MistakeNot4892:
## Description of changes
Fixes https://github.com/ScavStation/ScavStation/issues/799. More specifically:
- `equip_rank` -> `equip_job_title` to disambiguate from military rank datums, which are also passed around these procs.
- `equip()` to `equip_outfit()` and `equip_job()` for searchability.
- `equip_id` override merged into base proc.
- `equip_id` calls unified on the outfit datum in the `equip_outfit` chain.
- `equip_outfit` arguments somewhat simplified/cleaned up.
- `setup_account()` moved prior to outfit dressing.

Fixes issues with equip ordering and ID not having assignment or account info set sometimes.

## Authorship
Myself.

## Changelog
Nothing player-facing.

### DIFF
--- a/code/datums/outfits/outfit.dm
+++ b/code/datums/outfits/outfit.dm
@@ -74,9 +74,9 @@ var/global/list/outfits_decls_by_type_
 		J.toggle()
 		J.toggle_valve()
 
-/decl/hierarchy/outfit/proc/equip(mob/living/carbon/human/H, var/rank, var/assignment, var/equip_adjustments)
+/decl/hierarchy/outfit/proc/equip_outfit(mob/living/carbon/human/H, assignment, equip_adjustments, datum/job/job, datum/mil_rank/rank)
 	equip_base(H, equip_adjustments)
-
+	equip_id(H, assignment, equip_adjustments, job, rank)
 	for(var/path in backpack_contents)
 		var/number = backpack_contents[path]
 		for(var/i=0,i<number,i++)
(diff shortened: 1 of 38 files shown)


code/datums/outfits/outfit.dm
```diff
@@ -74,9 +74,9 @@ var/global/list/outfits_decls_by_type_
 		J.toggle()
 		J.toggle_valve()
 
-/decl/hierarchy/outfit/proc/equip(mob/living/carbon/human/H, var/rank, var/assignment, var/equip_adjustments)
+/decl/hierarchy/outfit/proc/equip_outfit(mob/living/carbon/human/H, assignment, equip_adjustments, datum/job/job, datum/mil_rank/rank)
```
out-of-phaze, line 77: Do alt titles just not get added to the ID anymore?

MistakeNot4892: I was using `rank` instead of `job.title`, fixed now.

code/datums/outfits/outfit.dm
```diff
@@ -169,30 +169,31 @@ var/global/list/outfits_decls_by_type_
...
 		return
 	var/obj/item/card/id/W = new id_type(H)
 	if(id_desc)
 		W.desc = id_desc
-	if(rank)
-		W.rank = rank
 	if(assignment)
 		W.assignment = assignment
 	if(job)
+		W.position = job.title
```
out-of-phaze, line 183: This still won't factor in the selected alt title, will it?

out-of-phaze (changes requested):

code/datums/outfits/outfit.dm
MistakeNot4892: Alt title is `assignment`.

out-of-phaze (approved):

PR #3455, opened 2023-10-19
Fix various matter-related issues

out-of-phaze:
## Description of changes
- Fixes single cable coils spawning with the wrong amount.
- Fixes network and electrical cables costing too much when fabricating.
- Fixes lit matches/lighters dealing brute damage instead of burn
- Fixes matches dealing 5 damage
- Fixes chewing tobacco and chewing gum being made of plastic and fiberglass
- Fix visible messages with bad deaf/self messages
- Removes rounding from matter amount modifier
- Update lighter/match object flags to fix overwriting/matter values
- Makes force-based item value calculation only apply to items with the new IS_WEAPON item flag.

Fixes the value and matter of a lot of objects being way too high. Stuff is going to be cheaper and have less matter in it across the board.

### DIFF
--- a/code/game/objects/items/weapons/candle/candle.dm
+++ b/code/game/objects/items/weapons/candle/candle.dm
@@ -51,6 +51,8 @@
 /obj/item/flame/candle/proc/light(mob/user)
 	if(!lit)
 		lit = TRUE
+		damtype = BURN
+		update_force()
 		visible_message(SPAN_NOTICE("\The [user] lights \the [src]."), SPAN_NOTICE("You light \the [src]."))
 		set_light(candle_range, candle_power)
 		START_PROCESSING(SSobj, src)
--- a/code/game/objects/items/weapons/candle/candle.dm
+++ b/code/game/objects/items/weapons/candle/candle.dm
@@ -48,12 +48,12 @@
 		var/obj/item/flame/candle/other_candle = A
 		other_candle.light()
 
-/obj/item/flame/candle/proc/light(mob/user)
+/obj/item/flame/candle/light(mob/user, no_message)
 	if(!lit)
-		lit = TRUE
-		damtype = BURN
+		..()
 		update_force()
-		user.visible_message(SPAN_NOTICE("\The [user] lights \the [src]."), SPAN_NOTICE("You light \the [src]."))
+		if(!no_message)
+			user.visible_message(SPAN_NOTICE("\The [user] lights \the [src]."), SPAN_NOTICE("You light \the [src]."))
 		set_light(candle_range, candle_power)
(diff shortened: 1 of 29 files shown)


code/game/objects/items/weapons/candle/candle.dm
```diff
@@ -71,7 +73,7 @@
 
 /obj/item/flame/candle/attack_self(mob/user)
 	if(lit)
-		lit = 0
+		lit = FALSE
```
comma: shouldn't this also change damtype back to brute?

code/modules/economy/worth_items.dm
```diff
@@ -16,12 +16,14 @@
...
+			if(tech[t] <= 1)
+				continue
+			var/next_tech_val = ((tech[t] - 1)**2) * 5
 			if(next_tech_val > largest_tech_val)
 				largest_tech_val = next_tech_val
-			. += largest_tech_val
+		. += largest_tech_val
 
-	if(force)
+	if(item_flags & ITEM_FLAG_IS_WEAPON && force)
```
comma: (item_flags & ITEM_FLAG_IS_WEAPON) && force

code/game/objects/items/weapons/storage/boxes.dm
```diff
@@ -438,12 +438,13 @@
 
 /obj/item/storage/box/matches/attackby(obj/item/flame/match/W, mob/user)
 	if(istype(W) && !W.lit && !W.burnt)
-		W.lit = 1
-		W.damtype = "burn"
+		W.lit = TRUE
```
MistakeNot4892: Maybe the lighting/unlighting should be a match proc.

code/game/objects/objs.dm
```diff
@@ -35,7 +35,7 @@
 	return ..()
 
 /obj/proc/get_matter_amount_modifier()
-	. = CEILING(w_class * BASE_OBJECT_MATTER_MULTPLIER)
+	. = w_class * BASE_OBJECT_MATTER_MULTPLIER
```
MistakeNot4892, line 38: Minor concern, but we should probably do rounding in the places where this is used as a mult now.

out-of-phaze: I think we do

out-of-phaze: yeah, every usage uses round()

code/game/objects/items/weapons/storage/boxes.dm
out-of-phaze: Unlighting is, it's `extinguish()`

MistakeNot4892 (approved):

PR #3460, opened 2023-10-20
Damage propagation for stock_parts and SMES overload (fixes #3457)

### LINKED ISSUE #3457: APCs overloaded by malfunctioning SMES are irrepairable
#### Description of issue

SMES total failure processing makes APCs ostensibly connected to get overloaded, with some chance breaking them entirely.

But the way they are broken (I suspect `set_broken(TRUE)` with implied generic-legacy reason of being broken) renders them impossible to repair or even disassemble: trying to pry the circuitboard causes machine state validation to fail (probably because `reason_broken=MACHINE_GENERIC_BROKEN` and not `MACHINE_BROKEN_CONSTRUCT`), and the `construct_state` is set to `null` after throwing an exception.

Afterwards the remnant APC frame is totally inoperable, irremovable and irrepairable; the only way to remove it is to use admin powers of deletion; yet they still count as APCs, so you cannot even build another one for this area, because "This area already has an APC"

Expected: Even if APCs are done for, the remnants should be either removable or repairable. Probably the frame should be still intact after overloading, so just fry the power components? Or something.
Actual: Broken APC disassembly fails, and the replacement cannot be installed.

#### Steps to reproduce

0. If we're starting on a Tradeship, maybe fuse box should be disabled (no idea whether it's really relevant)
1. Locate a fully-charged SMES
2. Disable safeties and grounding by cutting wires
3. Examine the wires to cause a system failure discharge upon the 
4. Reroll 1-3 (or just 3) until a proper kind of total system failure happens (when APCs are overloaded and broken by `/obj/machinery/power/smes/buildable/proc/apcs_overload`)

-- or, alternatively, just select a SMES and do Advanced ProcCall `apcs_overload` with args `100, 0, 0`

5. Find the APC destroyed by the SMES system failure
6. Attempt to disassemble it by: 
    a. unlocking the cover with proper ID,
    b. removing all the stock parts
    c. cutting the wires
    d. try to pry the circuit open
7. Now you get an exception, and the frame is hanging there, undisassemblable and unreassemblable, with `construct_state` of `null`.

> `Exception has occurred: %% code/game/machinery/_machines_base/machine_construction/_construction.dm,90 %% Machine Washroom APC (/obj/machinery/power/apc) ([0x2000bd4]) (the floor) (33,36,3 - Tradeship Ocelot Alpha) (/turf/simulated/floor/tiled/freezer) violated the state assumptions of the construction state /decl/machine_construction/wall_frame/no_wires/hackable!`

I suspect that something has to be done with `code\modules\power\smes_construction.dm`, line 274: maybe instead of setting generic `set_broken`, output an amount of electrocute damage to circuitboard / all power components of APC?

Probably since the machinery rework?

First achieved downstream on ScavStation:main (lagging a bit behind Nebula:stable).
Reproducible somewhat on Nebula:stable.

#### Issue bingo

cyberillithid:
## Description of changes
Changed APCs failure from plain `set_broken(TRUE)` to getting dealt electrocute damage in SMES failure procedures.

To allow that to break more than a singular `stock_part`, damage returned from `take_damage` should be changed from being reduced only by armor (unapplicable to stock parts) to keeping track of the health the item had before damage; thus, sufficient amount of damage to machinery will dismantle it in one hit.

Will make APCs repairable after SMES failures.

## Authorship
Myself

## Changelog
:cl:
tweak: changed the SMES overloading APC failure damage
bugfix: fixed APCs being stuck irrepairable on overloading
bugfix: fixed clamping any machinery damage to one part per hit
/:cl:

### DIFF
--- a/code/game/objects/item_damage.dm
+++ b/code/game/objects/item_damage.dm
@@ -18,6 +18,7 @@
 		return 0 //must return a number
 
 	//Apply damage
+	damage = min(health, damage)
 	health = clamp(health - damage, 0, max_health)
 	check_health(damage, damage_type, damage_flags)
 	return damage
--- a/code/modules/power/smes_construction.dm
+++ b/code/modules/power/smes_construction.dm
@@ -210,8 +210,8 @@
 				SET_STATUS_MAX(h_user, STAT_PARA, 6)
 			spawn(0)
 				empulse(src.loc, 8, 16)
-			charge = 0
 			apcs_overload(1, 10, 20)
+			charge = 0
 			energy_fail(10)
 			src.ping("Caution. Output regulators malfunction. Uncontrolled discharge detected.")
 
@@ -225,8 +225,8 @@
 			SET_STATUS_MAX(h_user, STAT_PARA, 8)
 			spawn(0)
 				empulse(src.loc, 32, 64)
-			charge = 0
 			apcs_overload(5, 25, 100)
+			charge = 0
 			energy_fail(30)
 			src.ping("Caution. Output regulators malfunction. Significant uncontrolled discharge detected.")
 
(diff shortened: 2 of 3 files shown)


code/game/objects/item_damage.dm
```diff
@@ -18,9 +18,11 @@
 		return 0 //must return a number
 
 	//Apply damage
+	var/damage_taken = health // Initialize with current health value...
```
MistakeNot4892: This doesn't seem right to me - I feel like just doing `damage = min(health, damage)` and otherwise leaving the logic unchanged would do the job.

code/modules/power/smes_construction.dm
```diff
@@ -258,23 +258,35 @@
...
-
+	var/list/obj/machinery/power/apc/apcs = list()
 	for(var/obj/machinery/power/terminal/T in powernet.nodes)
 		var/obj/machinery/power/apc/A = T.master_machine()
 		if(istype(A))
 			if (prob(overload_chance))
 				A.overload_lighting()
 			if (prob(failure_chance))
-				A.set_broken(TRUE)
+				if (!(A in apcs))
```
MistakeNot4892: `apcs |= A` will do a non-duplicating add, for reference.

code/game/objects/item_damage.dm
MistakeNot4892: The proc claims to return the damage taken by the object, not the amount of health left after taking damage.

cyberillithid: Yeah, that's what I'm doing here, but the `min` is definitely more elegant.
On the other side, since we're already removed the `<0` cases, even `clamp` could be removed? But alright.

MistakeNot4892 (approved):

PR #3465, opened 2023-10-22
Pulling in Scav's updated ministation to replace the existing ministation.

MistakeNot4892:
## Description of changes
- Adds head roles and rep role to ministation roster.
- Adds overmap and multiz support to ministation.
- Pulls in various other map changes from ScavStation.

Given that this is the map that ScavStation runs rounds on, it is a much more complete and tested experience than the existing ministation.

## Authorship
I just copied in the code and descavved it, I believe most of the changes are @Devchord and @Qumefox (and other scavs).

## Changelog
:cl:
tweak: Updated ministation from ScavStation. Lots of changes, see PR.
/:cl:


out-of-phaze:
I feel like we really need to still have a functional non-multiz/1z (at least mostly, I know there's already space.dmm) map for testing purposes. Up until now I've been using Ministation for that, but this seems to add 3 new z-levels.

MistakeNot4892:
Misclick :(

out-of-phaze (dismissed):
The concern with OpenDream is apparently no longer true, so this seems fine.

out-of-phaze (approved):

PR #3466, opened 2023-10-22
Corrects insectoid connection point.

MistakeNot4892:


out-of-phaze (changes requested):
There are now two definitions for `amputation_point` on each of the types modified in the PR. Please just have one or the other.
image
image

out-of-phaze (approved):

PR #3472, opened 2023-10-27
Rough pass on removing isSynthetic [DNM]

MistakeNot4892:
## Description of changes
Kills isSynthetic(). Opening this for input/review, it is in no state for a merge currently.

isSynthetic() is a holdover from the previous attempt at FBP and is pretty much redundant in view of the new bodytype system. Getting rid of it will make various systems more flexible and available to nonhuman synthetics and partial cyborgs.

## Authorship
Myself and @out-of-phaze.

## Changelog

### DIFF
--- a/code/_onclick/hud/human.dm
+++ b/code/_onclick/hud/human.dm
@@ -129,7 +129,7 @@
 		mymob.bodytemp.screen_loc = ui_temp
 		hud_elements |= mymob.bodytemp
 
-	if(target.isSynthetic())
+	if(GET_INTERNAL_ORGAN(target, BP_CELL))
 		target.cells = new /obj/screen()
 		target.cells.icon = 'icons/mob/screen1_robot.dmi'
 		target.cells.icon_state = "charge-empty"
--- a/code/datums/repositories/crew/binary.dm
+++ b/code/datums/repositories/crew/binary.dm
@@ -1,17 +1,16 @@
 /* Binary */
 /crew_sensor_modifier/binary/process_crew_data(var/mob/living/carbon/human/H, var/obj/item/clothing/under/C, var/turf/pos, var/list/crew_data)
 	crew_data["alert"] = FALSE
-	if(!H.isSynthetic() && H.should_have_organ(BP_HEART))
+	var/obj/item/organ/internal/cell/cell = H.should_have_organ(BP_CELL) && GET_INTERNAL_ORGAN(H, BP_CELL)
+	if(istype(cell))
+		crew_data["alert"] = cell.percent() < 10
+	else if(H.should_have_organ(BP_HEART))
 		var/obj/item/organ/internal/heart/O = H.get_organ(BP_HEART, /obj/item/organ/internal/heart)
 		if (!O || !BP_IS_PROSTHETIC(O)) // Don't make medical freak out over prosthetic hearts
 			var/pulse = H.get_pulse()
 			if(pulse == PULSE_NONE || pulse == PULSE_THREADY)
 				crew_data["alert"] = TRUE
 		if(H.get_blood_oxygenation() < BLOOD_VOLUME_SAFE)
-			crew_data["alert"] = TRUE
-	if(H.isSynthetic())
-		var/obj/item/organ/internal/cell/cell = H.get_organ(BP_CELL, /obj/item/organ/internal/cell)
-		if(!cell || cell.percent() < 10)
 			crew_data["alert"] = TRUE
 	return ..()
 
--- a/code/datums/traits/radiation_hardened.dm
+++ b/code/datums/traits/radiation_hardened.dm
@@ -1,0 +1,3 @@
+/decl/trait/radiation_hardened
+	name = "Radiation Hardened"
+	description = "While this creature can become radioactive, it will not suffer ill effects from being so."
--- a/code/game/machinery/rechargestation.dm
+++ b/code/game/machinery/rechargestation.dm
@@ -195,8 +195,8 @@
 		return (R.cell)
 	if(ishuman(M))
 		var/mob/living/carbon/human/H = M
-		if(H.isSynthetic())
-			return 1
+		if(H.should_have_organ(BP_CELL))
+			return TRUE
 		var/obj/item/rig/rig = H.get_rig()
 		if(rig)
 			return rig.cell
--- a/code/game/objects/items/weapons/implants/implants/imprinting.dm
+++ b/code/game/objects/items/weapons/implants/implants/imprinting.dm
@@ -103,11 +103,11 @@
 	. = ..()
 
 /obj/item/implant/imprinting/can_implant(mob/M, mob/user, target_zone)
-	var/mob/living/carbon/human/H = M	
+	var/mob/living/carbon/human/H = M
 	if(istype(H))
 		var/obj/item/organ/internal/B = GET_INTERNAL_ORGAN(H, BP_BRAIN)
-		if(!B || H.isSynthetic())
-			to_chat(user, "<span class='warning'>\The [M] cannot be imprinted.</span>")
+		if(!B || BP_IS_PROSTHETIC(B) || BP_IS_CRYSTAL(B))
+			to_chat(user, "<span class='warning'>\The [M] does not have an organic brain, and cannot be imprinted.</span>")
 			return FALSE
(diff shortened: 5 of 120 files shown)


code/datums/repositories/crew/binary.dm
```diff
@@ -1,18 +1,17 @@
 /* Binary */
 /crew_sensor_modifier/binary/process_crew_data(var/mob/living/carbon/human/H, var/obj/item/clothing/under/C, var/turf/pos, var/list/crew_data)
 	crew_data["alert"] = FALSE
-	if(!H.isSynthetic() && H.should_have_organ(BP_HEART))
+	var/obj/item/organ/internal/cell/cell = H.should_have_organ(BP_CELL) && GET_INTERNAL_ORGAN(H, BP_CELL)
+	if(istype(cell))
+		crew_data["alert"] = cell.percent() < 10
```
out-of-phaze, line 6: What if the cell is destroyed by an EMP?
```suggestion
    if(H.should_have_organ(BP_CELL))
	    var/obj/item/organ/internal/cell/cell = H.get_organ(BP_CELL, /obj/item/organ/internal/cell)
		crew_data["alert"] = !cell || (cell.percent() < 10)
```
Using `get_organ()` instead of `GET_INTERNAL_ORGAN()` lets us use `!cell` instead of `!istype(cell)` by taking advantage of the typecheck in `get_organ()`.

code/datums/traits/radiation_hardened.dm
```diff
@@ -0,0 +1,3 @@
+/decl/trait/radiation_hardened
+	name = "Radiation Hardened"
+	description = "While this creature can become radioactive, it will not suffer ill effects from being so."
```
out-of-phaze, line 3:
```suggestion
	description = "While this creature can become irradiated, it will not suffer ill effects from being so."
```

code/game/machinery/rechargestation.dm
out-of-phaze, line 1: Line 203 is redundant now unless there's some way to get a cell organ even if `should_have_organ(BP_CELL)` returns false. I guess for augments that could be true, but I don't think it's the case for cells.

code/modules/species/species_bodytype.dm
```diff
@@ -17,10 +18,43 @@ var/global/list/bodytypes_by_category = list()
...
+	var/list/cold_discomfort_strings = list(
+		"You feel chilly.",
+		"You shiver suddenly.",
+		"Your chilly flesh stands out in goosebumps."
+	)
+
+	var/list/traits = list() // An associative list of /decl/traits and trait level - See individual traits for valid levels
+
+	/// Damage mods
+	var/radiation_mod = 1 // Not as biologically fragile as meatboys.
```
out-of-phaze, line 57:
```suggestion
	var/radiation_mod = 1
```
Comment doesn't apply here, it looks like it was originally for the Ascent Alates?

code/_onclick/hud/human.dm
```diff
@@ -129,7 +129,7 @@
 		mymob.bodytemp.screen_loc = ui_temp
 		hud_elements |= mymob.bodytemp
 
-	if(target.isSynthetic())
+	if(GET_INTERNAL_ORGAN(target, BP_CELL))
```
PsyCommando, line 132: So, I was thinking. The code that modifies/adds UI components depending on the kind of organ in the mob could probably be transformed into some generic way for organs to specify a UI element to add to the player ui? So you don't have to have those checks for specific robot organs?

code/datums/repositories/crew/binary.dm
```diff
@@ -1,18 +1,17 @@
 /* Binary */
 /crew_sensor_modifier/binary/process_crew_data(var/mob/living/carbon/human/H, var/obj/item/clothing/under/C, var/turf/pos, var/list/crew_data)
 	crew_data["alert"] = FALSE
-	if(!H.isSynthetic() && H.should_have_organ(BP_HEART))
+	var/obj/item/organ/internal/cell/cell = H.should_have_organ(BP_CELL) && GET_INTERNAL_ORGAN(H, BP_CELL)
```
PsyCommando, line 4: Maybe an accessor on the /mob level to get the mob's vitals could be better, instead of having to dig for the right organ to get data from in the crew data stuff?

code/game/machinery/rechargestation.dm
```diff
@@ -195,8 +195,8 @@
 		return (R.cell)
 	if(ishuman(M))
 		var/mob/living/carbon/human/H = M
-		if(H.isSynthetic())
-			return 1
+		if(H.should_have_organ(BP_CELL))
```
PsyCommando, line 198: Maybe this proc could just use the base implementation of get_cell()? Since it's fairly low level. Mobs could just return the right thing themselves instead of having that big if/else type check.

code/game/objects/items/weapons/implants/implants/imprinting.dm
```diff
@@ -103,11 +103,11 @@
 	. = ..()
 
 /obj/item/implant/imprinting/can_implant(mob/M, mob/user, target_zone)
-	var/mob/living/carbon/human/H = M	
+	var/mob/living/carbon/human/H = M
 	if(istype(H))
 		var/obj/item/organ/internal/B = GET_INTERNAL_ORGAN(H, BP_BRAIN)
-		if(!B || H.isSynthetic())
-			to_chat(user, "<span class='warning'>\The [M] cannot be imprinted.</span>")
+		if(!B || BP_IS_PROSTHETIC(B) || BP_IS_CRYSTAL(B))
```
PsyCommando, line 109: I wonder if we couldn't generalize implant requirements and handle them in the base implant class? Cause most of them are just checking if an organ exist and what flags it has?

code/modules/client/preferences.dm
```diff
@@ -454,7 +454,7 @@ var/global/list/time_prefs_fixed = list()
 	character.flavor_texts["legs"] = flavor_texts["legs"]
 	character.flavor_texts["feet"] = flavor_texts["feet"]
 
-	if(!character.isSynthetic())
+	if(!character.get_bodytype().is_robotic)
```
PsyCommando, line 457: Will this error? Cause, there's no return type or ``:`` operator, and is_robotic wouldn't be defined?

code/modules/emotes/definitions/visible.dm
```diff
@@ -358,7 +358,7 @@
 	key = "vomit"
 
 /decl/emote/visible/vomit/check_user(var/mob/living/carbon/human/user)
-	. = ..() && user.check_has_mouth() && !user.isSynthetic()
+	. = ..() && user.check_has_mouth() && user.should_have_organ(BP_STOMACH)
```
PsyCommando, line 361: Maybe some organs could have a list of emotes they can provide to the organ holder instead? So no need to do checks like that? For instance if a critter has a stomach, and it has the emote for vomiting listed, it's pretty safe to say there's a way for the vomit to come out and so on?

code/modules/events/ion_storm.dm
```diff
@@ -18,7 +18,8 @@
 
 /datum/event/ionstorm/announce()
 	for(var/mob/living/carbon/S in SSmobs.mob_list)
-		if (!S.isSynthetic())
+		var/obj/item/organ/internal/brain/brain = S.should_have_organ(BP_BRAIN) && GET_INTERNAL_ORGAN(S, BP_BRAIN)
+		if(brain && !BP_IS_PROSTHETIC(brain))
```
PsyCommando, line 22: Maybe mobs could implement a proc to determine if they have anything vulnerable to a given damage type? Like EMP, Pain, Oxygen deprivation, Toxin, etc..

PsyCommando (commented):
Just a couple things I had some ideas on.

code/modules/client/preferences.dm
MistakeNot4892: `get_bodytype()` always returns an instance of /decl/bodytype - the linter is even aware of this thanks to RETURN_TYPE - and chaining calls like this is functionally identical to `:`. Might need a nullcheck though.

out-of-phaze (changes requested):
Requesting changes from a partial review I did a few weeks ago. I'll start over with a blank slate once I get back to my computer.

MistakeNot4892:
Will revisit/reopen this in single-PRs as they are completed. Have backed it up as `wip/issythetic_removal` on my fork.

PR #3473, opened 2023-10-27
Minor poster refactor.

MistakeNot4892:
## Description of changes
- Removes `/obj/item/contraband` path.
- Allows poster decls to set an icon.

Needed for scav posters and path cleanliness.

## Authorship
Myself.

## Changelog
Nothing player-facing.


out-of-phaze (approved):
This might be a good candidate for a migration script. `/obj/item/contraband/poster : /obj/item/poster{@OLD}` would be all you'd need though, so it's trivial and I really need to reorganize update_paths to not be a part of mapmerge anyway.

PR #3475, opened 2023-10-29
Allows spawn_atom to spawn invisible atoms.

MistakeNot4892:
So it turns out `input()` with a list of types checks the `see_invisible` of the mob against initial `invisibility` on the type.

image


comma (approved):

out-of-phaze (approved):
Theoretically you could avoid the text2path by passing an associative list to input() but I can never remember which way around it should be, i.e. if it displays keys and returns values or vice versa. Holding off on merge in case you want to try that.

MistakeNot4892:
Per Discord discussion I'm not sure there's a benefit to using an assoc list approach. I'm fine with the current method if there aren't any blocking objections.

PR #3476, opened 2023-10-29
Excluding unsimulated and space turfs from vessel mass calc.

MistakeNot4892:
## Description of changes
Adds `simulated = FALSE` to unsimulated turfs and excludes unsim turfs and the world area from vessel mass calc.

Mass calc is currently adding 500 mass units per unsimulated filler turf on the edge of the map.

## Authorship
@out-of-phaze found the issue and fix, I just committed it.

## Changelog
:cl:
bugfix: Vessel acceleration should be much more responsive now.
/:cl:

### DIFF
--- a/code/game/turfs/unsimulated.dm
+++ b/code/game/turfs/unsimulated.dm
@@ -1,6 +1,7 @@
 /turf/unsimulated
 	name = "command"
 	initial_gas = list(/decl/material/gas/oxygen = MOLES_O2STANDARD, /decl/material/gas/nitrogen = MOLES_N2STANDARD)
+	simulated = FALSE
 
 /turf/unsimulated/get_lumcount(var/minlum = 0, var/maxlum = 1)
 	return 0.8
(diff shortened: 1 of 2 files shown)


comma (approved):

code/game/turfs/unsimulated.dm
```diff
@@ -1,6 +1,7 @@
 /turf/unsimulated
 	name = "command"
 	initial_gas = list(/decl/material/gas/oxygen = MOLES_O2STANDARD, /decl/material/gas/nitrogen = MOLES_N2STANDARD)
+	simulated = FALSE
```
out-of-phaze, line 4: Woah woah. Nooope. `/turf/unsimulated` is just for lack of ZAS sim, isn't it? While `/turf/unsimulated/dark_filler` doesn't physically exist and so needs `simulated = FALSE`.

out-of-phaze (changes requested):
I really don't think `simulated = FALSE` on all unsim-type turfs is a good idea, we should at least discuss it further.

out-of-phaze (approved):
Discussed on Discord.

PR #3478, opened 2023-10-31
[IDB IGNORE] [MDB IGNORE] Unifying exterior and simulated flooring.

MistakeNot4892:
DNM, opening for diff and CI.

## Description of changes
- Unifies exterior and simulated flooring behavior on /turf by making turfs contain a lazy `flooring_layers` list of `/decl/flooring`.
- Moves flooring icon and smoothing behavior onto `/decl/flooring`.

## Why and what will this PR improve
Simpler code, shared code, more consistent behavior and easier maintenance.
Flooring icons no longer have two completely distinct icon state schemes.

## TODO
- [X] Coarse unification, flooring decl updates.
- [x] Verify all icons have appropriate states.
- [x] Generalize burned/broken to exterior.
- [x] Horizontal/vertical inner edge smoothing.
- [x] Horizontal/vertical outer edge smoothing.
- [x] Inner corner smoothing.
- [x] Outer corner smoothing.
- [x] Inner junction smoothing.
- [x] Outer junction smoothing.
- [x] Adjacent outer non-blending corner smoothing.
- [x] Profile/optimize to avoid bloating init time.
- [x] Add missing icon states for CI.
- [x] Verify facings are all correct for corners etc.
- [x] Correct decal directions on ninja and heist maps.
- [ ] Fix broken smoothing after inlining the checks.
- [ ] Get wild grass, water and sand working.

## Authorship
Myself.

## Changelog
:cl:
tweak: Flooring behavior has been rewritten - please report bugs with floor icons like carpet edges on the tracker.
/:cl:

PR #3479, opened 2023-10-31
Adding unit tests for turf icons and random spawner icons.

MistakeNot4892:
Noticed a ton of missing icons in #3478 so splitting it off into its own PR.


code/game/machinery/doors/airlock_subtypes.dm
```diff
@@ -18,6 +18,15 @@
...
 	stripe_color = COLOR_DEEP_SKY_BLUE
 
+/obj/machinery/door/airlock/medical/open
+	icon_state = "open"
+
+/obj/machinery/door/airlock/medical/open/Initialize()
+	set_density(FALSE)
+	set_opacity(FALSE)
+	layer = open_layer
+	return ..()
```
out-of-phaze:
```suggestion
	queue_icon_update()
	return ..()
```
So that it actually has the open icon state.

out-of-phaze (changes requested):

code/game/machinery/doors/airlock_subtypes.dm
MistakeNot4892: Generalized it to door level and added an icon update queue to airlocks.

out-of-phaze (approved):

PR #3484, opened 2023-11-02
Vitals sensors are now handled by an accessory.

MistakeNot4892:
## Description of changes
Moves suit sensors stuff onto an accessory.

Allows me to do dark things to the `/under` type.

## Authorship
Myself.

## Changelog
:cl:
tweak: Suit sensors are now an accessory on your uniform that can be removed.
/:cl:


out-of-phaze (changes requested):
Prisoner and test subject uniform sensors should use `removable = FALSE` or `canremove = FALSE`, and the accessory removal verb should check whichever one you go for. Other than that, good to go.

MistakeNot4892:
Implemented using `removable`, works in testing.

out-of-phaze (approved):
This does kind of make me wonder why we'd have the `removable` variable at all instead of just an `is_removable()` proc on accessories, but it works regardless.

PR #3487, opened 2023-11-03
Only brute and burn damage can dismantle machines.

### LINKED ISSUE #3486: Magbase EMP can destroy pipes, leading to CI failure.
#### Description of issue
The magbase away site has an EMP generator that can potentially reach a pipe. EMP damage destroys machines after the recent fixes, so this causes a leak, which fails CI apparently at random.

EMP should probably not physically destroy or disconnect pipes.

#### Steps to reproduce
Entirely random chance.

It's the pipes to the northeast of the EMP generator I believe.

Presumably since the fix that allowed damage to totally destroy machines if all components were destroyed.

Current dev.

#### Issue bingo

MistakeNot4892:
Hopefully fixes #3486.

### DIFF
--- a/code/game/machinery/_machines_base/machinery_damage.dm
+++ b/code/game/machinery/_machines_base/machinery_damage.dm
@@ -33,7 +33,7 @@
 	if(victim?.can_take_damage() && victim.is_functional())
 		amount -= victim.take_damage(amount, damtype)
 
-	if(amount)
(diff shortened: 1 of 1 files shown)


out-of-phaze (approved):
Hopefully this will at some point be superseded by a var or something on a declized damage type.

comma (approved):

PR #3490, opened 2023-11-03
Fix downward ambient lights propagation

out-of-phaze:
## Description of changes
With Lohikar's help, makes downward ambient lights propagation more robust to fix incorrect shadowing of the level below where ambiently-lit turfs shared a corner with a non-z-lighting-transparent turf.

Fixes some ugly shadowing artifacts from downward ambient light propagation.

## Authorship
Lohikar, me.


MistakeNot4892 (approved):
Requesting an approval from @Lohikar also as this is his baby

MistakeNot4892:
Going to go ahead and merge this as Lohi participated in development anyway.

PR #3491, opened 2023-11-03
Upstream dev branch features and tweaks from Lighthouse

out-of-phaze:
## Description of changes
Readds missing decals and other accidentally-cut content, adds a helper for grammatically-correct material amount strings, allows cable coil types to opt out of being recolored (for things like heavy-duty cable subtypes), optimizes random maps by about 70%, adds a console to control huge scrubbers based on ID tag (should maybe at some point make it use networks?), makes integrated circuit printers use the atom info repository for matter, and makes get_breath_from_environment check the mob's is_flooded instead of directly checking the turf. Also lets cablenets cross Z boundaries, using some of the recent mimic_edge turf work done by Loaf.

Various fixes and tweaks to either improve things directly or lay the groundwork for future expansion.

### DIFF
--- a/code/game/machinery/atmoalter/area_atmos_computer.dm
+++ b/code/game/machinery/atmoalter/area_atmos_computer.dm
@@ -1,5 +1,5 @@
 /obj/machinery/computer/area_atmos
-	name = "Area Air Control"
+	name = "area air control"
 	desc = "A computer used to control the stationary scrubbers and pumps in the area."
 	icon_keyboard = "atmos_key"
 	icon_screen = "area_atmos"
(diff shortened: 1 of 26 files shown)


code/game/machinery/atmoalter/area_atmos_computer.dm
```diff
@@ -173,3 +173,27 @@
 		status = "ERROR: No scrubber found!"
 
 	src.updateUsrDialog()
+
+/obj/machinery/computer/area_atmos/tag
```
MistakeNot4892, line 177: Would prefer something slightly more distinct than `tag` as the subtype name since `tag` is an inbuilt.

MistakeNot4892 (approved):
Looks good other than the use of `tag` as a subtype, I won't block over that but it makes me a little uneasy.

PR #3492, opened 2023-11-03
Nonhuman mob inventory rework.

MistakeNot4892:
## Description of changes
- `attack_hand()` and `attackby()` are now gated behind a dexterity flag.
- All living mobs can perform `attack_hand()` in theory. This was already the case in some contexts and the proc params are standardized so I don't expect this to cause huge issues.
- Crows, dionaea and Ascent nymphs use inventory slots now.

Less bespoke code, more standardized interactions.

## Authorship
Myself.

## Changelog
Nothing player-facing in practice.


code/modules/mob/living/simple_animal/crow/crow.dm
```diff
@@ -38,101 +35,36 @@
...
+	if(user.a_intent == I_HELP)
+		var/obj/item/backpack = get_equipped_item(slot_back_str)
+		if(backpack)
+			return backpack.attackby(I, user)
+	return ..()
 
 /mob/living/simple_animal/crow/on_update_icon()
 	..()
-	if(messenger_bag)
+	if(get_equipped_item(slot_back_str))
```
out-of-phaze: Should it maybe check the type of the equipped item?

out-of-phaze (changes requested):

code/modules/mob/living/simple_animal/crow/crow.dm
MistakeNot4892: The icon is hardcoded and indicates there's Something:tm: on their back - not much point checking type.

MistakeNot4892: Redid the logic so it checks for a crow icon in the equipped item's icon.

out-of-phaze: As mentioned on Discord, would be good for this to use the animal bodytype and slot overlay systems now that they're available on non-humantypes. Will need extra care to avoid using the fallback human icons, though.

MistakeNot4892: Implemented, seems to be working fine.

out-of-phaze (dismissed):

out-of-phaze (dismissed):

out-of-phaze (approved):

PR #3501, opened 2023-12-01
Implements can_be_injected_by().

MistakeNot4892:
## Description of changes
@PsyCommando pointed out this proc was unused earlier, so I reviewed injection code and added it where I think it was intended to go.

Slightly cleaner code.

## Authorship
@PsyCommando found the issue, I did the code.

## Changelog
Nothing player-facing.

### DIFF
--- a/code/modules/clothing/masks/smokable.dm
+++ b/code/modules/clothing/masks/smokable.dm
@@ -201,6 +201,9 @@
 	ignitermes = "<span class='notice'>USER fiddles with FLAME, and manages to light their NAME.</span>"
 	brand = "\improper Trans-Stellar Duty-free"
 
+/obj/item/clothing/mask/smokable/cigarette/can_be_injected_by(var/atom/injector)
+	return TRUE
+
 /obj/item/clothing/mask/smokable/cigarette/Initialize()
 	. = ..()
 	initialize_reagents()
--- a/code/modules/mob/living/simple_animal/familiars/familiars.dm
+++ b/code/modules/mob/living/simple_animal/familiars/familiars.dm
@@ -45,6 +45,8 @@
 	icon = 'icons/mob/simple_animal/spaceshark.dmi'
 	pixel_x = -16
 
+	offset_overhead_text_x = 16
+
 	speak_emote = list("gnashes")
 
 	health = 100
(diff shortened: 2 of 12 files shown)


MistakeNot4892:
Random pike commit too apparently.

code/modules/clothing/masks/smokable.dm
```diff
@@ -201,6 +201,9 @@
 	ignitermes = "<span class='notice'>USER fiddles with FLAME, and manages to light their NAME.</span>"
 	brand = "\improper Trans-Stellar Duty-free"
 
+/obj/item/clothing/mask/smokable/cigarette/can_be_injected_by(var/atom/injector)
+	return TRUE
```
out-of-phaze, line 205: Uh, why should this differ from `/atom/movable/can_be_injected_by(var/atom/injector)`? Now this returns true regardless of adjacency or free space. (Free space is checked in the one place `can_be_injected_by()` is called, so if it's unnecessary in the `/atom/movable` override, it should probably be removed from there.)

code/modules/mob/living/simple_animal/familiars/familiars.dm
```diff
@@ -45,6 +45,8 @@
 	icon = 'icons/mob/simple_animal/spaceshark.dmi'
 	pixel_x = -16
 
+	offset_overhead_text_x = 16
```
out-of-phaze, line 48: I assume you didn't mean to include the pike commit?

out-of-phaze (changes requested):

code/modules/mob/living/simple_animal/familiars/familiars.dm
MistakeNot4892: I didn't but it's also one line so I don't really care :U

out-of-phaze: I think when I originally wrote this comment you hadn't mentioned it and then when I posted it in the morning you had. Sorry! Yeah, it's fine.

code/modules/clothing/masks/smokable.dm
MistakeNot4892: Did some work on that, should be okay now.

out-of-phaze (approved):

PR #3502, opened 2023-12-01
Adds a lower body inventory slot. [IDB IGNORE]

MistakeNot4892:
## Description of changes
Adds a pants slot.

Allows for some more interesting layering interactions, also supports clothing schemes that aren't entirely based on jumpsuits a bit better.

## TODO
- [X] Add slot
- [X] Test slot
- [ ] Add shirt tucking
- [ ] Test layering

## Authorship
Myself.

## Changelog
:cl:
tweak: The uniform slot has been split into upper and lower body slots. Various items are now either pants items, or split into a pants and a shirt item. Jumpsuits, for example, now count as pants. This is a broad change so please report issues to the tracker.
/:cl:


MistakeNot4892:
Need to rebase this but when I do it will depend on #3798.

MistakeNot4892:
With the reworking of accessories and clothing stacking, I think this slot is probably unnecessary now. Going to close it for the time being unless a pressing need for it comes up again in the future.

PR #3503, opened 2023-12-01
Jetpacks in suit storage will now function as jetpacks.

MistakeNot4892:
Desired by democracy.

image


out-of-phaze (approved):
I was one of the no voters but the code is fine, c'est la vie.

out-of-phaze (approved):

PR #3507, opened 2023-12-08
Making tail icon gen a little less brittle.

MistakeNot4892:
## Description of changes
- Reworks tail icon retrieval to be a bit less brittle in case of missing or broken values.
- Removes the trailing `_s` from static tail states.
- Removes the redundant `tail_animation` icon var since everything can use the tail icon.
- Deletes `icons/effects/species.dmi` as it is unused and misleading.
- Adds tail validation to bodytypes.
- Somewhere amongst all this, fixes tails not rendering on staging.

make tail work better

## Authorship
Myself.

## Changelog
Nothing player-facing.


code/modules/species/species_bodytype.dm
```diff
@@ -422,3 +448,9 @@ var/global/list/bodytypes_by_category = list()
 		var/organ_type = replacing_organs[organ_tag]
 		var/obj/item/organ/internal/new_innard = new organ_type(limb.owner, null, limb.owner.dna, src)
 		limb.owner.add_organ(new_innard, GET_EXTERNAL_ORGAN(limb.owner, new_innard.parent_organ), FALSE, FALSE)
+
+/decl/bodytype/proc/get_typical_user_species()
```
out-of-phaze: If this is only used for validation, I'd like that to be specified—something like `get_validation_user_species()` instead.

out-of-phaze (dismissed):

MistakeNot4892:
Why the heck did I commit this straight to `staging`.

out-of-phaze (approved):

PR #3508, opened 2023-12-09
Holopad ID uniqueness pass.

MistakeNot4892:
## Description of changes
- Changes holopads to use an instance variable instead of area proper_name.
- Adds a uniqueness test for holopad IDs.

Prevents overlapping area names from making holopads inaccessible.
In theory allows for multiple holopads in the same area.

## Authorship
Myself.

## Changelog
Nothing player-facing.

### DIFF
--- a/code/game/machinery/hologram.dm
+++ b/code/game/machinery/hologram.dm
@@ -30,17 +30,17 @@ Possible to do for anyone motivated enough:
 #define AREA_BASED 6
 
 var/global/const/HOLOPAD_MODE = RANGE_BASED
+var/global/list/holopads = list()
 
 /obj/machinery/hologram/holopad
 	name = "\improper holopad"
 	desc = "It's a floor-mounted device for projecting holographic images."
 	icon = 'icons/obj/machines/holopad.dmi'
 	icon_state = "holopad-B0"
-
 	layer = ABOVE_TILE_LAYER
+	idle_power_usage = 5
 
 	var/power_per_hologram = 500 //per usage per hologram
-	idle_power_usage = 5
 
 	var/list/mob/living/silicon/ai/masters = new() //List of AIs that use the holopad
 	var/last_request = 0 //to prevent request spam. ~Carn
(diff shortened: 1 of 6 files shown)


comma (dismissed):

code/game/machinery/hologram.dm
```diff
@@ -58,10 +58,24 @@ var/global/const/HOLOPAD_MODE = RANGE_BASED
...
-	desc = "It's a floor-mounted device for projecting holographic images. Its ID is '[A.proper_name]'"
+
+	// Null ID means we want to use our area name.
+	global.holopads += src
+	if(isnull(holopad_id))
+		var/area/A = get_area(src)
+		holopad_id = A?.proper_name || "Unknown"
+
+	// For away sites, always tag the sector name so we have a unique discriminator for long range calls.
+	var/obj/effect/overmap/visitable/sector = global.overmap_sectors[num2text(z)]
```
out-of-phaze, line 73: Is it guaranteed that the main map doesn't have a sector/that this will only apply to away sites, as stated in the comment? (I have a map downstream that does, so if that's not allowed I need to fix that.)

maps/away/errant_pisces/errant_pisces_areas.dm
```diff
@@ -86,7 +86,7 @@
 	icon_state = "fishing_wing"
 
 /area/errant_pisces/bridge
-	name = "Bridge"
+	name = "Errant Pisces Bridge"
```
out-of-phaze: Why not use the sector name logic for ships as well, so that the name of the ship is added instead of `Errant Pisces`? Because that's not a player-facing name, `CV Ahab's Harpoon` is.

out-of-phaze (dismissed):

code/game/machinery/hologram.dm
MistakeNot4892: Well I meant 'away sites' as 'overmap sites', the comment is just slightly too narrow. Not sure what you're asking re: 'allowed', if there's no overmap sector it just won't modify the name.

maps/away/errant_pisces/errant_pisces_areas.dm
MistakeNot4892: The name logic is used for ships, this will end up as `CV Ahab's Harpoon - Errant Pisces Bridge`, so I'll fix that.

out-of-phaze: It's used only for sectors. Compare `/obj/effect/overmap/visitable/sector` and `/obj/effect/overmap/visitable/ship`. `overmap_sectors` contains any subtype of `/obj/effect/overmap/visitable`, I believe.

code/game/machinery/hologram.dm
out-of-phaze: The variable is named sector, I didn't realize. I thought it was the `/obj/effect/overmap/visitable/sector` type. Whoops.

maps/away/errant_pisces/errant_pisces_areas.dm
out-of-phaze: Nevermind, my bad.

out-of-phaze:
please rebase from dev onto stable

comma (approved):

out-of-phaze (approved):

PR #3531, opened 2023-12-16
Generalized the atom temperature system and removed restrictions.

MistakeNot4892:
## Description of changes
- Removed `ATOM_FLAG_NO_TEMP_CHANGE` in favour of checking `simulated`.
- `ProcessAtomTemperature()` will now enqueue physically contained atoms on parent atom temperature change.
- Renamed `HandleObjectHeating()` to `handle_external_heating()` and moved it to `/atom`.
- Bypassed heating message in `handle_external_heating()` if source and user aren't supplied.
- `handle_external_heating()` calls itself on physically contained external atoms within an atom.
- `fire_act()` now calls `handle_external_heating()` on affected atoms.

Causes heat to propagate to contents more consistently and lays the groundwork for fire and heat causing atoms to melt in the future.

## Authorship
Myself.

## Changelog
:cl:
tweak: Heating atoms like beakers with fire or a welding torch should be more consistent now.
/:cl:


code/game/objects/items/weapons/flame.dm
```diff
@@ -5,11 +5,6 @@
 	var/lit = FALSE
 	material = /decl/material/solid/organic/wood
 
-/obj/item/flame/afterattack(var/obj/O, var/mob/user, proximity)
```
out-of-phaze, line 8: Is this handled somewhere else now?

out-of-phaze: I see, item-level. I hope this won't fuck with things like igniting rags on bottles or lighting cigarettes.

code/game/atoms_temperature.dm
```diff
@@ -41,10 +58,19 @@
...
 		var/altered_temp = max(temperature + (ATOM_TEMPERATURE_EQUILIBRIUM_CONSTANT * temperature_coefficient * diff_temp), 0)
 		ADJUST_ATOM_TEMPERATURE(src, (diff_temp > 0) ? min(adjust_temp, altered_temp) : max(adjust_temp, altered_temp))
 	else
 		temperature = adjust_temp
-		return PROCESS_KILL
+		. = PROCESS_KILL
+
+	// If our temperature changed, our contents probably want to know about it.
+	if(temperature != old_temp)
+		var/my_atoms = get_contained_external_atoms()
```
out-of-phaze: Iiiiiii kind of think it might be good for this to use a separate helper, so things like insulated containers can choose not to affect their contents, or remove specific parts. Also, this would prevent components from heating up, I think? Which might not be desirable if we want to add actual heat-draining functionality to the um, heatsink component, or add fan components, etc.
Maybe it could be like `get_contained_matter()` and internally default to `get_contained_external_atoms()`?

out-of-phaze: Also, this might be partially redundant given `handle_external_heating` propagates heating to the stuff inside of it.

out-of-phaze (commented):
A related aside: Is it possible to set `/obj/screen` and similar types to have `simulated = FALSE` in order to avoid unnecessary temperature processing? Especially for things like the typing indicator, which IIRC is in mob contents and therefore can be heated by things that heat the mob?

code/game/objects/items/weapons/flame.dm
MistakeNot4892: Per Discord, igniting rags etc is attackby IIRC. Should be fine.

code/game/atoms_temperature.dm
MistakeNot4892: This just enqueues, `handle_external_heating` propagates temperature change then enqueues. Could maybe remove it from `handle_external_heating`.

MistakeNot4892: Trimmed it out of `handle_external_heating()` as it will be propagated as soon as the atom calls `ProcessAtomTemperature()`. Also added `get_contained_temperature_sensitive_atoms()` which defaults to `get_contained_external_atoms()`.

code/game/objects/items/weapons/flame.dm
MistakeNot4892: Yep, attackby on the rag and on the bottle.

out-of-phaze (approved):

PR #3535, opened 2023-12-17
ID card retrieval rework.

### LINKED ISSUE #3461: Serpentids cannot access nanoUI programs requiring access
#### Description of issue
Apparently, `/atom/movable/proc/GetIdCard()` works in such a way that the only ID Card worth considering for Serpentids is their system controller alien chip (with the only access of `ACCESS_ASCENT`). 

Expected: Serpentids should be able to use modular computers with their job accesses.
Factual: Serpentids do not get any accesses to modular computer programs from their ID cards.

#### Steps to reproduce
1. Login as a Serpentid Captain
2. Try to open "Command and Communications program"

Obviously, `OS flashes "Access Denied" warning`.

I suspect issues with  `/atom/movable/proc/GetIdCards()` ordering.

No idea.

Local `Nebula:stable`. 

#### Issue bingo

MistakeNot4892:
## Description of changes
- Collapses ID card retrieval procs a bit.
- Ensures that `exceptions` is actually passed through the chain and consistently checked.
- Semi-sorts mob ID cards so GetIDCard() will prefer to return held cards, then equipped cards, then any remaining cards (ID implants).

Fixes #3461.

## Authorship
Myself.

## Changelog
Nothing player-facing.

### DIFF
--- a/code/_helpers/lists.dm
+++ b/code/_helpers/lists.dm
@@ -77,10 +77,12 @@
 
 //Checks for specific types in a list
 /proc/is_type_in_list(var/atom/A, var/list/L)
+	if(!length(L) || !istype(A))
+		return FALSE
 	for(var/type in L)
 		if(istype(A, type))
-			return 1
-	return 0
+			return TRUE
+	return FALSE
 
 //Checks for specific paths in a list
 /proc/is_path_in_list(var/path, var/list/L)
--- a/code/_helpers/lists.dm
+++ b/code/_helpers/lists.dm
@@ -76,20 +76,22 @@
 	return counting_english_list(input, output_icons, determiners, nothing_text, and_text, comma_text, final_comma_text)
 
 //Checks for specific types in a list
-/proc/is_type_in_list(var/atom/A, var/list/L)
-	if(!length(L) || !istype(A))
+/proc/is_type_in_list(datum/thing, list/type_list)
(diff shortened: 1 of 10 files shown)


code/_helpers/lists.dm
```diff
@@ -76,18 +76,22 @@
...
+	return FALSE
 
 //Checks for specific paths in a list
-/proc/is_path_in_list(var/path, var/list/L)
-	for(var/type in L)
-		if(ispath(path, type))
-			return 1
-	return 0
+/proc/is_path_in_list(var/check_path, list/type_list)
+	if(!length(type_list) || !ispath(check_path))
```
out-of-phaze, line 89: `for(var/foo in bar)` includes an implicit null/length check on `bar`, this is just unnecessarily verbose in a proc that tends to be used in hot code like smoothing. `!ispath(check_path)` is fine though.
```suggestion
	if(!ispath(check_path))
```

out-of-phaze (approved):

code/_helpers/lists.dm
out-of-phaze: I meant to request changes but forgot, but like, this isn't a big deal either way.

MistakeNot4892: Per Discord this was a microopt for cases where the list is null or empty.

PR #3544, opened 2023-12-26
Updates space turfs in radius of simulated/exterior turfs during init.

### LINKED ISSUE #2746: Space turfs do not update their ambient light when a dynamically lit turf is created beside them.
#### Description of issue
See title.

Space turfs only generate/apply ambient light when they are bordering a dynamically lit turf, as otherwise you would have thousands of space turfs pointlessly creating corners etc. However as it stands, they should be updated when a dynamically lit turf is created beside them.

#### Steps to reproduce
- Enable starlight.
- Go to space.
- Spawn a turf.
- Put a non-fullbright light source on the turf or nearby.
- Observe that the edges are dark as though bordering a fulldark turf.

I believe this was previously handled in simulated turf init, but not sure.

Presumably since ambient lights were added..

Current dev.

#### Issue bingo

MistakeNot4892:
## Description of changes
Simulated and exterior turfs iterate neighboring space turfs and queue them in SSambience. Init times seem fine on tradeship (19s which is the normal init time on my hardware) but I haven't benchmarked exodus or the other maps.

- Fixes #2746

## Authorship
Myself.

## Changelog
Nothing player-facing.

### DIFF
--- a/code/game/turfs/exterior/_exterior.dm
+++ b/code/game/turfs/exterior/_exterior.dm
@@ -43,6 +43,11 @@
 
 	. = ..(mapload)	// second param is our own, don't pass to children
 
+	if(!mapload)
+		// Force neighboring space to update corners.
+		for(var/turf/space/space in RANGE_TURFS(src, 1))
+			SSambience.queued |= space
+
 	if (no_update_icon)
 		return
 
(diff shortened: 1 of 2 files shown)


MistakeNot4892:
Need to verify that this actually does anything, I seem to have misunderstood the issue.

MistakeNot4892:
Will revisit this when I can confirm it actually does the thing.

PR #3555, opened 2024-01-05
Staging fixes

MistakeNot4892:
## Description of changes
- Adds contents to obj matter cost calc for recipes/atom info repo. Should mean recipes are more accurate when considering matter costs.
- Adds arguments to human Initialize() overrides to try and preserve the chain due to some null bodytype runtimes in equip code. Also tweaks some overrides like corpses.
- Adds `?` to several `get_bodytype()` calls due to the above.
- Removes or rewrites several other `get_bodytype()` calls.

Fixes some stuff.

## Authorship
Myself for the human stuff and implementation, thanks to 54_23 for pointing out the issues with matter costs in recipes.

## Changelog
:cl:
tweak: Some crafting and fabricator recipe costs may have changed, please report inconsistencies on the tracker.
/:cl:


out-of-phaze:
Didn't we discuss that there was a reason contents weren't included in it? If it turns out that was resolved, then this looks good to go.

code/modules/fabrication/designs/_design.dm
```diff
@@ -57,14 +57,26 @@
...
 			var/decl/material/M = GET_DECL(material)
 			if(istype(M))
-				.[M.type] = matter[material]
+				.[M.type] += matter[material]
 	if(reagents && length(reagents.reagent_volumes))
+		LAZYINITLIST(.)
 		for(var/R in reagents.reagent_volumes)
-			.[R] = FLOOR(REAGENT_VOLUME(reagents, R) / REAGENT_UNITS_PER_MATERIAL_UNIT)
+			.[R] += FLOOR(REAGENT_VOLUME(reagents, R) / REAGENT_UNITS_PER_MATERIAL_UNIT)
+	for(var/obj/thing in contents)
```
out-of-phaze: Should this be `contents` or `get_contained_external_atoms()`? Note that this issue is why I made `get_contained_matter` for calculating this type of thing at runtime: it has some overlap with the objects in `get_contained_external_atoms` but adds/removes extra ones, like machine parts, etc.

MistakeNot4892:
> Didn't we discuss that there was a reason contents weren't included in it? If it turns out that was resolved, then this looks good to go.

I don't remember the reason. :(

code/modules/fabrication/designs/_design.dm
MistakeNot4892: Per discussion on Discord, I've merged `get_contained_matter()` and `building_cost()` as both of them had gaps, but were trying to achieve the same goal at different points in time. Going to see if it impacts the init numbers or causes recipes to fail CI but I think it should address this issue.

MistakeNot4892:
That's all the runtimes from the staging test.

code/game/atoms.dm
```diff
@@ -331,6 +331,19 @@
 	SHOULD_CALL_PARENT(FALSE) //Don't call the stub plz
 	return
 
+/**
+ * Returns the sum of this atoms's reagents plus the combined matter of all its contents.
+ * Obj adds matter contents. Other overrides may add extra handling for things like material storage.
+ * Most useful for calculating worth or deconstructing something along with its contents.
+ */
+/atom/proc/get_contained_matter()
+	if(length(reagents?.reagent_volumes))
```
out-of-phaze, line 340:
```
	if(!simulated || INSTANCE_IS_ABSTRACT(src))
		return
```
used to be here in `building_cost()`, is it still necessary?

MistakeNot4892: No, I don't think it was necessary initially. Unsimulated atoms shouldn't be being created by the atom repo or passed back to contained external atoms anyway.

out-of-phaze (approved):

PR #3558, opened 2024-01-06
Various map object from Polaris map conversion.

MistakeNot4892:
## Description of changes
Adds various objects (and repaths a few) used for the Polaris/Crux maps.

- Adds smartfridges for produce and material stacks.
- Adds box of flares.
- Adds several towel subtypes.
- Adds a random pill bottle spawner.
- Adds a marble bench subtype.
- Adds various job closets (pilot, medical L2, xenoflora, sidearm cabinet)
- Adds non-cargo exotic seed crate.
- Adds some modular computer presets.
- Adds a sugar-iron pill bottle.

Adds generally useful mapping objects.

## Authorship
Much of the new object code is adapted from Polaris.

## Changelog


code/game/objects/structures/crates_lockers/closets/walllocker.dm
```diff
@@ -28,4 +28,16 @@
...
-	)
\ No newline at end of file
+	)
+
+/obj/structure/closet/walllocker/emerglocker
+	name = "emergency locker"
+	desc = "A wall mounted locker with emergency supplies."
+	closet_appearance = /decl/closet_appearance/wall/emergency
+
+/obj/structure/closet/walllocker/emerglocker/WillContain()
```
out-of-phaze: If you look at how this type is implemented on Polaris, it's basically just `/obj/structure/emergency_dispenser`. Is this change (to be an actual closet) intentional?

code/game/turfs/flooring/flooring_shuttle.dm
```diff
@@ -3,6 +3,10 @@
 	icon = 'icons/turf/shuttle.dmi'
 	abstract_type = /turf/simulated/floor/shuttle
 
+/turf/simulated/floor/shuttle/plating
+	icon_state = "plating"
```
out-of-phaze, line 7: Why is this necessary? Does `/turf/simulated/floor` not work? Nothing does `istype(foo, /turf/simulated/floor/shuttle)` and there's no logic defined on this either.

code/game/objects/structures/crates_lockers/closets/secure/scientist.dm
```diff
@@ -60,3 +60,26 @@
 		/obj/item/storage/box/syringes,
 		/obj/item/chems/glass/bottle/sedatives
 	)
+
+/obj/structure/closet/secure_closet/pilot
+	name = "pilot locker"
+	req_access = list(access_xenobiology)
```
out-of-phaze, line 66: Is this intentional, or bad copypaste?

out-of-phaze (changes requested):
After these are addressed this would be good to merge, however I'm going to be a little nitpicky. Can you add a file named something like `3558_lowercase_closets.txt` in `/tools/map_migrations` (create the folder, it doesn't exist upstream) with the contents
```
/obj/structure/closet/secure_closet/CMO : /obj/structure/closet/secure_closet/cmo{@OLD}
/obj/structure/closet/secure_closet/RD : /obj/structure/closet/secure_closet/research_director{@OLD}
```
in order to help downstreams apply the repaths in this PR to their own maps?

code/game/objects/structures/crates_lockers/closets/secure/scientist.dm
MistakeNot4892: Intentional, we have no pilot access.

code/game/objects/structures/crates_lockers/closets/walllocker.dm
MistakeNot4892: I didn't realize the emergency dispenser existed.

code/game/turfs/flooring/flooring_shuttle.dm
MistakeNot4892: It was for a shuttle, so I decided to use the same root type until I can find time to go through and work out why shuttle floors need their own type at all.

code/game/objects/structures/crates_lockers/closets/walllocker.dm
out-of-phaze: They're terribly written but almost exactly the same as the emerglocker type, just without the crowbar.

out-of-phaze (dismissed):

out-of-phaze (approved):

PR #3562, opened 2024-01-06
Merges antitoxin pills and pill bottles.

MistakeNot4892:
We had two antitox pills/bottles that only differed in per-pill dosage. I unified them in favour of the larger dosage per pill.


out-of-phaze:
I vaguely recall there being a reason for there to be two dosages? Let me check

out-of-phaze (approved):

PR #3563, opened 2024-01-07
Exterior turf and shuttle fixes

NataKilar:
## Description of changes
Adjustments and fixes to ZAS and exterior/open turf logic. Also fixes some issues with shuttles while I'm in the area.

**ZAS**
* Unsimulated connection edges now keep track of the connecting turfs in the simulated zone in their ``inner_turfs`` list
* ``inner_turfs`` are now marked for update on zone merges. This fixes an issue where unsimulated connection edges were not retained post zone merge

**Exterior turf logic**
* ``update_external_atmos_participation`` should now be called whenever the external status of a turf may change. Currently, this is when ``set_outside`` is called directly, when the area of the turf changes, or if a turf above changes
* In most situations, turfs' ``air`` variable will be updated with the exterior mix when changing from exterior to ZAS participating
* Fixes issue where turfs on planets were automatically becoming pressurized when a non-external area was set

**Open turf logic**
* The ``open_turf_type`` variable on the bottom most turf in a Z-stack will now determine the type of open turf created when a turf above it is destroyed in most situations. The ``open_turf`` variable on the area will still take precedence if present.
* When the turf at the bottom of a Z-stack is changed, it will attempt to update all open turfs of its previous ``open_turf_type`` to its new ``open_turf_type``

**Shuttle fixes**
* Shuttles now have a ``force_ceiling_on_init`` variable which will overwrite any ceiling directly above it on Initialization. This is necessary because space turfs automatically create plating above the shuttle when they see non-dense spaces below them in ``LateInitialize``, which then must be overwritten
* Shuttles no longer create ceilings over background turfs (e.g. the catwalk tiles on the Bee)
* Fixes shuttle walls magically becoming reinforced when moving

Makes exterior and ZAS-participating turfs behave as expected. Holes in ceilings into space now act properly, and buildings on planets should not become magically pressurized.

The shuttle ceiling issue is mostly a pet peeve, since currently the Bee leaves behind an ugly plating outline above it when it takes off. I set ``force_ceiling_on_init`` to be ``TRUE`` by default, so any shuttles which are mapped to have ceilings immediately above them need to be adjusted. This isn't a perfect solution. Ideally I would simply have space turfs not turn into plating if they are above a shuttle, but I can't think of a clean way to do this currently due to init order

The open turf changes are mostly to make creation and removal of space turfs more consistent. Space turfs are set to have themselves as their ``open_turf_type``, so as long as there is space at the bottom, there will be space through the entire Z-stack. On the other hand, plating will always have ``/turf/simulated/open`` above it, rather than space, unless it has been mapped otherwise. In the short future, we should be able to unify ``/turf/exterior/open`` and ``/turf/simulated/open``, so this change shouldn't have much of an effect on multi-Z planets.  In the event that a turf cannot find an open turf type from the bottom of its Z-stack or its area, it will still fall back on its own ``open_turf_type``. Overall, I believe this is a better solution than having every turf deciding its own open turf type, as this can lead to inconsistencies in the Z-stack. Happy to discuss this change further though if more flexibility is needed or I'm missing an edge case.

Targeting staging for this, since these are mostly fixes

## Authorship
Myself. Thanks to Loaf for helpful discussions

### DIFF
--- a/code/game/base_turf.dm
+++ b/code/game/base_turf.dm
@@ -13,12 +13,9 @@
 			return A.open_turf
 
 		// Find the first non-open turf below and use its open_turf_type.
-		var/turf/below = T
-		while ((below = GetBelow(below)))
-			if(!below.is_open() || !HasBelow(below.z))
-				if(below.open_turf_type)
-					return below.open_turf_type
-				break
+		var/z_stack_type = get_open_turf_type(T)
+		if(z_stack_type)
+			return z_stack_type
 
 		// Otherwise, default to the open turf type set on the turf being removed.
 		if(T.open_turf_type)
@@ -26,6 +23,17 @@
 	if(istype(A) && A.base_turf)
 		return A.base_turf
 	return get_base_turf(T.z)
+
+// Returns the open turf of a Z-stack by finding the nearest non-open turf below.
+/proc/get_open_turf_type(var/turf/T)
+	if(!HasBelow(T.z))
+		return
+	var/turf/below = T
+	while ((below = GetBelow(below)))
+		if(!below.is_open() || !HasBelow(below.z))
+			if(below.open_turf_type)
+				return below.open_turf_type
+			return
 
 /client/proc/set_base_turf()
 	set category = "Debug"
--- a/code/game/base_turf.dm
+++ b/code/game/base_turf.dm
@@ -11,6 +11,16 @@
 	if(HasBelow(T.z))
 		if(istype(A) && A.open_turf)
 			return A.open_turf
+
+		// Find the first non-open turf below and use its open_turf_type.
+		var/turf/below = T
+		while ((below = GetBelow(below)))
+			if(!below.is_open() || !HasBelow(below.z))
+				if(below.open_turf_type)
+					return below.open_turf_type
+				break
+
+		// Otherwise, default to the open turf type set on the turf being removed.
 		if(T.open_turf_type)
 			return T.open_turf_type
 	if(istype(A) && A.base_turf)
--- a/code/game/turfs/space/space.dm
+++ b/code/game/turfs/space/space.dm
@@ -11,6 +11,8 @@
 	z_eventually_space = TRUE
 	turf_flags = TURF_FLAG_BACKGROUND
 
+	open_turf_type = /turf/space
+
 	/// If we're an edge.
 	var/edge = 0
 	/// Force this one to pretend it's an overedge turf.
@@ -80,6 +82,7 @@
 		var/turf/T = src
 		while ((T = GetAbove(T)))
 			T.z_eventually_space = FALSE
+
 	return ..()
 
 /turf/space/LateInitialize()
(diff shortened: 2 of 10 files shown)


code/game/turfs/space/space.dm
```diff
@@ -241,8 +244,8 @@
 					A.loc.Entered(A)
 	return
 
-/turf/space/ChangeTurf(var/turf/N, var/tell_universe = TRUE, var/force_lighting_update = FALSE, var/keep_air = FALSE)
-	return ..(N, tell_universe, TRUE, keep_air)
+/turf/space/ChangeTurf(var/turf/N, var/tell_universe = TRUE, var/force_lighting_update = FALSE, var/keep_air = FALSE, var/keep_air_below = FALSE, var/update_open_turfs_above = TRUE)
+	return ..(N, tell_universe, TRUE, keep_air, keep_air_below, update_open_turfs_above)
```
MistakeNot4892, line 248: This could just be `return ..(force_lighting_update = FALSE)` if I'm reading correctly.

code/game/base_turf.dm
```diff
@@ -11,6 +11,16 @@
 	if(HasBelow(T.z))
 		if(istype(A) && A.open_turf)
 			return A.open_turf
+
+		// Find the first non-open turf below and use its open_turf_type.
+		var/turf/below = T
+		while ((below = GetBelow(below)))
```
MistakeNot4892: Non-blocking note, it might be good to have a helper to get a type from up or down the z-stack.

code/game/turfs/turf.dm
```diff
@@ -445,7 +468,7 @@
 
 	// If we are in a multiz volume and not already inside, we return
 	// the outside value of the highest unenclosed turf in the stack.
-	if((. != OUTSIDE_NO) && HasAbove(z))
```
MistakeNot4892, line 448: If I remember correctly a hard `OUTSIDE_NO` means that the turf specifically has been given a roof - what's the requirement for removing the check here?

NataKilar: Most commonly, ``. == OUTSIDE_NO`` here because all space and non-planet areas are set to ``is_outside = OUTSIDE_NO`` (see the check above). If the check is left in place, turfs are instantly set to non-outside as soon as their area is changed to a player made or station area, regardless of the presence of a roof. Nothing really behaves as you would expect if this check is left in place.

The downside of removing this check is you're pretty much required to map in a roof (or in space, have the game initialize a roof for you) if the turfs aren't at the top of the Z-stack. I can try to add flexibility to this if you request, but I'm not really sure how to present a roofed turf that doesn't actually have a roof with regards to player mobility and Z-mimic.

code/game/turfs/space/space.dm
NataKilar: ``force_lighting_update = TRUE`` I think, but sounds good

NataKilar: Actually, this seemed to not pass the arguments up correctly when I tried it.

MistakeNot4892: Per Discord, it seems to work fine in Wrench, so unsure.
EDIT: turns out it drops all other arguments, rip.

MistakeNot4892 (approved):

PR #3564, opened 2024-01-07
minor stable fixes

43082ujfgrj:
fixed deconstruction extension typo
You start welding the Canister [He].
Jerry Auman unwelds /decl/machine_construction/pipe/welder.
You switch the welding gun off.

runtime in stirling process
[00:47:30] Runtime in stirling.dm,48: Division by zero
  proc name: Process (/obj/machinery/atmospherics/binary/stirling/Process)
  src: the stirling engine (/obj/machinery/atmospherics/binary/stirling)
  src.loc: the floor (40,59,3) (/turf/simulated/floor/tiled/monotile)
  call stack:
  the stirling engine (/obj/machinery/atmospherics/binary/stirling): Process(20)
  the stirling engine (/obj/machinery/atmospherics/binary/stirling): ProcessAll(20)
  Machines (/datum/controller/subsystem/machines): process machinery(0)
  Machines (/datum/controller/subsystem/machines): fire(0)
  Machines (/datum/controller/subsystem/machines): ignite(0)

fix the default pistol safety icon being reversed (glows green when safety is off)

fixed capacitor weapons runtime (happens when you choose null in the dial)
[01:05:43] Runtime in capacitor.dm,152: Cannot read null.name
  usr: Jerry Auman (/mob/living/carbon/human) ([0x300002e]) (the floor) (46,23,3 - Tradeship Ocelot Alpha) (/turf/simulated/floor/tiled/techfloor/grid) (a57b4a72o7)
  usr.loc: The floor (/turf/simulated/floor/tiled/techfloor/grid) ([0x10149dd]) (46,23,3 - Tradeship Ocelot Alpha) (/area/ship/trade/maintenance/power)

fixed a bunch of capacitor rifle overlays (misnamed and missing icon states plus upd proc), also added glowy icons when the rifle is on back

fixed power cell overlays (clarification: previously they had no charge indicator)


code/modules/power/cell.dm
```diff
@@ -51,6 +50,7 @@
 
 	if(new_overlay_state != overlay_state)
 		overlay_state = new_overlay_state
+		cut_overlays()
```
MistakeNot4892: It would be preferable to keep the parent call and just remove the state checking, the `..()` applies blood overlays and contamination overlays to the atom.

code/modules/projectiles/guns/energy/capacitor.dm
```diff
@@ -224,6 +224,7 @@ var/global/list/laser_wavelengths
 		M.update_inv_hands()
 
 /obj/item/gun/energy/capacitor/adjust_mob_overlay(var/mob/living/user_mob, var/bodytype,  var/image/overlay, var/slot, var/bodypart)
+	. = ..()
```
MistakeNot4892: Per the comments on `/obj/item/proc/adjust_mob_overlay`, you have to call `..()` after making any changes to the overlay, because the root proc does a bunch of adjustment logic.

MistakeNot4892 (dismissed):
Other changes look fine, just need the parent calls to be addressed.

43082ujfgrj:
> Other changes look fine, just need the parent calls to be addressed.

I changed the power cell icon code as you said, it's working fine.

But if I move the parent call to the end (as it was earlier) all the overlays on it just break, hence why I moved it up (because the parent call on guns adds a "-wielded" to it and checking it in capacitor rifle code is not good)

image

43082ujfgrj:
fixed cell overlays, waiting for icon system resolve idk

MistakeNot4892:
I opened a PR to your branch with a fix for the ordering issue with cap rifle overlays.

MistakeNot4892 (approved):

PR #3574, opened 2024-01-08
Added paint stripper.

MistakeNot4892:
## Description of changes
This is a very anemic feature PR. Adds paint stripper, which works like paint but backwards.

At the moment there's no way to properly reset the colour changes resulting from splashing a mob with paint. You can use white paint, but it's not perfect. This adds the bare bones of a paint stripper chem to do it properly.

## Authorship
Myself.

## Changelog
:cl:
add: Added paint stripper.
/:cl:

### DIFF
--- a/code/game/atoms.dm
+++ b/code/game/atoms.dm
@@ -743,9 +743,27 @@
 /atom/proc/get_color()
 	return color
 
-/// Set the color of this atom to `new_color`.
-/atom/proc/set_color(new_color)
-	color = new_color
+/* Set the atom colour. This is a stub effectively due to the broad use of direct setting. */
+// TODO: implement this everywhere that it should be used instead of direct setting.
+/atom/proc/set_color(var/new_color)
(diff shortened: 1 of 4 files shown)


PsiOmegaDelta (dismissed):

code/game/atoms.dm
```diff
@@ -743,9 +743,27 @@
...
+/atom/proc/set_color(var/new_color)
+	if(isnull(new_color))
+		return reset_color()
+	if(color != new_color)
+		color = new_color
+		return TRUE
+	return FALSE
+
+/atom/proc/reset_color()
+	if(color != COLOR_WHITE)
```
out-of-phaze: i am pretty sure that setting color to `COLOR_WHITE` will set color to null actually, so it might assign it unnecessarily if color is already null. not a big deal though

out-of-phaze (dismissed):

out-of-phaze (approved):

PR #3575, opened 2024-01-08
Adjusts human mob init args.

MistakeNot4892:
Turns out the previous pattern drops all the other args. Thanks DM.

### DIFF
--- a/code/modules/mob/living/carbon/human/human_species.dm
+++ b/code/modules/mob/living/carbon/human/human_species.dm
@@ -3,12 +3,12 @@
 	status_flags = GODMODE|CANPUSH
 	virtual_mob = null
 
-/mob/living/carbon/human/dummy/mannequin/Initialize()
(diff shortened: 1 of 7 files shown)


comma (approved):

code/modules/mob/living/carbon/human/human_species.dm
```diff
@@ -58,7 +58,8 @@
 /mob/living/carbon/human/monkey
 	gender = PLURAL
 
-/mob/living/carbon/human/monkey/Initialize()
+/mob/living/carbon/human/monkey/Initialize(mapload, species_name, datum/dna/new_dna, decl/bodytype/new_bodytype)
 	if(gender == PLURAL)
 		gender = pick(MALE, FEMALE)
-	. = ..(species_name = SPECIES_MONKEY)
```
out-of-phaze, line 64: FWIW, this specific pattern was the only thing that dropped all the other arguments, I think. Most of the other overrides don't technically need the arguments but there's no reason not to have them now that you've added them.

out-of-phaze (approved):

PR #3583, opened 2024-01-12
Refactoring human limb icon generation.

MistakeNot4892:
## Description of changes
- Simplifies human icon gen to use bodypart keys instead of generating a bespoke key.
- Simplifies limb icon gen to minimize extra icon ops and properly use caching.
- Generalizes some overlay and marking handling.

Cleaner code. Probably slightly faster? Needed for #3191.

## Authorship
Myself.

## Changelog
Nothing player-facing.

### DIFF
--- a/code/game/objects/alien_props.dm
+++ b/code/game/objects/alien_props.dm
@@ -8,13 +8,13 @@
 	icon_state = "unknown1"
 	maxcharge = 5000
 	origin_tech = @'{"powerstorage":7}'
-	var/static/base_icon
+	var/base_icon_state
 
 /obj/item/cell/alien/on_update_icon()
 	. = ..()
-	if(!base_icon)
-		base_icon = pick("instrument", "unknown1", "unknown3", "unknown4")
-	icon_state = base_icon
+	if(!base_icon_state)
+		base_icon_state = pick("instrument", "unknown1", "unknown3", "unknown4")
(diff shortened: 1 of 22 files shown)


code/game/objects/alien_props.dm
```diff
@@ -8,13 +8,13 @@
 	icon_state = "unknown1"
 	maxcharge = 5000
 	origin_tech = "{'powerstorage':7}"
-	var/static/base_icon
+	var/base_icon_state
```
out-of-phaze, line 11: I don't understand why this change was made, I think the original idea was that all alien cells in a round would have the same random icon. If the intent is to change that, that's fine I guess, I just don't know why it's in this PR.

out-of-phaze: Oh, I see. Every instance of `base_icon` was audited.

out-of-phaze (dismissed):

code/game/objects/alien_props.dm
MistakeNot4892: Yeah sorry, I wanted to make sure I got every instance so renamed for clarity.

out-of-phaze (dismissed):

out-of-phaze (dismissed):

out-of-phaze (approved):

deathride58 (approved):

PR #3596, opened 2024-01-18
Fix temperature sim causing infinite processing

out-of-phaze:
## Description of changes
Adds a variable to subsystems, `_internal_name`, for better bookkeeping with `is_processing`. Without this, the conversion from variable name to `is_processing` is one-way and so we can't tell what subsystem it last queued on. Tracking this helps prevent an issue where trying to dequeue something from SSobjs would instead dequeue it from SStemperature or vice versa. Now, a processing subsystem will only set a dequeued datum's `is_processing` to null if it's equal to their `_internal_name`.

Fixes lighters that run out of fuel spamming their extinguish message because their `is_processing` was set to null by temperature dequeueing.

## Authorship
Me.

### DIFF
--- a/code/__defines/MC.dm
+++ b/code/__defines/MC.dm
@@ -27,7 +27,7 @@ if (Datum.is_processing) {\
 		PRINT_STACK_TRACE("Failed to start processing. [log_info_line(Datum)] is already being processed by [Datum.is_processing] but queue attempt occured on [#Processor]."); \
 	}\
 } else {\
-	Datum.is_processing = #Processor;\
+	Datum.is_processing = Processor._internal_name;\
 	Processor.processing += Datum;\
 }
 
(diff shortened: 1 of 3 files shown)


code/__defines/MC.dm
```diff
@@ -137,4 +140,7 @@ if(Datum.is_processing) {\
 		processing = SS##X.processing; \
 	}\
 }\
-/datum/controller/subsystem/processing/##X
+/datum/controller/subsystem/processing/##X{\
+	_internal_name = "SS" + #X;\
+}\
+/datum/controller/subsystem/processing/##X
```
MistakeNot4892, line 144: I don't think this is needed with the above.

code/controllers/subsystems/processing/processing.dm
```diff
@@ -28,7 +28,7 @@ SUBSYSTEM_DEF(processing)
 		var/datum/thing = current_run[current_run.len]
 		current_run.len--
 		if(QDELETED(thing) || (call(thing, process_proc)(wait, times_fired, src) == PROCESS_KILL))
-			if(thing)
+			if(thing?.is_processing == _internal_name)
```
MistakeNot4892, line 31: Is `is_processing` ever actually set to `_internal_name`?

code/__defines/MC.dm
out-of-phaze: It is. I couldn't get it working otherwise.

MistakeNot4892 (approved):

PR #3601, opened 2024-01-19
Fancy food box rework.

MistakeNot4892:
## Description of changes
- Fancy boxes are now `/box/fancy`.
- Donut boxes are now `/fancy`.
- Egg boxes, cracker bags, and donut boxes now take overlays from their contents.
- Eggs, crackers and donuts now use the single icon system.
- Pizza boxes have been totally rewritten. 
- Pizzas use the single icon system.
- Pizzas boxes have inhands.
- Throwing a pizza may ruin it (visual only).
- You can stack pizza boxes higher than 5 but it's probably a terrible idea.

Video of pizza stuff:
https://cdn.discordapp.com/attachments/678820025355862048/1197845870255816734/dreamseeker_kQl6Ped6wK.mp4?ex=65bcbfe6&is=65aa4ae6&hm=196a370f964ee564ff96b1f51042d05072b89acf6e23718452384c3e1f0a0721&

Working towards breaking up `icons/obj/food.dmi`, this is a starting point.
Pizza stuff was a fey mood inspired by Evree posting ruined pizza photos.

## Authorship
Myself.

## Changelog
:cl:
tweak: Egg and donut boxes are slightly fancier.
tweak: Pizza stacks have been enhanced to dangerous levels.
/:cl:


out-of-phaze (dismissed):
I think the number of repaths in this PR would justify adding a migration in `tools/map_migrations/`.

Other than that, it looks fine, though I'd like if there were a more generalized system for the donut box/cracker sleeve behavior where we use a `_crackerbag` or `_donutbox` state suffix on the items inside the storage. It seems that the main thing to change is the suffix and the pixel offset, though it might also have a slight performance hit to manually check `istype(foo, key_item)`... though we don't do any typechecks for donuts/crackers currently, which I suppose is fine because we check the icon for a matching state.

MistakeNot4892:
Yeah I avoided the typecheck so that in theory other stuff added down the track can have its own overlay state without needing any additional handling. I could generalize it down to /fancy possibly.

out-of-phaze (approved):

code/game/objects/items/weapons/storage/fancy/_fancy.dm
```diff
@@ -0,0 +1,48 @@
+/*
+ * The 'fancy' path is for objects like candle boxes that show how many items are in the storage item on the sprite itself
+ */
+
+/obj/item/storage/box/fancy
+	abstract_type = /obj/item/storage/box/fancy
+	/// A string modifier used to generate overlays for contents.
+	var/use_single_icon_overlay_state
+	/// The root type  of the key item that this "fancy" container is meant to store.
```
out-of-phaze:
```suggestion
	/// The root type of the key item that this "fancy" container is meant to store.
```

code/game/objects/items/weapons/storage/fancy/_fancy.dm
```diff
@@ -0,0 +1,48 @@
...
+	var/obj/item/key_type
+
+/obj/item/storage/box/fancy/proc/offset_contents_overlay(var/overlay_index, var/image/overlay)
```
out-of-phaze: It might be good to rename this to something like `modify_contents_overlay` in the event that something other than offsetting is done in this by a subtype.

out-of-phaze (dismissed):

code/game/objects/items/weapons/storage/fancy/_fancy.dm
MistakeNot4892: Went with `adjust` to line up with the overlay adjustment proc on mobs.

out-of-phaze (approved):

PR #3608, opened 2024-01-20
Fixes runtimes from clientless mobs attempting to maneuver.

MistakeNot4892:


MistakeNot4892:
Speedmerge due to CI fails resulting.

PR #3617, opened 2024-01-23
Fixes maploader landmark.

MistakeNot4892:
Seems like the actual loading loop got clobbered at some point. Tested on Crux.


out-of-phaze:
https://github.com/out-of-phaze/LighthouseRelay/commit/2deefa17a0110a82d0f3f7f28c9aafd9f1cfe263

no time to review atm but this is the fix downstream

code/modules/maps/helper_landmarks.dm
```diff
@@ -4,8 +4,9 @@
 	var/centered = TRUE
 	var/list/map_template_names	//list of template names to pick from
 
-/obj/abstract/landmark/map_load_mark/New(loc)
-	..()
+INITIALIZE_IMMEDIATE(/obj/abstract/landmark/map_load_mark)
+/obj/abstract/landmark/map_load_mark/Initialize()
```
out-of-phaze, line 8: according to past-me in https://github.com/NebulaSS13/Nebula/pull/3330 this isn't feasible because this should never run during Initialize() in SSatoms? it should only run at compiletime or in SSmapping, i could be wrong about that though, i can always just revert/issue a fix if it breaks downstream

MistakeNot4892:
More or less identical, I just included a New to INITIALIZE_IMMEDIATE and an extra template category to avoid spawning the sub-templates on their own z-levels for testing,

comma (approved):

out-of-phaze (approved):

code/modules/maps/helper_landmarks.dm
MistakeNot4892: I don't really get what past-you means since the only time this would run the actual load logic in Initialize() is if someone spawns it, which is already strongly discouraged. Could mark the type as unspawnable to really formalize that though.

PR #3621, opened 2024-01-24
Split disposal.dmi

quardbreak:
## Description of changes
Cleanups disposal.dmi and puts disposals objects sprites in separate files.
Removed bunch of unused and old icons:
- Semi-transparency disposal pipe icons (T-Ray should use alpha on pipes already)
- "trashbin" (duplicate of old machinery component icon)
- old toilet 😔

Updated some code to make animation icon less hardcoded. 
Renamed `con[icon_state]` to `[icon_state]_construct` for better explanation.

Open as draft PR. I need to test that nothing is missing.

Less icon hoarding, easier to maintain and one step closer to disposals refactor.

## Authorship
Myself.


code/modules/recycling/disposal.dm
```diff
@@ -310,21 +310,21 @@ var/global/list/diversion_junctions = list()
 
 	// flush handle
 	if(flush)
-		overlays += image('icons/obj/pipes/disposal.dmi', "dispover-handle")
+		overlays += image(icon, "[icon_state]-handle")
```
MistakeNot4892: Should just be `overlays += "[icon_state]-handle"` here and below, wherever it's just using `icon`.
Would be good to replace the direct overlays modification with cut_overlays()/add_overlay() if you're willing.

MistakeNot4892 (changes requested):

code/modules/recycling/disposal.dm
quardbreak: Done.

MistakeNot4892 (approved):

PR #3647, opened 2024-01-31
Fluid reagents are now turf reagents.

MistakeNot4892:
## Description of changes
- Moves reagent holding from the fluid overlay onto the turf.
- Moves various processing and updating from the fluid overlay onto the turf.
- `on_reagent_change()` is now called by `/datum/reagents/Destroy()`.
- Several icon updates and on_reagent_change calls have been updated to check for null reagents due to the atom info repo now causing calls to on_reagent_change().
- Fluid glowing has been removed - it never worked anyway.

Simplifies fluid logic/handling and makes future atom-level reagent consumption like fires easier.

## Authorship
Myself.

## Changelog
Nothing player-facing.

### DIFF
--- a/code/game/objects/effects/fluids.dm
+++ b/code/game/objects/effects/fluids.dm
@@ -1,90 +1,44 @@
-/obj/effect/fluid_overlay
-	name = ""
-	icon = 'icons/effects/liquids.dmi'
-	icon_state = "puddle"
-	anchored = TRUE
-	simulated = 0
-	opacity = FALSE
-	mouse_opacity = MOUSE_OPACITY_UNCLICKABLE
-	layer = FLY_LAYER
-	alpha = 0
-	color = COLOR_LIQUID_WATER
-	var/update_lighting = FALSE
+/atom/movable/fluid_overlay
+	name                = ""
+	icon                = 'icons/effects/liquids.dmi'
+	icon_state          = "puddle"
+	anchored            = TRUE
+	simulated           = FALSE
+	opacity             = FALSE
+	mouse_opacity       = MOUSE_OPACITY_UNCLICKABLE
+	layer               = FLY_LAYER
+	alpha               = 0
+	color               = COLOR_LIQUID_WATER
+	is_spawnable_type   = FALSE
 
-/obj/effect/fluid_overlay/Initialize()
-	atom_flags |= ATOM_FLAG_OPEN_CONTAINER
-	icon_state = ""
-	. = ..()
-	var/turf/simulated/T = get_turf(src)
-	if(!isturf(T) || !T.CanFluidPass())
-		return INITIALIZE_HINT_QDEL
-	if(istype(T))
-		T.unwet_floor(FALSE)
-
-/obj/effect/fluid_overlay/airlock_crush()
-	qdel(src)
-
-/obj/effect/fluid_overlay/Move()
-	PRINT_STACK_TRACE("A fluid overlay had Move() called!")
-	return FALSE
-
-/obj/effect/fluid_overlay/Destroy()
-	var/turf/old_loc = get_turf(src)
-	for(var/checkdir in global.cardinal)
-		var/turf/neighbor = get_step(old_loc, checkdir)
-		if(neighbor)
-			ADD_ACTIVE_FLUID(neighbor)
-	REMOVE_ACTIVE_FLUID(loc)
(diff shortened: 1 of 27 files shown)


code/game/objects/effects/fluids.dm
```diff
@@ -9,82 +9,65 @@
...
-	for(var/checkdir in global.cardinal)
-		var/obj/effect/fluid/F = locate() in get_step(loc, checkdir)
-		if(F)
-			ADD_ACTIVE_FLUID(F)
-	update_lighting = TRUE
-	update_icon()
-
-/obj/effect/fluid/Destroy()
-	var/turf/simulated/T = get_turf(src)
+/obj/effect/fluid_overlay/Destroy()
```
out-of-phaze: I don't understand why the turf update logic is in the Destroy() override for the visual-only overlay...

code/modules/mob/living/living.dm
```diff
@@ -797,22 +797,21 @@ default behaviour is:
 
 /mob/living/handle_drowning()
 	var/turf/T = get_turf(src)
-	if(!can_drown() || !loc.is_flooded(lying))
+	if(!T || !can_drown() || !T.is_flooded(lying))
```
out-of-phaze: ~~you could be on the bottom of the ocean and not drown if you're in a locker with this change.~~ `loc.is_flooded(lying)` was correct I think

actually i think this has the opposite issue. atoms can't be not-flooded if their turf is flooded, which might cause issues for airtight lockers, exosuits, etc.

code/modules/mob/living/living.dm
```diff
@@ -797,22 +797,21 @@ default behaviour is:
...
 		return FALSE
 	if(!lying && T.above && T.above.is_open() && !T.above.is_flooded() && can_overcome_gravity())
 		return FALSE
 	if(prob(5))
 		var/datum/reagents/metabolism/inhaled = get_inhaled_reagents()
 		var/datum/reagents/metabolism/ingested = get_ingested_reagents()
-		var/obj/effect/fluid/F = locate() in loc
-		to_chat(src, SPAN_DANGER("You choke and splutter as you inhale [(F?.reagents && F.reagents.get_primary_reagent_name()) || "liquid"]!"))
+		to_chat(src, SPAN_DANGER("You choke and splutter as you inhale [T.reagents?.get_primary_reagent_name() || "liquid"]!"))
```
out-of-phaze: maybe use `get_primary_reagent()?.liquid_name` here, idk

mods/content/xenobiology/slime/life.dm
```diff
@@ -65,12 +65,10 @@
 
 /mob/living/slime/fluid_act(datum/reagents/fluids)
 	. = ..()
-	if(stat == DEAD)
-		var/obj/effect/fluid/F = locate() in loc
-		if(F && F.reagents?.total_volume >= FLUID_SHALLOW)
-			F.reagents.add_reagent(/decl/material/liquid/slimejelly, (is_adult ? rand(30, 40) : rand(10, 30)))
-			visible_message(SPAN_DANGER("\The [src] melts away...")) // Slimes are water soluble.
-			qdel(src)
+	if(stat == DEAD && loc?.reagents?.total_volume >= FLUID_SHALLOW)
```
out-of-phaze: may be worth changing this to check the volume of water in it specifically, or use some other value. otherwise this will cause a cascade of slime jelly puddles dissolving other slimes

out-of-phaze (dismissed):

mods/content/xenobiology/slime/life.dm
MistakeNot4892: Good thought, this logic predates non-water fluids.

code/modules/mob/living/living.dm
MistakeNot4892: Yeah no idea why I made this change.

code/modules/mob/living/living.dm
MistakeNot4892: There's actually a helper I should be using here.

MistakeNot4892:
Gutted the fluid overlay logic down to the bare visual essentials and made some other changes to enable the turf to handle everything.

MistakeNot4892:
Thinking I should consider adding fluid displacement to closing airlocks while I'm making these changes. Not sure how that should work though.

out-of-phaze (approved):

PR #3654, opened 2024-02-02
Prevents inflatables from exploding due to being next to a window containing high pressure.

MistakeNot4892:
Fixes https://github.com/ScavStation/ScavStation/issues/862

@Lohikar wrote all the macros, I just plugged them in.

### DIFF
--- a/code/_helpers/game.dm
+++ b/code/_helpers/game.dm
@@ -400,20 +400,31 @@
 * Gets the highest and lowest pressures from the tiles in cardinal directions
 * around us, then checks the difference.
 */
-/proc/get_surrounding_pressure_differential(var/turf/loc)
+/proc/get_surrounding_pressure_differential(var/turf/loc, atom/originator)
 	var/minp = INFINITY
 	var/maxp = 0
-	var/has_neighbor = FALSE
+	var/has_neighbour = FALSE
+	var/airblock // zeroed by ATMOS_CANPASS_TURF, declared early as microoptvar/airblock
 	for(var/dir in global.cardinal)
-		var/turf/neighbor = get_step(loc,dir)
-		if(!neighbor || (neighbor.c_airblock(loc) & AIR_BLOCKED))
+		var/turf/neighbour = get_step(loc,dir)
+		if(!neighbour)
 			continue
-		var/datum/gas_mixture/environment = neighbor.return_air()
+		for(var/obj/O in loc)
+			if(originator && O == originator)
+				continue
+			ATMOS_CANPASS_MOVABLE(airblock, O, neighbour)
+			. |= airblock
+		if(airblock & AIR_BLOCKED)
+			continue
+		ATMOS_CANPASS_TURF(airblock, neighbour, loc)
+		if(airblock & AIR_BLOCKED)
+			continue
+		var/datum/gas_mixture/environment = neighbour.return_air()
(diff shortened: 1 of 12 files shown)


comma (dismissed):

out-of-phaze:
I'm not sold on this as a solution. I think the proper solution is to make inflatables use `getOPressureDifferential` and have `getOPressureDifferential` check `neighbour.c_airblock(src.loc) & AIR_BLOCKED` just as firedoors do in `/obj/machinery/door/firedoor/proc/can_safely_open()`.

code/_helpers/game.dm
```diff
@@ -400,15 +400,17 @@
 * Gets the highest and lowest pressures from the tiles in cardinal directions
 * around us, then checks the difference.
 */
-/proc/getOPressureDifferential(var/turf/loc)
-	var/minp=16777216;
-	var/maxp=0;
+/proc/get_surrounding_pressure_differential(var/turf/loc)
+	var/minp = 16777216 // Arbitrary high value
```
out-of-phaze: Not `INFINITY`? Is there a reason this wasn't updated?

code/_helpers/game.dm
```diff
@@ -400,15 +400,17 @@
...
+	var/maxp = 0
 	for(var/dir in global.cardinal)
-		var/turf/T = get_step(loc,dir)
-		var/datum/gas_mixture/environment = T.return_air()
-		var/cp = environment?.return_pressure()
-		if(cp<minp)minp=cp
-		if(cp>maxp)maxp=cp
+		var/turf/neighbor = get_step(loc,dir)
+		if(!neighbor || (neighbor.c_airblock(loc) & AIR_BLOCKED))
```
out-of-phaze: I may be stupid. This needs to be done in a way that's at least a little smarter than I suggested, because firedoors block air, no? Or did you test this and it still worked?

out-of-phaze (dismissed):

code/_helpers/game.dm
MistakeNot4892: I don't know why it's such a hyperspecific value so I left it alone. 

As an aside, it also seems weird that calling this on a turf with no neighbors it can exchange atmos with would have infi ite/maximum pressure differential rather than just zero.

code/_helpers/game.dm
MistakeNot4892: I did not test this, no.

MistakeNot4892:
Ended up pulling in a bunch of atmos fixes. Tested inflatables with border windows and doors, all working.

code/_helpers/game.dm
```diff
@@ -400,16 +400,31 @@
...
 * around us, then checks the difference.
 */
-/proc/getOPressureDifferential(var/turf/loc)
-	var/minp=16777216;
-	var/maxp=0;
+/proc/get_surrounding_pressure_differential(var/turf/loc, atom/originator)
+	var/minp = INFINITY
+	var/maxp = 0
+	var/has_neighbour = FALSE
+	var/airblock // zeroed by ATMOS_CANPASS_TURF, declared early as microoptvar/airblock
```
out-of-phaze:
```suggestion
	var/airblock // zeroed by ATMOS_CANPASS_TURF, declared early as microopt
```

out-of-phaze (changes requested):

out-of-phaze (approved):

PR #3655, opened 2024-02-03
Moves /mob and /obj health vars to /atom.

MistakeNot4892:
## Description of changes
- `/mob` now uses `max_health` instead of `mob_default_max_health`.
- `max_health` and `get_max_health()` are now on `/atom`.
- `/mob` and `/obj` both use `get_max_health()` in most cases.
- `/mob` `current_health` var is now defined at `/atom` and `/obj` uses it instead of `health`.

Unifies some duplicate code, prepares for unifying damage procs.
Allows integrating changes in matter into modifying max health easier.

## Authorship
Myself.

## Changelog
Nothing player-facing.


PsyCommando:
So, I was wondering, why change health to current_health? It's a bit redundant and longer to type out?

MistakeNot4892:
> So, I was wondering, why change health to current_health? It's a bit redundant and longer to type out?

Visibility and searchability - there's a shit-ton of `health` in the codebase: 
image

Making it use something distinct and specific helps with future work.

out-of-phaze (approved):

PR #3659, opened 2024-02-03
Fixes cartridge message being 'you eat the cartridge'

MistakeNot4892:


code/modules/reagents/dispenser/cartridge.dm
```diff
@@ -15,6 +15,12 @@
 	if(populate && reagents.primary_reagent)
 		setLabel(reagents.get_primary_reagent_name())
 
+/obj/item/chems/chem_disp_cartridge/self_feed_message(var/mob/user)
+	to_chat(user, "<span class='notice'>You swallow a gulp from \the [src].</span>")
```
out-of-phaze: Would like to nudge here about `SPAN_NOTICE("foo")` being preferred over `"<span class='notice'>foo</span>"` in new code, but that's non-blocking.

out-of-phaze (dismissed):

out-of-phaze (approved):
I only meant the one line, but I'm not complaining.

PR #3661, opened 2024-02-04
Minor lock refactor/fix.

MistakeNot4892:
- Fixes no feedback on unlocking doors.
- Fixes crafted keys having material written to key data.

### DIFF
--- a/code/modules/reagents/reactions/reaction_other.dm
+++ b/code/modules/reagents/reactions/reaction_other.dm
@@ -13,8 +13,7 @@
 /decl/chemical_reaction/soap_key/on_reaction(var/datum/reagents/holder)
 	var/obj/item/soap/S = holder.get_reaction_loc(chemical_reaction_flags)
 	if(istype(S) && S.key_data)
-		var/obj/item/key/soap/key = new(get_turf(S), S.key_data)
-		key.uses = strength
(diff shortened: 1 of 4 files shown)


code/modules/reagents/reactions/reaction_other.dm
```diff
@@ -13,8 +13,7 @@
 /decl/chemical_reaction/soap_key/on_reaction(var/datum/reagents/holder)
 	var/obj/item/soap/S = holder.get_reaction_loc(chemical_reaction_flags)
 	if(istype(S) && S.key_data)
-		var/obj/item/key/soap/key = new(get_turf(S), S.key_data)
-		key.uses = strength
+		new /obj/item/key/temporary(get_turf(S), /decl/material/liquid/cleaner, S.key_data, strength)
```
out-of-phaze, line 16:
```suggestion
		new /obj/item/key/temporary(get_turf(S), /decl/material/liquid/cleaner/soap, S.key_data, strength)
```
Make sure it doesn't melt at room temperature.

out-of-phaze (changes requested):

code/modules/reagents/reactions/reaction_other.dm
out-of-phaze: Whoops, that's dev-only right now.

out-of-phaze (approved):

PR #3666, opened 2024-02-10
Atom fire system.

MistakeNot4892:
## Description of changes
Copying this from #3435 due to a branch rename.
Adds an atom-level fire system with the goal of replacing ZAS fires for most cases.

In summary:
- Atoms have a proc, `get_flammable_material`, which returns all constituent material that has a fuel value and ignites at or below the given temperature. This uses `reagents` by default and adds matter when it's an `/obj`.
- A new subsystem, `SSfires`, handles burning atoms as well as cross-turf spreading.

## TODO
- [ ] Get temperature from fires working to an acceptable level.
- [ ] Move on-turf spreading onto SSfires as well as cross-turf.
- [ ] Convert mob fires to atom fires.
- [ ] Convert fire sources to atom fires.

:fire:

## Authorship
Myself.

## Changelog
:cl:
add: Atoms can now burn. Report bugs with fires on the issue tracker!
/:cl:


MistakeNot4892:
Will revisit this when melting is stress tested and sorted out, if only because it will mean we've got better matter handling.

PR #3669, opened 2024-02-10
Added the ability for materials to transition into other materials based on atom temperature.

MistakeNot4892:
## Description of changes
- `ProcessAtomTemperature()` can losslessly/non-destructively replace material and matter type (sand to glass, clay to pottery).
- Added a pottery type for simple baked clay instead of using ceramic.

Notes towards a more full featured pottery system involving a kiln.

## Authorship
Myself.

## Changelog
Functionally nothing player-facing unless someone has a habit of setting lumps of clay on fire.

### DIFF
--- a/code/modules/materials/definitions/solids/materials_solid_mineral.dm
+++ b/code/modules/materials/definitions/solids/materials_solid_mineral.dm
@@ -219,7 +219,7 @@
 	uid = "solid_sand"
 	color = "#e2dbb5"
 	heating_products = list(/decl/material/solid/glass = 1)
-	heating_point = GENERIC_SMELTING_HEAT_POINT
+	heating_point = 2000 CELSIUS
 	heating_sound = null
 	heating_message = null
 	ore_compresses_to = /decl/material/solid/stone/sandstone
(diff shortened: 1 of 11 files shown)


code/modules/materials/definitions/solids/materials_solid_mineral.dm
```diff
@@ -219,7 +219,7 @@
 	uid = "solid_sand"
 	color = "#e2dbb5"
 	heating_products = list(/decl/material/solid/glass = 1)
-	heating_point = GENERIC_SMELTING_HEAT_POINT
+	heating_point = 2273
```
out-of-phaze:
```suggestion
	heating_point = 2000 CELSIUS
```
this is preferred stylistically and also makes it clear why it's 2273 and not 2000

code/modules/materials/definitions/solids/materials_solid_mineral.dm
```diff
@@ -236,18 +236,17 @@
...
-	heating_sound = null
-	heating_message = null
 	ore_compresses_to = null
-	ore_icon_overlay = "dust"
+	ore_icon_overlay = "lump"
 	value = 0.8
 	default_solid_form = /obj/item/stack/material/lump
+	bakes_into_material = /decl/material/solid/stone/pottery
+	melting_point = null // Clay is already almost a liquid...
+	bakes_into_at_temperature = 1330 // roughly the temperature expected from a kiln
```
out-of-phaze: non-blocking but i would personally round this to either 1050, 1060, or 1100C to be honest, 1056.85C is a bit odd

out-of-phaze (dismissed):
as discussed on Discord, this should probably *eventually* be made a gradual conversion rather than a binary one, but this works fine for now.

also i'm a little uncertain about the pottery/ceramic distinction but it's mostly a functional one so it's whatever

MistakeNot4892:
> also i'm a little uncertain about the pottery/ceramic distinction but it's mostly a functional one so it's whatever

Mostly colour, ceramic is off-white, baked clay is brick red. Baked clay should also be more brittle and less heat-resistant than proper ceramics.

code/modules/materials/definitions/solids/materials_solid_mineral.dm
out-of-phaze: @MistakeNot4892 my point here was to use the CELSIUS helper, not to change it to 1300K or 1500K x_x

if you're fine with it just being some weird non-round number when converted to celsius that's fine though, i can approve as-is

out-of-phaze (dismissed):

out-of-phaze (approved):

PR #3687, opened 2024-02-15
Limits scan title to avoid header covering up the print button.

MistakeNot4892:
Looks like some hydroponics scans can have a combined area and plant name long enough to cause the fakenano header bar to overflow and cover up the print button.

Fixes https://github.com/ScavStation/ScavStation/issues/867

### DIFF
--- a/code/game/objects/items/devices/scanners/_scanner.dm
+++ b/code/game/objects/items/devices/scanners/_scanner.dm
@@ -24,8 +24,8 @@
 	show_menu(user)
 
 /obj/item/scanner/proc/show_menu(mob/user)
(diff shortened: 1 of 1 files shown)


out-of-phaze (changes requested):
This is not a suitable fix because the issue is with the actual CSS. Also, it'll likely still occur for things like vending machine wires that use `holder.name` for their title. I'd suggest #3689 to fix it instead, but if you're still set on this scan title change for other reasons, we can revisit it after—maybe removing all variable-name popup titles for consistency's sake and moving the names into the contents.

PR #3681, opened 2024-02-12
Adding plating for food.

MistakeNot4892:
## Description of changes
This is a bit of a fey mood and is pretty anemic/rushed, willing to take input on implementation prior to merge.

- Removes plates, trays and platters from food icons in food.dmi.
- Adds plates/platters/trays that food can be placed on.
- Removes plate and tray trash items.
- Eating food off a plate dirties the plate, which can then be washed to clean it.

## Why and what will this PR improve
Trash plates/trays has always been a bit strange. This adds a slightly more interesting element to food.

## Authorship
Myself.

## Changelog
:cl:
tweak: Plates are no longer part of food when it emerges from the microwave, and must be retrieved from the dinnerware vendor and used to plate food.
/:cl:

PR #3688, opened 2024-02-15
Reworks utensils.

MistakeNot4892:
## Description of changes
- Repaths utensils, moves table knives to utensil type.
- Adds chopsticks.
- Reworks how utensils handle food interactions in the backend to benefit from the new edibility system.
- Changes the Chinese food vendor's wares to be slightly less racist.

Depends on #3681.

Better/cleaner utensil code with some skeleton code towards future features.

## Authorship
Myself.

## Changelog
:cl:
tweak: The Chinese food vendor has slightly different wares.
/:cl:


tools/map_migrations/3681_plates.txt
out-of-phaze, line 1: the new migrations in this file should probably be in `3688_utensils.txt` or something similar

code/modules/food/utensils/_utensil.dm
```diff
@@ -0,0 +1,120 @@
...
+	return loaded_food ? loaded_food.attack_self(user) : ..()
+
+/obj/item/utensil/handle_eaten_by_mob(var/mob/user, var/mob/target)
+	. = loaded_food ? loaded_food.handle_eaten_by_mob(user, target) : ..()
+	if(QDELETED(loaded_food))
+		loaded_food = null
+		update_icon()
+
+/obj/item/utensil/get_edible_material_amount(var/mob/eater)
+	return loaded_food ? loaded_food.reagents?.total_volume : ..()
```
out-of-phaze:
```suggestion
	return loaded_food ? loaded_food.get_edible_material_amount() : ..()
```

code/__defines/misc.dm
```diff
@@ -315,4 +315,10 @@
...
 // Defines for fluorescence (/atom/var/fluorescent)
 #define FLUORESCENT_GLOWS   1	// Glows when under flourescent light
-#define FLUORESCENT_GLOWING 2	// Currently glowing due to flourescent light
\ No newline at end of file
+#define FLUORESCENT_GLOWING 2	// Currently glowing due to flourescent light
+
+// Flags used for utensil-food interaction.
+#define UTENSIL_FLAG_COLLECT  BITFLAG(0)
+#define UTENSIL_FLAG_SCOOP  BITFLAG(1)
+#define UTENSIL_FLAG_SLICE  BITFLAG(2)
```
out-of-phaze: is `UTENSIL_FLAG_SLICE` really necessary when we use `has_edge()` for slicing foods? i'm worried that this is going to mean, for example, you can't slice foods with a survival knife or something like that

code/__defines/misc.dm
```diff
@@ -315,4 +315,10 @@
 
 // Defines for fluorescence (/atom/var/fluorescent)
 #define FLUORESCENT_GLOWS   1	// Glows when under flourescent light
-#define FLUORESCENT_GLOWING 2	// Currently glowing due to flourescent light
\ No newline at end of file
+#define FLUORESCENT_GLOWING 2	// Currently glowing due to flourescent light
+
+// Flags used for utensil-food interaction.
+#define UTENSIL_FLAG_COLLECT  BITFLAG(0)
```
out-of-phaze: is there an explanation anywhere of the intended difference between "collecting" and "scooping"?

out-of-phaze (dismissed):

code/__defines/misc.dm
MistakeNot4892: The intent is to not be able to use a fork to eat soup, basically. I can document my plans if you want?

code/__defines/misc.dm
MistakeNot4892: I suppose I should unify that with food slicability down the track. I just intended this for like, slicing part of a steak or a cake before using a fork.

code/__defines/misc.dm
out-of-phaze: That'd be nice. The distinction makes sense, I just wasn't sure which was which. I guess collecting is soup and scooping is mashed potatoes, ice cream, etc.?

MistakeNot4892: Collecting was original spearing but someone pointed out it was kinda odd to be spearing stuff on chopsticks. Scooping is spoons.

MistakeNot4892: I've added some comments to the defines.

code/__defines/misc.dm
MistakeNot4892: Added edge/sharp to the utensil logic.

MistakeNot4892: Sliceable flag also now handles the existing slicing mechanic.

code/__defines/misc.dm
```diff
@@ -314,5 +314,17 @@
...
+#define FLUORESCENT_GLOWS   1
+/// Currently glowing due to flourescent light
+#define FLUORESCENT_GLOWING 2
+
+// Flags used for utensil-food interaction.
+/// Solid or semi-solid food; chopsticks, forks.
+#define UTENSIL_FLAG_COLLECT BITFLAG(0)
+/// Soft, liquid or semi-liquid food; soups, stews, pudding.
+#define UTENSIL_FLAG_SLICE   BITFLAG(2)
+/// Foods that need to be sliced before eating; steak, grapefruit.
```
out-of-phaze: you seem to have swapped the comments here

out-of-phaze (dismissed):

code/__defines/misc.dm
MistakeNot4892: Ah I've messed up the define order when I wqs putting the comments on a new line.

out-of-phaze (approved):

PR #3690, opened 2024-02-15
Updating the emote system

MistakeNot4892:
## Description of changes
- Ports backend changes from Polaris that convert default emote lists to static lists, and remove the per-mob caching in favour of checking at time of invocation.
- Replaces a few emote() with custom_emote().
- Allows emote() to accept a decl type.
- Standardizes emote USER and TARGET tokens to `$USER$`, `$TARGET$`, etc.
- Validates emote strings for no erroneous tokens.
- Enforces emote key uniqueness.

Emotes are kinda crufty.

## Authorship
Myself.

## Changelog
Nothing player-facing.


code/modules/organs/pain.dm
```diff
@@ -47,7 +47,7 @@
 	if(.)
 		var/force_emote = species.get_pain_emote(src, power)
 		if(force_emote && prob(power))
-			var/decl/emote/use_emote = usable_emotes[force_emote]
+			var/decl/emote/use_emote = GET_DECL(force_emote)
```
out-of-phaze, line 50: `get_pain_emote` still returns the key rather than the decl typepath. This will runtime, I think.

out-of-phaze (changes requested):

out-of-phaze (dismissed):

out-of-phaze (approved):

PR #3691, opened 2024-02-15
Non-secret modes should be available again.

MistakeNot4892:


code/controllers/configuration.dm
```diff
@@ -288,14 +288,14 @@
 	var/list/all_modes = decls_repository.get_decls_of_subtype(/decl/game_mode)
 	for (var/mode_type in all_modes)
 		var/decl/game_mode/game_mode = all_modes[mode_type]
-		if (!game_mode.uid || !(game_mode.uid in modes))
```
out-of-phaze, line 291: I feel like maybe it was supposed to check `(game_mode.uid in modes)` rather than `!(game_mode.uid in modes)`, but since UIDs are enforced to be, well, unique, it shouldn't be necessary.

out-of-phaze (approved):

PR #3693, opened 2024-02-16
[MDB IGNORE] [IDB IGNORE] Standardizing death, gibbing and dusting handling.

MistakeNot4892:
## Description of changes
- `gib()` and `dust()` now take fewer args. dust() takes none, gib() takes a bool for whether or not it should create a gibspawner.
- Dust/gib animation state and remains type are handled by procs.
- Dust/gib call a shared handle_existence_failure() proc for most of their logic.
- Dust/gib have been reworked across the board to take the return type into account before creating debris etc.
- `death()` now only takes one argument, `gibbed`. Calls have been updated for this.
- Mobs now use `get_death_message()` and `get_self_death_message()` to return the relevant strings.
- Simple animals now have a `death_message` member var for this.
- Base death() cancels BYOND walk() movement.

Cleaner death code with fewer parameters to keep track of.

## Authorship
Myself.

## Changelog
Nothing player-facing.


code/modules/mechs/mech_life.dm
```diff
@@ -89,17 +89,20 @@
...
-/mob/living/exosuit/death(var/gibbed)
+/mob/living/exosuit/death(gibbed)
 	// Eject the pilot.
 	if(LAZYLEN(pilots))
 		hatch_locked = 0 // So they can get out.
 		for(var/pilot in pilots)
 			eject(pilot, silent=1)
 	new wreckage_path(get_turf(src), src, gibbed)
-	..(gibbed, (gibbed ? "explodes!" : "grinds to a halt before collapsing!"))
+	return ..(gibbed, (gibbed ? "explodes!" : "grinds to a halt before collapsing!"))
```
comma: still passing death message here?

MistakeNot4892: Aiee it dodged my regex...

MistakeNot4892:
This should also fix the stack overflows in #3694

MistakeNot4892:
I have buggered my branches so I'm calling this dependent on #3695 so I can resolve the inevitable conflicts after that goes in.

out-of-phaze (approved):
`handle_existence_failure` confused me a little at first (and kind of still does), but nothing about it looks *wrong—* the unconditional `death(gibbed = TRUE)` regardless of the value of `dusted` confused me, but I think it just skips the death message and does some extra handling on organs for 'existence failure' type deaths? Might be worth renaming the argument in the future.

MistakeNot4892:
`handle_existence_failure` is an oblique Hitchhiker's Guide joke, sorry, I wasn't sure what else to call the general discorporatin-handling proc.

out-of-phaze:
> `handle_existence_failure` is an oblique Hitchhiker's Guide joke, sorry, I wasn't sure what else to call the general discorporatin-handling proc.

oh, not the name, I meant the logic. the name is fine

PR #3695, opened 2024-02-16
Condensing blood code.

MistakeNot4892:
## Description of changes
- `blood.dm` has been split up between mob definitions.
- Some blood logic has been moved down to `/mob/living`.
- `gibs()` will now retrieve all relevant info from the mob.
- Bleeding on the floor uses `blood_splatter()` directly.
- Bullet blood impact effects have been condensed on `/mob/living`.

More consistent behavior, tidier code.

## Authorship
Myself.

## Changelog
Nothing player-facing.


code/modules/mob/living/living_blood.dm
```diff
@@ -0,0 +1,34 @@
...
+	var/chems = LAZYACCESS(injected_data, "trace_chem")
+	for(var/C in chems)
+		src.reagents.add_reagent(C, (text2num(chems[C]) / my_species.blood_volume) * amount)//adds trace chemicals to owner's blood
+
+/mob/living/get_blood_data()
+
+	var/data = ..()
+	data["blood_DNA"]  = get_unique_enzymes()
+	data["blood_type"] = get_blood_type()
+	data["species"]    = get_species_name()
```
out-of-phaze: worried that for mobs without a species, this would result in `list("species" = null)` and similar. maybe `get_species_name() || "Unknown"`? or a new proc like `get_species_display_name()` which returns `"Unknown"` if species is null?

out-of-phaze (dismissed):
looks good aside from my concern about `get_species_name()` in `/mob/living/get_blood_data()`, will approve after that's addressed one way or the other

code/modules/mob/living/living_blood.dm
MistakeNot4892: Species name is used as a key in some cases so not sure about that. Might be better not to have species in data in those cases?

out-of-phaze (approved):

PR #3696, opened 2024-02-16
Adding logs and trees.

MistakeNot4892:
## Description of changes
- Adds some more trees, and logs to cut from them. 
- Adds the ability to cut logs into planks.
- Ports tree shaking when attacked from Polaris.
 
Sprites aren't great but they're as good as I'm willing to get them.
image

Sources of wood, mapping props.

## Authorship
Log icons and tree shake are adapted from Polaris, all else is mine.

## Changelog
:cl:
tweak: Trees can be cut down into logs.
add: Logs can be cut up into planks.
/:cl:


code/modules/hydroponics/seed.dm
```diff
@@ -726,33 +726,35 @@
 				total_yield = max(1,total_yield)
 
 		. = list()
-		for(var/i = 0;i<total_yield;i++)
-			var/obj/item/product = new product_type(get_turf(user),name)
-			. += product
+		if(ispath(product_type, /obj/item/stack))
+			. += new product_type(get_turf(user), total_yield)
```
out-of-phaze: does this properly handle splitting stacks when `total_yield` is greater than the max stack size?

code/modules/materials/material_recipes.dm
```diff
@@ -17,10 +17,10 @@
 	if(holographic || phase_at_temperature() != MAT_PHASE_SOLID)
 		return list()
 
-	// By default we don't let anything be crafted with ore, as it's too raw.
+	// By default we don't let anything be crafted with ore or logs, as they are too raw.
 	// We make an exception for clay as it is being moulded by hand.
 	. = list()
-	if(ispath(stack_type, /obj/item/stack/material/ore))
+	if(ispath(stack_type, /obj/item/stack/material/ore) || ispath(stack_type, /obj/item/stack/material/log))
```
out-of-phaze, line 23: we already have brick walls, i'm unsure if adding log walls is out of scope but it'd make sense. admittedly it might not make sense without either cutting wedges into the logs so they'll stack without rolling off, or having some other kind of binding material like ropes or rags or mud or clay, and i don't *think* we have support for either of those when crafting

out-of-phaze (dismissed):
i have been waiting for this since i first started working on nebula. bless

code/modules/hydroponics/seed.dm
MistakeNot4892: Nope.

code/modules/materials/material_recipes.dm
MistakeNot4892: I'm calling log walls out of scope for this PR since we don't have sprites for them. No objection to adding them down the track though.

PlayerDeer:
Wow people have gotten a lot better at spriting trees!

code/modules/hydroponics/seed.dm
MistakeNot4892: Done.

out-of-phaze (approved):

PR #3707, opened 2024-02-19
Reworking damage overlays to try to avoid crashes.

MistakeNot4892:
## Description of changes
- Adds validation to damage overlay states.
- Moves damage overlay states to /mob/living.

Trying to get a better idea of the weird icon issues causing client crashes.

## Authorship
Myself.

## Changelog
Nothing player-facing.


MistakeNot4892:
This appears to have prevented client crashes from gibbing or losing a limb, which is an improvement over existing behavior.

comma (approved):

PR #3709, opened 2024-02-20
Moving vis_contents changes onto a subsystem.

MistakeNot4892:
## Description of changes
- Adds SSvis_contents_updates and appropriate procs.
- Changes existing vis_contents macros to be procs that use the subsystem.

vis_contents manipulation is expensive and current uses on /turf will benefit from being ordered during initialization.

## Authorship
Myself.

## Changelog
Nothing player-facing.


MistakeNot4892:
Per Discord it's unclear if this is a good idea or not. It increased Crux init by 5s (85s -> 90s).
EDIT: It's unclear if this was due to vis_contents or not, more profiling might be needed.

code/__defines/subsystems.dm
```diff
@@ -20,27 +20,28 @@
...
+#define SS_INIT_JOBS             11
+#define SS_INIT_CIRCUIT          10
+#define SS_INIT_GRAPH            9
+#define SS_INIT_OPEN_SPACE       8
+#define SS_INIT_ATOMS            7
+#define SS_INIT_PRE_CHAR_SETUP   6
+#define SS_INIT_CHAR_SETUP       5
+#define SS_INIT_MACHINES         4
+#define SS_INIT_ICON_UPDATE      3
+#define SS_INIT_VIS_CONTENTS     2
```
out-of-phaze: It should init after `SS_INIT_AIR` to avoid wasted updates before zones are built.

out-of-phaze (changes requested):
That optimization commit I linked used SSmisc_late because it runs after SSair, which lets us skip updates that will be invalidated by zone building (a lot of them, surprisingly). I'd like to see it tested running after SSair rather than before.

code/__defines/subsystems.dm
MistakeNot4892: Done.

out-of-phaze (approved):

PR #3710, opened 2024-02-20
Various flooding and fluids optimizations and fixes.

MistakeNot4892:
## Description of changes
- ChangeTurf now universally preserves flooded status. A helper landmark has been added to let people spawn flooded turfs ad hoc.
- Fluid updates and sleeping have been refined/optimized.
- Fluid and flood activation/sleeping have been optimized.
- Turfs now have a displace_all_reagents() proc that tries to move all reagent contents into neighbors. This is used when a turf discovers it fails CanFluidPass() during SSfluids processing.

Performance of the fluid system and flooding init.

## Authorship
Myself and @out-of-phaze.

## Changelog
Nothing player-facing.

### DIFF
--- a/code/game/turfs/turf_changing.dm
+++ b/code/game/turfs/turf_changing.dm
@@ -12,6 +12,7 @@
 	var/obj/structure/lattice/L = locate(/obj/structure/lattice, src)
 	if(L)
 		qdel(L)
+
 // Called after turf replaces old one
 /turf/proc/post_change()
 	levelupdate()
--- a/code/game/turfs/turf_changing.dm
+++ b/code/game/turfs/turf_changing.dm
@@ -100,7 +100,7 @@
 			qdel(old_fire)
 
 	if(old_flooded != W.flooded)
-		set_flooded(old_flooded && !W.density)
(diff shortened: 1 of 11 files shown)


code/game/turfs/turf_changing.dm
```diff
@@ -100,7 +100,7 @@
 			qdel(old_fire)
 
 	if(old_flooded != W.flooded)
-		set_flooded(old_flooded && !W.density)
+		set_flooded(old_flooded)
```
MistakeNot4892, line 104: Noting that this makes spawning flooded turfs not work. Possibly not ideal.

MistakeNot4892: Added a mappable landmark to solve this.

MistakeNot4892:
Tested fairly extensively on Crux, fluids take awhile to settle but flooding a 5x5 chunk of the main ring did go to 0 active fluids within a minute or two. Flooding the entire map also didn't cause any issues other than locking up all resources while it tried to resolve the flows.

out-of-phaze (approved):

PR #3714, opened 2024-02-21
Dismantles /turf/simulated.

### LINKED ISSUE #3711: Tradeship loses air to open turfs on init.
#### Description of issue
Air vanishes when you start round on Tradeship.

Atmos not venting into nowhere.

#### Steps to reproduce
Start a round on Tradeship.

I've seen it occur in the main compartment and the engine room.

Since the exterior turf merge.

Current dev.

#### Issue bingo

MistakeNot4892:
## Description of changes
- Moves floor wetness and floor dirtiness onto turf decals.
- Moves all `/turf/simulated` behavior onto `/turf`, gated by `simulated` (note `/turf/simulated/wall` and `/turf/simulated/floor` are unaffected).
- Moves turf holiness onto a turf flag.
- Replicates initial_gas and zone_membership_candidate setting on `/turf/open`, `/turf/simulated/floor` and `/turf/simulated/wall`.

Moves us a step closer to destroying `/turf/simulated` and should fix #3711.

## Authorship
Myself.

## Changelog
Nothing player-facing.


out-of-phaze (dismissed):
I like that we keep it as an abstract type for now. We could maybe repath it fully in r6, mostly to stop people from relying on `istype(foo, /turf/simulated)` checks for gating logic.

MistakeNot4892:
> I like that we keep it as an abstract type for now. We could maybe repath it fully in r6, mostly to stop people from relying on `istype(foo, /turf/simulated)` checks for gating logic.

Yeah I'm leaving it until I can collapse everything down to `/turf/wall` and `/turf/floor` in 20 years.

out-of-phaze (dismissed):

out-of-phaze (dismissed):

out-of-phaze (approved):

PR #3726, opened 2024-02-24
Port some init optimizations from Lighthouse

out-of-phaze:
## Description of changes
Ports a bunch of optimizations from Lighthouse, mostly to make init faster.

Note that the mutable appearances thing in copy_lighting is fine—the corruption/crash is a resolved issue in 515.1631 and it was specifically due to reusing mutable appearances. `Reusing mutable_appearance could sometimes result in certain appearances being deallocated early, leading to a crash.`

Makes init much faster—in my testing it went from 90 seconds on tether to 60 seconds, but your mileage may vary.

### DIFF
--- a/code/_helpers/type2type.dm
+++ b/code/_helpers/type2type.dm
@@ -1,17 +1,12 @@
 /*
  * Holds procs designed to change one type of value, into another.
  * Contains:
- *			hex2num & num2hex
  *			text2list & list2text
  *			file2list
  *			angle2dir
  *			angle2text
  *			worldtime2text
  */
-
-// Returns the hex value of a number given a value assumed to be a base-ten value
-/proc/num2hex(num, len=2)
-	return num2text(num, len, 16)
 
 /proc/text2numlist(text, delimiter="\n")
 	var/list/num_list = list()
--- a/code/_helpers/type2type.dm
+++ b/code/_helpers/type2type.dm
@@ -11,29 +11,7 @@
 
 // Returns the hex value of a number given a value assumed to be a base-ten value
 /proc/num2hex(num, len=2)
-	if(!isnum(num))
-		num = 0
-	num = round(abs(num))
-	. = ""
-	var/i=0
-	while(1)
-		if(len<=0)
-			if(!num)
-				break
-		else
-			if(i>=len)
-				break
-		var/remainder = num/16
-		num = round(remainder)
-		remainder = (remainder - num) * 16
-		switch(remainder)
-			if(9,8,7,6,5,4,3,2,1)
-				. = "[remainder]" + .
-			if(10,11,12,13,14,15)
-				. = ascii2text(remainder+87) + .
-			else
-				. = "0" + .
-		i++
+	return num2text(num, len, 16)
 
 /proc/text2numlist(text, delimiter="\n")
 	var/list/num_list = list()
--- a/code/_helpers/unsorted.dm
+++ b/code/_helpers/unsorted.dm
@@ -773,16 +773,10 @@ var/global/list/WALLITEMS = list(
 	return 0
 
 /proc/get_random_colour(var/simple = FALSE, var/lower = 0, var/upper = 255)
-	var/colour
 	if(simple)
-		colour = pick(list("FF0000","FF7F00","FFFF00","00FF00","0000FF","4B0082","8F00FF"))
+		return pick(list("#ff0000","#ff7f00","#ffff00","#00ff00","#0000ff","#4b0082","#8f00ff"))
 	else
-		for(var/i=1;i<=3;i++)
-			var/temp_col = "[num2hex(rand(lower,upper))]"
-			if(length(temp_col )<2)
-				temp_col = "0[temp_col]"
-			colour += temp_col
-	return "#[colour]"
+		return rgb(rand(lower, upper), rand(lower, upper), rand(lower, upper))
 
 // call to generate a stack trace and print to runtime logs
 /proc/get_stack_trace(msg, file, line)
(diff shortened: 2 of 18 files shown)


code/game/turfs/turf.dm
```diff
@@ -78,6 +80,8 @@
 	var/last_flow_strength = 0
 	var/last_flow_dir = 0
 	var/atom/movable/fluid_overlay/fluid_overlay
+	var/is_water_source = FALSE
```
MistakeNot4892: Obviated by the fluid optimization PR I think?

code/game/turfs/turf.dm
```diff
@@ -10,6 +10,8 @@
 	/// Will participate in external atmosphere simulation if the turf is outside and no zone is set.
 	var/external_atmosphere_participation = TRUE
 
+	var/queued_for_vis_contents_update = FALSE
```
MistakeNot4892: So should I be closing my SSvis_contents PR?

code/_helpers/type2type.dm
```diff
@@ -11,29 +11,7 @@
...
-		remainder = (remainder - num) * 16
-		switch(remainder)
-			if(9,8,7,6,5,4,3,2,1)
-				. = "[remainder]" + .
-			if(10,11,12,13,14,15)
-				. = ascii2text(remainder+87) + .
-			else
-				. = "0" + .
-		i++
+	return num2text(num, len, 16)
```
MistakeNot4892: Could make this a macro for less proc overhead?

code/modules/random_map/noise/noise.dm
```diff
@@ -127,32 +126,30 @@
...
 			total += map[TRANSLATE_COORD(3, y)]
-			next_map[TRANSLATE_COORD(2, y)] = round(total / 3)
+			map[TRANSLATE_COORD(2, y)] = round(total / 3)
 			for(var/x = 3 to limit_x-1) // should technically be 2 + blur_radius to limit_x - blur_radius
 				total -= map[TRANSLATE_COORD(x-2, y)] // x - blur_radius - 1
 				total += map[TRANSLATE_COORD(x+1, y)] // x + blur_radius
-				next_map[TRANSLATE_COORD(x, y)] = round(total / 3) // should technically be 2*blur_radius+1
+				map[TRANSLATE_COORD(x, y)] = round(total / 3) // should technically be 2*blur_radius+1
 		// now do the same in the x axis
-		map = next_map.Copy()
```
MistakeNot4892: The purpose of using a second map is so that smoothing is acting on an unaltered original state, this is going to make the smoothing top-left biased as far as I follow it.

out-of-phaze: i believe this is not actually the case because of the specific box blur we use, it adds the unblurred... oh you're right, damn. ok i'll drop the commit, it was negligible fractions of a second anyway

code/game/machinery/alarm.dm
```diff
@@ -859,22 +859,22 @@ FIRE ALARM
 	return overlays_cache[key]
 
 /obj/machinery/firealarm/on_update_icon()
-	overlays.Cut()
+	cut_overlays()
 	icon_state = "casing"
 	if(construct_state && !istype(construct_state, /decl/machine_construction/wall_frame/panel_closed))
-		overlays += get_cached_overlay(construct_state.type)
+		add_overlay(get_cached_overlay(construct_state.type))
```
Lohikar: As noted on Discord, `get_cached_overlay()` here seems redundant — just use the native istate cache in SSoverlays.

code/_helpers/type2type.dm
Lohikar: Given how uncommon this proc is I'd rather it just be inlined, but macro is also fine.

code/modules/multiz/zmimic/mimic_movable.dm
```diff
@@ -81,16 +81,15 @@
 	return ..()
 
 /atom/movable/openspace/multiplier/proc/copy_lighting(atom/movable/lighting_overlay/LO, use_shadower_mult = TRUE)
-	appearance = LO
-	layer = MIMICED_LIGHTING_LAYER
-	plane = OPENTURF_MAX_PLANE
-	blend_mode = BLEND_MULTIPLY
-	set_invisibility(INVISIBILITY_NONE)
+	var/mutable_appearance/MA = mutable_appearance(LO)
```
Lohikar: Shouldn't this be `new /mutable_appearance()`? Afaict, this looks like it shouldn't work. No point paying the proc overhead for the native behavior anyways though.

code/_helpers/type2type.dm
out-of-phaze: ehhhh. i think it's fine as-is, making it a macro would lose the default argument and i don't want to have to manually specify it in everything that uses it for lights. also it's important to note that the proc overhead is negligible compared to the overhead of string interning in the current implementation (prior to this PR).

code/modules/multiz/zmimic/mimic_movable.dm
out-of-phaze: oh fuck lol you're right

code/_helpers/type2type.dm
out-of-phaze: made it two separate macros and replaced a bunch of uses of it that weren't necessary/were wasted string interns.

code/_helpers/unsorted.dm
```diff
@@ -773,16 +773,10 @@ var/global/list/WALLITEMS = list(
 	return 0
 
 /proc/get_random_colour(var/simple = FALSE, var/lower = 0, var/upper = 255)
-	var/colour
 	if(simple)
-		colour = pick(list("FF0000","FF7F00","FFFF00","00FF00","0000FF","4B0082","8F00FF"))
+		return pick(list("#ff0000","#ff7f00","#ffff00","#00ff00","#0000ff","#4b0082","#8f00ff"))
 	else
```
MistakeNot4892, line 778: `else` is now unnecessary

MistakeNot4892 (dismissed):

MistakeNot4892 (approved):

PR #3728, opened 2024-02-24
Make initial contents for most storage objects lazy-initialize

out-of-phaze:
## Description of changes

Most storage objects will now wait until their first interaction to create their contents. Storage objects that display their contents do not do this.

Notably, this will cause issues for a few things like `in world`s, which is why I want maintainer consensus about what, if anything, to do about that sort of thing.

## Changelog
:cl:
experiment: Most storage contents will not initialize their contents until first interacted with, which makes startup faster but could cause bugs. Report any issues on the issue tracker!
/:cl:


MistakeNot4892:
I'd like us to defer this one until after the storage rework as this is going to conflict.

out-of-phaze:
Abandoning, someone else can salvage parts of it if they'd like.

PR #3729, opened 2024-02-24
Make closets lazy-initialize their default contents

out-of-phaze:
## Description of changes

Most closets will now wait until their first interaction (with their contents, not necessarily by players) to create their contents.

Notably, this will cause issues for a few things like `in world`s, which is why I want maintainer consensus about what, if anything, to do about that sort of thing.

## Changelog
:cl:
experiment: Closets will not initialize their contents until first interacted with, which makes startup faster but could cause bugs. Report any issues on the issue tracker!
/:cl:


out-of-phaze:
Abandoning, someone else can salvage parts of it if they'd like.

PR #3730, opened 2024-02-24
Make ammo magazines lazy-initialize

out-of-phaze:
## Description of changes

Most ammo magazines will now wait until their first interaction (with their contents, not necessarily by players) to create their contents. Chambering/consuming a round from an uninitialized magazine will lower the initial ammo count by one and create exactly one round, further delaying initialization where possible.

## Changelog
:cl:
experiment: Ammo magazines will not initialize their contents until first interacted with, which makes startup faster but could cause bugs. Report any issues on the issue tracker!
/:cl:

### DIFF
--- a/code/modules/mechs/equipment/combat_projectile.dm
+++ b/code/modules/mechs/equipment/combat_projectile.dm
@@ -17,11 +17,11 @@
 
 /obj/item/gun/projectile/automatic/get_hardpoint_status_value()
 	if(!isnull(ammo_magazine))
-		return ammo_magazine.stored_ammo.len
+		return ammo_magazine.get_stored_ammo_count()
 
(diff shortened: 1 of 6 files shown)


MistakeNot4892:
I almost feel like it would be preferable for mags to never instantiate their full contents - outside of people manually loading in shells, we only ever care about the same time 50 times or whatever, which could be factored into worth and matter calc pretty easily without creating the full set.

out-of-phaze:
It should only generate them all when you dump out the magazine or do other things that access all of them, iirc? I was pretty careful to delay instantiation as much as possible.

MistakeNot4892 (approved):
I think this is worth testing, looks safe enough to me and vending machines already do something similar.

PR #3606, opened 2024-01-20
Moves dreaming from carbon to living.

MistakeNot4892:
## Description of changes
Moves dreaming from carbon to living.

## Why and what will this PR improve
`/carbon` delenda est. Also lets dogs dream, as god intended.

## Authorship
Myself.

## Changelog
:cl:
tweak: All living mobs can now dream.
/:cl:

PR #3733, opened 2024-02-24
Fixes tail icon crashes

### LINKED ISSUE #3706: Periodic crashes on stable
#### Description of issue
Human limb code appears to be causing pretty frequent crashes. Putting this in as a note to self to investigate.

Not crashing.

#### Steps to reproduce
???

Appears to be related to damage overlays.

Unclear.

Current stable.

#### Issue bingo

MistakeNot4892:
- Adds/verifies that tails have a basic organ icon in the bodytype dmi for use when severed. We could probably make them use `tail_icon` for this universally but that's a bit out of scope here.
- Skips retrieving the icon for parts with `skip_body_icon_draw` set (tails).
- Brings forward some limb masking and get_cached_marking parameter changes (organ instead of organ vars).

~~Maybe fixes #3706~~? ~~Does not fix #3606~~, Fixes #3706!


MistakeNot4892:
Tails are still invisible after being removed from the body, which I am very confused about, but the crashing appears to have been fixed.

out-of-phaze (approved):

PR #3745, opened 2024-02-26
Fluid turfs and fluid overlays adjustments/masking.

MistakeNot4892:
## Description of changes
- Exterior water turfs now use the depth system and fluid reagents.
- Exterior turfs only lose fluid if it is above ground level (height > 0).
- Movables on exterior water turfs, or turfs with shallow fluid, will get an alpha overlay covering their feet/legs.
- Fluids will get an alpha mask to make their edges a bit more visually distinct.
- Fluids will transition across map edges.
- Fluids cause movement slowdown.

More immersive (ha) water/fluid visuals.

## Authorship
I sprited/coded this from scratch, but the project was inspired by https://github.com/tgstation/TerraGov-Marine-Corps/pull/12864

## Changelog
:cl:
add: Water and fluids will now partially cover atoms within them.
tweak: Static fluid turfs are now dynamic fluid turfs.
/:cl:


out-of-phaze (dismissed):
Was tempted to say DNM until the right-click menu issue is fixed, but it's really not that big of a deal and this is a huge leap in fluid sim, so it's worth it. It might still be worth making a tracking issue for it, though.

out-of-phaze (approved):

PR #3748, opened 2024-02-27
Default map species with no external prosthetics will use the basic human for internal prosthetics.

MistakeNot4892:
Organ fabs use the map default species as a fallback when generating organs for recipes/worth init, so this has to be set to something.


comma (approved):

out-of-phaze (approved):
Gripes about how I prefer the witchcraft that is `||=` aside, this is good.

PR #3749, opened 2024-02-27
Fixes a weird runtime when generating recipes on a held stack.

MistakeNot4892:
Seems to be related to difficulty levels with no corresponding label, not sure what the cause is.


comma (approved):

out-of-phaze:
A little hesitant about fixes like this that don't solve the root cause of "skill value not in skill labels list," but it's just a stopgap.

PR #3750, opened 2024-02-27
Adds trenches

MistakeNot4892:
## Description of changes

Depends on #3759
- Exterior turfs can be dug out to make trenches using a shovel.
- Trenches will cause fall damage if moved over without creeping and must be climbed into or out of to safely traverse, unless flooded.

holes

## Authorship
Myself.

## Changelog
:cl:
add: Trenches/moats can now be dug in exterior turfs. 
/:cl:


code/modules/materials/definitions/solids/materials_solid_wood.dm
```diff
@@ -74,7 +75,9 @@
 	. += new/datum/stack_recipe/prosthetic/right_hand(src)
 	. += new/datum/stack_recipe/prosthetic/left_foot(src)
 	. += new/datum/stack_recipe/prosthetic/right_foot(src)
-	. += new/datum/stack_recipe/campfire(src)
+
+	if(stack_type == /obj/item/stack/material/log)
```
out-of-phaze: `if(ispath(stack_type)) return` on lines 46-47 will make this not work. Maybe make it a separate override, or split that check up somehow? Remember we still need to check holographic and reinf_material.

code/modules/materials/material_sheets.dm
```diff
@@ -314,7 +314,7 @@
 	if(plank_type && (IS_HATCHET(W) || IS_SAW(W)))
 		var/tool_type = W.get_tool_quality(TOOL_HATCHET) >= W.get_tool_quality(TOOL_SAW) ? TOOL_HATCHET : TOOL_SAW
 		if(W.do_tool_interaction(tool_type, user, src, 1 SECOND, set_cooldown = TRUE) && !QDELETED(src))
-			var/obj/item/stack/planks = new plank_type(loc, 1, material?.type, reinf_material?.type)
+			var/obj/item/stack/planks = new plank_type(loc, rand(2,4), material?.type, reinf_material?.type) // todo: change plank amount based on carpentry skillcheck
 			planks.add_to_stacks(user, TRUE)
```
out-of-phaze:
```suggestion
			var/obj/item/stack/planks = new plank_type(get_turf(src), rand(2,4), material?.type, reinf_material?.type) // todo: change plank amount based on carpentry skillcheck
			if(!planks.add_to_stacks(user, TRUE))
				user.put_in_hands(planks)
```
The put_in_hands part is optional, but as-is, using `loc` means planks cut from held logs will appear inside your body.

out-of-phaze (changes requested):

code/modules/materials/definitions/solids/materials_solid_wood.dm
MistakeNot4892: I need to systematize the stack type system I think. Another dep incoming...

code/modules/materials/material_sheets.dm
MistakeNot4892: Had to tweak add_to_stacks to have a meaningful return value but good call.

out-of-phaze (dismissed):

MistakeNot4892:
Need to do some review on how this all works once everything is in before it's merged.

MistakeNot4892:
Calling this done for now. I'm not happy with the layering or movement hook stuff, but I would like to get this in alongside the natural floor rewrite PR before I do any further integration/generalizing as the conflicts are going to suck already.

out-of-phaze (approved):
I might modify `can_backfill_turf_type` to populate a typecache, but aside from future changes this is fine as-is.

PR #3759, opened 2024-03-01
Digging tweaks from trench PR.

MistakeNot4892:
## Description of changes
- Moves digging and height logic into their own files.
- Digging a 'non-fundament' turf like grass or snow will turn it into dirt.

Small improvements to the experience of interacting with exterior turfs.

## Authorship
Myself.

## Changelog
:cl:
tweak: Snow and grass can be removed with a shovel.
/:cl:


out-of-phaze (dismissed):

code/game/turfs/exterior/_exterior_icon.dm
```diff
@@ -0,0 +1,56 @@
+/turf/exterior/on_update_icon()
```
comma, line 1: Is this supposed to check height? It's called to update neighbors in height setting but doesn't seem to actually care about height differences.

out-of-phaze: I think the height icon stuff is done in the trench PR itself?

MistakeNot4892: Yeah setting height doesn't actually do anything visual in this PR, I've pulled it in by accident.

comma (approved):

out-of-phaze (approved):

PR #3766, opened 2024-03-04
Adding a generalized ability handler and converting psionics to use it.

MistakeNot4892:
## Description of changes
- Adds `/datum/extension/abilities` and `/datum/ability_handler`.
- Replaces hardcoded psionics hooks in various attack procs with this extension.
- Converted `/datum/psi_complexus` with `/datum/ability_handler/psionics`.

Generalizes interaction hooks in a way that modpacks can take advantage of.

In the future I would like to fold `ability_master` and `/spell` into this system, as well as all the ancient nasty stuff on `/datum/mind` around spells.

## Authorship
Myself.

## Changelog
Nothing player-facing.


out-of-phaze (approved):
A relatively straightforward generalization of the Psi complexus handler and the inhand ability items.

PR #3771, opened 2024-03-07
Added log and brick wall icons.

MistakeNot4892:
- Adds icons for log and brick walls.
- Adds brick walls.
- Makes log and brick walls use their own dismantle sounds and products.
- Makes log walls craftable.
- Makes structure doors blend with turfs.

image
image


comma (approved):

PlayerDeer:
Honestly stone blocks go hard but in my biased opinion of loving contrast I think the logs could be more contrasty 👉👈

out-of-phaze (approved):
I do kind of wish they had more contrast but my girlfriend says they look fine, so like, I'll merge it as-is and someone can update the contrast later if they want.

PlayerDeer:
Valid tbh. I know I'm too lazy to change something like that

PR #3778, opened 2024-03-09
Reworks temperature equalization for atoms to be more measured.

MistakeNot4892:
## Description of changes
- Removes `temperature_coefficient`.
- Adds `get_thermal_mass()` and `get_thermal_mass_coefficient()`.

Atoms will not instantly equalize with their environment.
Larger atoms will equalize slower.

## Authorship
Myself with help from Penny.

## Changelog
:cl:
tweak: Atom temperature will equalize slower.
/:cl:


out-of-phaze (approved):
Ideally this would eventually factor in matter and material properties as well, but this is good enough for basic sim right now. Also, `ATOM_TEMPERATURE_EQUILIBRIUM_CONSTANT` delenda est, this should give behavior that's a lot more predictable.

PR #3779, opened 2024-03-09
Adding temperature_damage_threshold and removing use of material melting points.

MistakeNot4892:
## Description of changes
- Meat, skin, wood, paper, several other materials, no longer melt.
- Added `temperature_damage_threshold` which takes the minimum non-null of `melting_point`, `ignition_point` and `boiling_point`.
- Grilles, exosuits, walls and windows now use `temperature_damage_threshold` instead of `melting_point`.

Cloth melting was dumb and used only for wall damage.

## Authorship
Myself.

## Changelog
Nothing player-facing.

### DIFF
--- a/code/modules/materials/_materials.dm
+++ b/code/modules/materials/_materials.dm
@@ -348,6 +348,7 @@ INITIALIZE_IMMEDIATE(/obj/effect/gas_overlay)
 		solvent_melt_dose            = 0
 		solvent_max_damage           = 0
 		slipperiness                 = 0
+		bakes_into_at_temperature    = null
 		ignition_point               = null
 		melting_point                = null
 		boiling_point                = null
@@ -356,7 +357,7 @@ INITIALIZE_IMMEDIATE(/obj/effect/gas_overlay)
 		burn_product                 = null
 		vapor_products               = null
 	else if(isnull(temperature_damage_threshold))
-		for(var/value in list(ignition_point, melting_point, boiling_point))
+		for(var/value in list(ignition_point, melting_point, boiling_point, heating_point, bakes_into_at_temperature))
 			if(!isnull(value) && (isnull(temperature_damage_threshold) || temperature_damage_threshold > value))
 				temperature_damage_threshold = value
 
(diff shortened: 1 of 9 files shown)


code/modules/materials/_materials.dm
```diff
@@ -330,28 +332,33 @@ INITIALIZE_IMMEDIATE(/obj/effect/gas_overlay)
...
+		slipperiness                 = 0
+		ignition_point               = null
+		melting_point                = null
+		boiling_point                = null
+		temperature_damage_threshold = INFINITY
+		accelerant_value             = FUEL_VALUE_NONE
+		burn_product                 = null
+		vapor_products               = null
+	else if(isnull(temperature_damage_threshold))
+		for(var/value in list(ignition_point, melting_point, boiling_point))
```
out-of-phaze: Should `bakes_into_at_temperature` be considered as well? If your mech suddenly turns to ash above a certain temperature, that temperature is probably unsafe for operation. Same with heating point and chilling point, maybe?

Also this block might be able to be one line, albeit maybe a little clunkier. `min(list(ignition_point, melting_point, boiling_point) - new /list(3))`. This will subtract up to 3 nulls (the maximum possible) from the list, and take the lowest of what remains. This effectively just does a minimum in a list excluding nulls, and it avoids loop overhead at the cost of always doing a fixed number of list removal attempts scaling with the length of the list, though it might actually be batched to be more efficient?

code/modules/materials/_materials.dm
```diff
@@ -330,28 +332,33 @@ INITIALIZE_IMMEDIATE(/obj/effect/gas_overlay)
...
+		solvent_melt_dose            = 0
+		solvent_max_damage           = 0
+		slipperiness                 = 0
+		ignition_point               = null
+		melting_point                = null
+		boiling_point                = null
+		temperature_damage_threshold = INFINITY
+		accelerant_value             = FUEL_VALUE_NONE
+		burn_product                 = null
+		vapor_products               = null
```
out-of-phaze, line 358: Seems like the `bakes_into_*` variables should be unset for holomaterials as well? Unless we want to be able to fire holoclay into holobrick...

code/modules/materials/_materials.dm
MistakeNot4892: I don't think it's worth the clunky syntax and list creation tbqh.

out-of-phaze (approved):

PR #3780, opened 2024-03-09
Corrects brain replacement in synthbrain aspect.

MistakeNot4892:
Synthbrain aspect no longer kills you when your old brain hard dels.

### DIFF
--- a/code/modules/brain_interface/_brain_interface.dm
+++ b/code/modules/brain_interface/_brain_interface.dm
@@ -122,7 +122,6 @@
 		rig.forced_move(direction, user)
 
 /obj/item/organ/internal/brain_interface/Destroy()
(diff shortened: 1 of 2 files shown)


code/modules/brain_interface/_brain_interface.dm
```diff
@@ -122,7 +122,6 @@
 		rig.forced_move(direction, user)
 
 /obj/item/organ/internal/brain_interface/Destroy()
-	STOP_PROCESSING(SSprocessing, src)
```
MistakeNot4892, line 125: Random unrelated runtime noticed during debugging. It never queues on SSprocessing.

out-of-phaze (approved):

PR #3781, opened 2024-03-09
Seeds spawning rework.

MistakeNot4892:
## Description of changes
- Seed packets take a seed argument in Initialize().
- `seed_type` has been merged with `seed` on seed packets.
- `/obj/structure/flora/plant` will spawn some seeds on being cut down.
- `/obj/structure/proc/dismantle()` is now `/obj/structure/proc/dismantle_structure(mob/user)`.

Nicer seeds code, nicer survivalist experience, more searchable dismantle proc.

## Authorship
Myself.
## Changelog
:cl:
tweak: Exoplanet plants will now drop seeds.
/:cl:


out-of-phaze (commented):
Is there a reason `/obj/item/chems/food/grown/var/plantname` wasn't also merged with its `seed` var?

MistakeNot4892:
Because it's unrelated to seed packets?

out-of-phaze (changes requested):
The chemistry unit test might need to explicitly test the predefined seed datums so that fruits that spawn acid don't destroy themselves. Alternatively, it'd be faster to just exclude growns from the test.

out-of-phaze (approved):

PR #3782, opened 2024-03-09
Reworking fire sources (campfires etc) and how they handle fuel and temperature.

MistakeNot4892:
## Description of changes
- Fire sources no longer use wood as a material. They are made out of stone/iron/etc.
- Fire sources use `burn_temperature` on their fuel material to determine how hot they burn.

Better fire source code.
Supporting code for kilns/pottery.

## Authorship
Myself and Penny.

## Changelog
Nothing player-facing.


comma (dismissed):

code/modules/materials/definitions/solids/materials_solid_wood.dm
```diff
@@ -21,6 +21,7 @@
 	brute_armor = 1
 	weight = MAT_VALUE_NORMAL
 	melting_point = T0C+300 //okay, not melting in this case, but hot enough to destroy wood
```
out-of-phaze:
```suggestion
	melting_point = null // Don't give us molten wood, please.
```
this is an actual issue in testing, put too many logs in at once and the excess melt :(

code/game/objects/structures/fires.dm
```diff
@@ -177,24 +186,20 @@
...
 	return (lit == FIRE_LIT)
 
 /obj/structure/fire_source/attackby(var/obj/item/thing, var/mob/user)
 
-	if(ATOM_SHOULD_TEMPERATURE_ENQUEUE(thing) && user.a_intent != I_HURT && lit == FIRE_LIT)
-		thing.handle_external_heating(DIRECT_HEAT, src, user)
+	if(ATOM_IS_OPEN_CONTAINER(thing) && thing.reagents?.total_volume)
+		user.visible_message(SPAN_DANGER("\The [user] pours the contents of \the [thing] into \the [src]!"))
+		take_reagents(thing.reagents)
 		return TRUE
```
out-of-phaze: Is this necessary? It seems redundant in the face of `standard_pour_into` existing on reagent containers, which also handles things like food better. Moreover this just seems to have far too high of a priority, preventing you from heating things or adding stuff to the fire if they have reagents.

code/game/objects/structures/fires.dm
```diff
@@ -213,58 +218,56 @@
...
+		try_light(thing.get_heat())
 		return TRUE
 
-	if((lit != FIRE_LIT || user.a_intent == I_HURT) && user.try_unequip(thing, src))
-		user.visible_message(SPAN_NOTICE("\The [user] drops \the [thing] into \the [src]."))
+	if((lit != FIRE_LIT || user.a_intent == I_HURT))
+		// Only drop in one log at a time.
+		if(istype(thing, /obj/item/stack))
+			var/obj/item/stack/stack = thing
+			thing = stack.split(1)
```
out-of-phaze, line 306: This will eventually cause issues for putting bricks into kilns. RSI is no joke, I think we should try to avoid repeated clicking.

code/game/objects/structures/fires.dm
```diff
@@ -213,58 +218,56 @@
...
+/obj/structure/fire_source/proc/process_fuel(ignition_temperature)
 
 	if(fuel >= IDEAL_FUEL)
 		return TRUE
 
+	// Slowly lose burn temperature.
+	// TODO: use temperature var and equalizing system?
+	last_fuel_burn_temperature = max(ignition_temperature, last_fuel_burn_temperature)
+	if(last_fuel_burn_temperature > T20C)
+		last_fuel_burn_temperature = max(T20C, round(last_fuel_burn_temperature * 0.9))
```
out-of-phaze, line 323: It's important to note that this is exponential. `0.9 ** ticks`. If fuel is less than `IDEAL_FUEL`, it will take 6 ticks (about 12 seconds) to reach half temperature, 20 ticks (about 40 seconds) to reach 10% of the temperature, etc. To go from 1000C to 20C it'd take 14 ticks (28 seconds), and this is just while `fuel < IDEAL_FUEL`. I would suggest changing the coefficient to something like 0.98 or 0.99, or gating the temperature loss by `fuel < LOW_FUEL`.

code/game/objects/effects/effect_system.dm
```diff
@@ -98,6 +98,7 @@ steam.start() -- spawns the effect
 	icon = 'icons/effects/effects.dmi'
 	anchored = TRUE
 	mouse_opacity = MOUSE_OPACITY_UNCLICKABLE
+	pass_flags = PASS_FLAG_TABLE | PASS_FLAG_GRILLE
```
out-of-phaze: `/obj/structure/fire_source` probably needs a `CanPass` override to make this work
```
/obj/structure/fire_source/CanPass(atom/movable/mover, turf/target, height=0, air_group=0)
	return ..() || (istype(mover) && mover.checkpass(PASS_FLAG_TABLE))
```
Also, should consider if we really want sparks to go through grilles or not. I think just tables is fine?

out-of-phaze (dismissed):

code/game/objects/structures/fires.dm
MistakeNot4892: Bricks into kilns use a separate interaction. This is to simplify fuel processing.

code/modules/materials/definitions/solids/materials_solid_wood.dm
MistakeNot4892: https://github.com/NebulaSS13/Nebula/pull/3779

MistakeNot4892:
Will depend on #3779 due to issues with the fuel melting.

out-of-phaze (dismissed):

out-of-phaze (dismissed):

out-of-phaze (approved):
It looks like that was just a conflict resolution, so we should be good once my prior approval hits 24 hours.

PR #3783, opened 2024-03-09
Collapses edge mimic turfs down into a single shared type.

MistakeNot4892:
## Description of changes
Merges mimic_edge types down to a base type.

We don't need all three separate types implementing identical behavior.

## Authorship
Myself.

## Changelog
Nothing player-facing.


out-of-phaze (approved):

out-of-phaze (changes requested):
~~Needs a map migration file, actually, but after that it'll be good.~~ wait, these are automatically spawned, nevermind

out-of-phaze (approved):

PR #3784, opened 2024-03-09
Removes /turf/simulated/shuttle.

MistakeNot4892:
## Description of changes
- Replaces `/turf/simulated/shuttle` with `/turf/simulated/wall/shuttle`.
- Breaks `shuttle.dmi` up between the various places it's used.
- Adds a unique turf ID that is checked when walls are blending together.

`/simulated` must die. Also adds some more flexibility to mapping in general.

## Authorship
Myself.

## Changelog
Nothing player-facing.


out-of-phaze (changes requested):
Needs a map migration file, but after that it's good to go.

out-of-phaze (approved):

PR #3786, opened 2024-03-09
Adding a recycler.

MistakeNot4892:
## Description of changes
- Removes item recycling from fabricators. Fabs can generally only take stacks now, with some exceptions for biomass.
- Adds a recycler machine similar to the fabricator that turns items in storage into cubes. By default this works at 65% efficiency, so as to avoid making ore processing pointless by letting the recycler process ores.
- Maps recyclers into the main maps near the fabricators.

Fixes three big issues:
- Accidentally dicing your ID with one click.
- Wasting materials by putting the item into a lathe that doesn't use the extra mats.
- Wasting item contents by putting an item with contents into a lathe.

'Fixes' this issue: https://github.com/ScavStation/ScavStation/issues/903

## Authorship
Myself.

## Changelog
:cl:
tweak: Fabricators like autolathes can no longer recycle arbitrary items.
add: A recycler machine has been added and should be found near fabricators.
/:cl:


MistakeNot4892:
Some possible TODO for future PRs:
- ~~Hopper capacity should depend on matter bin quality.~~
- ~~Material returns should be based on microlaser/manipulator quality.~~
- Adjacent autolathes should take material directly from a recycler until full.
- Conveyors bumping items into recyclers should autoload them.

out-of-phaze (approved):
I'm not entirely sure about removing `has_recycler` entirely rather than just defaulting it to off, but it works either way, really.

PR #3789, opened 2024-03-12
Fishes are less suicidal.

MistakeNot4892:
Stops fishes wandering out of the water.


code/modules/mob/living/simple_animal/simple_animal.dm
```diff
@@ -227,6 +227,15 @@ var/global/list/simplemob_icon_bitflag_cache = list()
 	do_delayed_life_action()
 	performing_delayed_life_action = FALSE
 
+/mob/living/simple_animal/proc/turf_is_safe(turf/target)
+	if(!istype(target))
+		return FALSE
+	if(target.is_open() && target.has_gravity() && !can_overcome_gravity())
+		return FALSE
+	if(is_aquatic != !target.submerged())
```
out-of-phaze: The double negative here makes me worry this check is inverted. This means an aquatic creature would consider a dry tile safe and a wet tile dry.

out-of-phaze (changes requested):

code/modules/mob/living/simple_animal/simple_animal.dm
MistakeNot4892: Ah yep I didn't commit that fix.

out-of-phaze (approved):

PR #3791, opened 2024-03-13
[MDB IGNORE] Removed /simulated.

MistakeNot4892:
## Description of changes
- Removes `/turf/simulated`.
- Changes `/turf/simulated/floor` to `/turf/floor`.
- Changes `/turf/simulated/wall` to `/turf/wall`.

Unifying turf behavior at root, laying the groundwork for basing exterior turfs on floor turfs. All the actual behavior from simulated has been relocated already, this is just removing the abstract type.

## Authorship
Myself.

## Changelog
Nothing player-facing.


deathride58 (approved):
if it compiles it works
image

out-of-phaze (approved):

PR #3792, opened 2024-03-13
[MDB IGNORE] Repaths /turf/exterior/wall to /turf/wall/natural.

MistakeNot4892:
## Description of changes
- Unifies exterior walls with regular walls.
- Optimizes some list tracking on natural walls in the xenoarch subsystem.
- Reorganizes the `simulated` folder into `walls` and `floors`, moves natural walls into that directory instead of exterior.

Unifies code and material handling, unifies behavior, generally simplifies the turf typetree.

Depends on #3791.

## Authorship
Myself.

## Changelog
Nothing player-facing.


code/game/turfs/exterior/_exterior.dm
```diff
@@ -18,8 +18,6 @@
 	var/icon_edge_layer = -1
 	var/icon_edge_states
 	var/icon_has_corners = FALSE
-	///If this turf is on a level that belongs to a planetoid, this is a reference to that planetoid.
-	var/datum/planetoid_data/owner
```
out-of-phaze, line 22: Surprised this was okay, but it really is only used in exterior's `Initialize()` and in `SSmaterials.get_rock_color()`.

code/modules/maps/helper_landmarks.dm
```diff
@@ -78,8 +78,8 @@ INITIALIZE_IMMEDIATE(/obj/abstract/landmark/map_load_mark)
 	var/turf/wall/simulated_wall = get_turf(src)
 	if(istype(simulated_wall))
 		simulated_wall.dismantle_wall(TRUE, TRUE, TRUE)
-	else if(istype(simulated_wall, /turf/exterior/wall))
-		var/turf/exterior/wall/exterior_wall = simulated_wall
+	else if(istype(simulated_wall, /turf/wall/natural))
+		var/turf/wall/natural/exterior_wall = simulated_wall
 		exterior_wall.dismantle_wall(TRUE)
```
out-of-phaze: This might be able to be simplified somehow now, but I'm not totally sure. Approving since, you know, I have no clue.

out-of-phaze (dismissed):
Moving the logic upwards into `/turf/Initialize()` will probably lead to an increase in init time. Remind me to profile and see if there's anything we can do to speed it up.

code/modules/maps/helper_landmarks.dm
MistakeNot4892: Ah yeah this looks redundant now.

code/game/turfs/exterior/_exterior.dm
MistakeNot4892: Yeah I was surprised too, but it's really limited in use.

MistakeNot4892:
Init impact should be minimal, one istype() per turf - space is the primary turf in world.area and doesn't call parent on Initialize.

out-of-phaze:
> Init impact should be minimal, one istype() per turf - space is the primary turf in world.area and doesn't call parent on Initialize.

Assumptions like this on TG led to forks with a non-standard world.turf having degraded performance, but fair enough.

code/modules/maps/helper_landmarks.dm
out-of-phaze: Important to note the args, though. That's always been the main difference.

MistakeNot4892: Mostly because exterior reimplemented the proc entirely and didn't have those args. I think it should be fine now, will check.

MistakeNot4892: Made the overrides use the same args but with ramp updates for natural, natural overrides the dismantled product logic to ignore the extra args.

out-of-phaze (approved):

PR #3793, opened 2024-03-13
[MDB IGNORE] [IDB IGNORE] Unifying /turf/exterior with /turf/floor.

MistakeNot4892:
## Description of changes
- Repaths /turf/exterior to /turf/floor/natural.
- Audits the entirety of the floor type tree, repathing or removing a few.
- Repaths several exterior turfs to the regular turf type (paths and concrete).
- Adds unit testing and validation for floor icon states and overlays.
- Restructures and cleans up the flooring and floor turf directories.

Has dependencies on #3792 and #3791. Currently opening for CI.

Working towards a unified flooring system and adding validation for what is currently a bit of a nightmare.

## Authorship
Myself.

## Changelog
Should be mostly nothing player-facing.


code/controllers/subsystems/fluids.dm
```diff
@@ -129,7 +129,7 @@ SUBSYSTEM_DEF(fluids)
 		// Wash our turf.
 		current_fluid_holder.fluid_act(reagent_holder)
 
-		if(isspaceturf(current_fluid_holder) || (istype(current_fluid_holder, /turf/exterior) && (current_fluid_holder.reagents?.total_volume + current_fluid_holder.get_physical_height()) > 0))
+		if(isspaceturf(current_fluid_holder) || (istype(current_fluid_holder, /turf/floor) && current_fluid_holder.is_outside() && (current_fluid_holder.reagents?.total_volume + current_fluid_holder.get_physical_height()) > 0))
```
MistakeNot4892: Not completely sure about this one. The old behavior was mainly to prevent exoplanets flooding, and this does a similar job.

code/modules/overmap/ships/computers/comms.dm
```diff
@@ -196,7 +196,7 @@
 	if(!istype(T))
 		return FALSE
 	for(var/turf/neighbor in RANGE_TURFS(T, 1))
-		if(istype(neighbor, /turf/space) || istype(neighbor, /turf/exterior))
+		if(istype(neighbor, /turf/space) || neighbor.is_outside())
```
MistakeNot4892, line 199: This is another one I'm not totally sure about. It's used by sensors to ensure you don't build them inside, so even in a case where they're built inside and the roof is removed so it has LOS to the sky, that seems fine to me.

code/modules/multiz/ladder.dm
```diff
@@ -42,7 +42,7 @@
 		var/turf/T = get_turf(src)
 		if((locate(/obj/structure/ladder) in GetBelow(src)) && (!(locate(/obj/structure/lattice) in loc) || !T.is_open()))
 			var/old_turf_type = T.type
-			T.ReplaceWithLattice()
+			T.physically_destroyed() // this does not feel correct, but it's the proc that clears the turf and spawns a lattice
```
MistakeNot4892: Pointing this out for review.

code/modules/atmospherics/datum_pipeline.dm
```diff
@@ -231,21 +231,6 @@
 		else
 			target.air.temperature += sharer_temperature_delta
 
-	else if(istype(target, /turf/exterior) && !target.blocks_air)
```
MistakeNot4892, line 234: This block is redundant in view of the block immediately above it.

code/game/turfs/floors/subtypes/floor_plating.dm
```diff
@@ -0,0 +1,6 @@
+// simulated/floor is currently plating by default, but there really should be an explicit plating type. // Why?
```
MistakeNot4892: Reiterating: why do we need a redundant subtype when `/turf/floor` is functionally identical?

code/game/turfs/floors/natural/natural_ice.dm
```diff
@@ -0,0 +1,44 @@
...
+			. += my_species.snow_slowdown_mod
+		. = max(., 0)
+
+/turf/floor/natural/snow/fire_act(datum/gas_mixture/air, exposed_temperature, exposed_volume)
+	handle_melting()
+	return ..()
+
+/turf/floor/natural/snow/handle_melting(list/meltable_materials)
+	. = ..()
+	ChangeTurf(/turf/floor/natural/ice)
```
MistakeNot4892: Noting to self that this should probably just turn snow and ice into mud with water.

code/game/turfs/floors/natural/_natural_icon.dm
```diff
@@ -40,9 +40,9 @@
 				var/res = (neighbors & direction)
 				if(res == 0)
 					draw_state = "edge[direction]"
-				else if(res == direction)
+				else if(res == direction && icon_has_corners)
 					draw_state = "corner[direction]"
-				if(draw_state && check_state_in_icon(draw_state, icon))
```
MistakeNot4892: All these `check_state_in_icon` should in theory be handled by the new unit test.

code/game/turfs/floors/natural/_natural.dm
```diff
@@ -51,33 +49,29 @@
...
 	update_icon()
-
 	if(reagent_type && height < 0)
 		add_to_reagents(reagent_type, abs(height))
 
-/turf/exterior/on_reagent_change()
+/turf/floor/natural/on_reagent_change()
 	. = ..()
-	if(reagent_type && height < 0 && reagents && reagents.total_volume < abs(height))
+	if(!QDELETED(src) && reagent_type && height < 0 && !QDELETED(reagents) && reagents.total_volume < abs(height))
```
MistakeNot4892, line 96: This may not be super necessary, my unit test does some extremely weird shit with turf spawning and was runtiming here.

code/modules/overmap/ships/computers/comms.dm
out-of-phaze: I think it's also useful for preventing sensors from working on landable ships while in a mothership's hangar, so this solution works fine as well.

code/game/turfs/floors/natural/natural_ice.dm
out-of-phaze: Noting that originally the resulting turf was named "permafrost," which should *probably* be (mostly-)unmeltable ice? Frozen soil? Dunno.

code/controllers/subsystems/fluids.dm
out-of-phaze: The change here seems fine, but the clause directly after might cause issues in other PRs (particularly, trenches—things like gently sloping hillsides and flat, slightly-inset areas could cause problems. Noting for future concern.

code/game/turfs/floors/natural/_natural_icon.dm
out-of-phaze: As long as there's no runtime icon mutation/creation, yeah.

code/game/turfs/floors/subtypes/floor_plating.dm
out-of-phaze: Probably user-friendliness? When searching for plating in build mode/a map editor/etc., something with the word "plating" in it is a lot clearer. Otherwise, you'd just have to guess that `/turf/floor` is plating.

code/modules/multiz/ladder.dm
out-of-phaze: I mean, I kind of think `physically_destroyed()` should be calling a helper so that on subtypes that expect *destruction* we don't do things like spawn shrapnel, but that's what it did before and you removed said helper in favor of this. If you think it's alright, it's alright.

code/game/turfs/floors/natural/_natural.dm
out-of-phaze: Your unit test probably uncovered a real issue with ChangeTurf on flooded turfs, so I think this is fine.

code/game/turfs/turf_changing.dm
```diff
@@ -1,28 +1,23 @@
...
+	return src
 
+/turf/physically_destroyed(var/skip_qdel)
+	SHOULD_CALL_PARENT(FALSE)
+	var/turf/new_turf = switch_to_base_turf()
+	if(istype(new_turf) && !new_turf.is_open() && !(locate(/obj/structure/lattice) in new_turf))
+		new /obj/structure/lattice(new_turf)
+	return !!new_turf
+
+/turf/floor/natiral
```
out-of-phaze:
```suggestion
```

code/modules/atmospherics/datum_pipeline.dm
out-of-phaze: This probably predates atmos unification and so the istype should probably be replaced with `(target.external_atmosphere_participation && target.is_outside()` or something similar.

out-of-phaze (dismissed):
This was originally going to be an approval, but I noticed one or two issues that required changes. Most of the comments are just comments and not requests for changes.

code/game/turfs/turf_changing.dm
MistakeNot4892: Where did THAT come from...

code/controllers/subsystems/fluids.dm
MistakeNot4892: I think it should probably be fine, since turfs being inset like that kind of implies you _want_ them to interact with fluids. Rain flowing downhill to pool at the bottom seems expected for example. Unsure though, we probably need a turf flag or something to absorb water.

code/modules/multiz/ladder.dm
MistakeNot4892: The helper was pretty specific to constructed turfs, is the thing.
I think what I'll do is generalize `dismantle_wall` to turf.

code/game/turfs/floors/subtypes/floor_plating.dm
MistakeNot4892: I put a stub type with no changes in _floor.dm instead since it's literally the exact same turf.

code/game/turfs/floors/natural/natural_ice.dm
MistakeNot4892: It doesn't make a huge amount of sense for permafrost to be caused by fire, though.

code/modules/multiz/ladder.dm
out-of-phaze: I think that has an argument to not spawn dismantle products, so that's fine.

code/game/turfs/floors/natural/natural_ice.dm
out-of-phaze: I assumed the intent was that the snow on top was melted away to reveal the permafrost beneath it, but now that we have actual turf height stuff, there may be a better way to represent snow accumulation and melting down the line.

code/controllers/subsystems/fluids.dm
MistakeNot4892: I added a turf flag for a subset of natural turfs to absorb fluid instead of doing it based on type.

code/modules/atmospherics/datum_pipeline.dm
```diff
@@ -231,6 +231,21 @@
 		else
 			target.air.temperature += sharer_temperature_delta
 
+	else if(target.external_atmosphere_participation && !target.blocks_air)
```
out-of-phaze, line 234: Does this need `target.is_outside()`? If `target` has `zone_membership_candidate = FALSE` but `external_atmosphere_participation = TRUE` while being indoors, it'd use `target.return_air()`. In fact, why do we not use that in all cases? Maybe this block should just check `!target.blocks_air`, since the `else` block ignores air entirely.

MistakeNot4892:
There's an inconsistent CI failure in my new floor tests where it can't ChangeTurf as expected. Not quite sure what could be causing that at random.

out-of-phaze (changes requested):
> There's an inconsistent CI failure in my new floor tests where it can't ChangeTurf as expected. Not quite sure what could be causing that at random.

The current one seems to be because of this:
```dm
/turf/floor/fixed/alium/ruin/Initialize()
	. = ..()
	if(prob(10))
		ChangeTurf(get_base_turf_by_area(src))
```
It changes to another turf 10% of the time, and that throws off your test, I assume.

code/modules/atmospherics/datum_pipeline.dm
out-of-phaze: this was an old comment and not intended to be posted :p

MistakeNot4892:
> ```dm
> urf/floor/fixed/alium/ruin/Initializ
> ```

I turned the logic into a landmark which should hopefully sort it out.

out-of-phaze (dismissed):

MistakeNot4892:
Carpets are broken :I

out-of-phaze (dismissed):

out-of-phaze (dismissed):

out-of-phaze (approved):

deathride58 (approved):

PR #3795, opened 2024-03-14
Adds a mortar

out-of-phaze:
## Description of changes
Adds mortars, used to grind small, soft objects into their reagents.

Low-tech option for chemical processing from fruits, etc.

## Authorship
Me.

## Changelog
:cl:
add: Adds a mortar. Add an item to the mortar and use an object made of a heavy material to crush it, at the cost of stamina.
/:cl:


code/modules/reagents/reagent_containers/mortar.dm
```diff
@@ -0,0 +1,60 @@
...
+		to_chat(user, SPAN_NOTICE("You add \the [using_item] to \the [src] for grinding."))
+		return TRUE
+	else if(using_item.force > 0 && attacking_material && crushing_material)
+		if(attacking_material.hardness <= crushing_material.hardness)
+			to_chat(user, SPAN_NOTICE("\The [using_item] is not hard enough to crush \the [crushing_item]."))
+			return TRUE
+		if(REAGENTS_FREE_SPACE(reagents) < crushing_item.reagents.total_volume)
+			to_chat(user, SPAN_WARNING("\The [src] is too full to grind \the [crushing_item], it'd spill everywhere!"))
+			return TRUE
+		if(crushing_item.reagents.total_volume) // if it has no reagents, skip all the fluff and destroy it instantly
```
MistakeNot4892: `reagents?` is probably needed here.

code/modules/reagents/reagent_containers/mortar.dm
```diff
@@ -0,0 +1,60 @@
...
+	var/skill_factor = CLAMP01(1 + 0.3*(user.get_skill_value(SKILL_CHEMISTRY) - SKILL_EXPERT)/(SKILL_EXPERT - SKILL_MIN))
+	if(!crushing_item)
+		if(!using_item.reagents?.total_volume)
+			to_chat(user, SPAN_NOTICE("\The [using_item] is not suitable for grinding."))
+			return TRUE
+		if(!user.try_unequip(using_item, src))
+			return FALSE
+		to_chat(user, SPAN_NOTICE("You add \the [using_item] to \the [src] for grinding."))
+		return TRUE
+	else if(using_item.force > 0 && attacking_material && crushing_material)
```
MistakeNot4892: You could remove this `else` and then deindent the entire rest of the block, which might help with readability. Not blocking.

code/modules/reagents/reagent_containers/mortar.dm
```diff
@@ -0,0 +1,60 @@
...
+		return
+	if(!istype(using_item))
+		return FALSE
+	if(!CanPhysicallyInteract(user) || !user.check_dexterity(DEXTERITY_HOLD_ITEM))
+		return TRUE
+	var/list/contained_atoms = get_contained_external_atoms()
+	var/obj/item/crushing_item = LAZYACCESS(contained_atoms, 1)
+	var/decl/material/attacking_material = using_item.get_material()
+	var/decl/material/crushing_material = crushing_item?.get_material()
+	var/skill_factor = CLAMP01(1 + 0.3*(user.get_skill_value(SKILL_CHEMISTRY) - SKILL_EXPERT)/(SKILL_EXPERT - SKILL_MIN))
```
MistakeNot4892, line 29: I could have sworn there was a helper for this, but I can't find it now. Might be worth putting it in a define or proc.

code/modules/reagents/reagent_containers/mortar.dm
```diff
@@ -0,0 +1,60 @@
...
+	material_alteration = MAT_FLAG_ALTERATION_ALL
+
+// todo: generalize to use TOOL_PESTLE
+/obj/item/chems/glass/mortar/attack_self(mob/user)
+	var/list/contained_atoms = get_contained_external_atoms()
+	var/obj/item/crushing_item = LAZYACCESS(contained_atoms, 1)
+	if(crushing_item)
+		crushing_item.dropInto(loc || get_turf(user))
+		return TRUE
+
```
MistakeNot4892, line 17: `return ..()`

code/modules/reagents/reagent_containers/mortar.dm
```diff
@@ -0,0 +1,60 @@
...
+	var/obj/item/crushing_item = LAZYACCESS(contained_atoms, 1)
+	var/decl/material/attacking_material = using_item.get_material()
+	var/decl/material/crushing_material = crushing_item?.get_material()
+	var/skill_factor = CLAMP01(1 + 0.3*(user.get_skill_value(SKILL_CHEMISTRY) - SKILL_EXPERT)/(SKILL_EXPERT - SKILL_MIN))
+	if(!crushing_item)
+		if(!using_item.reagents?.total_volume)
+			to_chat(user, SPAN_NOTICE("\The [using_item] is not suitable for grinding."))
+			return TRUE
+		if(!user.try_unequip(using_item, src))
+			return FALSE
```
MistakeNot4892: I think you still want to return TRUE here, FALSE indicates further processing should happen.

MistakeNot4892 (changes requested):
Bunch of minor stuff, generally looks fine to me.

code/modules/reagents/reagent_containers/mortar.dm
out-of-phaze: It's used in exactly one other place, grinder code. If it were used in a few more, I'd think about it, but as-is I'm not sure what I'd name the macro, what the parameters would be, or where I'd define it in the codebase.

MistakeNot4892 (approved):

code/modules/reagents/reagent_containers/mortar.dm
```diff
@@ -0,0 +1,62 @@
...
+	var/skill_factor = CLAMP01(1 + 0.3*(user.get_skill_value(SKILL_CHEMISTRY) - SKILL_EXPERT)/(SKILL_EXPERT - SKILL_MIN))
+	if(!crushing_item)
+		if(!using_item.reagents?.total_volume)
+			to_chat(user, SPAN_NOTICE("\The [using_item] is not suitable for grinding."))
+			return TRUE
+		if(!user.try_unequip(using_item, src))
+			return TRUE
+		to_chat(user, SPAN_NOTICE("You add \the [using_item] to \the [src] for grinding."))
+		return TRUE
+	if(using_item.force <= 0 || !attacking_material || !crushing_material)
```
comma, line 38: Do plants or pills have a material set? Wouldn't be able to grind them without that it seems

MistakeNot4892: Pills currently don't, but plants have plantmatter or pith set I think.

out-of-phaze: `material = /decl/material/solid/organic/plantmatter` for pills, for some reason

PR #3796, opened 2024-03-14
Fix recipe/material issues, adds missing brick presets

out-of-phaze:
## Description of changes
Adds materials and stumps to dead trees.
Fixes hardness-based recipes not checking hardness.
Adds more stone brick types and adjusts melting point for basalt, granite, etc.
Readds the log wall recipe.
Fixes recipes not being able to create more than one result object.
Fixes blank amount in stack crafting message.
Fixes `one_per_turf` recipes still letting you bulk-create objects on the same turf.

Some issues I noted while playtesting Kobaloi with recent changes.

## Authorship
Me.

### DIFF
--- a/code/game/objects/items/stacks/stack.dm
+++ b/code/game/objects/items/stacks/stack.dm
@@ -153,7 +153,9 @@
 
 /obj/item/stack/proc/produce_recipe(decl/stack_recipe/recipe, var/quantity, mob/user)
 	var/required = quantity*recipe.req_amount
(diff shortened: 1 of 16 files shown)


MistakeNot4892 (dismissed):

out-of-phaze:
got an issue, making 1x drying rack creates 3 drying racks somehow...?

MistakeNot4892 (approved):

PR #3716, opened 2024-02-23
[MDB IGNORE] Converts storage to a datum (sorry)

MistakeNot4892:
DNM, opening for CI.

## Description of changes
- Merges behavior from `/obj/item/suit/storage` `/obj/item/storage` `/obj/structure/mirror` into `/datum/extension/storage`.
- Removes/repaths storage types globally.
- Replaces typechecks etc. with extension checks.
- Adds a new contents getter, `/atom/proc/get_stored_inventory()`, which defaults to `get_contained_external_atoms()`.

## TODO
- [X] Coarse conversion.
- [X] Repaths to remove redundant types.
- [x] Test basic storage functionality.
- [x] Test pockets functionality on suits.
- [x] Allow objects to be removed from structure inventory.
- [x] Work out how to separate mirror interface and storage interactions.
- [x] Test structure functionality on mirrors.
- [x] Test pneumatic cannon functionality (or rewrite)
- [x] Consider `get_contained_external_atoms()` in storage contents helper?
- [X] Test post-conversion to datum.
- [X] Test recycler.
- [x] Fix holders not closing UI on move away from mob (throwing your backpack)
- [x] Fix storage not updating UI on qdel of item (recycler).

## Why and what will this PR improve
Less redundant storage code, more flexible storage functionality, shorter typepaths generally. Lays the groundwork for some clothing and accessory stuff I want to do down the track.

## Authorship
Myself.

## Changelog
Nothing player-facing.

PR #3798, opened 2024-03-16
[MDB IGNORE] [IDB IGNORE] Dismantling accessories and toggleable clothing.

MistakeNot4892:
VERY WIP, opening for CI/diff.

## Description of changes
- Removes/repaths all /accessory items.

Depends on #3805 #3716 #3806 #3807

Simplifies the type tree, makes behavior sharable without tying it to type. Working towards the pants PR as being able to freely stack/mix and match accessories with clothes will make it much, much easier.

## Authorship
Myself.

## Changelog
Ideally nothing player-facing.


MistakeNot4892:
I am definitely going to have to split this into several PRs. At least:
- ~~Accessory logic being moved to `/obj/item/clothing`~~
- ~~Clothing modifiers~~
- ~~Bodytype slot modifiers~~

MistakeNot4892:
Also depends on the slot storage rework so I can further maul the storage accessories.

out-of-phaze (approved):

PR #3801, opened 2024-03-16
Adjust exterior temperature for cold weather

out-of-phaze:
## Description of changes
Adds a `/cold` weather state, which is like `/calm` but colder, and can precede snowstorms.
`/cold`, `/snow`, and the other snow subtypes will lower outdoor temperature on their level by varying amounts.
Fixes `weather.adjust_temperature` not actually working.

Future changes: maybe I'll make weather states define a max and min temperature adjustment (and max/min temperature) and have the weather state machine object pick one at random?

Fleshes out some stub code for weather states adjusting level temperature. Also fixes a bug with `adjust_temperature()` not actually affecting the temperature if there were no fire sources around.

## Authorship
Me.

## Changelog
:cl:
add: Certain weather conditions now affect exterior temperature. Cold and snowy conditions will reduce the outdoor temperature accordingly.
/:cl:

### DIFF
--- a/code/modules/ZAS/Turf.dm
+++ b/code/modules/ZAS/Turf.dm
@@ -258,12 +258,15 @@
 	if(!include_heat_sources)
 		return gas
 
-	var/initial_temperature = weather ? weather.adjust_temperature(gas.temperature) : gas.temperature
+	if(weather)
+		gas.temperature = weather.adjust_temperature(gas.temperature)
+	var/initial_temperature = gas.temperature
 	if(length(affecting_heat_sources))
 		for(var/obj/structure/fire_source/heat_source as anything in affecting_heat_sources)
 			gas.temperature = gas.temperature + heat_source.exterior_temperature / max(1, get_dist(src, get_turf(heat_source)))
 			if(abs(gas.temperature - initial_temperature) >= 100)
 				break
+	gas.update_values()
 	return gas
 
 /turf/proc/c_copy_air()
--- a/code/modules/weather/weather_effects.dm
+++ b/code/modules/weather/weather_effects.dm
@@ -21,7 +21,10 @@
 	return 0
 
 /obj/abstract/weather_system/proc/adjust_temperature(initial_temperature)
-	return initial_temperature
+	. = initial_temperature
+	var/decl/state/weather/current_weather = weather_system?.current_state
+	if(istype(current_weather))
+		. = current_weather.adjust_temperature(initial_temperature)
 
 /obj/abstract/weather_system/proc/show_weather(var/mob/M)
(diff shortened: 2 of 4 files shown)


code/modules/ZAS/Turf.dm
```diff
@@ -258,7 +258,9 @@
 	if(!include_heat_sources)
 		return gas
 
-	var/initial_temperature = weather ? weather.adjust_temperature(gas.temperature) : gas.temperature
+	if(weather)
+		gas.temperature = weather.adjust_temperature(gas.temperature)
```
MistakeNot4892, line 262: I think this'll need an `update_values` if you want pressure etc. to reflect the temp difference.

code/modules/weather/weather_fsm_states.dm
```diff
@@ -106,6 +122,9 @@
 		"Flakes of snow drift gently past."
 	)
 
+/decl/state/weather/snow/adjust_temperature(initial_temperature)
+	return min(initial_temperature - 20, T0C)
```
MistakeNot4892, line 126: This could be a pretty massive drop depending on the exterior temperature couldn't it? I'm imagining a cold front on a volcanic planet and the temperature goes from 150C to 0C.

MistakeNot4892 (changes requested):
Couple of small concerns.

code/modules/weather/weather_fsm_states.dm
out-of-phaze: Yeah, so volcanic planets shouldn't get snow at all.

out-of-phaze: Currently they only ever get ash.

MistakeNot4892 (dismissed):
Per Discord I think there's room for improvement in the temp clamping but we can revisit that down the track.

code/modules/weather/weather_effects.dm
```diff
@@ -21,7 +21,11 @@
 	return 0
 
 /obj/abstract/weather_system/proc/adjust_temperature(initial_temperature)
-	return initial_temperature
+	. = initial_temperature
+	var/decl/state/weather/current_weather = weather_system?.current_state
+	if(!istype(current_weather))
+		return initial_temperature
```
MistakeNot4892: Idle thought, but since you set `.` couldn't this just be
````
if(istype(current_weather))
    . = current_weather.adjust_temperature(.)
````

out-of-phaze: Same indentation depth, one line shorter, why not.

MistakeNot4892 (approved):

PR #3802, opened 2024-03-16
Make survival box choices per-map

out-of-phaze:
## Description of changes
Makes survival box choices per-map. A null list will auto-populate while an empty list will have no survival box options at all.
Makes `/datum/preferences/var/survival_box_choice` a decl reference rather than a string. It's (de)serialized using the UID.

Maps where emergency kits don't make sense can avoid having any, or limit it to just certain ones.

## Authorship
Me.


MistakeNot4892 (approved):
We need to do this for credsticks too...

PR #3805, opened 2024-03-17
Generalized bodytype uniform modifiers.

MistakeNot4892:
`uniform_state_modifier` on /under -> `onmob_state_modifiers` on /clothing.
Nothing uses these currently except """feminine""" alt uniforms, so impact should be minimal.


out-of-phaze (approved):
wow, that really *was* a tiny PR

PR #3806, opened 2024-03-17
Move accessory attachment behavior down to /obj/item/clothing.

MistakeNot4892:
## Description of changes
- Lazylists `accessories` on `/obj/item/clothing`.
- Moves accessory attachment and handling procs and vars from `/obj/item/clothing/accessory` to `/obj/item/clothing`.

Working towards being able to freely mix and match clothing with other clothing without having to awkwardly work around typepaths (see: all the accessory shirts and jackets).

## Authorship
Myself.

## Changelog
Nothing player-facing.

### DIFF
--- a/code/modules/clothing/clothing_accessories.dm
+++ b/code/modules/clothing/clothing_accessories.dm
@@ -1,11 +1,22 @@
-/obj/item/clothing/proc/can_attach_accessory(obj/item/clothing/accessory/A)
-	if(valid_accessory_slots && istype(A) && (A.slot in valid_accessory_slots))
-		.=1
+/obj/item/clothing
+	var/accessory_slot
+	var/accessory_removable = FALSE
+	/// if it should appear on examine without detailed view
+	var/accessory_high_visibility
+	/// used when an accessory is meant to slow the wearer down when attached to clothing
+	var/accessory_slowdown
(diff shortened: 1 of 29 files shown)


MistakeNot4892:
Pls review :(

code/modules/clothing/clothing_accessories.dm
```diff
@@ -1,11 +1,22 @@
-/obj/item/clothing/proc/can_attach_accessory(obj/item/clothing/accessory/A)
-	if(valid_accessory_slots && istype(A) && (A.slot in valid_accessory_slots))
-		.=1
+/obj/item/clothing
+	var/accessory_slot
+	var/accessory_removable = FALSE
+	/// if it should appear on examine without detailed view
+	var/accessory_high_visibility
+	/// used when an accessory is meant to slow the wearer down when attached to clothing
+	var/accessory_slowdown
```
comma, line 7: I wonder if it's worth having it as a separate variable now that accesory stuff is on the clothing

MistakeNot4892: This PR is mainly just to make it all available on clothing, I'd rather not start messing with slowdown and stuff at this point. There's like three PRs of changes on top of this one currently. :(

out-of-phaze (approved):

PR #3807, opened 2024-03-17
[MDB IGNORE] [IDB IGNORE] Condensing clothing state modifiers into a decl system.

MistakeNot4892:
## Description of changes
- All the various clothing states (open, buttons, sleeves, tucked in, hood, etc) have been unified in a decl system.
- Accessory hiding vars have been rolled into a check against the state decls.
- Clothing icon validation has been expanded to check for onmob states based on clothing state modifiers.

Depends on #3805 #3806 

Consistent clothing behavior, shorter typetree, more flexible clothing code.

## TODO
- [X] Coarse implementation/conversion.
- [x] Get suits and accessories passing CI.
- [x] Work out why verbs aren't being removed on accessory removal.
- [x] Verify icon updating and clothes icon updating is working on toggle.
- [x] See what happens to unit test time if we test states for all clothing.
- [x] Get all the other clothing passing CI.

## Authorship
Myself.

## Changelog


MistakeNot4892:
Somehow testing every single clothing subtype for icon states didn't appreciably impact testing time. Perplexing.

mods/species/bayliens/skrell/_skrell.dm
```diff
@@ -1,5 +1,4 @@
 #define SPECIES_SKRELL "Skrell"
-#define BODYTYPE_SKRELL "skrellian body"
```
comma, line 2: Why remove this?

MistakeNot4892: It was not used correctly. Didn't do anything, states were all broken as a result.

comma: But you're using the same value, just plain string now instead of define, what's different?

MistakeNot4892: It's not a `BODYTYPE_` value, using that would be confusing (which I guess is why all the icon states used it) - I can make it a new define if you want but it's only used in like four places so I'd probably just swap the uses to `species.name == SPECIES_SKRELL` or something.

MistakeNot4892: Went ahead and did that.

out-of-phaze (approved):

PR #3808, opened 2024-03-18
Various rock and crafting tweaks.

MistakeNot4892:
## Description of changes
- Rocks are now a material stack.
- Single natural walls no longer have no icon on Initialize().
- Mining requires a pickaxe with hardness greater than the target turf.
- Tool head crafting requires rocks or planks depending on the tool head.
- Stack crafting recipes or stacks themselves can have verbs specified for crafting messages.

Various stuff to make toolmaking and mining more entwined.

## Authorship
Myself.
## Changelog
:cl:
tweak: Mining now requires a tool of sufficient hardness.
tweak: Tool crafting is now restricted to specific stack types.
/:cl:


code/game/objects/items/rock.dm
```diff
@@ -1,28 +1,63 @@
...
-	edge = TRUE
-	force = 3
-	material_alteration = MAT_FLAG_ALTERATION_COLOR | MAT_FLAG_ALTERATION_NAME
-
-/obj/item/rock/Initialize()
+/obj/item/stack/material/rock
+	name                = "rock"
+	desc                = "The secret is to bang the rocks together, guys."
+	singular_name       = "rock"
+	plural_name         = "rock"
```
out-of-phaze: `You need at least 4 rock` is not really ideal. Set it to rocks or like, remove it so it autosets?
```suggestion
	plural_name         = "rocks"
```

code/modules/materials/_materials.dm
out-of-phaze, line 1: Line 300 should use `\the` instead of `the`, or drop the explicit article entirely.

code/modules/tools/archetypes/tool_extension_variable.dm
```diff
@@ -44,6 +44,6 @@
 	var/tool_name = tool_archetype.name
 	if(tool_archetype.article)
 		tool_name = "\a [tool_name]"
-	to_chat(user, SPAN_NOTICE("You adjust \the [holder] to function as [tool_name]."))
+	to_chat(user, SPAN_NOTICE("You adjust your grip on \the [holder] to use it as [tool_name]."))
```
out-of-phaze: I think the original idea here was something like changing a drill's bit from 'drill' to 'screwdriver' or what have you? Still, this message does work better for non-mechanical tools. I kind of think maybe the variable tool extension just shouldn't be used for those... could make it contextual and pop up a radial menu if it has two? Then again, that could make `IS_TOOL` blocking so you'd have to do it delicately...

out-of-phaze (changes requested):

code/modules/tools/archetypes/tool_extension_variable.dm
MistakeNot4892: Oh you're right, I thought this was separate to the powertools but it's all the same extension type.

MistakeNot4892: Per Discord this isn't something I can really do in this PR.

code/game/objects/items/weapons/surgery_tools.dm
```diff
@@ -179,7 +179,7 @@
 
 /obj/item/incision_manager/Initialize()
 	. = ..()
-	set_extension(src, /datum/extension/tool/variable, list(
+	set_extension(src, /datum/extension/tool/variable/simple, list(
```
out-of-phaze:
```suggestion
	set_extension(src, /datum/extension/tool/variable, list(
```
This is a digital tool, isn't it?

out-of-phaze (dismissed):

code/game/objects/items/weapons/surgery_tools.dm
MistakeNot4892: Yeah I must have find and replace'd that one.

out-of-phaze (dismissed):

out-of-phaze (approved):
Conflict resolution, so it's been 24hrs.

PR #3811, opened 2024-03-19
Added functionality for dumping the codex out as a website.

MistakeNot4892:
## Description of changes
- Adds a verb and handlers for writing the entire codex out as HTML at runtime.

Opening this so I don't forget it exists. Still needs a solution to the searchability problem.

Self-generating wiki.

## Authorship
Myself.

## Changelog
Nothing player-facing.

### DIFF
--- a/code/controllers/subsystems/initialization/codex_dump.dm
+++ b/code/controllers/subsystems/initialization/codex_dump.dm
@@ -1,0 +1,208 @@
+/datum/codex_entry/proc/get_dump_link_name()
+	return name
+
+/datum/codex_entry/codex/get_dump_link_name()
+	return "index" // This is the central page of the website.
+
+/decl/codex_page_converter
+	var/write_location = "codex/"
+	var/file_prefix
+	var/file_suffix = ".html"
+	var/style_loc = "common.css"
+	var/sidebar
+	var/static/list/illegal_filename_characters = list(
+		"\\",
+		"/",
+		":",
+		"*",
+		"?",
+		"\"",
+		"<",
+		">",
+		"|",
+		".",
+		"\n",
+		"\[",
+		"]"
+	)
+
+/decl/codex_page_converter/proc/handle_additional_dump()
+	var/full_style_loc = "[write_location][style_loc]"
+	if(fexists(full_style_loc))
+		fdel(full_style_loc)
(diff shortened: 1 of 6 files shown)


MistakeNot4892:
Search isn't implemented but the rest is working. Calling this done for now.

code/modules/codex/entries/_codex_entry.dm
```diff
@@ -63,10 +63,10 @@
 	if(!name)
 		if(length(associated_strings))
 			name = associated_strings[1]
-		else
+		else if(store_codex_entry)
 			CRASH("Attempted to instantiate unnamed codex entry with no associated strings!")
 
-	if(store_codex_entry)
+	if(name && store_codex_entry)
```
out-of-phaze, line 76: I feel like this can be simplified or changed somehow, but it should work fine as-is

out-of-phaze (approved):
aside from searchability, this looks fine

code/controllers/subsystems/initialization/codex_dump.dm
```diff
@@ -0,0 +1,204 @@
...
+	var/list/address_to_entry = list()
+	var/list/entry_to_address = list()
+	//var/list/seen_name_count =  list()
+	for(var/datum/codex_entry/codex_entry as anything in all_entries)
+		var/address = convert.convert_filename(codex_entry.get_dump_link_name())
+		address_to_entry[address] = codex_entry
+		entry_to_address[codex_entry] = address
+		address_to_body[address] = convert.convert_body(codex_entry.name, codex_entry.get_codex_body(include_header = FALSE, include_footer = FALSE))
+
+	// Copied from del-the-world UT exceptions list.
```
out-of-phaze: not quite. you do all subtypes of `/atom/movable`, while dtw only instantiates `typesof(/obj/item, /obj/effect, /obj/structure, /obj/machinery, /obj/vehicle, /mob) - ignore`. for example, `/obj/random` is excluded implicitly from dtw (maybe uh, bit of an oversight there actually?). excluding it here would fix the runtime, i've got no clue why missiles are failing GC though

out-of-phaze (dismissed):

out-of-phaze (approved):

PR #3812, opened 2024-03-19
Fixing issues with mass testing of /atom/movable.

MistakeNot4892:
## Description of changes
Fixes numerous issues with atoms being manually spawned in nullspace.

Used this code for testing:
````
/client/verb/test_atom_init()

    set name = "Test Atom Init"
    set category = "Debug"
    set src = usr

    for(var/atom_type in typesof(/atom/movable))
        try
            var/atom/movable/atom = atom_type
            if(TYPE_IS_SPAWNABLE(atom))
                atom_info_repository.get_instance_of(atom)
        catch(var/exception/E)
            to_chat(usr, "Failed: [atom_type] - '[E]'")
````

Allows for some atom subtype checking in the codex dump PR. Also generally good for code quality.

## Authorship
Myself.

## Changelog
Nothing player-facing.

### CODE REVIEW

File: code/modules/mob/living/simple_animal/hostile/giant_spider.dm [hunk 1]
```diff
@@ -312,6 +316,8 @@ Nurse caste procs
 
 /datum/ai/giant_spider/nurse/do_process(time_elapsed)
 	. = ..()
+	if(!body.loc)
```
comma: Maybe these checks should go into whatever cycle calls do_process? Seems like none of processing should be happening in nullspace

code/modules/mob/living/simple_animal/hostile/giant_spider.dm
```diff
@@ -180,6 +180,8 @@
 	expected_type = /mob/living/simple_animal/hostile/giant_spider
 
 /datum/ai/giant_spider/do_process()
+	if(!body.loc)
```
comma: Missed one spot when reverting

code/datums/ai/ai.dm
```diff
@@ -20,7 +20,7 @@
 	. = ..()
 
 /datum/ai/proc/can_process()
-	if(!body || ((body.client || body.mind) && !(body.status_flags & ENABLE_AI)))
+	if(!body || !body.loc || ((body.client || body.mind) && !(body.status_flags & ENABLE_AI)))
```
out-of-phaze, line 23: Is this really sufficient? What issue was this for? If `body` is in something that is in nullspace, will it still have issues? Do we need to check `!get_turf(body)`?

code/game/machinery/Beacon.dm
```diff
@@ -36,11 +36,9 @@
 		icon_state = "[state]"
 
 /obj/machinery/tracking_beacon/Process()
-	if(!beacon)
+	if(!beacon && loc)
```
out-of-phaze: `get_turf(src)` could still be null on the next line.

code/game/machinery/computer/area_atmos.dm
```diff
@@ -140,11 +140,11 @@
 	wow this is stupid, someone help me
 	*/
 	var/turf/T_src = get_turf(src)
-	if(!T_src.loc) return 0
+	if(!T_src?.loc) return 0
 	var/area/A_src = T_src.loc
```
out-of-phaze: Could just use this instead, since T_src is not used other than to get the area. Same with T_scrub.
```
	var/area/A_src = get_area(src)
	if(!A.src) return FALSE
```

code/game/machinery/computer/area_atmos.dm
```diff
@@ -158,7 +158,7 @@
 	var/found = 0
 
 	var/turf/T = get_turf(src)
-	if(!T.loc) return
+	if(!T?.loc) return
 	var/area/A = T.loc
```
out-of-phaze: Once again, this could just use `get_area(src)`.

code/game/machinery/computer/area_atmos.dm
```diff
@@ -158,7 +158,7 @@
...
 	for(var/obj/machinery/portable_atmospherics/powered/scrubber/huge/scrubber in SSmachines.machinery)
 		var/turf/T2 = get_turf(scrubber)
```
out-of-phaze, line 146: This could also use `get_area(src)`!

code/game/objects/random/date_based.dm
```diff
@@ -13,7 +13,7 @@
 	return ..()
 
 /obj/random/date_based/proc/check_date()
-	if(date_check.IsValid())
+	if(date_check?.IsValid())
```
out-of-phaze, line 17: I would rather we just set `abstract_type = /obj/random/date_based`, personally?

code/modules/holodeck/HolodeckControl.dm
```diff
@@ -3,13 +3,12 @@
 	desc = "A computer used to control a nearby holodeck."
 	icon_keyboard = "tech_key"
 	icon_screen = "holocontrol"
-	var/lock_access = list(access_bridge)
-	var/islocked = 0
-
+	is_spawnable_type = FALSE
```
out-of-phaze: This is constructable, see `/obj/item/stock_parts/circuitboard/holodeckcontrol`. I don't think `is_spawnable_type = FALSE` should be set for anything directly constructable, to be honest!

code/modules/power/smes.dm
```diff
@@ -18,6 +18,7 @@
 	stat_immune = 0
 	stat = BROKEN         // Should be removed if the terminals initialize fully.
 	reason_broken = MACHINE_BROKEN_GENERIC
+	is_spawnable_type = FALSE
```
out-of-phaze: Is there a reason this isn't just using `abstract_type = /obj/machinery/power/smes`? Setting it back to `TRUE` on every child type seems to indicate it should use `abstract_type` instead.

code/unit_tests/del_the_world.dm
```diff
@@ -19,35 +19,15 @@
...
-		// Exclude only this type, since it's not meant to be spawned but its subtypes are.
-		// TODO: Consider whether this warrants abstract_type?
-		/obj/machinery/power/smes,
-		// This is evil, evil code and destroys other tests.
-		/obj/effect/bhole,
+		/obj/structure/stairs
 	)
+
+	// Suspend to avoid fluid flows shoving stuff off the testing turf.
+	SSfluids.suspend()
```
out-of-phaze, line 29: I still worry about fluid_act() causing issues.

maps/random_ruins/exoplanet_ruins/monoliths/monoliths.dm
```diff
@@ -19,7 +19,12 @@
...
 /obj/structure/monolith/Initialize()
+
 	. = ..()
+
+	if(!loc)
+		return INITIALIZE_HINT_QDEL
+
 	icon_state = "jaggy[rand(1,4)]"
 
 	var/datum/planetoid_data/E = SSmapping.planetoid_data_by_z[z]
```
out-of-phaze: I would rather just gate this section rather than disallowing it from being spawned in nullspace
```suggestion
	var/datum/planetoid_data/E = loc ? SSmapping.planetoid_data_by_z[get_z(src)] : null
```

out-of-phaze (changes requested):

code/datums/ai/ai.dm
MistakeNot4892: Generally for wandering or targeting adjacent turfs, so checking for null loc does the job. It's just to stop runtimes in the testing period.

code/unit_tests/del_the_world.dm
MistakeNot4892: Not a lot we can do about it other than disabling the subsystem.

code/modules/power/smes.dm
MistakeNot4892: There was a reason but I can't remember it now.

code/unit_tests/del_the_world.dm
out-of-phaze: Blacklist the fluid spawner effects from the test, as before...?

MistakeNot4892: Won't stop atoms spawning reagents on the test turf though, so the issue is just masked, not resolved.

In a perfect universe the exceptions list wouldn't exist - I would prefer to see how it goes with SSfluids suspended at least initially. If it causes issues we can add it back to the exception list.

out-of-phaze (approved):
holodeck_control fixes seem good. do we wanna do a snake_case pass on filenames at some point, by the way? now `holodeck_control` is in `HolodeckControl.dm` :p

MistakeNot4892:
Maybe we can do a snake_case pass when we merge `game` and `modules`? It'd be nice but disruptive.

PR #3816, opened 2024-03-19
Converted HUD alert variables to a decl list.

MistakeNot4892:
## Description of changes
Converts several HUD alert vars to use an `alerts` lazylist instead.

Generalized code, cherrypicked work from my broader HUD rewrite to minimize the future changeset.

## Authorship
Myself.

## Changelog
Nothing player-facing.

### DIFF
--- a/code/modules/mob/living/carbon/human/life.dm
+++ b/code/modules/mob/living/carbon/human/life.dm
@@ -21,7 +21,6 @@
 #define COLD_GAS_DAMAGE_LEVEL_3 3 //Amount of damage applied when the current breath's temperature passes the 120K point
 
 /mob/living/carbon/human
-	var/toxins_alert = 0
 	var/co2_alert = 0
 	var/stamina = 100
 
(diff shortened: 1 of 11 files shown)


code/modules/mob/living/carbon/human/life.dm
```diff
@@ -548,13 +541,13 @@
 				cells.icon_state = "charge-empty"
 
 		if(pressure)
-			pressure.icon_state = "pressure[pressure_alert]"
+			pressure.icon_state = "pressure[GET_HUD_ALERT(src, /decl/hud_element/condition/pressure)]"
 		if(toxin)
-			toxin.icon_state = "tox[toxins_alert ? "1" : "0"]"
+			toxin.icon_state = "tox[GET_HUD_ALERT(src, /decl/hud_element/condition/toxins)]"
```
out-of-phaze, line 546: `"tox[null]"` and `"tox[0]"` are not equivalent; if the alert has never been set this might cause issues?

out-of-phaze (changes requested):

code/modules/mob/living/carbon/human/life.dm
MistakeNot4892:
````
#define GET_HUD_ALERT(M, A)        ((istype(M?.hud_used, /datum/hud) && (A in M.hud_used.alerts)) ? M.hud_used.alerts[A] : 0)
````

out-of-phaze (approved):

PR #3818, opened 2024-03-21
Adding metalwork.

MistakeNot4892:
## Description of changes
- Adds coverage checking to hot items burning the mob holding them.
- Adds crucibles for melting down ore in a kiln.
- Adds clay molds taken from items and used to make metal versions of those items.

Another step towards bootstrap crafting from rocks to spacetech.

## TODO
- [x] moulds need icons
- ~~crucible needs to announce the contents have melted~~
- [x] moulds and crucible need to handle reagents transfer
- [x] Add intermediary mold recipes (rods, blooms, ingots)
- [x] Find an anvil icon or remove the anvil definition from this PR.
- [x] Add metadata transfer between mold and item (casting a key and keeping the lock data)

## Authorship
Crucible icon is adapted from Doe's work if I recall. Code is mine.

## Changelog
:cl:
add: Added metalwork.
:/cl:


PsyCommando:
So, just a thing about using clay for casting. That is, if you wanna keep it as authentic as possible.

Usually people doing small scale smelting use green sand to make molds to cast iron, steel, copper, bronze, brass, aluminium and other low-ish melting point metals. It can be re-used, and it works great for making molds of existing things. http://backyardmetalcasting.com/greensand.html
(They also use things like steel pans and muffin trays to cast ingots if its a low melting point metal.)

Using straight up clay/pottery ceramic would essentially be extremely dangerous because the water inside the clay would turn to steam instantly and project hot molten metal everywhere.

MistakeNot4892:
This is functionally complete. There's several remaining things that I'm not sure how to approach and don't directly impact the ability to make metal stuff:
- It's still using clay for moulds rather than foundry sand as Psy describes.
- The crucible has no indicator that the contents have melted other than the visual overlay.

code/game/objects/structures/fires.dm
```diff
@@ -185,7 +185,8 @@
 			user.put_in_hands(removing)
 			if(lit == FIRE_LIT)
 				visible_message(SPAN_DANGER("\The [user] fishes \the [removing] out of \the [src]!"))
-				user.fire_act(return_air(), last_fuel_burn_temperature, 500)
+				// Uncomment this when there's a way to take stuff out of a kiln or oven without setting yourself on fire.
+				//user.fire_act(return_air(), last_fuel_burn_temperature, 500)
```
out-of-phaze, line 189: could honestly just try calling `removing.try_burn_wearer(user)`? but maybe that'll happen anyways. honestly i kind of think holding a hot object is punishing enough for removing one from a lit kiln as it is

out-of-phaze (dismissed):

code/game/objects/structures/fires.dm
MistakeNot4892: That's called anyway when the item is equipped.

out-of-phaze (approved):

PR #3819, opened 2024-03-21
Various material stack and recipe tweaks.

MistakeNot4892:
## Description of changes
- Stacks can be pulverized or converted into other forms via a generalized list returned to a proc on the stack instance.
- Stacks will now handle matter multiplier on self and stack products appropriately.
- Stacks will automatically calculate maximum result amount.
- Stack recipes can be hidden via a map var.
- Tiles now use material and matter mult instead of hardcoding result amount.
- Ore, logs and lump stacks have more matter in them.

Generally more dynamic crafting system. Allows for freely using stack types with different matter multipliers without causing discrepancies in cost.

## Authorship
Myself.

## Changelog
:cl:
tweak: More ore per ore!
/:cl:

### DIFF
--- a/code/modules/materials/material_sheets.dm
+++ b/code/modules/materials/material_sheets.dm
@@ -17,7 +17,6 @@
 	plural_name = "sheets"
 	abstract_type = /obj/item/stack/material
 	is_spawnable_type = FALSE // Mapped subtypes set this so they can be spawned from the verb.
-	var/list/can_be_converted_into
 	var/can_be_pulverized = FALSE
 	var/can_be_reinforced = FALSE
 	var/decl/material/reinf_material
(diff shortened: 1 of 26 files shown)


code/modules/materials/material_sheets.dm
```diff
@@ -128,17 +131,22 @@
...
+	// TODO: convert to converts_into entry.
+	if(can_be_pulverized && IS_HAMMER(W) && material?.hardness >= MAT_VALUE_RIGID && user.a_intent == I_HURT)
+
 		if(W.material?.hardness < material.hardness)
 			to_chat(user, SPAN_WARNING("\The [W] is not hard enough to pulverize [material.solid_name]."))
 			return TRUE
+
 		var/converting = clamp(get_amount(), 0, 5)
-		if(converting)
+		world << "[converting] [get_amount()]"
```
out-of-phaze:
```suggestion
```

out-of-phaze (dismissed):

out-of-phaze (approved):
Presumably at some point we're going to have to turn stack transformations/conversions into decls so that we can add additional logic other than tool archetype, e.g. hardness checks for pulverizing, etc. But for now, this works.

PR #3820, opened 2024-03-22
Various fixes to survival and crafting stuff.

MistakeNot4892:
## Description of changes
- Fixes material stacks not getting a color.
- Fixes runtime with stack spawns.
- Fixes ordering issue with trees dropping logs.
- Fixes some display issues with the meat hook.
- Fixes some issues with alpha masks.
- Bunch of other small fixes from playtesting the survival map.

Fixes.

## Authorship
Myself.

## Changelog
Nothing player-facing.


code/game/turfs/walls/wall_natural.dm
```diff
@@ -65,11 +66,16 @@
...
+	if(IS_PICK(W) && !being_mined)
 		if(W.material?.hardness < max(material?.hardness, reinf_material?.hardness))
 			to_chat(user, SPAN_WARNING("\The [W] is not hard enough to dig through \the [src]."))
-		else if(W.do_tool_interaction(TOOL_PICK, user, src, 2 SECONDS, suffix_message = destroy_artifacts(W, INFINITY)))
-			dismantle_wall()
+		else
+			if(being_mined)
+				return TRUE
+			being_mined = TRUE
+			if(W.do_tool_interaction(TOOL_PICK, user, src, 2 SECONDS, suffix_message = destroy_artifacts(W, INFINITY)))
```
out-of-phaze, line 76: I believe that some things use `set_cooldown = TRUE` instead of a bespoke debounce var, but that *does* mean you still have the cooldown even when you cancel it, and additional vars have a constant (not scaling with number of instances) cost anyway.

out-of-phaze (approved):

code/game/turfs/walls/wall_natural.dm
MistakeNot4892: I don't want to use cooldown because that would interfere with the traditional mining pattern of doing three turfs at once. We really need a better handler for stacking do_afters.

out-of-phaze: Worse yet, it would mean bump-mining would still have no cooldown. Good call.

PR #3824, opened 2024-03-24
Make wildgrass a subtype of grass

out-of-phaze:
## Description of changes
Makes wildgrass a subtype of grass.

Makes the forage map work for both grass and wild grass, plus general tidiness.


MistakeNot4892:
I swear I did this already in two other branches. Maybe it never got cherrypicked. :(

MistakeNot4892 (approved):

PR #3827, opened 2024-03-24
Fix fingerprint merging causing runtimes

out-of-phaze:
## Description of changes
Makes fingerprint merging actually handle list-typed data.

Fixes fingerprint merging causing runtimes.

### DIFF
--- a/code/modules/detectivework/evidence/fingerprints.dm
+++ b/code/modules/detectivework/evidence/fingerprints.dm
@@ -11,12 +11,13 @@
 	if(F.completeness > 0)
 		add_data(F)
 
-/datum/forensics/fingerprints/add_data(var/datum/fingerprint/newprint)
-	if(!newprint || newprint.completeness <= 0)
-		return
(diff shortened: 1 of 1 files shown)


MistakeNot4892 (approved):

code/modules/detectivework/evidence/fingerprints.dm
```diff
@@ -11,12 +11,13 @@
 	if(F.completeness > 0)
 		add_data(F)
 
-/datum/forensics/fingerprints/add_data(var/datum/fingerprint/newprint)
-	if(!newprint || newprint.completeness <= 0)
-		return
-	for(var/datum/fingerprint/F in data)
-		if(F.merge(newprint))
-			return
+/datum/forensics/fingerprints/add_data(var/list/newdata)
```
MistakeNot4892, line 14: Isn't this now going to runtime if passed a single datum?

out-of-phaze: It is never passed a single datum

PR #3832, opened 2024-03-24
Fix some sprite accessory issues, misc. preferences fixes

out-of-phaze:
## Description of changes
Heritable sprite accessories on DNA now properly track color.
Changing the background in character setup does not regenerate the entire mannequin.
Sprite accessory availability caching now takes into account bodytype.
Bodytype default skin and eye color is now used in pref sanitization.
`reset_hair` and a few other bodytype-based accessory things now properly work. Before, the lack of `accessory_color` would cause it to fail to add the accessory.

DNA will no longer entirely clear sprite accessories (including hair!).
Changing the background in character setup shouldn't cause as much server load.
Bodytype-restricted accessories will now work properly.
Default bodytype accessories and skin and eye color now properly work in more situations.

### DIFF
--- a/code/modules/client/preferences.dm
+++ b/code/modules/client/preferences.dm
@@ -403,7 +403,7 @@ var/global/list/time_prefs_fixed = list()
 	character.backpack_setup = new(backpack, backpack_metadata["[backpack]"])
 
 	for(var/obj/item/organ/external/O in character.get_external_organs())
-		O.clear_sprite_accessories(skip_update = TRUE)
+		O.clear_sprite_accessories_by_category(SAC_MARKINGS, skip_update = TRUE)
 
 	for(var/accessory_category in sprite_accessories)
 		for(var/accessory in sprite_accessories[accessory_category])
--- a/code/modules/client/preferences.dm
+++ b/code/modules/client/preferences.dm
@@ -403,7 +403,10 @@ var/global/list/time_prefs_fixed = list()
 	character.backpack_setup = new(backpack, backpack_metadata["[backpack]"])
 
 	for(var/obj/item/organ/external/O in character.get_external_organs())
-		O.clear_sprite_accessories_by_category(SAC_MARKINGS, skip_update = TRUE)
+		for(var/decl/sprite_accessory_category/sprite_category in O.get_sprite_accessory_categories())
+			if(!sprite_category.clear_in_pref_apply)
+				continue
+			O.clear_sprite_accessories_by_category(sprite_category.type, skip_update = TRUE)
 
 	for(var/accessory_category in sprite_accessories)
 		for(var/accessory in sprite_accessories[accessory_category])
--- a/code/modules/mob/living/carbon/human/human_appearance.dm
+++ b/code/modules/mob/living/carbon/human/human_appearance.dm
@@ -63,21 +63,24 @@
 	var/decl/bodytype/root_bodytype = get_bodytype()
 	var/list/valid_hairstyles = species?.get_available_accessories(root_bodytype, SAC_HAIR)
 	if(length(valid_hairstyles))
-		SET_HAIR_STYLE(src, pick(valid_hairstyles), TRUE)
+		var/body_default_hairstyles = LAZYACCESS(root_bodytype?.default_sprite_accessories, SAC_HAIR)
+		// Forgive us, but we need to set both hair color AND style at the same time...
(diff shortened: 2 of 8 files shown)


code/modules/client/preferences.dm
```diff
@@ -403,7 +403,7 @@ var/global/list/time_prefs_fixed = list()
 	character.backpack_setup = new(backpack, backpack_metadata["[backpack]"])
 
 	for(var/obj/item/organ/external/O in character.get_external_organs())
-		O.clear_sprite_accessories(skip_update = TRUE)
+		O.clear_sprite_accessories_by_category(SAC_MARKINGS, skip_update = TRUE)
```
MistakeNot4892: Not sure if I like this. I almost feel like you should put a flag on the accessory categories and iterate the list to clear appropriately.

code/modules/mob/living/carbon/human/human_appearance.dm
```diff
@@ -63,21 +63,24 @@
...
 		//this shouldn't happen
 		var/new_hair = LAZYACCESS(root_bodytype?.default_sprite_accessories, SAC_HAIR) || /decl/sprite_accessory/hair/bald
 		if(new_hair)
-			SET_HAIR_STYLE(src, new_hair, TRUE)
+			set_organ_sprite_accessory_by_category(new_hair[1], SAC_HAIR, new_hair[new_hair[1]], TRUE, TRUE, BP_HEAD, TRUE)
 
 	var/list/valid_facial_hairstyles = species?.get_available_accessories(root_bodytype, SAC_FACIAL_HAIR)
 	if(length(valid_facial_hairstyles))
-		SET_FACIAL_HAIR_STYLE(src, pick(valid_facial_hairstyles), TRUE)
+		var/body_default_facial_hairstyles = LAZYACCESS(root_bodytype?.default_sprite_accessories, SAC_FACIAL_HAIR)
```
MistakeNot4892, line 79: This needs a `length()` check on the following lines I think, for cases where the mob has no hair or facial hair available, like unathi.

code/modules/species/species.dm
```diff
@@ -689,13 +689,14 @@ var/global/const/DEFAULT_SPECIES_HEALTH = 200
 	for(var/accessory in get_available_accessory_types(bodytype, accessory_category))
 		. += GET_DECL(accessory)
 
-/decl/species/proc/get_available_accessory_types(bodytype, accessory_category)
+/decl/species/proc/get_available_accessory_types(decl/bodytype/bodytype, accessory_category)
 	if(!bodytype)
 		bodytype = default_bodytype
-	var/list/available_accessories = LAZYACCESS(accessory_styles, accessory_category)
+	var/list/available_accessories = accessory_styles?[bodytype.type]?[accessory_category]
```
MistakeNot4892, line 695: This feels super brittle. :(

out-of-phaze: howso? bodytype should never be null here, and since we're not indexing numerically, `?[x]` is equivalent to `LAZYACCESS` here. it's just an easier way to do chained lazyaccess

code/modules/mob/living/carbon/human/human_appearance.dm
out-of-phaze: didn't cause issues downstream for a species without facial hair. where should it check? there's already a length check on line 76, and if body_default_facial_hairstyles has nothing then it'll just pass null as color (so maybe i should add `|| COLOR_BLACK` or something, idk)

code/modules/client/preferences.dm
out-of-phaze: i don't understand what the issue here is—that we only clear markings? I just wanted to keep parity with the non-broken prior behavior, but i could just put a `reset_hair()` after this at the cost of some wasted work the first time it's applied.

MistakeNot4892: More that under the current system, markings might not be the only category of sprite accessory that you want to clear here. I would personally just put a `clear_in_pref_apply` flag or something to the sprite accessory category and only set it on markings for the moment, which would be the same behavior that you want but future proofed.

code/modules/mob/living/carbon/human/human_appearance.dm
MistakeNot4892: Oh I am completely blind yeah it's already checked.

MistakeNot4892 (approved):

PR #3841, opened 2024-03-30
Updating Neb mini from Scav.

MistakeNot4892:
- Tons of map changes.
- Removal of the default x4, y4 overmap position for the overmap site.
- Re-themes the map from generic corporate to tradehouse, a la the tradeship.
- Retypes pilot voidsuits to expedition voidsuits for symmetry.

Nobody except Scav uses this map generally and I'm trying to make it easier to do a dev update downstream, so I'm not too fussed about any clobbering.


comma (dismissed):

out-of-phaze (dismissed):

out-of-phaze (dismissed):

code/game/objects/objs.dm
```diff
@@ -21,7 +21,7 @@
 	var/armor_penetration = 0
 	var/anchor_fall = FALSE
 	var/holographic = 0 //if the obj is a holographic object spawned by the holodeck
-	var/tmp/directional_offset ///JSON list of directions to x,y offsets to be applied to the object depending on its direction EX: @'{"NORTH":{"x":12,"y":5}, "EAST":{"x":10,"y":50}}'
+	var/list/directional_offset ///JSON list of directions to x,y offsets to be applied to the object depending on its direction EX: @'{"NORTH":{"x":12,"y":5}, "EAST":{"x":10,"y":50}}'
```
MistakeNot4892, line 24: This was needed in order to be able to null the list so switches/buttons mapped onto tables didn't offset.

out-of-phaze: Can't you do `var/tmp/list`?

MistakeNot4892: `tmp` hides it from the mapper.

out-of-phaze: Oh. It doesn't in StrongDMM, so I never noticed. Yeah, fair

out-of-phaze (approved):

PR #3850, opened 2024-04-02
Expands agriculture.

MistakeNot4892:
## Description of changes
- Adds a tool archetype for hoes.
- Makes farm plots dug into dirt turfs instead of crafted from bricks.

Depends on #3848.

## TODO
- [X] Make plots dug rather than built.
- [X] Add plot destruction.
- [x] Make rain and turf reagents hydrate/feed trays.
- [x] Make adjacent fluids feed/hydrate soil.
- [x] Make hoes a tool type.
- ~~Add compost bin that turns nutriment/meat/plant matter into compost/mulch that can feed hydrotrays.~~
- [x] Add walking/running on plots damaging the plants.
- [x] Add ability to reinforce a dug plot with bricks to prevent people walking on the plots.

## Why and what will this PR improve
More involved/interesting turf-based agriculture.

## Authorship
Myself.

## Changelog
:cl:
tweak: Farm plots are now dug rather than crafted.
tweak: Various agriculture changes! Check the PR for more details.
/:cl:

PR #3790, opened 2024-03-12
Adds fishing

MistakeNot4892:
## Description of changes
- Ports several fish icons from Polaris and adds subtypes to use them.
- Adds craftable fishing rods and fishing line.
- Adds infrastructure for fishing in submerged turfs.

## Why and what will this PR improve
Fishing good.

## TODO
- [x] Add some fish or other item rewards for fishing to exoplanets as an example for main code.
- [x] Add sound effects and visuals to fishing.

## Authorship
Code is mine, icons are adapted from Polaris.

## Changelog
:cl:
add: Added fishing!
/:cl:

PR #3851, opened 2024-04-02
Adds compost bins.

MistakeNot4892:
## Description of changes
- Adds craftable compost bins.
- Compost bins use worms to turn compostable materials into compost, which can be used to feed plants.

Provides a source of fertilizer for low-tech situations.

Depends on #3790 and #3850.

## Authorship
Myself.

## Changelog
:cl:
add: Added craftable compost bins.
/:cl:


out-of-phaze (approved):
as mentioned on Discord, I think `inline_counting_english_list` might be equivalent to the `summarize` arg you added to `english_list()`, but i'm not willing to block over it. ~~if it works it works, though i do think it might be nicer to use the existing helper with `DET_INDEFINITE` so you get "there is a banana, 2 oranges, and a pear in the composter" or whatever.~~ in fact, that proc currently says `2x oranges` instead of `2 oranges` so it's kinda ugly. the summarize arg is fine

PR #3859, opened 2024-04-04
Fixes planet colors for skybox image generation

comma:
## Description of changes

Overmap object gets these variables set by template data so they were always overwritten by default values. Moves them in the right place now.

Meat planets now look meat colored from space again

## Authorship

Me

## Changelog
:cl:
bugfix: Fixed planets having wrong colors when seen on skybox
/:cl:


MistakeNot4892 (approved):
muh indenting......

out-of-phaze (approved):

PR #3866, opened 2024-04-05
Allow recipes to set their skill and difficulty from material

out-of-phaze:
## Description of changes
Recipes with a null difficulty now calculate it from material properties.
Materials' construction difficulty var is now actually used.
Materials now have a construction skill var.
Recipe time is now calculated dynamically based on material if time is unset, rather than set once in init.
Does some (slight, experimental) optimizations to the inconsistent recipe materials test.

Makes it less fiddly to override recipe skills on other maps, and lets materials inform what skill they should use. This is a step towards breaking up the construction skill monolith.

### DIFF
--- a/code/datums/repositories/decls.dm
+++ b/code/datums/repositories/decls.dm
@@ -24,7 +24,9 @@ var/global/repository/decls/decls_repository = new
 	var/list/fetched_decls =                 list()
 	var/list/fetched_decl_ids =              list()
 	var/list/fetched_decl_types =            list()
+	var/list/fetched_decl_instances =        list()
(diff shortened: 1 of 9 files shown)


code/datums/repositories/decls.dm
```diff
@@ -114,6 +116,18 @@ var/global/repository/decls/decls_repository = new
 		if(decl)
 			. += decl
 
+/repository/decls/proc/get_decls_of_type_unassociated(var/decl_prototype)
```
MistakeNot4892, line 119: We can probably swap these procs into a dozen places.

MistakeNot4892 (approved):

PR #3871, opened 2024-04-08
Adds a shim for 515 :: syntax.

MistakeNot4892:


MistakeNot4892:
Selfmerging this because I can't compile without it.

PR #3875, opened 2024-04-09
Fixes various turf issues on dev.

MistakeNot4892:
- Fixes stacked AO overlays on turf changing.
- Fixes turfs never properly resetting to plating base icons.
- Fixes natural turfs eating tiles without changing their flooring.

### DIFF
--- a/code/game/turfs/floors/natural/_natural.dm
+++ b/code/game/turfs/floors/natural/_natural.dm
@@ -37,24 +37,35 @@
 		color = base_color
 	else
 		color = null
-	if(material)
-		set_turf_materials(material, skip_update = no_update_icon)
 
 	if(possible_states > 0)
 		icon_state = "[rand(0, possible_states)]"
+
+	// TEMP: set these so putting tiles over natural turfs doesn't make green sand.
+	base_name       = name
+	base_desc       = desc
+	base_icon       = icon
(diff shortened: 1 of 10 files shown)


code/game/turfs/floors/natural/_natural.dm
```diff
@@ -37,24 +37,35 @@
 		color = base_color
 	else
 		color = null
-	if(material)
-		set_turf_materials(material, skip_update = no_update_icon)
 
 	if(possible_states > 0)
 		icon_state = "[rand(0, possible_states)]"
+
+	// TEMP: set these so putting tiles over natural turfs doesn't make green sand.
```
out-of-phaze, line 44: what's this about? how is this going to be replaced/superseded later? not a big fan of temp code without a TODO explaining what we're keeping it around until

MistakeNot4892: It'll be replaced when natural turf stuff is moved onto flooring in my next PR, I can open the PR later today if you want a reference for it.

MistakeNot4892: https://github.com/NebulaSS13/Nebula/pull/3879

out-of-phaze (approved):

PR #3773, opened 2024-03-08
Reworking attack() -> use_on_mob()

MistakeNot4892:
## Description of changes
- Renamed `attack()` to `use_on_mob()` and removed the zone parameter.
- Standardize parameters to `mob/living/target` and `mob/living/user`.
- Add return values and generally clean up the procs.

## Why and what will this PR improve
Standardized return value, standard parameters, less carbon types.

## Authorship
Myself.

## Changelog
Nothing player-facing.

PR #3876, opened 2024-04-09
Removes /mob/living/carbon pending a full human retype.

MistakeNot4892:
## Description of changes
- Moves all vars and logic from /mob/living/carbon to `/mob/living` or `/mob/living/carbon/human`.
- Retains `/mob/living/carbon` only as a component of `/mob/living/carbon/human` as the human repath is like 3000 lines and would be even more unreviewable.

Depends on #3773.

/carbon delenda est.

## Authorship
Myself.

## Changelog
Nothing player-facing.


code/game/atoms_movable.dm
```diff
@@ -364,14 +364,12 @@
 		return TRUE
 
 /atom/movable/proc/buckle_mob(mob/living/M)
+
 	if(buckled_mob) //unless buckled_mob becomes a list this can cause problems
 		return FALSE
+
 	if(!istype(M) || (M.loc != loc) || M.buckled || LAZYLEN(M.pinned) || (buckle_require_restraints && !M.restrained()))
 		return FALSE
-	if(ismob(src))
```
MistakeNot4892, line 378: This is redundant in view of can_buckle_mob().

code/modules/mob/living/simple_animal/simple_animal.dm
```diff
@@ -501,17 +500,6 @@ var/global/list/simplemob_icon_bitflag_cache = list()
 /mob/living/simple_animal/get_digestion_product()
 	return /decl/material/liquid/nutriment
 
-/mob/living/simple_animal/eyecheck()
```
MistakeNot4892, line 487: Might need to split the flash rework out into a separate PR.

MistakeNot4892: Not going to bother.

MistakeNot4892:
Current issues:
- Humans are spawning with no blood, seemingly due to should_have_organ() returning false for BP_HEART during initialization.
- Humans are claiming not to take damage during CI, which is strange as they do take damage post-spawn.

MistakeNot4892:
Tested:
- General spawning
- Latejoin/roundstart spawning
- Taking damage via proc and falling off stuff
- Dismemberment
- Gibbing
- Breathing and pressure

I am almost certain this is going to break spectacularly, but I can't see any immediate obvious issues locally, so might as well shake them loose in dev. I don't think there are any conceptual problems with the PR at least.

code/__defines/mobs.dm
```diff
@@ -224,14 +224,14 @@
 #define MOB_PULL_SAME 2
 #define MOB_PULL_LARGER 3
 
-//carbon taste sensitivity defines, used in mob/living/carbon/proc/ingest
+//carbon taste sensitivity defines, used in mob/living/proc/ingest
```
out-of-phaze: nitpick, though i think this is the only reference to carbons in a comment on the right-hand side of the diff i saw
```suggestion
// taste sensitivity defines, used in mob/living/proc/ingest
```

code/game/machinery/oxygen_pump.dm
```diff
@@ -67,17 +67,17 @@
 	ui_interact(user)
 	return TRUE
 
-/obj/machinery/oxygen_pump/proc/attach_mask(var/mob/living/carbon/C)
-	if(C && istype(C))
+/obj/machinery/oxygen_pump/proc/attach_mask(var/mob/living/C)
+	if(ishuman(C))
```
out-of-phaze: kinda wondering why this has to be a humantype and not a living-type especially since the argument is the latter. `get_internals()` and `get_breath_from_internal()` and etc are all on living now

code/game/machinery/oxygen_pump.dm
```diff
@@ -67,17 +67,17 @@
...
 			tank.forceMove(C)
 		breather = C
 
-/obj/machinery/oxygen_pump/proc/set_internals(var/mob/living/carbon/C)
-	if(C && istype(C))
-		if(!C.internal && tank)
+/obj/machinery/oxygen_pump/proc/set_internals(var/mob/living/C)
+	if(ishuman(C))
+		if(!C.get_internals() && tank)
 			breather.set_internals(tank)
```
out-of-phaze, line 81: we pass `C` but use `breather` instead. should either be argless and just replace mentions of `C` with `breather`, or... probably just that, really

code/modules/admin/view_variables/topic.dm
```diff
@@ -445,11 +445,11 @@
 		possibleverbs += typesof(/mob/proc,/mob/verb,/mob/living/proc,/mob/living/verb)
 		switch(H.type)
 			if(/mob/living/carbon/human)
-				possibleverbs += typesof(/mob/living/carbon/proc,/mob/living/carbon/verb,/mob/living/carbon/human/verb,/mob/living/carbon/human/proc)
+				possibleverbs += typesof(/mob/living/carbon/human/verb, /mob/living/carbon/human/proc)
```
out-of-phaze, line 448: this isn't a request for changes i just want to note that this code is horrifying. nothing can be done about it either, it is probably the best way this code can be written without massively overengineering it. yikes

code/modules/admin/view_variables/topic.dm
```diff
@@ -482,9 +482,9 @@
 	else if(href_list["addorgan"])
 		if(!check_rights(R_SPAWN))	return
 
-		var/mob/living/carbon/M = locate(href_list["addorgan"])
+		var/mob/living/carbon/human/M = locate(href_list["addorgan"])
 		if(!istype(M))
```
out-of-phaze, line 486: it may be better to replace this with something like `should_have_organs()` or maybe just a `length(get_organs())` check since mobs with organs delete themselves upon losing the last one. we do still want to eventually support organs on non-humantypes, right?

code/modules/clothing/head/headphones.dm
```diff
@@ -64,10 +64,9 @@
 	stop_music(user)
 
 /obj/item/clothing/head/headphones/proc/play_music(mob/user)
-	var/static/list/allowed_slots = list(slot_l_ear_str, slot_r_ear_str, slot_head_str)
 	if(!user || !user.client)
 		return
-	if(!(user.get_equipped_slot_for_item(src) in allowed_slots))
+	if(!(user.get_equipped_slot_for_item(src) in global.headphone_slots))
```
out-of-phaze, line 69: ~~curious why this was done. not a request for changes, just a request for info~~ nevermind, `get_sound_volume_multiplier()` i presume

code/modules/mob/living/human/human.dm
```diff
@@ -37,9 +50,26 @@
...
 	QDEL_NULL(attack_selector)
 	QDEL_NULL(vessel)
-	. = ..()
+	QDEL_NULL(touching)
+
+	if(reagents == bloodstr)
+		reagents = null
+	else
+		QDEL_NULL(bloodstr)
+	QDEL_NULL(bloodstr)
```
out-of-phaze: this doesn't feel right. `reagents` should be deleted/nulled in `/atom/Destroy`. this probably results in hanging refs to carbons on current dev. i think maybe it should just be
```suggestion
	if(reagents == bloodstr)
		bloodstr = null
	else
		QDEL_NULL(bloodstr)
```

code/modules/mob/living/human/human_damage.dm
```diff
@@ -424,23 +430,21 @@ This function restores all organs.
...
 
-	var/total_damage = 0
+	shock_damage /= length(to_shock)
+	shock_damage = round(shock_damage, 0.1)
 
 	for(var/obj/item/organ/external/E in to_shock)
-		total_damage += ..(shock_damage, E.organ_tag, base_siemens_coeff * get_siemens_coefficient_organ(E))
+		var/actual_shock_damage = max(1, round(shock_damage * base_siemens_coeff * get_siemens_coefficient_organ(E)))
+		if(actual_shock_damage)
+			apply_damage(shock_damage, BURN, E.organ_tag, used_weapon="Electrocution")
```
out-of-phaze: this change looks a bit odd compared to the prior code (i believe that, previously, it'd propagate the shock to other organs via `trace_shock()` in the recursive call?) but honestly, the prior code looks weird too, so assuming electrocution was tested it's probably fine. not sure why `trace_shock()` would even need to run more than once anyway, you have one entry point that's being shocked per call. tl;dr did you test this? if so, cool

code/modules/mob/living/human/human_eating.dm
```diff
@@ -0,0 +1,10 @@
+/mob/living/carbon/human/ingest(var/datum/reagents/from, var/datum/reagents/target, var/amount = 1, var/multiplier = 1, var/copy = 0) //we kind of 'sneak' a proc in here for ingesting stuff so we can play with it.
```
out-of-phaze: honestly kind of wish that simplemobs could taste reagents, would be nice but i understand if it's too complex to move it down to `/mob/living/proc/ingest()` rather than up to `/mob/living/carbon/human/ingest()`.

code/modules/mob/transform_procs.dm
```diff
@@ -49,7 +49,7 @@
 	QDEL_NULL_LIST(worn_underwear)
 	return ..(move)
 
-/mob/living/carbon/AIize()
+/mob/living/AIize()
```
out-of-phaze, line 53: should probably do an early return for `/mob/living/silicon/ai/AIize()`? or maybe in that case it's actually desirable to create a new AI mob. if so, ignore

code/modules/organs/prosthetics/prosthetics_manufacturer.dm
```diff
@@ -10,6 +10,7 @@
 	eye_flash_mod = 1
 	eye_darksight_range = 2
 	associated_gender = PLURAL
+	retrievable_via_pronoun = FALSE
```
out-of-phaze: is there a reason not to just set `associated_gender = null` instead? i guess this would make it so that selecting the FBP bodytype sets your pronouns to they/them but i don't know if that's actually desirable

code/modules/reagents/Chemistry-Taste.dm
```diff
@@ -1,17 +1,3 @@
...
-/mob/living/carbon/ingest(var/datum/reagents/from, var/datum/reagents/target, var/amount = 1, var/multiplier = 1, var/copy = 0) //we kind of 'sneak' a proc in here for ingesting stuff so we can play with it.
-	if(last_taste_time + 50 < world.time)
-		var/datum/reagents/temp = new(amount, global.temp_reagents_holder) //temporary holder used to analyse what gets transfered.
-		from.trans_to_holder(temp, amount, multiplier, 1)
-		var/text_output = temp.generate_taste_message(src, from)
-		if(text_output && (text_output != last_taste_text || last_taste_time + 1 MINUTE < world.time)) //We dont want to spam the same message over and over again at the person. Give it a bit of a buffer.
-			to_chat(src, SPAN_NOTICE("You can taste [text_output].")) //no taste means there are too many tastes and not enough flavor.
-			last_taste_time = world.time
-			last_taste_text = text_output
-	. = ..()
```
out-of-phaze: ewwww leading whitespace at start of file. nitpick though, feel free to ignore

code/modules/reagents/reagent_containers/syringes.dm
```diff
@@ -218,15 +218,15 @@
 		mode = SYRINGE_DRAW
 		update_icon()
 
-/obj/item/chems/syringe/proc/handleBodyBag(var/obj/structure/closet/body_bag/bag, var/mob/living/carbon/user)
+/obj/item/chems/syringe/proc/handleBodyBag(var/obj/structure/closet/body_bag/bag, var/mob/living/user)
```
out-of-phaze, line 221: noting that we should probably `snake_case` these procs at some point, along with dbcore typepaths and such. not a request for changes

out-of-phaze (changes requested):
not looking forward to seeing `var/mob/living/C` all over just like `var/obj/item/W` but maybe i'll make a PR to clean those up eventually :p

code/modules/organs/prosthetics/prosthetics_manufacturer.dm
MistakeNot4892: That was my original solution but I thought about it and I don't think it's conceptually sound. Having a separate flag feels more future proofed to me.

code/modules/mob/living/living_taste.dm
```diff
@@ -0,0 +1,17 @@
+/mob/living
+	/// these two help govern taste. The first is the last time a taste message was shown to the plaer.
```
out-of-phaze:
```suggestion
	/// these two help govern taste. The first is the last time a taste message was shown to the player.
```

code/game/machinery/oxygen_pump.dm
```diff
@@ -10,7 +10,7 @@
 	anchored = TRUE
 
 	var/obj/item/tank/tank
-	var/mob/living/carbon/breather
+	var/mob/living/carbon/human/breather
```
out-of-phaze: `attach_mask` now takes a living mob but `breather` is still humantyped

out-of-phaze (dismissed):
the prior unresolved comments still stand, i just found some additional things. sorryyyy

code/modules/organs/prosthetics/prosthetics_manufacturer.dm
out-of-phaze: i'm not sure about selecting the FBP bodytype in character setup changing your character's pronouns to plural though... and if we gated that behind `retrievable_via_pronoun` as well it'd mean `associated_gender` is entirely unused for prostheses, i.e. it'd be identical to just setting `associated_gender = null`.

MistakeNot4892: I don't feel strongly enough about it to argue it, I've implemented that change.

code/modules/mob/living/human/human_damage.dm
MistakeNot4892: Testing shortly just to confirm, but this isn't really a functional change, there is no recursive call. I really just moved the logic up from the carbon definition to human.

MistakeNot4892: Tested, works.

out-of-phaze (approved):

PR #3879, opened 2024-04-11
[MDB IGNORE] [IDB IGNORE] Natural flooring now uses /decl/flooring

MistakeNot4892:
## Description of changes
- Converts all natural turfs to use `/decl/flooring`.
- Moves as much as possible from `/turf/floor/natural` to `/turf/floor` and `/decl/flooring`.

Unifies all the code in one spot and makes it much easier to avoid ChangeTurf().
Prep work for layered flooring in the future.

## TODO
- [X] Readd digging
- [x] Test digging
- [X] Test removing and replacing tiles
- [x] Readd everything else clobbered by merge resolution.
- [x] Finish converting natural turfs (lava)
- [x] Get edges working again.
- [x] Get edge layering working again.
- [x] Test all natural flooring.
- [x] Test digging products and trenches.
- [x] Test tilling natural floors.
- [x] Add map migration.

## Test log
- [x] Test tiling over space/open space.
- [x] removing and replacing tiles from plating
- [X] removing and replacing reinforcement from plating
- [X] removing and replacing wooden planks from plating
- [x] removing and replacing tiles from natural turfs
- [x] removing and replacing surface from natural turfs
- [x] removing and replacing wooden planks from natural turfs
- [x] dismantling walls - not updating AO, leaving open space holes
- [x] test roofing a natural turf
- [x] test roofing a built turf
- [x] test digging a trench
- [x] test backfilling a trench
- [X] test building and removing log/brick walls
- [x] test removing ship walls
- [x] test removing natural walls
 
## Testing notes
- Basic steel tile/steel reinforcement/wooden tiles can be removed but not put back on.
- AO is being offset as the turf moves when digging trenches, rather than layering beneath the southern turf.

## Authorship
Myself.

## Changelog


code/game/turfs/floors/subtypes/floor_height.dm
out-of-phaze, line 1: This almost certainly should not be in the subtypes folder.

code/game/turfs/flooring/flooring_lava.dm
```diff
@@ -1,16 +1,16 @@
...
 	movement_delay = 4
-	dirt_color = COLOR_GRAY20
-	footstep_type = /decl/footsteps/lava
-	is_fundament_turf = TRUE
-	material = /decl/material/solid/stone/basalt // TODO: inherent turf temperature
-	var/list/victims
+	color          = COLOR_GRAY20
+	footstep_type  = /decl/footsteps/lava
+/*
+	set_light(2, l_color = LIGHT_COLOR_LAVA)
```
out-of-phaze: lava not producing light is... kind of a little bit of a big deal, to be honest?

code/game/turfs/flooring/_flooring.dm
```diff
@@ -171,3 +180,82 @@ var/global/list/flooring_cache = list()
...
+			return TRUE
+		to_chat(user, SPAN_NOTICE("You unwrench and remove the [descriptor]."))
+		floor.set_flooring(null, place_product = TRUE)
+		playsound(floor, 'sound/items/Ratchet.ogg', 80, 1)
+		return TRUE
+
+	if(IS_SHOVEL(item) && (flags & TURF_REMOVE_SHOVEL))
+		if(!user.do_skilled(remove_timer, SKILL_CONSTRUCTION, floor) || floor.flooring != src)
+			return TRUE
+		to_chat(user, SPAN_NOTICE("You shovel off the [descriptor]."))
```
out-of-phaze: "you shovel off the grass" sounds kind of bad honestly

out-of-phaze (dismissed):
kinda worried about this concept turning flooring into a tangled mess like organ code, but hopefully taking a closer look in review will make me relax a bit. i do like the benefits of this, just worried about mapper-friendliness and extensibility.

anyway, file in the wrong spot. also it seems like natural turfs no longer have a material? and the tool vs floor hardness logic for digging seems to be completely gone. this seems like a lot of regressions, am i missing something?

out-of-phaze:
did not realize this wasnt review-ready, oops

code/game/turfs/flooring/flooring_lava.dm
MistakeNot4892: Per todo, I still need to integrate DOT and light to /decl/flooring. Thinking maybe it could be a function of the turf temperature.

MistakeNot4892:
image

MistakeNot4892:
PR is mostly fine other than issues with edge layering. Any input on this would be good, I find opening the PR exhausting at the moment.

code/game/turfs/floors/floor_icon.dm
```diff
@@ -7,14 +7,42 @@
...
+/turf/floor/proc/can_layer_over(turf/floor/turf_to_check)
+	if(!istype(turf_to_check))
+		return FALSE
+	var/decl/flooring/my_flooring = get_topmost_flooring()
+	if(!istype(my_flooring) || my_flooring.icon_edge_layer == FLOOR_EDGE_NONE)
+		return FALSE
+	var/decl/flooring/their_flooring = turf_to_check.get_topmost_flooring()
+	if(!istype(their_flooring))
+		return TRUE
+	if(their_flooring?.type == my_flooring.neighbour_type)
```
out-of-phaze, line 30: i guess this is fine since path material presets aren't subtypes now? so it'd just be `/decl/flooring/path/herringbone` instead of `/turf/floor/natural/path/herringbone/basalt`

code/game/turfs/flooring/_flooring.dm
```diff
@@ -123,33 +159,49 @@ var/global/list/flooring_cache = list()
 	if(target.icon_state != target.flooring_override)
 		target.icon_state = target.flooring_override
 
+	if(color)
+		target.color = color
+	else
+		var/decl/material/use_material = target.get_material()
+		target.color = use_material?.color
+
+	var/edge_layer = (icon_edge_layer == FLOOR_EDGE_NONE) ? layer + icon_edge_layer : layer
```
out-of-phaze:
```suggestion
	var/edge_layer = (icon_edge_layer != FLOOR_EDGE_NONE) ? layer + icon_edge_layer : layer
```
this logic is inverted

code/game/turfs/floors/_floor.dm
```diff
@@ -12,98 +12,101 @@
...
-	return (A.level <= LEVEL_BELOW_PLATING && !is_plating()) || ..()
+	if(ispath(material))
+		if(ml)
+			set_turf_materials(material, skip_update = TRUE)
+			queue_icon_update()
+		else
+			set_turf_materials(material)
+	else if(!istype(material))
+		material = null
+		update_icon()
```
out-of-phaze: these icon updates are too early, flooring isn't set yet. also the `else if` means we're doing an unnecessary update if material is already null. should just do one unified icon update at the end, and make sure not to queue it if it's not mapload so that we don't briefly have it have the wrong icon.
```suggestion
			set_turf_materials(material, skip_update = TRUE)
```

code/game/turfs/floors/_floor.dm
```diff
@@ -12,98 +12,101 @@
...
-/turf/floor/Initialize(var/ml, var/floortype)
-	. = ..(ml)
-	if(!floortype && initial_flooring)
-		floortype = initial_flooring
 	if(floortype)
 		if(ml)
 			set_flooring(GET_DECL(floortype), skip_update = TRUE)
 			queue_icon_update()
 		else
 			set_flooring(GET_DECL(floortype))
```
out-of-phaze: we need to handle icon updates later in order to properly handle things like turfs updating neighbors when mined out
```suggestion
			set_flooring(GET_DECL(floortype), skip_update = TRUE)
```

code/game/turfs/floors/_floor.dm
```diff
@@ -12,98 +12,101 @@
...
 
-/turf/floor/proc/set_flooring(var/decl/flooring/newflooring, skip_update)
+	if(fill_reagent_type && get_physical_height() < 0)
+		add_to_reagents(fill_reagent_type, abs(height))
```
out-of-phaze, line 48: move the icon update code here and readd the neighbor update code, or else neighboring edges won't be added or removed on turf change
```suggestion
		add_to_reagents(fill_reagent_type, abs(height))
		
	if(material || flooring || base_flooring)
		if(ml)
			queue_icon_update()
		else
			for(var/direction in global.alldirs)
				var/turf/target_turf = get_step_resolving_mimic(src, direction)
				if(istype(target_turf))
					if(TICK_CHECK) // not CHECK_TICK -- only queue if the server is overloaded
						target_turf.queue_icon_update()
					else
						target_turf.update_icon()
			update_icon()
```

code/game/turfs/floors/floor_icon.dm
```diff
@@ -158,41 +196,41 @@
...
+		if(space_smooth == SMOOTH_ALL)
+			. = TRUE
+
+	//If we get here then its a normal floor
+	else if (istype(opponent, /turf/floor))
+		var/turf/floor/floor_opponent = opponent
+		if (floor_smooth == SMOOTH_ALL)
+			. = TRUE
+		//If the floor is the same as us,then we're linked,
+		else if (floor_opponent.flooring?.type == neighbour_type)
```
out-of-phaze:
```suggestion
		else if (istype(floor_opponent.flooring, neighbour_type))
```
the point of this is so that path subtypes (like herringbone, running, etc) don't overlap each other. without this, cobble paths draw edges over brick paths, which is not great visually and a regression from current behavior

code/game/turfs/flooring/_flooring.dm
```diff
@@ -13,43 +13,53 @@ var/global/list/flooring_cache = list()
 	var/desc
 	var/icon
 	var/icon_base
-	var/color
+	var/color = COLOR_WHITE
 	var/footstep_type = /decl/footsteps/blank
+	var/growth_value = 0
+
+	var/neighbour_type
```
out-of-phaze, line 21: do we have like. a policy on AmEng vs BrEng in variable names btw? and in strings? may be worth having a unified style when it comes to that

icons/turf/flooring/dirt.dmi
out-of-phaze, line 1: trench icon is not greyscaled

icons/turf/flooring/clay.dmi
out-of-phaze, line 1: trench icon is not greyscaled, same goes for a bunch of other ones. i won't list them all, should probably just do a pass? or make 'use atom color' a flag somewhere

out-of-phaze (dismissed):
fixes for some issues i found in local testing, including the layering issue

would also appreciate clarification on why corners are in the edge icon states, and what exactly the corner state is for

code/game/turfs/floors/floor_icon.dm
MistakeNot4892: Yeah exactly.

code/game/turfs/flooring/_flooring.dm
MistakeNot4892: It will cause bloodshed

icons/turf/flooring/clay.dmi
MistakeNot4892: I'll do a pass over em for this PR yeah, I think they actually got greyscaled on main but I clobbered them with the mass icon moves.

MistakeNot4892:
image
Trench edges are overlaying the turf they are meant to be below.

MistakeNot4892:
- Trenches seem to be working.
- Base turf on brick/log walls is set to dirt.
- Base turf for Shadyhills wooden turfs is set to dirt.

MistakeNot4892:
AO bake-in issue persists but otherwise I think this is ready to go.
image

out-of-phaze (dismissed):

code/game/turfs/floors/_floor.dm
MistakeNot4892: Implemented in a slightly different way that did not overlap with this line.

comma (approved):

out-of-phaze (approved):

PR #3880, opened 2024-04-11
Swaps world.turf from space to dark filler.

MistakeNot4892:
## Description of changes
Changes `world.turf` to `/turf/unsimulated/dark_filler` instead of `/turf/space`.

Loaded z-levels aren't randomly surrounded by space.

## Authorship
Myself.

## Changelog
Nothing player-facing.

### DIFF
--- a/code/controllers/subsystems/mapping.dm
+++ b/code/controllers/subsystems/mapping.dm
@@ -59,7 +59,42 @@ SUBSYSTEM_DEF(mapping)
 /datum/controller/subsystem/mapping/PreInit()
 	reindex_lists()
 
+#ifdef UNIT_TEST
+/datum/controller/subsystem/mapping/proc/test_load_map_templates()
+	for(var/map_template_name in map_templates)
+		var/datum/map_template/map_template = get_template(map_template_name)
+		// Away sites are supposed to be tested separately in the Away Site environment
+		if(SSunit_tests.is_tested_separately(map_template))
+			report_progress("Skipping template '[map_template]' ([map_template.type]): Is tested separately.")
+			continue
+		if(map_template.is_runtime_generated())
+			report_progress("Skipping template '[map_template]' ([map_template.type]): Is generated at runtime.")
+			continue
+		load_template(map_template)
+		if(map_template.template_flags & TEMPLATE_FLAG_TEST_DUPLICATES)
+			load_template(map_template)
+	log_unit_test("Map templates loaded.")
+
+/datum/controller/subsystem/mapping/proc/load_template(datum/map_template/map_template)
+	// Suggestion: Do smart things here to squeeze as many templates as possible into the same Z-level
+	if(map_template.tallness == 1)
+		increment_world_z_size(/datum/level_data/unit_test)
+		var/turf/center = WORLD_CENTER_TURF(world.maxz)
+		if(!center)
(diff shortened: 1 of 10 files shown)


PsyCommando:
I'm really confused here.. Isn't that changing a long established assumption? Like, feels like it would cause other problems to surface­, for the benefit of an edge case? Unless I don't understand what you mean by random space turfs surrounding a level?

Also, the transition edge stuff should be able to handle putting whatever filler turf you want around your levels?

out-of-phaze (dismissed):
```dm
///Initialize the turfs on the z-level.
/datum/level_data/proc/initialize_new_level()
	//#TODO: Is it really necessary to do this for any and all levels? Seems mainly useful for space levels?
	var/picked_turf = filler_turf || base_turf //Pick the filler_turf for filling if it's set, otherwise use the base_turf
	var/change_turf = (picked_turf && picked_turf != world.turf)
	var/change_area = (base_area && base_area != world.area)
	if(!change_turf && !change_area)
		return
	var/corner_start = locate(1, 1, level_z)
	var/corner_end =   locate(world.maxx, world.maxy, level_z)
	var/area/A = change_area ? get_base_area_instance() : null
	for(var/turf/T as anything in block(corner_start, corner_end))
		if(change_turf)
			T = T.ChangeTurf(picked_turf)
```
Because of `filler_turf` being `null` and `base_turf` being `/turf/space`, which no longer matches `world.turf`, now every turf created by the unit test template loader has to run `ChangeTurf(/turf/space)`, which takes a long time. `/datum/level_data` or at least `/datum/level_data/unit_test` should have its `filler_turf` set to `/turf/unsimulated/dark_filler`, or else this makes space-level loading dramatically more expensive.

image

However, it also seems that this breaks exoplanet generation somehow.

PsyCommando:
> However, it also seems that this breaks exoplanet generation somehow.

I'm pretty sure random map will replace only turfs set to a specific type. And, the generators all expect space turf by default. That's cause world.turf is a runtime var.

Example:
```
/datum/random_map/noise/desert/replace_space
	descriptor = "desert (replacement)"
	target_turf_type = /turf/space
```

MistakeNot4892:
`Unless I don't understand what you mean by random space turfs surrounding a level?`
When world.maxx and world.maxy change, the extra space is filled with world.turf. This leads to stuff like the overmap and exoplanets being surrounded by space tiles. It also means any newly created level is always going to be space.

I don't think it's an edge case. It's a conceptual problem. If Neb is going to support a variety of maps and settings I don't think every new level being deep space is desirable. Additionally, the adaptive client view can lead to seeing past the border turfs on some levels, into space.

There's also the overhead of initializing starlight and multiz stuff for space when it's in a filler context.

I'll have a look at template target turfs and unit tests later. Obviously if the performance impacts are this significant the change isn't feasible, but in general breaking the hard dependency on /turf/space is a good idea imo.

out-of-phaze:
> `Unless I don't understand what you mean by random space turfs surrounding a level?` When world.maxx and world.maxy change, the extra space is filled with world.turf. This leads to stuff like the overmap and exoplanets being surrounded by space tiles. It also means any newly created level is always going to be space.

so, i think at one point i talked with someone about having each level_data handle filling in new space on world resize, but that adds a significant ChangeTurf cost i think. might be better to just not explicitly support post-SSmapping-init existing-level resizing though, and this PR would be one way

MistakeNot4892:
- Turbolifts on tradeship work with no issues.
- Exoplanets generate fine during round setup.
- Placed Bearcat on a new Z, no issues.

out-of-phaze (approved):

PR #3885, opened 2024-04-15
Lock, key and keyring tweaks.

MistakeNot4892:
## Description of changes
- Locks, locked structures and keys will show their lock data ID so that you can actually tell locks apart.
- Keyrings are now a separate storage item for holding keys.
- `lock_keys` on `/datum/job` can be populated to provide the job with a key/keyring on join.
- Closets can use locks.
- `/obj/abstract/landmark/lock_preset` can be used to set up locks on doors and closets.
- Walking into a locked door with keys equipped on help intent will attempt to unlock the door before opening it.

Better QOL for mapping locks and using keys.

## Authorship
Myself.

## Changelog
:cl:
tweak: Closets can now be locked with a physical lock.
/:cl:


code/game/objects/structures/crates_lockers/closets/__closet.dm
```diff
@@ -55,6 +55,12 @@ var/global/list/closets = list()
 	if(!opened && mapload) // if closed and it's the map loading phase, relevant items at the crate's loc are put in the contents
 		store_contents()
 
+/obj/structure/closet/update_lock_overlay()
+	return // TODO
+
+/obj/structure/closet/can_install_lock()
```
out-of-phaze, line 61: Requesting `/obj/structure/closet/body_bag/can_install_lock() return FALSE` in `bodybag.dm` to avoid installing a lock on what amounts to a large plastic bag. Also unsure if you should be able to put one on `/obj/structure/closet/secure_closet` since it already has a digital lock.

code/game/objects/structures/crates_lockers/closets/__closet.dm
```diff
@@ -83,10 +89,15 @@ var/global/list/closets = list()
...
 	if((setup & CLOSET_HAS_LOCK) && locked)
-		return 0
+		return FALSE
 	if((setup & CLOSET_CAN_BE_WELDED) && welded)
-		return 0
-	return 1
+		return FALSE
+	if(lock)
+		if(usr)
+			try_unlock(usr, usr.get_active_held_item())
```
out-of-phaze: Not sure if I'm a huge fan of using `usr` here, would prefer passing `mob/user` from `toggle()` -> `open()` -> `can_open()`. Is there a reason we can't do that?

out-of-phaze (dismissed):
some minor nitpicks, then it's good to merge

code/game/objects/structures/crates_lockers/closets/__closet.dm
MistakeNot4892: Not wanting to update a bunch of overrides mostly.

out-of-phaze: didn't you do that for doors anyway?

MistakeNot4892: Doors don't have any subtypes that override the logic like secure lockers etc.
I'll have a look.

MistakeNot4892: Wasn't as bad as I thought so just added user params.

code/game/objects/structures/crates_lockers/closets/__closet.dm
MistakeNot4892: Added an explicit return to bodybags and made the base types only physically lockable if not set up for a secure lock.

out-of-phaze (approved):

code/game/objects/structures/crates_lockers/closets/__closet.dm
```diff
@@ -81,18 +87,23 @@ var/global/list/closets = list()
...
+	return TRUE
 
-/obj/structure/closet/proc/can_close()
+/obj/structure/closet/proc/can_close(mob/user)
 	for(var/obj/structure/closet/closet in get_turf(src))
 		if(closet != src)
-			return 0
-	return 1
+			return TRUE
+	return FALSE
```
out-of-phaze:
```suggestion
			return FALSE
	return TRUE
```

out-of-phaze (changes requested):

out-of-phaze (approved):

PR #3899, opened 2024-04-16
Add OpenDream to unit testing

out-of-phaze:
## Description of changes
Ports OpenDream unit testing from tgstation/Goonstation/Paradise/etc.

OpenDream's parsing effectively acts as a more advanced linter, and in the future we'll be able to annotate return types on things. Presently, BYOND accepts but ignores those type annotations, but OD has a PR open (that I've been contributing to) to check them.

Also, once we do this, OD will add us to their CI, ensuring better compatibility.

## Authorship
Took `dm_annotate` from tgstation, and adapted parts of the workflow from them as well. Cross-referenced with Goonstation's implementation. I improved some aspects of it like passing `--version` using the BYOND version environment variables.


ike709:
Pragma documentation: https://github.com/OpenDreamProject/OpenDream/wiki/Pragmas---Error-Emissions

out-of-phaze:
Waiting on https://github.com/OpenDreamProject/OpenDream/issues/1754 now. Still open to feedback.

out-of-phaze:
Ready to go after https://github.com/OpenDreamProject/OpenDream/pull/1762 is merged to OpenDream.

code/datums/mind/mind.dm
```diff
@@ -399,6 +399,7 @@
 				H.implant_loyalty(H, override = TRUE)
 				log_admin("[key_name_admin(usr)] has loyalty implanted [current].")
 			else
+				pass()
```
MistakeNot4892, line 402: What is the point of this compared to just deleting the else?

code/modules/client/client_procs.dm
```diff
@@ -125,6 +125,7 @@ var/global/list/localhost_addresses = list(
 
 	switch (connection)
 		if ("seeker", "web") // check for invalid connection type. do nothing if valid
+			pass()
```
MistakeNot4892, line 128: Wouldn't this be cleaner if the switch was inverted and it just returned instead of this structure?

MistakeNot4892 (approved):

code/datums/mind/mind.dm
out-of-phaze: spacemandmm sometimes throws inexhaustive switch errors for certain things like rand() in a switch condition and *maybe* alert() and a few other things. in this case specifically it also makes it clear that the case was explicitly expected/accounted for by whoever wrote the code. i also just wanted to keep the structure of existing code, though i mean, i can easily remove it

MistakeNot4892: Doesn't impact much either way, just looked odd.

code/modules/client/client_procs.dm
out-of-phaze: what do you mean? would have to do `if(connection == "seeker" || connection == "web")` which doesn't feel any cleaner to me

MistakeNot4892: It wouldn't need a switch, else or pass(), which imo would be cleaner. Not really fussed.

PR #3901, opened 2024-04-17
Repaths sensors off /accessory.

MistakeNot4892:
## Description of changes
Splits buddy tags and vitals sensors off accessory onto their own subtype.

Breaks #3798 up for ease of review.
Working towards destroying /obj/item/clothing/accessory.

## Authorship
Myself.

## Changelog
Nothing player-facing.


MistakeNot4892:
Working.

code/modules/clothing/sensors/_sensor.dm
```diff
@@ -0,0 +1,12 @@
+/obj/item/clothing/sensor
+	name = "sensor"
+	abstract_type = /obj/item/clothing/sensor
+	slot_flags = SLOT_TIE
+	icon_state = ICON_STATE_WORLD
+	accessory_slot = ACCESSORY_SLOT_SENSORS
```
out-of-phaze, line 6: does this mean you won't be able to have both a buddy tag and a vitals sensor on the same uniform?

MistakeNot4892: Shouldn't do, it's not limited to one per slot type.

out-of-phaze (dismissed):

out-of-phaze (approved):

PR #3903, opened 2024-04-17
Repathing ties into their own subtype.

MistakeNot4892:
Splitting off the big accessory rework. Repaths ties onto their own subtype.


MistakeNot4892:
Tested, appears to all be working.

out-of-phaze (dismissed):

out-of-phaze (approved):

PR #3910, opened 2024-04-18
Optimizing storage tests.

MistakeNot4892:
Hopefully should prevent the runtime we're seeing sometimes in CI.

### DIFF
--- a/code/unit_tests/map_tests.dm
+++ b/code/unit_tests/map_tests.dm
@@ -264,10 +264,13 @@
 /datum/unit_test/storage_map_test/start_test()
 	var/bad_tests = 0
 
-	for(var/datum/storage/storage in world)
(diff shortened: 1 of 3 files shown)


comma (approved):

code/unit_tests/map_tests.dm
```diff
@@ -264,18 +264,20 @@
 /datum/unit_test/storage_map_test/start_test()
 	var/bad_tests = 0
 
-	for(var/datum/storage/storage in world)
+// We have to ifdef this because _test_storage_items doesn't exist when UNIT_TEST isn't defined.
+#ifdef UNIT_TEST
```
out-of-phaze, line 268: i think maybe if UNIT_TEST isn't defined this should always fail, since you can run map tests manually. that or make it fall back to `in world`

out-of-phaze (changes requested):

out-of-phaze (approved):

PR #3922, opened 2024-04-19
Repaths badges and medals off /accessory.

MistakeNot4892:
Splitting off accessory PR, etc etc.


comma (dismissed):

out-of-phaze (approved):
24h including conflict resolution rule, so I'm merging.

PR #3940, opened 2024-04-26
Adds grafadreka.

MistakeNot4892:
DNM WIP opening for diff, will probably split the systems out into several PRs.

## Description of changes
- Adds a category system to limbs and uses it for a generalized stance calculation that should support both bipeds and quadrupeds.
- Adds drakes as an example quadrupedal species.

More flexible bodytypes.

## TODO
- [x] Work out why gibbing a drake crashes the client.
- [x] Readd mouth gripper.
- [x] Add resting/sitting state icons.
- [x] Add alt icons for backpacks.
- [x] Test/fix stance code
- [X] Add a way to distinguish between sitting and lying.
- [x] Add drake sap reserve and spitting.
- [x] Add drake toxin and melee bonus on prone/stunned/confused mobs.
- [x] Restrict drakes from jobs by default.
- [x] Add drake healing.
- [x] Write drake lore strings.
 
## Authorship
Myself.

## Changelog
Nothing player-facing.


code/modules/clothing/_clothing.dm
```diff
@@ -65,6 +68,15 @@
 /obj/item/clothing/proc/is_accessory()
 	return istype(loc, /obj/item/clothing)
 
+/obj/item/clothing/proc/setup_equip_flags()
+	if(bodytype_equip_flags & BODY_FLAG_EXCLUDE)
+		bodytype_equip_flags |= BODY_FLAG_QUADRUPED
+	else
+		bodytype_equip_flags &= ~BODY_FLAG_QUADRUPED
+
+/obj/item/clothing/proc/setup_sprite_sheets()
```
out-of-phaze: this proc is overridden on `/obj/item/backpack` and `/obj/item/card/id` later, it should probably be on item

out-of-phaze (dismissed):
`cards_ids.dm:290` should probably only specify either width or height and rely on auto-setting to maintain the aspect ratio, in order to prevent the 64x32 drake sprite from being squashed to be 80x80. This might? also be an issue with unathi and other 32x40 sprites, it was probably just overlooked because it's less obvious.

MistakeNot4892:
Depends on #3952 now.

MistakeNot4892:
Depends on #4017

out-of-phaze (approved):

PR #3947, opened 2024-04-28
Fix being unable to open storage on accessories

### LINKED ISSUE #3946: The pouches inventory on the armor doesn't open
#### Description of issue
The pouches inventory doesn't open when I click on the armor that has the pouches accessory

The pouches inventory on the armor should open when I click on the armor

#### Steps to reproduce
Eqiup the armor with pouches
Click the armor
The pouches inventory doesn't open

Any pouches on any armor I guess

???

Doesn't matter

#### Issue bingo

out-of-phaze:
## Description of changes
Clothing `attack_hand` propagates to accessories properly. `can_interact_with_storage()` now checks for accessory membership (so you can't interact with storage pouches inside storage pouches) on clothes, so that it can actually open. Picking up items checks `can_be_picked_up()` so that we can prevent accessories from being plucked off of clothes on accident.

Fixes #3946.

### DIFF
--- a/code/modules/clothing/_clothing_accessories.dm
+++ b/code/modules/clothing/_clothing_accessories.dm
@@ -137,9 +137,23 @@
 			accessory.emp_act(severity)
 	..()
 
-/obj/item/clothing/attack_hand(var/mob/user)
+// Make 'strict' a bit less strict, for accessories.
+// If we're checking strictly and our parent is an accessory,
+// This will need to be handled differently if we ever allow non-clothing accessories!
+/obj/item/clothing/can_interact_with_storage(user, strict = FALSE)
+	if((. = ..(user, FALSE)) || !strict) // Ignore the parent strictness check.
+		return .
 	if(istype(loc, /obj/item/clothing))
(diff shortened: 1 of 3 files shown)


code/modules/clothing/_clothing_accessories.dm
```diff
@@ -137,9 +137,23 @@
...
 	..()
 
-/obj/item/clothing/attack_hand(var/mob/user)
+// Make 'strict' a bit less strict, for accessories.
+// If we're checking strictly and our parent is an accessory,
+// This will need to be handled differently if we ever allow non-clothing accessories!
+/obj/item/clothing/can_interact_with_storage(user, strict = FALSE)
+	if((. = ..(user, FALSE)) || !strict) // Ignore the parent strictness check.
+		return .
 	if(istype(loc, /obj/item/clothing))
```
MistakeNot4892, line 146: There's a couple of places that typecheck this, they could possibly just use `is_accessory()` (which is just the same typecheck).

MistakeNot4892 (approved):

PR #3948, opened 2024-04-28
[MDB IGNORE] [IDB IGNORE] Splitting up uniforms and implementing component clothing.

MistakeNot4892:
## Description of changes
- Implements dresses, shirts, skirts, jumpsuits and several other types as their own subtypes of clothing instead of sitting under uniforms.
- Splits a lot of uniforms up into ensembles using starting_accessories.

More flexibility

## TODO
- [x] Finish this pass over uniforms or decide the PR is too large.
- ~~Adjust equip procs so that they try to equip the accessory to the already-equipped item.~~ split to a separate PR.
- ~~Reorganize the clothing icons folder.~~

## Authorship
Myself.

## Changelog
Will fill this out before merge.


MistakeNot4892:
Will depend on #3951 for some accessory and state CI stuff.

code/modules/clothing/webbing/holster.dm
```diff
@@ -32,9 +32,9 @@
 	var/datum/extension/holster/H = get_extension(src, /datum/extension/holster)
 	H.examine_holster(user)
 
-/obj/item/clothing/webbing/holster/on_attached(obj/item/clothing/under/S, mob/user)
+/obj/item/clothing/webbing/holster/on_attached(var/obj/item/clothing/holder, var/mob/user)
 	. = ..()
-	var/obj/item/clothing/holder = loc
+	holder = loc
```
out-of-phaze: uh, was this ever valid/necessary? like... on_attached can return FALSE meaning it didn't attach and nothing got moved, but like, if it does succeed, `loc` will be `holder` (because `..()` does `forceMove(holder)`). this line just feels like it's redundant at best and a bad idea at worst
```suggestion
```

out-of-phaze (changes requested):
aside form this, looks good. dreading having to convert stuff downstream eventually but at least i can keep them mostly as-is now

out-of-phaze (approved):

PR #3968, opened 2024-05-04
Adding plant oil as a lantern fuel for the fantasy map.

MistakeNot4892:
## Description of changes
- Adds a plant oil reagent usable as nutrition or fuel.
- Adds plant oil to cotton, sunflowers, soybeans and peanuts.
- Adds plant oil to seeds of all kinds.
- Adds `default_liquid_fuel_type` to `/datum/map`, defaulting to welding fuel, but set to plant oil for Shaded Hills.
- Tweaks lanterns to take the map default liquid fuel type.

Allows lanterns to be refuelled on Shaded Hills.
Also provides a potential reagent for frying/cooking or biodiesel in the future.

## Authorship
Myself.

## Changelog
Nothing player facing, really.

### DIFF
--- a/code/modules/hydroponics/seed_packets.dm
+++ b/code/modules/hydroponics/seed_packets.dm
@@ -6,6 +6,7 @@
 	w_class = ITEM_SIZE_SMALL
 	abstract_type = /obj/item/seeds
 	max_health = 10 //Can't set a material, otherwise extracting seeds would generate free materials
+	material = /decl/material/solid/organic/plantmatter/pith
 
 	var/seed_mask_icon = 'icons/obj/seeds/seed_masks.dmi'
 	var/seed_base_name = "packet"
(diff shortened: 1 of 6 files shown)


code/modules/hydroponics/seed_datums.dm
```diff
@@ -4,6 +4,7 @@
 	display_name = "cotton"
 	product_material = /decl/material/solid/organic/plantmatter/pith/husk
 	chems = list(
+		/decl/material/liquid/nutriment/plant_oil = list(3,10),
```
out-of-phaze: pretty sure the oil is specifically in the seeds when it comes to cotton, same goes for sunflowers. but i guess there's no way to have seed-specific chems at the moment?

code/modules/hydroponics/seed_packets.dm
```diff
@@ -7,6 +7,11 @@
 	abstract_type = /obj/item/seeds
 	max_health = 10 //Can't set a material, otherwise extracting seeds would generate free materials
 
+	material = /decl/material/solid/organic/plantmatter/pith
+	matter = list(
+		/decl/material/liquid/nutriment/plant_oil = MATTER_AMOUNT_SECONDARY
```
out-of-phaze: this makes them melt at room temperature which causes gc fails

code/modules/hydroponics/seed_datums.dm
MistakeNot4892: Yeah no support for it. Happy to revise this in the future (making roasting the seeds part of the process etc) but this is the quickest way to get the reagent in that I can think of.

out-of-phaze (dismissed):

out-of-phaze (approved):
Conflict resolution, still good to go

PR #3975, opened 2024-05-06
Mining walls will now use the lower material hardness for checks.

### LINKED ISSUE #3924: Diamond cannot be mined
#### Description of issue
No tool material other than diamond is capable of mining diamonds, because the material check means it needs a material with at least a hardness of 100 (`MAT_VALUE_VERY_HARD + 20`), while no accessible material with a hardness greater than or equal to that exists save for diamond itself.

Presumably at least some alloyed material would be able to mine diamond ore, or explosives would be available for it.

#### Steps to reproduce
1. Try to mine diamond with any tool other than a diamond one.
2. Fail.

Since #3808, which added the hardness check.

N/A. An issue as of latest dev commit, `1675f78dc0f23535bea14a0e35dff98c5f9bf979`.

#### Issue bingo

MistakeNot4892:
Fixes #3924.

### DIFF
--- a/code/game/turfs/walls/wall_natural.dm
+++ b/code/game/turfs/walls/wall_natural.dm
@@ -67,16 +67,21 @@
 // Drill out natural walls.
 /turf/wall/natural/handle_wall_tool_interactions(obj/item/W, mob/user)
 	if(IS_PICK(W) && !being_mined)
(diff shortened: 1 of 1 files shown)


out-of-phaze (approved):
kind of a stopgap, but better than being unable to mine diamond at all.

PR #3976, opened 2024-05-07
Generalizing player-triggered sleeping, shaking awake and hugging.

MistakeNot4892:
## Description of changes
- Shaking/hugging behavior is merged into `default_help_act()`.
- Sleep verb is now on `/mob/living`.

Dogs can snooze.
Player-driven dogs will snooze when logged out.

## Authorship
Myself.

## Changelog
Nothing player-facing.


code/datums/extensions/abilities/abilities_mob.dm
```diff
@@ -2,10 +2,10 @@
 	var/datum/extension/abilities/abilities
 	if(create_if_missing)
 		abilities = get_or_create_extension(src, /datum/extension/abilities)
-	else
+	else if(has_extension(src, /datum/extension/abilities))
 		abilities = get_extension(src, /datum/extension/abilities)
-		if(!abilities)
-			return null
+	if(!abilities)
```
out-of-phaze, line 7: this breaks `create_if_missing`. also you removed a lot of cases where that arg was passed in, is it deprecated now?

out-of-phaze (changes requested):

code/datums/extensions/abilities/abilities_mob.dm
MistakeNot4892: It was being used erroneously and creating empty handlers/extensions. `get_or_create_extension()` is used for creating if missing, worked fine in testing.

out-of-phaze: right, i was looking at lines 10-11, since `add_ability_handler()` also does `get_or_create_extension()`. missed that it was checked earlier. looks good

out-of-phaze (approved):

PR #3977, opened 2024-05-07
Shaded Hills testing tweaks.

MistakeNot4892:
- Storage structures can now have items taken out again.
- Meat, fish, etc. have the correct icons and icon states in the mapper.
- Soup recipes overlap less.
- Cooking vessels can be placed on fire sources without automatically dropping them IN to the fire source.
- Shaded Hills has some props and stuff added to the various dwellings.
- A bunch of other small stuff idk I'm going to bed


code/game/objects/structures/barrel.dm
```diff
@@ -1,15 +1,52 @@
...
+	..()
+	return INITIALIZE_HINT_LATELOAD
+
+/obj/structure/reagent_dispensers/barrel/LateInitialize()
+	..()
+	for(var/obj/item/thing in loc)
+		if(!thing.simulated || thing.anchored)
+			continue
+		if(storage.can_be_inserted(thing, null))
+			storage.handle_item_insertion(null, thing)
```
out-of-phaze: this results in `/datum/unit_test/chemistry_premade_bottles_shall_not_melt` failing because the loc of the spawned stuff is no longer the turf. maybe this should be a mapload only thing?

code/game/objects/structures/barrel.dm
```diff
@@ -1,15 +1,52 @@
...
+	storage                   = /datum/storage/barrel
+	amount_dispensed          = 10
+	possible_transfer_amounts = @"[10,25,50,100]"
+	volume                    = 7500
+	atom_flags                = ATOM_FLAG_CLIMBABLE
+	movable_flags             = MOVABLE_FLAG_WHEELED
+
+/obj/structure/reagent_dispensers/barrel/Initialize()
+	..()
+	return INITIALIZE_HINT_LATELOAD
```
out-of-phaze, line 21:
```suggestion
	. = ..()
	if(mapload)
		return INITIALIZE_HINT_LATELOAD
```
see other comment, this seems to solve the unit testing issue and gives more reliable behavior. if it's created or constructed outside of mapload stuff should have to be manually inserted IMO

out-of-phaze (changes requested):
CI fail, see comments, otherwise looks good

code/game/objects/structures/barrel.dm
MistakeNot4892: Yeah looks like closets use mapload too.

out-of-phaze (approved):

PR #3980, opened 2024-05-08
Xenobotany genetics rewrite.

MistakeNot4892:
WIP DNM, opening for CI and general thoughts before I test it exhaustively.

## Description of changes
- Converts the xenobotany gene and trait system into a true decl system. `/decl/plant_gene` replaces `GENE_X`, `/decl/plant_trait` replaces `TRAIT_Y` (though the defines are still in place to keep the diff size down). 
- Makes trait values lazy. If a trait is unset or set to the default value, it will be dropped from the trait list. `get_trait()` returns the default value if not present in the trait list.
- Generally cleans up existing xenobotany gene code, lazylists things where possible.

A cleanup of this code is desperately overdue. This makes it much, much easier to see what an individual gene does, add or remove genes/traits, and generally get an understanding of the xenobotany code. It also lays the groundwork for future development/cleanup.

## Authorship
Myself.

## Changelog
Nothing player-facing.


code/modules/genetics/plants/trait_requires_nutrients.dm
```diff
@@ -0,0 +1,13 @@
+/// The plant can starve.
+/decl/plant_trait/requires_nutrients
+	name = "required nutrients"
+	shows_extended_data = TRUE
+	default_value = 1
```
comma, line 5: should this be 1? 
If 0.2 and up is 'heavy supply' shouldn't this be something between 0.2 and 0.05?

MistakeNot4892: There's one trait for a boolean 'requires nutrients at all' and a second trait for 'how much nutrients are required', it's odd. I might condense them down into one trait once this is tested.

MistakeNot4892:
Botany machines, disks, products and seeds appear to be working.

out-of-phaze (approved):

PR #3986, opened 2024-05-10
Allow pipes above fatigue pressure to leak

out-of-phaze:
## Description of changes
Pipes above fatigue pressure will now leak.
I, uh. Need to test this, but the idea of doing it fills me with dread.
- [ ] Tested

A ~~ten~~ thirteen year old TODO is fulfilled, and also this provides some telegraphing before that 5% burst chance inevitably procs and the pipe explodes violently.

## Authorship
Me.

## Changelog
:cl:
add: Pipes will now leak if above fatigue pressure. If this happens, act quick, because it'll burst eventually!
/:cl:

### DIFF
--- a/code/modules/atmospherics/pipes.dm
+++ b/code/modules/atmospherics/pipes.dm
@@ -241,9 +241,11 @@
 		burst()
 
 	else if(pressure_difference > fatigue_pressure)
-		//TODO: leak to turf, doing pfshhhhh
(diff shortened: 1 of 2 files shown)


MistakeNot4892:
Seems to have caused a bunch of pipe CI failures.

out-of-phaze:
I've got no clue why this is failing and don't really want to pursue this further right now.

PR #4007, opened 2024-05-13
Fix lighting artifacts on mimic edge boundaries

out-of-phaze:
## Description of changes
Fixes lighting artifacts with mimic edge boundaries by fixing the coordinates used for corner-dir-getting. There's probably a more optimized way to do this without adding an extra variable, but all my attempts kept breaking, so I'll let someone else handle that or take a stab at it myself later.

Fixes lighting artifacts from edge mimic violating the assumption that a turf to the west of a corner will always have an X coordinate lower than the corner's, that a turf to the east will have an X coordinate greater, etc., and the same for north/south and the Y coordinate.

## Authorship
Me. No one else in Coderbus helped at all and I solved it on my own, wtf.


MistakeNot4892 (approved):

out-of-phaze:
Taking a stab at optimizing.

out-of-phaze:
Loaf wins, I'll do it in another PR

PR #4017, opened 2024-05-15
Drake-supporting code pulled out of grafadreka PR.

MistakeNot4892:
- Fixes an odd runtime when readied up with a species who does not have `SPECIES_CAN_JOIN`.
- Moves appearance descriptors (including age) onto bodytype.
- Fixes an issue with circular qdel of ability handlers.
- Passes total dose to affect_overdose() as a parameter.
- Adds a helper proc for retrieving injured organs.
- Adds capacity for organs to register for Stat() updates.
- Adds capacity to flag a species to blacklist itself from all map jobs it is not explicitly whitelisted for. May apply this to vox.
- Adds a has_aura() helper proc.
- A couple of stray icon changes that snuck in.
- Adjusts grab code so that humantype mobs can be ridable.


code/modules/materials/_materials.dm
```diff
@@ -868,7 +868,7 @@ INITIALIZE_IMMEDIATE(/obj/effect/gas_overlay)
 		holder.remove_reagent(type, REAGENT_VOLUME(holder, type))
 		. = TRUE
 
-/decl/material/proc/affect_overdose(var/mob/living/M) // Overdose effect. Doesn't happen instantly.
+/decl/material/proc/affect_overdose(mob/living/M, total_dose) // Overdose effect. Doesn't happen instantly.
```
out-of-phaze, line 871: is `tick_dosage_tracker` a private variable or something? how is total_dose better than just `M.tick_dosage_tracker[src]`?

out-of-phaze (approved):

code/modules/materials/_materials.dm
MistakeNot4892: Per Discord, it's local to the overdose loop.

PR #4022, opened 2024-05-16
Add hue shifting to wood tile sprites

out-of-phaze:
Before:
baae0132-9068-48a8-8b35-5075d7db75e7
After:
image

The intent here is to add more depth to the sprite, since currently it feels flat and bland.


MistakeNot4892 (approved):
Github made me post a comment containing just `- [ ]` and now won't let me delete it.

Issue #3999, opened 2024-05-12
Multiz shuttles replace open turfs with space.


MistakeNot4892:
#### Description of issue
Copied from https://github.com/ScavStation/ScavStation/issues/888
Moving multiz shuttles with open space replace that open space with `/turf/space`.

#### Difference between expected and actual behavior
Space type should be preserved between moves.

#### Steps to reproduce
Build a multiz shuttle, try to move it.

#### Specific information for locating
Unsure, cannot repro this locally due to not having a multiz shuttle handy.

#### Length of time in which bug has been known to occur
Unclear.

#### Client version, Server revision & Game ID
Reported on Scav main, which is current Neb stable.

#### Issue bingo

- [X] Issue could be reproduced at least once
- [ ] Issue could be reproduced by different players
- [ ] Issue could be reproduced in multiple rounds
- [ ] Issue happened in a recent (less than 7 days ago) round
- [X] [Couldn't find an existing issue about this](https://github.com/NebulaSS13/Nebula/issues)

out-of-phaze:
I............ don't think this is an issue anymore, please open a new issue if it still is

PR #4027, opened 2024-05-18
Assorted ZAS fixes

NataKilar:
## Description of changes
An assortment of ZAS fixes, mostly related to the creation of unsimulated edges between zones and space.
* Creating a hole in the floor into space will now create an unsimulated edge as expected. This is done by checking if the ``zone_membership_candidate`` variable has changed to ``FALSE`` during a turf change. Every other case should be covered by the typical checks for exterior atmosphere participation.
* Changed the default ``open_turf`` for flooring and natural flooring to airless variants of open space. This fixes a bug where space tiles weren't being removed from Z-stacks when the bottom most space tile was replaced by plating. I haven't tested how this affects issue #3999 yet.
* ~~Doors and other atoms that need to update ZAS connections will now update their ``loc``'s neighbors if the atom is located on top of an exterior turf. This fixes a bug where unsimulated edges weren't being updated or created as expected~~ Supplanted by new update code.
* Fixes a longstanding issue where connections for ZAS edges in the up and down direction weren't being added to turfs' connection managers

New Changes 06/08/2024
* You can now safely call ``SSair.mark_for_update`` on non-ZAS participating turfs to create the appropriate unsimulated edges
* Improved consistency of updating exterior air participation. Newly ZAS participating turfs will now only assume exterior air if they were previously outside
* Adds ``keep_air = TRUE`` to a variety of ChangeTurf calls where appropriate
* Elevators no longer move air by default.
* Shuttles now create roofs at their destination before moving, and only remove the roof after they have moved.

As a consequence of these changes, ``keep_air_below`` has been cut entirely.

Fixes a variety of ZAS issues stemming from the simulated/unsimulated turf merger. There is still an issue with gas conservation when merging two zones between z-levels, which I still need to resolve.

## Authorship
Myself. Thanks to Penny for the multi-Z ``get_dir`` code.
## Changelog
:cl:
bugfix: Fixed space tiles not creating vacuums in certain conditions
/:cl:


MistakeNot4892 (dismissed):
Fixes all look good to me and you justified/explained the ones I was confused about on Discord, so looks fine to me.

code/modules/multiz/basic.dm
```diff
@@ -33,4 +33,25 @@ var/global/list/z_levels = list() // Each Z-level is associated with the relevan
 	else if (dir == DOWN)
 		. = GetBelow(ref)?.resolve_to_actual_turf()
 	else
-		. = get_step_resolving_mimic(ref, dir)
\ No newline at end of file
+		. = get_step_resolving_mimic(ref, dir)
+
+/proc/get_dir_multiz(turf/us, turf/them)
+	us = get_turf(us)
```
comma: why do get turf here, we're already getting a turf don't we?

code/modules/multiz/basic.dm
```diff
@@ -33,4 +33,25 @@ var/global/list/z_levels = list() // Each Z-level is associated with the relevan
...
+	them = get_turf(them)
+	if(!us || !them)
+		return 0
+	if(us.z == them.z)
+		return get_dir(us, them)
+	else
+		var/turf/T = GET_ABOVE(us)
```
comma: Maybe this can be optimized a little by first comparing z levels and then getting above/below turf.
something like if them.z > us.z && get_above(us) and vice versa for below

MistakeNot4892 (changes requested):
Per Discord, issues with elevators need to be sorted out.

code/game/area/areas.dm
```diff
@@ -140,13 +140,14 @@ var/global/list/areas = list()
 		if(adjacent_turf)
 			adjacent_turf.update_registrations_on_adjacent_area_change()
 
-	T.last_outside_check = OUTSIDE_UNCERTAIN
```
MistakeNot4892, line 143: Was it intentional to remove the cache invalidation?

MistakeNot4892: Noting that you moved it into update_external_atmos_participation(), but it will now only be nulled if T.is_outside is OUTSIDE_AREA.

NataKilar: That seems correct. Looking at ``proc/is_outside``, I don't think changing the area can affect the outside status of turfs which don't have ``is_outside = OUTSIDE_AREA``

MistakeNot4892: Ah I actually didn't grok that this was area updates specifically. Yeah looks fine in that case.

MistakeNot4892 (approved):

PR #4028, opened 2024-05-19
Script tweaks from trying to run 515 on testing server

MistakeNot4892:
Previous timestamp included forward slashes.

### DIFF
--- a/scripts/server.sh
+++ b/scripts/server.sh
@@ -23,8 +23,6 @@ fi
 touch server_running
 trap "cleanup" EXIT
 
-exec 5>&1 # duplicate fd 5 to fd 1 (stdout); this allows us to echo the log during compilation, but also capture it for saving to logs in the case of a failure
-
 [[ -e stopserver ]] && rm stopserver
 while [[ ! -e stopserver ]]; do
 	MAP="$(cat use_map || echo "example")"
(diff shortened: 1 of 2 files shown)


out-of-phaze (dismissed):

scripts/server.sh
```diff
@@ -46,7 +44,7 @@ while [[ ! -e stopserver ]]; do
 
 	# Compile
 	echo "Compiling..."
-	DMoutput="$(./scripts/dm.sh $EXTRA_DM_SH_ARGS -M$MAP $DME.dme | tee /dev/fd/5)" # duplicate output to fd 5 (which is redirected to stdout at the top of this script)
+	DMoutput="$(./scripts/dm.sh $EXTRA_DM_SH_ARGS -M$MAP $DME.dme | tee /dev/fd/2)" # duplicate output to fd 2 for logging purposes
```
out-of-phaze, line 47: fd 2 is stderr which is displayed by default, isn't it? then again fd 5 is duplicated to fdout. mostly i'm worried this will end up throwing extra junk into the log file, though maybe that's okay? if it isn't then the only solution is to make it mandatory to run with root permissions.

MistakeNot4892: I think the idea is just to duplicate it to -some- pipe other than stdout so we can return a compile log as well as print the output to the console. My fd understanding is -extremely- limited though.

out-of-phaze: well if it works i guess it's fine

out-of-phaze (approved):

MistakeNot4892:
Waiting to see if Qume can test these on Scav since that's on 514, just to make sure we aren't soft-removing 514 support.

MistakeNot4892:
Ship has sailed on this one, it's needed if we want to test the next staging push.

PR #4040, opened 2024-05-21
Fixing bad proc ref in timers.

MistakeNot4892:
Other timer cmp procs just use `/proc` and this one is generating a 'bad proc' runtime on dev testing.

### DIFF
--- a/code/controllers/subsystems/timer.dm
+++ b/code/controllers/subsystems/timer.dm
@@ -273,7 +273,7 @@ SUBSYSTEM_DEF(timer)
 		return
 
 	// Sort all timers by time to run
-	sortTim(alltimers, PROC_REF(cmp_timer))
+	sortTim(alltimers, /proc/cmp_timer)
 
 	// Get the earliest timer, and if the TTR is earlier than the current world.time,
 	// then set the head offset appropriately to be the earliest time tracked by the


code/controllers/subsystems/timer.dm
```diff
@@ -273,7 +273,7 @@ SUBSYSTEM_DEF(timer)
 		return
 
 	// Sort all timers by time to run
-	sortTim(alltimers, PROC_REF(cmp_timer))
+	sortTim(alltimers, /proc/cmp_timer)
```
out-of-phaze, line 276: technically the correct thing here would be `GLOBAL_PROC_REF(cmp_timer)`, although i think that's mostly just for future-proofing. `PROC_REF()` tries to get `/datum/controller/subsystem/timer/proc/cmp_timer`, which doesn't exist. `GLOBAL_PROC_REF(cmp_timer)` just gets (the name of) `/proc/cmp_timer`.

aside from weird ref/GC/harddel issues, using a direct ref instead of the PROC_REF macro can lead to the wrong override getting called (base definition instead of most-overridden). but since global procs can't (currently) be overridden, that's not an issue. not blocking esp. since other sortTim calls also don't use GLOBAL_PROC_REF (i'll probably get them all later) but figured i'd note it.

out-of-phaze (approved):

PR #4043, opened 2024-05-21
Fixing some issues with printing tape from a fabricator.

MistakeNot4892:
Fixes https://github.com/ScavStation/ScavStation/issues/974

### DIFF
--- a/code/modules/fabrication/designs/_design.dm
+++ b/code/modules/fabrication/designs/_design.dm
@@ -11,6 +11,9 @@
 	var/max_amount = 1 // How many instances can be queued at once
 	var/list/required_technology
 	var/list/species_locked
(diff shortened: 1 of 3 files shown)


code/modules/fabrication/designs/_design.dm
```diff
@@ -11,6 +11,7 @@
 	var/max_amount = 1 // How many instances can be queued at once
 	var/list/required_technology
 	var/list/species_locked
+	var/handle_as_stack
```
out-of-phaze: i kind of would like this to be renamed `pass_order_multiplier` or something tbh but like, you don't HAVE to do that

out-of-phaze (changes requested):

out-of-phaze (approved):

out-of-phaze (approved):

PR #4044, opened 2024-05-21
Fix Shaded Hills issues for May 21, 2024

out-of-phaze:
## Description of changes
Makes flint strikers not colorable in loadout, and changes the name from striker to flint striker.
Adds density and ClickCross checks to strikers, so you can't create sparks inside a dense turf or object.
Makes folding blades not display their icon when closed, as intended.
Adds a `get_striking_material()` to represent "which material we're hitting something with," to disambiguate between blade/hilt strikes. TODO: integrate this elsewhere, tie into dynamic intents.
Makes folding blades and folding knives with non-ferrous handles unable to be used to make sparks
Adds a commit that was supposed to be in #4021 (`Replaces non-furniture uses of ebony on Shaded Hills with walnut, and makes walnut the dominant tree instead of ebony.`)

Fixes some broken code and a few bugs/oversights, plus a commit I forgot to include in a prior PR.

### DIFF
--- a/code/game/objects/items/blades/folding.dm
+++ b/code/game/objects/items/blades/folding.dm
@@ -49,5 +49,6 @@
 		attack_verb = closed_attack_verbs
 	..()
 
-/obj/item/knife/folding/get_mob_overlay(mob/user_mob, slot, bodypart, use_fallback_if_icon_missing = TRUE, skip_adjustment = FALSE)
+// Only show the inhand sprite when open.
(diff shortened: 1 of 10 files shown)


code/game/objects/items/blades/folding.dm
```diff
@@ -49,5 +49,10 @@
 		attack_verb = closed_attack_verbs
 	..()
 
-/obj/item/knife/folding/get_mob_overlay(mob/user_mob, slot, bodypart, use_fallback_if_icon_missing = TRUE, skip_adjustment = FALSE)
```
MistakeNot4892, line 52: Oh god dammit, that explains why the inhands were wonky.

MistakeNot4892 (approved):

PR #4047, opened 2024-05-23
[IDB IGNORE] [MDB IGNORE] More uniform conversions.

MistakeNot4892:
DNM, opening for CI. Depends on #3948.


MistakeNot4892:
This one is a bit halfassed but it does the job of killing /under. I'll do another final PR after this to go through /costume and convert what can be converted.

mods/content/government/away_sites/icarus/icarus.dm
```diff
@@ -164,11 +164,6 @@
 	icon = 'mods/content/government/away_sites/icarus/icarus_sprites.dmi'
 	icon_state = "solgovseal"
 
-/obj/item/clothing/under/ec_uniform
-	name = "expeditionary uniform"
-	desc = "An older model of the utility uniform of the SCG Expeditionary Corps. It has a patch on the left sleeve signifying the wearer served on the SEV Icarus."
-	icon = 'icons/clothing/under/uniform_ec.dmi'
```
MistakeNot4892, line 170: This icon turned out to not have anything in it, for the record.

out-of-phaze (dismissed):
doesn't look like this has a map migration, and it needs one

MistakeNot4892:
> doesn't look like this has a map migration, and it needs one

Done.

out-of-phaze (approved):

PR #4048, opened 2024-05-23
Fix Shaded Hills and fantasy modpack issues, May 22, 2024

### LINKED ISSUE #4049: Cannot pick up clothing storage items off the ground
#### Description of issue
Clothing items with storage, like labcoats, cannot be picked up off the ground and must be dragged into your hand.

You should pick it up.

#### Steps to reproduce
1. Try to pick it up
2. Open storage instead

Since I caused it, I'm literally just making this issue for bookkeeping purposes

#### Issue bingo

out-of-phaze:
## Description of changes
Fixes some Shaded Hills/fantasy modpack issues discovered in testing on May 22, 2024.

Fixes a few issues that came up in testing.

### DIFF
--- a/code/modules/mob/living/living_attackhand.dm
+++ b/code/modules/mob/living/living_attackhand.dm
@@ -89,9 +89,6 @@
 // Returns TRUE if further interactions should be halted, FALSE otherwise.
 /mob/living/proc/try_awaken(mob/user)
 
-	if(isSynthetic() || stat == CONSCIOUS)
-		return user.attempt_hug(src)
-
 	var/decl/pronouns/pronouns = get_pronouns()
 	var/obj/item/uniform = get_equipped_item(slot_w_uniform_str)
 	if(uniform)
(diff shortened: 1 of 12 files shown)


code/modules/mob/living/living_attackhand.dm
```diff
@@ -89,9 +89,6 @@
 // Returns TRUE if further interactions should be halted, FALSE otherwise.
 /mob/living/proc/try_awaken(mob/user)
 
-	if(isSynthetic() || stat == CONSCIOUS)
```
MistakeNot4892, line 92: I thought this was to stop you from shaking robots awake.

out-of-phaze: silicons already have overrides that do that, and i think you should still be able to shake FBPs to get them up honestly. if that has to be changed though, it's trivial to just add it back without the `stat == CONSCIOUS` check

MistakeNot4892: Eh I don't care much either way.

MistakeNot4892 (approved):

PR #4050, opened 2024-05-24
Fixes clamp() args order (at old `between` helper calls)

cyberillithid:
## Description of changes
PR #2704 had removed the redundant clamp helper functions. But all calls of `between(lo, mid, hi)` had been replaced by `clamp(lo, mid, hi)`, while the documentation names its arguments as `clamp(Number, Low, High)`. Thus, this PR changes order of arguments for all places where `between` had been called once.
Old `Clamp` has the same order of arguments as modern `clamp`, and all usages of `dd_range(lo,hi,mid)` had been already reordered previously.
Technically changing args order still doesn't preserve erroneous behaviour of old helper functions (when lo>hi, those may have returned `lo`) and probably should change nothing in the current behaviour.

Improves readability of code. 

## Changelog
:cl:
refactor: `clamp` args order changed to comprehensible where `between` was used once
/:cl:

### DIFF
--- a/code/modules/shield_generators/shield_generator.dm
+++ b/code/modules/shield_generators/shield_generator.dm
@@ -88,12 +88,12 @@
 	for(var/obj/item/stock_parts/smes_coil/S in component_parts)
 		full_shield_strength += (S.ChargeCapacity / CELLRATE) * 5
 	max_energy = full_shield_strength * 20
-	current_energy = clamp(0, current_energy, max_energy)
+	current_energy = clamp(current_energy, 0, max_energy)
 
 	mitigation_max = MAX_MITIGATION_BASE + MAX_MITIGATION_RESEARCH * total_component_rating_of_type(/obj/item/stock_parts/capacitor)
-	mitigation_em = clamp(0, mitigation_em, mitigation_max)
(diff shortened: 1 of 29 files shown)


code/modules/shield_generators/shield_generator.dm
```diff
@@ -88,12 +88,12 @@
 	for(var/obj/item/stock_parts/smes_coil/S in component_parts)
 		full_shield_strength += (S.ChargeCapacity / CELLRATE) * 5
 	max_energy = full_shield_strength * 20
-	current_energy = clamp(0, current_energy, max_energy)
+	current_energy = clamp(current_energy, 0, max_energy)
```
MistakeNot4892, line 91: This might be why shield generators don't work at the moment @Qumefox

MistakeNot4892 (approved):

code/modules/shield_generators/shield_generator.dm
out-of-phaze: it shouldn't, max(0, current_energy) and max(current_energy, 0) are equivalent. i think the main change is when the second and third arguments are out of order (third smaller than second) but we shouldn't have any of those.

out-of-phaze (approved):

PR #4051, opened 2024-05-25
Allow hnolls to wear gloves

out-of-phaze:
## Description of changes
Allows hnolls to wear gloves, using the same code as neoavians.

Hnolls can now wear gloves. According to Loaf their claws are more like dog claws so there shouldn't be any conceptual issues with them tearing gloves.


MistakeNot4892:
That was more of a suggestion/thought than a statement of fact, but I don't mind either way.

MistakeNot4892 (approved):

PR #4063, opened 2024-05-26
Syncing changes from ScavStation.

MistakeNot4892:
Pulling a bunch of changes from Scav that I've been meaning to grab for awhile.

- Split vital sensor equipping into a proc so it can be overridden/reordered downstream.
- Entertainment intercoms broadcast by default.
- Cleans suit refitting names up a bit.
- TODO, still filtering out the changes we don't want.

Things I'm not sure about:
- Traitor collaborators get an uplink.


MistakeNot4892:
Not sure what MapDiff2 is choking on, I fixed the missing paths in the maps I've edited.

out-of-phaze:
MDB issue should be fixed now, so when you get back it'll just need a push to re-run.

I think it might be okay to give collaborators an uplink *with no telecrystals.* That way, if they want gear, they'll need to ask an actual antagonist for TC. On the other hand though, they don't have an antagonist role and shouldn't have access to antagonist things like uplinks... maybe make them provocateurs? But then what if they don't encounter any traitors at all?

MistakeNot4892:
Will come back to this after staging.

PR #4065, opened 2024-05-27
Update spacemandmm.toml to fix mapdiffbot issues

out-of-phaze:
## Description of changes
Excludes certain types from rendering in mapdiffbot, and excludes some render passes that don't apply to us.
Made with reference to https://github.com/ParadiseSS13/Paradise/blob/9f963ffd2226191923d7871ef6c1965676443c7e/SpacemanDMM.toml and 

The mapdiffbot should render things better, since some TG-only features were disabled.

## Authorship
Made with advice from AA07.

### DIFF
--- a/SpacemanDMM.toml
+++ b/SpacemanDMM.toml
@@ -51,3 +51,22 @@ macro_undefined_no_definition = "error"
 
 # Raised by Object Tree
 override_precedes_definition = "error"
+
+[map_renderer]
+hide_invisible = [
+	"/obj/effect/step_trigger",
+	"/obj/abstract",
+]
+
+[map_renderer.render_passes]
+overlays = false
+gravity-gen = false
+pretty = false
+icon-smoothing = false
+icon-smoothing-2016 = true
+smart-cables = false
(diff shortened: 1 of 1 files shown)


SpacemanDMM.toml
```diff
@@ -51,3 +51,22 @@ macro_undefined_no_definition = "error"
...
+[map_renderer.render_passes]
+overlays = false
+gravity-gen = false
+pretty = false
+icon-smoothing = false
+icon-smoothing-2016 = true
+smart-cables = false
+# TODO: add support for directional_offsets var to Jupyterkat's fork?
+offsets = false
+# maybe also add support for our random spawners?
```
MistakeNot4892, line 70: What does this mean for random spawners being mapped?

out-of-phaze: Hmm. For TG random spawners it replaces things like random vendor spawners with a vendor, etc. I'm not sure how feasible that is. It could be useful to enable for things like regenerating the nanoui maps, though.

I also need to add a TODO about a Bay-specific wall smoothing pass

MistakeNot4892 (approved):

PR #4068, opened 2024-05-31
Fix many double variable definitions

out-of-phaze:
## Description of changes
Fixes issues where a variable was set twice in one type definition block, which can lead to confusion as only the second value applies.
I tried doing it in CI but automating it would be incredibly difficult; only ripgrep is powerful enough to handle it, and the grep is really brittle. Here's the regex anyway, it has to be used in VSC because that automatically adds `--crlf` and does some annoying conversion of `\n` to `\r\n` in the regex.
`^(?!\/\/)\/?[\w\/]+\/\w+(?:\s*\/\/.*)?$(?:\n+^\s*[^\s\/].*$)*?\n+\t(\w+\b)\s*=\s*.+$\n+(?:^\s*?(?:[^\s\/]|\/\/).*$\n+)*?\t\1\b`

Fixes issues where the actual value of a variable was unclear, or where the intended value was clobbered by a later definition. A lot of these were cases where `obj_flags` was defined twice, so I kept both sets of flags and just combined them into one.

### DIFF
--- a/code/datums/inventory_slots/slots/slot_cuffs.dm
+++ b/code/datums/inventory_slots/slots/slot_cuffs.dm
@@ -1,5 +1,5 @@
 /datum/inventory_slot/handcuffs
-	slot_id = "Handcuffs"
+	slot_name = "Handcuffs"
 	slot_id = slot_handcuffed_str
 	skip_on_inventory_display = TRUE // Handcuffs have their own logic on examine.
 	skip_on_strip_display = TRUE // Handcuffs are removed with their own button in the strip menu.
--- a/code/game/objects/items/weapons/circuitboards/machinery/research.dm
+++ b/code/game/objects/items/weapons/circuitboards/machinery/research.dm
@@ -229,8 +229,8 @@
 	build_path = /obj/machinery/merchant_pad
 	board_type = "machine"
 	origin_tech = @'{"programming":6,"wormholes":6,"esoteric":1}'
-	req_components = list(/obj/item/stack/cable_coil = 15)
 	req_components = list(
+		/obj/item/stack/cable_coil = 15,
 		/obj/item/stock_parts/subspace/amplifier = 1,
 		/obj/item/stock_parts/subspace/ansible = 1,
 		/obj/item/stock_parts/subspace/crystal = 1,
--- a/code/game/objects/items/weapons/storage/boxes.dm
+++ b/code/game/objects/items/weapons/storage/boxes.dm
@@ -706,7 +706,6 @@
 	name = "assorted parts pack"
 	icon = 'icons/obj/items/storage/part_pack.dmi'
 	icon_state = "big"
-	icon_state = "part"
 	w_class = ITEM_SIZE_NORMAL
 
 /obj/item/box/parts/WillContain()
--- a/code/modules/materials/definitions/gasses/material_gas_mundane.dm
+++ b/code/modules/materials/definitions/gasses/material_gas_mundane.dm
@@ -236,7 +236,6 @@
 	boiling_point = -33 CELSIUS
 	gas_symbol_html = "NH<sub>3</sub>"
 	gas_symbol = "NH3"
-	metabolism = 0.05 // So that low dosages have a chance to build up in the body.
 	taste_description = "mordant"
 	taste_mult = 2
 	lore_text = "A caustic substance commonly used in fertilizer or household cleaners."
--- a/code/modules/tools/subtypes/xenoarchaeology_picks.dm
+++ b/code/modules/tools/subtypes/xenoarchaeology_picks.dm
@@ -106,7 +106,6 @@
 	name              = "excavation pick set"
 	icon              = 'icons/obj/items/storage/excavation.dmi'
 	icon_state        = "excavation"
-	item_state        = "utility"
 	desc              = "A rugged case containing a set of standardized picks used in archaeological digs."
 	item_state        = "syringe_kit"
 	slot_flags        = SLOT_LOWER_BODY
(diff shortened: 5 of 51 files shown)


comma (dismissed):

code/datums/inventory_slots/slots/slot_cuffs.dm
```diff
@@ -1,5 +1,4 @@
 /datum/inventory_slot/handcuffs
-	slot_id = "Handcuffs"
```
MistakeNot4892, line 2: This is actually an error rather than a duplicate var, should be `slot_name`

code/game/objects/items/weapons/circuitboards/machinery/research.dm
```diff
@@ -229,7 +229,6 @@
 	build_path = /obj/machinery/merchant_pad
 	board_type = "machine"
 	origin_tech = @'{"programming":6,"wormholes":6,"esoteric":1}'
-	req_components = list(/obj/item/stack/cable_coil = 15)
```
MistakeNot4892, line 232: Should the cable coil be moved into the second list?

code/game/objects/items/weapons/storage/boxes.dm
```diff
@@ -706,7 +706,6 @@
 	name = "assorted parts pack"
 	icon = 'icons/obj/items/storage/part_pack.dmi'
 	icon_state = "big"
-	icon_state = "part"
```
MistakeNot4892, line 709: `part` seems to be the correct state here.

code/modules/materials/definitions/gasses/material_gas_mundane.dm
```diff
@@ -236,7 +236,6 @@
 	boiling_point = -33 CELSIUS
 	gas_symbol_html = "NH<sub>3</sub>"
 	gas_symbol = "NH3"
-	metabolism = 0.05 // So that low dosages have a chance to build up in the body.
```
MistakeNot4892, line 239: This doesn't seem to be set elsewhere. Base `gas` doesn't set metabolism and `/decl/material` has it set to `0.2`.

MistakeNot4892: Disregard, Github wasn't showing me context.

code/modules/tools/subtypes/xenoarchaeology_picks.dm
```diff
@@ -106,7 +106,6 @@
 	name              = "excavation pick set"
 	icon              = 'icons/obj/items/storage/excavation.dmi'
 	icon_state        = "excavation"
-	item_state        = "utility"
```
MistakeNot4892, line 109: I think `utility` might have been the correct state.

mods/content/corporate/datum/antagonists/commando.dm
```diff
@@ -21,7 +21,7 @@
 	shoes =   /obj/item/clothing/shoes/jackboots/swat
 	glasses = /obj/item/clothing/glasses/thermal
 	mask =    /obj/item/clothing/mask/gas/syndicate
-	l_ear =   /obj/item/radio/headset/hacked
+	r_ear =   /obj/item/radio/headset/hacked
```
MistakeNot4892: Are they really meant to have two headsets?

mods/content/corporate/datum/antagonists/deathsquad.dm
```diff
@@ -31,7 +30,7 @@
 	name =     "Special Role - Deathsquad Commando"
 	l_ear =    /obj/item/radio/headset/ert
 	uniform =  /obj/item/clothing/jumpsuit/green
-	l_ear =    /obj/item/radio/headset/hacked
+	r_ear =    /obj/item/radio/headset/hacked
```
MistakeNot4892: As above, it seems kinda bizzare to have a headset in each ear.

MistakeNot4892 (changes requested):
Left a handful of comments.

code/modules/tools/subtypes/xenoarchaeology_picks.dm
out-of-phaze: it literally does not exist, i checked

mods/content/corporate/datum/antagonists/commando.dm
out-of-phaze: if you were present i would've asked. but i assume so, one lets you intercept the station's comms (but it's useless for speaking into, thus right ear) and the other is used for communicating with your team.

code/game/objects/items/weapons/storage/boxes.dm
out-of-phaze: no, i assumed `part` was for the single-type packs later in the file. `big` is unused elsewhere and makes sense given that this is `ITEM_SIZE_NORMAL` and has more parts, while the other one is `ITEM_SIZE_SMALL` and has fewer

mods/content/corporate/datum/antagonists/deathsquad.dm
MistakeNot4892: Would suggest adding this to the ERT encryption key and removing the r_ear set:
````
/obj/item/encryptionkey/ert/Initialize(ml, material_key)
	. = ..()
	can_decrypt |= get_all_station_access()
````

mods/content/corporate/datum/antagonists/commando.dm
MistakeNot4892: Would suggest adding this and removing the other headsets:
````
/obj/item/radio/headset/mercenary/commando
	encryption_keys = list(
		/obj/item/encryptionkey/mercenary,
		/obj/item/encryptionkey/hacked
	)
````

mods/content/corporate/datum/antagonists/deathsquad.dm
out-of-phaze: it's something only deathsquad is supposed to have though, not normal ERT

MistakeNot4892: Per Discord, ERT's meant to have all station channel access.

MistakeNot4892 (approved):

PR #4071, opened 2024-06-02
Disables temperature-based melting until it is refined.

MistakeNot4892:
Melting is really problematic at the moment and needs more time in the oven, disabling temperature-based melting for the time being.


out-of-phaze (approved):

out-of-phaze:
this seems to make it so that you can't smelt ores in the kiln on shadyhills

MistakeNot4892:
Yeah it needs to be a type check on loc for crucibles on dev.

PR #4079, opened 2024-06-04
Fixes issues with object embedding.

MistakeNot4892:
- Embedding is now unified under `try_embed_in_mob()`.
- Embedding will proactively try to find an appropriate limb and create a wound if unsupplied.
- Failed embedding will not teleport the item inside the mob with no way to remove it.
- Pinning should work again.
- Slightly lowered the damage threshold for embedding.

### DIFF
--- a/code/game/atoms_movable.dm
+++ b/code/game/atoms_movable.dm
@@ -274,6 +274,7 @@
 		hit_atom.hitby(src, TT)
 
 /atom/movable/proc/throw_at(atom/target, range, speed, mob/thrower, spin = TRUE, datum/callback/callback) //If this returns FALSE then callback will not be called.
+
 	. = TRUE
 	if (!target || speed <= 0 || QDELETED(src) || (target.z != src.z))
 		return FALSE
--- a/code/modules/mob/living/carbon/carbon_defense.dm
+++ b/code/modules/mob/living/carbon/carbon_defense.dm
@@ -1,25 +1,6 @@
 
 /mob/living/carbon/standard_weapon_hit_effects(obj/item/I, mob/living/user, var/effective_force, var/hit_zone)
-	if(!effective_force)
-		return 0
-
-	//Apply weapon damage
-	var/damage_flags = I.damage_flags()
-	var/datum/wound/created_wound = apply_damage(effective_force, I.damtype, hit_zone, damage_flags, used_weapon=I, armor_pen=I.armor_penetration)
-
-	//Melee weapon embedded object code.
-	if(istype(created_wound) && I && I.can_embed() && I.damtype == BRUTE && !I.anchored && !is_robot_module(I))
-		var/weapon_sharp = (damage_flags & DAM_SHARP)
-		var/damage = effective_force //just the effective damage used for sorting out embedding, no further damage is applied here
-		damage *= 1 - get_blocked_ratio(hit_zone, I.damtype, I.damage_flags(), I.armor_penetration, I.force)
-
-		//blunt objects should really not be embedding in things unless a huge amount of force is involved
-		var/embed_chance = weapon_sharp? damage/I.w_class : damage/(I.w_class*3)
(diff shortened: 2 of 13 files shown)


code/game/atoms_movable.dm
```diff
@@ -274,6 +274,7 @@
 		hit_atom.hitby(src, TT)
 
 /atom/movable/proc/throw_at(atom/target, range, speed, mob/thrower, spin = TRUE, datum/callback/callback) //If this returns FALSE then callback will not be called.
+
```
out-of-phaze, line 277: ?

code/modules/organs/external/wounds/wound.dm
```diff
@@ -203,7 +206,7 @@
 			return FALSE
 	if(bandaged || clamped)
```
out-of-phaze: could maybe replace the bandaged check here with a !is_open() check, removing the one in the final return

code/game/atoms_movable.dm
MistakeNot4892: What?

out-of-phaze: there's a blank line here

code/modules/mob/living/carbon/carbon_defense.dm
```diff
@@ -1,25 +1,6 @@
...
-		var/embed_threshold = weapon_sharp? 5*I.w_class : 15*I.w_class
-
-		//Sharp objects will always embed if they do enough damage.
-		if((weapon_sharp && damage > (10*I.w_class)) || (damage > embed_threshold && prob(embed_chance)))
-			embed_in_mob(I, hit_zone, damage, I.damtype, supplied_wound = created_wound)
-			I.has_embedded()
-
-	return 1
+	if(effective_force)
+		try_embed_in_mob(I, hit_zone, effective_force, direction = get_dir(user, src))
```
out-of-phaze, line 4: this doesn't call parent or `apply_damage()` which means... i don't think weapons will do damage?

out-of-phaze (changes requested):

code/modules/mob/living/living_defense.dm
```diff
@@ -198,53 +196,56 @@
...
-
-	. = ..()
+/mob/living/proc/try_embed_in_mob(obj/O, def_zone, embed_damage = 0, dtype = BRUTE, datum/wound/supplied_wound, obj/item/organ/external/affecting, direction)
 
-/mob/living/proc/pin_to_wall(var/obj/O, var/direction)
-	if(!istype(O) || O.loc != src || !O.can_embed())//Projectile is suitable for pinning.
-		return
-
-	var/turf/T = near_wall(direction,2)
+	if(!istype(O) || !O.can_embed())
```
out-of-phaze: ok so it seems like only weapons that can embed can deal damage now. should probably do `if(!O.can_embed()) return FALSE` *after* the apply_damage call?

out-of-phaze (changes requested):

code/modules/mob/living/carbon/carbon_defense.dm
MistakeNot4892: It calls apply_damage if no wound is supplied to the embed proc.

out-of-phaze: see my review, before these changes it returned after apply_damage, now it returns before

code/modules/mob/living/living_defense.dm
MistakeNot4892: Fixed ordering.

code/game/atoms_movable.dm
MistakeNot4892: Per Discord it was an accidental inclusion, but it looks nicer than having it crammed up against the header so I'm going to leave it.

out-of-phaze (approved):

PR #4085, opened 2024-06-05
Reduced the smelting temperature to something less batshit insane.

MistakeNot4892:
Fixes https://github.com/ScavStation/ScavStation/issues/992


code/modules/mining/machinery/material_smelter.dm
```diff
@@ -38,7 +38,7 @@
 	var/adjusted_air = FALSE
 	for(var/mtype in reagents?.reagent_volumes)
 		var/decl/material/mat = GET_DECL(mtype)
-		if(mat.boiling_point && temperature >= mat.boiling_point)
+		if(!isnull(mat.boiling_point) && temperature >= mat.boiling_point)
```
out-of-phaze, line 41: zero kelvins is still an invalid boiling point so i don't know if this is a necessary change

out-of-phaze (approved):

code/modules/mining/machinery/material_smelter.dm
MistakeNot4892: It's not invalid, it's just strange. The usual value for 'does not melt' and 'does not boil' is null.

out-of-phaze (approved):

PR #4086, opened 2024-06-05
Tweaking crucible behavior and melting values.

MistakeNot4892:
- Dropped smelting/alloying temperatures to a range that the kiln can hit when burning coal.
- Added slag skimming to crucibles.
- Made crucibles use the storage system.
- Added ability to skip reagents to `trans_to_holder()`.
- Made storage items consolidate stacks on insertion.
- Drastically reduces ore amounts across the board, to increase scarcity and make smelting more manageable.
- Drastically increases crucible capacity so you can do more than 10 units of ore at a time.


code/modules/crafting/metalwork/metalwork_items.dm
```diff
@@ -1,3 +1,11 @@
+/datum/storage/crucible
+	max_w_class = ITEM_SIZE_LARGE
+	max_storage_space = BASE_STORAGE_CAPACITY(ITEM_SIZE_LARGE)
+	can_hold = list(
```
out-of-phaze, line 4: honestly not sure how to feel about this, it was nice to be able to melt down smallish metal objects like spoons, chopsticks, cups, etc.

code/modules/reagents/Chemistry-Holder.dm
```diff
@@ -407,9 +408,21 @@ var/global/obj/temp_reagents_holder = new
...
+			using_volume -= LAZYACCESS(reagent_volumes, rtype)
+		if(using_volume <= 0)
+			return
+		part /= using_volume
+	else
+		part /= total_volume
+
 	. = 0
 	for(var/rtype in reagent_volumes)
+		if(LAZYISIN(skip_reagents, rtype))
```
out-of-phaze: this is O(mn), i wonder if it might be better to remove all entries in skip_reagents from a copy of reagent_volumes first? that would mean it'd be O(n+m), because we loop over m elements of skip_reagents followed by somewhere between n and n-m elements of reagent_volumes

actually judging from some quick wrench tests, you can just do `for(var/rtype in reagent_volumes - skip_reagents)`

code/modules/crafting/metalwork/metalwork_items.dm
```diff
@@ -7,52 +15,45 @@
...
+			if(pouring.standard_pour_into(user, src))
+				return TRUE
 
-		if(transferring)
-			transferring.forceMove(src)
-			visible_message(SPAN_NOTICE("\The [user] drops \the [transferring] into \the [src]."))
+		// Attempting to skim off slag.
+		// TODO: check for appropriate vessel material? Check melting point against temperature of crucible?
+		if(reagents?.total_volume && length(reagents.reagent_volumes) > 1)
+			var/removing = min(amount_per_transfer_from_this, W.reagents.maximum_volume - W.reagents.total_volume)
```
out-of-phaze: i feel like it should use `W.amount_per_transfer_from_this` since we know it's `/obj/item/chems`, i was confused why it kept only transferring 5u
```suggestion
			var/removing = min(W.amount_per_transfer_from_this, REAGENTS_FREE_SPACE(W.reagents))
```

code/modules/crafting/metalwork/metalwork_items.dm
MistakeNot4892: This is replicating the typechecks in attackby, you couldn't put those items in a crucible as far as I recall.

MistakeNot4892: Tested on dev, you can't put small items in the crucible.

code/modules/crafting/metalwork/metalwork_items.dm
MistakeNot4892: We're transferring from the crucible, though, not the bucket. This is consistent with reagent dispenser behavior.

out-of-phaze: hhhhhhuh. i guess, but i would just assume it would use the transfer amount of the thing you're holding... either way i guess it *works*

out-of-phaze (approved):

PR #4088, opened 2024-06-05
Make yellow/green text readable in lightmode

PsyCommando:
## Description of changes
Light text color would be hard to read on the light theme. So made them darker.
Also added a verb to reload only the CSS for the theme at runtime.

## Changelog
:cl:
code: Added verb to reload light/dark mode CSS file. 
config: Made yellow and green text more readable for users of the light mode theme.
/:cl:

### DIFF
--- a/code/modules/admin/admin_verbs.dm
+++ b/code/modules/admin/admin_verbs.dm
@@ -217,7 +217,8 @@ var/global/list/admin_verbs_debug = list(
 	/client/proc/spawn_ore_pile,
 	/datum/admins/proc/force_initialize_weather,
 	/datum/admins/proc/force_weather_state,
-	/datum/admins/proc/force_kill_weather
+	/datum/admins/proc/force_kill_weather,
+	/client/proc/force_reload_theme_css,
 	)
 
 var/global/list/admin_verbs_paranoid_debug = list(
@@ -918,3 +919,13 @@ var/global/list/admin_verbs_mod = list(
 
 		if("Cancel")
 			return
+
+/client/proc/force_reload_theme_css()
+	set category = "Debug"
+	set name     = "Reload UI Theme CSS"
+	set desc     = "Forces the client to reload its UI theme css file."
+	if(!check_rights(R_DEBUG))
+		return
+	var/datum/D = src
+	ReloadModeThemeCss(D.get_client())
+
--- a/code/modules/admin/admin_verbs.dm
+++ b/code/modules/admin/admin_verbs.dm
@@ -926,6 +926,4 @@ var/global/list/admin_verbs_mod = list(
 	set desc     = "Forces the client to reload its UI theme css file."
 	if(!check_rights(R_DEBUG))
 		return
-	var/datum/D = src
-	ReloadModeThemeCss(D.get_client())
-
+	ReloadThemeCss(src)
--- a/code/modules/client/lightmode.css
+++ b/code/modules/client/lightmode.css
@@ -85,7 +85,7 @@ h1.alert, h2.alert		{color: #000000;}
 .info					{color: #0000cc;}
 .notice					{color: #000099;}
 .subtle					{color: #000099; font-size: 75%; font-style: italic;}
-.alium					{color: #00ff00;}
+.alium					{color: #439C16;}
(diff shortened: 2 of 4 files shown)


code/modules/admin/admin_verbs.dm
```diff
@@ -918,3 +919,13 @@ var/global/list/admin_verbs_mod = list(
 
 		if("Cancel")
 			return
+
+/client/proc/force_reload_theme_css()
```
out-of-phaze, line 923: why is this on `/client` and not `/datum/admins` if we're just assuming src is `/datum/admins` anyway??

out-of-phaze (dismissed):

code/modules/client/lightmode.css
```diff
@@ -85,7 +85,7 @@ h1.alert, h2.alert		{color: #000000;}
 .info					{color: #0000cc;}
 .notice					{color: #000099;}
 .subtle					{color: #000099; font-size: 75%; font-style: italic;}
-.alium					{color: #00ff00;}
+.alium					{color: #439C16;}
```
MistakeNot4892: Surprised this isn't failing the CI tests, but I guess they specifically look at DM files. Expectation is usually that hex color values are lowercase due to that being the BYOND standard, if I remember correctly.

code/modules/admin/admin_verbs.dm
PsyCommando: Well, mainly because it's a clientside thing. It doesn't really makes sense to me to put it under the admin datum and end up dereferencing a bunch of vars to get the client, while I could just call it on the client. Idk.

PsyCommando: Also, I'm not assuming that it's being called on /datum/admin. I'm using get_client() cause it's more reliable than just assuming src is a client.

code/modules/client/lightmode.css
PsyCommando: That's a css file. So that's probably why it's not tested. You'd have to check the loaded css string that's injected into the UI to validate that.

PsyCommando: I could probably change it to lowercase if it's really a problem?
Also, another thing I noticed, is that, this reaally shouldn't be assumed to be CSS, because byond's style attribute doesn't support a lot of what css does? Like if you use hsla() to specify a color, it just won't get parsed. It's really a subset of css, and it seems to have it's own standard.

out-of-phaze: hsla was first supported in IE9, byond's internal css handling has partial parity with like. IE6 or something. don't expect things newer than IE7 to work reliably

out-of-phaze: also technically the reason we do lowercase in dm files is because if you do `color = "#FF0000"; color == "#FF0000"` it'll be false because it's internally converted to `"#ff0000"`. not technically necessary for css but it's still a preferred stylistic thing

PsyCommando: Alright, that explains it.

code/modules/admin/admin_verbs.dm
MistakeNot4892: That doesn't make any sense. You could just call `get_client()` without casting if that was a concern, but even then, `get_client()` will just return `src`. This is pretty strange code.

PsyCommando: > That doesn't make any sense. You could just call `get_client()` without casting if that was a concern, but even then, `get_client()` will just return `src`. This is pretty strange code.

Yeah. I think you're right. I probably should have re-read that thing more closely...

code/modules/client/lightmode.css
MistakeNot4892: Happy to approve this if the new colour strings are made lowercase.

PsyCommando: Sounds good, I'll make the change.

PsyCommando: Alright, I made the change, and added some comments to the css files to make it clear it doesn't support the same css subset as everything else. And why the color hex values are lowercase.

MistakeNot4892 (approved):

out-of-phaze (approved):

PR #4089, opened 2024-06-05
Fixed missing storage removal check

PsyCommando:
## Description of changes
In ``/obj/item/handle_mouse_drop`` and ``/obj/item/attack_hand``, the return value of remove_from_storage wouldn't be checked for success before assuming it worked. It doesn't seem to have caused any obvious problems so far, but it would probably make more sense to check it first.

Also, NOTE: I'm a bit unsure if returning to the base proc is the right thing to do on line 355? From what I can see it should be. But, I'm not 100% sure since a lot of things changed with inventories.

### DIFF
--- a/code/game/objects/items/__item.dm
+++ b/code/game/objects/items/__item.dm
@@ -352,7 +352,8 @@
 	if(user.client && istype(inv) && inv.slot_id && (over in user.client.screen))
 		// Remove the item from our bag if necessary.
 		if(istype(loc?.storage))
-			loc.storage.remove_from_storage(user, src)
+			if(!loc.storage.remove_from_storage(user, src))
+				return ..()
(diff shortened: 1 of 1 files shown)


MistakeNot4892 (approved):
I'm not even sure if we have a case where you fail to remove an item from storage, but can't see any harm in it.

PsyCommando:
> I'm not even sure if we have a case where you fail to remove an item from storage, but can't see any harm in it.

Well, quick_empty on storage comes to mind. It doesn't check if it can actually drop anything. So, if the container is locked, it'll empty it's contents regardless. The only reasons it's not happening, is because using the lockable storage in your hands steals self-interacts to open it's lockable UI.

I could have added a check to quick_empty, but it makes more sense to use something more thorough?

PR #4090, opened 2024-06-06
Drake fixes/implementation.

MistakeNot4892:
- Drakes are called 'meredrakes' for the fantasy modpack (as opposed to grafadreka). Their culture entries have also been adjusted to not mention Sif.
- Drakes can no longer use items in their mouth for things like attacking or mining.
- Drakes can only speak their own language.
- Drakes can see in the dark.
- Drakes can join Shaded Hills as cave dweller or traveller.
- Drakes can no longer join as Tradeship Captain.
- Kobaloi can join all Shaded Hills roles except Innkeeper.
- Shaded Hills outfits no longer have vitals sensors.
- Shaded Hills cultures no longer show 'distance to Sol'.


maps/shaded_hills/shaded_hills.dm
```diff
@@ -3,6 +3,7 @@
 	#include "../../mods/content/matchmaking/_matchmaking.dme"
 	#include "../../mods/content/mouse_highlights/_mouse_highlight.dme"
 	#include "../../mods/content/scaling_descriptors.dm"
+	#include "../../mods/species/drakes/_drakes.dme" // include before _fantasy.dme so overrides work
```
out-of-phaze, line 6: could warn if we detect the fantasy modpack has already been loaded? `#ifdef MODPACK_FANTASY #warn Fantasy modpack loaded before Drakes modpack, compatibility features will be missing.` or something

out-of-phaze (dismissed):
wonder if maybe you should do something about sifsap and sifian metabolism too

out-of-phaze (dismissed):

maps/shaded_hills/shaded_hills.dm
MistakeNot4892: I put an ifdef in there but it doesn't seem to be tripped for whatever reason.

out-of-phaze (approved):
conflict resolution, prior review time still valid

maps/shaded_hills/shaded_hills.dm
out-of-phaze: works for me locally if _drakes.dme is included after _fantasy.dme

PR #4103, opened 2024-06-12
Add interaction to switch folding blade tool mode

out-of-phaze:
## Description of changes
Adds an alt interaction to folding blades that lets you open/close the blade, as well as switch tool modes. Replaces current open/close behavior in `attack_self()` when on help intent; when not on help intent you just open/close the blade, to hopefully prevent this from getting in the way of combat.

The code is a little weird, so feel free to ask questions and I'll answer and add additional comments to the code if needed.

Future work, not in this PR (because I can't sprite): Replace the knife icons for tool modes with per-tool-archetype radial menu icons. Would probably be useful elsewhere as well.

Currently there's no way to switch the tool mode with a folding blade. This fixes that.
Also fixes an issue in `code/_onclick/hud/radial.dm` where `maptext_y` was not set. If this PR is closed that needs to be salvaged as its own PR.

## Authorship
Me.

## Changelog
:cl:
add: Added the ability to switch tool mode with folding blades.
/:cl:

### DIFF
--- a/code/game/objects/items/blades/folding.dm
+++ b/code/game/objects/items/blades/folding.dm
@@ -22,16 +22,26 @@
 	update_force()
 
 /obj/item/bladed/folding/attack_self(mob/user)
-	open = !open
+	if(user.a_intent != I_HELP)
+		set_open(!open, user)
+		return TRUE
+	var/decl/interaction_handler/folding_knife/interaction = GET_DECL(/decl/interaction_handler/folding_knife)
+	if(!interaction.is_possible(src, user))
+		return FALSE
+	interaction.invoked(src, user)
+	return TRUE
+
+/obj/item/bladed/folding/proc/set_open(new_state, mob/user)
+	open = new_state
 	update_force()
 	update_icon()
-	if(open)
-		user.visible_message(SPAN_NOTICE("\The [user] opens \the [src]."))
-		playsound(user, 'sound/weapons/flipblade.ogg', 15, 1)
-	else
-		user.visible_message(SPAN_NOTICE("\The [user] closes \the [src]."))
-	add_fingerprint(user)
-	return TRUE
+	if(user)
+		if(open)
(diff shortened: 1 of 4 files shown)


code/game/objects/items/blades/folding.dm
```diff
@@ -74,3 +84,73 @@
...
+		I.appearance_flags = APPEARANCE_UI_IGNORE_ALPHA
+		var/tool_name = tool_archetype.name
+		if(tool_archetype.article)
+			tool_name = "\a [tool_name]"
+		var/image/underlay = new /image
+		underlay.appearance = folding_knife
+		// reset a bunch of appearance vars we don't need
+		underlay.pixel_x = 0
+		underlay.pixel_y = 0
+		underlay.layer = FLOAT_LAYER
```
MistakeNot4892, line 138: May also need to set plane if the knife is beind held and is on the HUD.

MistakeNot4892:
Generally looks fine other than the noted plane setting.

code/game/objects/items/blades/folding.dm
out-of-phaze: radials are always on the hud plane aren't they, so this is fine

MistakeNot4892: By that argument you don't need to set layer either. I guarantee there's going to be some edge case where plane not being set causes some visual malarkey.

out-of-phaze: you do need to set layer, `hud_layerise()` sets the knife's layer to `HUD_ITEM_LAYER` while the radial menu is on `HUD_ABOVE_ITEM_LAYER` and radial button images are on `HUD_ABOVE_HUD_LAYER`. however all of them are already on `HUD_PLANE`.

MistakeNot4892: Alright.

MistakeNot4892 (approved):

PR #4104, opened 2024-06-12
Adds config option to disable swapping pronouns with bodytype in chargen

out-of-phaze:
## Description of changes
Adds a `CISNORMATIVITY` config option to disable setting bodytype in chargen changing the character's pronouns. Sorry if the comment's a bit too pass-agg, happy to change it if needed.

Also fixes hair being reset even if the bodytype has no default accessories when changing bodytype in prefs. Now, we do a more intelligent reset taken from sanitize_character code, to remove accessories that aren't valid on the new bodytype. Default accessories are applied as before.

Allows me to disable this feature on a per-server basis, because I really don't like it and don't view avoiding "confusion" as a good justification for it.

## Authorship
Me.

## Changelog
:cl:
config: added a config option to disable pronouns switching on bodytype selection in character setup
/:cl:

### DIFF
--- a/code/datums/config/_config.dm
+++ b/code/datums/config/_config.dm
@@ -13,7 +13,7 @@
 
 	// Get our actual value (loading from text etc. may change data type)
 	var/decl/config/config_option = GET_DECL(config_decl)
-	if((config_option.config_flags & CONFIG_FLAG_NUM) && istext(new_value))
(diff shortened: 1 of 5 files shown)


MistakeNot4892 (approved):
Comment is a bit extra but heigh ho.

PR #4107, opened 2024-06-15
Move Ninja antagonist and gamemode into a modpack

out-of-phaze:
## Description of changes
Moves the Ninja special role and gamemode into a modpack.
Makes unnamed pAIs use the AI namelist instead of the ninja namelist.

Part of a project to move gamemodes into modpacks to avoid repeats of the Load-Bearing Deity Code incident of 2019 (three dead, 27 injured)


MistakeNot4892 (changes requested):
Per Discord, I would like antags with a runtime-loaded map to be included in away_site_testing.

out-of-phaze:
> Per Discord, I would like antags with a runtime-loaded map to be included in away_site_testing.

fwiw the map is already loaded on all maps that include the modpack, but I can do that, sure

MistakeNot4892 (dismissed):
Per Discord this is tested without being included in away_sites.

out-of-phaze:
> Per Discord this is tested without being included in away_sites.

to be more precise, `/decl/special_role/validate` loads it

MistakeNot4892 (approved):

PR #4109, opened 2024-06-15
Various fixes from dev testing.

MistakeNot4892:


code/game/turfs/floors/natural/_natural.dm
```diff
@@ -112,6 +113,16 @@
...
+	if(reagent_type && height < 0 && reagents?.total_volume < abs(height))
+		try_refill_turf()
+
+/turf/floor/natural/proc/try_refill_turf()
+	set waitfor = FALSE
+	if(turf_is_refilling)
+		return
+	turf_is_refilling = TRUE
+	sleep(0)
+	turf_is_refilling = FALSE
```
out-of-phaze: this will not help, it needs to be unset after the call to `add_to_reagents`.
also i think defer_update = TRUE on add_to_reagents might be better, to be honest? that would prevent the infinite recursion
if you still want to do the debounce solution what if we set turf_is_refilling to world.time and check `if(turf_refill_time >= world.time)` which would limit it to once per tick without having to sleep?

out-of-phaze (changes requested):

out-of-phaze (approved):

code/game/turfs/floors/natural/_natural.dm
MistakeNot4892: Dropped the commit for now, will revisit during the week. It's live on the Crux server though so don't be too confused if you jump on and the behavior is different.

out-of-phaze:
talked it through on discord, readd the commit pls

MistakeNot4892:
It's in #4120

PR #4112, opened 2024-06-16
Repaths humans to /mob/living/human

out-of-phaze:
## Description of changes
Removes `/mob/living/carbon` by repathing `/mob/living/carbon/human` to `/mob/living/human`.

`/mob/living/carbon` presently does nothing, we just kept it so that the human repath diff could be split out into another PR.


out-of-phaze:
firefox freezes when I try to check the diff to see where I accidentally fucked up the includes, so i literally cannot fix CI on this

MistakeNot4892:
Obviated by #4113 so removing salvage tag

PR #4118, opened 2024-06-16
Move Deity gamemode and content into a modpack

### LINKED ISSUE #4115: Deity gamemode needs to be modpacked
Deity needs to be modpacked, and because it relies on wizard and cult code, it has to go first.

out-of-phaze:
## Description of changes
Moves Deity content into a modpack, including:
- `deity_be_near` extension (I fixed a bug with this also)
- deity forms
- deity-specific spells
- deity-specific spell code (`take_charge`, etc) has been modularized
- deity-related chaplain code (holy water, nullrod effects) has been modularized
 - null rods now use an atom-level `nullrod_act()` proc instead of typechecks
 - holy water now uses an `/affect_holy` proc for easy overriding in modpacks

Deity is also no longer included on any playable maps by default, just in the modpack testing map.

Untangles the mess that is Deity from the rest of the codebase. Also has the effect of removing Deity from all maps except modpack testing, because honestly, I'm not even sure if we want to keep it.


MistakeNot4892 (approved):
I hate this code so much.

PR #4124, opened 2024-06-18
Moves Cult into a modpack

out-of-phaze:
## Description of changes
Moves Cult into a modpack.
Makes the Nar'sie godform in the Deity modpack require the Cult modpack.
Moves a bunch of stuff (soulstones, etc.) into the Cult modpack.
Makes a bunch of things have interactions with the Cult modpack.

Cult was and honestly still is tangled through the entire codebase. This is *terrible.* It is outright *miserable.* I kind of feel like, after having seen this code, we should just totally remove cult and start over or something. It's so bad.


MistakeNot4892 (approved):
Given the quality of the code and the fact I couldn't get a single straight answer when I polled Neb's userbase about if anyone had actually played cult on our code, I don't really care what happens to it. Modpacking it at least makes it easier to jettison down the track.

PR #4125, opened 2024-06-19
Remove unnecessary uses of unsafe var access operator

out-of-phaze:
## Description of changes
Removes uses of the unsafe var access operator `:` outside of View Variables, SDQL, lighting, atmos, and Z-mimic.
Also removes an unset global variable, `global.host`, and replaces its one use with the hosted_by config text.

This operator shouldn't be used without a very, very good reason. I also cleaned up some code a little while I was in the area.

### DIFF
--- a/code/game/machinery/computer/shuttle.dm
+++ b/code/game/machinery/computer/shuttle.dm
@@ -9,44 +9,46 @@
 	var/list/authorized = list(  )
 
 
-/obj/machinery/computer/shuttle/attackby(var/obj/item/card/W as obj, var/mob/user as mob)
+/obj/machinery/computer/shuttle/attackby(var/obj/item/card as obj, var/mob/user as mob)
 	if(stat & (BROKEN|NOPOWER))	return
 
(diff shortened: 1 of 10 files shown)


code/game/machinery/computer/shuttle.dm
```diff
@@ -9,44 +9,46 @@
 	var/list/authorized = list(  )
 
 
-/obj/machinery/computer/shuttle/attackby(var/obj/item/card/W as obj, var/mob/user as mob)
+/obj/machinery/computer/shuttle/attackby(var/obj/item/card as obj, var/mob/user as mob)
```
MistakeNot4892, line 12: Might be worth standardizing the params and removing `as obj` `as mob` while you're here.

out-of-phaze: typemaker (opendream) is going to let us lint for those eventually. that's why i didn't remove it

MistakeNot4892 (approved):

PR #4129, opened 2024-06-20
Merging aspects into traits.

MistakeNot4892:
## Description of changes
- Removes `/decl/aspect`.
- Adds aspect behavior/definitions to `/decl/trait`.

Aspects were effectively a superset of traits.

## Authorship
Myself.

## Changelog
Nothing player-facing.


out-of-phaze (dismissed):
`/decl/trait/ascent`, `/decl/trait/utility_frame`, and `/decl/trait/vox` need `abstract_type` set

code/__defines/traits.dm
```diff
@@ -0,0 +1,64 @@
+#define TRAIT_LEVEL_EXISTS   0
+#define TRAIT_LEVEL_MINOR    1
+#define TRAIT_LEVEL_MODERATE 2
+#define TRAIT_LEVEL_MAJOR    3
+
+#define TRAITS_PHYSICAL  BITFLAG(0)
+#define TRAITS_MENTAL    BITFLAG(1)
+
+#define DEFINE_ROBOLIMB_MODEL_TRAITS(MODEL_PATH, MODEL_ID, COST) \
+/decl/trait/prosthetic_limb/left_hand/##MODEL_ID {                                                        \
```
out-of-phaze: the spacing on the slashes here seems a bit excessive, and also they don't even seem to line up?

code/datums/traits/_aspects.dm
out-of-phaze, line 1: empty file?

code/__defines/traits.dm
```diff
@@ -0,0 +1,64 @@
+#define TRAIT_LEVEL_EXISTS   0
+#define TRAIT_LEVEL_MINOR    1
+#define TRAIT_LEVEL_MODERATE 2
+#define TRAIT_LEVEL_MAJOR    3
+
+#define TRAITS_PHYSICAL  BITFLAG(0)
```
out-of-phaze: is this define unused? downstream we use `aspect_flags` for stuff iirc, but it seems that's gone and now we just have `reapply_on_rejuvenation`? those two hardly feel equivalent

out-of-phaze (changes requested):

code/__defines/traits.dm
MistakeNot4892: they were used to micro optimize character prefs and that's it. Removed them in favour of a flag.

code/datums/traits/_aspects.dm
MistakeNot4892: Yeah meant to delete it.

code/__defines/traits.dm
MistakeNot4892: Fixed.

code/__defines/traits.dm
out-of-phaze: i kind of feel that physical/mental traits are an important distinction outside of just reapplying on rejuv though, but I guess i don't have anything concrete. downstream we stored physical aspects on a 'body record' file and mental ones on a 'mind backup' file, but i guess i could just readd that distinction downstream. mostly i just see three categories:
- physical traits that initialize something about your character, like prostheses. if your character were cloned or copied, these would be lost
- physical traits that represent something constitutive about your character, like allergies. if your character were cloned or copied, these would carry over
- mental traits, which are a bit finnicky because you start getting into fuzzy mind/brain stuff like "do you keep mental aspect X if your mind is uploaded into a computer, or is it part of your Brain?"
etc.
i dunno, i guess that'd be a bridge to cross later when we have stuff that actually uses it. are aspects stored in mob snapshots at all?

MistakeNot4892: I don't have any particular objection to categorization but it's not in use currently so I didn't bother keeping it.

Re: snapshots, nope, not currently. Genetic conditions are their own category of thing.

out-of-phaze (approved):

PR #4131, opened 2024-06-20
Fix flashlights not working in space

### LINKED ISSUE #2090: flashlight does not work in space
#### Description of issue
flashlight does not work in space
its turned on:
изображение

flashlight should make light

#### Steps to reproduce
1) go to the space tiles or lattice
2) turn it on

#### Issue bingo

out-of-phaze:
## Description of changes
Fixes #2090. Currently targeting dev because I'm unsure if this change is valid or not, will probably retarget stable if Lohikar says it's okay.

dreamseeker_Vm1T7dr6jR

Fixes #2090.

### DIFF
--- a/code/modules/lighting/lighting_source.dm
+++ b/code/modules/lighting/lighting_source.dm
@@ -358,7 +358,7 @@
 			if ((DETERMINANT(limit_a_x, limit_a_y, test_x, test_y) > 0) || DETERMINANT(test_x, test_y, limit_b_x, limit_b_y) > 0)
 				continue
 
-		if (TURF_IS_DYNAMICALLY_LIT_UNSAFE(T) || T.light_source_solo || T.light_source_multi)
+		if (TURF_IS_DYNAMICALLY_LIT_UNSAFE(T) || T.corners || T.light_source_solo || T.light_source_multi)
 			Tcorners = T.corners
 			if (!T.lighting_corners_initialised)
 				T.lighting_corners_initialised = TRUE


code/modules/lighting/lighting_source.dm
```diff
@@ -358,7 +358,7 @@
 			if ((DETERMINANT(limit_a_x, limit_a_y, test_x, test_y) > 0) || DETERMINANT(test_x, test_y, limit_b_x, limit_b_y) > 0)
 				continue
 
-		if (TURF_IS_DYNAMICALLY_LIT_UNSAFE(T) || T.light_source_solo || T.light_source_multi)
+		if (TURF_IS_DYNAMICALLY_LIT_UNSAFE(T) || T.corners || T.light_source_solo || T.light_source_multi)
```
Lohikar, line 361: I'm pretty sure the correct fix is:
```suggestion
		if (TURF_IS_DYNAMICALLY_LIT_UNSAFE(T) || T.ambient_light || T.light_source_solo || T.light_source_multi)
```

out-of-phaze: corners is wrong, but ambient_light doesn't cover the case where space ambience/starlight is disabled.

MistakeNot4892:
Per Discord, this issue will need to be fixed on stable by Lohikar, so closing this one.

code/modules/lighting/lighting_source.dm
out-of-phaze: wait. no, in that case wouldn't corners be right?

PR #4137, opened 2024-06-22
Converting existing automated movement to use a subsystem.

MistakeNot4892:
## Description of changes
- Adds `SSautomove` to replace `walk_*()` inbuilts (and eventually the direct unmanaged `AStar()` calls that bots are using).
- Adds movement intent to simple animals.
- Renames vars on simple animals (`speed` to `base_movement_delay` and `turns_per_move` to `turns_per_wander`) for clarity.

## TODO
- [x] Get the basic system working.
- ~~Add movement intents to bots.~~ why did I put this on the todo
- [x] Work out why animals can't change to run intent.
- [x] Implement avoidance to replace `walk_away()`.
- [x] Go back over simple_animal values and fold base_movement_delay into movement intent or automove metadata where possible.
- [X] Check simple_animal use of `step()` in case it can be tied to this system.
- ~~*Maybe* get bots using this system as well. Probably something for another PR.~~
- ~~Get meteors working.~~
- [x] Maybe just swap obj back to `walk_to()`.

Condenses various movement values for simple animals and makes them use movement intents properly.

## Authorship
Myself, with some reference to how /tg/ handles their automovement for ideas.

## Changelog
Nothing player-facing.


MistakeNot4892:
Got `walk_to()` and `walk()` functionally replaced, will work on `walk_away()` in the morning.

code/game/objects/objs.dm
```diff
@@ -399,3 +399,6 @@
 	if(obj_flags & OBJ_FLAG_INSULATED_HANDLE)
 		return
 	return ..()
+
+/obj/get_default_automove_controller_type()
+	return /decl/automove_controller/obj
```
MistakeNot4892: Note to self, this is an abstract type, Do Not.

MistakeNot4892:
Current big issue, automove ends after one step. No runtimes, just stops moving and sits there like a lemon.

MistakeNot4892:
- Spiders are working. They hunt you down, attack you, and continue attacking you while visible, and change target if you disappear.
- Deer are working, to an extent. They flee appropriately.

Main thing missing from this implementation versus walk() is that the mob will only ever move in the direction away from you, they won't try to skirt around obstacles at all. I have logic in place that should also try diagonal moves if the cardinal move fails, but it does not work for reasons that escape me. :(

MistakeNot4892:
Added a movement delay reset function that has resolved the issue, all seems more or less equivalent now.

code/datums/movement/automove_controller.dm
```diff
@@ -0,0 +1,63 @@
...
+			else
+				return TRUE // no idea how we would get into this position
+
+		if(mover.SelfMove(target_dir) && (old_loc != mover.loc))
+			return TRUE
+
+		// Try to move around any obstacle.
+		var/static/list/_alt_dir_rot = list(45, -45)
+		for(var/alt_dir in shuffle(_alt_dir_rot))
+			mover.reset_movement_delay()
```
out-of-phaze, line 57: it might be a little more robust to set the delay to what it was prior to the first movement attempt rather than resetting it to the current `world.time` entirely, but i don't feel comfortable asking that to be changed unless the current implementation actually causes issues when tested

out-of-phaze (approved):

code/datums/movement/automove_controller.dm
MistakeNot4892: We would need to do some really invasive lookups to keep track of the previous delay, but I don't expect this to cause issues - it returns early if the mob is already moving, so the capacity to get a bonus move via reset is pretty minimal I think.

out-of-phaze: pretty sure you could just get the movement delay of `mover` and store it in a local var. but yeah, like i said, no real reason to do it atm

PR #4141, opened 2024-06-24
Replace click-level dexterity check with more granular checks

out-of-phaze:
## Description of changes
Replaces click-level dexterity checks with a bunch of granular checks in different spots.
Also rewrites structure window (de)construction code, currently untested.

Drakes were unable to do a lot of really basic things like place things on tables or in bags. This should fix that. I would rather we be too permissive than too restrictive when it comes to dexterity, to be honest


MistakeNot4892 (approved):
Per Discord I'm really leery of the 'fix it as issues arise' approach for this since drakes are already going to get side-eyed (speaking from how things went on Polaris) but I suppose I am being paranoid.

I would like checks that they can't arm grenades or use shotguns with their mouths though.

MistakeNot4892 (changes requested):
Hit the wrong button.

MistakeNot4892:
Per Discord, I pulled this down and tested for ten minutes. Drakes can:
- cock a shotgun
- change gun firemodes
- toggle gun safeties
- use hand teleporters
 - interact with bomb assemblies
- drill through walls

This is a small handful of the potentially problematic interactions. I am not in favour of this change generally.

out-of-phaze:
> Per Discord, I pulled this down and tested for ten minutes. Drakes can:
> 
>    * ~~cock a shotgun~~
> 
>    * ~~change gun firemodes~~
> 
>    * ~~toggle gun safeties~~
> 
>    * use hand teleporters
> 
>    * interact with bomb assemblies
> 
>    * ~~drill through walls~~
> 
> 
> This is a small handful of the potentially problematic interactions. I am not in favour of this change generally.

the two non-crossed-out ones shouldn't be caused by this as far as i know. they use `attack_self()` which doesn't necessarily go through `Click()` at all (which is a good example of why putting the check in `Click()` is kind of a bad stopgap). i'll include them with the fixes in this PR but i'm going to test on dev to confirm

MistakeNot4892 (approved):
Approving this mostly because it's not a fight I'm particularly invested in. I am predicting we're going to have some absurd edge cases pop up down the track though.

out-of-phaze:
> Approving this mostly because it's not a fight I'm particularly invested in. I am predicting we're going to have some absurd edge cases pop up down the track though.

even with the click-level check in, there were a bunch of edge cases with tool crafting and such anyway, i'm pretty sure. ran into them in testing, i was also able to pick up someone's tool crafting holder and disassemble it which was funny

PR #4151, opened 2024-07-01
Fix for shuttle ceiling creation

NataKilar:
## Description of changes
Fixes shuttle ceiling creation in the case that the shuttle is landing directly below where it previously was.

## Authorship
Myself. Thanks to Penny for the report.

### DIFF
--- a/code/modules/shuttles/shuttle.dm
+++ b/code/modules/shuttles/shuttle.dm
@@ -340,9 +340,9 @@
 		if(force || (istype(TAD, get_base_turf_by_area(TAD)) || TAD.is_open()))
 			TAD.ChangeTurf(ceiling_type, TRUE, TRUE, TRUE)
 
-		// TODO: Ideally the second check here would remain is_open() rather than a direct type check, but unfortunately we can't do that with only the path.
+		// TODO: Ideally the latter checks here would remain is_open() rather than direct type checks, but unfortunately we can't do that with only the path.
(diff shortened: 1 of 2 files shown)


code/modules/shuttles/shuttle.dm
```diff
@@ -328,12 +331,21 @@
...
+			continue
+
+
 		if(force || (istype(TAD, get_base_turf_by_area(TAD)) || TAD.is_open()))
-			// Check for multi-z shuttles. Don't create a ceiling where the shuttle is about to be.
-			if((istype(TAS) && (get_area(TAS) in shuttle_area)))
-				continue
 			TAD.ChangeTurf(ceiling_type, TRUE, TRUE, TRUE)
 
+		// TODO: Ideally the second check here would remain is_open() rather than a direct type check, but unfortunately we can't do that with only the path.
```
out-of-phaze: it excludes space, which is probably a pretty big deal. the base-turf-by-area check miiiight be handling it in some cases? but like. definitely need to account for both `/turf/space` and `/turf/open`.

out-of-phaze (changes requested):

code/modules/shuttles/shuttle.dm
NataKilar: Addressed

out-of-phaze (approved):

PR #4154, opened 2024-07-04
Change non-furniture wood on Shaded Hills to walnut, tweak lanterns

out-of-phaze:
## Description of changes
- Consolidates some wooden door types to reduce duplication.
- Fixes a blank tool entry when examining certain bladed tools.
- Adds wood type variants of railings, and ebony/walnut compost bins.
- Reduces the lantern max fuel amount to 60, but reduces fuel consumption to 1/3 of the current amount. This means that it'll last an hour from full to empty.
- Makes wall sconces start off unlit, so that the lights don't all go out at once 40 minutes in. It can be changed per sconce, for mapping flexibility.
- Replaces non-furniture uses of ebony on Shaded Hills with walnut. This was supposed to be done a while back but I kept forgetting to include the mapping commit.

Makes lanterns less of a pain to deal with.
Fixes compost bins being generic wood.
Adds more wood variants of objects and removes duplicate ones.
Makes the map less oppressively dark.

## Authorship
Me.

## Changelog
:cl:
tweak: Ebony walls, floors, doors, and railings on Shaded Hills have all been replaced with walnut variants.
tweak: Most wall sconces now start unlit.
balance: A full lantern now burns for an hour, but only holds 60 units of fuel.
/:cl:

### DIFF
--- a/code/game/objects/items/blades/_blade.dm
+++ b/code/game/objects/items/blades/_blade.dm
@@ -65,7 +65,7 @@
 		)
 		switch(w_class)
 			if(0 to ITEM_SIZE_SMALL)
-				initial_tool_qualities[TOOL_HATCHET] = TOOL_QUALITY_NONE
+				pass()
 			if(ITEM_SIZE_SMALL to ITEM_SIZE_NORMAL) // Since ITEM_SIZE_SMALL was already covered, this is just ITEM_SIZE_NORMAL.
 				initial_tool_qualities[TOOL_HATCHET] = TOOL_QUALITY_WORST
 			if(ITEM_SIZE_NORMAL to ITEM_SIZE_LARGE)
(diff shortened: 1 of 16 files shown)


code/game/objects/items/blades/_blade.dm
```diff
@@ -65,7 +65,7 @@
 		)
 		switch(w_class)
 			if(0 to ITEM_SIZE_SMALL)
-				initial_tool_qualities[TOOL_HATCHET] = TOOL_QUALITY_NONE
+				pass()
```
MistakeNot4892, line 68: Strictly speaking I think you can just snip this case out and remove ITEM_SIZE_SMALL from the second case.

MistakeNot4892 (dismissed):

code/game/objects/items/blades/_blade.dm
out-of-phaze: there's an `else`

MistakeNot4892 (approved):

PR #4156, opened 2024-07-04
Fix wineskin not filling with wine

### LINKED ISSUE #4155: Beggar Knight wineskin lacks wine
#### Description of issue
The wineskin that Beggar Knights start with doesn't have wine.

It should have wine.

#### Steps to reproduce
1. Spawn in as a Beggar Knight on Shaded Hills.
2. Try to sip from your wineskin
3. Smack yourself in the face instead.

N/A

Since the wineskin was added.

N/A

#### Issue bingo

out-of-phaze:
## Description of changes
Changes `reagents?.total_volume` to be `reagents?.maximum_volume` instead.

The wineskin will now fill with wine. Fixes #4155.

### DIFF
--- a/code/game/objects/items/waterskin.dm
+++ b/code/game/objects/items/waterskin.dm
@@ -17,4 +17,4 @@
 
 /obj/item/chems/waterskin/crafted/wine/populate_reagents()
 	. = ..()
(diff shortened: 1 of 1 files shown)


MistakeNot4892 (approved):
Thanks for that.

PR #4157, opened 2024-07-04
Replace a number of hooks with observation decls, remove deprecated hooks

out-of-phaze:
## Description of changes
Replaces `/hook/submap_join`, `/hook/player_latejoin`, `/hook/sell_crate`, `/hook/terminate_employee`, `/hook/reassign_employee`, `/hook/change_account_status`, `/hook/revoke_payroll`, `/hook/borgify`, and `/hook/debrain` with corresponding observation decls.
Removes `/hook/death` because it handled on-species death effects which are now handled elsewhere.
Removes `/hook/clone` because we don't have cloning.
Makes auxtools debugging not use hooks.

Hooks are an old and gross system and I want to deprecate them in favour of observation events. hooks-def.dm was one of the oldest files in our codebase before this.

### DIFF
--- a/code/datums/observation/debrain.dm
+++ b/code/datums/observation/debrain.dm
@@ -1,0 +1,13 @@
+/**
+ * Observer Pattern Implementation: Debrained
+ *
+ * Raised when: A brainmob is created by the removal of a brain.
+ *
+ * Arguments that the called proc should expect:
+ *     /mob/living/brainmob: The brainmob that was created.
+ *     /obj/item/organ/internal/brain: The brain that was removed.
+ *     /mob/living/owner: The mob the brain was formerly installed in.
+ */
+/decl/observ/debrain
+	name = "Debrained"
+	expected_type = /mob/living
--- a/code/_helpers/auxtools.dm
+++ b/code/_helpers/auxtools.dm
@@ -9,13 +9,21 @@ var/global/auxtools_debug_server = world.GetConfig("env", "AUXTOOLS_DEBUG_DLL")
 /proc/enable_debugging(mode, port)
 	CRASH("auxtools not loaded")
 
-/hook/startup/proc/auxtools_init()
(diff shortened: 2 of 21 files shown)


code/datums/observation/debrain.dm
```diff
@@ -0,0 +1,13 @@
...
+ * Observer Pattern Implementation: Debrained
+ *
+ * Raised when: A brainmob is created by the removal of a brain.
+ *
+ * Arguments that the called proc should expect:
+ *     /mob/living/brainmob: The brainmob that was created.
+ *     /obj/item/organ/internal/brain: The brain that was removed.
+ *     /mob/living/owner: The mob the brain was formerly installed in.
+ */
+/decl/observ/debrain
```
MistakeNot4892, line 11: Is it really worth keeping the ones that are just flat out not used for anything?

MistakeNot4892:
Changes generally look fine, but honestly I would rather we just blow away all these extraneous events/hooks that have no associated logic.

code/datums/observation/debrain.dm
out-of-phaze: i have no clue which ones are used on downstreams, iirc several are used on hearth? and psy was very concerned about removing hooks because outreach uses them
```
/hook/startup/proc/tgs_on_world_start()
	world.TgsNew()
	return TRUE

/hook/global_init/proc/tgs_on_init()
	world.TgsInitializationComplete()
	return TRUE

/hook/shutdown/proc/tgs_on_shutdown()
	world.TgsReboot()
	return TRUE
```
as an example of just the tgs ones

MistakeNot4892:
Per Discord it's unclear what is in current use so happy to merge this as-is and revisit cleaning up unneeded observs later.

MistakeNot4892 (approved):

PR #4159, opened 2024-07-05
Fix code that indexes datum-typed variables

out-of-phaze:
## Description of changes
Fixes `datum[foo]` which is going to be a runtime error in 516.

Two of these were due to type-abused vars. One was an actual issue (circuits code) which would never work due to bad copypasta.

### DIFF
--- a/code/modules/integrated_electronics/core/helpers.dm
+++ b/code/modules/integrated_electronics/core/helpers.dm
@@ -24,9 +24,10 @@
 
 /obj/item/integrated_circuit/proc/set_pin_data(pin_type, pin_number, datum/new_data)
 	if(islist(new_data))
+		var/list/new_list = new_data
 		for(var/i in 1 to length(new_data))
-			if (istype(new_data) && !isweakref(new_data))
-				new_data[i] = weakref(new_data[i])
+			if (istype(new_list[i], /datum) && !isweakref(new_list[i]))
+				new_list[i] = weakref(new_list[i])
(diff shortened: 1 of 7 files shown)


code/modules/integrated_electronics/core/helpers.dm
```diff
@@ -24,9 +24,10 @@
 
 /obj/item/integrated_circuit/proc/set_pin_data(pin_type, pin_number, datum/new_data)
 	if(islist(new_data))
+		var/datum/new_list = new_data
 		for(var/i in 1 to length(new_data))
-			if (istype(new_data) && !isweakref(new_data))
-				new_data[i] = weakref(new_data[i])
+			if (istype(new_list[i], /datum) && !isweakref(new_list[i]))
+				new_data[i] = weakref(new_list[i])
```
MistakeNot4892: Still appears to be failing on these ones - looks like new_list is mistyped?

code/modules/integrated_electronics/core/helpers.dm
```diff
@@ -24,9 +24,10 @@
 
 /obj/item/integrated_circuit/proc/set_pin_data(pin_type, pin_number, datum/new_data)
 	if(islist(new_data))
+		var/datum/new_list = new_data
```
MistakeNot4892:
```suggestion
		var/list/new_list = new_data
```

MistakeNot4892 (approved):

PR #4160, opened 2024-07-05
Clean up topic commands to use UIDs instead of a global list

out-of-phaze:
## Description of changes
Cleans up topic commands. They now use abstract_type and UIDs instead of a bespoke `base_type` var and a global list of names -> commands.

Might be a good idea to go back and remove the `name` var or have it auto-set UIDs (if we allow that?) or whatever.

Improves the code quality of topic commands, which hadn't been touched in like three years.

### DIFF
--- a/code/game/world.dm
+++ b/code/game/world.dm
@@ -121,11 +121,13 @@ var/global/world_topic_last = world.timeofday
 	if(!length(params))
 		return
 	var/command_key = params[1]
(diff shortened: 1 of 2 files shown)


MistakeNot4892:
I was sure we had decls that autoset UID somewhere, but my regex-fu cannot find them. Seems fine conceptually so long as it's consistent.

MistakeNot4892 (approved):

PR #4168, opened 2024-07-06
Prevent space from freezing things via gas conduction

### LINKED ISSUE #4163: Vacuum makes things cold
#### Description of issue
Vacuum makes things cold because there's no difference between 2000kPa of gas at 2.7K and 0kPa of gas at 2.7K.

0mol of gas shouldn't cool anything down.

#### Steps to reproduce
1. Try to pick up an item in vacuum, like stuff in the magshield away sites.
2. Get burnt by the cold.

N/A

Since local environment processing was added to ProcessAtomTemperature, I guess?

Latest dev

#### Issue bingo

out-of-phaze:
## Description of changes
Heat transfer rate from local air to atoms now scales with air pressure.

Fixes #4163.

Currently, vacuum freezes things because it's at 2.7K even though there's no gas to transfer heat to. Now, at 0kPa no heat transfer will occur, and at 10kPa it'll equalize ten times as quickly (though maybe it'd be better to cap it at a max of 1x?)

In the future we can add radiation separately, but I would much rather have the issue of hot things not cooling down (since everything that does that presently has it hardcoded separately) than *everything* becoming ice-cold no matter what. This is gamebreaking, a lack of radiative cooling is relatively small.

## Authorship
Me.

### DIFF
--- a/code/game/atoms_temperature.dm
+++ b/code/game/atoms_temperature.dm
@@ -32,18 +32,19 @@
 	// Get our location temperature if possible.
 	// Nullspace is room temperature, clearly.
 	var/adjust_temp = T20C
+	var/thermal_mass_coefficient = get_thermal_mass_coefficient()
 	if(isturf(loc))
 		var/turf/T = loc
 		var/datum/gas_mixture/environment = T.return_air()
-		if(environment)
-			adjust_temp = environment.temperature
+		adjust_temp = environment.temperature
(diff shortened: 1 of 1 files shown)


MistakeNot4892 (approved):
Would like to revisit this if we get issues where post-fire situations are full of white-hot metal rods or whatever, but in the meantime this seems unlikely to cause serious issues.

PR #4171, opened 2024-07-06
Fix smelter being unable to melt graphite

### LINKED ISSUE #4161: Smelter cannot process graphite
#### Description of issue
Graphite melts at 4074K, but the smelter now only processes ores at 1000C. That means it can't process graphite.

You should be able to process graphite in the smelter so that you can alloy it.

#### Steps to reproduce
1. Try to process graphite in a smelter, making sure the smelter is on and has been heating up for a while.
2. The graphite pops out on the other side unchanged.

N/A

Since the smelting temperature was lowered.

#### Issue bingo

out-of-phaze:
## Description of changes
Increases the high smelting temperature to 4000C (4273.15K), enough to melt graphite (melting point of 4074K).

Fixes #4161. I understand that before it was set way too high, but like, it does need to be hot enough to melt graphite...

## Authorship
Me.

### DIFF
--- a/code/__defines/research.dm
+++ b/code/__defines/research.dm
@@ -20,7 +20,7 @@
 #define BASE_OBJECT_MATTER_MULTPLIER    0.25
 
 #define GENERIC_SMELTING_HEAT_POINT 1000 CELSIUS
-#define HIGH_SMELTING_HEAT_POINT    1500 CELSIUS
+#define HIGH_SMELTING_HEAT_POINT    4000 CELSIUS // must be at least 4074K (3800 C) to melt graphite
 #define TECH_MATERIAL      "materials"
 #define TECH_ENGINEERING   "engineering"
 #define TECH_EXOTIC_MATTER "exoticmatter"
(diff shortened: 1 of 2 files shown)


MistakeNot4892:
Per Discord I'm not super sure about this one. Scav had issues where they put stuff into the smelter and ended up with like, white-hot vaporized metal in the atmosphere. Smelting graphite into plastic is pretty absurd as-is, if there was a chemical process we could use to replace it I think that would probably be better - maybe a plastic precursor synthesised from powdered graphite, similar to the one you find in the botany mushrooms? 

I guess the other alternative would be checking that none of the common ore/mineral materials vaporize at 4kK, or just removing that aspect of the smeltery entirely.

out-of-phaze:
> Per Discord I'm not super sure about this one. Scav had issues where they put stuff into the smelter and ended up with like, white-hot vaporized metal in the atmosphere. Smelting graphite into plastic is pretty absurd as-is, if there was a chemical process we could use to replace it I think that would probably be better - maybe a plastic precursor synthesised from powdered graphite, similar to the one you find in the botany mushrooms?
> 
> I guess the other alternative would be checking that none of the common ore/mineral materials vaporize at 4kK, or just removing that aspect of the smeltery entirely.

you need to smelt graphite to alloy it with iron and make steel, it's not the compressor, which is what turns graphite into plastic. i'd be fine with removing it personally, or limiting it to just water vapor so that you get punished for just shoving ice in there

MistakeNot4892 (approved):

PR #4179, opened 2024-07-09
Excepts simple_animals from weather effects.

### LINKED ISSUE #4153: Wildlife on Shaded Hills die of hail
#### Description of issue
Hail kills all the wildlife.

Hail should not kill all the wildlife. Either they should be immune, or they should run for cover.

#### Steps to reproduce
1. Cover yourself in oil.
2. Wait for it to hail.
3. Ow ow ow ow ow
4. All the animals are dead.

N/A

Since wildlife and weather both coexisted on this map.

Latest dev.

#### Issue bingo

MistakeNot4892:
Should fix #4153

### DIFF
--- a/code/modules/mob/living/life.dm
+++ b/code/modules/mob/living/life.dm
@@ -229,41 +229,62 @@
 	if(!.) // If we're under or inside shelter, use the z-level rain (for ambience)
 		. = SSweather.weather_by_z[my_turf.z]
 
-/mob/living/proc/handle_environment(var/datum/gas_mixture/environment)
+/* Accidentally copied from dev - leaving commented for easier merge resolution when this commit goes back up to dev.
+/mob/living/proc/handle_contact_reagent_dripping()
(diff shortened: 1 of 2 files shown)


out-of-phaze (approved):
contact reagent dripping seems undocumented but i'm not opposed

MistakeNot4892:
Undocumented how? I just moved the logic.

out-of-phaze (changes requested):
> Undocumented how? I just moved the logic.

it's only in the added part of the diff, so i can't tell where it was moved from. did you not remove the old copy or something?

oh lol this pr targets stable but the code you're moving is only on dev. think this should get retargeted to dev then

MistakeNot4892:
Oh whoops.

MistakeNot4892:
I wanted this fix to go to stable, but I've commented out the contact reagent stuff - will make it easier to resolve when I pull the commit into dev.

out-of-phaze (approved):

PR #4182, opened 2024-07-10
Added a mild toxin to replace bromide in food.

MistakeNot4892:
## Description of changes
- Adds acrylamide, a mild toxin, to replace the weird use of bromide in foods following the removal of generic `/decl/material/liquid/toxin` ages ago.
- Swaps meat/protein for acrylamide in burned recipes.
- Swaps bromide for ammonia in raw organs.

Generally makes more sense, patches over an unintended or undocumented change to cause burned recipes to have meat, which let them be farmed to make meatballs.

## Authorship
Myself.

## Changelog
:cl:
tweak: Burned recipes no longer contain meat, burned foods contain acrylamide, and organs now contain ammonia instead of bromide.
/:cl:


out-of-phaze (approved):
feels like the new chems should go to dev, but on the other hand this is fixing burnt food magic meatballs... stable will do

Issue #4147, opened 2024-06-27
Manual heating of atoms is too quick.


MistakeNot4892:
Copied from https://github.com/ScavStation/ScavStation/issues/1007

#### Description of issue
I found out that touching any container with alcohol in it once with a lit cigar (and presumably cigs and really any fiery heat source you can light like welders too) is enough to instantly get it hot enough to boil it into ethanol, I assume this works with other heat dependent reactions but I didn't test those 

#### Difference between expected and actual behavior
I am used to watching skozzy having to spam clicks to get the same effect so it was a bit strange to me I was able to do it with just one tap and with a lit cigar on the keg of beer of all things

#### Steps to reproduce
get some booze container
light a cigar or any other smokeable object
click container with your lit cigar or whatever

#### Specific information for locating

cigar or other smokeable object (possibly lit welders too) any container with alcohol in it (also possibly milk with enzyme, space drugs production, and other such heat dependent reactions) 

#### Length of time in which bug has been known to occur

found today on 6/13/2024

#### Client version, Server revision & Game ID

Date
2024-06-14
Branch
main
Game ID
cud-aU3B
Runtime Version
514.1589
Compiler Version
514.1589
Revision
0342fd8a2ec437311ac7fb9bb9b727503d129072

#### Issue bingo

- [ ] Issue could be reproduced at least once
- [ ] Issue could be reproduced by different players
- [ ] Issue could be reproduced in multiple rounds
- [x] Issue happened in a recent (less than 7 days ago) round
- [x] [Couldn't find an existing issue about this](https://github.com/ScavStation/ScavStation/issues)

PR #4183, opened 2024-07-10
Smaller atoms now manually heat target atoms slower or faster based on size.

MistakeNot4892:
## Description of changes
Adds procs to modify the effective heat transfer when 'carefully heating' atoms with things like welding torches.

Addresses #4147.

## Authorship
Myself.

## Changelog
:cl:
tweak: Cigarettes and welding torches will now heat atom reagents slower.
/:cl:

### DIFF
--- a/code/game/atoms_movable.dm
+++ b/code/game/atoms_movable.dm
@@ -490,6 +490,9 @@
 /atom/movable/proc/get_object_size()
 	return ITEM_SIZE_NORMAL
 
+/atom/movable/get_manual_heat_source_coefficient()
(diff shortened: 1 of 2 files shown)


out-of-phaze (approved):
as mentioned on Discord, i think part of the issue with cigarettes/cigars/smokables is that they return an obnoxiously high temperature of 1500K. this change is fine but I think we should probably change the heat value for smokables as well, if you're not going to then I can get to it later

MistakeNot4892:
Yeah I probably will not get to it.

PR #4186, opened 2024-07-11
Optimize initialization and maploading

out-of-phaze:
## Description of changes
Performs a lot of optimizations across the codebase, some more experimental than others. I tested and got some pretty impressive results downstream on Spacefarers (loading in a new ship after roundstart took about 4 seconds instead of 16) and it seems like a lot of trouble spots in initialization are faster too.

The commit messages are descriptive, I would suggest referencing those for a description of what's included in this PR. Of particular note are these changes:
- ~~/turf/unsimulated no longer goes through Initialize~~ Separated into #4386 
- ~~/obj/effect/floor_decal no longer goes through Initialize or Destroy~~ Separated into #4387 
- update_alpha_mask now sleeps until, at minimum, the end of the tick (instead of only sleeping if the server was overloaded)
- A bunch of event registrations are now *explicitly* cleaned up in Destroy(), to avoid costly event cleanup code.
- Unregistering from the Moved event also unregisters from its parent. ~~(I'm unsure about this one but I couldn't see any other way to fix a bunch of hanging `recursive_move` event registrations.)~~ Seems to work like a charm.
- A few things were changed to respect level data bounds, which is helpful to optimize smaller away sites in large worlds.
- SSambience avoids duplicate queueing and avoids costly removal attempts when we know the turf is not in the queue.
- `initial_gas = list(/decl/material/gas/oxygen = MOLES_O2STANDARD, /decl/material/gas/nitrogen = MOLES_N2STANDARD)` now has a special hot-path: `initial_gas = GAS_STANDARD_AIRMIX`. This avoids creating a list per turf (including a unsimulated fillter turfs!), which saves memory and init time.
- Turf height appearance icons are now cached.
- Unsimulated atoms skip visualnet updates, which makes ChangeTurf() and Destroy() even faster for them.
- Extra argument handling in `qdel()` after `force` was removed. Nothing used this and it was a waste of processor time.

Lots and lots of optimizations.

## Authorship
One particular commit, `Port InitAtom optimizations from tgstation`, is adapted from some of the changes made by Mothblocks on tgstation in [this commit](https://github.com/tgstation/tgstation/commit/f338a209d7c1484da6a92c66a7ede04c2b8d0404). The rest was all done by me, mostly for the downstream *Spacefarers.*


out-of-phaze:
image
hell yeah

code/game/turfs/unsimulated.dm
```diff
@@ -1,16 +1,26 @@
...
-	)
+	initial_gas = GAS_STANDARD_AIRMIX
 	abstract_type = /turf/unsimulated
 	simulated = FALSE
 
 // Shortcut a bunch of simulation stuff since this turf just needs to sit there.
-/turf/unsimulated/Initialize()
-	SHOULD_CALL_PARENT(FALSE)
+// We don't even call Initialize(), how cool is that???
+/turf/unsimulated/New()
```
MistakeNot4892: Per Discord, I don't like going back to using New() on atoms, would like a justification there. I'm also unsure how overriding Initialize entirely is going to interact with things like unsimulated turfs in dynamically lit areas.

MistakeNot4892: Discussed on Discord - needed to avoid overhead in InitAtom. Really don't like it but not for any sane reason that I can articulate, other than tidiness.

MistakeNot4892: Unsim turfs also seem to work fine with dynamic lighting post-change, which is spooky but fine I guess.

code/game/turfs/unsimulated/beach.dm
```diff
@@ -18,6 +18,6 @@
 	icon_state = "water"
 	turf_flags = TURF_FLAG_BACKGROUND | TURF_IS_WET | TURF_IS_HOLOMAP_PATH
 
-/turf/unsimulated/beach/water/Initialize(var/ml)
-	. = ..()
+/turf/unsimulated/beach/water/New()
+	..()
 	overlays += image("icon"='icons/misc/beach.dmi',"icon_state"="water2","layer"=MOB_LAYER+0.1)
```
MistakeNot4892: Augh. At some point I hope we can forbid direct overlays modification.

code/modules/fabrication/designs/_design.dm
```diff
@@ -29,13 +29,7 @@
...
 	if(required_technology == TRUE)
 		if(ispath(path, /obj/item))
-			var/list/res = build(null, new/datum/fabricator_build_order(src, 1))
-			if(length(res))
-				var/obj/item/O = res[1]
-				var/tech = O.get_origin_tech()
-				if(tech)
-					required_technology = cached_json_decode(tech)
-				QDEL_NULL_LIST(res)
+			required_technology = cached_json_decode(atom_info_repository.get_origin_tech_for(path, amount = 1))
```
MistakeNot4892: Why not cache the decoded origin tech list on the atom info repo?

out-of-phaze: huh. good point, hadn't thought of that

code/unit_tests/turf_icons.dm
```diff
@@ -55,8 +55,8 @@
 		SSair.tiles_to_update    -= floor
 		SSairflow.processing     -= floor
 		SSoverlays.processing    -= floor
-		SSambience.queued        -= floor
 		SSicon_update.queue_refs -= floor
+		AMBIENCE_UNQUEUE_TURF(floor)
```
MistakeNot4892: Noting from earlier that this should probably be DEQUEUE rather than UNQUEUE.

MistakeNot4892:
Generally the changes look good. I'm skittish about the New() stuff as discussed but I think that's mostly just inertia, not a serious technical objection. It might be nice to get a quorum or at least get @Lohikar's opinion on the use of EarlyDestroy and skipping of InitAtom if only for future reference.

out-of-phaze:
decals don't work because of layer stuff... should maybe update decal layer on turf init or something.

out-of-phaze:
decals are fixed, had to modify is_plating() a bit. also fixed an unreported issue where decals on natural turfs didn't work—if they're not supposed to that should probably be explicitly handled

code/controllers/subsystems/zcopy.dm
```diff
@@ -198,7 +198,8 @@ SUBSYSTEM_DEF(zcopy)
 	calculate_zstack_limits()
 
 	for (var/zlev in 1 to world.maxz)
-		for (var/turf/T in block(locate(1, 1, zlev), locate(world.maxx, world.maxy, zlev)))
+		var/datum/level_data/level = SSmapping.levels_by_z[zlev]
+		for (var/turf/T as anything in block(level.level_inner_min_x, level.level_inner_min_y, zlev, level.level_inner_max_x, level.level_inner_max_y))
```
Lohikar, line 202: Is it impossible for turfs to be placed outside of this? This is supposed to be a universal reset, though it's explicitly unsupported anyways.

code/game/turfs/flooring/_flooring_decals.dm
```diff
@@ -13,12 +13,18 @@ var/global/list/floor_decals = list()
...
-// Have to wait for turfs to set up their flooring, so we can better guess at our layers.
-/obj/effect/floor_decal/Initialize()
-	..()
-	return INITIALIZE_HINT_LATELOAD
+/obj/effect/floor_decal/Initialize(mapload)
+	SHOULD_CALL_PARENT(FALSE)
+	SHOULD_NOT_OVERRIDE(TRUE)
+	return INITIALIZE_HINT_NORMAL
+
+/obj/effect/floor_decal/New(loc, var/newdir, var/newcolour, var/newappearance)
```
Lohikar: I don't like this being `New()`, there shouldn't be enough of these for the overhead saved here to be remotely meaningful.

I barely justify merging atom/New() into turf/New() on aurora.

code/game/turfs/unsimulated/beach.dm
```diff
@@ -18,6 +18,6 @@
 	icon_state = "water"
 	turf_flags = TURF_FLAG_BACKGROUND | TURF_IS_WET | TURF_IS_HOLOMAP_PATH
 
-/turf/unsimulated/beach/water/Initialize(var/ml)
-	. = ..()
-	overlays += image("icon"='icons/misc/beach.dmi',"icon_state"="water2","layer"=MOB_LAYER+0.1)
+/turf/unsimulated/beach/water/New()
+	..()
+	add_overlay(image("icon"='icons/misc/beach.dmi',"icon_state"="water2","layer"=MOB_LAYER+0.1))
```
Lohikar: `add_overlay()` in a turf New is unsafe, they can be instanced before subsystems are iirc.

code/controllers/subsystems/zcopy.dm
out-of-phaze: technically turfs outside of it exist, but should not participate in sim _at all_. it is currently invalid for unsim filler turfs outside level bounds to have any sort of behavior especially ZM behavior. i guess at some point if we get a virtual-Z system a la shiptest it could be, but we'd probably still have a level data datum associated with it even then

code/game/turfs/unsimulated/beach.dm
out-of-phaze: according to [TG init order documentation](https://github.com/tgstation/tgstation/blob/470cff519e57263dc0f7adcbbf04febd90de0088/code/game/world.dm#L25-L26) this is incorrect, `Master` is instantiated via a static var initializer, which runs before mapped atom New(). `/datum/controller/master/New()` is what instantiates the subsystems. this will always happen before mapped atoms are instantiated.

code/game/turfs/flooring/_flooring_decals.dm
out-of-phaze: for post-roundstart maploading like we do downstream it's actually pretty significant, non-mapload initatoms is pretty damn slow and ships tend to have a bunch of decals. normalized so that both traces have the same number of calls, it saved 87.25%
image
also consider lush exoplanets; `/datum/random_map/noise/exoplanet/proc/spawn_grass` spawns a *lot* of floor decals. i tested 3 lush exoplanets and compared with just an optimized `/Initialize` compared to `/New`, and with `/New`, `spawn_grass` is 71.92% cheaper.
image

Lohikar: 1 second of init time is nothing, I don't really care about the percentage improvement if the absolute init time is already very fast.

out-of-phaze: it's kind of hard to profile this because I can't exactly profile the missing InitAtom calls due to this change, but i think it's pretty significant. i'll profile against a larger map like Tether at some point

out-of-phaze: @Lohikar i believe this particular change was an 8 second (out of ~64 seconds) time save on Tether, which is pretty significant imo. also with this PR as a whole Shaded Hills went from around 42 seconds to 19 seconds in local testing, though idk if this particular change is as relevant there

code/game/turfs/floors/natural/_natural.dm
```diff
@@ -35,6 +35,9 @@
 /turf/floor/natural/get_plant_growth_rate()
 	return 0.1
 
+/turf/floor/natural/is_plating()
```
MistakeNot4892: Noting from Pyrelight testing that this prevents building log/stone walls over the top of natural turfs.

MistakeNot4892 (approved):

Issue #4191, opened 2024-07-13
Collected issues from Pyrelight test


MistakeNot4892:
Posting this mainly for my own reference, so pardon the condensed ticket.

- ~~shrine needs salt, hoe, bowls/cups~~
- ~~area ambience is still using space sounds~~
- ~~days are way too short~~
- ~~weather needs to be weighted to clear conditions, most of round was spent in rain or hail~~
- ~~drakes cannot eat items held in mouth (dexterity check)~~
- ~~carrot sticks are white (chopped carrot is not)~~
- ~~broth is being created in 3u increments instead of 10/15u~~ pot was full
- ~~barrels in the rain don't refill (might not be an issue, barrels have a lid)~~ calling it not an issue due to need for barrels to process to refill like this
- ~~flowers should fit in ear slots~~ too complicated, todo
- ~~moving onto lit fires should check deliberate movement.~~
- ~~can't use a grab to position things on open space~~
- ~~dead hnoll still animate their tail wags~~

- tree shadows are off by 1 tile to the bottom left and leave baked in ambient lighting
- bandages and poultices have no failure message on use, and will attack even on help intent

Greenjoe12345:
To add here,  drakes cannot eat small mobs such as mice whole.

MistakeNot4892:
Going to sort out the rest of these via a list in Pyrelight, or open them as Neb issues.

PR #4196, opened 2024-07-13
Fixes some dex issues with food and natural weapons.

### LINKED ISSUE #4193: Simple mobs do not reset their attack animations on failure
#### Description of issue
Simple mobs like giant spiders do not reset if they pull back to strike but fail to resolve. This causes them to just creep further and further away from their actual location.

They should reset to neutral regardless of the outcome of their attack.

#### Steps to reproduce
Spawn a giant spider and have it attack a dummy human.

Likely since the big AI rewrite a few weeks ago.

Current dev.

#### Issue bingo

### LINKED ISSUE #4194: Simple mobs do not show successful attack animations, but apply attack effects like venom.
#### Description of issue
Title.

Attacks should play a sound and show a message prior to applying attack effects.

#### Steps to reproduce
Giant spiders are a good example case currently. Spawn one and let it nibble you.

Probably since AI rewrite.

Current dev.

#### Issue bingo

MistakeNot4892:
- Fixes #4193
- Fixes #4194
- Fixes drake food issue in #4191

### DIFF
--- a/code/_onclick/other_mobs.dm
+++ b/code/_onclick/other_mobs.dm
@@ -74,11 +74,11 @@
 	if(.)
 		return
 
-	a_intent = I_HURT
 	var/attacking_with = get_natural_weapon()
 	if(a_intent == I_HELP || !attacking_with)
 		return A.attack_animal(src)
 
+	a_intent = I_HURT
 	. = A.attackby(attacking_with, src)
 	if(!.)
 		reset_offsets(anim_time = 2)
--- a/code/modules/reagents/reagent_containers/food.dm
+++ b/code/modules/reagents/reagent_containers/food.dm
@@ -22,6 +22,7 @@
 	center_of_mass = @'{"x":16,"y":16}'
 	w_class = ITEM_SIZE_SMALL
 	abstract_type = /obj/item/chems/food
+	needs_attack_dexterity = DEXTERITY_NONE
 
 	/// Indicates the food should give a stress effect on eating.
 	// This is set to 1 if the food is created by a recipe, -1 if the food is raw.
--- a/code/_onclick/item_attack.dm
+++ b/code/_onclick/item_attack.dm
@@ -57,10 +57,11 @@ avoid code duplication. This includes items that may sometimes act as a standard
 /mob/living/attackby(obj/item/used_item, mob/user)
 	if(!ismob(user))
 		return TRUE
-	if(can_operate(src, user) != OPERATE_DENY && used_item.do_surgery(src,user)) //Surgery
-		return TRUE
-	if(try_butcher_in_place(user, used_item))
-		return TRUE
+	if(user.a_intent != I_HURT)
+		if(can_operate(src, user) != OPERATE_DENY && used_item.do_surgery(src,user)) //Surgery
+			return TRUE
+		if(try_butcher_in_place(user, used_item))
+			return TRUE
 	var/oldhealth = current_health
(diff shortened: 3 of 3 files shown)


code/modules/mob/animations.dm
```diff
@@ -186,3 +186,7 @@
 	else
 		animate(src, pixel_x = pixel_x + pixel_x_diff, pixel_y = pixel_y + pixel_y_diff, time = windup_time-2)
 		animate(pixel_x = default_pixel_x, pixel_y = default_pixel_y, time = 2)
+
+// reset wherever the attack animation got us to.
+/atom/movable/proc/reset_windup_animation()
```
out-of-phaze: Uh, what's the difference between this and `reset_offsets(anim_time = 2)` (the default)? It just seems like this is less robust? Does it interact with grabs properly?

code/modules/reagents/reagent_containers/food.dm
```diff
@@ -22,6 +22,7 @@
 	center_of_mass = @'{"x":16,"y":16}'
 	w_class = ITEM_SIZE_SMALL
 	abstract_type = /obj/item/chems/food
+	needs_attack_dexterity = DEXTERITY_NONE
```
out-of-phaze, line 25: I don't like this. We confirmed drakes can eat foods fine by using it in-hand, which I thought was the intended way for them to do it? This only affects clicking themselves (and other things) with food, and will undoubtedly have other side effects.

out-of-phaze (changes requested):

code/modules/reagents/reagent_containers/food.dm
MistakeNot4892: This is fixing them not being able to eat food inhand.

MistakeNot4892: Either that or Greenjoe misreported the issue.

out-of-phaze: it was misreported, we confirmed it on discord. using it inhand worked, clicking yourself is what didn't and that's intended afaik

MistakeNot4892: Per Discord I think I was misunderstood re: intentionality. I don't think it's good UX to arbitrarily have one method work and the other not for something as fundamental as eating. With respect to side effects, I can't think of anything offhand other than drakes feeding food to people, which is a bit whatever.

code/modules/mob/animations.dm
MistakeNot4892: It was just extracting the existing line. I'll swap it out for reset_offsets() I guess.

code/modules/reagents/reagent_containers/food.dm
out-of-phaze: in that case yeah, this is fine

out-of-phaze (dismissed):

code/_onclick/other_mobs.dm
```diff
@@ -79,6 +79,7 @@
 	if(a_intent == I_HELP || !attacking_with)
 		return A.attack_animal(src)
 
+	a_intent = I_HURT
```
out-of-phaze, line 81: already exists on line 77

MistakeNot4892: Line 77 is actually in error. Resolved now anyway.

out-of-phaze (approved):
still 24h under conflict resolution rule, yay

PR #4198, opened 2024-07-13
Fixes several Shaded Hills map issues

MistakeNot4892:
Fixes several issues from #4191.
- Renames several daycycle vars/types to be clearer.
- Increases length of default daycycle to an hour, and shaded hills to 2 hours. The effective daylight period is less than half the total, so it was very short on Shadyhills.
- Uncommented weather transition weighting and tweaked the weightings.
- Allowed weather to be supplied a list of states to forbid, and forbids snow on Shaded Hills.
- Moves Shaded Hills weather into a new finalize_map_generation() proc on /datum/map, called at the end of SSmapping Initialize().
- Adds GLUT_TINY to drakes so they can eat mice. This is important apparently.


code/modules/weather/weather_fsm.dm
```diff
@@ -6,5 +6,5 @@
 	for(var/decl/state_transition/weather/state_transition in valid_transitions)
 		transitions[state_transition] = state_transition.likelihood_weighting
 	if(length(transitions))
-		return pick(transitions) //pickweight(transitions)
+		return pickweight(transitions)
```
out-of-phaze, line 9: ***WHAT***

out-of-phaze (approved):

code/modules/weather/weather_fsm.dm
MistakeNot4892: What?

out-of-phaze: i had no clue the transition weighting was disabled...

MistakeNot4892: I thought pickweight was broken.

PR #4199, opened 2024-07-13
Improve shutters on fantasy map

out-of-phaze:
## Description of changes
- Makes wall shutters propagate light in a cone. Works pretty neat, but it'd be nice if it automatically updated the icon when the adjacent turf's lighting changed...
- Adds more shutters to the fantasy map for adequate daytime lighting without needing to have a lantern lit.
- Adds `throwpass` to textile benches so you can reach over them when trying to interact with shutters.

Better indoor lighting during the day/without lanterns on the fantasy map.

## Authorship
Me

## Changelog
:cl:
add: Added light propagation to open wall shutters
/:cl:


code/modules/lighting/lighting_atom.dm
```diff
@@ -9,6 +9,8 @@
 	//  Useful for objects like light fixtures that aren't visually in the middle of the turf, but aren't offset either.
 	var/light_offset_x
 	var/light_offset_y
+	/// An override for cases where the light is not facing the same direction as the object.
+	var/light_dir
```
out-of-phaze, line 13: (yes, this is a new atom var, but it has no memory cost unless changed, and most things won't have it changed)

MistakeNot4892 (approved):
Worked very well in the Pyrelight test today.

PR #4206, opened 2024-07-15
Fixes quote marks getting scrambled in custom emotes.

### LINKED ISSUE #4201: putting symbols in 'me' statements causes weird syntax issues
#### Description of issue
image
This is an example of putting quote marks into me statements, but I think I saw someone experience similar for ' half quote/apostrophes' 

ideally the symbols would just appear when we put them into 'me' statmeents

#### Steps to reproduce
enter 'me' and then hit space and enter a statement with special symbols like " or '

In your text chat

last couple games

Client Version: 515
Server Revision: 4af3f265d980627663f96ce1a32d889ffc411a21 - dev - 2024-07-14
Game ID: cuH-afFs
Current map: Shaded Hills

#### Issue bingo

MistakeNot4892:
Fixes #4201. 
Not sure if this is a good fix, or why html_encode is being used in the first place, but this appears to fix it.

### DIFF
--- a/code/modules/emotes/emote_mob.dm
+++ b/code/modules/emotes/emote_mob.dm
@@ -168,6 +168,7 @@
 	message = trim(html_encode(message))
 	message = filter_modify_message(message)
 	message = format_emote(src, message)
+	message = trim(html_decode(message))
 
 	if (message)
 		log_emote("[name]/[key] : [message]")


code/modules/emotes/emote_mob.dm
```diff
@@ -168,6 +168,7 @@
 	message = trim(html_encode(message))
 	message = filter_modify_message(message)
 	message = format_emote(src, message)
+	message = trim(html_decode(message))
```
out-of-phaze, line 171: I forget why it was encoded in the first place, I swear that at one point it was decoded after, but I can't find it now. Anyway, this trim had to be added regardless. I added trims to the markdown filters but I don't necessarily trust them to be exhaustive.

out-of-phaze (approved):

PR #4213, opened 2024-07-15
Adds map-specific skill names descriptions for Shaded Hills

Greenjoe12345:
## Description of changes
Adds some overrides for some skill names and descriptions to make them more fitting for Shaded Hills. Also removes high finance skill giving Legalease. Adds map-specific skill level names.
## Why and what will this PR improve
It doesn't mention modern things like microwaves that don't exist in Shaded Hills, as well as removing references to modern jobs such as EMTs. Legalese is unwanted on shadedhills.

## Authorship
Myself, Loaf

## Changelog
:cl:
add: Added map-specific skill level, name and description overrides for Shaded hills.
/:cl:

PR #4214, opened 2024-07-16
Renames the skill ranks on Shaded Hills.

MistakeNot4892:
Shaded Hill skills get slightly more thematically appropriate names.


MistakeNot4892:
Greenjoe will put this one in #4213.

PR #4222, opened 2024-07-18
Generalizing the weather processing system for atoms.

MistakeNot4892:
- `SSweather_atoms` handles processing atoms that care about weather (humans and cleanable decals currently).
- Most cleanable decals will slowly fade away while rained on. 
- Weather ambience for mobs is handled by `handle_environment()` while exposure effects are handled by `process_weather()`.
- `get_affecting_weather()` and `is_outside()` have been generalized to `/atom`.

### DIFF
--- a/code/modules/mob/living/life.dm
+++ b/code/modules/mob/living/life.dm
@@ -264,20 +264,6 @@
 /mob/living/proc/handle_random_events()
 	return
 
-/mob/living/proc/is_outside()
-	var/turf/T = loc
-	return istype(T) && T.is_outside()
-
-/mob/living/proc/get_affecting_weather()
-	var/turf/my_turf = get_turf(src)
-	if(!istype(my_turf))
-		return
-	var/turf/actual_loc = loc
-	// If we're standing in the rain, use the turf weather.
-	. = istype(actual_loc) && actual_loc.weather
-	if(!.) // If we're under or inside shelter, use the z-level rain (for ambience)
-		. = SSweather.weather_by_z[my_turf.z]
-
 /mob/living/proc/handle_contact_reagent_dripping()
 	// TODO: process dripping outside of Life() so corpses don't become sponges.
 	// TODO: factor temperature and vapor into this so warmer locations dry you off.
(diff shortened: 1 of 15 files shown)


code/modules/mob/living/life.dm
```diff
@@ -291,18 +277,22 @@
...
-/mob/living/proc/handle_weather_effects(obj/abstract/weather_system/weather)
-	// Handle physical effects of weather.
-	if(!istype(weather))
-		return
-	var/decl/state/weather/weather_state = weather.weather_system.current_state
-	if(istype(weather_state))
-		weather_state.handle_exposure(src, get_weather_exposure(weather), weather)
+/mob/living/process_weather(obj/abstract/weather_system/weather, decl/state/weather/weather_state)
+	// Handle physical effects of weather. Ambience is handled in handle_environment
+	// as clients should get wind noise etc. regardless of weather sensitivity.
```
out-of-phaze: i don't think this is correct, i think `WEATHER_IGNORE` should block ambience, while `WEATHER_EXPOSED`, `WEATHER_ROOFED`, and `WEATHER_PROTECTION` should allow it

MistakeNot4892: Why would WEATHER_IGNORE block ambience? Ambience is stuff like rain drumming on the roof or wind noises.

MistakeNot4892: Nvm I'm stupid and forgot my own code, WEATHER_IGNORE skips ambience anyway. Dunno if I like that on reflection, it'd be nice to hear rain from inside a locker or tent or such.

MistakeNot4892: Updated the comment but I think the structure is sound. Mobs without clients don't need to do any weather ambience processing and the other proc has a client check.

out-of-phaze: I'm pretty sure `WEATHER_IGNORE` is specifically for "you are two z-levels down and there is no way the weather could reach you," IIRC? Like how we wouldn't want rain sounds to be audible from within the deep caverns

MistakeNot4892: I'm not touching that logic so I'm a bit confused by what this comment thread wants me to do.

out-of-phaze: was just replying to `Dunno if I like that on reflection, it'd be nice to hear rain from inside a locker or tent or such.`, no request for changes. as evidenced by how there were not changes requested

out-of-phaze (changes requested):
Has an extra commit that shouldn't be in there.

out-of-phaze (approved):

PR #4226, opened 2024-07-19
You can now use honey to disinfect limbs.

MistakeNot4892:
Medieval tech inclusion. Still need to test it locally.

### DIFF
--- a/code/modules/surgery/other.dm
+++ b/code/modules/surgery/other.dm
@@ -154,7 +154,7 @@
 //////////////////////////////////////////////////////////////////
 /decl/surgery_step/sterilize
 	name = "Sterilize wound"
-	description = "This procedure sterilizes a wound with alcohol."
+	description = "This procedure sterilizes a wound with antiseptic substances such as alcohol or raw honey."
 	allowed_tools = list(
 		/obj/item/chems/spray = 100,
 		/obj/item/chems/dropper = 100,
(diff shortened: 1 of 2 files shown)


code/modules/surgery/other.dm
```diff
@@ -229,12 +240,15 @@
...
-	if(container.reagents.has_reagent(/decl/material/liquid/antiseptic))
-		return TRUE
+	if(!container.reagents?.total_volume)
+		return FALSE
 
-	for(var/rtype in container?.reagents?.reagent_volumes)
-		var/decl/material/liquid/ethanol/booze = GET_DECL(rtype)
-		if(istype(booze) && booze.strength <= 40)
-			return TRUE
+	// Must be impure.
```
out-of-phaze:
```suggestion
	// This check means it's impure.
```
the way this is phrased makes it sound like we're saying that the reagents "must be impure," as in it's a requirement to proceed. i guess you don't need to change it, but like, still

out-of-phaze (changes requested):

code/modules/surgery/other.dm
MistakeNot4892: Blocking over the phrasing of comments made me side-eye this but shrug. Committed your suggestion.

out-of-phaze: i did say "I guess you don't need to change it," i guess maybe i should've been clearer and said "non-blocking" explicitly

also ironic to have this miscommunication when i was complaining about the phrasing of a comment being unclear. whoops. should remove the log from my own eye before criticizing the speck in my neighbor's etc

out-of-phaze (approved):

PR #4227, opened 2024-07-19
Loadout now uses the decl system and UID.

MistakeNot4892:
Pointing this at stable as it fixes an issue with loadout being broken if you try to load your character before someone has looked at loadout options.

- Fixes https://github.com/ScavStation/ScavStation/issues/1030
- Loadout uses UID instead of name.
- Loadout uses the decls repository instead of a global list.
- `colour select` find and replaced to `color select` for consistency and accuracy (since it is setting `color`).
- UID added to all loadout options.


out-of-phaze (approved):
this will cause a shitton of conflicts when merged into dev btw, have fun

PR #4229, opened 2024-07-19
Various fixes from Scav issues.

MistakeNot4892:
- Fixes https://github.com/ScavStation/ScavStation/issues/1033
- Fixes https://github.com/ScavStation/ScavStation/issues/1028
- Fixes https://github.com/ScavStation/ScavStation/issues/1032


out-of-phaze (dismissed):
definitely need temperature simulation on machinery imo, i believe smelters use it, and even if they don't, downstream we use it on cooking appliances :(

out-of-phaze (approved):

PR #4233, opened 2024-07-20
Item damage calculation rework.

MistakeNot4892:
## Description of changes
- Removes `throwforce`, `force`, various related variables. 
- Blows away a lot of specific var overrides of the above to clean the slate.
- Removes /twohanded and repaths the existing twohanded weapons to other types.
- Twohanding now checks slot or hand dexterity against weapon attack dexterity.
- Revisits/simplifies weapon force calculation.
- Renames several multipliers to shorter names.

## TODO
- [x] Coarse retype and unification.
- [x] Update pipe wrench to play nicely with base wrench code.
- [x] Convert axes to use the various /bladed material and icon handling.
- [x] Convert spears to above.
- [x] Update gun twohanding.
- [x] Verify baseball bats work as expected.
- [x] Wielded icons for broadsword
- [x] Wielded icons for longsword
- [x] Readd base attack force to atoms that previously relied on the material damage multiplier.
- [x] Reimplement material based damage calculation.
- [x] Revisit thrown, material and wielding multipliers. Do a spreadsheet to make sure weapons are sane with a variety of materials.
- [X] Verify force and damage flag updates are correct.
- [x] Add migration txt

- Cleaner code.
- More predictable, saner and cleaner damage calculation.
- Makes twohanding generally available to all item types.
- Makes force calculation more intuitive/cleaner.

## Authorship
Myself.

## Changelog
TODO before merge.


MistakeNot4892:
Christ this got out of hand.

MistakeNot4892:
Alright so the new damage calc is basically:
- base damage, plus or minus a component based on the hardness and weight of the material (currently defaulting to 25%, so plus or minus 12%). Wielding (when possible) then multiplies the damage by a further 25%.
- throw damage is the above, multiplied by a further thrown force multiplier (defaults to 30% of total).
- Sharp/edge are calculated based on material procs, currently mostly still using the hardness checks used previously.

MistakeNot4892:
Generally speaking, this PR limits the impact of material on weapon force, which makes weapon damage bounds a bit more measured and predictable. I think the only way to really thrash the impact of these changes out is playtesting. I'm happy to pull it into Pyrelight and see how that goes if we can get some combat in.

MistakeNot4892:
Spreadsheet of weapon values:
[weapon_stats_dump.csv](https://github.com/user-attachments/files/16323764/weapon_stats_dump.csv)

out-of-phaze (dismissed):
do we want to just merge this on pyrelight for the test and see how it all shakes out?

MistakeNot4892:
Yeah I've pulled it into volatile for tomorrow.

MistakeNot4892:
That push was conflict resolution.

out-of-phaze (approved):
more cornflake rizzolution, this PR is still ready to merge once it's been balance-tested in-game

code/game/objects/items/_item_materials.dm
```diff
@@ -113,18 +88,24 @@
...
 			else
 				remove_extension(src, armor_type)
+
 	queue_icon_update()
 
+/obj/item/proc/update_name()
+	if(material_alteration & MAT_FLAG_ALTERATION_NAME)
+		SetName("[material.adjective_name] [initial(name)]")
+	else
+		SetName(initial(name))
```
out-of-phaze: this causes growns to be named produce, should probably fix that. this will probably also lead to a bunch of other bugs... i am really not looking forward to having to dig through these

out-of-phaze (changes requested):

code/game/objects/items/_item_materials.dm
MistakeNot4892: I made /item keep track of a base_name set in Initialize() a la chems and posters which should take care of mapped names and growns.

MistakeNot4892: Also edited set_custom_name().

out-of-phaze (dismissed):
revised PR is DNM pending further testing on Pyrelight

MistakeNot4892:
Sick of dragging this out. Nothing exploded on Pyrelight, bugs can be sorted out like any other bug.

out-of-phaze (dismissed):
we got the big ones (item names, etc) so sure we can do this

out-of-phaze (approved):
cornflakes eaten

PR #4249, opened 2024-07-26
Increased towercap hardness to allow it to be used in crafting.

MistakeNot4892:
Bumps fungal wood up to the minimum hardness for most furniture and item crafting.


comma (approved):

out-of-phaze:
I don't know if I like this. It's considered softwood in the code for a reason, isn't it...?

MistakeNot4892:
It is one of the only sources of wood in hydrotrays, not being able to make wooden furniture with it sucks. I can change it to hardwood in lore if you want but the distinction is aesthetic.

PR #4258, opened 2024-07-27
General cleanup and rework of medical stack types.

MistakeNot4892:
## Description of changes
- Separates bandages and damage overlays into distinct layers.
- Moves poultice crafting down to the root bruise pack level.
- Repaths `/obj/item/stack/medical/bruise_pack` to `/obj/item/stack/medical/bandage` and `/obj/item/stack/medical/bruise_pack/bandage` to `/obj/item/stack/medical/bandage/crafted`.
- Repaths `/obj/item/stack/medical/ointment/poultice` to `/obj/item/stack/medical/ointment/crafted`.
- Repaths `/obj/item/stack/medical/advanced/bruise_pack` to `/obj/item/stack/medical/bandage/advanced`, cleans up and unifies behavior.
- Repaths `/obj/item/stack/medical/advanced/ointment` to `/obj/item/stack/medical/ointment/advanced`, cleans up and unifies behavior.
- Repaths crafted stacks and resin globs to `/crafted` and `/improvised`.
- Cleans up and modernizes medical stack applications, returns a message if the target is uninjured, and uses function overrides instead of overriding the entirety of use_on_mob().

Should make medical packs a lot nicer to use ingame and maintain in code.

## Authorship
Myself.

## Changelog
:cl:
tweak: Regular gauze packs can now be used to make herbal poultices.
/:cl:


code/datums/trading/traders/goods.dm
```diff
@@ -224,43 +224,44 @@ Sells devices, odds and ends, and medical stuff
...
+		/obj/item/camera                          = TRADER_THIS_TYPE,
+		/obj/item/destTagger                      = TRADER_THIS_TYPE,
+		/obj/item/gps                             = TRADER_THIS_TYPE,
+		/obj/item/measuring_tape                  = TRADER_THIS_TYPE,
+		/obj/item/ano_scanner                     = TRADER_THIS_TYPE,
+		/obj/item/core_sampler                    = TRADER_THIS_TYPE,
+		/obj/item/depth_scanner                   = TRADER_THIS_TYPE,
+		/obj/item/pinpointer/radio                = TRADER_THIS_TYPE,
+		/obj/item/stack/medical/ointment/advanced = TRADER_BLACKLIST,
+		/obj/item/stack/medical/bandage/advanced  = TRADER_BLACKLIST
```
out-of-phaze: looking at the old code i think it just specifically blacklisted selling `/obj/item/stack/medical/advanced` because that's an abstract type that shouldn't be accessible? while the same isn't true for these two. `TRADER_BLACKLIST` is just specifically that type, while `TRADER_BLACKLIST_SUB` is that and subtypes.

it also seems like Kearel accidentally removed `/obj/item/stack/medical  = TRADER_SUBTYPES_ONLY` eight whole years ago, and without it blacklisting any subtypes is pointless anyway.

out-of-phaze (dismissed):

code/datums/trading/traders/goods.dm
MistakeNot4892: I thought it was blacklisting all advanced types because this is a 'drugstore' rather than a medical specialist.

MistakeNot4892: Updated.

out-of-phaze (approved):

PR #4263, opened 2024-07-27
Limit Zalgo-potential

PsiOmegaDelta:
`/proc/sanitize()` now also limits Unicode diacritical marks to at most one for a given character.
Still allows for characters such as é, ñ, å, ä, ö but might have an impact on Arabic and other Asian languages.

Goes from
Pre-sanitation
to 
Post-sanitation

The regex can also be changed to `[\u0300-\u036F\u1AB0-\u1ACE\u1DC0-\u1DFF\u20D0-\u20F0\uFE20-\uFE2F]` (and the replacement changed from `"$1"` to simply `""`) in which case the result is `ååääööÅÄÖäñéèÄáasdasdasdsadååååäöäöaäsödd`

Note: `\p{M}` should cover diacritical marks more cleanly, but BYOND's regex engine doesn't appear to support it.


MistakeNot4892:
I wonder if it'd be worth putting this behind a config so servers can enable it if so desired.

MistakeNot4892 (approved):

PR #4266, opened 2024-07-28
Reworks bees and beehives.

MistakeNot4892:
## Description of changes
- Reworks beehive frames into a general frame type that is filled with reagents, rather than a bespoke item subtype.
- Removes the existing beehive machinery and replaces it with `/obj/structure/apiary`, which is crafted directly from planks rather than using a bespoke parts item.
- Implements `/datum/extension/insect_hive` and `/decl/insect_species` to allow for nests to go on any structure, and contain a variety of insects with a variety of products.
- Implements bees using the above. Beehives will generate swarms, which will gather pollen (not currently modeled via reagent, just a counter) from nearby flowers, then return the pollen to the hive, where it will be used to fill a frame or create a comb.
- Swarms will need to be smoked before harvesting frames to avoid agitating the hive and getting stung.

## TODO
- [x] Rewrite beehive/implement apiary.
- [x] Implement and test frame filling/creating behavior.
- [x] Implement and test swarm behavior.
- [x] Add wasp and bee hives to dead trees.
- [x] Balance swarm growth and honey generation time.
- [x] Balance stinging.
- [x] Reimplement smoking.
- ~~Implement wasps venom and wasp stinging.~~

Ants and silkworms can come in another PR, this one needs to get finished.

Generalizes beekeeping in a way that allows for an expansion into other species (ants and wasps) and allows hives to be placed on arbitrary structures like dead trees or old closets.

## Authorship
Myself.

## Changelog
:cl:
add: Beekeeping has been rewritten almost entirely, please refer to the Beewrite PR on the GitHub repo for details.
/:cl:


MistakeNot4892:
Depends on #4855 and #4846

MistakeNot4892:
Not realistically going to finish the last 10% of this. It's functional other than balancing and reimplementation of smoking if someone wants to pick it up down the track.

MistakeNot4892:
TODO:
- ~~Run through building a beehive, adding the frame, planting flowers, etc. and see how it feels and if it all works.~~
- ~~Test stinging behaviors and swarm aggression.~~
- ~~Test smokers.~~

MistakeNot4892:
This is probably as good as it'll get for now. Swarm splitting would be nice to get working but this is a functional replacement for old bee mechanics.

code/game/objects/structures/flora/plant.dm
```diff
@@ -7,14 +7,20 @@
 	var/dead = FALSE
 	var/sampled = FALSE
 	var/datum/seed/plant
-	var/harvestable
+	var/harvestable = 0
```
out-of-phaze: comment explaining that this is the number of harvests rather than a legacy bool (e.g. hasn't been changed to use `FALSE` instead of `0`) would be nice but isnt mandatory

code/game/objects/structures/flora/plant.dm
```diff
@@ -7,14 +7,20 @@
...
 
 /obj/structure/flora/plant/large
 	opacity = TRUE
 	density = TRUE
 
-/* Notes for future work moving logic off hydrotrays onto plants themselves:
 /obj/structure/flora/plant/Process()
+	if(plant?.produces_pollen <= 0)
+		return PROCESS_KILL
+	if(pollen < 10)
```
out-of-phaze: not a huge fan of bare constants like this, would prefer making it a member var or a local const like `var/const/max_pollen = 10` at the top of the proc. sometimes code like this gets copied around (e.g. someone wants to copy pollen behavior to their new custom type) and the original constant gets misused/misinterpreted. not mandatory though

code/game/objects/structures/flora/plant.dm
```diff
@@ -147,3 +157,23 @@
...
+
+/obj/structure/flora/plant/random_flower
+	name = "flower"
+	color = COLOR_PINK
+	icon_state = "flower5"
+	is_spawnable_type = TRUE
+
+/obj/structure/flora/plant/random_flower/proc/get_flower_variants()
+	var/static/list/flower_variants
+	if(isnull(flower_variants))
```
out-of-phaze, line 170: noting that this will only ever contain roundstart plants, but that's probably desirable even if not explicitly made clear in the code

code/modules/client/preference_setup/occupation/occupation.dm
```diff
@@ -185,7 +185,7 @@
 					skill_link = "<a href='byond://?src=\ref[src];set_skills=[title]'>View Skills</a>"
 				skill_link = "<td>[skill_link]</td>"
 
-				if(!user.skillset.skills_transferable)
+				if(!user.skillset?.skills_transferable)
```
out-of-phaze: no longer necessary because of https://github.com/NebulaSS13/Nebula/pull/5150

mods/content/beekeeping/hives/hive_swarm.dm
```diff
@@ -0,0 +1,358 @@
+/obj/effect/insect_swarm
```
out-of-phaze, line 1: a little curious why this is an effect and not a simple_animal. too much boilerplate you'd have to work around, i assume?parrots already do a similar species/type thing, so i figured it wouldn't be impossible, and mobs-as-effects has historically been a pain (see spiderlings for an example)

edit: oh it might be the merging? mob gc churn would be more annoying than doing the same with effects

mods/content/beekeeping/hives/hive_swarm.dm
```diff
@@ -0,0 +1,358 @@
...
+		swarm_transform.Turn(pick(90, 180, 270))
+
+/obj/effect/insect_swarm/proc/is_agitated()
+	return QDELETED(owner) || (swarm_agitation > 0 && !is_smoked())
+
+/obj/effect/insect_swarm/proc/find_sting_target()
+	for(var/mob/living/victim in view(7, src))
+		if(!victim.simulated || victim.stat || victim.current_posture?.prone)
+			continue
+		if(victim.isSynthetic())
```
out-of-phaze: another one for the list of 'isSynthetic uses to remove someday'. though i might as well ask: what is this meant to represent? is it "swarms won't target someone that isn't alive (because they don't exhale, aren't warm, aren't made of meat materials-wise, etc)" or "swarms can't sting someone if their stinger isn't strong enough to break the skin (e.g. if the skin is made of metal, stone, whatever)"?

mods/content/beekeeping/hives/hive_swarm.dm
```diff
@@ -0,0 +1,358 @@
...
+			break
+	return FALSE
+
+/obj/effect/insect_swarm/get_automove_target(datum/automove_metadata/metadata)
+	return move_target
+
+/obj/effect/insect_swarm/stop_automove()
+	SHOULD_CALL_PARENT(FALSE)
+	move_target = null
+	//. = ..() // TODO work out why they're not automoving
```
out-of-phaze: noting this todo here, is it still the case?

mods/content/beekeeping/hives/hive_swarm.dm
```diff
@@ -0,0 +1,358 @@
...
+	walk(src, 0)
+
+/obj/effect/insect_swarm/start_automove(target, movement_type, datum/automove_metadata/metadata)
+	SHOULD_CALL_PARENT(FALSE)
+	move_target = target
+	//. = ..() // TODO work out why they're not automoving
```
out-of-phaze: ditto, is this TODO still current? was under the impression the PR has been ready to review for a bit

mods/content/beekeeping/hives/hive_swarm.dm
```diff
@@ -0,0 +1,358 @@
...
+			closest_target = thing
+			closest_dist = next_dist
+
+	if(closest_target)
+		start_automove(closest_target)
+	else
+		start_automove(owner.holder)
+
+// TODO: update icon (twitching on ground?)
+// TODO: lower agitation
```
out-of-phaze: are these work for a future PR or for this one?

mods/content/beekeeping/items.dm
```diff
@@ -1,47 +1,34 @@
...
 	material = /decl/material/solid/metal/steel
 
-/obj/item/bee_pack
-	name = "bee pack"
-	desc = "Contains a queen bee and some worker bees. Everything you'll need to start a hive!"
-	icon = 'mods/content/beekeeping/icons/beekeeping.dmi'
-	icon_state = "beepack"
-	material = /decl/material/solid/organic/plastic
-	var/full = 1
+// TODO: consume reagents or charges? Unnecessary complexity?
```
out-of-phaze, line 9: > There are many options for beehive smoker fuel including cotton, bailing twine, pine needles, and wood chips. For best results, use fuel that produces cool, white smoke and avoid harsh smoke that can irritate bees.

seems pretty simple irl, just any flammable waste matter. i think the only thing that wouldnt work is lamp oil. actually it seems a useful thing irl for this is cardboard pellets/scraps which would work perfectly on a space map, probably?

mods/content/beekeeping/hives/hive_swarm.dm
MistakeNot4892: Both, I was thinking a bee swarm wouldn't identify a machine as alive and can't sting it. The sting proc does check can_inject so this could probably be relaxed.

mods/content/beekeeping/hives/hive_swarm.dm
MistakeNot4892: Yeah this is fixed, I just missed those. That actually explains why the automove isn't very snappy.

mods/content/beekeeping/items.dm
MistakeNot4892: I guess I could use that general proc for burning materials and returning waste products, and make the smoker a container that can hold a stack. My concern is mostly that it will mean anyone who wants to do beekeeping now needs to source smoker fuel.

mods/content/beekeeping/hives/hive_swarm.dm
MistakeNot4892: Missed this one, I'll hack out some sleepy states.

mods/content/beekeeping/hives/hive_swarm.dm
MistakeNot4892: The merging and also general health and processing behavior - I don't want you to be able to beat up a swarm with a baseball bat or inject it with a syringe like you can with mobs in general. That said, it would be nice to be able to put someone in control of a swarm of bees, and maybe swarm intensity could be handled via health or something, but... yeah.

code/game/objects/structures/flora/plant.dm
MistakeNot4892: Yeah I'll add a comment to make it clear.

code/game/objects/structures/flora/plant.dm
MistakeNot4892: I'll add a define alongside the other pollen constants.

code/game/objects/structures/flora/plant.dm
MistakeNot4892: I'll add a comment.

mods/content/beekeeping/items.dm
out-of-phaze: yeah, but 'cardboard cubes/pellets/etc comes in the beekeeping kit in cargo' would be an okay solution to that

MistakeNot4892: Had a poke but I think I want to call this out of scope. We don't have a pellets stack and I don't really want to add one, add recipes, etc. As well as redoing the smoker to handle refuelling and such.

mods/content/beekeeping/hives/hive_swarm.dm
MistakeNot4892: Done.

mods/content/beekeeping/hives/hive_swarm.dm
MistakeNot4892: This is removed, not sure why it's not showing up as out of date.

MistakeNot4892:
- Automove is not working.
- ~~Pollen generation is wrong, it's adding 0.000003 per SSplants run instead of 0.2.~~ I thought it generating 12 per run was an error but nope SSplants is just on a massive wait.

mods/content/beekeeping/hives/hive_swarm.dm
```diff
@@ -0,0 +1,374 @@
+/obj/effect/insect_swarm
+	anchored          = TRUE
+	is_spawnable_type = FALSE
+	icon_state        = "0"
+	gender            = NEUTER
+	default_pixel_z   = 8
+	layer             = ABOVE_HUMAN_LAYER
+	pass_flags        = PASS_FLAG_TABLE
+	movement_handlers = list(/datum/movement_handler/delay/insect_swarm)
```
out-of-phaze:
```dm
/datum/movement_handler/delay/New(var/host, var/delay)
	..()
	src.delay = max(1, delay)
```
it's gonna get its delay set wrong bc of this inheritance, maybe `/delay` needs an isnull to avoid setting it if we pass null. that or just do `list(/datum/movement_handler/delay = 1 SECOND)`

mods/content/beekeeping/hives/hive_swarm.dm
```diff
@@ -0,0 +1,374 @@
...
+/obj/effect/insect_swarm/stop_automove()
+	SHOULD_CALL_PARENT(FALSE)
+	move_target = null
+	. = ..()
+
+/obj/effect/insect_swarm/can_do_automated_move(variant_move_delay)
+	return !is_smoked()
+
+/obj/effect/insect_swarm/start_automove(target, movement_type, datum/automove_metadata/metadata)
+	SHOULD_CALL_PARENT(FALSE)
```
out-of-phaze:
```suggestion
	move_target = null
	. = ..()

/obj/effect/insect_swarm/can_do_automated_move(variant_move_delay)
	return !is_smoked()

/obj/effect/insect_swarm/start_automove(target, movement_type, datum/automove_metadata/metadata)
```

out-of-phaze (changes requested):

mods/content/beekeeping/hives/hive_swarm.dm
MistakeNot4892: I did do the latter already.

code/game/objects/items/devices/chameleonproj.dm
```diff
@@ -102,7 +102,7 @@
 	density = FALSE
 	anchored = TRUE
 	is_spawnable_type = FALSE
-	movement_handlers = list(/datum/movement_handler/delay/chameleon_projector)
+	movement_handlers = list(/datum/movement_handler/delay/chameleon_projector = list(2.5 SECONDS))
```
out-of-phaze, line 105:
```suggestion
	movement_handlers = list(/datum/movement_handler/delay = list(2.5 SECONDS))
```
subtype no longer needed

mods/content/beekeeping/hives/hive_swarm.dm
```diff
@@ -0,0 +1,371 @@
+/obj/effect/insect_swarm
+	anchored          = TRUE
+	is_spawnable_type = FALSE
+	icon_state        = "0"
+	gender            = NEUTER
+	default_pixel_z   = 8
+	layer             = ABOVE_HUMAN_LAYER
+	pass_flags        = PASS_FLAG_TABLE
+	movement_handlers = list(/datum/movement_handler/delay/insect_swarm = list(1 SECOND))
```
out-of-phaze, line 9:
```suggestion
	movement_handlers = list(/datum/movement_handler/delay = list(1 SECOND))
```
subtype no longer necessary

out-of-phaze (dismissed):

code/game/objects/items/devices/chameleonproj.dm
MistakeNot4892: Subtype needed for MayMove/DoMove.

mods/content/beekeeping/hives/hive_swarm.dm
MistakeNot4892: Subtype is needed for DoMove.

MistakeNot4892:
Still need to fix automoving on this.

MistakeNot4892:
Automove is now working, tyvm Penny.

Have also tweaked sting amount and sting pain as one sting put 100u of bee venom into me, so that should be more reasonable now.

out-of-phaze (dismissed):

out-of-phaze (approved):

PR #4267, opened 2024-07-28
Separates solid and liquid reagents

NataKilar:
## Description of changes
Separates solid and liquid reagents in reagent holders, allowing for more interesting chemistry, fluid handling and (in the near future) state changes.
* Reagent holders now contain two additional lazy-initialized lists, which keep track of the states of the contained reagents. The original ``reagent_volumes`` list remains.
* Most procs on ``/datum/reagents`` now contain a flag for what phases should be transferred. By default, both types of reagents will be transferred. In cases where there is a limited amount of space, transfer of liquids will be prioritized.
* Fluids in the world now mostly look at the total liquid volume rather than total reagent volume when determining if fluids should evaporate or not. When fluids drop beneath the minimum ``FLUID_SLURRY`` depth, solids are left behind in piles.

On the technical side, this doesn't seem to affect performance significantly. As mentioned above, the ``reagent_volumes`` list remains, but it could reasonably be replaced by summing the contents of the two lists via the ``REAGENT_VOLUME`` macro in many current instances. The vast majority of references to reagents in the game still only reference the total reagent volumes and do not care about the phase. Changing this should probably be done piecemeal where necessary.

This is an important first step into improving our chemistry gameplay, and eventually allowing for state changes and liquid flow via pipes. On the player facing side, this doesn't change much yet, but it stop reagents which are solid at STP from being named "molten iron" and the like when poured onto the ground.

In the near future, I plan on reworking reagent temperature to account for different specific heats of substances. After, I hope to tackle the creation of solutions (in our representation, just moving solid and gaseous reagents into the liquids list) and making chemical reactions sensitive to reagent phase. Proper state changes as a function of temperature is the biggest hurdle, and will probably come last.

## Authorship
Myself

## TODO
- [x] Benchmark fluid performance
- [x] Check player facing verbiage wrt reagents
- [x] Test chemical reactions
## Changelog
:cl:
tweak: Solid state reagents now behave mostly like solid state reagents.
/:cl:


code/__defines/materials.dm
```diff
@@ -64,10 +64,10 @@
...
 
 //Phase of matter placeholders
-#define MAT_PHASE_SOLID     "solid"
-#define MAT_PHASE_LIQUID    "liquid"
-#define MAT_PHASE_GAS       "gas"
-#define MAT_PHASE_PLASMA    "plasma"
+#define MAT_PHASE_SOLID     BITFLAG(0)
+#define MAT_PHASE_LIQUID    BITFLAG(1)
+#define MAT_PHASE_GAS       BITFLAG(2)
+#define MAT_PHASE_PLASMA    BITFLAG(3)
```
MistakeNot4892, line 70: Per Discord, just leaving a note to consider if anything is using phases as keys or such currently.

code/controllers/subsystems/fluids.dm
```diff
@@ -129,7 +132,7 @@ SUBSYSTEM_DEF(fluids)
 		// Wash our turf.
 		current_fluid_holder.fluid_act(reagent_holder)
 
-		if(isspaceturf(current_fluid_holder) || (istype(current_fluid_holder, /turf/floor) && (current_fluid_holder.turf_flags & TURF_FLAG_ABSORB_LIQUID) && (current_fluid_holder.reagents?.total_volume + current_fluid_holder.get_physical_height()) > 0))
+		if(isspaceturf(current_fluid_holder) || (istype(current_fluid_holder, /turf/floor) && (current_fluid_holder.turf_flags & TURF_FLAG_ABSORB_LIQUID) && (current_fluid_holder.reagents?.total_liquid_volume + current_fluid_holder.get_physical_height()) > 0))
```
MistakeNot4892: Per Discord, imo flow depth should be total volume not liquids alone, as you will end up with strange flows that equalize in unexpected directions otherwise.

code/_helpers/medical_scans.dm
```diff
@@ -49,10 +49,18 @@
 	scan["reagents"] = list()
 
 	if(H.reagents?.total_volume)
-		for(var/reagent_type in H.reagents.reagent_volumes)
-			var/decl/material/R = GET_DECL(reagent_type)
+		for(var/liquid_type in H.reagents.liquid_volumes)
```
MistakeNot4892, line 52: Per Discord (also non-blocking), I really dislike the boilerplate needed just for the sake of variant phases of the same reagent. Maybe a `reagents.get_reagent_names()` proc that returns all names and associated instances for iteration like this. Might need to be like `list(list(name, instance))` due to non-enforced solid/liquid name uniqueness but it would be cleaner code than this.

code/game/turfs/turf_fluids.dm
```diff
@@ -177,12 +184,35 @@
...
 			ADD_ACTIVE_FLUID(neighbor)
+
+/turf/proc/dump_solid_reagents(datum/reagents/solids)
+	if(!istype(solids))
+		solids = reagents
+	if(LAZYLEN(solids?.solid_volumes))
+		var/list/matter_list = list()
+		for(var/reagent_type in solids.solid_volumes)
+			var/reagent_amount = solids.solid_volumes[reagent_type]
+			matter_list[reagent_type] = reagent_amount/REAGENT_UNITS_PER_MATERIAL_UNIT
```
MistakeNot4892: Needs rounding.

code/game/turfs/turf_fluids.dm
```diff
@@ -177,12 +184,35 @@
...
+		var/obj/item/debris/scraps/chemical/scraps = locate() in contents
+		if(!scraps)
+			scraps = new(src)
+		if(!LAZYLEN(scraps.matter))
+			scraps.matter = matter_list
+		else
+			for(var/mat_type in matter_list)
+				scraps.matter[mat_type] += matter_list[mat_type]
+
+		scraps.update_primary_material()
```
MistakeNot4892, line 216: Per Discord, it might be worth putting a block in update_primary_material() to make scraps split into several items based on matter exceeding a sheet, to avoid a single scrap with 1000 sheets of matter in it.

code/modules/materials/material_debris.dm
```diff
@@ -65,6 +72,64 @@
 /obj/item/debris/scraps/squash_item(skip_qdel = FALSE)
 	return
 
+// Physical object for holding solid reagents which are out of solution or slurry.
+/obj/item/debris/scraps/chemical
+	desc = "A pile of dust and small filings"
```
MistakeNot4892: `.`

code/modules/materials/material_debris.dm
```diff
@@ -10,15 +10,22 @@
 	name = "scraps"
 	desc = "A small pile of tailings and scraps."
 
+	var/total_matter = 0
```
MistakeNot4892: What's the purpose of caching this? Seems like a cheap calc to do at the point it's needed, and doesn't potentially get out of sync with the matter list.

code/__defines/materials.dm
NataKilar: Doesn't look like anything did.

code/controllers/subsystems/fluids.dm
NataKilar: Reverted these to use total volume - total liquid volume is still checked for slurry threshold and some fluid interaction procs.

code/_helpers/medical_scans.dm
NataKilar: I agree, but if it's alright I might leave it for a different PR.

code/game/turfs/turf_fluids.dm
NataKilar: Done

code/game/turfs/turf_fluids.dm
NataKilar: Done, scraps can only hold up to 5 sheets now (any less and large chemical spills become unmanageable)

code/modules/materials/material_debris.dm
NataKilar: ``.`` added

code/modules/materials/material_debris.dm
NataKilar: Removed. A few more things do use this now (since we have to split piles up dynamically) but it's easy enough to make into a proc.

MistakeNot4892:
Limited testing on Pyrelight today didn't show any immediate issues, though it wasn't exhaustive.

MistakeNot4892 (approved):

PR #4270, opened 2024-07-29
Fix hydrotrays and soil plots being flooded by rain

out-of-phaze:
## Description of changes
Makes rain stop adding water to soil plots/hydrotrays when their water level is full.

Fixes rain causing soil plots and hydrotrays to be completely drenched in water, to the point where there's not enough room to add food for the plants.


out-of-phaze:
duplicate, rip

PR #4279, opened 2024-07-30
Fix the taste of grown foods.

liz-lavenza:
## Description of changes
While investigating some reagent transfer issues for #4278, I noticed that nutriment from a non-dried tea leaf had no taste. I then saw the code for grown food reagent initialisation while breakpointing it to investigate rounding errors, and noticed that taste had been changed to use a separate list everywhere but there. Seems like a simple enough fix.

Tea leaves no longer give the message `You taste  .`, and will instead say `You taste tea leaf.`... which is at least a little better?


MistakeNot4892 (approved):
Augh of course I missed a spot. Thanks.

PR #4282, opened 2024-07-30
Fix misordered proc arguments in fire_source attackby

liz-lavenza:
## Description of changes
Corrects the order of proc arguments to `standard_pour_into()`.

The other changes were accidentally kept in from when I was doing some other edits, I can remove them if needed. I was trying to see if OpenDream could detect errors like this, but it seems it can't.

Fixes a runtime when clicking a fire_source structure (like a campfire or stove) with a reagent container.

### DIFF
--- a/code/game/objects/structures/fires.dm
+++ b/code/game/objects/structures/fires.dm
@@ -256,7 +256,7 @@
 		// Pour fuel or water into a fire.
 		if(istype(thing, /obj/item/chems))
 			var/obj/item/chems/chems = thing
-			if(chems.standard_pour_into(src, user))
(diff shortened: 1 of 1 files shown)


code/game/objects/structures/fires.dm
```diff
@@ -270,11 +270,12 @@
 
 	if((lit != FIRE_LIT || user.a_intent == I_HURT))
 		// Only drop in one log at a time.
+		var/obj/item/dropped = thing
```
MistakeNot4892: Per Discord, I would like these changes snipped out for neatness' sake.

MistakeNot4892 (approved):

PR #4285, opened 2024-07-31
Adding tail customization via sprite accessories.

MistakeNot4892:
- Adds a (currently unused) tail sprite accessory category that overrides tail appearance.
- Modifies some code to support tail hiding and tail colours.

## TODO
- [x] Get two-tone tails to work.
- [x] Get two-tone tails to save.
- [x] Work out why tails have disappeared from Unathi and Taj.
- [x] Check if two-tone tail icons work with animations.
- [x] Get hiding of accessory categories to work.
- [x] Get whitelisting to stop removing accessories during load.


MistakeNot4892:
Will depend on #4286 for two-tone tails.

code/_helpers/lists.dm
```diff
@@ -856,3 +856,21 @@ var/global/list/json_cache = list()
 /// Is this a dense (all keys have non-null values) associative list with at least one entry?
 /proc/is_dense_assoc(var/list/L)
 	return length(L) > 0 && !isnull(L[L[1]])
+
+// I have a feeling this is a redundant proc but I couldn't find another implementation.
+/proc/lists_are_equivalent(list/A, list/B, associative)
```
out-of-phaze: `/proc/same_entries` on line 209 of this file

out-of-phaze (dismissed):

code/_helpers/lists.dm
out-of-phaze: actually that doesn't properly handle the recursive case, so it should prboably be edited to do `(islist(first[entry]) && islist(second[entry])) ? same_entries(first[entry], second[entry]) : first[entry == second[entry]` or something

code/_helpers/lists.dm
```diff
@@ -205,14 +205,25 @@
...
 	if(length(first) != length(second))
 		return 0
 	for(var/entry in first)
-		if(!(entry in second) || (first[entry] != second[entry]))
+		if(!(entry in second))
+			return 0
+		var/first_entry = first[entry]
+		if(islist(first_entry))
+			var/second_entry = second[entry]
+			if(!islist(second_entry))
```
out-of-phaze, line 222: technically the first line of same_entries should handle this but it's fine

out-of-phaze (approved):

PR #4274, opened 2024-07-30
Refactor food to not be a reagent container subtype

out-of-phaze:
## Description of changes
Makes food not a subtype of `/obj/item/chems` to avoid issues with accidentally pouring food out.

In a future PR, I'll remove `ATOM_FLAG_OPEN_CONTAINER` from foods to fix them having their reagents washed out by fluids.

Tested:
- [x] Eating food directly
- [x] Eating food with a utensil
- [x] Adding reagents to food
- [x] Eating canned food
  - [x] Open
  - [x] Closed
- [x] Adding reagents to canned food
  - [x] Open
  - [ ] Closed

## Why and what will this PR improve
Fixes strange and undesirable interactions with foods due to them being reagent containers with half of the functionality stubbed out.

## Authorship
Me.

PR #4291, opened 2024-08-02
Make food not an open container

out-of-phaze:
## Description of changes
Makes food objects no longer open containers, preventing a bunch of unwanted reagent interactions with fluids, pouring, etc. Also removes a lot of typechecks that disabled (or tried to disable) those interactions since they aren't necessary anymore.

Relies on #4274.

Tested:
- [x] Pouring condiments/spices/etc. into food
- [x] Unable to pour out of food
- [x] Unable to eat closed canned food
- [x] Able to eat closed canned food
- [x] Unable to pour out of open canned food 
- [x] Unable to pour into closed canned food
- [x] Able to pour into open canned food

Prevents a bunch of unwanted interactions between food reagent contents and fluid/pouring/etc.

### DIFF
--- a/code/game/turfs/turf.dm
+++ b/code/game/turfs/turf.dm
@@ -280,9 +280,7 @@
 		return TRUE
 
 	if(reagents?.total_volume >= FLUID_PUDDLE)
-		// Must be open, but food items should not be filled from sources like this. They're open in order to add condiments, not to be poured into/out of.
-		// TODO: Rewrite open-container-ness or food to make this unnecessary!
-		if(ATOM_IS_OPEN_CONTAINER(W) && !istype(W, /obj/item/food) && W.reagents)
+		if(ATOM_IS_OPEN_CONTAINER(W) && W.reagents)
 			var/taking = min(reagents.total_volume, REAGENTS_FREE_SPACE(W.reagents))
 			if(taking > 0)
 				to_chat(user, SPAN_NOTICE("You fill \the [W] with [reagents.get_primary_reagent_name()] from \the [src]."))
--- a/code/modules/reagents/reagent_containers/food/canned.dm
+++ b/code/modules/reagents/reagent_containers/food/canned.dm
@@ -60,6 +60,9 @@
 /obj/item/food/can/apply_filling_overlay()
 	return //Bypass searching through the whole icon file for a filling icon
 
+/obj/item/food/can/can_be_poured_into(atom/source)
+	return (reagents?.maximum_volume > 0) && ATOM_IS_OPEN_CONTAINER(src)
+
 //Just a short line of Canned Consumables, great for treasure in faraway abandoned outposts
 
 /obj/item/food/can/beef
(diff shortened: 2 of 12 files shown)


code/game/turfs/turf.dm
```diff
@@ -280,9 +280,7 @@
 		return TRUE
 
 	if(reagents?.total_volume >= FLUID_PUDDLE)
-		// Must be open, but food items should not be filled from sources like this. They're open in order to add condiments, not to be poured into/out of.
-		// TODO: Rewrite open-container-ness or food to make this unnecessary!
-		if(ATOM_IS_OPEN_CONTAINER(W) && !istype(W, /obj/item/food) && W.reagents)
+		if(ATOM_IS_OPEN_CONTAINER(W) && W.reagents)
```
MistakeNot4892, line 283: Should this be `can_be_poured_into()`?

code/modules/reagents/reagent_dispenser.dm
```diff
@@ -86,10 +86,7 @@
 
 	// We do this here to avoid putting the vessel straight into storage.
 	// This is usually handled by afterattack on /chems.
-	// The item must be an open container, but food items should not be filled from sources like this.
-	// They're open in order to add condiments, not to be poured into/out of.
-	// TODO: Rewrite open-container-ness or food to make this unnecessary!
-	if(storage && ATOM_IS_OPEN_CONTAINER(W) && !istype(W, /obj/item/food) && user.a_intent == I_HELP)
+	if(storage && ATOM_IS_OPEN_CONTAINER(W) && user.a_intent == I_HELP)
```
MistakeNot4892, line 89: Wondering if there would be a use case for a `can_be_poured_from()` for some of these checks.

code/modules/reagents/reagent_containers/food/canned.dm
```diff
@@ -60,6 +60,9 @@
 /obj/item/food/can/apply_filling_overlay()
 	return //Bypass searching through the whole icon file for a filling icon
 
+/obj/item/food/can/can_be_poured_into(atom/source)
```
MistakeNot4892, line 63: I would suggest something like `reagents?.maximum_volume` maybe rather than just the reagent datum existing.

code/game/turfs/turf.dm
out-of-phaze: no because then we'd have to readd the food istype?

code/modules/reagents/reagent_containers/food/canned.dm
out-of-phaze: turns out i already did that on the base definition and just forgot to copy it over

code/modules/reagents/reagent_dispenser.dm
out-of-phaze: presently i think that's just `ATOM_IS_OPEN_CONTAINER(X)` but it could be nice to make it more contextual i suppose. ~~unsure if it fits in this PR bc there's not really much use for it atm~~ you can pour pills into fire_sources because they don't check `ATOM_IS_OPEN_CONTAINER(W)` oops. yeah there's a good use-case for this

MistakeNot4892: After the recent commit it looks like this is still using opencontainer, should it be using can_be_poured_from()?

out-of-phaze: it's checked in `standard_pour_into()`, i wasn't sure

MistakeNot4892 (approved):

PR #4303, opened 2024-08-07
Replace rounding macros with 515 procs

out-of-phaze:
## Description of changes
Since we require 515 now, I replaced `FLOOR()` with `floor()`, `CEILING()` with `ceil()`, and the unused `ROUND()` (identical to `MINMAX()` in lighting code) with `trunc()`.

Better code quality, fewer ugly redundant macros.


MistakeNot4892 (approved):
I got 70 files through, I am going to assume the rest is fine.

PR #4309, opened 2024-08-08
Delete plates spawned by cooked food

out-of-phaze:
## Description of changes
Deletes plates spawned by recipes. I was too lazy to do something like 'tally up the plates in the ingredients and delete any in the products that exceed it, and dump out any plates that didn't get used." This is simple and easy.

Cooking food won't spawn plates, and this doesn't have to add a new argument to skip plate creation.


MistakeNot4892:
Not a fan of this solution but I'll do a nicer fix on dev.

MistakeNot4892 (approved):

PR #4312, opened 2024-08-09
Adds a null decl check to loadout sanitizing.

MistakeNot4892:


MistakeNot4892:
Fixes the reported issue on Scav.

out-of-phaze (approved):

PR #4315, opened 2024-08-09
Food can now skip plate creation.

MistakeNot4892:
## Description of changes
- Adds a parameter to food init to skip creating a plate.
- Updating parameters in other food spawning to use the parameter.
- Adding code to food combination to take a plate if one is available.

Broader fix for potato fries getting a plate ex nihilo.

## Authorship
Myself.

## Changelog

### DIFF
--- a/code/modules/butchery/butchery_products.dm
+++ b/code/modules/butchery/butchery_products.dm
@@ -16,7 +16,7 @@
 	var/fat_material    = /decl/material/solid/organic/meat/gut
 	var/meat_name       = "meat"
 
-/obj/item/food/butchery/Initialize(ml, material_key, mob/living/donor)
+/obj/item/food/butchery/Initialize(mapload, material_key, skip_plate = FALSE, mob/living/donor)
 	var/decl/butchery_data/butchery_decl = GET_DECL(donor?.butchery_data)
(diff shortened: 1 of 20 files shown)


code/modules/butchery/butchery_products.dm
```diff
@@ -16,7 +16,7 @@
 	var/fat_material    = /decl/material/solid/organic/meat/gut
 	var/meat_name       = "meat"
 
-/obj/item/food/butchery/Initialize(ml, material_key, mob/living/donor)
+/obj/item/food/butchery/Initialize(mapload, material_key, skip_plate = FALSE, mob/living/donor)
```
out-of-phaze, line 19: are you sure the `= FALSE` doesn't make it get treated as a keyword argument? is there a reason it's all `= FALSE` instead of just letting it be null?

MistakeNot4892: Per Discord I don't understand what you're asking or what you mean by keyword argument. It could be null but I wanted to be explicit that it was a boolean, like tons of other parameters (`silent`, `update_icons` etc).

out-of-phaze (approved):
well if it doesn't break then presumably it's fine and I was misremembering

PR #4321, opened 2024-08-10
Adding a signpost to despawn from Shaded Hills.

MistakeNot4892:
## Description of changes
- Breaks some cryopod despawn code out into separate procs.
- Adds a signpost effect to Shaded Hills to despawn your character.
 
Adds a way to clear a slot from maps with no cryopods.

## Authorship
Myself.

## Changelog
:cl:
add: Added signposts to despawn from Shaded Hills.
/:cl:


code/modules/mob/living/living.dm
```diff
@@ -1800,3 +1800,25 @@ default behaviour is:
 			return FALSE
 
 	return TRUE
+
+/mob/living/proc/prepare_for_despawn()
+	//Update any existing objectives involving this mob.
```
out-of-phaze, line 1805: maybe warrants a `SHOULD_CALL_PARENT(TRUE)` in the future

out-of-phaze (approved):
noting that any equipped items will just vanish, but that's fine for a prototype. potential alternatives to a cryo control computer could be, like, a 'lost and found' bulk storage box somewhere in town?

PR #4323, opened 2024-08-10
Make cooking recipes that use stacks not consume the entire stack

out-of-phaze:
## Description of changes
bone broth would consume an entire stack of bones instead of just one bone. this is sad! i'm sad. lets fix that!

cooking recipes are now stack-aware. i did my best to handle as many edge-cases as i could think of


MistakeNot4892 (approved):
Seemed to work fine on Pyrelight testing.

PR #4331, opened 2024-08-12
Fixing issues with traits not serializing due to UID.

MistakeNot4892:
- Enforces UIDs on traits.
- Enforces UIDs on bodytypes.


MistakeNot4892:
I was originally using bodytype to generate UID for prosthetic traits but couldn't use that due to the decl repo relying on initial(), so the bodytype UIDs are technically unneeded, but it seems to be fine as a change.

out-of-phaze (approved):

PR #4332, opened 2024-08-12
Accessories can now check traits for availability.

MistakeNot4892:
Needed for some free-range foxperson stuff on Pyrelight.


out-of-phaze (approved):
feels a bit overly-specific but i don't really see a good reason to object over that. code seems fine

PR #4334, opened 2024-08-13
Fixes from August 10, 2024 Pyrelight testing

out-of-phaze:
## Description of changes
- Fixes a runtime with null sprite accessory metadata.
- Allows psychotropics, amatoxin (and carpotoxin, because why not) to be composted.
- Moves simple animal attack effect triggering to natural weapon hit effects.
- Torches now only become 'burnt' and fail to relight if they run out of fuel.
- Torches drenched in blood will now fail to light until they are cleaned.
- Fixes some issues with prosthetics traits to allow wooden prosthetics to be picked on Shaded Hills.
- Sparks are now light sources.
- Sparks only last for 2 seconds instead of 5.
- Modifies unit testing for material temperature transitions.
- Temperature damage threshold is no longer set by ignition point (pending atom fires) or baking temperature. We enforce that baking temperature and ignition point are less than the damage threshold in `/decl/material/validate()` now.
- Simplifies temperature damage threshold autosetting to just use `max()` with more than two arguments.
- Automatic temperature damage thresholds will only be set if they're greater than the ignition point/it has no ignition point.

The intent with the torch coating check is that being doused in water will immediately call `clean()`, which removes the water coating. I don't necessarily want to touch that in this PR, need to have a bit of a discussion on how to redo coatings/cleaning like that, maybe as its own PR

- Fixes a runtime.
- Compost bins no longer fill up with psychoactive compounds.
- Spiders no longer inject venom even on miss.
- Torches can be relit if manually quenched/extinguished with fuel left.
- Lays the basic groundwork for torches not being able to be relit until dried off after you fall into deep water.
- Wooden prostheses can be selected on Shaded Hills.
- Base prosthetic traits now show what model they use, based on selected species.
- Sparks now produce light.
- Sparks no longer appear to hang in midair or get 'stuck' for a few seconds.
- Improves material temperature transition unit testing and auto-setting.
- Clay and logs will no longer melt just because they have bake/ignition temperatures set.

## Authorship
Me.

## Changelog
:cl:
add: sparks now produce light
add: wooden prostheses can be selected on Shaded Hills
tweak: amatoxin and psychotropics can be composted in a composter
/:cl:

### DIFF
--- a/code/datums/traits/prosthetics/prosthetic_limbs.dm
+++ b/code/datums/traits/prosthetics/prosthetic_limbs.dm
@@ -3,7 +3,7 @@
 	trait_cost = 1
 	category = "Prosthetic Limbs"
 	available_at_chargen = TRUE
-	available_at_map_tech = MAP_TECH_LEVEL_SPACE
+	available_at_map_tech = MAP_TECH_LEVEL_ANY // the base trait must be available so that wooden prostheses are available
 	abstract_type = /decl/trait/prosthetic_limb
 	reapply_on_rejuvenation = TRUE
 	var/fullbody_synthetic_only = FALSE
@@ -20,6 +20,18 @@
 	var/decl/species/species = species_name ? get_species_by_key(species_name) : global.using_map.default_species
 	return species?.base_external_prosthetics_model
 
+/decl/trait/prosthetic_limb/get_chargen_name(datum/preferences/pref)
+	var/decl/bodytype/prosthetic/our_model = get_base_model(pref.species)
+	if(!model && our_model.name)
+		return "[capitalize_words(our_model.name)] [bodypart_name]"
+	return ..()
+
+/decl/trait/prosthetic_limb/get_chargen_desc(datum/preferences/pref)
+	var/decl/bodytype/prosthetic/our_model = get_base_model(pref.species)
+	if(!model && our_model.name)
+		return "You have been fitted with [ADD_ARTICLE(our_model.name)] [lowertext(bodypart_name)] prosthesis."
+	return ..()
+
 /decl/trait/prosthetic_limb/Initialize()
 	. = ..()
 	// Macro can generate float costs, round to closest point value.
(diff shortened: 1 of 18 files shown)


code/datums/traits/prosthetics/prosthetic_limbs.dm
```diff
@@ -3,7 +3,7 @@
 	trait_cost = 1
 	category = "Prosthetic Limbs"
 	available_at_chargen = TRUE
-	available_at_map_tech = MAP_TECH_LEVEL_SPACE
+	available_at_map_tech = MAP_TECH_LEVEL_ANY // the base trait must be available so that wooden prostheses are available
```
MistakeNot4892, line 6: This might be overridden by the prosthesis model. I guess that's not an issue really since we're setting it to wood on Shadyhills.

code/modules/client/preference_setup/general/02_body.dm
```diff
@@ -257,7 +257,7 @@
 				var/list/metadata_strings = list()
 				for(var/metadata_type in accessory_decl.accessory_metadata_types)
 					var/decl/sprite_accessory_metadata/sam = GET_DECL(metadata_type)
-					metadata_strings += sam.get_metadata_options_string(src, accessory_cat_decl, accessory_decl, accessory_metadata[metadata_type])
+					metadata_strings += sam.get_metadata_options_string(src, accessory_cat_decl, accessory_decl, LAZYACCESS(accessory_metadata, metadata_type))
```
MistakeNot4892, line 260: I swear I did this already. Maybe on staging.

MistakeNot4892 (approved):

code/datums/traits/prosthetics/prosthetic_limbs.dm
out-of-phaze: it is only overridden if `model` is non-null actually

code/modules/client/preference_setup/general/02_body.dm
out-of-phaze: i think you fixed a different tails runtime? maybe?

PR #4337, opened 2024-08-13
Make skills not subtypes of hierarchy, redo Shaded Hills skill categories

out-of-phaze:
## Description of changes
- Makes skills not a subtype of `/decl/hierarchy`.
- Separates skill categories and skills into separate types.
- Makes skill category header rows stretch the full length of the skill table.
- Redoes Shaded Hills skill categories.

I would appreciate a proper review rather than a speedmerge, I'm a little skittish about how well this seems to work...

Allows skill categories to be overridden modularly.
Improves Shaded Hills skill categories for future Pyrelight work regarding anima.

## Authorship
Me.


code/modules/client/preference_setup/occupation/skill_selection.dm
```diff
@@ -66,9 +66,9 @@
 		if("[job.type]" in pref.skills_saved)
 			var/S = pref.skills_saved["[job.type]"]
 			var/L = list()
-			for(var/decl/hierarchy/skill/skill in global.using_map.get_available_skills())
-				if("[skill.type]" in S)
-					L[skill] = S["[skill.type]"]
+			for(var/decl/skill/skill in global.using_map.get_available_skills())
+				if("[skill.uid]" in S)
```
MistakeNot4892: Unneeded stringification here.

code/modules/client/preference_setup/occupation/skill_selection.dm
```diff
@@ -188,25 +188,22 @@
 
 	dat += "<table>"
 	var/list/available_skills = global.using_map.get_available_skills()
-	var/decl/hierarchy/skill/skill = GET_DECL(/decl/hierarchy/skill)
-	for(var/decl/hierarchy/skill/cat in skill.children)
-
+	var/list/categories = decls_repository.get_decls_of_subtype_unassociated(/decl/skill_category)
+	categories = dd_sortedObjectList(categories)
```
MistakeNot4892: Aren't we supposed to be using sortTim() by preference these days?

code/modules/client/preference_setup/occupation/skill_selection.dm
```diff
@@ -188,25 +188,22 @@
 
 	dat += "<table>"
 	var/list/available_skills = global.using_map.get_available_skills()
-	var/decl/hierarchy/skill/skill = GET_DECL(/decl/hierarchy/skill)
-	for(var/decl/hierarchy/skill/cat in skill.children)
-
+	var/list/categories = decls_repository.get_decls_of_subtype_unassociated(/decl/skill_category)
```
MistakeNot4892: Non-blocking thought, maybe the decls repo should sort itself on instance creation.

maps/~mapsystem/map_skills.dm
```diff
@@ -4,14 +4,14 @@
...
+	for(var/skill_type in decls_repository.get_decls_of_type(/decl/skill))
 		. += skill_type
 
 /datum/map/proc/build_available_skills()
 	_available_skills = list()
 	for(var/skill_type in get_available_skill_types())
-		var/decl/hierarchy/skill/skill = GET_DECL(skill_type)
-		if(INSTANCE_IS_ABSTRACT(skill))
+		var/decl/skill/skill = GET_DECL(skill_type)
+		if(!skill) // abstract, didn't instantiate
```
MistakeNot4892: Should we just let this runtime? It implies something incorrect is in `get_available_skill_types()`.

code/modules/client/preference_setup/occupation/skill_selection.dm
out-of-phaze: yeeeeah but it already used dd_SortValue so...

out-of-phaze:
On hold, need to do this in steps. Serialization has to be changed to use uids first so that people on Pyrelight, etc. don't have to redo their skills.

out-of-phaze:
Just went with a different method of migration, worked fine in testing

MistakeNot4892 (dismissed):

MistakeNot4892 (approved):

PR #4338, opened 2024-08-13
Allows liquids to boil off from open containers.

MistakeNot4892:
## Description of changes
- `/obj/item/chems` has additional code in `ProcessAtomTemperature()` for any open container to pass any contents that are above boiling point to loc via a new proc.
- Other `/obj/item/chems` that receive these reagents will store them if the atom is closed.
- Fire sources will show a more detailed list of descriptors for how hot they are burning.
- Fire sources other than firepits can be controlled or extinguished via an Adjust Draught alt interaction.
- Adds a wood burner and retort for medieval chemistry.

Prep work for medieval chemistry downstream, and generally nice to have for stuff like soup.

## TODO
- [x] Get draughts to modify fire source temperature properly to allow for precise temperature control.
- [x] Ensure the wood burner can separate ethanol and water.
- ~~Ensure the stove can separate water and soup.~~

## Authorship
Myself for some of the alembic and retort code.
@PlayerDeer for the base icons for the retort and alembic.
@noelle-lavenza for reworking temperature so that this actually does what we want it to.

## Changelog
:cl:
tweak: Open containers, when heated, can boil off their liquid contents.
/:cl:


MistakeNot4892:
Calling this done as it achieves the goal (letting ethanol get extracted from water).

code/modules/admin/view_variables/view_variables.dm
```diff
@@ -8,6 +8,9 @@ var/global/list/view_variables_no_assoc = list("verbs", "contents","screen","ima
 	set category = "Debug"
 	set name = "View Variables"
 
+	debug_variables_inner(D)
+
+/client/proc/debug_variables_inner(datum/D, filter = "" as null|text)
```
out-of-phaze, line 13: should probably have a comment explaining why this was done, i believe it's because otherwise using the VV verb would would prompt you to input a filter string

in fact it might be worth pulling the VV changes out into a separate PR so that it can actually be reviewed (by loaf)

code/modules/crafting/stack_recipes/recipes_bricks.dm
```diff
@@ -42,6 +42,9 @@
 /decl/stack_recipe/bricks/fire_source/kiln
 	result_type                 = /obj/structure/fire_source/kiln
 
+/decl/stack_recipe/bricks/fire_source/alembic
+	result_type                 = /obj/structure/fire_source/heater
```
out-of-phaze, line 48: should this need a higher skill requirement than just the default? presumably it should be more difficult to make than a kiln, given it has an adjustable draught and all

code/modules/food/cooking/cooking_vessels/_cooking_vessel.dm
```diff
@@ -150,3 +150,4 @@
 	. = ..()
 	if(. != PROCESS_KILL && !is_processing)
 		START_PROCESSING(SSobj, src)
+	update_icon()
```
out-of-phaze: this feels kind of gross given that it'll be an icon update every 2 seconds, but cooking vessels are relatively light in terms of icon stuff compared to humantype mobs and similar, so i figure it's probably fine. would still like if there was a way to more intelligently update stuff re: state transitions

actually, with nata's PR, will solid reagents be moved into the liquid list when they reach their melting point? and when water starts boiling it'll start to vaporize and cause a reagent change which will cause an icon update. are you certain this update_icon() call is still necessary?

code/modules/materials/definitions/solids/materials_solid_wood.dm
```diff
@@ -43,12 +43,14 @@
...
 	sound_dropped = 'sound/foley/wooddrop1.ogg'
 	compost_value = 0.2
+	temperature_burn_milestone_material = TRUE
 
 /decl/material/solid/organic/wood/fungal
 	name = "towercap"
 	uid = "solid_wood_fungal"
 	color = "#e6d8dd"
 	hardness = MAT_VALUE_FLEXIBLE + 10
+	temperature_burn_milestone_material = FALSE
```
out-of-phaze: thought: `temperature_burn_milestone_material` could be a typepath instead to avoid inheritance causing issues? so if `src.type == src.temperature_burn_milestone_material` it'd be a milestone material

code/modules/reagents/chems/chems_ethanol.dm
```diff
@@ -9,23 +9,27 @@
...
+	chilling_point = T0C
 	chilling_products = list(
 		/decl/material/liquid/ethanol = 0.75,
-		/decl/material/liquid/water =   0.25
+		/decl/material/solid/ice =      0.25
 	)
 	bypass_chilling_products_for_root_type = /decl/material/liquid/ethanol
 	affect_blood_on_ingest = FALSE // prevents automatic toxins/inebriation as though injected
 
+	temperature_burn_milestone_material = TRUE
```
out-of-phaze: this is not unset on subtypes of ethanol, i think the only reason it doesn't explode is because they don't have a boiling point set. if that ever changes it will say "it is burning hot enough to boil [every single subtype of ethanol]" and that's... unpleasant

my suggestion is either set it to false in the Initialize override below, or go with the prior suggestion of making `temperature_burn_milestone_material` a typepath instead of a bool.

code/modules/reagents/chems/chems_ethanol.dm
```diff
@@ -36,6 +40,12 @@
 	glass_desc = "A well-known alcohol with a variety of applications."
 	value = 1.2
 
+/decl/material/liquid/ethanol/Initialize()
+	. = ..()
+	// Impure ethanol doesn't boil, it has to separate first.
+	if(type != bypass_heating_products_for_root_type)
```
out-of-phaze, line 44: `bypass_heating_products_for_root_type` is only ever used for ethanol, and presumably it exists to avoid exactly this initialize override, so it could probably be changed to just a bespoke `type != /decl/material/liquid/ethanol` check that just unsets all the relevant vars here. idk, just a nitpick, i don't really like that this (and the chilling products counterpart) is so niche and only ever used for ethanol...

feel free to ignore this one, though, just idle musing

code/modules/reagents/reactions/reaction_herbal.dm
```diff
@@ -1,7 +1,7 @@
 /decl/chemical_reaction/drug/herbal
 	abstract_type       = /decl/chemical_reaction/drug/herbal
 	result_amount       = 2
-	minimum_temperature = 100 CELSIUS
+	minimum_temperature = 50 CELSIUS
```
out-of-phaze: i don't thiiink there was a specific reason 50C was chosen aside from the fact that it's just halfway to 100C. i feel like maybe the tea ones should require 100C (or close to it) still, and the tincture ones should have a lower temperature. i mean i guess there's no reason it can't just be arbitrary, this isn't really a big deal, just thoughts

code/modules/reagents/reagent_containers.dm
```diff
@@ -180,6 +180,35 @@
...
+		if(!isnull(mat.boiling_point) && temperature >= mat.boiling_point)
+			// TODO: reduce atom temperature?
+			var/removing = min(5, reagents.reagent_volumes[reagent])
+			reagents.remove_reagent(reagent, removing, defer_update = TRUE, removed_phases = MAT_PHASE_LIQUID)
+			loc.take_vaporized_reagent(reagent, removing)
+			. = null // Don't return PROCESS_KILL, we may have more to vaporise
+	reagents.handle_update()
+
+/obj/item/chems/take_vaporized_reagent(reagent, amount)
+	// TODO: check ATOM_IS_OPEN_CONTAINER(src) when there is a closed container for alembic vessels
```
out-of-phaze: as you mentioned on discord, this comment may not apply anymore? should look into that

out-of-phaze (dismissed):
some requested changes, i can't approve once this is changed but figured i'd offer feedback regardless.

also, i'm not suuuuuper happy with the burn strings solution, i'd _personally_ prefer a mercury thermometer or some other method of determining temperature (like, yknow, checking to see if water boils to know that it's over 100C, etc), but regardless it's a decent solution for the time being and i'm willing to make that compromise in the name of usability/player convenience

code/modules/crafting/stack_recipes/recipes_bricks.dm
MistakeNot4892: Everything except firepits has an adjustable draught.

code/modules/food/cooking/cooking_vessels/_cooking_vessel.dm
MistakeNot4892: I don't believe reagents move between lists currently. I could make it only trigger an update if boiling status changes I guess.

code/modules/reagents/chems/chems_ethanol.dm
MistakeNot4892: Typepath suggestion looks good to me.

code/modules/reagents/chems/chems_ethanol.dm
MistakeNot4892: Yeah I'm not touching that in this PR sorry.

code/modules/reagents/reactions/reaction_herbal.dm
MistakeNot4892: I think 'any heat source at all' is probably not a bad baseline. Increasing it makes it more annoying to find the sweet spot with the draught.

code/modules/crafting/stack_recipes/recipes_bricks.dm
MistakeNot4892: Bumped it up to NORMAL for kilns and heaters.

code/modules/admin/view_variables/view_variables.dm
MistakeNot4892: Per Discord, that seems like a good idea to me.

code/modules/food/cooking/cooking_vessels/_cooking_vessel.dm
MistakeNot4892: I moved the logic onto the cooking pot specifically and gated it behind temperature and boiling state changes. I am not sure about the phase transition aspect of the new reagent system.

code/modules/materials/definitions/solids/materials_solid_wood.dm
MistakeNot4892: Done.

comma (approved):

out-of-phaze:
chinsky's review is 24h, merging since we've waited long enough

PR #4343, opened 2024-08-14
Added some supporting code to allow equipment to modify your character preview.

MistakeNot4892:
This is a bit specific, but it's also fairly invasive so figured I would upstream it.

Specifically it's needed downstream to allow shapeshifter masks to add the transformed preview to the character screen.

### DIFF
--- a/code/modules/client/preference_setup/loadout/loadout.dm
+++ b/code/modules/client/preference_setup/loadout/loadout.dm
@@ -59,6 +59,11 @@
 		if(!found_species)
 			return FALSE
 
+	if(required_traits)
+		for(var/trait in required_traits)
+			if(!(trait in pref.traits))
+				return FALSE
+
 	if(faction_restricted)
 		var/has_correct_faction = FALSE
 		for(var/token in ALL_CULTURAL_TAGS)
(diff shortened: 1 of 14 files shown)


out-of-phaze (dismissed):
pretty niche but pretty low-impact, so it seems fine

MistakeNot4892:
This snowballed a bit.
- Adds a preload step to save loading to fix issues with the trait list not being populated when body first sanitizes.
- Adds a required_traits list to loadout options, and a flag for modifying an existing item preferentially.
- Dummy tails are force set to owner species and bodytype to avoid issue with whitelist checks on tail accessories getting the map default species.

code/modules/client/preference_setup/loadout/loadout.dm
```diff
@@ -521,7 +553,12 @@
 /decl/loadout_option/proc/spawn_and_validate_item(mob/living/human/H, metadata)
 	PRIVATE_PROC(TRUE)
 
-	var/obj/item/item = spawn_item(H, H, metadata)
+	var/obj/item/item
+	if(apply_to_existing_if_possible)
+		item = locate(path) in H.get_equipped_items(include_carried = TRUE)
```
out-of-phaze: maybe a little out of scope but i feel that there should be support for more complicated predicates, so something like
```dm
		for(var/obj/item/candidate in H.get_equipped_items(include_carried = TRUE))
			if(can_replace_existing(candidate))
				item = candidate
				break
```
and as a default
```dm
/decl/loadout_option/proc/can_replace_existing(obj/item/candidate)
	return istype(candidate, path)
```

code/modules/client/preferences.dm
```diff
@@ -403,6 +403,10 @@ var/global/list/time_prefs_fixed = list()
 
 	character.backpack_setup = new(backpack, backpack_metadata["[backpack]"])
 
+	if(length(traits))
+		for(var/trait_type in traits)
+			character.set_trait(trait_type, (traits[trait_type] || TRAIT_LEVEL_EXISTS))
```
out-of-phaze, line 408: nothing that this is probably fine as limbs/organs have been created by this point. i was initially a little worried about amputation traits etc but it seems a-okay

code/modules/sprite_accessories/tails/_accessory_tail.dm
```diff
@@ -10,7 +10,12 @@
 	// Give us a tail if we need one.
 	var/decl/sprite_accessory/tail_data = GET_DECL(accessories[1])
 	if(tail_data?.draw_accessory && !character.get_organ(BP_TAIL))
-		character.add_organ(new /obj/item/organ/external/tail(null, character.get_mob_snapshot()), null, TRUE, FALSE, FALSE, TRUE)
+		var/obj/item/organ/external/tail/new_tail = new(null, character.get_mob_snapshot())
+		// Unclear why the mob snapshot isn't setting these, but
+		// somehow the tail is ending up as the map default species.
+		new_tail.set_species(character.get_species_name())
```
out-of-phaze: I feel like this should probably be investigated further rather than just slapping a stopgap on there...

out-of-phaze (changes requested):
only the last comment about the stopgap for mob_snapshot not setting tail species is actually blocking, the first one is a suggestion and the second one is a comment

code/modules/sprite_accessories/tails/_accessory_tail.dm
```diff
@@ -10,7 +10,12 @@
...
+		new_tail.set_bodytype(character.get_bodytype())
```
out-of-phaze: image
womp womp, `material_key` missing.
```suggestion
		var/obj/item/organ/external/tail/new_tail = new(null, null, character.get_mob_snapshot())
```

out-of-phaze (changes requested):

code/modules/client/preferences.dm
MistakeNot4892: Yeah this actually speeds up accessories a bit by skipping applying to bodyparts that would later be amputated or made prosthetic.

out-of-phaze (approved):

PR #4344, opened 2024-08-14
[IDB IGNORE] Converting a bunch of existing soup items into bowls with soup in them.

MistakeNot4892:
## Description of changes
- Moves curries, soups, stews and chili onto the soup system.
- Moves food segmenting down to /obj/item.
- Added soup bowls.
- Added mapped soup bowls with predefined soup contents for previously mapped soup items.
- Removes milo (miso?) soup for now. Left a comment in a todo list for future implementation when more ingredients are added.

## TODO
- [X] Test soups.
- ~~Get soups looking presentable~~.
- ~~Get soup to calculate `mask_color` from ingredients/mix data~~.

Leaving the remaining cosmetic work for a followup PR.

- More variety in things you can make out of arbitrary meat and growns.
- Less bespoke food code.

## Authorship
Myself for a lot of the code/icon work.
@noelle-lavenza for testing and for writing a unit test to check the soup recipes. Saved me probably hours of manual testing, thank you.

## Changelog
:cl:
tweak: Curries, soups and stews have been rewritten, please refer to the codex for recipes.
/:cl:


MistakeNot4892:
image
Some progress.

MistakeNot4892:
If anyone wants to pull this down and test the soup recipes in the todo I'd appreciate the assistance, otherwise I will likely get through this tomorrow evening.

MistakeNot4892:
I did not get through this tomorrow evening.

code/modules/food/cooking/recipes/recipe_soup_simple.dm
```diff
@@ -0,0 +1,65 @@
...
+	abstract_type      = /decl/recipe/soup/simple/milk
+	reagents           = list(/decl/material/liquid/drink/milk = 10)
+	precursor_type     = null
+
+/decl/recipe/soup/simple/milk/get_result_data(atom/container, list/used_ingredients)
+	. = ..()
+	// Consider prefixing 'cream of' for milk soups?
+	LAZYINITLIST(.)
+	.[DATA_TASTE]            |= list("cream" = 1)
+	.[DATA_INGREDIENT_FLAGS] |= ALLERGEN_DAIRY
```
noelle-lavenza: This runtimes because you try to `|=` null, I guess. The following code fixed it in testing but there might be a more idiomatic way to do it?
```suggestion
	var/cream = LAZYACCESS(.[DATA_TASTE], "cream")
	LAZYSET(.[DATA_TASTE], "cream", cream + 1)
	LAZYDISTINCTADD(.[DATA_INGREDIENT_FLAGS], ALLERGEN_DAIRY)
```
```

File: code/modules/food/cooking/recipes/recipe_soup.dm
Comment (noelle-lavenza) [inline code/modules/food/cooking/recipes/recipe_soup.dm:1]: The types that were split out into separate files (simple, stew, stock, etc.) still exist here and cause typepath conflicts.

**Review by noelle-lavenza — CHANGES_REQUESTED** [review]
Hi I wrote a food unit test to test this, PR soon

File: code/modules/reagents/reactions/reaction_compounds.dm [hunk 2]
```diff
@@ -146,3 +146,15 @@
 	name = "Sulfuric Acid"
 	required_reagents = list(/decl/material/solid/sulfur = 1, /decl/material/solid/sodium = 2)
 	result_amount = 3
+
+/decl/chemical_reaction/compound/condensed_capsaicin
```
noelle-lavenza, line 150: is there a reason this doesn't use heating_products anymore?

MistakeNot4892: Making chili caused it to happen.

noelle-lavenza: Can you not just increase the heating product temperature?

MistakeNot4892: Not really? I would need to increase it beyond the max range of fire sources, which is hundreds of degrees. We could potentially make heating products only separate out when not in a grown, but idk.

noelle-lavenza: But fire sources have a draught. I guess maybe you wouldn't be able to cook soup on a campfire then? The other thing is like, technically a pot full of water shouldn't be able to heat above the boiling temperature of water, but that's kind of a pain to check.
Also there's no inhibitors set for this so I don't know how this is any different than heating products, do growns have NOREACT or something? I guess they must to avoid being able to inject water into a banana for a quick explosion

MistakeNot4892: Most growns don't have heating products, so it doesn't come up.

MistakeNot4892:
````
Also there's no inhibitors set for this so I don't know how this is any different than heating products, do growns have NOREACT or something? I guess they must to avoid being able to inject water into a banana for a quick explosion
````
Misread this, sorry. I thiiiiink I used a reaction because you can't set a maximum temperature on a heating product transition. If you could, yeah, the draught would handle this.

Comment (noelle-lavenza) [reply code/modules/reagents/reactions/reaction_compounds.dm:150]: Oh, I missed the max temperature bit. Huh. Fair enough

**Comment by MistakeNot4892** [comment]
Depends on #4525

**Comment by MistakeNot4892** [comment]
I am honestly willing to cop soups looking kind of crap for the sake of this getting merged.

File: code/__defines/reagent_data_fields.dm [hunk 3]
```diff
@@ -0,0 +1,60 @@
+// Currently just used for indexing reagent data without
+// having to use strings. UID is provided for serde.
+
+/decl/reagent_data_field
```
out-of-phaze, line 4: not super sure why this is preferable to string defines? if it works it works though

edit: obviously not blocking

code/modules/food/cooking/_recipe.dm
```diff
@@ -37,7 +42,7 @@ var/global/list/_cooking_recipe_cache = list()
 	var/display_name              // Descriptive name of the recipe, should be unique to avoid codex pages being unsearchable. If not set, codex uses initial name of product.
 	var/list/reagents             // example: = list(/decl/material/liquid/drink/juice/berry = 5) // do not list same reagent twice
 	var/list/items                // example: = list(/obj/item/crowbar, /obj/item/welder, /obj/item/screwdriver = 2) // place /foo/bar before /foo
-	var/list/fruit                // example: = list("fruit" = 3)
+	var/list/fruit                // example: = list(RECIPE_COMPONENT_FRUIT = 3)
```
out-of-phaze:
```suggestion
	var/list/fruit                // example: = list("tomato" = 3)
```

code/modules/reagents/reagent_containers/food.dm
```diff
@@ -71,7 +71,21 @@
 		reagents.maximum_volume = max(reagents.maximum_volume, volume)
 	return ..()
 
-/obj/item/chems/on_reagent_change()
+// Dummy type used solely for soup bowls/soup spoons.
+/obj/item/food/lump
```
out-of-phaze, line 75: a little worried in the future that this will result in you scooping up soup as a food lump on a spoon, then the lump immediately melting because it's liquid at room temperature, soaking you and your spoon in soup. but for now it's probably fine, even if it miiiight be better to use `reagents` in the long run for reagent transfer.

edit: also obviously not blocking

out-of-phaze (changes requested):

code/__defines/reagent_data_fields.dm
MistakeNot4892: Doesn't need string creation mostly, I also have the vague idea that we can do data validation with the decl at some point.

code/modules/reagents/reagent_containers/food.dm
MistakeNot4892: This is mainly future proofing for foods using matter down the track, will need some review for reagent based stuff at that point though yeah.

out-of-phaze (approved):

code/__defines/reagent_data_fields.dm
out-of-phaze: string creation happens once and if the string is compiled in its fine, it's only the creation of new strings that causes string tree rebalancing. that said, a set_data proc that does validation would be nice

PR #4353, opened 2024-08-14
Make wild plants have a lower chance of harvestable produce

### LINKED ISSUE #48: Wild plant yield is too high
it seems like maybe wild plant yield should be lowered a lot, we have toooons of plants without even doing any farming.

out-of-phaze:
## Description of changes
Lowers the chance of harvestable fruit from wild plants by about half.

fixes broken examine info on plants, `length(harvested)` is invalid because `harvested` is just a number

Incentivizes actually doing farming instead of just gathering a ton of wild plants.
also fixes a bug with the examine info thing

## Authorship
me

## Changelog
:cl:
balance: lowered the chance of wild plants bearing harvestable fruit by about half, to incentivize farming
/:cl:

### DIFF
--- a/code/game/objects/structures/flora/plant.dm
+++ b/code/game/objects/structures/flora/plant.dm
@@ -30,8 +30,8 @@
 	. = ..()
 	if(dead)
 		to_chat(user, SPAN_OCCULT("It is dead."))
-	else if(length(harvestable))
-		to_chat(user, SPAN_NOTICE("You can see [length(harvestable)] harvestable fruit\s."))
+	else if(harvestable)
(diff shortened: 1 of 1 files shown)


MistakeNot4892:
I'm not sure about this one. Foraging is a limited resource, if we have a more active round I can see it running out pretty quick even without this adjustment. Is there a particular reason to incentivise farming more than it already is?

MistakeNot4892 (approved):
Per Discord, just remembered you can get seeds from unharvestable plants so my concerns aren't really a big deal.

PR #4366, opened 2024-08-16
Lowers base move delay on simple animals across the board to compensate for move intent decl delay.

MistakeNot4892:


code/modules/mob/living/simple_animal/_simple_animal.dm
```diff
@@ -17,7 +17,7 @@
 		/decl/move_intent/run/animal
 	)
 
-	var/base_movement_delay = 4
+	var/base_movement_delay = 0 // Added to the delaye expected from movement decls.
```
out-of-phaze:
```suggestion
	var/base_movement_delay = 0 // Added to the delay expected from movement decls.
```

out-of-phaze (dismissed):

out-of-phaze (dismissed):

out-of-phaze (approved):
24h with cornflake resolution

PR #4357, opened 2024-08-15
Reworks fox and deer icons.

MistakeNot4892:
Greyscales @PlayerDeer's fox icon and breaks the markings/socks/eyes out into overlays.

image

PR #4368, opened 2024-08-17
Greyscaling deer icons.

MistakeNot4892:
Similar to #4357 but for deer.


LowFP:
RIP

MistakeNot4892:
I just merged it into the fox one since there were some other changes.

PR #4386, opened 2024-08-21
Optimize unsimulated turfs

out-of-phaze:
## Description of changes
Changes split out from #4186 because they were contentious enough to block that PR for months.

Optimizes unsim turfs by making them skip most Initialize logic because, well, they don't need it.

## Authorship
Me.


MistakeNot4892 (approved):
I don't like the use of New() but I don't care that much about unsim turfs ultimately.

PR #4387, opened 2024-08-21
Optimize floor decals

out-of-phaze:
## Description of changes
Changes split out from #4186 because they were contentious enough to block that PR for months.

Need to find a way to make decals layer over natural turfs without preventing you from building on them (`is_plating` is insufficient).

Optimizes floor decals by making them skip Initialize and Destroy logic because, well, they don't need it.
Fixes a bug where floor decals wouldn't show up on natural turfs.

## Authorship
Me.

### DIFF
--- a/code/game/atoms_init.dm
+++ b/code/game/atoms_init.dm
@@ -9,6 +9,7 @@
 		if(!istype(storage))
 			storage = null
 
+	// This preloader code is duplicated in /obj/effect/floor_decal/Initialize(). If you change this, be sure to change it there, too.
 	//atom creation method that preloads variables at creation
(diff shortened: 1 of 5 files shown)


MistakeNot4892:
Once the other two opt PRs are in, I'd like to benchmark this to get an idea of how much benefit there actually is, given the Discord discussion.

out-of-phaze:
so on exodus with like thousands of floor decals, it's only 0.3 seconds out of 51 during init. that said, something needs to be done about decals not showing up on turfs with only base flooring. i think maybe it should check if the topmost flooring allows decals and `/decl/flooring/plating` should just say "no"

basically: legitimate salvage for that particular aspect

PR #4391, opened 2024-08-21
Add animal husbandry to Shaded Hills

out-of-phaze:
## Description of changes
Adds an animal husbandry skill to Shaded Hills used in the following interactions, using the botany skill as a fallback on the main map unless otherwise specified:
- Examining baby chicks now gives you an estimate of their age.
- Examining a held egg while holding a light source now lets you determine if the egg is fertile, and how far along the egg's development is.
- Cows will now flee from someone who fails a skill check when attempting to milk them. If they fail repeatedly, the cow will get stressed and run away.
- Cows cannot be milked while attempting to flee, except if the milker has max level in the skill used for milking, in which case they will stop trying to move.
- Goats will now become hostile against someone (who is not a friend) who fails a skill check when attempting to milk them.
- Goats cannot be milked by someone they're hostile to, except if the milker has max level in the skill used for milking, in which case they will stop being enemies with the milker.
- Passive animals now have a skill-based chance to flee when picked up on Shaded Hills. They will always flee if an enemy attempts to pick them up, and will never flee if a friend picks them up.
- Chickens lay fewer eggs less often that are fertilized more often but grow slower as both eggs and chicks.

Gives farmers/foresters another advantage over randos when dealing with animals.

## Authorship
Me.

## Changelog
:cl:
add: Added an animal husbandry skill to Shaded Hills.
add: Added skill checks to milking animals, using the animal husbandry skill. On maps where that skill is unavailable, the botany skill is used instead.
add: Added candling eggs. If sufficiently skilled, you can examine a held egg while also holding a light source in order to determine if it has a chick inside it or not.
add: Examining a baby chick with at least basic botany (animal handling on Shaded Hills) skill will tell you how close the chick is to growing into a chicken.
add: Passive animals now have a skill-based chance to flee when picked up on Shaded Hills. They will always flee if an enemy attempts to pick them up, and will never flee if a friend picks them up.
balance: Chickens lay fewer eggs less often that are fertilized more often but grow slower as both eggs and chicks.
/:cl:

### DIFF
--- a/code/datums/ai/passive.dm
+++ b/code/datums/ai/passive.dm
@@ -54,3 +54,20 @@
 		if(istype(source))
 			flee_target = weakref(source)
 			update_targets()
+
+/datum/mob_controller/passive
+	var/decl/skill/scooping_skill // If overridden (on a subtype, on a map/downstream, etc) check this skill to see if scooping should succeed uncontested.
+	var/scooping_skill_req = SKILL_ADEPT
+
+/datum/mob_controller/passive/scooped_by(mob/living/initiator)
+	if(is_friend(initiator))
+		return TRUE
+	if(is_enemy(initiator) || (scooping_skill && initiator.skill_fail_prob(scooping_skill, 50, scooping_skill_req))) // scary, try to wriggle away
+		retaliate(initiator) // run! run like the wind!
+		if(!initiator.skill_fail_prob(SKILL_HAULING, 100, SKILL_EXPERT))
+			to_chat(initiator, SPAN_WARNING("\The [body] tries to wriggle out of your grasp, but you hold on tight!"))
+			return TRUE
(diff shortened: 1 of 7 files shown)


code/datums/ai/passive.dm
```diff
@@ -54,3 +54,20 @@
 		if(istype(source))
 			flee_target = weakref(source)
 			update_targets()
+
+/datum/mob_controller/passive
+	var/decl/skill/scooping_skill // If overridden (on a subtype, on a map/downstream, etc) check this skill to see if scooping should succeed uncontested.
+	var/scooping_skill_req = SKILL_ADEPT
+
+/datum/mob_controller/passive/scooped_by(mob/living/initiator)
```
MistakeNot4892, line 62: It might be fun to add this to grabs down the track.

MistakeNot4892 (approved):

PR #4396, opened 2024-08-23
Paper, stamps and scrolls rework.

MistakeNot4892:
## Description of changes
- Converts teleportation scrolls into a general scroll type with a teleportation scroll subtype.
- Converts paper items to use the single icon system.
- Converts stamps to use the single icon system.
- Adds an ITEM_FLAG_STAMP used to indicate an item can stamp paper.

- Allows scrolls to be used as generic paper (useful downstream).
- Simplifies cigarette rolling checks.
- Simplifies stamp checking.

## TODO
- [X] Convert scrolls to /paper.
- [X] Convert stamps/signet rings to single icon.
- [X] Convert paper to single icon.
- [x] Convert cigarette paper to single icon.

 ## Testing
- [x] Test basic paper (blank, written, crumpled)
- [x] Test scrolls (above + furling)
- [x] Test paper planes.
- [x] Test teleportation scroll.
- [x] Test stamps
- [x] Test talismans.

## Authorship
Myself.

## Changelog
Nothing player-facing.


mods/content/dungeon_loot/subtypes/maint.dm
```diff
@@ -151,8 +151,7 @@
 		/obj/item/stack/material/cardstock/mapped/cardboard/five,
 		/obj/item/poster,
 		/obj/item/newspaper,
-		/obj/item/paper/crumpled,
-		/obj/item/paper/crumpled/bloody,
```
MistakeNot4892, line 155: This type has been broken (missing icon) for ages.

code/modules/paperwork/paper.dm
```diff
@@ -394,10 +395,8 @@
 			interact(user, readonly = FALSE)
 		return TRUE
 
-	else if(istype(P, /obj/item/stamp) || istype(P, /obj/item/clothing/gloves/ring/seal))
-		apply_custom_stamp(
-			image('icons/obj/bureaucracy.dmi', icon_state = "paper_[P.icon_state]", pixel_x = rand(-2, 2), pixel_y = rand(-2, 2)),
-			"with \the [P]")
+	else if(P.item_flags & ITEM_FLAG_STAMP)
```
out-of-phaze: not super jazzed about using an entire item flag for this given that we're over 3/4 through our supply of them, i would much rather have like. a tool archetype a la pens, or an extension a la cells. feel like maybe we should establish what flags should be used for (things that are common or general enough that they will be useful for a large number of items) and what they shouldn't (anything that has niche functionality that could be pulled out into an extension, anything that could use per-item behavior changes, etc).

that said, this works for now, i understand if you'd rather not bother rewriting it again given all that's on your plate with the wizard rewrites

code/modules/paperwork/paper.dm
```diff
@@ -462,9 +461,26 @@
...
+		stamp.pixel_x = rand(-2, 2)
+	else if(offset_x != 0)
+		stamp.pixel_x = offset_x
+
+	if(isnull(offset_y))
+		stamp.pixel_y = rand(-2, 2)
+	else if(offset_y != 0)
+		stamp.pixel_y = offset_y
+
+	// TODO: alpha mask against paper.
```
out-of-phaze: can we not just use `blend_mode = BLEND_INSET_OVERLAY` on the stamp image? for reference
> [BLEND_INSET_OVERLAY](https://www.byond.com/docs/ref/index.html#/atom/var/blend_mode) overlays the icon, but masks it by the image being drawn on. This is pretty much not at all useful directly on the map, but can be very useful for an overlay for an atom that uses KEEP_TOGETHER (see [appearance_flags](https://www.byond.com/docs/ref/info.html#/atom/var/appearance_flags)), or for the [layering filter](https://www.byond.com/docs/ref/info.html#/{notes}/filters/layer).

code/modules/paperwork/paper.dm
```diff
@@ -87,10 +86,10 @@
 /obj/item/paper/on_update_icon()
 	. = ..()
 
+	icon_state = get_world_inventory_state()
 	if(is_crumpled)
-		icon_state = "scrap"
+		icon_state = "[icon_state]-scrap"
```
out-of-phaze, line 91: does scroll-scrap exist?

code/modules/paperwork/paper.dm
```diff
@@ -570,11 +586,48 @@ var/global/datum/topic_state/default/paper_state/paper_topic_state = new
 // Crumpled Paper
 ///////////////////////////////////////////////////
 /obj/item/paper/crumpled
-	name       = "paper scrap"
-	icon_state = "scrap"
+	name        = "paper scrap"
+	is_crumpled = TRUE
+
+// Stub type for moving teleportation scrolls into a modpack.
+/obj/item/paper/scroll
```
out-of-phaze, line 594: can you turn these into paper planes? do those turn back into normal paper? i would hope not

code/modules/paperwork/stamps.dm
```diff
@@ -1,55 +1,46 @@
...
-
-	var/list/stamp_types = typesof(/obj/item/stamp) - src.type // Get all stamp types except our own
 	var/list/stamps = list()
-
 	// Generate them into a list
-	for(var/stamp_type in stamp_types)
-		var/obj/item/stamp/S = new stamp_type
-		stamps[capitalize(S.name)] = S
-
+	for(var/stamp_type in subtypesof(/obj/item/stamp))
```
out-of-phaze: noooo. `typesof(/obj/item/stamp) - /obj/item/stamp/chameleon` is not equivalent to `subtypesof(/obj/item/stamp)`. this will exclude the deck chief stamp but include the chameleon stamp itself.
```suggestion
	for(var/stamp_type in typesof(/obj/item/stamp) - type) // exclude the chameleon stamp subtype
```

code/modules/paperwork/stamps.dm
```diff
@@ -1,55 +1,46 @@
...
-
-	if(user && (src in user.contents))
-
+	if(input_stamp && !QDELETED(user) && !QDELETED(src) && user.get_active_held_item() == src)
 		var/obj/item/stamp/chosen_stamp = stamps[capitalize(input_stamp)]
-
 		if(chosen_stamp)
-			SetName(chosen_stamp.name)
-			icon_state = chosen_stamp.icon_state
+			appearance = chosen_stamp
```
out-of-phaze, line 49: this still didn't work before, but i think the boss stamp won't work properly here. it'll say `boss's rubber stamp` instead of `[global.using_map.boss_name]'s' rubber stamp`.

out-of-phaze (changes requested):

code/modules/paperwork/paper.dm
MistakeNot4892: Should do, afaicr I just copied the scrap state from paper.

code/modules/paperwork/paper.dm
MistakeNot4892: Yes, and no.

code/modules/paperwork/stamps.dm
MistakeNot4892: Ah yep I misread it.

code/modules/paperwork/paper.dm
MistakeNot4892: Neat, I didn't know that was a thing.

code/modules/paperwork/paper.dm
MistakeNot4892: If we're not actually out of flags I don't really want to have to commit to writing a whole new extension to replace an istype check, honestly.

code/modules/paperwork/stamps.dm
MistakeNot4892: Augh. I really don't want to instantiate all these stamps for something so niche. Do we care if the centcomm stamp name is off?

out-of-phaze: can we... use the atom info repository for this?

out-of-phaze: yep. just do the appearance set and then `SetName(atom_info_repository.get_name_for(chosen_stamp))` probably

code/modules/paperwork/paper.dm
MistakeNot4892: Made it a bool on /obj/item.

code/modules/paperwork/paper.dm
MistakeNot4892: Done.

code/modules/paperwork/stamps.dm
MistakeNot4892: Fixed.

code/modules/paperwork/stamps.dm
MistakeNot4892: Fixed.

code/modules/paperwork/paper.dm
MistakeNot4892: Confirmed, it's in there.

code/modules/paperwork/paper.dm
out-of-phaze: i don't super like that to be honest, but it's whatever, i'll write an extension later

out-of-phaze (dismissed):

code/modules/paperwork/paper.dm
MistakeNot4892: Why does it need to be an extension? I cannot stress this enough, I am replacing a single typecheck.

MistakeNot4892: Swapped it to a tool quality after discussion on Discord.

out-of-phaze (approved):

PR #4400, opened 2024-08-23
Skintone defaults to the middle of the range.

MistakeNot4892:
Previous logic had `pref.skin_tone` as null unless explicitly set via the prefs Topic(), which meant an adjustment to the default skin tone that Polaris wanted to do was completely ignored.

Also included the Polaris adjustment because setting the default skintone to the middle of the range makes decent sense.

Ty @Cerebulon for pointing the bug out.

### DIFF
--- a/code/modules/client/preference_setup/general/02_body.dm
+++ b/code/modules/client/preference_setup/general/02_body.dm
@@ -1,10 +1,9 @@
 /datum/preferences
 	var/species
-	var/blood_type                           //blood type
-
+	var/blood_type
 	var/eye_colour = COLOR_BLACK
 	var/skin_colour = COLOR_BLACK
-	var/skin_tone = 0                    //Skin tone
+	var/skin_tone = -75
 	var/list/sprite_accessories = list()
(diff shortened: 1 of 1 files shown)


code/modules/client/preference_setup/general/02_body.dm
```diff
@@ -1,10 +1,9 @@
 /datum/preferences
 	var/species
-	var/blood_type                           //blood type
-
+	var/blood_type
 	var/eye_colour = COLOR_BLACK
 	var/skin_colour = COLOR_BLACK
-	var/skin_tone = 0                    //Skin tone
+	var/skin_tone = -75
```
out-of-phaze, line 6: could maybe use a comment explaining the range at some point

out-of-phaze (approved):

code/modules/client/preference_setup/general/02_body.dm
MistakeNot4892: I don't understand it sufficiently to comment it.

PR #4403, opened 2024-08-23
Adding a sieve for filtering liquids and solids.

MistakeNot4892:
- Adds a filter stand structure built from planks.
- Adds a filter made from cloth.
- Putting a filter and a vessel into a filtration stand then pouring reagents into the stand will filter into the loaded vessel.

Icons done by @PlayerDeer before I mauled them.

## TODO
- [x] Get pouring into the stand to work.
- [x] Work out why filters don't leak onto the turf properly. 
- [x] Work out why salt leaks out of the filter onto the turf, but not into the beaker.

PR #4404, opened 2024-08-24
Fix issues with solid/liquid reagent transfer

noelle-lavenza:
## Description of changes
Reagent apportionment (the `part` local variable in `trans_to_holder()`) now respects `transferred_phases`.
Remainder handling in `trans_to_holder()` now respects `transferred_phases`.
The `primary_solid` and `primary_liquid` local variables in `update_total()` are now public variables on `/datum/reagents`.

Fixes two issues discovered by testing #4403.
- Reagent apportionment (splitting the transfer amount between all the relevant reagents) gave incorrect transfer amounts when only transferring one phase of a multi-phase reagent mixture, because it was dividing by total_volume instead of the total solid or liquid volume.
- Reagent remainder handling (topping off transfers in cases where the split amount doesn't add up to the desired amount due to rounding) didn't properly handle single-phase transfers.

## Authorship
Myself.

### DIFF
--- a/code/modules/reagents/Chemistry-Holder.dm
+++ b/code/modules/reagents/Chemistry-Holder.dm
@@ -550,7 +550,12 @@ var/global/obj/temp_reagents_holder = new
 			return
 		part /= using_volume
 	else
-		part /= total_volume
+		var/using_volume = total_volume
+		if(!(transferred_phases & MAT_PHASE_LIQUID))
+			using_volume -= total_liquid_volume
+		else if(!(transferred_phases & MAT_PHASE_SOLID))
+			using_volume = total_liquid_volume
+		part /= using_volume
 
 	. = 0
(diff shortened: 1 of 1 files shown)


code/modules/reagents/Chemistry-Holder.dm
```diff
@@ -133,7 +135,7 @@ var/global/obj/temp_reagents_holder = new
 	primary_reagent = null
 
 	reagent_volumes = list()
-	var/primary_liquid = null
+	primary_liquid = null
```
MistakeNot4892, line 138: Strongly suggest nulling both of these in clear_reagent() as well.

MistakeNot4892:
Appears to work very well in the filter stand PR.

MistakeNot4892 (approved):

PR #4405, opened 2024-08-24
Adding mouth gripper to passive animals.

### LINKED ISSUE #73: simple animals cannot hold mice in their mouth slots
"fox trying to pick up a mouse in their mouth makes it drop on the ground"

MistakeNot4892:
- Adds mouth gripper to passive animal type. Mice can't pick stuff up so no big balance implications there.
- Removes simple_animal put_in_hands() override that is covered by lower level behavior.
- Merged human add_grab() down to /mob/living behind a new bool set the first time a hand slot is added.
- Fixes https://github.com/PyrelightSS13/Pyrelight/issues/73
- Fixes unreported issues with simple animal grabs in general.

### DIFF
--- a/code/modules/mob/living/living_defines.dm
+++ b/code/modules/mob/living/living_defines.dm
@@ -92,3 +92,6 @@
 
 	/// Var used to track current step for footsteps sounds.
 	var/tmp/step_count
+
+	/// Has this mob -ever- had a gripper? Used to skip hand checks in some cases.
+	var/has_had_gripper = FALSE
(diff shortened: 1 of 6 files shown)


code/modules/mob/living/living_defines.dm
```diff
@@ -91,4 +91,7 @@
 	var/weather_sensitive = FALSE
 
 	/// Var used to track current step for footsteps sounds.
-	var/tmp/step_count
\ No newline at end of file
+	var/tmp/step_count
+
+	/// Has this mob -ever- had a gripper? Used to skip hand checks in some cases.
+	var/has_had_gripper = FALSE
```
out-of-phaze, line 97: not totally convinced this is necessary but like, it's also not an issue

out-of-phaze: oh i see why it's necessary. yeah okay looks good

out-of-phaze (approved):

PR #4413, opened 2024-08-28
Culture tags are now a decl system.

MistakeNot4892:
## Description of changes
- `/decl/cultural_info` is now `/decl/background_detail` and `culture` is now `heritage` so that `/decl/cultural_info/culture/culture` is less confusing/easier to read.
- Instead of TAG_FOO, background categories are defined via `/decl/background_category` (including grandfather save info)
- Background categories and info are saved via UID.
- Backgrounds can be overridden per-map.
- Backgrounds are retrieved via flags.

Testing todo:
- [x] Pref availability in chargen.
- [x] Pref save, load, sanitize.
- [x] Pref name randomization.
- [x] Pref copy to mob on join.
- [x] Corpses
- [x] Random event location

Modular/extensible background code.

## Authorship
Myself.

## Changelog
Nothing player-facing.


code/game/movietitles.dm
```diff
@@ -118,8 +118,8 @@ var/global/list/end_titles
 		if(!C.holder)
 			continue
 		if(C.holder.rights & (R_DEBUG|R_ADMIN))
-			var/list/all_cultures = decls_repository.get_decls_of_subtype(/decl/cultural_info/culture)
-			var/decl/cultural_info/cult = all_cultures[pick(all_cultures)]
+			var/list/all_backgrounds = decls_repository.get_decls_of_subtype(/decl/background_detail/heritage)
+			var/decl/background_detail/cult = all_backgrounds[pick(all_backgrounds)]
 			staff += "[uppertext(pick(staffjobs))] - [cult.get_random_name(pick(MALE, FEMALE))] a.k.a. '[C.key]'"
```
out-of-phaze, line 123: as an aside, i think client.gender is a thing that could be worth plugging in here someday?

code/unit_tests/backgrounds.dm
```diff
@@ -0,0 +1,99 @@
+/datum/unit_test/background
+	name = "BACKGROUND - All Species Background Values Shall Be Of Valid Types And Length"
+
+/datum/unit_test/background/start_test()
+
+	var/list/all_background_tokens = global.using_map.get_background_categories()
+
+	var/fails = 0
+	for(var/species_name in get_all_species())
+		var/decl/species/species = get_species_by_key(species_name)
```
out-of-phaze, line 10: kind of wondering if this should be on `/decl/species/validate()` instead?

edit: oh this already existed, fair enough

out-of-phaze (approved):

code/game/movietitles.dm
MistakeNot4892: Cilents have gender??

code/unit_tests/backgrounds.dm
MistakeNot4892: Yeah this is an older one.

code/game/movietitles.dm
out-of-phaze: you can set it on your byond profile and it uses that, yeah. new_player probably has it too? `This is the client's gender, which is an attribute of the player's key. By default, when a new mob is made for a player (in client.New()), the new mob gets the same name and gender as the player's key.`

MistakeNot4892: That's amazing, I never knew that.
I wonder if we should give clients a `get_pronouns()` using the same logic as mobs, and let them choose an override in global prefs.

PR #4427, opened 2024-09-03
Improve map editor preview colours/sprites

liz-lavenza:
## Description of changes
- Adds proper preview colours to material stacks, thread, cloth, bowstrings, rocks, banners, waterskins, bags, clothes, lanterns, textile benches, fire sources, etc.
- Adds a hanging banner frame subtype with wall offsets.
- Adds custom preview icons for simple beds and variants thereof, as well as preview colours.
- Adds a subtype of cloak that takes material colour, available in loadout.
- Adds a handful of material subtypes for racks, doors, etc. Also includes a non-inset concrete tile.

Bumps the BYOND version used in CI to fix an issue with `::` not compiling properly.

This makes it easier to see what a map will look like while in the map editor. It also adds some additional props for decorating medieval maps.


MistakeNot4892 (approved):
I could have sworn we already had a craftable cloak subtype, but I might be thinking of mantles.

PR #4428, opened 2024-09-03
Fix icon colours in view-variables menu

liz-lavenza:
## Description of changes
The icon sent to the VV menu is uncoloured by default. This blends the colour in manually. The blend call is probably minimal in terms of overhead compared to browse_rsc, and it's only for an admin feature anyway, so I figure the icon operation overhead is negligible.

The VV menu will now have a coloured preview icon for the datum you're viewing, if it's an atom. Previously, it was monochrome only.
It could be made more accurate if you used getFlatIcon, but that would be too much for just a small preview icon; this handles things like grass, meat, and most clothes just fine.


MistakeNot4892 (approved):
Would be nice to cache this in future to avoid network overhead but heigh ho.

PR #4431, opened 2024-09-04
Adds drinking and hand washing, adds a wall fountain prop.

MistakeNot4892:
## Description of changes
- Alt-click on flooded turfs or structures that act as open containers (like wells) will allow you to drink or wash your hands (if you have any).
- Adds a wall fountain prop for dungeons or such.

Some long desired interactions.

## Authorship
My code, @PlayerDeer's fountain icon.

## Changelog
:cl:
tweak: You can now drink from and wash your hands in wells and flooded turfs.
/:cl:


code/game/objects/structures/wall_fountain.dm
```diff
@@ -0,0 +1,44 @@
+// Due to positioning and having a single-direction icon, this one shouldn't be constructable.
```
out-of-phaze: huh. really? could probably restrict it to certain directions or even just have it always face south and be offset onto the wall to the north. could even use the wall-supported checks for it.

edit: to be clear, non blocking, just a comment/question

code/game/objects/items/_item_interactions.dm
```diff
@@ -40,3 +40,76 @@
...
+	return ..() && target?.reagents?.total_volume && user.check_has_mouth() && !istype(target, /obj/item)
+
+/decl/interaction_handler/drink/invoked(atom/target, mob/user, obj/item/prop)
+
+	// Items can be picked up and drunk from, this interaction is for turfs and structures.
+	if(istype(target, /obj/item))
+		return
+
+	if(!user.check_has_mouth())
+		return
```
out-of-phaze:
```suggestion
	if(!user.check_has_mouth())
		show_food_no_mouth_message(user, target)
		return
```
maybe?

code/game/objects/items/_item_interactions.dm
```diff
@@ -40,3 +40,76 @@
...
+
+	if(!target?.reagents?.total_volume)
+		to_chat(user, SPAN_WARNING("\The [src] is empty of reagents."))
```
out-of-phaze: might need to be `show_food_empty_message(user, target)` but that doesn't specify reagents i suppose

code/modules/mob/living/living_eating.dm
```diff
@@ -2,8 +2,15 @@
 	user = user || src
 	if(get_food_satiation(consumption_method) < get_max_nutrition())
```
out-of-phaze, line 3: uh as an aside, isn't this wrong? it'd compare hydration against max nutrition instead of max hydration. maybe calls for adding `get_food_max_satiety(consumption_method)`?

code/game/objects/items/_item_interactions.dm
```diff
@@ -40,3 +40,76 @@
...
+			break
+	if(!found_hand)
+		return
+
+	var/decl/pronouns/pronouns = user.get_pronouns()
+	user.visible_message(
+		SPAN_NOTICE("\The [user] washes [pronouns.his] hands in \the [target]."),
+		SPAN_NOTICE("You wash your hands in \the [target].")
+	)
+	user.clean()
```
out-of-phaze, line 88: wondering if we should have a more specific handwashing proc given that user.clean() is going to wash their entire body. maybe it should just find the covering objects on your hands, wash those, and if there is no covering object, wash your hand organs.
but then again the germ level reduction from washing your hands is in user.clean(), but also germ level stuff needs to get looked at and fixed up on its own...
not blocking, just musing

code/game/objects/items/_item_interactions.dm
```diff
@@ -40,3 +40,76 @@
...
+	for(var/hand_slot in global.all_washable_hand_slots)
+		if(user.get_organ(hand_slot))
+			found_hand = TRUE
+			break
+	if(!found_hand)
+		return
+
+	var/decl/pronouns/pronouns = user.get_pronouns()
+	user.visible_message(
+		SPAN_NOTICE("\The [user] washes [pronouns.his] hands in \the [target]."),
```
out-of-phaze: this will, for river turfs, say `Penelope Haze washes her hands in the mud.`, which is admittedly kind of funny but gives a misleading impression. wonder if we could generalize `/turf/proc/get_trench_name()` somehow or if it'd be better to just hardcode a turf check here

out-of-phaze (changes requested):
not a huge fan of the hardcoded handwashing organ list but i suppose the alternative would either let you wash your head in a sink as a drake (lol) or require something like... put `can_handwash` on gripper slot datums, loop over all of them, and check 1) has associated organ 2) can_handwash. dunno, maybe it's worth it, i just hate those hardcoded hand tag lists. maybe my next project will be to obliterate them

out-of-phaze (dismissed):

code/game/objects/items/_item_interactions.dm
MistakeNot4892: Done.

code/game/objects/items/_item_interactions.dm
MistakeNot4892: Done.

code/modules/mob/living/living_eating.dm
MistakeNot4892: I thiiiiink this is intended to be like... appetite/stomach capacity. Really not sure, that whole mechanic needs looking at, you lose your appetite entirely if you eat like 3 bites of food on dev.

code/game/objects/items/_item_interactions.dm
MistakeNot4892: Yeah it's not very granular currently.

code/game/objects/items/_item_interactions.dm
MistakeNot4892: Swapped it to `get_fluid_name()` for turfs which will be `Penelope Haze washes her hands in the water.`

out-of-phaze: oh yeah `get_trench_name()` is just for trenches filled with water, and in handwashing it will always be filled with water

code/modules/mob/living/living_eating.dm
out-of-phaze: get_food_satiation returns current hydration/nutrition level, get_max_nutrition is only nutrition, is the point

MistakeNot4892:
> not a huge fan of the hardcoded handwashing organ list but i suppose the alternative would either let you wash your head in a sink as a drake (lol) or require something like... put `can_handwash` on gripper slot datums, loop over all of them, and check 1) has associated organ 2) can_handwash. dunno, maybe it's worth it, i just hate those hardcoded hand tag lists. maybe my next project will be to obliterate them

Made it use a bool on the organ.

out-of-phaze (approved):

PR #4432, opened 2024-09-04
Tweaks radial label positioning.

MistakeNot4892:
## Description of changes
- Moves the default label position up eight pixels to associate the labels closer with their button.
- Adds a `use_labels` value that centers the value on the button.

image
image

Improves UI readability for radials. Stops me repeatedly clicking on cancel because the label is closer to cancel than its own button.

## Authorship
Myself.

## Changelog
Nothing really player relevant.


out-of-phaze (approved):
ideally we'd replace the text-only ones with actual icons, buuuuuut this is a good stopgap to improve UX

PR #4433, opened 2024-09-04
Adding lockpicks.

MistakeNot4892:
## Description of changes
Adds a set of lockpicks with high lockpicking effectiveness.

Currently the best lockpicking tool in the codebase is a stick or a metal rod.

## Authorship
Code by myself, base icon by Zirc.

## Changelog
Nothing player-facing.


out-of-phaze (commented):
huh i kinda figured these would be a storage object. doesn't this mean the whole thing will just disintegrate when you fail to pick a lock?

code/game/objects/items/lockpicks.dm
```diff
@@ -0,0 +1,79 @@
...
+	name                = "roll of lockpicks"
+	desc                = "A stitched roll used to store thin, strangely-shaped tools commonly used used to pick locks."
+	icon                = 'icons/obj/items/lockpick_roll.dmi'
+	icon_state          = ICON_STATE_WORLD
+	material            = /decl/material/solid/organic/leather
+	color               = /decl/material/solid/organic/leather::color
+	material_alteration = MAT_FLAG_ALTERATION_ALL
+	storage             = /datum/storage/lockpick_roll
+
+/obj/item/lockpick_roll/filled/Initialize(ml, material_key)
```
out-of-phaze: this should be WillContain, shouldn't it?

out-of-phaze (changes requested):

code/game/objects/items/lockpicks.dm
MistakeNot4892: Oh true.

code/game/objects/items/lockpicks.dm
```diff
@@ -0,0 +1,78 @@
...
+/obj/item/lockpick/hook
+	use_icon_state = "2"
+/obj/item/lockpick/lever
+	use_icon_state = "3"
+
+/obj/item/lockpick/Initialize(ml, material_key)
+	. = ..()
+	if(isnull(use_icon_state))
+		use_icon_state = num2text(rand(1,3))
+		icon_state = "[get_world_inventory_state()][use_icon_state]"
```
out-of-phaze: is there a reason this isn't just an `update_icon()` call?

code/game/objects/items/lockpicks.dm
```diff
@@ -0,0 +1,78 @@
...
+	. = ..()
+	icon_state = ICON_STATE_WORLD
+	if(!storage?.opened)
+		icon_state = "[icon_state]-rolled"
+	else if(length(contents))
+		var/x_offset = -9
+		for(var/i = 1 to length(contents))
+			var/obj/item/lockpick/thing = contents[i]
+			if(istype(thing))
+				var/image/pick         = new /image
```
out-of-phaze: might be better to just do an appearance copy?
```dm
				var/image/pick         = new /image
				pick.appearance = thing
				pick.appearance_flags |= RESET_COLOR
				pick.pixel_x = x_offset
```
also pick is a global proc (builtin, technically), not super certain about using it as a keyword :(

out-of-phaze (changes requested):

code/game/objects/items/lockpicks.dm
MistakeNot4892: I did an appearance copy in the previous revision of this and it broke colouration and layering for some reason.

code/game/objects/items/lockpicks.dm
MistakeNot4892: There is a reason but I don't remember what it is.

out-of-phaze (approved):

PR #4434, opened 2024-09-04
Adding horses and saddles.

MistakeNot4892:
## Description of changes
- Cleans up some mob riding layering code so it works more consistently.
- Adds horses and saddles that allow horses to be ridden.

image

Improves the base mob riding code and adds horses for the fantasy map.

## Authorship

## Changelog
Nothing immediately player-facing.


code/modules/mob/mob_layering.dm
```diff
@@ -7,22 +7,24 @@
 	var/last_layer = layer
 	var/new_layer = get_base_layer()
 	if(isturf(loc))
-
 		var/turf/T = loc
 		if(T.pixel_z < 0)
 			new_layer = T.layer + 0.25
 		else if(buckled && buckled.buckle_layer_above)
-			new_layer = buckled.layer + ((buckled.dir == SOUTH) ? -0.1 : 0.1)
+			new_layer = buckled.layer + ((buckled.dir == SOUTH) ? -0.01 : 0.01)
```
out-of-phaze, line 14: fyi there is a point if layers get too large (this is an issue with wallening on tg, iirc, they multiplied everything by 1000) or if the subdivisions get too small that floating point errors will fuck this up. it's probably not an issue here but like, figured i'd comment on it just in case

out-of-phaze (approved):

code/modules/mob/mob_layering.dm
MistakeNot4892: I hate how we handle layering currently, particularly when trenches are involved. Really unsure what to do about a lot of it.

PR #4435, opened 2024-09-04
Adding gargoyle statues.

MistakeNot4892:
Simple PR to add @PlayerDeer's neato gargoyle icons as map decorations.


out-of-phaze (dismissed):
would be neat to add these as a high tier stonemasonry recipe...

also would be nice to have cursed gargoyles on pyrelight that leap up and attack you >:)

out-of-phaze (approved):

PR #4436, opened 2024-09-04
Adding a town bell.

MistakeNot4892:
## Description of changes
This might need to be pointed at Pyrelight instead of main Neb, it's highly specific.

Adds a bell structure that can be used to send everyone in the z-chunk a message. 

The use case for this is on Shaded Hills which has no way to call people to the town square or otherwise communicate en masse.

## Authorship
Near sound: https://freesound.org/people/HMTSCCSound/sounds/554655/
Far sound: https://freesound.org/people/rattatatiana/sounds/511681/

## Changelog
:cl:
tweak: Adds a town bell to Shaded Hills.
/:cl:


code/_helpers/turfs.dm
```diff
@@ -211,3 +211,38 @@
...
+	target = get_turf(target)
+	if(!istype(source) || !istype(target))
+		return "somewhere"
+	if(source == target)
+		return "right next to you"
+
+	if(source.z > target.z)
+		. += "upwards and to"
+	else if(source.z < target.z)
+		. += "downwards and to"
```
out-of-phaze: I kind of prefer "above" and "below" rather than "upwards" and "downwards"? Feels less clunky, but not a blocker

code/game/objects/structures/town_bell.dm
```diff
@@ -0,0 +1,103 @@
...
+/obj/structure/town_bell/Initialize(mapload)
+	if(bell_material)
+		LAZYSET(matter, bell_material, MATTER_AMOUNT_REINFORCEMENT)
+	. = ..()
+	update_icon()
+
+/obj/structure/town_bell/update_material_desc(override_desc)
+	. = ..()
+	if(material && bell_material)
+		desc += " The frame is made of [material.name] and the bell has been cast from [bell_material::name]."
```
out-of-phaze: ughhhh sorry to nitpick but this should really use `var/decl/material/bell_material = GET_DECL(bell_material); bell_material.use_name`
1) it should use use_name or solid_name instead of name, probably, and
2) use_name and solid_name are autoset on a lot of materials so `::` would only give the compile-time value, meaning we'd need to get the actual decl to find the runtime value
sorryyyy

code/_helpers/turfs.dm
```diff
@@ -211,3 +211,38 @@
 		return list(new_turf, source)
 
 	return new_turf
+
+/proc/get_dir_z_text(turf/source, turf/target)
```
out-of-phaze: I'm not super sure if I like the way this is currently structured? I feel like it should use `get_dir()` followed by `dir2text()` or maybe a new `dir2text_multiz()` method that just extends it to support above and below. Most of this is either already in existing helpers, or should have helpers created for it.

also note to self, we need a `get_dir_multiz()` variant that handles adjacent levels...

edit: again not really a blocker though since it's separate to the main concern of this PR

out-of-phaze (changes requested):

code/_helpers/turfs.dm
MistakeNot4892: I am using this over get_dir()/dir2text() specifically because they don't handle lateral multiz.

out-of-phaze: this won't either, will it? if you're 15 tiles to the west of the bell on one z level, source.x < target.x, and so it'll say the bell is coming from the east, which is correct. but if you're 15 tiles to the west of the bell with a z-level transition in between, source.x > target.x, because you end up on the right side of the new level, so it'll say it's from the west which is wrong

out-of-phaze (changes requested):

out-of-phaze (dismissed):
i misunderstood 💀

MistakeNot4892:
Had a bit of a fey mood and hacked out a solution to the z-level distance/direction stuff.

code/modules/multiz/level_data.dm
```diff
@@ -224,6 +233,111 @@
...
+		// Update our lateral coords based on the direction of our connection.
+		// This does NOT appropriately handle looping or non-horizontal/vertical
+		// connections, which violate our assumptions.
+		if(previous)
+
+			checking.z_volume_level_x = previous.z_volume_level_x
+			checking.z_volume_level_y = previous.z_volume_level_y
+			checking.z_volume_level_z = previous.z_volume_level_z
+
+			var/connect_dir = checking.connected_levels[previous.level_id]
```
out-of-phaze: previous can be non-null even when `checking.connected_levels` is null which causes a bad index runtime here. should probably do uh
```suggestion
			var/connect_dir = checking.connected_levels?[previous.level_id]
```
or use lazyaccess, but in this case they're equivalent

out-of-phaze (changes requested):
CI fail

code/modules/multiz/level_data.dm
MistakeNot4892: I genuinely just assumed that was always a list since we have like, 20 level objects max.

out-of-phaze (approved):

PR #4442, opened 2024-09-09
Fix issues with ointment, ailments, and medical stacks

out-of-phaze:
## Description of changes
Makes treating ailments take priority over other interactions, to avoid conflicts with organ treatment.
Replaces `LAZYREMOVE` with `qdel` in `remove_ailment()`.
Removes ailments on organ rejuvenation.
Removes a redundant `use(1)` in ointment code.

Fixes being unable to treat rashes and sprains using ointment and gauze.
Fixes ointment being twice as expensive, using (up to) two stacks per interaction.
Fixes removing ailments via VV not ending ailment effects.
Fixes rejuvenation not removing ailments.

### DIFF
--- a/code/_onclick/item_attack.dm
+++ b/code/_onclick/item_attack.dm
@@ -58,6 +58,14 @@ avoid code duplication. This includes items that may sometimes act as a standard
 	if(!ismob(user))
 		return TRUE
 
+	if(!QDELETED(used_item) && user.a_intent == I_HELP)
+		var/obj/item/organ/external/E = GET_EXTERNAL_ORGAN(src, user.get_target_zone())
+		if(length(E?.ailments))
+			for(var/datum/ailment/ailment in E.ailments)
+				if(ailment.treated_by_item(used_item))
+					ailment.was_treated_by_item(used_item, user, src)
+					return TRUE
+
 	if(user.a_intent != I_HURT)
 		if(can_operate(src, user) != OPERATE_DENY && used_item.do_surgery(src,user)) //Surgery
 			return TRUE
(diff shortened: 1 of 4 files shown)


code/_onclick/item_attack.dm
```diff
@@ -58,6 +58,14 @@ avoid code duplication. This includes items that may sometimes act as a standard
 	if(!ismob(user))
 		return TRUE
 
+	if(!QDELETED(used_item) && user.a_intent == I_HELP)
+		var/obj/item/organ/external/E = GET_EXTERNAL_ORGAN(src, user.get_target_zone())
+		if(length(E?.ailments))
+			for(var/datum/ailment/ailment in E.ailments)
+				if(ailment.treated_by_item(used_item))
+					ailment.was_treated_by_item(used_item, user, src)
+					return
```
MistakeNot4892: Should this return TRUE?

code/_onclick/item_attack.dm
```diff
@@ -76,24 +84,10 @@ avoid code duplication. This includes items that may sometimes act as a standard
 
 	var/oldhealth = current_health
 	. = used_item.use_on_mob(src, user)
-	if(used_item.get_attack_force(user) && istype(ai) && current_health < oldhealth)
+	if(used_item.get_attack_force(user) && user != src && istype(ai) && current_health < oldhealth)
```
MistakeNot4892: I would suggest doing this in ai.retaliate() itself

code/modules/organs/organ.dm
```diff
@@ -522,12 +523,12 @@ var/global/list/ailment_reference_cache = list()
 	if(ispath(ailment, /datum/ailment))
 		for(var/datum/ailment/ext_ailment in ailments)
 			if(ailment == ext_ailment.type)
-				LAZYREMOVE(ailments, ext_ailment)
+				qdel(ext_ailment)
```
MistakeNot4892, line 526: Shouldn't this be in addition to the LAZYREMOVE?

code/_onclick/item_attack.dm
out-of-phaze: i did not mean to commit that oops

code/_onclick/item_attack.dm
out-of-phaze: yep

code/modules/organs/organ.dm
out-of-phaze: no, `/datum/ailment/Destroy()` removes it from the list. this is exactly how the treatment procs remove them too

MistakeNot4892 (approved):

PR #4444, opened 2024-09-09
Fix ambiguous 'in' order warnings from OpenDream

out-of-phaze:
## Description of changes
Adds parentheses to disambiguate 'in' clauses with ambiguous precedence, caught by OpenDream.
Also elevates the warning to an error.

Fixes things broken due to 'in' precedence being weird

### DIFF
--- a/code/modules/xenoarcheaology/finds/find_types/statuette.dm
+++ b/code/modules/xenoarcheaology/finds/find_types/statuette.dm
@@ -40,7 +40,7 @@
 	//see if we've identified anyone nearby
 	if(world.time - last_bloodcall > bloodcall_interval && nearby_mobs.len)
 		var/mob/living/human/M = pop(nearby_mobs)
-		if(M in view(7,src) && M.current_health > 20)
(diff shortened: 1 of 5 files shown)


code/modules/xenoarcheaology/finds/find_types/statuette.dm
```diff
@@ -40,7 +40,7 @@
 	//see if we've identified anyone nearby
 	if(world.time - last_bloodcall > bloodcall_interval && nearby_mobs.len)
 		var/mob/living/human/M = pop(nearby_mobs)
-		if(M in view(7,src) && M.current_health > 20)
+		if((M in viewers(7,src)) && M.current_health > 20)
```
MistakeNot4892, line 43: Why swap to viewers()?

out-of-phaze: M is always a mob so it's identical but faster

MistakeNot4892 (approved):

PR #4447, opened 2024-09-10
Remove or implement unused variables and procs

out-of-phaze:
## Description of changes
Removes a bunch of unused variables and procs, implements/uses some other ones.

In terms of new features:
- Re-adds the name/job labels to PDAs. At some point I might make this use the labels extension, right now I can't be assed.
- Some optimizations I forgot to nab before

Also includes a few optimizations I forgot to pull in :(

code quality, restoring a feature that got clobbered in the move from device->modcomp PDAs, some missing optimizations oops

## Authorship
Me

## Changelog
:cl:
add: Readds name and job labels to PDAs obtained via job equipment
/:cl:

### DIFF
--- a/code/datums/mind/mind.dm
+++ b/code/datums/mind/mind.dm
@@ -283,7 +283,7 @@
 			if(!def_value)//If it's a custom objective, it will be an empty string.
 				def_value = "custom"
 
-		var/new_obj_type = input("Select objective type:", "Objective type", def_value) as null|anything in list("assassinate", "debrain", "protect", "prevent", "harm", "brig", "hijack", "escape", "survive", "steal", "download", "mercenary", "capture", "custom")
+		var/new_obj_type = input("Select objective type:", "Objective type", def_value) as null|anything in list("assassinate", "debrain", "protect", "prevent", "harm", "brig", "hijack", "escape", "survive", "steal", "download", "mercenary", "custom")
 		if (!new_obj_type) return
 
 		var/datum/objective/new_objective = null
@@ -341,7 +341,7 @@
 				new_objective = new /datum/objective/steal
 				new_objective.owner = src
 
-			if("download","capture")
+			if("download")
 				var/def_num
 				if(objective&&objective.type==text2path("/datum/objective/[new_obj_type]"))
 					def_num = objective.target_amount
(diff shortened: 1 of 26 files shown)


out-of-phaze:
I can pull the optimizations out into another PR I was just being lazy. Would also appreciate a closer look at the state machine stuff

code/datums/mind/mind.dm
```diff
@@ -354,9 +354,6 @@
 					if("download")
 						new_objective = new /datum/objective/download
 						new_objective.explanation_text = "Download [target_number] research levels."
-					if("capture")
```
MistakeNot4892, line 357: Why are you removing this one?

code/game/objects/item_mob_overlay.dm
```diff
@@ -16,7 +16,7 @@ var/global/list/bodypart_to_slot_lookup_table = list(
 	else
 		has_inventory_icon = FALSE
 	if(. || update_icon)
-		update_icon()
+		queue_icon_update()
```
MistakeNot4892: This is going to flood the queue during atoms init.

out-of-phaze: from what I recall I profiled it and it was a net positive but I can easily drop the commit

code/datums/mind/mind.dm
out-of-phaze: what even are "capture points"

MistakeNot4892: They're a function of species rarity iirc. If you're removing this you might as well remove the data theft objective too as neither of them are actually tracked for greentext.

code/game/objects/item_mob_overlay.dm
MistakeNot4892: If it doesn't have an impact in profiling then I don't mind but as far as I can tell this is going to make every item queue.

code/datums/mind/mind.dm
out-of-phaze: that's still something you can subjectively measure, though? research levels exist, data disks exist. capture points are not mentioned anywhere else in the codebase

MistakeNot4892 (approved):

PR #4451, opened 2024-09-13
Fix everyone starting off with $0

### LINKED ISSUE #4445: Cash doesn't spawn in your bag
#### Description of issue
Selecting to have money spawn in your bag as cash leads to the money not spawning at all.

You should get money if you select money

#### Steps to reproduce
Select 'all in cash' or another option that gives cash in the bag, see there is no cash.

A week or a bit longer?

Occurred on multiple server revisions and multiple rounds.

#### Issue bingo

out-of-phaze:
## Description of changes
`human.background_info` is an associative list of background category tokens to background detail instances. `prefs.background_info` is an associative list of background category tokens to background detail *types.* `GET_DECL()` should only be used in the latter case. I used `human.get_background_datum(token)` just for thoroughness.

Also reworks the finance skill bonus so that it doesn't apply at unskilled. This means everyone gets a slight reduction in starting pay across the board. Sorry!!

oh and I changed the variable names in the related procs because `M` for `/datum/money_account` instead of `/mob` was tripping me up.

Characters start with money again.

## Authorship
me

### DIFF
--- a/code/game/jobs/job/_job.dm
+++ b/code/game/jobs/job/_job.dm
@@ -123,20 +123,20 @@
 	. ||= outfit_type
 	return GET_DECL(.)
 
-/datum/job/proc/create_cash_on_hand(var/mob/living/human/H, var/datum/money_account/M)
-	if(!istype(M) || !H.client?.prefs?.starting_cash_choice)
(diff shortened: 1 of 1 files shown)


MistakeNot4892 (approved):
Whoops. Thanks for fixing that.

comma (approved):

PR #4466, opened 2024-09-16
Converting many grab attackby() blocks into grab_attack().

MistakeNot4892:
## Description of changes
Breaks various blocks out of attackby() into grab_attack().

Allows grab attack overrides to function properly without wonky proc flow like structure base grab interactions talking precedence over attackby().

## Authorship
Myself.

## Changelog
Nothing player-facing.


MistakeNot4892:
Depends on #4462 railing changes.

MistakeNot4892:
All appears to be working.

out-of-phaze (dismissed):
a little worried about the reversed arg order (user then item instead of item then user) causing issues down the line/when migrating legacy code, but it'll probably be fine. ~~it's something typemaker will hopefully be able to detect eventually, at least~~

MistakeNot4892:
> a little worried about the reversed arg order (user then item instead of item then user) causing issues down the line/when migrating legacy code, but it'll probably be fine. ~it's something typemaker will hopefully be able to detect eventually, at least~

Easy enough to tweak it.

out-of-phaze (dismissed):

out-of-phaze (approved):

Issue #4465, opened 2024-09-16
when standing in deep water, pepper spray cans on your person "leak" 


sharah-za-fluffy:
#### Description of issue
standing in deep water while having pepper spray cans on your person has said cans "leak" some pepper spray into the water, hitting you and others with pepper spray laced water

#### Difference between expected and actual behavior
pepper spray should be waterproof and not turn the flooded medbay into hellish pepper water for those without gasmasks

#### Steps to reproduce
be shitcurity on a round where flooding event happened long enough to totally flood a department
have pepper spray cans on your person
walk though deep water
watch as you get hit with pain or messages about face mask protecting you

#### Specific information for locating

pepper spray

#### Length of time in which bug has been known to occur

found it today by accidentally gassing mawreek in this exact manner

#### Client version, Server revision & Game ID

Server Revision: 5c689e810a0ec8bd3dd7c6030d0c823f0ea5e852 - main - 2024-09-16
Game ID: cvJ-cRb3
Current map: Tradepost Mollusc

#### Issue bingo

- [x] Issue could be reproduced at least once
- [x] Issue could be reproduced by different players
- [x] Issue could be reproduced in multiple rounds
- [x] Issue happened in a recent (less than 7 days ago) round
- [x] [Couldn't find an existing issue about this](https://github.com/NebulaSS13/Nebula/issues)

Issue #4458, opened 2024-09-15
batch crafting of cones and other unstackables causes waste of material 


sharah-za-fluffy:
#### Description of issue
I attempted to make 24 warning cones from 24 plastic sheets and only got one 

#### Difference between expected and actual behavior
either I should not be allowed to attempt to batch craft cones and other unstackable objects or I should just get like 23 cones under me after making the one

#### Steps to reproduce
get a lot of materials, click a crafting amount more than 1 (for example 5x) and watch what happens

#### Specific information for locating

construction skill menu

#### Length of time in which bug has been known to occur

since a while but I am reporting it now to stop waste of more important materials in the future 

#### Client version, Server revision & Game ID

Server Revision: 5c689e810a0ec8bd3dd7c6030d0c823f0ea5e852 - main - 2024-09-14

#### Issue bingo

- [x] Issue could be reproduced at least once
- [ ] Issue could be reproduced by different players
- [ ] Issue could be reproduced in multiple rounds
- [x] Issue happened in a recent (less than 7 days ago) round
- [x] [Couldn't find an existing issue about this](https://github.com/NebulaSS13/Nebula/issues)

MistakeNot4892:
![image](https://github.com/user-attachments/assets/ffc746f3-eb74-4627-bc58-281fb2687d87)
Could you do me a favour and check if this is reproducable on Scav this weekend? I can't repro it locally on stable, you can't multicraft cones.

MistakeNot4892:
Never mind, reproduced on stable, I had dev checked out.

PR #4469, opened 2024-09-17
Scav fixes.

MistakeNot4892:
- Fixes https://github.com/ScavStation/ScavStation/issues/1079
- Fixes https://github.com/ScavStation/ScavStation/issues/1069
- Fixes https://github.com/ScavStation/ScavStation/issues/1068
- Fixes https://github.com/ScavStation/ScavStation/issues/1053
- Fixes https://github.com/NebulaSS13/Nebula/issues/4465
- Fixes https://github.com/NebulaSS13/Nebula/issues/4458


code/game/atoms_fluids.dm
```diff
@@ -4,7 +4,7 @@
 /atom/proc/fluid_act(var/datum/reagents/fluids)
 	SHOULD_CALL_PARENT(TRUE)
 	// TODO: fix food open container behavior jesus christ
-	if(reagents && reagents != fluids && fluids?.total_volume >= FLUID_SHALLOW && ATOM_IS_OPEN_CONTAINER(src) && !istype(src, /obj/item/chems/food))
+	if(reagents && reagents != fluids && fluids?.total_volume >= FLUID_SHALLOW && !is_watertight() && !istype(src, /obj/item/chems/food))
```
out-of-phaze, line 7: when this is pulled into dev, do we want to set `watertight = TRUE` on lanterns? or would we rather just add a toggleable cap?

out-of-phaze (approved):

code/game/atoms_fluids.dm
MistakeNot4892: I was planning to set watertight = TRUE for simplicity.

Issue #3509, opened 2023-12-10
people in photos do not rotate to face camera but their clothes and hair do


MistakeNot4892:
Copying from https://github.com/ScavStation/ScavStation/issues/837

#### Description of issue
in past versions when you take a photo all players in photo face south, now only their clothes and hair do
![Screenshot 2023-12-09 125522](https://github.com/ScavStation/ScavStation/assets/65055421/af0f7f85-f220-434d-bfa2-b27085e58a23)

#### Difference between expected and actual behavior
either players face south in a photo or don't rotate at all, this bug is a unhappy inbetween.

#### Steps to reproduce
as with all camera oddities just get a camera, then take photo of someone not facing south

#### Specific information for locating

camera

#### Length of time in which bug has been known to occur

12/9/2023

#### Client version, Server revision & Game ID

Date
2023-12-09
Branch
main
Game ID
crb-cUhK
Runtime Version
514.1589
Compiler Version
514.1589
Revision
b8eae690be904a37869af8645d900b449dd55dbc

#### Issue bingo

- [ ] Issue could be reproduced at least once
- [ ] Issue could be reproduced by different players
- [ ] Issue could be reproduced in multiple rounds
- [x] Issue happened in a recent (less than 7 days ago) round
- [x] [Couldn't find an existing issue about this](https://github.com/ScavStation/ScavStation/issues)

PR #4479, opened 2024-09-19
Cleans up some SOUTH usage in getFlatIcon().

MistakeNot4892:
Fixes https://github.com/NebulaSS13/Nebula/issues/3509.

### DIFF
--- a/code/_helpers/icons.dm
+++ b/code/_helpers/icons.dm
@@ -635,7 +635,7 @@ The _flatIcons list is a cache for generated icon files.
 */
 
 // Creates a single icon from a given /atom or /image.  Only the first argument is required.
-/proc/getFlatIcon(image/A, defdir=2, deficon=null, defstate="", defblend=BLEND_DEFAULT, always_use_defdir = 0)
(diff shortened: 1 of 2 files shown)


code/_helpers/icons.dm
```diff
@@ -792,7 +791,7 @@ The _flatIcons list is a cache for generated icon files.
 	if(A.alpha < 255 && !islist(A.color))
 		flat.Blend(rgb(255, 255, 255, A.alpha), ICON_MULTIPLY)
 
-	return icon(flat, "", SOUTH)
+	return icon(flat, "", defdir)
```
out-of-phaze: uhhhhh, i don't actually think this is correct. i'm *pretty* sure it should be south because south is the default state? whereas if you specify like, EAST, it'll have four states and all but EAST will be blank

MistakeNot4892: Why are we providing it a dir at all then?

out-of-phaze: because we might want the single icon to be facing north/east/west instead of south

out-of-phaze (approved):

PR #4484, opened 2024-09-23
Non-reinforced wall demolition is now handled by an extension.

MistakeNot4892:
## Description of changes
- Wall dismantling is handled on an extension.
- Sledgehammers can break down non-reinforced walls.

`spaceasshole.ogg`

## Authorship
Myself.

## Changelog
:cl:
tweak: Sledgehammers can demolish non-reinforced walls.
/:cl:


code/datums/extensions/demolisher/_demolisher.dm
```diff
@@ -0,0 +1,38 @@
...
+	. = ..()
+	if(_delay)
+		demolish_delay = _delay
+	if(_verb)
+		demolish_verb = _verb
+	if(_sound)
+		demolish_sound = _sound
+
+/datum/extension/demolisher/proc/get_demolish_delay(turf/wall/wall)
+	return demolish_delay - wall?.get_material()?.cut_delay
```
comma: Is it supposed to be - delay here? Wouldn't that make demolishing diamond wall instantenous?

MistakeNot4892: Good question. Existing code was `var/cut_delay = (6 SECONDS) - material.cut_delay` but that does look inverted.

comma (approved):

out-of-phaze (approved):

PR #4489, opened 2024-09-25
Cherrypicks some ailment code and tidies up nanopaste code.

### LINKED ISSUE #4482: prostetic fault repair does not work seemingly
#### Description of issue
got a noisemaker fault in arm, opened it with screwdriver and crowbar and got the 3 nanopaste, however it just says "nothing to fix!" and that's it, I am stuck with a noisy arm and have to chop it off

I should be able to fix the fucking thing

#### Steps to reproduce
get funny RNG bullshit
do science to get nanopaste
attempt repair
get angry when it don't happen

nanopaste yinglet prosthetic arm

found it now

Server Revision: b59f29390d495fcc59222343642885e1d17992ed - main - 2024-09-21
Game ID: cvO-dpuu

#### Issue bingo

MistakeNot4892:
- Fixes #4482
- Allows nanopaste to repair limbs with less than thirty damage without using surgery.

### DIFF
--- a/code/_onclick/item_attack.dm
+++ b/code/_onclick/item_attack.dm
@@ -44,35 +44,31 @@ avoid code duplication. This includes items that may sometimes act as a standard
 	visible_message("<span class='danger'>[src] has been hit by [user] with [W].</span>")
 	return TRUE
 
-/mob/living/attackby(obj/item/I, mob/user)
+/mob/living/attackby(obj/item/used_item, mob/user)
 	if(!ismob(user))
 		return TRUE
-	if(can_operate(src,user) != OPERATE_DENY && I.do_surgery(src,user)) //Surgery
-		return TRUE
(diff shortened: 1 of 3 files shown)


code/_onclick/item_attack.dm
out-of-phaze, line 1: this is going to cause unreal conflicts when we merge stable->dev, isn't it

out-of-phaze (approved):

code/_onclick/item_attack.dm
MistakeNot4892: No?

MistakeNot4892: Double-checked, it will conflict in terms of changes, but the fix will just be taking stable and renaming the attack() proc, maybe some parent call stuff.

out-of-phaze: i thought i combined these two procs into one? because carbon was removed and the human level stuff got moved down

PR #4491, opened 2024-09-26
Expanding the existing whitelist config to support several levels of severity.

MistakeNot4892:
- `USEWHITELIST` is gone.
- `SERVER_WHITELIST` does the same job with four options:
    - `none`: no server whitelist checking.
    - `jobs`: unwhitelisted keys cannot join as jobs that are considered 'sensitive' (barred from guest joins).
    - `join`: unwhitelisted keys cannot join rounds except as observers.
    - `connect`: unwhitelisted keys can't connect to the server (similar to guestbans).
    
I haven't been able to test the IsBanned() change due to it apparently not being called on local joins, so will test that on Pyrelight this weekend.

### DIFF
--- a/code/__defines/misc.dm
+++ b/code/__defines/misc.dm
@@ -131,7 +131,7 @@
 #define PROJECTILE_CONTINUE   -1 //if the projectile should continue flying after calling bullet_act()
 #define PROJECTILE_FORCE_MISS -2 //if the projectile should treat the attack as a miss (suppresses attack and admin logs) - only applies to mobs.
 
-// Objective cofig enum values.
+// Objective config enum values.
 #define CONFIG_OBJECTIVE_NONE 2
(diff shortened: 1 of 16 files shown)


code/__defines/misc.dm
```diff
@@ -131,11 +131,20 @@
 #define PROJECTILE_CONTINUE   -1 //if the projectile should continue flying after calling bullet_act()
 #define PROJECTILE_FORCE_MISS -2 //if the projectile should treat the attack as a miss (suppresses attack and admin logs) - only applies to mobs.
 
-//objectives
+// Objective cofig enum values.
```
out-of-phaze: cofig
```suggestion
// Objective config enum values.
```

out-of-phaze (changes requested):

out-of-phaze (approved):

PR #4492, opened 2024-09-28
Small Pyrelight fixes, September 28 2024

out-of-phaze:
Rabbits have a butchery offset so that they're centered on the hook.
All props made of generic 'wood' on Shaded Hills are now made of walnut or ebony.
Fixed erroneous extra linebreak before mob examine messages. (sadly, blockquote doesn't work in a native BYOND output control and you can't fake it with CSS either)


MistakeNot4892 (commented):
Looks good once the include is fixed.

MistakeNot4892 (approved):

PR #4495, opened 2024-09-29
Simplifies vox markings.

MistakeNot4892:
Condenses vox markings down into a single set of decls.


MistakeNot4892:
Doing more investigation into vox/ute markings and species issue. Appears to be related to DNA/mob snapshot changes, specifically BODY_FLAG_NO_DNA.

MistakeNot4892:
Depends on #4497

mods/content/anima/mob.dm
```diff
@@ -0,0 +1,54 @@
+/decl/anima_aspect
```
MistakeNot4892: oops

out-of-phaze (approved):

PR #4497, opened 2024-09-30
Removes bodytype DNA flag checking from organ code.

MistakeNot4892:
- BODY_FLAG_NO_DNA is no longer checked when organs are being setup from a mob snapshot.
- Mob snapshots now only care about genetics when being requested by the organ printer, where genetics are the purpose.
- Fixes vox and utility frames not getting species set on bodyparts.
- Fixes vox and utility frame markings and hair.

### DIFF
--- a/code/game/gamemodes/endgame/ftl_jump/ftl_jump.dm
+++ b/code/game/gamemodes/endgame/ftl_jump/ftl_jump.dm
@@ -121,7 +121,7 @@
 /obj/effect/bluegoast/proc/blueswitch()
 	var/mob/living/human/H
 	if(ishuman(daddy))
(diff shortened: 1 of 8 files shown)


out-of-phaze:
> Mob snapshots now only care about genetics when being requested by the organ printer, where genetics are the purpose.

what's this mean? I use mob snapshots downstream for cloning, will this cause issues with that

MistakeNot4892:
````
[10:41 AM]Loaf: @Penny 
"what's this mean? I use mob snapshots downstream for cloning, will this cause issues with that"
the force param on the snapshot proc is now check_genetics and defaults to false
[10:41 AM]Loaf: should not impact your cloning code significantly, you might just need to rename the param in the call
[10:41 AM]Loaf: or invert it
````

out-of-phaze (approved):

PR #4509, opened 2024-10-06
Adds a unit test for food recipes

noelle-lavenza:
## Description of changes
Adds a unit test that attempts to produce all food recipes from their ingredients. Kind of a WIP, would appreciate thoughts on how to make this less bespoke for things like stacks, paper, etc., but also this caught some issues already so I'm pretty happy with it.

Should make debugging recipes easier, especially for soup which I understand has been a bit of a headache.

## Authorship
Me.

### DIFF
--- a/code/modules/food/cooking/_recipe.dm
+++ b/code/modules/food/cooking/_recipe.dm
@@ -212,11 +212,11 @@ var/global/list/_cooking_recipe_cache = list()
 	return new result(container)
 
 /// Return a data list to pass to a reagent creation proc. Allows for overriding/mutation based on ingredients.
-/// Actually place or create the result of the recipe. Returns the produced item, or the container for reagents.
 /decl/recipe/proc/get_result_data(atom/container, list/used_ingredients)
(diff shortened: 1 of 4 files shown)


code/modules/food/cooking/_recipe.dm
```diff
@@ -229,7 +230,7 @@ var/global/list/_cooking_recipe_cache = list()
 	// the result proc needs to check the list for procedural products.
 	var/list/used_ingredients = list(
 		"items"    = list(),
-		"fruit"    = list(),
+		"fruits"    = list(),
```
MistakeNot4892, line 233: This is going to conflict with the soup PR rip.

MistakeNot4892 (approved):

code/modules/food/cooking/_recipe.dm
noelle-lavenza: Oh, that explains why it didn't runtime when it was based off of the soup PR! Okay, that makes sense

PR #4510, opened 2024-10-06
qdel() now warns if a datum is still processing after Destroy()

noelle-lavenza:
qdel() now prints a stack trace (non-aborting runtime error that marks a CI run as failed) if a datum is still processing after Destroy() resolves. I can't put it in `/datum/Destroy()` because of the geothermal extension doing `STOP_PROCESSING()` after the parent call for some reason, and this is more robust against downstreams doing things like that, so...

This won't catch all cases (for example, temperature processing doesn't use is_processing) but it should catch a bunch.

Requires #4507 as a dependency, but it should be noted that before I rebased it, it also caught those as well.

### DIFF
--- a/code/controllers/subsystems/garbage.dm
+++ b/code/controllers/subsystems/garbage.dm
@@ -312,6 +312,9 @@ SUBSYSTEM_DEF(garbage)
 		if(isnull(D))
 			return
 
+		if(D.is_processing)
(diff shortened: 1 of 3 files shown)


MistakeNot4892 (approved):
I'll close mine since this covers both.

PR #4511, opened 2024-10-06
Small optimizations to proximity_trigger and get_mobs_and_objs_in_view_fast

healdove:
## Description of changes

`proximity_trigger`:
- Makes turfs_in_range into a turf->TRUE assoc list to speed up lookups.
- Puts a /turf filter in the FOR_DVIEW so we don't check if random atoms are in turfs_in_range.

`get_mobs_and_objs_in_view_fast`:
- Renamed to get_listeners_in_range to better represent what it does.
- No longer returns objs that aren't in global.listening_objects; this is where most of the benefit is.
- Use assoc lists for the turf check; this is minor but does help.
- Adds holopad + replicator to listening_objects so that they still work after the above.

Situations where there's a lot of opaque turf changes will be significantly faster within proximity_trigger.

Situations where there's a lot of visible_messages will be somewhat faster, especially if there's a lot of non-listening objects nearby. Airflow shmack messages in particular benefit from this.

Basically: large explosions are a bit faster now. :)

Tested with a 10/20/30/40 explosion on tradeship bridge. Profile before changes:

```
                                                   Profile results (total time)
Proc Name                                                                           Self CPU    Total CPU    Real Time     Overtime        Calls
-------------------------------------------------------------------------------    ---------    ---------    ---------    ---------    ---------
/datum/proximity_trigger/proc/get_seen_turfs                                           0.703        0.703        0.738        0.251          204
...
/proc/get_mobs_and_objs_in_view_fast                                                   0.448        0.449       -1.084        0.174         1452
/atom/proc/visible_message                                                             0.299        0.832       -0.682        0.145         1421
...
```

After changes:
```
                                                   Profile results (total time)
Proc Name                                                                           Self CPU    Total CPU    Real Time     Overtime        Calls
-------------------------------------------------------------------------------    ---------    ---------    ---------    ---------    ---------
...
/proc/get_listeners_in_range                                                           0.254        0.254        0.268        0.103         1459
...
/datum/proximity_trigger/proc/get_seen_turfs                                           0.070        0.070        0.074        0.002          193
...
/atom/proc/visible_message                                                             0.018        0.275        0.289        0.009         1453
...
```

## Authorship
Just me.

### DIFF
--- a/code/game/objects/items/weapons/grenades/flashbang.dm
+++ b/code/game/objects/items/weapons/grenades/flashbang.dm
@@ -10,13 +10,14 @@
 	var/list/victims = list()
 	var/list/objs = list()
 	var/turf/T = get_turf(src)
-	get_mobs_and_objs_in_view_fast(T, 7, victims, objs)
+	get_listeners_in_range(T, 7, victims, objs)
 	for(var/mob/living/M in victims)
 		bang(T, M)
 
-	for(var/obj/effect/blob/B in objs) //Blob damage here
+	FOR_DVIEW(var/obj/effect/blob/B, 7, T, INVISIBILITY_MAXIMUM) //Blob damage here
 		var/damage = round(30/(get_dist(B,T)+1))
 		B.take_damage(damage, BURN)
+	END_FOR_DVIEW
 
 	new /obj/effect/sparks(loc)
 	new /obj/effect/effect/smoke/illumination(loc, 5, 30, 1, "#ffffff")
--- a/code/_helpers/game.dm
+++ b/code/_helpers/game.dm
@@ -172,7 +172,6 @@
 	return L
 
 // Returns a list of mobs and/or objects in range of R from source. Used in radio and say code.
-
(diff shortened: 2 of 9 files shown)


code/game/objects/items/weapons/grenades/flashbang.dm
```diff
@@ -10,13 +10,14 @@
 	var/list/victims = list()
 	var/list/objs = list()
 	var/turf/T = get_turf(src)
-	get_mobs_and_objs_in_view_fast(T, 7, victims, objs)
+	get_listeners_in_range(T, 7, victims, objs)
```
MistakeNot4892, line 13: Just as a thought, could add a null check to get_listeners_in_range() for mobs/objs and skip their respective loop if no list was supplied. In this case `objs` is no longer used so declaring/populating it is pointless.

MistakeNot4892 (approved):
Generally looks good, if there was some way to remove the now extraneous objs list that would also be good.

PR #4514, opened 2024-10-09
Cleans up Swiss Army knife tool quality reporting.

### LINKED ISSUE #4513: steel combi-knife only works as knife, the screwdriver bits are useless.
#### Description of issue
Despite the description saying:
```
 That's a steel combi-knife. It is a small item.

A small, colourable, multi-purpose folding knife.
It looks like a mediocre crowbar, a mediocre screwdriver, mediocre wirecutters, and a mediocre hatchet.
```
only the knife part seems to work, and only for stabbing things. The screwdriver parts when unfolded do not seem to work as screwdriver, and none of the parts work as crowbar. 

as per description it should be able to un-screw simple panels and stuff, but it doesn't. It does not even seem able to interact with anything aside from stabbing things. (I don't get any feedback or text in log, and the character doesn't seem to even attempt to use it.)

#### Steps to reproduce

Just try to use it as you would any other screwdriver.

  That's a steel combi-knife. It is a small item.

A small, colourable, multi-purpose folding knife.
It looks like a mediocre crowbar, a mediocre screwdriver, mediocre wirecutters, and a mediocre hatchet.

I only started playing three days ago, and tried to use it for two, but it didn't work at all for me. (tried clearing cache)

Client Version: 515
Server Revision: eb6d8da69286af915d6e7d4ad6f80b2c5e564472 - main - 2024-10-07
Game ID: cv4-biNh

#### Issue bingo

MistakeNot4892:
Fixes #4513

### DIFF
--- a/code/game/objects/items/weapons/material/swiss.dm
+++ b/code/game/objects/items/weapons/material/swiss.dm
@@ -19,12 +19,12 @@
 	material_alteration = MAT_FLAG_ALTERATION_COLOR | MAT_FLAG_ALTERATION_NAME
 
 	var/active_tool = SWISSKNF_CLOSED
-	var/can_use_tools = FALSE
 	var/list/tools = list(SWISSKNF_LBLADE, SWISSKNF_CLIFTER, SWISSKNF_COPENER)
 	var/static/list/sharp_tools = list(SWISSKNF_LBLADE, SWISSKNF_SBLADE, SWISSKNF_GBLADE, SWISSKNF_WBLADE)
 
 /obj/item/knife/folding/swiss/Initialize(ml, material_key)
-	set_extension(src, /datum/extension/tool/variable, list(
+	// Variable tool qualities are handled by proc below.
+	set_extension(src, /datum/extension/tool, list(
 		TOOL_CROWBAR =     TOOL_QUALITY_MEDIOCRE,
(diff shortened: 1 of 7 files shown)


code/game/objects/items/weapons/material/swiss.dm
```diff
@@ -115,16 +120,8 @@
 /obj/item/knife/folding/swiss/resolve_attackby(obj/target, mob/user)
 	if((istype(target, /obj/structure/window) || istype(target, /obj/structure/grille)) && active_tool == SWISSKNF_GBLADE)
 		force = force * 8
-		. = ..()
-		update_force()
-		return
-	if(istype(target, /obj/item))
```
MistakeNot4892, line 121: I assume this is to limit the usefulness of the tools, but we have tool quality as this seems pointless.

out-of-phaze, line 121: noting that this changes it so that it can work on things other than items that are normal-sized or smaller. which like, is fine, idrk why bay did it this way to begin with? but it's def a balance change

out-of-phaze (approved):

code/game/objects/items/weapons/material/swiss.dm
MistakeNot4892: Oh I noted this but it's 'pending'.

PR #4520, opened 2024-10-10
Fix various issues with flooring

noelle-lavenza:
- Fixes carpets not having an edge layer defined.
- Repaths fantasy wood tiles to the normal tile type to avoid the issues that come with them being a subtype of dirt.
- Sets the base flooring for all floor tiles to dirt in the fantasy modpack. (TODO: null/TRUE/etc base_flooring -> check z-level data?)
- Makes get_base_flooring() and get_topmost_flooring() work pre-init, by handling the vars being paths.
- Refactors the forage noisemaps to check flooring rather than type.
- Optimises turf init.
  - test_link is now 0.75 seconds faster on Shaded Hills on my hardware, due to some small optimisations like early returns and not using `.`. I also used some faster syntax for inverting the blacklist check; `if(foo) {} else` is faster than `if(!foo)` because `!` adds an extra opcode, rather than just being a branch variant. It doesn't matter in most places but in here and maybe atmos it would be justified.
  - Floor initial reagents are faster; they're always added in the liquid phase explicitly, so we don't need `phase_at_temperature()` (which is relatively expensive). If this ever needs to change, we could probably do some other things like memoisation to make `phase_at_temperature()` faster.

This fixes a bunch of flooring issues and also makes init slightly faster, from ~25 to ~23 seconds on my hardware.


MistakeNot4892:
I don't understand what issues are arising from dirt floors. Making floor base type depend on modpack is really untidy imo, as noted on Discord the floor type should be irrelevant ideally.

mods/content/fantasy/turfs.dm
```diff
@@ -0,0 +1,2 @@
+/turf/floor
+	_base_flooring = /decl/flooring/dirt
```
MistakeNot4892, line 2: Per Discord, I am not a fan of this change. Issues like the mushrooms in the Shrine are better fixed by what you've done in this PR with the forage map - in an ideal universe I'd like turf type to not matter at all beyond `/turf/floor` and `/turf/wall` so that we can optimize things like ChangeTurf() to not touch the underlying turf object, which means changes like this aren't required. Not going to block on it in this case particularly because the flooring system still needs work for layering etc. but yeah, wanted to make it known.

code/game/turfs/floors/_floor.dm
```diff
@@ -110,11 +110,17 @@
 
 /turf/floor/proc/get_base_flooring()
 	RETURN_TYPE(/decl/flooring)
-	return istype(_base_flooring) ? _base_flooring : null
+	if(ispath(_base_flooring))
```
MistakeNot4892, line 113: Not sure if I like this and below since it implies stuff is accessing neighbor flooring before turfs have initialised, but I guess walls do this with material updates too, so it's not a limited scope problem. Probably something to look at down the track.

code/game/turfs/floors/floor_icon.dm
```diff
@@ -209,39 +209,37 @@
...
-			. = TRUE
-	//If is_hole is true, then it's space or openspace
-	else if(opponent.is_open())
-		if(space_smooth == SMOOTH_ALL)
-			. = TRUE
-
-	//If we get here then its a normal floor
-	else if (istype(opponent, /turf/floor))
+	// Just a normal floor
+	if (istype(opponent, /turf/floor))
```
MistakeNot4892, line 213: Thanks for doing this, I did very similar changes in my floor branch which then totally broke smoothing so I decided to leave it alone. Glad someone more competent has taken a hammer to it.

MistakeNot4892 (approved):

code/game/turfs/floors/_floor.dm
noelle-lavenza: not just neighbor flooring! self-flooring as well. anything in SSmapping, like map generation, which wants to check the initial state of the turf, has to go through this path.

PR #4521, opened 2024-10-11
Fix soup issues

noelle-lavenza:
- Fixes soup's `soup_ingredients` infinitely accumulating on transfer
- Fixes soup being solid at room temperature :)

### DIFF
--- a/code/modules/materials/definitions/liquids/materials_liquid_soup.dm
+++ b/code/modules/materials/definitions/liquids/materials_liquid_soup.dm
@@ -28,19 +28,21 @@
 
 	var/allergen_flags = ALLERGEN_NONE
 	var/list/ingredients = list()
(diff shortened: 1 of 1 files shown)


code/modules/materials/definitions/liquids/materials_liquid_soup.dm
```diff
@@ -4,6 +4,8 @@
 	nutriment_factor     = 4
 	hydration_factor     = 5 // Per removed amount each tick
 	glass_name           = "soup"
+	melting_point        = T0C // We assume soup is water-based by default and so it freezes at 0C.
+	boiling_point        = null // It kind of sucks for your soup to boil away honestly
```
MistakeNot4892, line 8: I actually disabled the 'boiling away' mechanic for everything except water on the soups branch, I feel like it should be opt-in.

MistakeNot4892 (approved):

PR #4522, opened 2024-10-12
Neb dev update

Atermonera:
## Description of changes
Re-syncs with Nebula/dev

There's kinda a lot. Keeps us on the bleeding edge.

## Authorship
See commits.

## Changelog
Still not sure how to succinctly handle this for upstream updates...


Atermonera:
This... is not the repository I meant to PR against.

PR #4526, opened 2024-10-13
Debugging proc chain for species setup.

MistakeNot4892:
## Description of changes
Attempts to rewrite the base shapeshifter species/bodytype to actually work.
Promethean logic is from Neerti and will be removed before this is merged, I include it so the species/bodytype can be debugged.

Shapeshifters are completely screwed following the bodytype and species changes over the last few years.

## Authorship
@Neerti has been attempting to port Prometheans, the other code is my attempt to fix the underlying species/bodytype logic.

## Changelog
Nothing player-facing.


MistakeNot4892:
The issue as far as I can tell is that `global.wrapped_bodytypes_by_ref` is not getting set by some proc chain somewhere in the depths of mob/organ init. I have put an hour into this and need to go to sleep so I am opening it for less tired eyes to look at.

MistakeNot4892:
Shapeshifters are pretty screwed, going to ditch this rewrite in favour of a total shapeshifter rewrite starting with #4639.

PR #3882, opened 2024-04-12
Kills /datum/dna.

MistakeNot4892:
## Description of changes
Biting the bullet on this one so there's some kind of forward movement. 
- Deletes `/datum/dna` and all associated hooks.
- Replaces it in organ and mob contexts with `/datum/mob_snapshot` which is generated on demand
- Converts genetic conditions (superpowers and disabilities) to `/decl/genetic_condition`.

## Why and what will this PR improve
With current DNA, the tail is wagging the dog. The overcomplexity makes it prohibitive to work with considering the actual use cases. The replacement is generated from the mob when needed rather than constantly maintained, is MUCH simpler, and doesn't get as tangled up in the ideas implied by naming it 'DNA' as the previous implementation.

## TODO
- [X] Test and see that basic setup and joining actually works.
- [X] Reimpement mutation toggles (decl system?) 
- [x] Readd genetics spells
- [x] Readd anomaly stuff.
- [x] Fix up markings tracking and have it track bodytype as well.
- [x] Readd DNA scrambling.
- [x] Test species and bodytype setting.
- [x] Test bioprinter.
- [x] Test mutagen.
- [x] Test DNA switching artifact.
- [x] Test genetics overlay.

## Authorship
Myself.

## Changelog
Nothing player-facing.

PR #4527, opened 2024-10-14
Fixes Heart Attacks

Neerti:
## Description of changes

Re-adds human override for `is_asystole()` that was removed seemingly by accident in #3882 , which currently causes heart attacks to be permanent and making defibs not do anything useful.

People undergoing cardiac arrest can now potentially be saved again. `is_asystole()` is also used in a few other places and it will accurately give an answer instead of always returning FALSE.

## Authorship

Technically me even if I just brought back the old code as-is, just with correct pathing.


MistakeNot4892 (approved):
Whoops.

out-of-phaze (approved):

PR #4530, opened 2024-10-16
Port missing fixes from Spacefarers

out-of-phaze:
apparently i have like over a year of fixes i just never cherrypicked lmao
there will be another PR for dev-specific stuff

### DIFF
--- a/code/modules/client/preference_setup/general/01_basic.dm
+++ b/code/modules/client/preference_setup/general/01_basic.dm
@@ -36,7 +36,8 @@
 	W.write("real_name",             pref.real_name)
 	W.write("name_is_always_random", pref.be_random_name)
 	var/decl/spawnpoint/spawnpoint = GET_DECL(pref.spawnpoint)
-	W.write("spawnpoint", spawnpoint.uid)
+	if(spawnpoint)
+		W.write("spawnpoint", spawnpoint.uid)
 
 /datum/category_item/player_setup_item/physical/basic/sanitize_character()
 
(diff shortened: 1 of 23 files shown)


code/modules/client/preference_setup/general/01_basic.dm
```diff
@@ -36,7 +36,7 @@
 	W.write("real_name",             pref.real_name)
 	W.write("name_is_always_random", pref.be_random_name)
 	var/decl/spawnpoint/spawnpoint = GET_DECL(pref.spawnpoint)
-	W.write("spawnpoint", spawnpoint.uid)
+	W.write("spawnpoint", spawnpoint?.uid)
```
MistakeNot4892: This is going to write nulls, which seems... bad.

out-of-phaze: i guess? it's worked fine on spacefarers, this was a really early change I just never mirrored. the alternative was runtiming

MistakeNot4892: I would honestly say this should be sanitized elsewhere in prefs, it should use one of the map spawnpoints or something instead of keeping it null. Not a huge fan of writing nulls to the save file json.

out-of-phaze: but there *are* no spawnpoints for spacefarers, because there's no roundstart jobs

MistakeNot4892: Yeah you explained on Discord.
As we discussed I do think it should just not write anything if there's no spawnpoint decl rather than writing a null.

MistakeNot4892 (changes requested):
Please change the spawn point pref saving to just skip if there's no spawnpoint decl, otherwise this is good to merge.

MistakeNot4892 (approved):

PR #4531, opened 2024-10-16
Port missing dev branch fixes from Spacefarers

out-of-phaze:
Same as #4530 but for stuff that has to go to dev.

### DIFF
--- a/code/modules/reagents/chems/chems_blood.dm
+++ b/code/modules/reagents/chems/chems_blood.dm
@@ -70,6 +70,10 @@
 		holder.remove_reagent(type, volume)
 	. = ..()
 
+/decl/material/liquid/blood/get_reagent_color(datum/reagents/holder)
+	var/list/blood_data = REAGENT_DATA(holder, type)
+	return blood_data?["blood_color"] || ..()
+
 /decl/material/liquid/coagulated_blood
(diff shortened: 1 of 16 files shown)


code/modules/reagents/chems/chems_blood.dm
```diff
@@ -70,6 +70,10 @@
 		holder.remove_reagent(type, volume)
 	. = ..()
 
+/decl/material/liquid/blood/get_reagent_color(datum/reagents/holder)
+	var/list/blood_data = REAGENT_DATA(holder, type)
+	return blood_data["blood_color"] || ..()
```
MistakeNot4892: Data is lazylisted isn't it?

out-of-phaze: ? yeah? so if it's null we want to return parent

MistakeNot4892: So if it's null this will runtime.

MistakeNot4892 (changes requested):
Please amend the blood data check to use LAZYACCESS() to avoid a null runtime, otherwise this PR is good to merge.

MistakeNot4892 (approved):

PR #4533, opened 2024-10-17
Reworking monkey cubes and monkey cube boxes.

MistakeNot4892:
- Monkey and spider cubes share an abstract subtype.
- Animal cube icons are in their own files instead of food.dmi.


out-of-phaze (approved):
might be worth greyscaling the cube at some point so we could have a brown cube for a monkey, a black or purple cube for a spider (ooh spooky) etc?

PR #4535, opened 2024-10-17
Premium drinks and ration water bottles will no longer freeze en masse.

### LINKED ISSUE #4523: why premium drinks are just ethanol discovered
#### Description of issue
for the longest time whenever i buy "premium drinks" crate it always ends up just ethanol, I realized why when I found out that because they come in a freezer crate all the wine in it freezes and separates into ice and ethanol, proof of this can be seen by the bottle of wine that spawns unfrozen in the matriarch's quarters in molluscstation's mining enclave on scavstation  

I should get more than just pure alchohol and ice in premium drinks and thus it should not come in a freezer crate Image showing off the diffrence between normal and frozen wine: 
image

#### Steps to reproduce
order premium drinks crate, put frozen wine bottle in a drink dispencer and note it's all ethanol and ice as is what happens when booze is cold enough to freeze in this codebase

all the random(?) fancy pants drink bottle names such as 
image
freezer crate, premium drinks crate

for a long time now but we all assumed it was a joke that the premium drinks where just engine degreaser with pretentious names, only now I realise it's a by product of our hardcore freezer crates

Client Version: 515
Server Revision: dd304f526d85f869f1a486d7736947df29380032 - main - 2024-10-12
Game ID: cv9-dpBx

#### Issue bingo

MistakeNot4892:
- Fixes #4523.
- Fixes the bubbling sound at roundstart on tradeship (rations crate in the cargo bay was freezing).

### DIFF
--- a/code/datums/supplypacks/galley.dm
+++ b/code/datums/supplypacks/galley.dm
@@ -108,7 +108,7 @@
 		/obj/item/chems/drinks/bottle/premiumvodka = 3,
 		/obj/item/chems/drinks/bottle/whiskey =      3
 	)
-	containertype = /obj/structure/closet/crate/freezer
+	containertype = /obj/structure/closet/crate/plastic
 	containername = "premium drinks crate"
 
 /decl/hierarchy/supply_pack/galley/barsupplies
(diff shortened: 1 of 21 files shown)


out-of-phaze (approved):
technically this could use a map migration but i'm not pressed abt it

PR #4543, opened 2024-10-29
Apply fixes for TypeMaker issues

out-of-phaze:
Lots of fixes. This is targeting the dev branch because it shook out of an experiment I was doing where I added typehints to like, 30% of the codebase. Woo.

### DIFF
--- a/code/game/machinery/kitchen/icecream.dm
+++ b/code/game/machinery/kitchen/icecream.dm
@@ -206,7 +206,7 @@
 	volume = 20
 	nutriment_amt = 5
 	nutriment_type = /decl/material/liquid/nutriment
-	nutriment_desc = "crunchy waffle cone"
+	nutriment_desc = list("crunchy waffle cone" = 1)
 	var/ice_creamed
 	var/cone_type
 
--- a/code/modules/mob/mob.dm
+++ b/code/modules/mob/mob.dm
@@ -874,7 +874,8 @@
 		if(!canface() || current_posture.prone || restrained())
 			facing_dir = null
 		else if(buckled)
-			if(buckled.obj_flags & OBJ_FLAG_ROTATABLE)
+			var/obj/buckled_obj = buckled
+			if(!isobj(buckled) || (buckled_obj.obj_flags & OBJ_FLAG_ROTATABLE))
 				buckled.set_dir(facing_dir)
 				return ..(facing_dir)
 			else
--- a/code/modules/projectiles/guns/energy/capacitor.dm
+++ b/code/modules/projectiles/guns/energy/capacitor.dm
@@ -64,10 +64,11 @@ var/global/list/laser_wavelengths
 
 	var/wiring_color = COLOR_CYAN_BLUE
 	var/max_capacitors = 2
-	var/list/capacitors = /obj/item/stock_parts/capacitor
+	var/list/capacitors
+	var/initial_capacitor_type = /obj/item/stock_parts/capacitor
 	var/const/charge_iteration_delay = 3
 	var/const/capacitor_charge_constant = 10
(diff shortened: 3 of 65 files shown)


code/game/machinery/kitchen/icecream.dm
```diff
@@ -206,7 +206,7 @@
 	volume = 20
 	nutriment_amt = 5
 	nutriment_type = /decl/material/liquid/nutriment
-	nutriment_desc = "crunchy waffle cone"
+	nutriment_desc = list("crunchy waffle cone" = 1)
```
MistakeNot4892, line 209: I wonder if we could put taste validation into the food icon state tests, since we're already hitting the atoms repo for all the food items.

code/modules/mob/mob.dm
```diff
@@ -874,7 +874,8 @@
 		if(!canface() || current_posture.prone || restrained())
 			facing_dir = null
 		else if(buckled)
-			if(buckled.obj_flags & OBJ_FLAG_ROTATABLE)
+			var/obj/buckled_obj = buckled
+			if(!isobj(buckled) || buckled_obj.obj_flags & OBJ_FLAG_ROTATABLE)
```
MistakeNot4892: Missing parens around the bitflag check.

code/modules/projectiles/guns/energy/capacitor.dm
```diff
@@ -64,10 +64,11 @@ var/global/list/laser_wavelengths
 
 	var/wiring_color = COLOR_CYAN_BLUE
 	var/max_capacitors = 2
-	var/list/capacitors = /obj/item/stock_parts/capacitor
+	var/list/capacitors
+	var/initial_capacitor_type = /obj/item/stock_parts/capacitor
```
MistakeNot4892, line 68: Hmm, the intent of this was avoiding an extra member var that is only used during init and never used again

code/modules/reagents/reagent_containers/food/baking/unleavened_dough.dm
```diff
@@ -7,7 +7,7 @@
 	backyard_grilling_product = /obj/item/food/flatbread
 	backyard_grilling_announcement = "is baked into a simple flatbread."
 	nutriment_amt = 4
-	nutriment_desc = "raw dough"
+	nutriment_desc = list("raw dough" = 1)
```
MistakeNot4892, line 10: Noting to self that it might be feasible to do this wrapping during taste generation instead of compile/init time.

code/modules/projectiles/guns/energy/capacitor.dm
out-of-phaze: i accidentally included some more opinionated changes when I meant for this pr to just be the low hanging fruit, so I can just remove it tomorrow

MistakeNot4892: Nah it's fine. Just thinking aloud really, it would be nice if there was some way around the issue.

code/modules/reagents/reagent_containers/food/baking/unleavened_dough.dm
out-of-phaze: we do that I just don't like the inconsistency
```dm
		if(!islist(nutriment_desc))
			nutriment_desc = list(nutriment_desc)
		for(var/taste in nutriment_desc)
			if(nutriment_desc[taste] <= 0)
				nutriment_desc[taste] = 1
```

MistakeNot4892: But now we've got the overhead of doing the list check AND the overhead of (init).

MistakeNot4892 (approved):

PR #4545, opened 2024-11-01
Rewrites mining drills.

Neerti:
## Description of changes

Completely rewrites the mining drills and their braces, making the code functional, cleaner, and adding a few new things on the side.

All changes done:
* Repaths mining drills from `/obj/machinery/mining/drill` to `/obj/machinery/mining_drill`
* Repaths braces from `/obj/machinery/mining/brace` to `/obj/structure/drill_brace`
* Drill braces are no longer machinery, but are now structures.
    * Drill braces can be crafted in hand with steel sheets now, since they are no longer machinery.
    * Drill brace circuitboard has been removed.
* Drill brace code moved to its own file.
* Drill head uses a state machine to manage what to do now.
    * The drill's light will change color based on the drill's current state.
    * The drill makes sounds when switching between states, e.g. buzzing when it gets full, or pinging when it's finished.
* Adds some sound feedback to certain actions, like when unloading the ore from the drill or when toggling the drill on or off.
* General drilling code broken up into different functions instead of everything being shoved into `Process()`
* Upgrading has been changed:
    * Micro lasers no longer multiply ore out of the ground, but instead cause the drill to go 'faster', and extract more ore for that tick. Power consumption increases linearly.
    * Scanning modules can now be added, which increases the radius that the drill will mine from.
        * The radius is shown to players when they turn the drill on and the drill enters the 'scanning' state, with a brief visual effect.
    * Efficiency formula used when adding capacitors has been changed from `1 / x` to `0.8^(x - 1)`.
        * This makes the efficiency gains less potent with less/lower quality capacitors.
        * A graph that shows the difference between two formulas for energy usage. The green line slopes downwards much slower compared to the red line.
            * X axis: Summed capacitor ratings.
            * Y axis: Energy usage.
            * Red line: Current formula.
            * Green line: New formula.
    * Matter bins remain unchanged.
    * The various upgrades appear to be relatively self-balancing. Stacking mining speed is rough on the cell without also adding efficiency, and higher radii means the drill will likely fill up before finishing without bigger ore storage.
* Icon file has been changed, making the drilling animation and the lights blinking independent of each other. The lights can also be any color now.
* Drill glows slightly while it's active, with the color based on the drill's current state.
* Codex now has an entry for the drill head.
* Drills can be unloaded by clickdragging the drill onto an ore box, as an alternative to using the verb.

Mining drills are functional once more, but now with code that isn't superold, and with a few new things for players to mess around with. It also means you won't experience FOMO with unupgraded drills, and how much ore a place has will not possibly vary wildly because of upgrades.

## Authorship

Me

## Changelog
Player facing changes:
:cl:
del: Mining drill brace circuitboard removed.
tweak: Mining drill braces are now crafted from steel sheets directly.
balance: Microlasers added to mining drills no longer multiply ore out of the ground, but make the drill mine faster, proportionally increasing the energy usage.
add: Mining drills can have scanning modules added to them, allowing the drill to mine a larger radius.
balance: Capacitors added to mining drills are less powerful.
add: You can now clickdrag a mining drill onto an ore box to empty the drill into that ore box.
/:cl:


code/modules/mining/drilling/brace.dm
```diff
@@ -0,0 +1,64 @@
+/obj/structure/drill_brace
+	name = "mining drill brace"
+	desc = "A machinery brace for an industrial drill. It looks like it's about half a meter thick."
```
MistakeNot4892: `metre`

code/modules/mining/drilling/brace.dm
```diff
@@ -0,0 +1,64 @@
...
+	icon = 'icons/obj/mining_drill.dmi'
+	icon_state = "mining_brace"
+	density = TRUE
+	layer = ABOVE_HUMAN_LAYER
+	obj_flags = OBJ_FLAG_ROTATABLE
```
MistakeNot4892: `OBJ_FLAG_ANCHORABLE` would let you remove a lot of the toggle code here, you could move the gating logic into a `wrench_floor_bolts()` override.

code/modules/mining/drilling/brace.dm
```diff
@@ -0,0 +1,64 @@
...
+		visible_message(
+			SPAN_NOTICE("\The [user] [anchored ? "" : "un"]anchors the brace."),
+			SPAN_NOTICE("You [anchored ? "" : "un"]anchor the brace."),
+			SPAN_NOTICE("You hear bolts being adjusted.")
+		)
+
+		if(anchored && connect_to_drill())
+			to_chat(user, SPAN_NOTICE("You attach \the [src] to \the [connected]."))
+		else if(disconnect_from_drill())
+			to_chat(user, SPAN_NOTICE("You detatch \the [src]."))
```
MistakeNot4892, line 34: Probably should return TRUE after this whole IS_WRENCH block.

tools/map_migrations/4545_mining_drills.txt
```diff
@@ -0,0 +1,2 @@
+/obj/machinery/mining_drill/@SUBTYPES : /obj/machinery/mining/drill/@SUBTYPES{@OLD}
```
MistakeNot4892: I think these are backwards, left path is the one you're replacing.

code/modules/mining/drilling/drill_fsm.dm
```diff
@@ -0,0 +1,218 @@
...
+		/decl/state_transition/drill/unpowered,
+		/decl/state_transition/drill/error
+	)
+
+/decl/state_transition/drill/finished
+	target = /decl/state/drill/finished
+
+/decl/state_transition/drill/finished/is_open(obj/machinery/mining_drill/drill)
+	. = ..()
+	return . && !drill.current_turf && !drill.turfs_to_mine.len
```
MistakeNot4892: Minor, but could just do `return ..() && ...`

code/modules/mining/drilling/drill.dm
```diff
@@ -1,299 +1,203 @@
...
-	connected.check_supports()
-	connected = null
-
-/obj/machinery/mining/brace/dismantle()
-	if(connected)
-		disconnect()
-	..()
\ No newline at end of file
+	if(box?.Adjacent(src))
+		if(!contained_ore.len)
```
MistakeNot4892: Minor but `length()` is probably preferred.

code/modules/mining/drilling/brace.dm
```diff
@@ -0,0 +1,64 @@
...
+	return ..()
+
+/obj/structure/drill_brace/proc/connect_to_drill()
+	var/turf/front_turf = get_step(get_turf(src), dir)
+	if(!istype(front_turf))
+		return FALSE
+	var/obj/machinery/mining_drill/drill = locate(/obj/machinery/mining_drill) in front_turf
+	if(drill)
+		connected = drill
+		connected.supports += src
```
MistakeNot4892, line 43: Should this be |=? I'm not seeing supports being cleared.

MistakeNot4892: Might also be worth making supports a list of weakrefs rather than hard references.

Neerti: The drill brace removes itself from the drill head's support list, either when the player unanchors it, or when the drill brace or head gets deleted for whatever reason, so there shouldn't be any dangling refs when either the drill head or brace gets qdel'd. Only reason I can think of for a drill brace to get added twice (which `|=` would prevent iirc) is if something external overwrites the `anchored` var and it becomes inconsistent.

MistakeNot4892 (approved):

PR #4546, opened 2024-11-02
Fixes exoplanets not having any atmosphere

Neerti:
## Description of changes

Exoplanets currently always have vacuum instead of the atmosphere they're supposed to get.
This is because `/datum/level_data/proc/setup_exterior_atmosphere()` gets called first, and sets `exterior_atmosphere` to an empty gas mixture datum, and later on `/datum/level_data/planetoid/proc/apply_planet_atmosphere(var/datum/planetoid_data/P)` gets called, but doesn't do anything because `exterior_atmosphere` is already set, and thus the correct atmosphere held inside of `P.atmosphere` won't be applied.

I was told to just have `exterior_atmosphere` get overwritten regardless if it's set, since it seems that it was changed so it always gets set, so that's what this PR does.

Exoplanets will have atmospheres once again.

## Authorship

Me

### DIFF
--- a/code/modules/maps/template_types/random_exoplanet/random_planet_level_data.dm
+++ b/code/modules/maps/template_types/random_exoplanet/random_planet_level_data.dm
@@ -82,11 +82,12 @@
 	//Rename the surface area if we have one yet
 	adapt_location_name(parent_planetoid.name)
 
-///If we're getting atmos from our parent planet, decide if we're going to apply it, or ignore it
+///If we're getting atmos from our parent planet, apply it.
 /datum/level_data/planetoid/proc/apply_planet_atmosphere(var/datum/planetoid_data/P)
-	if(istype(exterior_atmosphere))
-		return //level atmos takes priority over planet atmos
-	exterior_atmosphere = P.atmosphere.Clone() //Make sure we get one instance per level
+	if(istype(P) && istype(P.atmosphere))
+		exterior_atmosphere = P.atmosphere.Clone()
+		exterior_atmosphere.update_values()
+		exterior_atmosphere.check_tile_graphic()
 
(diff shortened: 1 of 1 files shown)


code/modules/maps/template_types/random_exoplanet/random_planet_level_data.dm
```diff
@@ -82,11 +82,12 @@
...
 	adapt_location_name(parent_planetoid.name)
 
-///If we're getting atmos from our parent planet, decide if we're going to apply it, or ignore it
+///If we're getting atmos from our parent planet, apply it.
 /datum/level_data/planetoid/proc/apply_planet_atmosphere(var/datum/planetoid_data/P)
-	if(istype(exterior_atmosphere))
-		return //level atmos takes priority over planet atmos
-	exterior_atmosphere = P.atmosphere.Clone() //Make sure we get one instance per level
+	if(istype(P))
+		exterior_atmosphere = P.atmosphere.Clone()
```
MistakeNot4892, line 88: Is `P.atmosphere` guaranteed not null?

Neerti: For randomized exoplanets it seems to always create one with `/datum/planetoid_data/random/proc/generate_atmosphere()`. It's possible for non-random exoplanets to have it be null if they didn't supply a preset composition, but a quick examination at the code suggests that there's some other planet stuff like `setup_flora()` that expects a gas mixture datum and has no null checks, so a null atmosphere would probably break stuff like that as well.
I'll still put a check here however since the rest of the exoplanet code seems to have null checks for atmosphere.

MistakeNot4892 (approved):

PR #4555, opened 2024-11-10
[MDB IGNORE] Moving broken/burned icons onto the flooring decl.

MistakeNot4892:
Allows flooring decls to validate or override broken/burned states on a per-icon basis. Needs testing.


MistakeNot4892:
Tested via explosion and fire on Tradeship.

out-of-phaze:
> Tested via explosion and fire on Tradeship.

was it tested with wood floor tiles? previously it only worked for tile

MistakeNot4892:
It's only set up on plating and tiles. Wood etc don't have broken states defined currently and will need them added.

code/unit_tests/turf_icons.dm
out-of-phaze, line 1: line 18, `if(istype(floor))` should be `if(istype(floor, floor_type))` and also `var/turf/floor/check_floor` might just need to be `var/turf/check_turf` tbh... actually that var is unnecessary it should just do `if(TYPE_IS_ABSTRACT(floor_type))` shouldn't it

out-of-phaze (changes requested):

out-of-phaze (dismissed):

out-of-phaze (approved):

PR #4558, opened 2024-11-11
Moving sensor equipping into a separate proc for override/reuse downstream.

MistakeNot4892:


out-of-phaze (approved):
swear i already did this for replika stuff on spacefarers, i'll have to double check i didn't forget to atomize or cherry pick it

PR #4564, opened 2024-11-15
Moves solar daycycle onto base level data type.

MistakeNot4892:
The default 'player level' on SS13 is generally a space level, which should probably have solars, but doesn't use the space level type. This commit makes the default assumption for a level to be having a solars daycycle.


out-of-phaze (approved):
need to make a mental note for myself to change daycycle_type on tether whenever i get around to updating lighthouse but aside from that this looks good

PR #4579, opened 2024-11-22
Refactor injury self-examination, readd wound pressure via agrab

noelle-lavenza:
## Description of changes
Fixes applying wound pressure to yourself.
Fixes applying wound pressure to others (using the prior functionality of help intent with an aggressive grab).
Clicking your health doll now gives you the same status info as clicking your mob on help intent, which is good for cases like drakes where the default help-intent self-interaction is disabled.

The injury self-examination rewrite is used in my Caves of Qud inspired mutations modpack I'm working on.

## Authorship
Me.

## Changelog
:cl:
add: You can apply pressure to another's wounds by aggressively grabbing them and then clicking them with the grab on help intent, with the affected zone targeted.
add: Clicking the health doll now checks yourself for injuries, the same as clicking yourself on help intent if no other action is available.
bugfix: You can once again apply pressure to your own wounds, by clicking yourself on help intent with the affected zone targeted.
/:cl:

### DIFF
--- a/code/_onclick/hud/screen/screen_warnings.dm
+++ b/code/_onclick/hud/screen/screen_warnings.dm
@@ -8,6 +8,12 @@
 	icon_state = "health0"
 	screen_loc = ui_health
 
+/obj/screen/health_warning/handle_click(mob/user, params)
+	if(ishuman(user))
+		var/mob/living/human/human_user = user
+		human_user.check_self_injuries()
+	return TRUE
+
 /obj/screen/warning_cells
 	name = "cell"
 	icon_state = "charge-empty"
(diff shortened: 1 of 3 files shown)


noelle-lavenza:
This could possibly target staging because of the fixes but it's also a rewrite, and those aren't really able to be extricated easily, so.

code/_onclick/hud/screen/screen_warnings.dm
```diff
@@ -8,6 +8,12 @@
 	icon_state = "health0"
 	screen_loc = ui_health
 
+/obj/screen/health_warning/handle_click(mob/user, params)
+	if(ishuman(user))
+		var/mob/living/human/human_user = user
+		human_user.check_self_injuries()
```
MistakeNot4892, line 14: Just a thought, but having the base proc be on /mob would be good. Potentially we could have all the health markers show the status messages down the track.

MistakeNot4892 (approved):

PR #4582, opened 2024-11-24
Implementing Doe's new banners and signs.

MistakeNot4892:
Random debug types to show the concepts:
image
image

Icons by @PlayerDeer.


code/game/objects/structures/banners.dm
```diff
@@ -64,95 +79,279 @@
...
-		var/image/I = image(banner.icon, "banner_base")
+	if(!istype(banner))
+		return
+
+	var/image/I = image(banner.icon, "[banner.icon_state]-hanging")
+	I.appearance_flags |= RESET_COLOR
+	I.color = banner.color
+	add_overlay(I)
+
+	if(banner.trim_color)
```
out-of-phaze: kind of feel like the decals should be added to the banner state first, with `BLEND_INSET_OVERLAY`, and then we apply the trim on top to keep the layering sane. that way you could have user-added decals like stripes/crosses/etc that aren't mutually exclusive with other decals and work regardless of banner shape. this probably works fine as-is i just wanted to point out that inset overlay blend mode could be useful here now or in the future

out-of-phaze (dismissed):
presumably at some point modpacks should be able to add additional banner icons, e.g. corporate logos as decals in the corporate modpack, holy symbols as decals in the mundane/modern earth modpack, etc.

would also be nice to, in the future, have it list decals on examine, including their color if supported. `"This is a <font color="white">banner</font> with a <font color="red">sword</span> embroidered on it."`, `"This is a walnut sign with a beer mug embroidered on it." etc.

for now this looks good, might be worth adding a unit test in the future that checks if the various states exist for every banner type's base_icon_state, like banner-hanging etc.

code/game/objects/structures/banners.dm
MistakeNot4892: Done.

MistakeNot4892:
````
presumably at some point modpacks should be able to add additional banner icons, e.g. corporate logos as decals in the corporate modpack, holy symbols as decals in the mundane/modern earth modpack, etc.
````
Symbols are now decls, so this is done.

````
would also be nice to, in the future, have it list decals on examine, including their color if supported. "This is a <font color="white">banner</font> with a <font color="red">sword</span> embroidered on it.", `"This is a walnut sign with a beer mug embroidered on it." etc.
````
Simple version of this is done.

out-of-phaze (approved):

PR #4585, opened 2024-11-28
Intents are now /decl/intent.

MistakeNot4892:
## Description of changes
- Adds getters/setters and helpers to intent checking, replacing all use of `a_intent`.
- Rewrites intent selectors to condense/improve code and remove separate intent tracking.
- Removes direct intent setting on mob definitions in favour of setting via AI init.
- Changes intent strings to `/decl/intent` with handling/validation for such.
- Removes `intent_numeric()` in favour of using get_next_in_list() and get_previous_in_list() for hotkey cycling.
- Removes outdated robot intent selector limitations.
- Adds intent flags preferentially used check checking intent, rather than strict intent types.

This is more or less just an infrastructure update to support dynamic intents when I get to them, but it does also clean up some 12 year old crufty code. Tested with humans, robots and nymphs.

## Authorship
Myself.

## Changelog
Nothing player-facing.


MistakeNot4892:
There's a good chance the deity changes break deity, but I am intending to delete it shortly so I am not going to test it.

MistakeNot4892:
Need to retest with the diona and psionics changes.

MistakeNot4892:
Commented out nymph intents/intent selectors entirely, they need the full selector for interactions to work and I am unwilling to do more icon work for this PR to fix such a niche side of it. Will revisit with future intent PRs.

out-of-phaze (dismissed):
I'm not super sure why they're flags now, is it so you can check if someone has either of two intents? That makes sense, I guess...

MistakeNot4892:
> I'm not super sure why they're flags now, is it so you can check if someone has either of two intents? That makes sense, I guess...

Initially it was so the binary intent selector could be both harm and grab or disarm and hurt. That broke interactions though. I could roll it back if desired but it might be useful down the track.

out-of-phaze (approved):
36hrs with conflict resolution, looks good to go for testing on dev

PR #4587, opened 2024-11-30
Adding a priority list system for mouse cursor icons.

MistakeNot4892:
## Description of changes
Adds infrastructure for handling mouse pointer overrides.

I need to start adding more cursors for bows and melee weapons so need some infrastructure.

## Authorship
Myself.

## Changelog
Nothing player-facing.


MistakeNot4892:
Fixes/improvements to sideport from my melee branch, DNM.

code/modules/client/mouse_pointer/_mouse_pointer.dm
```diff
@@ -0,0 +1,69 @@
...
+/client/proc/add_mouse_pointer(pointer_type, pointer_priority = 1, icon_index = 1)
+
+	var/decl/mouse_pointer/pointer_decl = GET_DECL(pointer_type)
+	if(!isnum(icon_index) || icon_index < 1 || icon_index > length(pointer_decl.icons))
+		CRASH("Invalid icon_index passed to add_mouse_pointer() for [pointer_type].")
+
+	// Is an identical pointer already being tracked?
+	var/pointer_icon = pointer_decl.icons[icon_index]
+	if(pointer_type in _mouse_pointers)
+		var/list/comparing = _mouse_pointers[pointer_type]
```
out-of-phaze: not hot code, but i would personally do something like
```suggestion
	var/list/comparing = _mouse_pointers?[pointer_type]
	if(comparing)
```
i'm a little wary of using `in` on assoc lists because a lookup is faster, iirc. again, not performance critical and arguably the current code is easier to understand

code/modules/client/mouse_pointer/_mouse_pointer.dm
```diff
@@ -0,0 +1,69 @@
...
+	if(LAZYLEN(_mouse_pointers) > 1)
+		_mouse_pointers = sortTim(_mouse_pointers, /proc/cmp_priority_list, TRUE)
+		need_update = TRUE
+
+	// Refresh if needed.
+	if(need_update)
+		update_mouse_pointer()
+
+/client/proc/remove_mouse_pointer(pointer_type)
+	if(!LAZYLEN(_mouse_pointers) || !_mouse_pointers[pointer_type])
```
out-of-phaze: could be written more succinctly by using the null indexing operator. just down to personal preference though
```suggestion
	if(!_mouse_pointers?[pointer_type])
```

code/modules/client/mouse_pointer/mouse_pointer_definitions.dm
```diff
@@ -0,0 +1,31 @@
...
+			"",
+			"over",
+			"drag",
+			"drop",
+			"all"
+		)
+		for(var/icon in icons)
+			for(var/check_state in check_states)
+				if(!check_state_in_icon(check_state, icon))
+					. += "missing state '[check_state]' from icon '[icon]'"
```
out-of-phaze, line 24: i wonder if it would be possible to make a helper that checks multiple icon states faster? either through list intersection or some other method. not necessary for functionality but i figure this is probably done elsewhere

code/modules/client/mouse_pointer/mouse_pointer_definitions.dm
```diff
@@ -0,0 +1,31 @@
...
+	else
+		. += "null or empty icon list"
+
+// Subtypes for use in add_mouse_pointer() below.
```
out-of-phaze, line 28: there is nothing below :)

code/modules/mob/mob_defines.dm
```diff
@@ -70,9 +70,6 @@
 	*/
 	var/obj/screen/zone_selector/zone_sel = null
 
-	/// Cursor icon used when holding shift over things.
-	var/examine_cursor_icon = 'icons/effects/mouse_pointers/examine_pointer.dmi'
```
out-of-phaze, line 74: for parity with the prior system we could specify an examine cursor *decl* on mob instead

out-of-phaze (dismissed):
honestly i kind of assumed this would be using `/datum/priority_queue`, but this is fine too, especially since that type is super old

code/modules/mob/mob_defines.dm
MistakeNot4892: I don't understand what you mean by parity with the prior system. This var was never changed or used outside of being passed to the pointer.

code/modules/client/mouse_pointer/mouse_pointer_definitions.dm
MistakeNot4892: The only thing you'd save would be lookup time in the icon_states cache, doesn't seem worth the additional complexity for the sake of unit tests. Wouldn't oppose it, but not going to touch that proc in this PR.

LowFP:
Is there a cursor size? I can try making some.

MistakeNot4892:
The existing icon is 32x32

code/modules/client/mouse_pointer/mouse_pointer_definitions.dm
MistakeNot4892: image

out-of-phaze: i misunderstood and thought it meant the proc was below lol

code/modules/mob/mob_defines.dm
out-of-phaze: mostly that it's a mob var that can be overridden

out-of-phaze (approved):

PR #4593, opened 2024-12-02
Adding gemstones.

MistakeNot4892:
## Description of changes
- Adds gemstones that can be cut to various shapes. Cutting improves value but requires a skill check.
- Adds possible gemstone drops to turf digging and natural wall mining.

Prep work for jewellery stuff (rings and necklaces with gems).

## Authorship
@PlayerDeer did the icons for the gems, I wrote the code.

## Changelog
:cl:
add: Mining granite or marble, or digging clay, may drop a gemstone.
/:cl:


code/game/turfs/floors/floor_digging.dm
```diff
@@ -1,3 +1,6 @@
+/turf/floor
+	var/gemstone_dropped = FALSE
```
out-of-phaze, line 2: should this be shared on /turf to avoid the duplicate natural wall var? should it persist through changeturf or should it be reset?

MistakeNot4892:
Tested, ready to go.

code/game/turfs/turf_digging.dm
```diff
@@ -1,5 +1,5 @@
-/// Return an assoc list of resource item type to a base and a random component
-/// ex. return list(/obj/item/stack/material/ore/handful/sand = list(3, 2))
+/// Return an assoc list of resource item type to a metadata list containing base amount, random component, and material override
+/// ex. return list(/obj/item/stack/material/ore/handful/sand = list("amount" = 3, "variance" = 2, "material" = /decl/material/foo))
```
out-of-phaze, line 2: can material be skipped/null?

code/game/turfs/floors/floor_digging.dm
MistakeNot4892: Keeping it on just the two specific types that need it seems fine, short of a less bespoke solution to one-shot turf resources. With regards to changeturf, I'm not sure. I guess you could potentially build a wall over clay and dismantle it to reset them gem counter as it currently stands.

code/game/turfs/turf_digging.dm
MistakeNot4892: Yep, it only really needs the type. An entry with no metadata list will spawn 1 of the type with whatever the type has as default material.

out-of-phaze (dismissed):

out-of-phaze (approved):
Reapproving after conflict resolution with intents PR.

PR #4592, opened 2024-12-02
Updating necklaces and rings to use new icons.

MistakeNot4892:
## Description of changes
- Items now have a list of decorations (/decl/item_decoration) that modify onmob and inv/world icons.
- Material rings have been retyped to just be the basic ring type.
- Rings and necklaces now use colourable icons and overlays.
- Rings and necklaces have handling for gem materials and overlays.
- Dummy gem materials have been added (with crap colours).
- Necklaces support a 'pendant' item that draws on the item and mob.

~~Noting that the gem stuff is very placeholder, I'd like to do another PR down the track to add gems and gemcutting using Doe's other icons, but I am not going to get around to it in the short term so slapped in a crappier version on this one.~~ Got around to doing this.

## Why and what will this PR improve
Nicer jewellery, more interesting jewellery code.
Handing for future stuff like glazes for ceramics or insets on armor.

## Authorship
@PlayerDeer did the icons, I just plugged in the code.

## Changelog
Nothing player-facing.

PR #4594, opened 2024-12-02
Adding various fancy goods from Doe's icon set.

MistakeNot4892:
Adds bowls, goblets, decanters and vases from the giant pile of icons @PlayerDeer has made recently. Nothing uses them yet but they are available for subtyping/mapping.

image


MistakeNot4892:
Depends on #4592

code/modules/crafting/stack_recipes/recipes_planks.dm
```diff
@@ -78,6 +78,26 @@
 /decl/stack_recipe/planks/bowl
 	result_type            = /obj/item/chems/glass/handmade/bowl
 
+/decl/stack_recipe/planks/fancy
```
out-of-phaze, line 81: huh, why are the fancy cups... wooden?

out-of-phaze (approved):

code/modules/crafting/stack_recipes/recipes_planks.dm
MistakeNot4892: Because we don't really have metalcrafting yet.

MistakeNot4892: Oh, I remembered, this is so you can make molds of them. Could put them on stone instead possibly.

out-of-phaze: no that makes sense!

PR #4596, opened 2024-12-03
Adding a silhouette that flashes when hit by an attack.

MistakeNot4892:
Trying to add a full silhouette solid colour flash when hit by an attack - having some issues, so opening a draft for review/input:
- Silhouette does not respect animated transforms (giant spider attack windup for example)
- Silhouette includes attack animation overlays (wrench etc).
- Humans don't proc the flash despite the stuff I've added to their code.


MistakeNot4892:
Nice to have but hard to implement.

PR #4598, opened 2024-12-03
Cleaning up a lot of sprite_sheets setting.

MistakeNot4892:
## Description of changes
All sprite_sheets setting in modpacks now uses item vars to track relevant icons.

Allows overriding sprite_sheets on subtypes.

## Authorship
Myself.

## Changelog
Nothing player-facing.

### DIFF
--- a/mods/species/ascent/items/rig.dm
+++ b/mods/species/ascent/items/rig.dm
@@ -3,7 +3,6 @@
 	name = "alate support exosuit"
 	desc = "A powerful support exosuit with integrated power supply, weapon and atmosphere. It's closer to a mech than a rig."
 	icon = 'mods/species/ascent/icons/rig/rig.dmi'
-
 	suit_type = "support exosuit"
 	armor = list(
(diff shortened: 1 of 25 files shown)


mods/species/ascent/items/rig.dm
```diff
@@ -1,9 +1,19 @@
+/obj/item
```
out-of-phaze: placement in this file feels weird but it's probably fine

mods/species/bayliens/tajaran/machinery/suit_cycler.dm
```diff
@@ -1,71 +1,62 @@
+/obj/item
```
out-of-phaze: same here, kind of weird for it to be in a machinery file

out-of-phaze (dismissed):
kinda weird choice of files to put the vars in but if it works it works

out-of-phaze (approved):

PR #4604, opened 2024-12-04
Adding a lounge.

MistakeNot4892:
Adds a lounge bench subtype using @PlayerDeer's fancy couch icons.


code/game/objects/structures/stool_bed_chair_nest_sofa/pew.dm
```diff
@@ -132,4 +132,20 @@
 
 /obj/structure/bed/chair/bench/pew/ebony
 	color    = /decl/material/solid/organic/wood/ebony::color
-	material = /decl/material/solid/organic/wood/ebony
\ No newline at end of file
+	material = /decl/material/solid/organic/wood/ebony
+
+/obj/structure/bed/chair/bench/lounge
+	name       = "lounge"
```
out-of-phaze, line 138: to be totally honest I've never heard the word "lounge" used for a lounge chair or fancy sofa, might be an aussie thing, but it's probably fine

edit: for reference, at least in the US, a "lounge" refers exclusively to a type of room. it would probably not be unexpected to find this type of sofa in a lounge, but it wouldn't really be called a lounge

out-of-phaze (approved):

code/game/objects/structures/stool_bed_chair_nest_sofa/pew.dm
MistakeNot4892: https://en.wikipedia.org/wiki/Chaise_longue

````
This variant has been documented in British[3] texts since at least 1811 and in American texts[4] since 1824.[5]
````

PR #4605, opened 2024-12-05
Removed deity mode.

MistakeNot4892:
Completely removes deity mode.

Rationale:
- It is unmaintained.
- It has a very questionable space within lore/gameplay on Neb.
- It makes it actively difficult to change or rework things like cult and wizard without also reworking deity due to how it wraps/integrates parts of those modes.
- Participant vote was in favour.


out-of-phaze (dismissed):

out-of-phaze (approved):
conflict resolution, >24h

PR #4608, opened 2024-12-05
Abolishes the modules folder.

MistakeNot4892:
## Description of changes
- Moves everything from `code/modules` into `code/game`.
- Moves some atmos defines into `atmospherics.dm`.

`modules` folder has been a lie for years, none of it is modular. This is clearer and lays the groundwork for moving stuff that actually is modular into the mods folder.

## Authorship
Myself.

## Changelog
Nothing player-facing.


out-of-phaze (changes requested):
can you open the dme in dreammaker and compile, then save the dme? that will hopefully re-alphabetize things properly so that the checks pass. also the commit name has a `d` at the start

MistakeNot4892:
If I open it in DM on my VM it changes the line encodings and I've yet to work out how to change it back without checking out and manually editing the dme :(

MistakeNot4892:
The time has not yet arrived...

PR #4609, opened 2024-12-05
Replace direct color sets with set_color, various color tweaks

out-of-phaze:
## Description of changes
Renames a file, makes cables use paint_color, replaces a bunch of item/structure color sets with set_color, and simplifies shard alpha calculation. Also rewrites table icon code, now it's split, which should make desks less sucks (eventually)? maybe?

code :)


MistakeNot4892 (approved):

MistakeNot4892 (changes requested):
Needs a fix for tables getting double colour applied per Pyrelight volatile test.

MistakeNot4892 (approved):

PR #4611, opened 2024-12-06
Cleaning up some submap archetype code

MistakeNot4892:
## Description of changes
- Submap archetype code uses get_decls_of_type_unassociated().
- Maps can supply distinct submap archetype lists instead of always using the full set.
- Submap `descriptor` is now `name` so that they'll behave nicely in input() in the future.
- Submap name uniqueness validation is now in submap archetype validate()

No crashed pods on Shaded Hills.

## Authorship
Myself.

## Changelog
Nothing player-facing.

### DIFF
--- a/code/modules/submaps/submap_archetype.dm
+++ b/code/modules/submaps/submap_archetype.dm
@@ -1,5 +1,6 @@
 /decl/submap_archetype
-	var/descriptor = "generic ship archetype"
+	// TODO: use UID instead of name for pref saving.
+	var/name = "generic ship archetype"
 	var/list/whitelisted_species = list()
 	var/list/blacklisted_species = list()
 	var/call_webhook
@@ -13,12 +14,18 @@
 
 /decl/submap_archetype/validate()
 	. = ..()
-	if(!descriptor)
-		. += "no descriptor set"
+	if(name)
+		var/static/list/submaps_by_name = list( (global.using_map.name) = global.using_map.type)
+		if(submaps_by_name[name])
+			. += "name '[name]' ([type]) collides with submap type '[submaps_by_name[name]]'"
+		else
+			submaps_by_name[name] = type
+	else
+		. += "no name set"
 
 // Generic ships to populate the list.
 /decl/submap_archetype/derelict
-	descriptor = "drifting wreck"
+	name = "drifting wreck"
 
 /decl/submap_archetype/abandoned_ship
-	descriptor = "abandoned ship"
+	name = "abandoned ship"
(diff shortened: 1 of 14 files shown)


code/modules/submaps/submap_archetype.dm
```diff
@@ -13,12 +13,22 @@
 
 /decl/submap_archetype/validate()
 	. = ..()
-	if(!descriptor)
-		. += "no descriptor set"
+	if(!name)
+		. += "no name set"
+	else
+		var/static/list/submaps_by_name
```
out-of-phaze: i don't think this is a good idea, it'll loop over N submaps in N submaps' `validate()` proc. this was a uniqueness unit test for a reason

out-of-phaze (dismissed):

code/modules/submaps/submap_archetype.dm
MistakeNot4892: It was a uniqueness test because validate() didn't exist when I wrote it. I'll move it to a global instead of static and it will only be built once, if that addresses the issue.

out-of-phaze: in my opinion nothing that references/compares with other decls of the same type should be in validate, except for maybe things like parent/child relationship stuff on hierarchies/traits/etc.

MistakeNot4892: Why? We have decl validation as a single test that condenses all decl problems in one spot in the log, I don't see why we need a full fat seperate uniqueness test for this when one list and a collision check does the job.

code/modules/submaps/submap_archetype.dm
```diff
@@ -11,14 +11,24 @@
...
 /decl/submap_archetype/validate()
 	. = ..()
-	if(!descriptor)
-		. += "no descriptor set"
+	if(!name)
+		. += "no name set"
+	else
+		if(!global.submaps_by_name)
+			global.submaps_by_name = list()
+			for(var/decl/submap_archetype/submap as anything in decls_repository.get_decls_of_type_unassociated(/decl/submap_archetype))
```
out-of-phaze: i wonder if this is actually necessary? we're already looping over all submap archetypes in the validation test, so theoretically it could just be
```dm
		if(global.submaps_by_name[name])
			. += "name '[name]' conflicts with submap type '[global.submaps_by_name[name]]'"
		else
			global.submaps_by_name[name] = type
```
and obviously submaps_by_name would have to always be a list for this code to work exactly as i wrote it
that said i think it should be a static var to avoid someone trying to access it outside of validation

MistakeNot4892: Yeah good call, makes it much simpler.

out-of-phaze (approved):

PR #4612, opened 2024-12-07
Various clothing and equipment tweaks for Karzerfeste.

MistakeNot4892:
- Outfits can be given lists of types instead of single types.
- Added sleeved robes.
- Added various clothing and tool subtypes for fantasy.


code/datums/outfits/outfit.dm
```diff
@@ -136,36 +170,37 @@
...
-		wearer.equip_to_slot_or_del(new r_pocket(wearer),slot_r_store_str)
-	if(suit_store)
-		wearer.equip_to_slot_or_del(new suit_store(wearer),slot_s_store_str)
+		for(var/r_ear_type in resolve_equip_to_list(r_ear))
+			var/r_ear_path = (OUTFIT_ADJUSTMENT_PLAIN_HEADSET & equip_adjustments) && ispath(r_ear_type, /obj/item/radio/headset) ? /obj/item/radio/headset : r_ear_type
+			try_equip(wearer, r_ear_path, slot_r_ear_str)
+
+	// Ears, backpack and uniforms are handled individually.
+	var/list/slot_to_gear  = list(
+		slot_wear_suit_str = resolve_equip_to_list(suit),
```
out-of-phaze: this is a define, so it might actually work, but i'm a little worried about using bare list keys like this. actually no i know for a fact it'll work until someone makes a typo or include order error when adding a new one, and then it'll insert the nonexistent variable as a string. so... dunno. fine as is i guess but wrapping it in parens is technically safer

out-of-phaze (dismissed):

code/datums/outfits/outfit.dm
MistakeNot4892: I can do that for clarity easily enough.

out-of-phaze (approved):

PR #4613, opened 2024-12-07
Adding a large cauldron that cannot be picked up.

MistakeNot4892:
Adds a large cauldron for mapping purposes. Icon by @PlayerDeer.


out-of-phaze (approved):
i wonder if there are some interactions that aren't possible/feasible without being able to pick it up and pour out of it. guess that's what testing is for

PR #4617, opened 2024-12-07
Adds a grandfather clock.

MistakeNot4892:
Adds a very basic standing clock. Icon by @PlayerDeer.


code/game/objects/structures/grandfather_clock.dm
```diff
@@ -0,0 +1,65 @@
...
+// TODO: looping 2 second tick tock sound, somehow aligned with pendulum (may not be possible in DM)
+/obj/structure/grandfather_clock
+	name       = "grandfather clock"
+	desc       = "A tall, stately timepiece."
+	icon       = 'icons/obj/structures/grandfather_clock.dmi'
+	icon_state = ICON_STATE_WORLD
+	density    = TRUE
+	material   = /decl/material/solid/organic/wood/mahogany
+	var/last_time
+	var/decl/material/face_mat      = /decl/material/solid/organic/bone
```
out-of-phaze: apparently common materials for grandfather clock faces/dials were uh. brass, painted wood, painted iron, painted/engraved (opaque?) glass. a lot of them were painted though
for the purpose of this, if we don't wanna add a 'face_paint_color' variable, i would maybe suggest porcelain instead of bone lol? as in `/decl/material/solid/stone/ceramic`

MistakeNot4892: Sucked it up and went with a color var. Can revisit this when we work out a system for constructing complex mechanisms.

out-of-phaze (approved):

PR #4619, opened 2024-12-07
Fix duplicate `name_prefix` var

ike709:
## Description of changes
`var/name_prefix` was declared on both `/obj/item` and `/obj/item/banner`

This is due to https://github.com/NebulaSS13/Nebula/pull/4582 and https://github.com/NebulaSS13/Nebula/pull/4603 adding the same var to each type.

This errors in BYOND and is also breaking OpenDream CI.


out-of-phaze:
ha gottem. thanks though

PR #4628, opened 2024-12-08
Synthetics do not get hungry or thirsty.

### LINKED ISSUE #4626: FBPs suffer from hunger effects
Copied from https://github.com/ScavStation/ScavStation/issues/1137

#### Description of issue
despite being all robot I found myself walking weirdly slow, then I realized what was going on when I tried to jump over a table... "YOU ARE TOO HUNGRY TO JUMP!" ...despite being a FBP robot and thus cannot eat.

robots should not get hungry especially if they cannot eat

#### Steps to reproduce
be a yinglet gynoid or something
play until you get weirdly slow
attempt a jump

yinglet gynoid 

found it now

Client Version: 515
Server Revision: db8bd2dc81b2282378fb43076f5089a91f23cfe7 - dev - 2024-11-29
Game ID: cwV-dMe8

#### Issue bingo

MistakeNot4892:
- Fixes #4626

### DIFF
--- a/code/modules/mob/living/life.dm
+++ b/code/modules/mob/living/life.dm
@@ -121,7 +121,7 @@
 	return TRUE
 
 /mob/living/proc/experiences_hunger_and_thirst()
-	return TRUE
+	return !isSynthetic() // Doesn't really apply to robots. Maybe unify this with cells in the future.
 
 /mob/living/proc/get_hunger_factor()
 	var/decl/species/my_species = get_species()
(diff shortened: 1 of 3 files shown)


code/modules/mob/living/life.dm
```diff
@@ -121,7 +121,7 @@
 	return TRUE
 
 /mob/living/proc/experiences_hunger_and_thirst()
-	return TRUE
+	return !isSynthetic() // Doesn't really apply to robots. Maybe unify this with cells in the future.
```
out-of-phaze, line 124: really wary of an isSynthetic check, this should probably be a bodytype check if we're doing it right. i get wanting to slap in a quick fix though

out-of-phaze (approved):

PR #4630, opened 2024-12-09
Removed slavers' den.

MistakeNot4892:
It's a pretty two-dimensional and crap away site full of hostile mobs and the 'anti-slavery abolitionists' being insane berserkers has always struck me as weird political messaging that leaves a bad taste in my mouth.


out-of-phaze (approved):
iirc bay had some sort of weird justification for why the abolitionists tried to kill the torch on sight, but uh. it sucks! begone

PR #4631, opened 2024-12-09
More robust handling for slapcrafting holder creation.

### LINKED ISSUE #4625: Slapcrafting does not properly update storage UI if ingredient is in backpack
Copied from https://github.com/ScavStation/ScavStation/issues/1140

#### Description of issue

With zip gun, if you stick the pipe into a stock while stock is in your backpack, it drops a half-assembled weapon at your feet while the original stock remains in the backpack and loses its "size" sprite so if not pulled out right away, it becomes completely unremovable (although it also stops taking up space).
With coilgun applying sheets to stock while it's in the bag produces same result.
No further assembly steps cause a similar effect.

I'm pretty sure it is not supposed to duplicate itself or become permanently stuck in the backpack.

#### Steps to reproduce

- Make a stock from wooden planks.
- Put stock in the bag.
- Apply pipe/sheets (depending on weapon type) to the stock while it is still in the bag.

Zipgun frame -> zipgun frame assembly
Coilgun stock -> coilgun stock assembly

Noticed this shortly before we switched to staging, so it was already  present on regular build, and it is still present two stagings later at the day this is written.

Client Version: 515
Server Revision: db8bd2dc81b2282378fb43076f5089a91f23cfe7 - dev - 2024-12-08
Game ID: cw4-angZ

#### Issue bingo

MistakeNot4892:
- Fixes #4625
- `5 sheetss` -> `5 steel sheets`

### DIFF
--- a/code/modules/crafting/slapcrafting/_crafting_holder.dm
+++ b/code/modules/crafting/slapcrafting/_crafting_holder.dm
@@ -13,17 +13,13 @@
 		var/list/next_products = list()
 		for(var/decl/crafting_stage/next_stage in current_crafting_stage.next_stages)
 			if(ispath(next_stage.completion_trigger_type))
-				var/atom/next_tool = next_stage.completion_trigger_type
-				var/tool_string = initial(next_tool.name)
-				if(next_stage.stack_consume_amount > 1)
-					tool_string = "[next_stage.stack_consume_amount] [tool_string]\s"
-				else
-					tool_string = "\a [tool_string]"
+				var/tool_string = next_stage.generate_completion_string()
 				if(ispath(next_stage.product))
 					var/atom/next_product = next_stage.product
 					next_products[tool_string] = "\a [initial(next_product.name)]"
 				else
 					next_steps += tool_string
+
 		if(length(next_products))
 			for(var/thing in next_products)
 				to_chat(user, SPAN_NOTICE("With <b>[thing]</b>, you could finish building <b>[next_products[thing]]</b>."))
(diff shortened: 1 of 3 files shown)


code/modules/crafting/slapcrafting/_crafting_stage.dm
```diff
@@ -20,6 +20,28 @@
...
+	var/list/names = assemble_name_strings()
+	if(ispath(completion_trigger_type, /obj/item/stack) || stack_consume_amount)
+		names.Insert(1, max(stack_consume_amount, 1))
+	return jointext(names, " ")
+
+/decl/crafting_stage/proc/assemble_name_strings()
+	SHOULD_CALL_PARENT(TRUE)
+	var/list/names = list()
+	var/obj/item/prop = completion_trigger_type
+	if(ispath(prop, /obj/item/stack))
```
out-of-phaze, line 33: occurs to me that stack.get_string_for_amount is an instance proc which is unpleasant, otherwise it could be used here. that STATIC_CALL macro I wrote might work but it's also scary

out-of-phaze (dismissed):

code/modules/crafting/slapcrafting/_crafting_holder.dm
```diff
@@ -13,32 +13,38 @@
...
+
+	// Move our component into the new holder.
+	if(ismob(target.loc))
+		var/mob/M = target.loc
 		M.drop_from_inventory(target)
-	target.forceMove(src)
+		target.forceMove(src)
+	else if(ismovable(target.loc))
+		var/atom/movable/holder = target.loc
+		holder.storage?.remove_from_storage(user, target, src)
```
MistakeNot4892: Noting atoms without storage won't forcemove the target, needs a second forcemove call

MistakeNot4892: Fixed.

out-of-phaze (approved):

PR #4632, opened 2024-12-09
Vendor prices now round to avoid partial worth issues.

### LINKED ISSUE #4624: Vending machines require 1 more than listed money and produce a 0-value pile
Copied from https://github.com/ScavStation/ScavStation/issues/1142

#### Description of issue

Attempting to pay with a pile of credits that contains exactly as many as vendor requested says that is not enough money. Paying with a pile that has one more credit than requested results in you having an invisible pile of 0 credits in your hand.

Vendor has some weird calculations regarding how much money it takes.

#### Steps to reproduce

- Pick an item to buy at vendor machine
- Separate exactly as much credits as it asks for from your cash stack
- Get called a lowlife
- Separate a pile that has one more credit than it asks for and pay with it
- End up with a pile of zero credits

Any vendor machine that sell stuff for credits.

  That is not enough money.
You add 53 credits to the pile.
It holds 1505 credits now.
Terket inserts some cash into Getmore Chocolate Corp.
Getmore Chocolate Corp clunks as it vends the lollipop.
You add 0 credits to the pile.
It holds 1451 credits now.

- In this case it was a lollipop that costed 53 credits. Also it was a welding fuel lollipop, wtf.

Had it happen way before the switch to staging but didn't pinpoint the specific circumstances until the day it was written.

Client Version: 515
Server Revision: db8bd2dc81b2282378fb43076f5089a91f23cfe7 - dev - 2024-12-08
Game ID: cw4-angZ

#### Issue bingo

MistakeNot4892:
Fixes #4624.

### DIFF
--- a/code/game/machinery/vending/_vending.dm
+++ b/code/game/machinery/vending/_vending.dm
@@ -107,7 +107,7 @@
 		var/category = current_list[2]
 		for(var/entry in current_list[1])
 			var/datum/stored_items/vending_products/product = new(src, entry)
-			product.price = atom_info_repository.get_combined_worth_for(entry) * markup
+			product.price = ceil(atom_info_repository.get_combined_worth_for(entry) * markup)
 			product.category = category
 			if(product && populate_parts)
 				product.amount = (current_list[1][entry]) ? current_list[1][entry] : 1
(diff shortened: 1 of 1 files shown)


code/game/machinery/vending/_vending.dm
```diff
@@ -213,9 +213,9 @@
 	if(currently_vending.price > cashmoney.absolute_worth)
 		// This is not a status display message, since it's something the character
 		// themselves is meant to see BEFORE putting the money in
-		to_chat(usr, "[html_icon(cashmoney)] <span class='warning'>That is not enough money.</span>")
+		to_chat(usr, SPAN_WARNING("[html_icon(cashmoney)] That is not enough money."))
```
out-of-phaze, line 216: out of scope but i think these REALLY need a `mob/user` param at some point

out-of-phaze (approved):

Issue #4073, opened 2024-06-02
Rabbit did not behave as expected in Shaded Hills


ZirconiumWit:
#### Description of issue
Was messing around in Shaded Hills last night and hunted a rabbit.  Rabbit seemed to bleed out because it suddenly stopped running and just sat there.  My character moved in to try to finish it, and found that I could not interact with it.  My character did not seem able to stab it anymore, but the rabbit did not transition to a death icon.  My character also could not butcher it with my knife or grab it to drag it away.

#### Difference between expected and actual behavior
Hunted animals should not be suspended in an endless limbo of bleeding sadly into the pasture, unable to move or react to the world around them

#### Steps to reproduce
Unsure.  I stabbed the rabbit with the bronze knife, chased it across the river.  It forgot to be frightened of me after I chased it across the river so I stabbed it again. I mention this because we had both just been swimming when things went weird. It fled briefly, acted oddly (It had been fleeing northeast, it suddenly stopped and acted like it couldn't go further north) and then changed direction to start running south and then stopped moving.  It was facing south when it stopped and it seemed to be stuck because I could no longer interact with it at this stage.

#### Specific information for locating

shaded hills map.  The rabbit near the mine processing spot.

#### Length of time in which bug has been known to occur

Discovered last night

#### Client version, Server revision & Game ID

Client Version: 515
Server Revision: 996ac692463405ee41e687d6e9b5c4524f3c8171 - dev - 2024-06-02
Game ID: ct1-a8Aj
Current map: Shaded Hills

#### Issue bingo

- [ ] Issue could be reproduced at least once
- [ ] Issue could be reproduced by different players
- [ ] Issue could be reproduced in multiple rounds
- [x] Issue happened in a recent (less than 7 days ago) round
- [x] [Couldn't find an existing issue about this](https://github.com/NebulaSS13/Nebula/issues)

out-of-phaze:
I wonder if this happened because of the overflow in movement code? that wouldn't prevent other non-movement interactions though... would be real useful to have a runtime log

MistakeNot4892:
Honestly I think this is going to be due to transitioning across the level border. It might be possible to use the dist stuff from the town bell PR to make a z-aware flee distance proc but we'll need something like a z-aware viewers proc too.

out-of-phaze:
> It forgot to be frightened of me after I chased it across the river 

isn't this because they run to the level edge in the direction away from their flee target, in order to flee?

[...]

PR #4633, opened 2024-12-09
Adding z-aware distance procs for better AI behavior.

MistakeNot4892:
This is basically just a set of notes regarding #4073.

- [X] Add procs to AI.
- [ ] Add procs to automove.
- [ ] Test.


MistakeNot4892:
Not going to get to this for awhile.

PR #4634, opened 2024-12-10
Add wattle, daubed, plastered, and timber-framed walls

noelle-lavenza:
Adds wattle, daubed, plastered, and timber-framed walls. Sprites by Doe.
Fixes strings that misspell `separate` as `seperate`.
Updates the cabinet sprites since I forgot to commit that at some point.

There's currently no way to plaster walls. I should probably also make it so you can't plaster (or recolor) them using a paint sprayer? I also haven't tested the daubing part. And I haven't added a crafting recipe for wattle walls. Uhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhh maybe I should draft this PR


code/game/turfs/walls/wall_wattle.dm
```diff
@@ -0,0 +1,155 @@
...
+
+/turf/wall/wattle/Initialize(ml, materialtype, rmaterialtype)
+	if(ispath(daubing_material))
+		daubing_material = GET_DECL(daubing_material)
+	return ..()
+
+// Daubing with clay or soil
+/turf/wall/wattle/attackby(obj/item/W, mob/user, click_params)
+	if(isnull(daubing_material))
+		var/static/list/daub_materials = list( // Does not include subtypes.
```
MistakeNot4892, line 26: Dirt material could be used to represent adobe?

MistakeNot4892: Nvm that's soil.

code/unit_tests/turf_icons.dm
```diff
@@ -140,6 +140,9 @@
 /turf/wall/natural/get_turf_validation_corner_states()
 	return list("", "shine")
 
+/turf/wall/log/get_turf_validation_corner_states()
```
MistakeNot4892, line 143: Does `brick` need one too?

noelle-lavenza: No LOL it didn't fail CI, they have paint/etc states

code/game/turfs/walls/wall_icon.dm
```diff
@@ -54,6 +58,14 @@
 /turf/wall/proc/apply_reinf_overlay()
 	. = istype(reinf_material)
 
+/// Gets the base wall colour for icon rendering. Can be overridden on wall subtypes. Not equivalent to get_color().
+/turf/wall/proc/get_base_color()
+	return material.color
```
MistakeNot4892, line 64: Might need an istype() check on material depending on how early you can expect this to get called.

code/game/turfs/walls/wall_icon.dm
```diff
@@ -54,6 +58,14 @@
...
+
+/// Gets the reinforcement colour. Can be overridden so that some wall types don't apply paint colour to their reinforcements.
+/turf/wall/proc/get_reinf_color()
+	return paint_color ? paint_color : reinf_material?.color
```
MistakeNot4892: `paint_color || reinf_material?.color` ?

code/game/turfs/walls/wall_icon.dm
MistakeNot4892: Looks like it's in icon code so would be unsurprised if this runtimes due to material being a type on some neighbor that gets poked during turf init.

code/game/turfs/walls/wall_icon.dm
```diff
@@ -124,17 +136,17 @@
 
 	if(apply_reinf_overlay())
```
MistakeNot4892, line 138: Null check on `reinf_icon` here possibly?

MistakeNot4892: Or in `apply_reinf_overlay()`

code/unit_tests/turf_icons.dm
```diff
@@ -127,15 +127,22 @@
...
-	if(paint_color)
+	. = list("")
+	if(!material)
+		CRASH("[type] lacks a material!")
+	if(material?.wall_flags & WALL_HAS_EDGES)
+		. |= "other"
+	if(paint_color || material?.wall_flags & PAINT_PAINTABLE)
 		. |= "paint"
-	if(stripe_color)
+	if(stripe_color || material?.wall_flags & PAINT_STRIPABLE)
```
MistakeNot4892: Parens

code/game/turfs/walls/wall_icon.dm
noelle-lavenza:
```
	if(!istype(material))
		return

	refresh_connections()
	update_wall_icon()
```
update_wall_icon never runs if material is not a datum

code/game/turfs/walls/wall_icon.dm
noelle-lavenza: reinf_icon must always be set if reinf_material is set and use_reinf_state is also set—actually i guess the material could just not have one.

code/game/turfs/walls/wall_icon.dm
noelle-lavenza: i added a comment to make that clear

noelle-lavenza:
> There's currently no way to plaster walls. I should probably also make it so you can't plaster (or recolor) them using a paint sprayer? I also haven't tested the daubing part. And I haven't added a crafting recipe for wattle walls. Uhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhh maybe I should draft this PR

i am going to say:
- the paint sprayer can do whatever for now, i'm not up for making a generalisable solution to that right now. it's an edge case of an edge case
- daubing works hooray
- i am honestly okay with it being mapping only for now but could also just make it crafted with wood rods

basically: ready for review

code/game/turfs/walls/wall_icon.dm
MistakeNot4892: That is my concern, yeah.

MistakeNot4892: Oh you added it to the proc, easy.

MistakeNot4892 (approved):
I think the implementation needs some refinement but we're going to need better systems for variable turfs like this to make that worthwhile. Fine with this going in as-is.

PR #4636, opened 2024-12-11
Remove unnecessary uses of usr

out-of-phaze:
Removes a lot of unnecessary uses of `usr` (around 500, by my count). Using `usr` when `user` exists can result in funny bugs like "an admin's character is inexplicably electrocuted to death because they triggered the random camera damage event", and more mundane bugs like stuff just plain not working right. Mostly this is a code quality thing rather than bugfixes, though, so to dev it goes.

Completely untested as of writing this.

### DIFF
--- a/code/_onclick/ai.dm
+++ b/code/_onclick/ai.dm
@@ -60,7 +60,7 @@
 
 	if(silicon_camera.in_camera_mode)
 		silicon_camera.camera_mode_off()
-		silicon_camera.captureimage(A, usr)
+		silicon_camera.captureimage(A, src)
 		return
 
 	A.add_hiddenprint(src)
@@ -118,10 +118,10 @@
 	I have no idea why it was in atoms.dm instead of respective files.
 */
 
-/atom/proc/AICtrlAltClick()
+/atom/proc/AICtrlAltClick(mob/living/silicon/user)
 
-/obj/machinery/door/airlock/AICtrlAltClick() // Electrifies doors.
-	if(usr.incapacitated())
+/obj/machinery/door/airlock/AICtrlAltClick(mob/living/silicon/user) // Electrifies doors.
+	if(user.incapacitated())
 		return
 	if(!electrified_until)
 		// permanent shock
@@ -131,14 +131,14 @@
 		Topic(src, list("command"="electrify_permanently", "activate" = "0"))
 	return 1
 
-/atom/proc/AICtrlShiftClick()
+/atom/proc/AICtrlShiftClick(mob/living/silicon/user)
 	return
 
-/atom/proc/AIShiftClick()
+/atom/proc/AIShiftClick(mob/living/silicon/user)
 	return
 
-/obj/machinery/door/airlock/AIShiftClick()  // Opens and closes doors!
-	if(usr.incapacitated())
+/obj/machinery/door/airlock/AIShiftClick(mob/living/silicon/user)  // Opens and closes doors!
+	if(user.incapacitated())
 		return
 	if(density)
 		Topic(src, list("command"="open", "activate" = "1"))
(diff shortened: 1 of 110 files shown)


code/_onclick/ai.dm
```diff
@@ -131,62 +131,62 @@
 		Topic(src, list("command"="electrify_permanently", "activate" = "0"))
 	return 1
 
-/atom/proc/AICtrlShiftClick()
+/atom/proc/AICtrlShiftClick(mob/living/silicon/robot/user)
```
MistakeNot4892: Shouldn't this be /silicon rather than /silicon/robot?

out-of-phaze: yep

code/modules/mob/mob.dm
```diff
@@ -561,10 +561,10 @@
 		update_flavor_text(href_list["flavor_change"])
 		return TOPIC_HANDLED
 
-// If usr != src, or if usr == src but the Topic call was not resolved, this is called next.
 /mob/proc/get_comments_record()
 	return
 
+// If usr != src, or if usr == src but the Topic call was not resolved, this is called next.
 /mob/OnTopic(mob/user, href_list, datum/topic_state/state)
```
MistakeNot4892, line 568: Idle thought but we should probably put an `exactly` proc in the test scripts to prevent adding more Topic() overrides carelessly.

code/modules/power/singularity/emitter.dm
```diff
@@ -77,8 +77,8 @@
 				to_chat(user, "You turn on \the [src].")
 				shot_number = 0
 				fire_delay = get_initial_fire_delay()
-				log_and_message_admins("turned on \the [src]")
-				investigate_log("turned <font color='green'>on</font> by [key_name_admin(user || usr)]","singulo")
+				log_and_message_admins("turned on \the [src]", user)
```
MistakeNot4892, line 80: I think the purpose of this was due to setup procs/events potentially toggling emitters, but I don't know if that's still a thing on our codebase.

mods/content/matchmaking/matchmaker.dm
```diff
@@ -271,8 +271,8 @@ var/global/datum/matchmaker/matchmaker = new()
 			var/list/relations = matchmaker.get_relationships(mind)
 			for(var/datum/relation/R in relations)
 				R.finalize()
-			show_browser(usr,null, "window=relations")
+			show_browser(src,null, "window=relations")
```
MistakeNot4892, line 274: Will this interfere with admin messing with matchmaking stuff? Guessing not since it's a self call.

MistakeNot4892 (dismissed):

code/modules/mob/mob.dm
out-of-phaze: i think Topic is still preferred for some cases? i think OnTopic doesn't always call parent but Topic must or something

code/modules/power/singularity/emitter.dm
out-of-phaze:
```
/proc/log_and_message_admins(var/message as text, var/mob/user = usr, var/turf/location)
	var/turf/T = location ? location : (user ? get_turf(user) : null)
	message = append_admin_tools(message, user, T)

	log_admin(user ? "[key_name(user)] [message]" : "EVENT [message]")
	message_admins(user ? "[key_name_admin(user)] [message]" : "EVENT [message]")
```
it is, that happens specifically if user is null. if user is null it'll still happen, but i guess the issue is specifically if someone triggers something such that `activate()` runs without a `user` but it still has a `usr`? i can restore it if you want but i think if the passed user param is null that `var/mob/user = usr` bit might make it use usr for the arg anyway, not sure

mods/content/matchmaking/matchmaker.dm
out-of-phaze: oh good catch, yeah since it's OnSelfTopic this is always safe, but that's not necessarily true in other cases. i'll make sure to keep that in mind when rewriting Topic stuff to use OnTopic later

PR #4637, opened 2024-12-11
Meat and nutriment should no longer cause organ damage on ingestion.

### LINKED ISSUE #4588: Eating food causes liver damage
#### Description of issue

Eating food seems to cause minor liver damage. 

No liver damage is expected. Liver damage right now.

#### Steps to reproduce
Eat food
I specifically have gotten it as neo avian with cybernetic organs eating an energy bar twice, and once other with chocolate cake.

Liver
Energy Bar

Since the update that altered a lot of sprites a few weeks ago, sooo 11/1/24 or so.

Client Version: 515
Server Revision: db8bd2dc81b2282378fb43076f5089a91f23cfe7 - dev - 2024-11-29
Game ID: cwV-dMe8

#### Issue bingo

MistakeNot4892:
Fixes #4588

### DIFF
--- a/code/modules/reagents/chems/chems_drinks.dm
+++ b/code/modules/reagents/chems/chems_drinks.dm
@@ -9,6 +9,7 @@
 	nutriment_factor = 0
 	hydration_factor = 6
 	affect_blood_on_ingest = FALSE
+	affect_blood_on_inhale = FALSE
 
 	var/adj_dizzy = 0 // Per tick
 	var/adj_drowsy = 0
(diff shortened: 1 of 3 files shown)


code/modules/reagents/chems/chems_drinks.dm
```diff
@@ -9,6 +9,7 @@
 	nutriment_factor = 0
 	hydration_factor = 6
 	affect_blood_on_ingest = FALSE
+	affect_blood_on_inhale = FALSE
```
out-of-phaze, line 12: what I did for ethanol was make affect_inhale or w/e call affect_ingest with a multiplier, so that inhaling alcohol would get you drunk rather than kill you from injecting alcohol directly into your bloodstream

out-of-phaze (approved):

PR #4639, opened 2024-12-12
Infrastructure work for shapeshifting fixes/rewrite.

MistakeNot4892:
## Description of changes
- Removes legacy shapeshifting species/bodytype, they are extremely broken due to bodytypes not existing when they were written.
- Cleans up/privates equipment adjustment lists.
- Allows organs to set an `appearance_bodytype` that overrides regular bodytype for accessories and icon gen.
- Allows bodytypes to provide a different bodytype when asked for equipment offsets.

Tested with a debug verb that I have removed for this PR.

Supporting work for completely rewriting shapeshifters in view of how badly they are broken currently.

## Authorship
Myself.

## Changelog
Nothing player-facing.


code/modules/bodytype/bodytype_offsets.dm
```diff
@@ -18,15 +18,19 @@ The slots that you can use are found in items_clothing.dm and are the inventory
 */
```
out-of-phaze, line 18: might be nice to note you can use `BP_R_HAND` etc for held items as well? i think?
also maybe would be good to explain why we need `(foo)` as the key. todo `FOOTGUNS.MD` i guess

out-of-phaze (approved):

PR #4641, opened 2024-12-13
Relocating/cleaning up some bodytype logic.

MistakeNot4892:
- Moves bodytype stuff into a `modules` folder.
- Removes an unused eye offset and the obviated icon_uid (uses decl uid instead).
- Comments/formats the main bodytype file.
- Privates the bodytype organ category and tag lists.
- Moves surgery overlay icons onto bodytype.
- Probably some other stuff I missed.


out-of-phaze:
would like the eye offsets back, i use them for replikas on a downstream

MistakeNot4892:
> would like the eye offsets back, i use them for replikas on a downstream

Restored, added a comment so I don't do it again sometime.

code/modules/bodytype/_bodytype.dm
```diff
@@ -372,15 +411,19 @@ var/global/list/bodytypes_by_category = list()
...
+			LAZYINITLIST(_organ_tags_by_category)
+			LAZYADD(_organ_tags_by_category[organ_cat], organ_tag)
 
 	for(var/organ_tag in has_organ)
 		var/obj/item/organ/organ = has_organ[organ_tag]
 		var/organ_cat = initial(organ.organ_category)
 		if(organ_cat)
-			LAZYADD(organs_by_category[organ_cat], organ)
-			LAZYADD(organ_tags_by_category[organ_cat], organ_tag)
+			LAZYINITLIST(_organs_by_category)
```
MistakeNot4892, line 423: I've just realised these weren't lazylisted because they're on bodytypes which are singletons. I'm not bright.

out-of-phaze: some decls are lazy instantiated but bodytypes are not one

out-of-phaze (approved):
sorry, thought you were going to change the lazylist stuff so i waited to make sure the approval wouldn't be invalidated

PR #4643, opened 2024-12-15
Tweaking biogenerator/seed extractor attackby().

### LINKED ISSUE #4642: biogenerator does not accept plants at the moment
#### Description of issue
normally you can feed plants into the biogenerator but not on current version, not even replacing the bottle inside it with a actual beaker helps.

I should be able to put produce into the biogen to make the bio mass

#### Steps to reproduce
get fruit from any plant (besides towercap) and attempt to feed it into the machine 

biogenerator

since 12/13/2024

Client Version: 515
Server Revision: eca6451d87eb1c495f8b8397ab9ffc7a7459bdb1 - dev - 2024-12-14
Game ID: cxa-cTbm

#### Issue bingo

MistakeNot4892:
Fixes #4642


out-of-phaze (approved):
here's hoping we never add a screwdriver tree that grows screwdriver fruits or else we'll have the vendor screwdriver issue all over again :p (dynamic intents when?)

PR #4644, opened 2024-12-15
Atom fires attempt #2

MistakeNot4892:
## Description of changes
- Moves fire behavior onto /atom.
- Makes fires consume atoms instead of being turf-only.

This is an old PR that I have rebased, it is very WIP.

Better fire behavior.

## Authorship
Myself.

## Changelog
TODO


MistakeNot4892:
Will come back to this after my break (per Discord)

PR #4645, opened 2024-12-15
Stubs and getters/setters for atom fires.

MistakeNot4892:
Procs/varname changes to support #4644


code/game/machinery/igniter.dm
```diff
@@ -51,7 +51,7 @@
 			location.hotspot_expose(1000,500,1)
 	return 1
 
-/obj/machinery/igniter/proc/ignite()
+/obj/machinery/igniter/proc/create_sparks()
```
out-of-phaze: this doesn't actually produce sparks at all

code/game/objects/items/flame/_flame.dm
```diff
@@ -216,17 +216,14 @@
...
-		loc.update_icon()
-
-	// TODO: generalized ignition proc
-	if(isliving(loc))
-		var/mob/living/M = loc
-		M.IgniteMob()
+	if(loc)
+		if(istype(loc, /obj/structure/wall_sconce))
+			loc.update_icon()
+		loc.ignite_fire()
```
out-of-phaze, line 220: could put an icon update in sconces' ignite_fire?

code/game/objects/items/flame/_flame.dm
```diff
@@ -115,7 +115,7 @@
 
 	return ..()
 
-/obj/item/flame/proc/extinguish(var/mob/user, var/no_message)
+/obj/item/flame/extinguish_fire(mob/user, no_message = FALSE)
```
out-of-phaze: not sure how to feel about this. i feel like a lighter should be able to catch fire separately from it being on/off or lit/unlit? idk

out-of-phaze (dismissed):
i feel like "does this have the ability to be lit, a la lighters" and "is the material this is made of *on fire?*" should be separate, conceptually, but I'm not hard tied to that

code/game/machinery/igniter.dm
MistakeNot4892: Oh true, I assumed. Will rename it to `toggle_igniter` or something.

code/game/objects/items/flame/_flame.dm
MistakeNot4892: This actually might cause a recursion issue if the sconce then tries to light the torch again.

code/game/objects/items/flame/_flame.dm
MistakeNot4892: I wanted to condense all the procs but I was iffy on this too, I'll give it its own proc.

code/game/objects/items/flame/_flame.dm
MistakeNot4892: Looks like sconces don't try to light their contents via that proc so should be fine.

out-of-phaze (approved):

PR #4647, opened 2024-12-17
Misc Scav feedback fixes.

MistakeNot4892:
- Lowers default click cooldown (items/attacks already set their own, so this largely impacts non-harm-intent interactions).
- Fixes autobinders not taking paper along with several other fab intake issues.
- Fixes https://github.com/ScavStation/ScavStation/issues/1148


out-of-phaze (approved):
I wonder if this will stop tiles with mapped/random dirt from persisting dirt until they're cleaned and get dirty again, even if it goes from the default to max dirt.

MistakeNot4892:
Mapped decals should take precedence over persistent ones anyway, so not sure.

PR #4651, opened 2024-12-18
[MDB IGNORE] Struts are now largely replaced by rods.

MistakeNot4892:
- Recipes that used to use struts now use rods.
- Rods now use the material stack system and crafting system properly.
- Grilles are a crafting recipe that uses rods.
- Maps have been updated with mapped subtypes of rods.

Struts were an interesting idea but they are really just a kind of painful bit of unnecessary complexity a lot of the time.


code/game/objects/structures/girders.dm
```diff
@@ -30,9 +30,9 @@
 
 	if(reinf_material)
 		playsound(src.loc, 'sound/items/Screwdriver.ogg', 100, 1)
-		visible_message(SPAN_NOTICE("\The [user] begins unscrewing \the [reinf_material.solid_name] struts from \the [src]."))
```
MistakeNot4892, line 33: I know these aren't the same kind of struts as the rest of the PR but it has bothered me for a bit.

code/game/objects/structures/girders.dm
```diff
@@ -30,9 +30,9 @@
...
+		visible_message(SPAN_NOTICE("\The [user] begins unscrewing \the [reinf_material.solid_name] rods from \the [src]."))
```
out-of-phaze, line 33: while we're at it these might need to be `reinf_material.adjective_name` so that you'd get e.g. "oaken rods" rather than "oak rods". it's not a huge deal though because usually both are valid, but i like using the adjective form where possible

edit: this is an issue in more places than just here, i will do a pass on it at some point

out-of-phaze (approved):

PR #4660, opened 2024-12-20
Barrel/well interaction rework.

MistakeNot4892:
## Description of changes
_Really_ not sure about the implementation of this one. Going to pull it into Pyrelight volatile for testing over the weekend.

- Adds `get_standard_interactions()` and calls it similarly to alt interactions at the root of `attackby()` and `attack_hand()`.
- Moves dipping, filling and emptying into interaction handlers provided as needed by the alt interaction proc.
- Barrels and wells also provide these interaction handlers to their standard interaction procs. Basically: you can pick what interaction you want if you have a bucket to fill or empty, or a wrench to unanchor with or put into storage.
- Generalized several structure procs to allow for tool interactions in a similar vein.
- Let barrels be anchored/unanchored.

We are at a point where the UX for some structures is getting stupid. Barrels and wells are the most obvious/severe example: if you have a half full bucket, and a half full well, and you click the bucket on the well, the interaction has to be mediated by flag checks and a toggle on the well atom, or workarounds like drag-drop. 

This puts support in place for similar interactions in the future, and solves the immediate barrel/well issues.

## Authorship
Myself.

## Changelog
:cl:
tweak: The way you interact with barrels and well has been significantly reworked; clicking with a bucket or tool should give a list of options to pick from. Please report bugs with this on the tracker.
/:cl:


MistakeNot4892:
Something has broken in the fill/pour interactions and now they aren't showing up on barrels. Augh. Issue for tomorrow.

MistakeNot4892:
https://www.dropbox.com/scl/fi/9buseoj32imp3nh1x7uxq/tvylQ8LZPw.mp4?rlkey=hfitnnb4t2j4ko854vf2g7alx&st=q1i0uxe9&dl=0

MistakeNot4892:
DNM, need to check if this has clobbered hammer deconstruction.

MistakeNot4892:
Ready to go.

code/modules/interactions/interactions_atom.dm
```diff
@@ -18,9 +18,20 @@
...
-		if(!istype(choice) || QDELETED(user) || !(choice.type in get_alt_interactions(user)) || !choice.is_possible(src, user, prop))
+		if(!istype(choice) || QDELETED(user) || QDELETED(src))
+			return TRUE
+		// This is not ideal but I don't want to pass a callback through here as a param and call it. :(
+		if(check_alt_interactions)
+			if(!(choice.type in get_alt_interactions(user)))
+				return TRUE
+		else
+			if(!(choice.type in get_standard_interactions(user)))
+				return TRUE
```
out-of-phaze: maybe `var/list/new_interactions = check_alt_interactions ? get_alt_interactions(user) : get_standard_interactions(user)`? that way you would only need one `!(choice.type in new_interactions)` check. it wouldn't fix the coupling issue (it's a niche control flag) but it'd make the code marginally nicer
```suggestion
		// This is not ideal but I don't want to pass a callback through here as a param and call it. :(
		var/list/new_interactions = check_alt_interactions ? get_alt_interactions(user) : get_standard_interactions(user)
		if(!(choice.type in new_interactions))
				return TRUE
```
that said, it should probably use a callback, because someone might accidentally call it with standard interactions but use the alt interactions flag (or the opposite). but. meh. i might do it at some point

code/modules/hydroponics/trays/tray_soil.dm
```diff
@@ -19,11 +19,6 @@
 
 	var/obj/item/stack/material/brick/reinforced_with
 
-/obj/machinery/portable_atmospherics/hydroponics/soil/get_alt_interactions(var/mob/user)
-	. = ..()
-	. -= /decl/interaction_handler/drink
```
out-of-phaze, line 24: was this moved? if so i can't find it, but you definitely shouldn't be able to wash your hands in or drink from a plot of soil. it's not got standing water, it's absorbed into the soil

out-of-phaze (changes requested):

code/modules/interactions/interactions_atom.dm
out-of-phaze: this was not meant to be blocking, i was gonna approve until i noticed the soil plot thing

code/modules/hydroponics/trays/tray_soil.dm
MistakeNot4892: I think I removed it thinking it was an addition. Will revert.

out-of-phaze: still has wash hands. for that matter uhhh, i think all the reagent interactions except `/decl/interaction_handler/empty_into` should be removed on this. that or `can_be_poured_from` should return FALSE on soil trays, which should lead to uh. dip into and fill from being disabled. maybe wash hands should also check that and then we could pretty tidily solve it without needing a get_alt_interactions override at all?

out-of-phaze (changes requested):

out-of-phaze (approved):
works for me

PR #4662, opened 2024-12-21
Staging fixes from Shaded Hills scav feedback.

### LINKED ISSUE #4656: Crucibles made from sand or dirt are watertight.
Current dev branch, reported on Scav. Free 5000u reagent storage made of dirt is probably not desirable.

### LINKED ISSUE #4657: Barrels cannot be unanchored or moved.
Title, noting for my reference.

### LINKED ISSUE #4658: Medieval map needs additional medical reagents.
- Antiseptic (suggest putting STERILE on ethanol and overriding on subtypes)
- Antibiotics (suggest garlic juice/garlic oil)

### LINKED ISSUE #4659: Hammer should be able to be used for dismantling wooden furniture or removing wooden/cobble floor tiles.
Currently you have to use a knife to remove flooring on the fantasy map and cannot remove cobblestones at all. Also cannot dismantle furniture short of smashing it to bits.

MistakeNot4892:
- Fixes #4656
- Fixes #4657
- Fixes #4658
- Fixes #4659
- Fixes https://github.com/ScavStation/ScavStation/issues/1154
- Fixes https://github.com/ScavStation/ScavStation/issues/1149
- Converts ethanol to an abstract type and adds a /spirits subtype for pure ethanol.
- Prevents ethanol from boiling as a short-term fix for the annoying nature of refining ethanol.
- Prevents bowls from being lidded.
- Moves spoon interactions to /glass to allow most vessels to use it.
- Prevents reinforced material being dropped from structures as planks/struts.
- Fixed some issues preventing structures from dumping their contents into loc on physical destruction.
- Cherrypicks some barrel/well interactions from dev due to issues with them on staging.
- Repaths waterskins to /glass to allow pouring from them into cups.


code/modules/reagents/reagent_containers/food/lunch.dm
```diff
@@ -107,11 +107,11 @@ var/global/list/lunchables_ethanol_reagents_ = list(
 		.[initial(O.name)] = lunch
 	return sortTim(., /proc/cmp_text_asc)
 
-/proc/init_lunchable_reagent_list(var/list/banned_reagents, var/reagent_types)
+/proc/init_lunchable_reagent_list(var/list/banned_reagents, var/reagent_type)
 	. = list()
-	for(var/reagent_type in subtypesof(reagent_types))
-		if(reagent_type in banned_reagents)
+	for(var/reagent_subtype in decls_repository.get_decls_of_subtype(reagent_type))
```
out-of-phaze: the `get_decls_of_subtype` changes for ethanol aren't actually necessary because ethanol is an abstract type anyway. in fact this might cause issues if you pass in a non-abstract type to this proc ever. technically it's not a big deal but eh

code/modules/reagents/reagent_containers/_glass.dm
```diff
@@ -19,7 +19,8 @@
 	pickup_sound = 'sound/foley/bottlepickup1.ogg'
 	watertight = FALSE // /glass uses the open container flag for this
 
-	var/list/can_be_placed_into = list(
+/obj/item/chems/glass/proc/get_atoms_can_be_placed_into()
```
out-of-phaze, line 22: not super sure why this was necessary? it complicates overriding it and runs into the potential of overrides accidentally causing list mutations affecting all procs if they do `. = ..(); . -= /obj/machinery/whatever` to remove something from a subtype. obviously an edge case but one that's annoying enough i'm a little worried

code/modules/reagents/chems/chems_ethanol.dm
```diff
@@ -99,6 +89,22 @@
 	if(halluci)
 		M.adjust_hallucination(halluci, halluci)
 
+// Somewhat a dummy type for 'pure ethanol' to avoid having to set dirtiness/heating products/etc on literally everything else.
+/decl/material/liquid/ethanol/spirits
+	dirtiness         = DIRTINESS_STERILE
+	uid               = "chem_ethanol"
+
+	// Uncomment when refining spirits is less annoying
```
out-of-phaze: maybe note that it's specifically about draught/heating controls? currently they aren't finely-grained enough to allow a useful degree of control. we might look into a stopgap based on autogenerating the levels from burn milestone materials in the meantime but i think that failed to work earlier for some reason

code/modules/reagents/chems/chems_ethanol.dm
```diff
@@ -99,6 +89,22 @@
 	if(halluci)
 		M.adjust_hallucination(halluci, halluci)
 
+// Somewhat a dummy type for 'pure ethanol' to avoid having to set dirtiness/heating products/etc on literally everything else.
+/decl/material/liquid/ethanol/spirits
```
out-of-phaze: not super sure about calling it 'spirits' generally. aurora has iirc `/decl/reagent/alcohol/ethanol` which could be better, since it'd open the door for having e.g. rubbing alcohol etc. in the future, or alien alcohols that don't affect humans/aliens the same, etc.? not a big deal though

code/modules/reagents/chems/chems_ethanol.dm
```diff
@@ -7,46 +8,35 @@
...
-		/decl/material/liquid/ethanol = 0.75,
-		/decl/material/solid/ice =      0.25
+		/decl/material/liquid/ethanol/spirits = 0.75,
+		/decl/material/solid/ice              = 0.25
 	)
-	bypass_chilling_products_for_root_type = /decl/material/liquid/ethanol
 	affect_blood_on_ingest = FALSE // prevents automatic toxins/inebriation as though injected
 	affect_blood_on_inhale = FALSE
-	can_boil_to_gas = TRUE
-	temperature_burn_milestone_material = /decl/material/liquid/ethanol
```
out-of-phaze, line 32: this reminded me that because wood is an abstract material now, this doesn't show wood burning anymore i think
```dm
		for(var/decl/material/material as anything in decls_repository.get_decls_of_subtype_unassociated(/decl/material))
			if(material.type != material.temperature_burn_milestone_material)
				continue
```
get_decls_of_subtype_unassociated can never return abstract decls unless they have the abstract decl init flag, which materials don't. so there will never be an instantiated type where `material.type != material.temperature_burn_milestone_material` succeeds for like. wood. this in fact was done explicitly because alcohol was not an abstract reagent and i think switching it back to true/false would be better (although we probably want to deduplicate it somehow anyway to avoid having burning points listed for every single type of wood?)

code/modules/reagents/chems/chems_drinks.dm
```diff
@@ -178,15 +178,16 @@
 	allergen_flags = ALLERGEN_VEGETABLE
 
 /decl/material/liquid/drink/juice/garlic
-	name = "garlic juice"
-	lore_text = "Who would even drink this?"
+	name = "garlic oil"
```
out-of-phaze, line 181: clearly we need `lipophilicity` as a material var to generalize oils in the code (joke)

out-of-phaze (changes requested):
the comment about `temperature_burn_milestone_material` is the only change that imo is non-negotiable since it's a bugfix. everything else is either jokes or minor nitpicks or suggestions

code/modules/reagents/reagent_containers/_glass.dm
MistakeNot4892: Mostly just hundreds of items having a twenty item list that is never changed but can't be made static due to the rag subtype.

code/modules/reagents/chems/chems_ethanol.dm
MistakeNot4892: Repathing it to alcohol will be a much bigger changeset. Would prefer to do that on dev since it's purely organisational.

MistakeNot4892: Went ahead and did it anyway.

code/modules/reagents/chems/chems_ethanol.dm
out-of-phaze: i was wrong, it was moved to oak LMFAO

MistakeNot4892: That's an issue on dev, not staging, so can't fix it in this PR. Also yeah it was for both ethanol and wood as they have 200 billion subtypes each.

out-of-phaze (dismissed):

out-of-phaze (approved):

PR #4664, opened 2024-12-21
Add fluid heights to slots, fix overzealous dousing

out-of-phaze:
added heights to slots, currently unscaled but could easily be affected by mob height and size modifiers. unfortunately it'd need a rewrite to properly handle like... a head below your hands, or whatever. would probably want to make that organ-driven somehow.
expands submerged checks to add negative turf height when floating or being thrown. this means you can leap over deep water or throw a lantern across water.
makes all fluid act sources from movement check submersion
basically you might get your shoes wet in a puddle but you won't have your lantern immediately doused.

this is moderately tested but should be pulled into volatile tbh

### DIFF
--- a/code/datums/inventory_slots/_inventory_slot.dm
+++ b/code/datums/inventory_slots/_inventory_slot.dm
@@ -20,7 +20,7 @@
 	var/requires_organ_tag
 	var/quick_equip_priority = 0 // Higher priority means it will be checked first. If null, will not be considered for quick equip.
 	/// What depth of fluid is necessary for an item in this slot to be considered submerged?
-	var/fluid_height = FLUID_SHALLOW
+	var/fluid_height = FLUID_SHALLOW // we're treating FLUID_SHALLOW as waist level, basically
 
 	var/mob_overlay_layer
 	var/alt_mob_overlay_layer
--- a/code/datums/inventory_slots/_inventory_slot.dm
+++ b/code/datums/inventory_slots/_inventory_slot.dm
@@ -19,6 +19,8 @@
 	var/requires_slot_flags
 	var/requires_organ_tag
 	var/quick_equip_priority = 0 // Higher priority means it will be checked first. If null, will not be considered for quick equip.
+	/// What depth of fluid is necessary for an item in this slot to be considered submerged?
+	var/fluid_height = FLUID_SHALLOW
 
 	var/mob_overlay_layer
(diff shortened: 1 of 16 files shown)


code/datums/inventory_slots/_inventory_slot.dm
```diff
@@ -19,6 +19,8 @@
 	var/requires_slot_flags
 	var/requires_organ_tag
 	var/quick_equip_priority = 0 // Higher priority means it will be checked first. If null, will not be considered for quick equip.
+	/// What depth of fluid is necessary for an item in this slot to be considered submerged?
+	var/fluid_height = FLUID_SHALLOW // we're treating FLUID_SHALLOW as waist level, basically
```
MistakeNot4892, line 23: Nit: var name isn't very informative.

code/game/atoms_fluids.dm
```diff
@@ -23,19 +23,64 @@
...
+		var/turf_height = T.get_physical_height()
+		// If we're not on the surface of the turf (floating, leaping, or other sources)
+		// then we add the turf height to the depth, so you can jump over a water-filled pit
+		// or throw something over it
+		if(turf_height < 0 && above_turf)
+			depth += turf_height
 	if(ismob(loc))
 		return depth >= FLUID_SHALLOW
 	if(isturf(loc))
+		if(locate(/obj/structure/table))
```
MistakeNot4892, line 42: Possibly out of scope for this specifically, but this should probably use the height offset value that structure implements for standing on tables rather than just checking for tables.

out-of-phaze: that's applied for mobs in the mob override below. i'm not suuuper sure about applying it to non-mobs on the table...

out-of-phaze: also isn't mob_offset only on table? if not then I was mistaken and can change it

MistakeNot4892: `mob_offset` is just a vertical pixel value on /structure, you should be able to use it to offset more or less anything. Items use the center_of_mass/auto_align stuff though. I mentioned this being out of scope because you would need some way to map pixel height to fluid depth which might not even be a sane way to approach it.

MistakeNot4892 (approved):
This is probably fine as-is though I wouldn't mind fluid_height being renamed to `submerge_depth` or something.

PR #4665, opened 2024-12-21
Fix misc. issues with fluid appearances

out-of-phaze:
- Fixes an issue with fluids on recessed natural turfs flickering above/below mobs.
- Fixes torch head colors breaking when an alpha mask is applied.
- Fixes KEEP_TOGETHER never being unapplied when an alpha mask is removed, because remove_filter returned the wrong value.
- Adds held and worn icon update calls to torches when they're extinguished.
- Fixes ticks2readable giving a blank output for values under 10 ticks (relevant for ability cooldown etc).

### DIFF
--- a/code/game/objects/items/flame/_flame.dm
+++ b/code/game/objects/items/flame/_flame.dm
@@ -125,6 +125,9 @@
 	update_attack_force()
 
 	update_icon()
+	if(ismob(loc)) // not very robust for things like accessories...
+		update_held_icon()
+		update_clothing_icon()
 	if(istype(loc, /obj/structure/wall_sconce))
 		loc.update_icon()
 
--- a/code/game/objects/items/flame/flame_torch.dm
+++ b/code/game/objects/items/flame/flame_torch.dm
@@ -76,11 +76,11 @@
 	if(head_material)
 		var/decl/material/head_mat = GET_DECL(head_material)
 		if(burnt)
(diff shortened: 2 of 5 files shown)


code/game/objects/items/flame/_flame.dm
```diff
@@ -125,6 +125,9 @@
 	update_attack_force()
 
 	update_icon()
+	if(ismob(loc)) // not very robust for things like accessories...
```
MistakeNot4892, line 128: get_recursive_loc_of_type() or whatever that proc is called might be good here?

code/game/objects/items/flame/flame_torch.dm
```diff
@@ -76,11 +76,11 @@
 	if(head_material)
 		var/decl/material/head_mat = GET_DECL(head_material)
 		if(burnt)
-			add_overlay(overlay_image(icon, "[icon_state]-burnt", head_mat.color, flags = RESET_COLOR))
+			add_overlay(overlay_image(icon, "[icon_state]-burnt", head_mat.color, flags = RESET_COLOR|KEEP_APART))
```
MistakeNot4892, line 79: Does this work as expected? I had a lot of trouble with overlays like this on Polaris and I remember this not working.

MistakeNot4892 (approved):
Curious about the KEEP_APART comment but not really blocking.

code/game/objects/items/flame/flame_torch.dm
out-of-phaze: it worked fine in my testing

PR #4667, opened 2024-12-22
Drake fixes.

MistakeNot4892:
- Buffs drake unarmed damage, as they cannot use held weapons at all.
- Adds native amor to drakes, similar to above - they cannot equip armor.
- Fixes native armor applying to entire mob via species, now applies per organ via bodytype.
- Adds manual cooldowns to unarmed attacks to compensate for the default quick cooldown being too short.
- Fixes damage overlays not respecting stance modifier.
- Adds base damage overlay textures for drakes.


out-of-phaze (commented):
salt pr because loaf got their ass kicked by an npc last karzerfeste test (joke)

also drake natural armor is not in the pr description and is also not really a bugfix? that feels more like a dev branch thing than a staging one. up to you though

MistakeNot4892:
Unrelated to me getting got, it was something pointed out on Scav. Drakes can't wear armour at all, it's as much a fix for an oversight as half this PR :P

MistakeNot4892:
Updated body.

out-of-phaze (approved):
first Pyrelight salt pr let's gooooo

MistakeNot4892:
If it was a salt PR I would have disabled AI punches from stunning you...

PR #4668, opened 2024-12-22
Drakes can dig.

MistakeNot4892:
## Description of changes
Digging handlers and predator melee invocation will check for a hand organ and use it as a shovel if nothing else is provided.

People want drakes to dig.

## Authorship
Myself.

## Changelog
:cl:
add: Drakes can dig holes!
/:cl:


out-of-phaze (approved):
sneaky staging pr for the scavs...

PR #4669, opened 2024-12-22
Fixing organs not burning in fires.

MistakeNot4892:
Organs will now burn to ash on a fire.

### DIFF
--- a/code/modules/materials/_material_stack.dm
+++ b/code/modules/materials/_material_stack.dm
@@ -219,3 +219,8 @@
 		. += " [singular_name]"
 		return indefinite_article ? "[indefinite_article] [.]" : ADD_ARTICLE(.)
 	return "[amount] [.] [plural_name]"
+
+// Horrible solution to heat damage for atoms causing logs and
+// fuel to vanish. Replace this when the atom fire system exists.
+/obj/item/stack/material/should_take_heat_damage(datum/gas_mixture/air, exposed_temperature)
+	return istype(loc, /obj/structure/fire_source) ? FALSE : ..()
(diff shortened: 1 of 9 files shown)


MistakeNot4892:
Temperature damage is stupid.

code/modules/materials/_material_stack.dm
```diff
@@ -219,3 +219,8 @@
 		. += " [singular_name]"
 		return indefinite_article ? "[indefinite_article] [.]" : ADD_ARTICLE(.)
 	return "[amount] [.] [plural_name]"
+
+// Horrible solution to heat damage for atoms causing logs and
+// fuel to vanish. Replace this when the atom fire system exists.
+/obj/item/stack/material/should_take_heat_damage(datum/gas_mixture/air, exposed_temperature)
+	return istype(loc, /obj/structure/fire_source) ? FALSE : ..()
```
out-of-phaze, line 226: does nothing to stop molds taking damage when you fire them in a kiln

out-of-phaze (changes requested):

code/modules/materials/_material_stack.dm
MistakeNot4892: Molds generally don't have an ignition temperature, these changes shouldn't impact them.

MistakeNot4892: Per Discord, I'm not sure what change is being requested here. I haven't adjusted the behavior for materials like clay that don't solely have an ignition point.

out-of-phaze (approved):
as i explained on discord i misunderstood

PR #4671, opened 2024-12-22
Ginseng will now restore blood.

MistakeNot4892:
Fills a need for blood loss treatment on shadyhills.

### DIFF
--- a/code/modules/reagents/chems/chems_herbal.dm
+++ b/code/modules/reagents/chems/chems_herbal.dm
@@ -15,9 +15,15 @@
 /decl/material/liquid/antitoxins/ginseng
 	name               = "powdered ginseng"
 	uid                = "chem_antitoxins_herbal"
-	lore_text          = "Ginseng root has curative properties and encourages organ recovery after poisoning."
+	lore_text          = "Ginseng root has curative properties and encourages organ recovery and restoration of blood volume after poisoning or blood loss."
 	taste_description  = "bitter herbs"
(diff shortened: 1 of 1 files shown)


code/modules/reagents/chems/chems_herbal.dm
```diff
@@ -15,10 +15,16 @@
...
-	lore_text          = "Ginseng root has curative properties and encourages organ recovery after poisoning."
+	lore_text          = "Ginseng root has curative properties and encourages organ recovery and restoration of blood volume after poisoning or blood loss."
 	taste_description  = "bitter herbs"
 	antitoxin_strength = 0.35
 
+/decl/material/liquid/antitoxins/ginseng/affect_ingest(var/mob/living/M, var/removed, var/datum/reagents/holder)
+	. = ..()
+	if(M.has_trait(/decl/trait/metabolically_inert))
+		return
+	M.add_chemical_effect(CE_BLOODRESTORE, 8 * removed)
```
out-of-phaze, line 26: this will mean ginseng tea and tincture don't apply it, no? or at least that it wouldn't be more effective

MistakeNot4892: Tea/tincture subtypes the base herb so will apply it. Do we need this effect to be stronger for tea/tinctures? I can make it a var easily enough, but it seemed fairly niche.

out-of-phaze (approved):

PR #4672, opened 2024-12-22
Adds blacksmithy

MistakeNot4892:
## Description of changes
This is a partial implementation of blacksmithy and metalworking to bracket the existing low-tech crafting systems like casting.

image

## Longer-term TODO (not in this PR)
- Update bladed weapons to be crafted like tools, by combining handle and blade blank.
- Implement quenching/hardening and grinding/polishing as final steps after producing the blanks.
- Implement barrel, chest and armour fittings.
- Implement barrel crafting.
- Implement nails/uses for nails.
- Move metal crafting recipes off stacks and onto either the autolathe, forging or both.
- Add anvil forging (ingots on an anvil instead of a billet?)
- Punch up the forging strings and messages to be less brusque.
- Find icons for armor, helmet, blade blanks.

- Implements a metalworking minigame to replace using psychic powers and spaceman bare hands to turn ingots into things.
- Clang clang clang clang.

## Authorship
Code is mine, icons and a lot of the design work belong to @PlayerDeer.

## Changelog
:cl:
feature: Added blacksmithing!
/:cl:


code/modules/crafting/forging/forging_step.dm
```diff
@@ -0,0 +1,193 @@
...
+
+/decl/forging_step/product/shield_fasteners
+	billet_name  = "shield fasteners"
+
+/decl/forging_step/product/tongs
+	billet_name  = "tongs"
+	product_type = /obj/item/tongs
+
+/decl/forging_step/product/chisel
+	billet_name  = "chisl"
```
out-of-phaze:
```suggestion
	billet_name  = "chisel"
```

out-of-phaze (changes requested):

code/game/objects/objs.dm
```diff
@@ -403,3 +403,25 @@
...
+	if(icon_temperature < scale_max)
+		h_r = 64 + (h_r - 64)*scale
+		h_g = 64 + (h_g - 64)*scale
+		h_b = 64 + (h_b - 64)*scale
+
+	var/scale_color = rgb(h_r, h_g, h_b)
+	var/list/animate_targets = get_above_oo() + src
+	for (var/thing in animate_targets)
+		var/atom/movable/AM = thing
+		animate(AM, color = scale_color, time = 2 SECONDS, easing = SINE_EASING)
```
out-of-phaze: this is why they're all grey when cold, because `heat2color_foo` gives grey at lower temperatures. need to factor in base color somewhere. maybe lerp between `get_color()` and scale_color?
that or you could maybe make the `heat2color_foo` procs defined on /obj instead of being global procs, and they could use `src.get_color()` as a basis.

out-of-phaze (changes requested):

code/modules/crafting/forging/forge_anvil.dm
```diff
@@ -0,0 +1,66 @@
...
+	// Put the hammer on the forge.
+	if(!user.check_intent(I_FLAG_HARM) && istype(used_item, /obj/item/tool/hammer/forge))
+		if(user.try_unequip(used_item, get_turf(src)))
+			auto_align(used_item, click_params)
+		return TRUE
+
+	// Put the bar onto the anvil (need to do this to avoid repairs in ..())
+	if(istype(used_item, /obj/item/stack/material/bar))
+		var/obj/item/stack/material/bar/bar = used_item
+		if(bar.get_amount() == 1 && current_health >= get_max_health())
```
out-of-phaze: this means the moment it takes any damage from use you stop being able to place a bar on it...

code/game/objects/objs.dm
MistakeNot4892: Tried using base colour instead of the `64` component and the bars just turn #ff0000 immediately. Will look into further.

out-of-phaze (dismissed):

code/game/objects/objs.dm
out-of-phaze:
```dm
		var/list/base_color = rgb2num(get_color())
		h_r = base_color[1] + (h_r - base_color[1])*scale
		h_g = base_color[2] + (h_g - base_color[2])*scale
		h_b = base_color[3] + (h_b - base_color[3])*scale
```

code/modules/crafting/forging/forge_anvil.dm
MistakeNot4892: Yes. Because otherwise you cannot repair an anvil.

code/game/objects/objs.dm
MistakeNot4892: Unsure what I was doing wrong previously but it does appear to be working now using the HEX_GREEN etc helpers. Bars are going to need to keep track of their glow temperature possibly since update_icon() appears to be called frequently on them and causes the glow animation to reset.

code/modules/crafting/forging/forge_anvil.dm
MistakeNot4892: Added a material check and removed the amount check. If you want to put a single iron  bar onto a damaged iron anvil you will need to use throw-place but that seems like a niche enough problem to not matter.

MistakeNot4892:
Breaking this up a bit for ease of workflow/review:
~~#4709~~, ~~#4710~~, ~~#4711~~, ~~#4712~~, ~~#4713~~

MistakeNot4892:
I am flagging on this one a bit. It's basically ready to go other than some balancing around how long billets retain heat, and getting a better sound effect for the forging. Plus icons for the various blanks.

MistakeNot4892:
Happy with where this is sitting now, other stuff can be a follow up.

MistakeNot4892:
Depends on #4802

out-of-phaze (approved):
note to self add smithing stuff to karzerfeste

PR #4673, opened 2024-12-26
Add covered wells

out-of-phaze:
## Description of changes
Adds covered wells as an alternative to the open one we currently have.
image
image

Wells with a reinforcement material have posts and a roof. There is currently no way to add this to an existing well but it probably wouldn't be too hard?

Also replaces the well on Shaded Hills' inn map with a covered well, see screenshot.

I just think it looks neat 😃 
Also it looks better than cramming it between two walls.

## Authorship
Original well sprites by Doe/PlayerDeer, I think? (EDIT: actually it was by Loaf apparently) I kitbashed the components of the covered well sprite from some of her stuff, liiiike I think the posts are taken from timber frame walls? But the roof is from the log wall frame which was by Noelle I... think, but maybe it was originally done by Doe also I can't remember
I did the very minor amounts of code in this

## Changelog
:cl:
add: Added a covered well, a (for now) cosmetic variant of normal wells.
/:cl:


MistakeNot4892 (approved):
Would be nice to have a basic bitch attackby() block to add a canopy since it's just cosmetic, but not fussed really. Looks nice.

PR #4677, opened 2024-12-27
Adding fish oil.

MistakeNot4892:
## Description of changes
- Adds cave fish and lantern fish to the Shaded Hills cave river.
- Adds a base lantern fuel oil type which is the parent type of plant oil and corn oil.
- Adds fish oil.
- Adds recipes to make fuel oil from fish, plant or corn oil and powdered coal.

- More characterful fishes for the cave river.
- More sources of fuel on the medieval map.

## Authorship
Code is mine, sprites are @PlayerDeer <3

## Changelog
:cl:
tweak: Raw plant oil no longer works as lantern fuel; it must be mixed with powdered graphite first.
/:cl:


MistakeNot4892:
~~I do not have decent sprites for the cave fish/lantern fish, I will kitbash some when I'm back home after the 1st but if anyone with more artistic skills than me wants to have a crack, it would be appreciated.~~ Thank you Doe!

code/modules/mob/living/simple_animal/aquatic/aquatic_fish_lantern.dm
```diff
@@ -0,0 +1,26 @@
+/mob/living/simple_animal/aquatic/fish/large/lantern
+	name = "lantern fish"
+	desc = "An oily, glowing fish with a habit of nipping people. They are sometimes caught in cave rivers, and are rumoured to have cousins in the deep ocean."
```
out-of-phaze: technically I think the original joke with the lanternfish was that it was the supposed explanation for someone being inexplicably covered in oil (because they'd actually fallen into the slaughterhouse water pit, which mixed their lantern's oil with the water. this then set them on fire

basically: it would be really funny if they got oil on you when they attack you

edit: not a requested change btw i was gonna approve but then i saw the other thing

code/modules/organs/internal/_internal.dm
```diff
@@ -171,7 +171,7 @@
 
 		// We clamp/round here so that we don't accidentally heal past the threshold and
 		// cheat our way into a full second threshold of healing.
-		damage = clamp(damage-get_organ_heal_amount(), min_heal_val, absolute_max_damage)
+		damage = clamp(damage - max(0, get_organ_heal_amount() * GET_CHEMICAL_EFFECT(owner, CE_BRAIN_REGEN)), min_heal_val, absolute_max_damage)
```
out-of-phaze: doesn't this apply to all organs and not just the brain? i'm pretty sure this turns fish oil/neuroannealer into budget peridaxon. also it means it won't regenerate at all without any in your system, you might want like. `* (1 + GET_CHEMICAL_EFFECT(owner, CE_BRAIN_REGEN))` or something

if that's not desired my proposal would be to give organs a var like `healing_chem_effect` and do something like
```dm
		var/healing_multiplier = 1 + (healing_effect ? GET_CHEMICAL_EFFECT(owner, healing_effect) : 0)
		damage = clamp(damage - max(0, get_organ_heal_amount() * healing_multiplier), min_heal_val, absolute_max_damage)
```

out-of-phaze (changes requested):

code/modules/organs/internal/_internal.dm
MistakeNot4892: Might just move it into a proc and override it on the brain.

code/modules/mob/living/simple_animal/aquatic/aquatic_fish_lantern.dm
MistakeNot4892: Fish don't attack currently, but we could consider a lantern carp?

code/modules/organs/internal/_internal.dm
MistakeNot4892: Made CE_BRAIN_REGEN modify get_organ_heal_amount() directly.

code/modules/organs/internal/brain.dm
```diff
@@ -99,13 +99,13 @@
...
-	return (damage && owner && GET_CHEMICAL_EFFECT(owner, CE_BRAIN_REGEN)) || ..()
+	return (damage && owner && GET_CHEMICAL_EFFECT(owner, CE_BRAIN_REGEN) > 0) || ..()
 
 /obj/item/organ/internal/brain/has_limited_healing()
-	return (!owner || !GET_CHEMICAL_EFFECT(owner, CE_BRAIN_REGEN)) && ..()
+	return (!owner || GET_CHEMICAL_EFFECT(owner, CE_BRAIN_REGEN) <= 0) && ..()
 
 /obj/item/organ/internal/brain/get_organ_heal_amount()
-	return 1
+	return owner ? GET_CHEMICAL_EFFECT(owner, CE_BRAIN_REGEN) : 1
```
out-of-phaze: oh, so it *is* intended that it never heals on its own if you don't have fish oil? otherwise you should probably be adding 1 to the effect

out-of-phaze (approved):

code/modules/organs/internal/brain.dm
MistakeNot4892: I forgor brains regen a little without neuroannealer. Dammit.

out-of-phaze: well, I guess thats a change to request?
```suggestion
	return 1 + (owner ? GET_CHEMICAL_EFFECT(owner, CE_BRAIN_REGEN) : 0)
```
this will mean it heals 1 base, and then gains more based on the regen effect strength.

out-of-phaze (changes requested):

code/modules/organs/internal/brain.dm
MistakeNot4892: Okay, getting it to heal 1 + brain heal if not under a threshold, brain heal only if over damage threshold.

out-of-phaze (approved):

PR #4678, opened 2024-12-27
Remove and condense unnecessary side-overrides in core code

out-of-phaze:
## Description of changes
Removes or condenses a lot of unnecessary side-overrides in core code.
~~Some of these commits are going to be sent to stable/staging as applicable because they're fixes rather than code quality edits.~~ Done. #4679

Found a lot of bugs/bad copypasta and also just improves code quality overall. Simplifies control flow and somewhat improves performance on hot procs like update_icon.

## Authorship
Me

### DIFF
--- a/code/_helpers/unsorted.dm
+++ b/code/_helpers/unsorted.dm
@@ -719,11 +719,8 @@ Turf and target are seperate in case you want to teleport some distance from a t
 /obj/item/weldingtool/can_puncture()
 	return 1
 
-/obj/item/screwdriver/can_puncture()
-	return 1
-
 /obj/item/clothing/mask/smokable/cigarette/can_puncture()
-	return src.lit
+	return ..() || lit // in case someone has a sharp cigarette for some reason
 
 /*
 Checks if that loc and dir has a item on the wall
--- a/code/game/objects/items/__item.dm
+++ b/code/game/objects/items/__item.dm
@@ -378,7 +378,7 @@
 	return ..(user, distance, "", jointext(desc_comp, "<br/>"))
 
 /obj/item/check_mousedrop_adjacency(var/atom/over, var/mob/user)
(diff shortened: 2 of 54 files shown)


code/_helpers/unsorted.dm
```diff
@@ -719,11 +719,8 @@ Turf and target are seperate in case you want to teleport some distance from a t
 /obj/item/weldingtool/can_puncture()
 	return 1
```
MistakeNot4892, line 720: This should return `isOn()` I suspect.

MistakeNot4892: Not blocking, just a note.

code/game/objects/items/__item.dm
```diff
@@ -378,7 +378,7 @@
 	return ..(user, distance, "", jointext(desc_comp, "<br/>"))
 
 /obj/item/check_mousedrop_adjacency(var/atom/over, var/mob/user)
-	. = (loc == user && istype(over, /obj/screen/inventory)) || ..()
```
MistakeNot4892, line 381: I think /inventory might be correct, there's no need to drag and drop over other elements that I can think of.

out-of-phaze:
```
/obj/item/organ/internal/augment/active/cyberbrain/handle_mouse_drop(atom/over, mob/user, params)
	if(!istype(over, /obj/screen))
		attack_self(user)
		return TRUE
	. = ..()
```
dunno, but it's not the only place that checks for specifically screen as opposed to inventory. was just going for what current code did, since by checking both with `||` we're defaulting to the less-specific one

MistakeNot4892 (dismissed):
Comments are not blocking.

MistakeNot4892 (approved):

PR #4679, opened 2024-12-27
Uninclude corporate modpack from core code, fix misc. pathing/side-override bugs

out-of-phaze:
## Description of changes
- Fixes all maps having the corporate modpack.
  - Moves oldpod ruin into the corporate modpack.
  - Moves hacked headsets into core code.
  - Adds a corporate modpack specific override for random spawner loot.
- Fixes core code not compiling without the core modpack. (TORCH 2 OH NOOO)
- Makes the example map fail CI if it detects modpacks have been loaded.
- Fixes a bunch of mistyped/mispathed proc overrides.

Issues found via the Dreamchecker `redefined_procs` lint. It's disabled in Nebula CI because we use a lot of those for legitimate reasons, including modpacks—so instead we (aka I) periodically audit the results of that lint to find issues to fix.

## Authorship
Me


code/game/objects/structures/crates_lockers/closets/secure/scientist.dm
```diff
@@ -64,12 +64,14 @@
...
 	req_access = list(access_xenobiology)
+	/// The jumpsuit type spawned for this locker. Exists to be overridden by the corporate modpack, which adds pilot jumpsuits.
+	var/jumpsuit_type = /obj/item/clothing/jumpsuit/white
 
 /obj/structure/closet/secure_closet/pilot/WillContain()
 	return list(
 		/obj/item/backpack/parachute,
 		/obj/item/knife/utility,
-		/obj/item/clothing/jumpsuit/pilot,
+		jumpsuit_type,
```
MistakeNot4892, line 74: Parens on this to stop it getting interpreted as a key, possibly? Not sure if that is only an issue with assoc list init.

MistakeNot4892: Nvm it is only assoc.

code/modules/random_map/drop/drop_types.dm
```diff
@@ -51,18 +51,6 @@ var/global/list/datum/supply_drop_loot/supply_drop
 		/obj/item/gun/projectile/automatic/smg,
 		/obj/item/gun/projectile/automatic/assault_rifle)
 
-/datum/supply_drop_loot/ballistics
```
MistakeNot4892, line 54: Note to self to turn this type into a /decl.

MistakeNot4892:
Looks good, will approve with CI working.

MistakeNot4892 (approved):

PR #4681, opened 2024-12-27
Compiled tweaks from old codex dump branch.

MistakeNot4892:
Cleaning up old branches, there were a couple of small changes in my working copy that I missed when I got it merged to main apparently.

### DIFF
--- a/code/game/objects/items/weapons/policetape.dm
+++ b/code/game/objects/items/weapons/policetape.dm
@@ -250,7 +250,7 @@ var/global/list/image/hazard_overlays //Cached hazard floor overlays for the bar
 /obj/structure/tape_barricade/proc/update_neighbors(var/location = loc)
 	for (var/look_dir in global.cardinal)
 		var/obj/structure/tape_barricade/B = locate(/obj/structure/tape_barricade, get_step(location, look_dir))
-		if(B && !QDELETED(B))
+		if(!QDELETED(B))
 			B.update_icon()
 
 	if(!QDELETED(src))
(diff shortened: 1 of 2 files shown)


code/game/objects/items/weapons/policetape.dm
```diff
@@ -261,10 +261,11 @@ var/global/list/image/hazard_overlays //Cached hazard floor overlays for the bar
 	neighbors = 0
 	for (var/look_dir in global.cardinal)
 		var/turf/target_turf = get_step(src, look_dir)
-		var/obj/structure/tape_barricade/B = locate(/obj/structure/tape_barricade, target_turf)
-		//We connect to walls and other tape_barricades
-		if((B && !QDELETED(B)) || (!B && target_turf?.is_wall()))
-			neighbors |= look_dir
+		if(target_turf)
+			var/obj/structure/tape_barricade/B = locate(/obj/structure/tape_barricade, target_turf)
```
out-of-phaze:
```suggestion
			var/obj/structure/tape_barricade/B = locate(/obj/structure/tape_barricade) in target_turf
```
uhhhh. i don't know if the original syntax is valid or not but it definitely gives me the heebie-jeebies. please change it and i will get around to changing the other cases of this also

out-of-phaze (changes requested):
well i sure *hope* they've been compiled! (joke)

code/game/objects/items/weapons/policetape.dm
MistakeNot4892: It's valid syntax, but I'll tweak it.

out-of-phaze (approved):

PR #4682, opened 2024-12-27
Remove mind.original

out-of-phaze:
There is a ~12 year old TODO about removing this. 

Also updates the Vox raider mirror to use the special role system properly. Same with some traitor borg stuff. Certain Law 0 handling was moved out of the traitor modpack as well.


out-of-phaze:
I have no idea how much this will break things

MistakeNot4892 (approved):
Thought I had approved this - if it breaks stuff we can fix it. Ez pz.

Issue #4684, opened 2024-12-28
Janibot does not automop floors


Recon104:
#### Description of issue

Janibot currently only cleans up a few select tiles like oilstain footsteps and should be automopping most dirt/filth on floor

#### Difference between expected and actual behavior

 Cleans up a few select tiles 
Should be automopping most dirt/filth on floor

#### Steps to reproduce

Spawn as janitor cyborg on filthy station, try to mop by moving over the tiles 

#### Specific information for locating

Janitor Cyborg
Filth
Dirt
Vomit
Blood
etc

#### Length of time in which bug has been known to occur

Few weeks

#### Client version, Server revision & Game ID

Revision: ScavStation
Date
2024-12-28
Branch
dev
Game ID
cxo-aaE6
Runtime Version
515.1643
Compiler Version
515.1643
Revision
1e6967ca994e71cbc630e54e49980272d4b23be6

#### Issue bingo

- [X] Issue could be reproduced at least once
- [X] Issue could be reproduced by different players
- [X] Issue could be reproduced in multiple rounds
- [X] Issue happened in a recent (less than 7 days ago) round
- [X] [Couldn't find an existing issue about this](https://github.com/NebulaSS13/Nebula/issues)

MistakeNot4892:
They appear to be getting stuck trying to get to dirt overlays on turfs they can't path to.

MistakeNot4892:
Oh I totally mistread this as being about cleanbots not janitor borgs. Oops.

PR #4686, opened 2024-12-29
Misc staging fixes.

MistakeNot4892:
- Fixes https://github.com/ScavStation/ScavStation/issues/1160.
- Attempts to mitigate the worst of #4684 (not necessarily successfully).
- Simplifies AI attack windups.
- Allows ingested items to melt (mostly for people eating water cubes or nutriment scraps).

### DIFF
--- a/code/game/objects/items/_item_melting.dm
+++ b/code/game/objects/items/_item_melting.dm
@@ -16,16 +16,22 @@
 		try_burn_wearer(holder, holder.get_equipped_slot_for_item(src))
 
 	// Temp gate until generalized temperature-based melting works properly.
-	if(istype(loc, /obj/item/chems/crucible))
-		// Check if this is meltable at all.
-		var/list/meltable_materials
-		for(var/mat in matter)
-			var/decl/material/melt_material = GET_DECL(mat)
-			if(!isnull(melt_material.melting_point) && temperature >= melt_material.melting_point)
-				LAZYDISTINCTADD(meltable_materials, melt_material)
-		if(length(meltable_materials))
-			. = null // Don't return PROCESS_KILL here.
-			handle_melting(meltable_materials)
+	var/static/list/_melting_containers = list(
+		/obj/item/chems/crucible,
+		/obj/item/organ/internal/stomach
+	)
+	if(!is_type_in_list(loc, _melting_containers))
+		return
+
+	// Check if this is meltable at all.
+	var/list/meltable_materials
+	for(var/mat in matter)
+		var/decl/material/melt_material = GET_DECL(mat)
+		if(!isnull(melt_material.melting_point) && temperature >= melt_material.melting_point)
+			LAZYDISTINCTADD(meltable_materials, melt_material)
+	if(length(meltable_materials))
+		. = null // Don't return PROCESS_KILL here.
+		handle_melting(meltable_materials)
 
 /obj/item/place_melted_product(list/meltable_materials)
 
--- a/code/modules/mob/living/bot/floorbot.dm
+++ b/code/modules/mob/living/bot/floorbot.dm
@@ -122,15 +122,11 @@
 		if(S.material?.type == /decl/material/solid/metal/steel)
 			return (amount < maxAmount && maketiles)
 
-	if(target.loc.name == "Space")
-		return 0
+	var/turf/floor/my_turf = target
+	if(!istype(my_turf) || (isturf(my_turf) && my_turf.is_open()))
+		return FALSE
 
-	var/turf/floor/T = target
-	if(istype(T))
-		if(emagged)
-			return 1
-		else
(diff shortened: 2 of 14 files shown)


code/game/objects/items/_item_melting.dm
```diff
@@ -16,16 +16,22 @@
...
-				LAZYDISTINCTADD(meltable_materials, melt_material)
-		if(length(meltable_materials))
-			. = null // Don't return PROCESS_KILL here.
-			handle_melting(meltable_materials)
+	var/static/list/_melting_containers = list(
+		/obj/item/chems/crucible,
+		/obj/item/organ/internal/stomach
+	)
+	if(!is_type_in_list(loc, _melting_containers))
+		return
```
out-of-phaze, line 24: i would ideally like for this to use a typecache, since it's marginally faster if we ever add more types to this, but it should be fine for now

code/modules/mob/living/bot/floorbot.dm
```diff
@@ -109,23 +109,23 @@
...
 
-	if(istype(A, /obj/item/stack/material))
-		var/obj/item/stack/material/S = A
+	if(istype(target, /obj/item/stack/material))
+		var/obj/item/stack/material/S = target
 		if(S.material?.type == /decl/material/solid/metal/steel)
 			return (amount < maxAmount && maketiles)
 
-	if(A.loc.name == "Space")
+	if(target.loc.name == "Space")
```
out-of-phaze: WHAT.
this should probably be `is_open()` or whatever rather than... that. christ. oh god. how many other cases of this are there?

code/modules/organs/prosthetics/prosthetics_manufacturer_models.dm
```diff
@@ -10,7 +10,7 @@
 	icon_base = 'icons/mob/human_races/cyberlimbs/morgan/morgan_main.dmi'
 	modifier_string = "wooden"
 	hardiness = 0.75
-	manual_dexterity = DEXTERITY_SIMPLE_MACHINES
+	manual_dexterity = DEXTERITY_SIMPLE_MACHINES | DEXTERITY_HOLD_ITEM | DEXTERITY_EQUIP_ITEM | DEXTERITY_KEYBOARDS | DEXTERITY_GRAPPLE
```
out-of-phaze, line 13: i don't know how to feel about that. what does this even prevent you from doing, now? it kind of feels like we're making wooden limbs not actually much worse than modern prostheses, especially given that DEXTERITY_COMPLEX_TOOLS or whatever isn't really used much on shaded hills. maybe we should just make bumping airlocks not check dexterity if it's worn and not held? and to open structure doors, we could make it so that it just needs DEXTERITY_HOLD_ITEM.

alternatively i'm open to hearing why wooden prostheses should have all of these? but I kind of liked the niche they had before.

out-of-phaze (changes requested):

code/modules/organs/prosthetics/prosthetics_manufacturer_models.dm
MistakeNot4892: Not being able to hold and equip items makes them worse than mouth slots. They are almost completely pointless. Keyboards seems like an oversight, you can peck type with a hook. Grapple I'm ambivalent about, put it on there because you need it to drag. 

Not being able to use complex electronics or weapons makes them still notably different/worse than regular prosthetics imo.

code/game/objects/items/_item_melting.dm
MistakeNot4892: Seems like overkill for what amounts to a temporary workaround to be honest.

code/modules/organs/prosthetics/prosthetics_manufacturer_models.dm
MistakeNot4892: I could override them on Shaded Hills but they are really crap to use at the moment.

out-of-phaze: sure, I'll take that. the `name == "Space"` thing should still go though, otherwise that's the only change requested

code/game/objects/items/_item_melting.dm
out-of-phaze: yeah fair enough, like I said, fine for now

code/modules/organs/prosthetics/prosthetics_manufacturer_models.dm
MistakeNot4892: Yeah I just pushed a fix for that.

MistakeNot4892: As a case in point for how useless they are, you cannot currently remove your headset with a prosthetic wooden hand. Would like to just buff them overall to be honest.

out-of-phaze (approved):

PR #4690, opened 2025-01-03
Add modpack compatibility patch system, improve modularity

out-of-phaze:
## Description of changes
- Adds a new modpack compatibility patch system to get around the annoying include order issues the old system had.
- Removes `get_alternate_reaction_indicator()` in favor of using `send_data()` instead, because only soulstone synthesis used it.
- Allows modular code to override/add to certain lists of fluff strings in core code.
- Repaths `/supermatter` material presets to `/exotic_matter` in preparation for moving supermatter content into a modpack. Also, consistency; that's the material path anyway.
- Replaces hardcoded singularity stage size checks with specific variables, for more granular control over stage behavior (and to make adding them in modpacks a lot easier).
- Makes the atmos grief fixer use a decl step system in preparation for moving the "disable the SM" step into a modpack. Theoretically this could be expanded and/or generalized too.

# DNM until I do at least some basic testing that the compatibility patches actually work properly.

Fixes some potential headaches/issues with modpack include order and compatibility patches. Improves general code quality and consistency. Prepares for moving the SM and related content into a modpack.

## Authorship
Me

## Changelog
No user-facing changes.


MistakeNot4892 (dismissed):
Meant to approve this one, not the one with the dep, but yeah this seems like a good solution.

MistakeNot4892 (approved):

PR #4691, opened 2025-01-03
Move the supermatter crystal and related content into a modpack

out-of-phaze:
## Description of changes
Moves the supermatter crystal and related content into a modpack.

Depends on #4690.

I have no clue how I'd test this. Engine setup verb, check that the UIs work, fix atmos grief verb, try a delam?

Moves the SM into a modpack, which is a good proof-of-concept for "what if we modpacked base-game SS13 features?" Like, I know that Shaded Hills doesn't suffer at all from "supermatter crystals existing in the game code but being inaccessible," but modpacking it highlights places our modularity is lacking, and it makes it easier for codebases to pick-and-choose what they want in their lore; some might not want supermatter crystals to be orderable at all, for example.

## Authorship
Me.

## Changelog
Shouldn't be any user-facing changes?


MistakeNot4892 (dismissed):
This all seems conceptually fine to me. Have to admit I did not read the copied/moved code in great depth.

mods/content/supermatter/overrides/sm_singularity.dm
```diff
@@ -0,0 +1,46 @@
...
+	em_heavy_range = 12
+	em_light_range = 16
+	mesmerize_text = "helpless"
+	the_goggles_do_nothing = TRUE
+	ignore_obstacles = TRUE
+
+/decl/singularity_stage/stage_super/grow_to(obj/effect/singularity/source)
+	source.visible_message(SPAN_SINISTER("<font size='3'>You witness the creation of a destructive force that cannot possibly be stopped by human hands.</font>"))
+
+// why is this not shrink_from or something?
```
MistakeNot4892, line 31: Because it's shrinking to this stage from stage six.

MistakeNot4892 (approved):

PR #4693, opened 2025-01-04
Staging fixes.

MistakeNot4892:
- Fixes https://github.com/ScavStation/ScavStation/issues/1169.
- Mitigates https://github.com/ScavStation/ScavStation/issues/1164.
- Fixes https://github.com/ScavStation/ScavStation/issues/1162.
- Fixes https://github.com/ScavStation/ScavStation/issues/1161.
- Fixes https://github.com/ScavStation/ScavStation/issues/1172.
- Fixes https://github.com/ScavStation/ScavStation/issues/1163.
- Dripping is now faster.
- fluid_act() is now applied to mobs only after meeting certain thresholds.
- Paint is not applied except when the volume exceeds item size for movables and a minimum fluid threshold for turfs.
- Floor slipperiness is only applied after meeting a minimum fluid threshold.
- Dirt will no longer visibly fall through open turfs.
- Adds level data list validation to /datum/map and tweaks maps to pass the checks.
- Removes testing maps from map votes.


code/controllers/subsystems/overlays.dm
```diff
@@ -209,7 +209,7 @@ SUBSYSTEM_DEF(overlays)
 		if(cut_old)
 			our_overlays = cached_other.Copy()
 		else
-			our_overlays |= cached_other
+			LAZYDISTINCTADD(our_overlays, cached_other)
```
out-of-phaze, line 212: this change doesn't do anything, does it? `null |= list(foo)` == `list(foo)`. i think `null |= null` might be an issue but we know `cached_other` is non-null here.

code/game/objects/items/_item_reagents.dm
```diff
@@ -64,12 +64,13 @@
...
+	var/had_units   = target.reagents?.total_volume
+	reagents.trans_to(target, amount)
 
 	if(had_liquids)
 		playsound(src, 'sound/effects/pour.ogg', 25, 1)
 	else
 		// Sounds more like pouring small pellets or dust.
 		playsound(src, 'sound/effects/refill.ogg', 25, 1)
-	to_chat(user, SPAN_NOTICE("You transfer [trans] unit\s of the solution to \the [target].  \The [src] now contains [src.reagents.total_volume] units."))
+	to_chat(user, SPAN_NOTICE("You transfer [target.reagents?.total_volume - had_units] unit\s of the solution to \the [target]. \The [src] now contains [reagents.total_volume] unit\s."))
```
out-of-phaze: why's this necessary? shouldn't we be using the return value of `reagents.trans_to()`?

code/modules/client/preference_setup/background/01_species.dm
```diff
@@ -49,6 +49,8 @@
 	. += "<tr><td colspan=3><center>"
 	for(var/s in get_playable_species())
 		var/decl/species/list_species = get_species_by_key(s)
+		if(!istype(list_species))
+			continue // Runtime observed in a round where the playable species list was edited.
```
out-of-phaze: it was edited to the wrong value because i was silly, this check is pointless and i fixed it in-game immediately

code/modules/crafting/slapcrafting/_crafting_stage.dm
```diff
@@ -76,17 +76,18 @@
...
+		if(stack.amount > stack_consume_amount)
+			used_stack = stack.split(stack_consume_amount)
+		else
 			if(!user.try_unequip(thing, target))
 				return FALSE
 			used_stack = stack
-		else
-			used_stack = stack.split(stack_consume_amount)
-		used_stack.forceMove(target)
+		if(used_stack)
```
out-of-phaze:
```suggestion
		if(!QDELETED(used_stack))
```
unsure if this is necessary or not but it probably wouldn't hurt if you feel up to changing it

code/modules/events/dust.dm
```diff
@@ -14,13 +14,17 @@ The "dust" will damage the hull of the station causin minor hull breaches.
...
+	if(world.time > last_wave + min_delay && prob(10) && length(affecting_z))
 		dust_swarm(severity, affecting_z)
 
 /datum/event/dust/end()
 	command_announcement.Announce("The [location_name()] has now passed through the belt of space dust.", "[location_name()] Sensor Array", zlevels = affecting_z)
 
 /proc/dust_swarm(var/strength = EVENT_LEVEL_MUNDANE, var/list/zlevels)
+
+	if(!length(zlevels))
+		return // Not sure how this happened, but saw it in a runtime on Pyrelight.
```
out-of-phaze, line 26: `affecting_z = SSmapping.station_levels`
karzerfeste must not have level data set properly or something, if there were no station levels

code/game/objects/structures/well.dm
```diff
@@ -57,13 +57,6 @@
 	if(!is_processing && auto_refill)
 		START_PROCESSING(SSobj, src)
 
-/obj/structure/reagent_dispensers/well/attackby(obj/item/W, mob/user)
```
out-of-phaze, line 60: Why was this removed? Is it just totally redundant now?

out-of-phaze (dismissed):
there seem to be a lot of confusing and poorly atomized changes in this PR, but it's a fix pr so i don't mind too much. i do have a few nitpicks though

code/controllers/subsystems/overlays.dm
MistakeNot4892: There was a `null |= 0` runtime on Scav that I didn't narrow down past this line, just saw that every other use of this list is LAZY.

code/game/objects/items/_item_reagents.dm
MistakeNot4892: It was returning null, resulting in a broken message.

code/game/objects/structures/well.dm
MistakeNot4892: Yeah there's an alt interaction to handle it. Keeping it in attackby is going to intercept clicks/confuse people.

MistakeNot4892: Likewise the removal in turf attackby below.

code/modules/client/preference_setup/background/01_species.dm
MistakeNot4892: I can revert if desired, but it's harmless at worst imo.

MistakeNot4892:
Sorry about the non-atomic changes, this was a bit of a stream of consciousness PR. Thought I had them mostly broken up, alas.

code/game/objects/items/_item_reagents.dm
MistakeNot4892: Looks like `trans_to_turf()` reliably returns null.

code/modules/events/dust.dm
MistakeNot4892: I've added a validation step to this PR to check for that.

code/modules/client/preference_setup/background/01_species.dm
MistakeNot4892: Reverted.

code/game/objects/items/_item_reagents.dm
out-of-phaze: yeah i think it's one of the only ones that doesn't. i really should do another opendream typemaker pass to catch these things tbh. trans_to_turf and all the other trans_to helpers should return whatever `trans_to_holder` does

code/controllers/subsystems/overlays.dm
out-of-phaze: uh. it is entirely impossible for cached_other to ever be `0`, what was the runtime specifically? i mean maybe it was `"0"` in which case it'd be fine. but actually no, cached_other should always be a list or null

MistakeNot4892: I don't have the runtime anymore. I also don't think the specifics of the runtime matter - this is a consistency change, every other use of this list is LAZY.

code/game/objects/items/_item_reagents.dm
MistakeNot4892: Should be more consistent now at least.

out-of-phaze (approved):
this is really poorly atomized. welp! good enough for me.

also someone needs to fix the nanoui map generation to only take into account the level data bounds or whatever. i'll probably do it later

PR #4695, opened 2025-01-05
Vomit and mud can now be tracked around the floor.

MistakeNot4892:
## Description of changes
- Added mob-level general coating proc for shoes/feet coatings.
- Vomit and mud can coat shoes/feet and get tracked around like blood.
- Quadrupeds will get coatings on all four feet, not just back feet.

image
image

Janitor can yell at people for tracking mud/sick around.

## Authorship
Myself.

## Changelog
:cl:
tweak: Mud and blood can now leave footprints.
/:cl:

### DIFF
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -1552,7 +1552,7 @@ default behaviour is:
 /mob/living/OnSimulatedTurfEntered(turf/T, old_loc)
 	T.add_dirt(0.5)
 
-	HandleBloodTrail(T, old_loc)
+	handle_walking_tracks(T, old_loc)
 
 	if(current_posture.prone)
 		return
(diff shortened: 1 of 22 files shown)


code/modules/mob/living/living.dm
```diff
@@ -1583,8 +1583,46 @@ default behaviour is:
...
+
+	if(!T.can_show_footsteps())
+		return
+
+	// Tracking blood or other contaminants
+	var/obj/item/source
+	var/obj/item/clothing/shoes/shoes = get_equipped_item(slot_shoes_str)
+	if(istype(shoes))
+		shoes.handle_movement(src, MOVING_QUICKLY(src))
+		if(shoes.coating && shoes.coating.total_volume > 1)
```
out-of-phaze, line 1596: I wonder if this should more generally loop over items that cover the zones associated with child stance limbs, or something? I don't think it'll be an issue for now, though.

out-of-phaze (approved):

code/modules/mob/living/living.dm
MistakeNot4892: I considered it but drakes can't wear gloves so no current need.

PR #4696, opened 2025-01-05
Move miscellaneous species (golems, aliums) to modpacks

out-of-phaze:
## Description of changes
- Moves golems to the xenobiology modpack.
- Moves random humanoid aliens to their own modpack.
- Adds a basic example of "spawner overridden by a modpack" so that maps can have optional modpack content.

This results in the random humanoid aliens, accessible only via the alien monolith in the ninja/heist bases, being unavailable on all current maps. It's sort of a soft-removal aside from in the modpack testing map—so it'll be maintained but isn't a super high priority or something that I, personally, think is worth having around in its current state.

Isolates core code better.

## Authorship
Me.


code/unit_tests/~unit_test_overrides.dm
```diff
@@ -26,7 +26,7 @@ var/global/atom/movable/unit_test_last_obj_random_creation
 	global.unit_test_last_obj_random_creation = create_instance(unit_test_select_heaviest(spawn_choices()), loc)
 
 /proc/unit_test_select_heaviest(var/list/choices)
-	if(ispath(choices) || istype(choices, /datum))
+	if(ispath(choices) || istype(choices, /datum) || !choices)
```
out-of-phaze: note to self: remove this, forgot to hit save

maps/__map_modpack_compatibility.dm
```diff
@@ -0,0 +1,25 @@
+// This must be included before maps and any modpacks that use it.
+// That includes before _map_include.dm.
+
+/// A spawner that can have its path overridden by a modpack, so that maps can have optional modpack compatibility.
+/obj/abstract/modpack_compat
+	abstract_type = /obj/abstract/modpack_compat
+	name = "optional modpack object"
+	desc = "This item type is used to spawn a movable at roundstart, whether a modpack is included or not."
```
MistakeNot4892, line 8: Oh I need to use these for scav stuff on Tradeship/Mini.

MistakeNot4892 (approved):

PR #4703, opened 2025-01-07
Couple of small fixes.

MistakeNot4892:
- Fixes non-watertight atoms only being watertight if they were open containers.
- Moves nanite fluid into compounds


out-of-phaze:
nanite thing isn't an issue upstream but it's fine to move it tbh

out-of-phaze (approved):

PR #4712, opened 2025-01-09
[IDB IGNORE] Adding grindstones and whetstones.

MistakeNot4892:
## Description of changes
- Converts sharp/edge to use getters on the object.
- Adds an item effect category for damage modification along with associated tweaks to item effect code.
- Adds a modpack containing whetstones and grindstones.
- Adds a sharpness item modifier on bladed weapons that can be reset by sharpening.

Actual use case for item effects, ability to add a damage buff from taking care of your kit.
Prep for blacksmithy (grindstone and whetstone will be used to turn blanks into swords)

## Authorship

## Changelog
:cl:
tweak: Swords on Shaded Hills can be sharpened using whetstones or grindstones.
/:cl:


out-of-phaze (dismissed):
not a huge fan of dry_run, i would personally prefer different procs i think? with `get_attack_force()` being a pure proc that returns a number and does nothing else, and then having another proc that both returns it and consumes effects.

for now this is fine though, unless you want to rewrite the whole thing yourself before the 24h is up :p

out-of-phaze (dismissed):
i'll go through later and try separating get_attack_force into two separate procs myself

code/game/machinery/doors/airlock.dm
```diff
@@ -802,7 +802,7 @@ About the new airlock wires panel:
 	else if((stat & (BROKEN|NOPOWER)) && isanimal(user))
 		var/mob/living/simple_animal/A = user
 		var/obj/item/I = A.get_natural_weapon()
-		if(I?.get_attack_force(user) >= 10)
+		if(I?.expend_attack_force(user) >= 10)
```
out-of-phaze, line 805: it's funny to imagine this dulling someone's claws. or that you could file your natural weapon claws to a point, but honestly no that rules.

mods/content/item_sharpening/item_sharpen.dm
```diff
@@ -0,0 +1,52 @@
...
+/obj/item/proc/can_sharpen_with(obj/item/sharpening_with)
+	if(!has_item_effect(/decl/item_effect/sharpened, IE_CAT_DAMAGE))
+		return FALSE
+	var/list/params = get_item_effect_parameters(/decl/item_effect/sharpened, IE_CAT_DAMAGE)
+	if(!islist(params) || params[IE_PAR_USES] >= params[IE_PAR_MAX_USES])
+		return FALSE
+	if(istype(sharpening_with, /obj/structure/working/grindstone))
+		var/obj/structure/working/grindstone/stone = sharpening_with
+		return !material || material.hardness <= stone.stone_material?.hardness
+	return !material || material.hardness <= sharpening_with.material?.hardness
```
out-of-phaze: kind of feel like this could maybe use a `get_sharpening_material()` proc overridden on grindstones/etc, but that's not necessary. mostly i just think it'd be nice to do
```suggestion
	var/decl/material/sharpening_material = sharpening_with.get_sharpening_material()
	return !material || material.hardness <= sharpening_material?.hardness
```
and then something like
```dm
/obj/proc/get_sharpening_material()
    return get_material() // <-- returns material vars on item and structure

/obj/structure/working/grindstone/get_sharpening_material()
    return stone_material
```
and then if the grindstone eventually has the actual stone used split out into an item in its contents or something, we could even do
```dm
/obj/structure/working/grindstone/get_sharpening_material()
    return stone?.get_sharpening_material()
```

mods/content/item_sharpening/item_sharpen.dm
```diff
@@ -0,0 +1,52 @@
+/obj/item/update_name()
+	. = ..()
+	if(has_item_effect(/decl/item_effect/sharpened, IE_CAT_EXAMINE) && get_item_effect_parameter(/decl/item_effect/sharpened, IE_CAT_DAMAGE, IE_PAR_USES) <= 0)
+		SetName("dulled [name]")
+
+/obj/item/proc/can_sharpen_with(obj/item/sharpening_with)
```
out-of-phaze: i don't like lying about parameters, this can take any `obj`.

mods/content/item_sharpening/item_sharpen.dm
```diff
@@ -0,0 +1,52 @@
...
+		return FALSE
+	var/list/params = get_item_effect_parameters(/decl/item_effect/sharpened, IE_CAT_DAMAGE)
+	if(!islist(params) || params[IE_PAR_USES] >= params[IE_PAR_MAX_USES])
+		return FALSE
+	if(istype(sharpening_with, /obj/structure/working/grindstone))
+		var/obj/structure/working/grindstone/stone = sharpening_with
+		return !material || material.hardness <= stone.stone_material?.hardness
+	return !material || material.hardness <= sharpening_with.material?.hardness
+
+/obj/item/proc/sharpen_with(mob/user, obj/item/sharpen_with)
```
out-of-phaze:
```suggestion
/obj/item/proc/sharpen_with(mob/user, obj/sharpen_with)
```
type doesn't match its caller's equivalent arg in `try_sharpen_with`

code/game/objects/items/_item_sharpness.dm
```diff
@@ -0,0 +1,21 @@
+/obj/item
+	var/_sharp = FALSE
```
out-of-phaze: wonder if it might be useful to do
```suggestion
	VAR_PROTECTED/sharp = FALSE
```
which will avoid the var rename and properly lint for undesired access in most spots (anything not descended from /obj/item). anything accessing it that's not derived from `/obj/item` would have to use the helper. it wouldn't enforce using set_sharp on `/obj/item` subtypes itself though, so idk

also, not blocking.
edit: could also just add `VAR_PROTECTED` and keep the rename to reduce the amount of work you'd have to do lol

out-of-phaze (dismissed):
few nits, will approve after they're addressed

code/game/machinery/doors/airlock.dm
MistakeNot4892: This proc will also allow for stuff like an ability granting charges or buffs, so seems fine to me.

code/game/objects/items/_item_sharpness.dm
MistakeNot4892: I thought I tried that and got linter complaints about overrides, but that may have been VAR_PRIVATE.

MistakeNot4892:
No idea why IDB is not running.

code/game/objects/items/blades/folding.dm
```diff
@@ -53,9 +53,8 @@
 
 /obj/item/bladed/folding/update_attack_force()
 	..()
-	// TODO: check sharp/edge.
-	edge  = open
-	sharp = open
+	set_edge(open)
```
out-of-phaze, line 56: out of scope for now but i genuinely wonder if maybe is_edge and is_sharp should just have overrides on this that return false if not open, rather than changing the var under the hood, lol. would mean a knife made of cardboard wouldn't become sharp when you open it

out-of-phaze (approved):

PR #4713, opened 2025-01-09
Adding bellows.

MistakeNot4892:
## Description of changes
Adds a (mappable only) bellows that can boost the temperature of fire sources directly east of it by 25%.

Useful prop for forging/smelting. Icon's great.

## Authorship
Code is mine, icon is @PlayerDeer.

## Changelog
Nothing player-facing.


code/game/objects/structures/fires.dm
```diff
@@ -386,7 +388,9 @@
 		return 0
 	var/ambient_temperature = get_ambient_temperature(absolute = TRUE)
 	// The effective burn temperature can't go below ambient (no cold flames) or above the actual burn temperature.
-	return clamp((last_fuel_burn_temperature - T0C) * draught_mult + T0C, ambient_temperature, last_fuel_burn_temperature)
+	. = clamp((last_fuel_burn_temperature - T0C) * draught_mult + T0C, ambient_temperature, last_fuel_burn_temperature)
+	if(bellows_oxygenation)
+		. = round(. * 1.25) // Burns 25% hotter while oxygenated.
```
out-of-phaze: this will result in it being ~68C hotter because it's applied after `+ T0C`. the oxygenation should maybe go in `get_draught_multiplier()`?

out-of-phaze (changes requested):

code/game/objects/structures/fires.dm
```diff
@@ -308,7 +310,9 @@
 	return ..()
 
 /obj/structure/fire_source/proc/get_draught_multiplier()
-	return has_draught ? draught_values[draught_values[current_draught]] : 1
+	. = has_draught ? draught_values[draught_values[current_draught]] : 1
+	if(bellows_oxygenation)
+		. += 0.25 // Burns 25% hotter while oxygenated.
```
out-of-phaze: i thought it would be * 1.25, because this means that closing the draught won't work while the bellows is going

out-of-phaze (changes requested):

code/game/objects/structures/fires.dm
MistakeNot4892: I figured it was air getting pumped in via the bellows, not via the draught.

out-of-phaze: that's a good point then tbh. i feel like we'd want closing the draught to turn it off entirely though, for ux reasons

out-of-phaze (approved):

PR #4717, opened 2025-01-10
Moving Shaded Hills level data and submap logic into fantasy modpack.

MistakeNot4892:
See title. This is largely just reorganisation and proc merging as Pyrelight wants to use this logic for other maps.


code/game/area/areas.dm
```diff
@@ -93,6 +94,10 @@ var/global/list/areas = list()
 	..()
 
 /area/Initialize()
+	if(additional_fishing_results)
```
out-of-phaze: miiiiight be worth using a LAZYLEN check just to avoid creating a zero-len fishing_results list if someone sets `additional_fishing_results` to an empty list, but it's not a huge deal

also i think we could probably null out additional_fishing_results after this but i'm not sure

EDIT: not blocking, the other comment is though

mods/content/fantasy/submaps/_submaps.dm
```diff
@@ -0,0 +1,52 @@
...
+/area/fantasy
+	abstract_type = /area/fantasy
+	allow_xenoarchaeology_finds = FALSE
+	icon = 'mods/content/fantasy/icons/areas.dmi'
+	icon_state = "area"
+	base_turf = /turf/floor/rock/basalt
+	sound_env = GENERIC
+	ambience = list()
+
+/area/fantasy/outside
```
out-of-phaze, line 33: this area has no name by default. it should probably be set to `"\improper Wilderness"` or something to match `/area/shaded_hills/outside` being named `"\improper Grasslands"`?

out-of-phaze (dismissed):

out-of-phaze (approved):

PR #4718, opened 2025-01-10
Update Create & Destroy unit test

out-of-phaze:
## Description of changes
Fixes the inconsistency with false positives (and false negatives) in the Create & Destroy test by checking based on gc_destroyed rather than queue time.
Also checks all queues and removes code duplication that handled the queues separately.
Also removes some unnecessary harddel hints by making them harddel normally.
Increases the filter queue time to 1 second rather than 0 seconds.

should make gc tests more consistent?

## Authorship
Taken from latest TG. Lots of PRs included, I'm not going to link them all unless someone specifically asks :(

### DIFF
--- a/code/___compile_options.dm
+++ b/code/___compile_options.dm
@@ -1,7 +1,3 @@
-// The default value for all uses of set background. Set background can cause gradual lag and is recommended you only turn this on if necessary.
-// 1 will enable set background. 0 will disable set background.
-#define BACKGROUND_ENABLED 0
-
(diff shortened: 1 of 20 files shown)


MistakeNot4892 (dismissed):

out-of-phaze:
Not closing, but this PR needs to be split up across multiple branches. It's too big to review like this and runs too high a risk of breaking stuff.

out-of-phaze:
Ready for review now, I split it into separate PRs.

MistakeNot4892 (approved):

PR #4720, opened 2025-01-12
Tweaking flooring procs to allow for multiple flooring layers.

MistakeNot4892:
## Description of changes
- Adds handling to `_flooring` for multiple floor types on the same turf.
- Changes wild grass to be `dirt, grass, wild grass` and adds a snowy grass type (`dirt, grass, snow`).

I would like some input on this one as it is blowing out init time on Shaded Hills 250% (almost a full minute) and there are some odd instances of completely black turfs. 

Allows for things like snow accumulation without ChangeTurf() or overriding the lower turf types.

## Authorship
Myself.

## Changelog
Nothing player-facing currently.

### DIFF
--- a/code/game/turfs/floors/_floor.dm
+++ b/code/game/turfs/floors/_floor.dm
@@ -24,10 +24,6 @@
 	// Flooring data.
 	var/floor_icon_state_override
 
-	// TODO:
-	VAR_PROTECTED/decl/flooring/_base_flooring = /decl/flooring/plating
-	VAR_PROTECTED/decl/flooring/_flooring
-
 	var/const/TRENCH_DEPTH_PER_ACTION = 100
 
 /turf/floor/Initialize(var/ml, var/floortype)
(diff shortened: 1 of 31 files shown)


MistakeNot4892:
Init time is still very bloated, up to 30 seconds from 20, but I fixed the icon issues.

code/game/turfs/floors/_floor.dm
```diff
@@ -96,118 +90,6 @@
...
-		enable_zmimic(z_flags)
-	else
-		disable_zmimic()
-
-	if(copy_from.has_environment_proc && !is_processing)
-		START_PROCESSING(SSobj, src)
-
-	levelupdate()
-
-	for(var/obj/effect/footprints/print in src)
```
MistakeNot4892, line 206: Note to self to reimplement this.

out-of-phaze:
as discussed on discord i think we'll need to do the 3-var approach where we cache the topmost flooring value, and the intermediary layers are in a list var. this will probably result in it accessing being _faster_ than it was before lol

MistakeNot4892:
Appears to be working. Tradeship inits in 10, Ministation inits in 80, Shaded Hills inits in 30, which is same as usual.

out-of-phaze (approved):

PR #4721, opened 2025-01-12
Soft turf interactions.

MistakeNot4892:
## Description of changes
- Snow, mud, and sand can be collected by hand in small amounts (prep for snowballs or mudfights).
- Snow, mud, and sand will leave visible footprints when moved across.
- Snowmen can be crafted.

image

More dynamic terrain, groundwork for snowballs/~~snowmen~~.
Potentially a newer system to replace bloody footprints down the track.

## Authorship
Code is mine, splitting of snowman icons was mine.
Original snowman icons came from Polaris, not sure of the original author.

## Changelog
:cl:
tweak: Snow, mud and sand will now show footprints.
add: Snowmen can be made from snow/mud/clay.
/:cl:


MistakeNot4892:
Went ahead and also ported snowmen: 
image

code/game/objects/structures/snowman.dm
```diff
@@ -0,0 +1,59 @@
...
+	// TODO: make carrot/stick arms/coal require items?
+	add_overlay(overlay_image(icon, "[icon_state]-decorations", COLOR_WHITE, RESET_COLOR))
+	compile_overlays()
+
+/obj/structure/snowman/proc/user_destroyed(user)
+	to_chat(user, SPAN_DANGER("\The [src] crumples into a pile of [material.solid_name] after a single solid hit. You monster."))
+	physically_destroyed()
+
+/obj/structure/snowman/attackby(obj/item/used_item, mob/user)
+	if(user.check_intent(I_FLAG_HARM) && used_item.get_base_attack_force())
```
out-of-phaze, line 40: there's a pr that will require adding `dry_run = TRUE` to this to avoid it double-spending sharpening charges, right? just noting it for later

code/game/turfs/flooring/_flooring.dm
```diff
@@ -261,15 +267,27 @@ var/global/list/flooring_cache = list()
...
 	return movement_delay
 
 /decl/flooring/proc/get_movable_alpha_mask_state(atom/movable/mover)
 	return
 
+/decl/flooring/proc/handle_hand_interaction(turf/floor/floor, mob/user)
+	if(!force_material || !can_collect)
+		return FALSE
+	user.visible_message(SPAN_NOTICE("\The [user] begins scraping together some of \the [name]..."))
+	if(do_after(user, 3 SECONDS, floor) && !QDELETED(floor) && !QDELETED(user) && floor.get_topmost_flooring() == src && isnull(user.get_active_held_item()))
```
out-of-phaze, line 284: does this not lower the height at all? it's okay if so, i just expected it to

out-of-phaze (approved):
move_trail should definitely be moved onto bodytype i think, note to self

code/game/objects/structures/snowman.dm
MistakeNot4892: Nope, this is base attack force, which is usually static.

code/game/turfs/flooring/_flooring.dm
MistakeNot4892: I could make it lower height but the behavior on Polaris is basically infinite snow. I guess we need to be more rigorous if it includes clay and sand since they can be used to make pottery and eventually glass. Will have a thinky thonk.

PR #4732, opened 2025-01-13
Reagent heaters can now heat skillets/pots.

MistakeNot4892:
## Description of changes
- Reagent heaters are renamed to hotplates.
- Reagent heaters will heat cooking vessels on their turf.
- Tradeship and Ministation kitchens have pots/skillets/hotplates.

Does not fix pots not showing their boiling overlay, cannot work out the cause of that.

Space maps can now make soup.

## Authorship
Myself.

## Changelog
:cl:
tweak: Reagent heaters/hotplates can be used to heat pots and skillets to make soup.
/:cl:

### DIFF
--- a/code/modules/food/cooking/cooking_vessels/_cooking_vessel.dm
+++ b/code/modules/food/cooking/cooking_vessels/_cooking_vessel.dm
@@ -40,6 +40,8 @@
 	return FALSE
 
 /obj/item/chems/cooking_vessel/afterattack(var/obj/target, var/mob/user, var/proximity)
+	if(!proximity || istype(target, /obj/machinery/reagent_temperature))
+		return FALSE
 	if(!ATOM_IS_OPEN_CONTAINER(src) || !proximity) //Is the container open & are they next to whatever they're clicking?
 		return FALSE //If not, do nothing.
 	if(target?.storage)
--- a/code/modules/reagents/heat_sources/_heat_source.dm
+++ b/code/modules/reagents/heat_sources/_heat_source.dm
@@ -97,6 +97,7 @@
 			return TRUE
 		thing.reset_offsets(anim_time = 0)
 		user.visible_message(SPAN_NOTICE("\The [user] places \the [thing] onto \the [src]."))
+		return TRUE
 
 	if(IS_WRENCH(thing))
 		if(use_power == POWER_USE_ACTIVE)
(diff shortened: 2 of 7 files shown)


out-of-phaze:
> Does not fix pots not showing their boiling overlay, cannot work out the cause of that.

like, it doesn't show the reagents overlay at all, or it just doesn't switch to the boiling one?

code/modules/food/cooking/cooking_vessels/_cooking_vessel.dm
```diff
@@ -40,6 +40,8 @@
 	return FALSE
 
 /obj/item/chems/cooking_vessel/afterattack(var/obj/target, var/mob/user, var/proximity)
+	if(!proximity || istype(target, /obj/machinery/reagent_temperature))
```
out-of-phaze, line 43: i kind of wonder if it might be a good idea to make these, eventually, use a single-item slot-based storage datum, since we already have support for intercepting afterattack based on `storage` below

code/modules/reagents/heat_sources/_heat_source.dm
```diff
@@ -79,10 +80,25 @@
 			if(container)
 				queue_temperature_atoms(container)
 			queue_icon_update()
+
+		// Hackery to heat pots placed onto a hotplate without also grilling/baking stuff.
+		if(isturf(loc))
+			var/datum/gas_mixture/environment = loc.return_air()
+			for(var/obj/item/chems/cooking_vessel/pot in loc.get_contained_external_atoms())
```
out-of-phaze, line 87: no skillet? fair enough I guess

out-of-phaze (approved):

code/modules/reagents/heat_sources/_heat_source.dm
MistakeNot4892: `pot` is the local variable name, it's iterating all cooking vessels.

code/modules/food/cooking/cooking_vessels/_cooking_vessel.dm
MistakeNot4892: Maybe yeah, this was a quick and dirty solution. I also need to add an electric oven at some point so I might do a broader rewrite then.

code/modules/reagents/heat_sources/_heat_source.dm
out-of-phaze: oop

PR #4734, opened 2025-01-13
Fixing cash vanishing when dispensed.

MistakeNot4892:
- Fixes https://github.com/ScavStation/ScavStation/issues/1192

### DIFF
--- a/code/modules/economy/cael/ATM.dm
+++ b/code/modules/economy/cael/ATM.dm
@@ -360,11 +360,15 @@
 					alert("That is not a valid amount.")
 				else if(authenticated_account && amount > 0)
 					//remove the money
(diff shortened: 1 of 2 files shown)


out-of-phaze (approved):

code/modules/economy/cael/ATM.dm
```diff
@@ -360,11 +360,15 @@
 					alert("That is not a valid amount.")
 				else if(authenticated_account && amount > 0)
 					//remove the money
+					// TODO: Jesus Christ why does this entire proc use usr
```
out-of-phaze, line 363: it's Topic(), you'd have to make it an OnTopic override to use `user` instead

PR #4739, opened 2025-01-15
Attack dexterity is bypassed for items in rigs.

MistakeNot4892:
## Description of changes
- Adds a getter for required attack dexterity.
- Provides required interaction dexterity to get_required_interaction_dexterity() for items.
- Bypasses required attack dexterity checks for items in rigs.
- Removes hardcoded gun dex check from special_check() in favour of attack dexterity check.

Dex system less wonky.
Drakes can use back-mounted machine guns in the future.

## Authorship
@Ofpelh found the issue, I just authored an alternative fix. 

## Changelog
:cl:
tweak: Guns can now be fired from rigs regardless of dexterity.
/:cl:

### DIFF
--- a/code/_onclick/item_attack.dm
+++ b/code/_onclick/item_attack.dm
@@ -26,6 +26,8 @@ avoid code duplication. This includes items that may sometimes act as a standard
 
 // If TRUE, prevent afterattack from running.
 /obj/item/proc/resolve_attackby(atom/A, mob/user, var/click_params)
+	if(!user.check_dexterity(get_required_attack_dexterity()))
+		return TRUE
 	if(!(item_flags & ITEM_FLAG_NO_PRINT))
 		add_fingerprint(user)
 	return A.attackby(src, user, click_params)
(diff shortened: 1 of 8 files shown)


code/_onclick/item_attack.dm
```diff
@@ -26,6 +26,8 @@ avoid code duplication. This includes items that may sometimes act as a standard
 
 // If TRUE, prevent afterattack from running.
 /obj/item/proc/resolve_attackby(atom/A, mob/user, var/click_params)
+	if(!user.check_dexterity(get_required_attack_dexterity()))
```
out-of-phaze: wasn't this check removed bc it caused problems with putting stuff on tables? i remember there being a reason needs_attack_dexterity isn't blanket checked in resolve_attackby.

out-of-phaze (dismissed):

code/_onclick/item_attack.dm
MistakeNot4892: Argh, maybe. Maybe we could pass the target as a param to get_required() and override it for tables?

MistakeNot4892: Okay, tables/racks set it to DEXTERITY_HOLD_ITEM when the user is on help intent (so not striking), atoms with storage set it to DEXTERITY_HOLD_ITEM universally. This may allow drakes to attack stuff with crowbars, but I'm not sure of a nice way to gate off the storage dexterity change.

out-of-phaze (approved):
let's give it a shot

PR #4746, opened 2025-01-16
Convert many Topic overrides to use OnTopic instead

out-of-phaze:
## Description of changes
HEAVY WIP. Will probably need testing but that'll be a huge pain with how much this touches, so it might need to be broken up.

Overriding Topic instead of OnTopic causes issues with missing parent call functionality and TOPIC_BLAH return values not being respected. This is better.


MistakeNot4892 (dismissed):
Not going to pretend I read the diff in detail, but nothing leaped out as obviously wrong.

MistakeNot4892 (approved):

PR #4756, opened 2025-01-18
Fix lighting overlays being created multiple times per turf

out-of-phaze:
## Description of changes
Fixes unnecessary/redundant z-level initialization and inlines a method that shouldn't ever be reused elsewhere.
Also re-enables some guard code to prevent this from being reintroduced.
Finally, we *always* go through lighting_build_overlay to make overlays, rather than just using a bare `new()` in SSlighting init.

Restores some assumptions that O7L uses, potentially optimizes some parts of lighting

### DIFF
--- a/code/controllers/subsystems/lighting.dm
+++ b/code/controllers/subsystems/lighting.dm
@@ -54,7 +54,12 @@ SUBSYSTEM_DEF(lighting)
 
 	// Generate overlays.
 	for (var/zlevel = 1 to world.maxz)
-		overlaycount += InitializeZlev(zlevel)
+		var/datum/level_data/level = SSmapping.levels_by_z[zlevel]
+		for (var/turf/tile as anything in block(1, 1, zlevel, level.level_max_width, level.level_max_height)) // include TRANSITIONEDGE turfs
+			if (TURF_IS_DYNAMICALLY_LIT_UNSAFE(tile))
+				tile.lighting_build_overlay()
+				overlaycount++
+			CHECK_TICK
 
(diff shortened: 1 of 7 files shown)


MistakeNot4892 (approved):

code/controllers/subsystems/lighting.dm
```diff
@@ -54,7 +54,12 @@ SUBSYSTEM_DEF(lighting)
 
 	// Generate overlays.
 	for (var/zlevel = 1 to world.maxz)
-		overlaycount += InitializeZlev(zlevel)
+		var/datum/level_data/level = SSmapping.levels_by_z[zlevel]
+		for (var/turf/tile as anything in block(1, 1, zlevel, level.level_max_width, level.level_max_height)) // include TRANSITIONEDGE turfs
+			if (TURF_IS_DYNAMICALLY_LIT_UNSAFE(tile))
+				tile.lighting_build_overlay()
```
Lohikar, line 60: Is using this helper a correctness fix? It was done manually for performance, though this is not particularly hot anyway.

out-of-phaze: well, i guess i could copy the CRASH or assert to this loop. mostly i just wanted to make damn well sure nothing was creating an overlay without it also checking to see if it already had one

Lohikar: It looks like O7 just has logging in this block, I'm fine with either approach assuming calling TBO here doesn't blow out lighting init times (which I assume it doesn't).

PR #4758, opened 2025-01-18
Optimize atom init significantly (init time halved)

out-of-phaze:
## Description of changes
Makes /atom/New during mapload take half as long to complete, bringing away sites testing init from 130 to 85 seconds on my machine—that means init is twice as fast. This was done by making the created_atoms list non-associative, since we don't rely on it for deduplication or membership checking, just to associate atoms to their arguments.

Init time drops drastically on maps with lots of atoms.


out-of-phaze:
Uh, issue. This won't pass CI until #4718 is in... so I'll just disable reftracking in CI for now, and then if that gets in first I'll rebase it, and if this gets in first I'll re-enable it in that PR.

MistakeNot4892 (approved):

PR #4759, opened 2025-01-18
Optimize parts of atmos code, plus some misc. optimizations

out-of-phaze:
## Description of changes
Optimizes and modernizes certain parts of atmos code.

Optimizes `_fetch_icon_state_cache_entry()` by using the icon as a key rather than the icon's ref, because there's a 1:1 correspondence. Using a ref is just slower for no benefit. (As far as I can tell it works fine for files, but I guess it might cause issues if you `qdel()` a file object? But I don't think that's even valid, and goes against the idea of wanting to cache this anyway. I think we only softdel them, and I'm pretty sure file() objects are reused under the hood.)

Makes unsim turfs not have alpha masks. I *might* need to adjust this in the future to just be the specific unsim turf subtypes that have no functionality, but I mostly don't want to waste time on alpha masks for these turfs during maploading, and we don't have any unsim turfs that mask their contents anyway.

Performance improvements, mostly relevant for init and any sort of heavy atmos load.

### DIFF
--- a/code/__defines/ZAS.dm
+++ b/code/__defines/ZAS.dm
@@ -57,29 +57,25 @@
 	}
 
 #ifdef MULTIZAS
-
-var/global/list/csrfz_check = list(NORTHEAST, NORTHWEST, SOUTHEAST, SOUTHWEST, NORTHUP, EASTUP, WESTUP, SOUTHUP, NORTHDOWN, EASTDOWN, WESTDOWN, SOUTHDOWN)
-var/global/list/gzn_check = list(NORTH, SOUTH, EAST, WEST, UP, DOWN)
+#define ZAS_CSRFZ_CHECK global.cornerdirsz
+#define ZAS_GZN_CHECK global.cardinalz
 
 #define ATMOS_CANPASS_TURF(ret, A, B) \
 	if (A.blocks_air & AIR_BLOCKED || B.blocks_air & AIR_BLOCKED) { \
 		ret = BLOCKED; \
 	} \
-	else if (B.z != A.z) { \
-		if (B.z < A.z) { \
-			ret = (A.z_flags & ZM_ALLOW_ATMOS) ? ZONE_BLOCKED : BLOCKED; \
-		} \
-		else { \
-			ret = (B.z_flags & ZM_ALLOW_ATMOS) ? ZONE_BLOCKED : BLOCKED; \
-		} \
+	else if (B.z < A.z) { \
+		ret = (A.z_flags & ZM_ALLOW_ATMOS) ? ZONE_BLOCKED : BLOCKED; \
 	} \
-	else if (A.blocks_air & ZONE_BLOCKED || B.blocks_air & ZONE_BLOCKED) { \
-		ret = (A.z == B.z) ? ZONE_BLOCKED : AIR_BLOCKED; \
+	else if(B.z > A.z) { \
+		ret = (B.z_flags & ZM_ALLOW_ATMOS) ? ZONE_BLOCKED : BLOCKED; \
 	} \
-	else if (A.contents.len) { \
+	else if ((A.blocks_air & ZONE_BLOCKED) || (B.blocks_air & ZONE_BLOCKED)) { \
+		ret = ZONE_BLOCKED; \
+	} \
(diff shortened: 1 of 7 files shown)


code/__defines/ZAS.dm
```diff
@@ -58,28 +58,25 @@
 
 #ifdef MULTIZAS
 
-var/global/list/csrfz_check = list(NORTHEAST, NORTHWEST, SOUTHEAST, SOUTHWEST, NORTHUP, EASTUP, WESTUP, SOUTHUP, NORTHDOWN, EASTDOWN, WESTDOWN, SOUTHDOWN)
-var/global/list/gzn_check = list(NORTH, SOUTH, EAST, WEST, UP, DOWN)
+var/global/list/csrfz_check = global.cornerdirsz
```
MistakeNot4892: Should this have a `.Copy()?` Unsure why we need to be declaring new global lists versus just using the existing ones directly.

code/__defines/ZAS.dm
```diff
@@ -58,28 +58,25 @@
...
-		else { \
-			ret = (B.z_flags & ZM_ALLOW_ATMOS) ? ZONE_BLOCKED : BLOCKED; \
-		} \
+	else if (B.z < A.z) { \
+		ret = (A.z_flags & ZM_ALLOW_ATMOS) ? ZONE_BLOCKED : BLOCKED; \
+	} \
+	else if(B.z > A.z) { \
+		ret = (B.z_flags & ZM_ALLOW_ATMOS) ? ZONE_BLOCKED : BLOCKED; \
 	} \
 	else if (A.blocks_air & ZONE_BLOCKED || B.blocks_air & ZONE_BLOCKED) { \
```
MistakeNot4892: Evidently not -required- but this could probably use parens around the bitwise op.

code/__defines/ZAS.dm
```diff
@@ -88,8 +85,8 @@ var/global/list/gzn_check = list(NORTH, SOUTH, EAST, WEST, UP, DOWN)
 	}
 #else
 
-var/global/list/csrfz_check = list(NORTHEAST, NORTHWEST, SOUTHEAST, SOUTHWEST)
-var/global/list/gzn_check = list(NORTH, SOUTH, EAST, WEST)
+var/global/list/csrfz_check = global.cornerdirs
+var/global/list/gzn_check = global.cardinal
```
MistakeNot4892: As above, why do we need these to be global lists? Could it just be a #define if the varname is important?

code/game/objects/item_mob_overlay.dm
```diff
@@ -45,13 +45,15 @@ var/global/list/icon_state_cache = list()
 /proc/_fetch_icon_state_cache_entry(checkicon)
 	if(isnull(checkicon) || !(isfile(checkicon) || isicon(checkicon)))
 		return null
-	var/checkkey = "\ref[checkicon]"
```
MistakeNot4892, line 48: As far as I recall we use \ref because generated icons don't work as keys. Not sure if that is exactly the case but I do think we used \ref for an assoc key reason.

out-of-phaze: if you stringify them, sure, because when stringified they're just `""` or `"/icon"` or whatever. but we aren't stringifying them

code/__defines/ZAS.dm
out-of-phaze: iirc the reason is ordering, global.cornerdirs is—oh wait it expands
yeah i could do that. that said i want to nuke support for not having MULTIZAS so this will hopefully just be able to use `global.cornerdirsz` directly in the future

code/__defines/ZAS.dm
out-of-phaze: i thought it was ordering but i was silly

code/game/objects/item_mob_overlay.dm
MistakeNot4892: That might have been it, yeah. Alrighty.

MistakeNot4892 (changes requested):
Would like the parens thing addressed and the new global lists removed but overall code looks fine.

MistakeNot4892 (approved):

PR #4760, opened 2025-01-18
Improve code quality grep checks

out-of-phaze:
## Description of changes
- Adds a check for the `tag` var in code, expanding it from just checking in maps.
- Removes unnecessary usage of the `tag` var.
- Replaces all current uses of `direct_output` with specialized helpers, in preparation for allowing drop-in optional replacement for downstreams that want to opt into things like rust-g to improve performance.
- Improves a lot of our grep checks to reduce false positives.
- Removes 15 `in world` loops, either deleting them entirely because they're dead code, or replacing them with other lists.
- Fixes redundant code in supply pack setup. This just got thrown in because eh, why not.

Improves code quality across the board.

### DIFF
--- a/code/modules/overmap/contacts/_contacts.dm
+++ b/code/modules/overmap/contacts/_contacts.dm
@@ -29,7 +29,6 @@
 
 	radar = image(loc = effect, icon = 'icons/obj/overmap.dmi', icon_state = "sensor_range")
 	radar.color = source.color
-	radar.tag = "radar"
 	radar.add_filter("blur", 1, list(type = "blur", size = 1))
 	radar.appearance_flags |= RESET_TRANSFORM | KEEP_APART
 	radar.appearance_flags &= ~PIXEL_SCALE
--- a/code/unit_tests/mob_tests.dm
+++ b/code/unit_tests/mob_tests.dm
@@ -60,19 +60,9 @@
 
 var/global/default_mobloc = null
 
-/proc/create_test_mob_with_mind(var/turf/mobloc = null, var/mobtype = /mob/living/human)
+/proc/create_test_mob_with_mind(var/turf/mobloc, var/mobtype = /mob/living/human)
 	var/list/test_result = list("result" = FAILURE, "msg"    = "", "mobref" = null)
 
-	if(isnull(mobloc))
-		if(!default_mobloc)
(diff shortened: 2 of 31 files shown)


code/modules/overmap/contacts/_contacts.dm
```diff
@@ -29,7 +29,6 @@
 
 	radar = image(loc = effect, icon = 'icons/obj/overmap.dmi', icon_state = "sensor_range")
 	radar.color = source.color
-	radar.tag = "radar"
```
MistakeNot4892, line 32: Is this not used in a locate() somewhere?

code/unit_tests/mob_tests.dm
```diff
@@ -60,19 +60,9 @@
 
 var/global/default_mobloc = null
 
-/proc/create_test_mob_with_mind(var/turf/mobloc = null, var/mobtype = /mob/living/human)
+/proc/create_test_mob_with_mind(var/turf/mobloc, var/mobtype = /mob/living/human)
 	var/list/test_result = list("result" = FAILURE, "msg"    = "", "mobref" = null)
 
-	if(isnull(mobloc))
```
MistakeNot4892, line 66: Does this block not need to be replaced?

MistakeNot4892: If you're removing this you also need to remove the global cache var above.

MistakeNot4892 (changes requested):
Default mobloc cache var needs removal or reimplementation.

code/modules/overmap/contacts/_contacts.dm
out-of-phaze: it is not. maybe it's supposed to be, but it isn't

code/unit_tests/mob_tests.dm
out-of-phaze: i'll do that. it's removed bc it violates the premise of unit tests not modifying stuff outside a designated area, it evades the post-test cleanup checks too. if this was a UT proc and not a global proc i could probably simplify it a lot actually

MistakeNot4892:
Unit tests appear to be timing out.

MistakeNot4892 (dismissed):

MistakeNot4892 (approved):

PR #4764, opened 2025-01-19
Staging fixes from Pyrelight test.

MistakeNot4892:
## Description of changes
Commit changes should be descriptive.

Fixes.

## Authorship
Myself.

## Changelog
:cl:
tweak: Rags can be used to wipe down turfs.
/:cl:


code/game/objects/effects/decals/Cleanable/robots.dm
```diff
@@ -44,6 +44,11 @@
 	chemical = /decl/material/liquid/lube
 	cleanable_scent = "industrial lubricant"
 
+/obj/effect/decal/cleanable/blood/oil/Initialize(mapload)
+	if(basecolor)
```
out-of-phaze: this should just be an update icon call at the end of Initialize, in parent /blood/Initialize or even /cleanable/Initialize, tbh. because only setting color will not fix the name being "blood"

out-of-phaze (dismissed):

out-of-phaze (approved):

code/modules/mob/hugs.dm
```diff
@@ -46,7 +46,7 @@ var/global/list/_default_hug_messages = list(
 
 	playsound(loc, 'sound/weapons/thudswoosh.ogg', 50, 1, -1)
 
-	if(src != target)
+	if(src != target && client && key && target.client && target.key && !target.incapacitated())
```
out-of-phaze, line 49: this is technically redundant, client implies key, but a key could be a disconnected player mob. i think just `client` would be fine to avoid it working on ssd and non-player mobs but it's nbd

PR #4766, opened 2025-01-19
Fix miscellaneous code quality and consistency issues

out-of-phaze:
## Description of changes
Removes a lot of spawn() and sleep() abuse in favor of QDEL_IN and timers.
Makes a bunch of things use abstract_type instead of bespoke vars that do the same thing.
Makes attack filters a datum subtype.
Makes the prometheus metrics system use decls.
Fixes some old outdated code comments.
Fixes horse coloration.

Improves code quality and consistency.

### DIFF
--- a/code/controllers/subsystems/garbage.dm
+++ b/code/controllers/subsystems/garbage.dm
@@ -279,13 +279,13 @@ SUBSYSTEM_DEF(garbage)
 	qdel(D, force)
 #endif
 
-// Should be treated as a replacement for the 'del' keyword.
-// Datums passed to this will be given a chance to clean up references to allow the GC to collect them.
+/// Should be treated as a replacement for the 'del' keyword.
+/// Datums passed to this will be given a chance to clean up references to allow the GC to collect them.
+/// Non-datums passed to this will be hard-deleted.
 /proc/qdel(datum/D, force=FALSE)
 	if(isnull(D))
 		return
 	if(!istype(D))
-		PRINT_STACK_TRACE("qdel() can only handle /datum (sub)types, was passed: [log_info_line(D)]")
 		del(D)
 		return
 	var/datum/qdel_item/I = SSgarbage.items[D.type]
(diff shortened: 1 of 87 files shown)


code/controllers/subsystems/garbage.dm
```diff
@@ -281,13 +281,13 @@ SUBSYSTEM_DEF(garbage)
...
-// Should be treated as a replacement for the 'del' keyword.
-// Datums passed to this will be given a chance to clean up references to allow the GC to collect them.
+/// Should be treated as a replacement for the 'del' keyword.
+/// Datums passed to this will be given a chance to clean up references to allow the GC to collect them.
+/// Non-datums passed to this will be hard-deleted.
 /proc/qdel(datum/D, force=FALSE)
 	if(isnull(D))
 		return
 	if(!istype(D))
-		PRINT_STACK_TRACE("qdel() can only handle /datum (sub)types, was passed: [log_info_line(D)]")
```
MistakeNot4892, line 288: Do we not want qdel() on non-instances to error?

out-of-phaze: images :)

MistakeNot4892: I don't understand this response. Images harddel'd prior to this PR so I'm not sure what the purpose of these changes is.

out-of-phaze: ? images still get qdel'd though. you literally added code in the footprints pr that qdel_list'd a list of images. and it should be fine to qdel an image, but this code makes that throw a useless error. istype(/image, /datum) is false, it will throw a pointless error if you try to qdel an image, and if you do this you can't do `QDEL_IN(image, X)`. the only real case where it should error is if we pass a primitive like text or a number, which shouldn't get del'd anyway, and if you want i can add a `CRASH` for that but it seems unnecessary

MistakeNot4892: Thank you for the explanation.
The qdel list in footprints is a mistake (per Discord) - I'm not sure we should be implicitly supporting the qdel of images/lists, since they're better left to the GC. No strong feelings on the matter though. The CRASH on an invalid qdel param has been useful in the past for me however.

MistakeNot4892 (approved):

PR #4788, opened 2025-01-22
Fixing mob electrocution explicitly checking gloves.

MistakeNot4892:
## Description of changes
- Cleans up some single-letter var names in SMES code.
- Adds a proc to get Siemen's coefficient based on coverage slots.
- Collapses some shock/electrocution procs and cleans up the chain.
- electrocute_mob() now grabs the Siemen's coefficient based on coverage of the current active hand slots.

Drakes or baxxid will be able to rely on the Siemen's coefficient of their hardsuit or such rather than explicitly needing to wear gloves even if using their mouth slot.

## Authorship
Myself.

## Changelog
Nothing player-facing.

### DIFF
--- a/code/modules/mob/inventory.dm
+++ b/code/modules/mob/inventory.dm
@@ -405,3 +405,9 @@
 		var/org = GET_EXTERNAL_ORGAN(src, hand_slot)
 		if(org)
 			LAZYDISTINCTADD(., org)
+
+/mob/proc/get_active_hand_bodypart_flags()
+	var/datum/inventory_slot/gripper/inv_slot = get_inventory_slot_datum(get_active_held_item_slot())
+	if(istype(inv_slot))
+		. = inv_slot.covering_slot_flags
+	. ||= SLOT_HANDS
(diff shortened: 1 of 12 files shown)


code/modules/mob/inventory.dm
```diff
@@ -405,3 +405,9 @@
 		var/org = GET_EXTERNAL_ORGAN(src, hand_slot)
 		if(org)
 			LAZYDISTINCTADD(., org)
+
+/mob/proc/get_active_hand_bodypart_flags()
+	var/datum/inventory_slot/gripper/inv_slot = get_inventory_slot_datum(get_active_held_item_slot())
+	if(istype(inv_slot))
+		. = inv_slot.covering_slot_flags
+	. ||= SLOT_HANDS
```
out-of-phaze, line 413: i assume it's intentional that this is "use hands if nothing else is set" rather than "add hands to the flags always"

out-of-phaze (approved):

code/modules/mob/inventory.dm
MistakeNot4892: Yeah.

PR #4789, opened 2025-01-22
Fixes bad istype() in hand slot code.

MistakeNot4892:


comma (dismissed):

out-of-phaze (approved):

code/modules/mob/mob.dm
```diff
@@ -1402,7 +1402,7 @@
 /// THIS DOES NOT RELATE TO HELD ITEM SLOTS. It is very specifically a functional BP_L_HAND or BP_R_HAND organ, not necessarily a gripper.
 /mob/proc/get_usable_hand_slot_organ()
 	var/obj/item/organ/external/paw = GET_EXTERNAL_ORGAN(src, BP_L_HAND)
-	if(!istype(paw) && !paw.is_usable())
+	if(istype(paw) && !paw.is_usable())
```
out-of-phaze: are we sure this shouldn't be
```suggestion
	if(!istype(paw) || !paw.is_usable())
```

out-of-phaze (changes requested):

code/modules/mob/mob.dm
MistakeNot4892: True.

code/modules/mob/mob.dm
```diff
@@ -1401,8 +1401,8 @@
 
 /// THIS DOES NOT RELATE TO HELD ITEM SLOTS. It is very specifically a functional BP_L_HAND or BP_R_HAND organ, not necessarily a gripper.
 /mob/proc/get_usable_hand_slot_organ()
-	var/obj/item/organ/external/paw = GET_EXTERNAL_ORGAN(src, BP_L_HAND)
-	if(!istype(paw) && !paw.is_usable())
-		paw = GET_EXTERNAL_ORGAN(src, BP_R_HAND)
-	if(istype(paw) && paw.is_usable())
-		return paw
+	var/static/list/hand_slots = list(BP_L_HAND, BP_R_HAND)
+	for(var/slot in shuffle(hand_slots))
```
out-of-phaze, line 1405: iirc shuffling is actually relatively expensive, i feel a little uncomfortable about this, but it's probably fine

out-of-phaze (approved):

code/modules/mob/mob.dm
MistakeNot4892: This isn't hot code, it's used only by a couple of specific interactions (there is also a use in get_quick_interaction_handler() but that is an error and is removed in the holsters PR).

PR #4794, opened 2025-01-23
Holster fixes, quick interactions.

### LINKED ISSUE #4776: Can Holster Two Guns in a Two Slot Holster
Upstreaming https://github.com/ScavStation/ScavStation/issues/1214

#### Description of issue

You are able to holster two guns in a holster with only two slots, with both slots already occupied. In this case, I was able to holster a _small energy gun_ while my holster belt already contained a _telescopic baton_ and _energy gun_.
Screenshot 2025-01-16 011042

Holster belts should only be able to holster one gun normally, and another object (usually a melee weapon).

#### Steps to reproduce

1) Have a _holster belt_ equipped with two items in it
2) Find another gun
3) Holster it

holster belt, telescopic baton, energy gun, small energy gun

Within the month of January 2025 (I vaguely recall this issue being present earlier this month but cannot remember when).

Client Version: 515
Server Revision: 9c5446490567e759e414574162a473fbebb70ebf - dev - 2025-01-16
Game ID: cxH-aD64

#### Issue bingo

MistakeNot4892:
- Fixes #4776 properly this time.
- Adds ctrl-click to unholster from accessories and belts.

### DIFF
--- a/code/game/objects/items/weapons/storage/belt.dm
+++ b/code/game/objects/items/weapons/storage/belt.dm
@@ -24,11 +24,11 @@
 	. = ..()
 	if(overlay_flags & BELT_OVERLAY_ITEMS)
 		var/list/cur_overlays
-		for(var/obj/item/I in contents)
-			if(I.use_single_icon)
-				LAZYADD(cur_overlays, I.get_on_belt_overlay())
+		for(var/obj/item/thing in contents)
+			if(thing.use_single_icon)
+				LAZYADD(cur_overlays, thing.get_on_belt_overlay())
 			else
-				LAZYADD(cur_overlays, overlay_image('icons/obj/clothing/obj_belt_overlays.dmi', I.icon_state))
+				LAZYADD(cur_overlays, overlay_image('icons/obj/clothing/obj_belt_overlays.dmi', thing.icon_state))
 
 		if(LAZYLEN(cur_overlays))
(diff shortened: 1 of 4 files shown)


comma (approved):

code/game/objects/items/weapons/storage/belt.dm
```diff
@@ -57,13 +57,6 @@
 	. = ..()
 	set_extension(src, /datum/extension/holster, storage, sound_in, sound_out, can_holster)
 
-/obj/item/belt/holster/get_stored_inventory()
```
out-of-phaze, line 60: wasn't the intent of this to make holstering an item not put it in storage contents? is this a bugfix or are we just deciding we want it to be in both? will removing it via storage contents properly unholster it still or will it teleport when you use the verb/interaction handler/etc?

out-of-phaze (changes requested):

code/game/objects/items/weapons/storage/belt.dm
MistakeNot4892: Taking it out of storage was my previous fix, I misunderstood the holster code. The extension/holster storage datum work in tandem. The actual fix is amending a `loc != storage` to  `storage.holder` in `check_holster()`

out-of-phaze (approved):

PR #4802, opened 2025-01-24
Non-blacksmithy changes from blacksmithy PR.

MistakeNot4892:
- Merges `shake()` and `shake_animation()` at `/atom` level.
- Privates `filter_data` and adds a `has_filter()` proc.
- Adds a delta param to `get_thermal_mass_coefficient()` to allow for faster cooling than heating.
- Generalizes HE pipe glow effects to `/obj`.
- Adds a silent spark subtype.
- Adds a bar recipe to soft materials.
- Adds an iron knife subtype and replaces knives on Shaded Hills with it.
- Adds debug strings to admin examining hot atoms.
- Skips structure hitsound if `silent` is `TRUE`.
- Adds an iron forging hammer.
- Renames the 'meat hook' item to 'hook'.
- Adds mapped iron and copper bar subtypes.


MistakeNot4892:
Will atomise this if requested, but I just tried and got really bored.

out-of-phaze (approved):

PR #4806, opened 2025-01-25
Various changes from Pyrelight.

MistakeNot4892:
Commit list should be descriptive.


MistakeNot4892:
Oop, brainfart, thought this was going to Pyrelight, not from Pyrelight. Tiny changeset at least :(

PR #4810, opened 2025-01-25
Nerfing blob defensive values.

MistakeNot4892:
## Description of changes
Syncs upstream blob with downstream blob. Generally significantly nerfs blob survivability.

My impression is that this blob is more appropriate for the lowpop rounds Neb tends to have. Main purpose is to avoid conflicts with downstream in this file every time I update.

## Authorship
@Qumefox is responsible for the original nerfs I believe.

## Changelog
tweaks: Blob survivability has been heavily nerfed.
/:cl:

### DIFF
--- a/code/modules/blob/blob.dm
+++ b/code/modules/blob/blob.dm
@@ -12,19 +12,19 @@
 
 	layer = BLOB_SHIELD_LAYER
 
-	max_health = 30
+	max_health = 15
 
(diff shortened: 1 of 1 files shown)


out-of-phaze (approved):

out-of-phaze:
fuck blob everyone hates blob kill blob woo

PR #4816, opened 2025-01-25
Couple of small fixes for Scav.

MistakeNot4892:
- Fixes https://github.com/ScavStation/ScavStation/issues/1227
- Fixes https://github.com/ScavStation/ScavStation/issues/1226


out-of-phaze (approved):
should maybe make the implants enumerate available emotes at some point so it can't let you pick one you're not actually able to do, maybe change that on dev?

out-of-phaze (approved):
shockingly the material stacker thing is the only match for the regex `\\(?:[Aa]|[Tt]he) \[([^\]]+)\][^["]*\\(?:[Aa]|[Tt]he) \[\1\]`, pretty neat?

PR #4822, opened 2025-01-28
Adding item-based intent.

MistakeNot4892:
## Description of changes
- Intent list is dynamically generated based on inhand items.
- Intents can prune other intents off the available intent list.
- Intent display can show less than four intents without exploding.

Finishes putting stuff in place for full dynamic intents.

## Authorship
Myself.

## Changelog
Nothing player-facing.


out-of-phaze (dismissed):
deceptively small PR, huh. i assume the purpose of the blocked intents thing is so that a sword's special slashing/cutting intents could block the default harm intent?

out-of-phaze (approved):

PR #4823, opened 2025-01-28
Remove arguments from update_icon overrides

out-of-phaze:
## Description of changes
Saves almost two seconds on tradeship init by:
- removing `arglist()` arguments passing to `on_update_icon()`
- removing an assoc list insertion when using `queue_icon_update()`
- optimizing icon update queue processing on `SSicon_update` to be faster
- making all icon updates prior to the first `SSicon_update` flush queue (I tested this, it is in fact much faster)
- removing an oversafe check in `_fetch_icon_state_cache_entry()` that should honestly just runtime in the icon_states check. It should never be passed a non-icon tbh.

I've yet to test railings, valves, and shield effects but they should work. The other stuff was unused args, or I tested them (lights were definitely tested, and I made sure normal updates for things like walls, floors, and machines worked by uh. Well, I fucked it up on accident and compared it to that to make sure everything had properly updated, lol.)

Prevents a ton of duplicate update_icon() calls in init. Everything is buttery smooth.
Also fixes an issue with rigged lights.

## Authorship
Me.

### DIFF
--- a/code/game/objects/items/weapons/tanks/tanks.dm
+++ b/code/game/objects/items/weapons/tanks/tanks.dm
@@ -35,7 +35,6 @@ var/global/list/global/tank_gauge_cache = list()
 
 	var/gauge_icon = "indicator_tank"
 	var/gauge_cap = 6
-	var/previous_gauge_pressure = null
 
 	var/datum/gas_mixture/air_contents = null
 	var/distribute_pressure = ONE_ATMOSPHERE
@@ -63,7 +62,7 @@ var/global/list/global/tank_gauge_cache = list()
 	air_contents.update_values()
 
 	START_PROCESSING(SSobj, src)
-	update_icon(TRUE)
+	update_icon()
 
 /obj/item/tank/Destroy()
 	QDEL_NULL(air_contents)
@@ -134,7 +133,7 @@ var/global/list/global/tank_gauge_cache = list()
 		if(C.use(1))
 			wired = 1
 			to_chat(user, "<span class='notice'>You attach the wires to the tank.</span>")
-			update_icon(TRUE)
+			update_icon()
 		return TRUE
 
 	if(IS_WIRECUTTER(W))
@@ -157,7 +156,7 @@ var/global/list/global/tank_gauge_cache = list()
 						assy.a_right = null
 						proxyassembly.assembly = null
 						qdel(assy)
-				update_icon(TRUE)
+				update_icon()
 
 			else
 				to_chat(user, "<span class='danger'>You slip and bump the igniter!</span>")
@@ -169,7 +168,7 @@ var/global/list/global/tank_gauge_cache = list()
 			if(do_after(user, 10, src))
 				to_chat(user, "<span class='notice'>You quickly clip the wire from the tank.</span>")
 				wired = 0
-				update_icon(TRUE)
+				update_icon()
 
 		else
 			to_chat(user, "<span class='notice'>There are no wires to cut!</span>")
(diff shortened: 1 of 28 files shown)


out-of-phaze:
Saves about ten seconds on away_sites_testing in CI. Also involves a nontrivial amount of code cleanup.

code/game/objects/items/weapons/tanks/tanks.dm
```diff
@@ -378,10 +377,11 @@ var/global/list/global/tank_gauge_cache = list()
 	air_contents.react()
 	check_status()
 
-/obj/item/tank/on_update_icon(var/override)
+// TODO: Check if this works without the override argument. Everything in tank code seems to call it, so...
+/obj/item/tank/on_update_icon()
 	. = ..()
 	var/list/overlays_to_add
```
MistakeNot4892: This looks like the pre-SSoverlays pattern, probably not much benefit to it now.

code/game/objects/structures/fishtanks.dm
```diff
@@ -129,17 +139,15 @@ var/global/list/global/aquarium_states_and_layers = list(
 				tank_overlay.add_overlay(global.fishtank_cache[cache_key])
 
 	// Update overlays with contents.
+	// TODO: Can this just use vis_contents...?
+	// Or add its contents to vis_contents on some sort of helper atom,
+	// which has the VIS_UNDERLAY flag so it shows up under the transparent part?
 	icon_state = "base"
 	..()
 	for(var/atom/movable/AM in contents)
```
MistakeNot4892: Could be `get_contained_external_atoms()` to skip the simulated check.

code/game/turfs/floors/floor_icon.dm
```diff
@@ -102,25 +102,20 @@
...
 	update_height_appearance() // Also refreshes out base layer.
-	update_floor_icon(update_neighbors)
+	update_floor_icon()
 
 	for(var/image/I in decals)
 		if(I.layer < layer)
 			continue
 		add_overlay(I)
 
-	if(update_neighbors)
```
MistakeNot4892, line 119: Is this replicated elsewhere? Not seeing it in the diff.

code/modules/games/cards.dm
```diff
@@ -337,73 +340,77 @@ var/global/list/card_decks = list()
...
 				I.pixel_y = -6+(offset*i)
 			if(EAST)
 				I.pixel_y = 8-(offset*i)
-			else
+			if(NORTH)
 				I.pixel_x = -7+(offset*i)
+			// other dirs are explicitly unsupported!
 		I.transform = M
-		overlays += I
+		add_overlay(I)
```
MistakeNot4892, line 397: I would suggest a `compile_overlays()` at the end of this one since cards are going to be updating pretty frequently and the flicker will be noticeable.

code/modules/games/cards.dm
```diff
@@ -337,73 +340,77 @@ var/global/list/card_decks = list()
...
 		i++
 
 /obj/item/hand/dropped(mob/user)
 	..()
-	if(locate(/obj/structure/table, loc))
-		src.update_icon(user.dir)
+	if(locate(/obj/structure/table) in loc)
```
MistakeNot4892, line 403: This isn't going to be reliable if the table is deconstructed or the hand moves without entering an inventory first.

code/game/turfs/floors/floor_icon.dm
out-of-phaze: it was never ever ever passed

code/modules/games/cards.dm
out-of-phaze: it doesn't need to update when the table is deconstructed imo, it's mostly important for placement and when it moves into your inventory (so it resets the translation/rotation matrix when in your inventory)

MistakeNot4892 (dismissed):
`compile_overlays()` on cards is the only thing I would really ask you to add, and even that isn't blocking, so.

MistakeNot4892 (approved):

PR #4826, opened 2025-01-30
Reorganize files into module and feature-level folders

out-of-phaze:
## Description of changes
Commit names should be descriptive.
Makes scent stuff not use define strings because it's sort of overkill, nothing checks for them and they're just displayed.
Also cleans up some electric welder code?

Bunch of reorganization and some comments laying out future groundwork for splitting up some of the more obnoxious types (`/obj/structure/bed/chair/bench/pew` being a good example)


MistakeNot4892:
Scent includes out of order apparently.

MistakeNot4892 (dismissed):
real talk I didn't read this diff

PR #4833, opened 2025-02-01
Collapsing organ damage procs down into take_damage().

MistakeNot4892:
## Description of changes
- Merges `take_internal_damage()`, `take_general_damage()` and `take_external_damage()` into `take_damage()`.
- Replaces `damage` on organs with `_organ_damage`, preferring `get_organ_damage()`, `set_organ_damage()` and `adjust_organ_damage()` until someone merges it into `current_health`.
- Renames a bunch of single-letter I/O/E vars appropriate to their context.

Cleaner code.

## Authorship
Myself.

## Changelog
Nothing player-facing.


MistakeNot4892:
Organ damage vars need to get merged with item health at some point but I have not done that in this PR.

out-of-phaze (dismissed):
wait why did we have three separate procs like this anyway

comma (dismissed):

MistakeNot4892:
Should be ready for review now, I think. CI is passing so damage is being passed through appropriately.

out-of-phaze (approved):
a little concerned about adjust_organ_damage and similar stuff on external organs since it seems to be exclusively used on internal organs. i feel like calling it directly on external organs would cause issues bc it'll be overwritten by update_damages? but like. this seems fine.

PR #4835, opened 2025-02-03
Separate beds and chairs out of furniture.dmi, groundwork for eventual de-bedding of chairs

out-of-phaze:
## Description of changes
Removes `base_icon` from beds.
Makes single pews/benches into chairs.
Splits up furniture.dmi.
Moves some grab-buckle logic onto structures.
Makes the travois not a bed.
Adds a description to the travois.

general code quality improvement, plus eventually we will be able to destroy `/obj/structure/bed/chair/bench/pew`.


code/game/objects/structures/defensive_barrier.dm
```diff
@@ -7,7 +7,7 @@
 	throwpass =  TRUE
 	anchored =   TRUE
 	atom_flags = ATOM_FLAG_CLIMBABLE | ATOM_FLAG_CHECKS_BORDER
-	can_buckle = TRUE
+	can_buckle = TRUE // TODO: Is it actually... intended that you can buckle stuff to this?
```
MistakeNot4892, line 10: Believe it's so you can avoid being shoved out of a defensive position

MistakeNot4892 (approved):

PR #4841, opened 2025-02-06
Moving organ damage var/procs onto /internal.

MistakeNot4892:
Rider to address comment on #4833


out-of-phaze (dismissed):

MistakeNot4892:
Push was conflict resolution.

out-of-phaze (approved):

PR #4844, opened 2025-02-09
Adds an industrial centrifuge to replace the honey extractor.

MistakeNot4892:
## Description of changes
Adds a centrifuge to replace the honey extractor. Pulled out of #4266.

Less bespoke machinery, more useful/extensible code.

## Authorship
Myself.

## Changelog
Nothing player-facing beyond the name of the machine, really.


MistakeNot4892:
Depends on #4845

code/game/machinery/centrifuge.dm
```diff
@@ -0,0 +1,183 @@
...
+		update_use_power(POWER_USE_IDLE)
+		return
+
+	var/obj/item/thing = processing_items[1]
+	thing.handle_centrifuge_process(src)
+	if(!QDELETED(thing) && loc)
+		thing.dropInto(loc)
+
+/obj/machinery/centrifuge/Initialize()
+	sound_id = "[work_sound]"
```
out-of-phaze, line 142: why are we stringifying it here? what's the deal with this anyway?

out-of-phaze (approved):

code/game/machinery/centrifuge.dm
MistakeNot4892: Every other usage does it. No idea if it's necessary but it's not exactly hot code.

PR #4846, opened 2025-02-09
Bees need flowers to produce honey.

MistakeNot4892:
## Description of changes
- Some specific seed types generate pollen.
- Pollen is consumed by bees to make honey.

Bees are less magic, prepares the ground for pollinators in beewrite.

## Authorship
Myself.

## Changelog
:cl:
add: Flowers or herbs must be planted near a beehive if you want to make honey.
/:cl:

### DIFF
--- a/code/modules/hydroponics/trays/tray_process.dm
+++ b/code/modules/hydroponics/trays/tray_process.dm
@@ -65,6 +65,9 @@
 			mutate((rand(100) < 15) ? 2 : 1)
 			mutation_level = 0
 
+	if(pollen < 10)
+		pollen += seed?.produces_pollen
+
 	// Maintain tray nutrient and water levels.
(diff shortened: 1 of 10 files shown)


out-of-phaze (approved):

code/modules/hydroponics/trays/tray_process.dm
```diff
@@ -65,6 +65,9 @@
 			mutate((rand(100) < 15) ? 2 : 1)
 			mutation_level = 0
 
+	if(pollen < 10)
```
out-of-phaze, line 68: this can mean it might go above 10 if something ever produces more than 1 pollen per tick, e.g. we have 9.5 pollen and produce 1.5, which gives us 11. but that's not a huge deal tbh

out-of-phaze (approved):

code/modules/hydroponics/trays/tray_process.dm
MistakeNot4892: Pollen consumption takes it by neat 1 values so should be fine.

PR #4856, opened 2025-02-12
Updating robot inventory to use general inventory system.

MistakeNot4892:
## Description of changes
Converts robot inventory to use the general hand slot system.
Converts robot module equipment to be a storage item.

Removes a lot of very old, very bad bespoke inventory code.

## Authorship
Myself

## Changelog
:cl:
tweak: Old browser-based module selection for robots has been removed, use the INV button to open storage like you would a belt/backpack.
/:cl:


code/modules/xenoarcheaology/boulder.dm
```diff
@@ -88,13 +88,14 @@
 
 /obj/structure/boulder/Bumped(AM)
 	. = ..()
-	if(ishuman(AM))
-		var/mob/living/human/H = AM
-		for(var/obj/item/P in H.get_inactive_held_items())
-			if(IS_PICK(P))
-				attackby(P, H)
+	if(isliving(AM))
```
out-of-phaze, line 91: Robots are living mobs, so this means the block starting on line 97 will never run. Maybe it should go isrobot, else if is living? Or nest the isrobot check inside the isliving check?

out-of-phaze (changes requested):

code/modules/xenoarcheaology/boulder.dm
MistakeNot4892: Simplified it down so that boulder walk-mining is the same as natural wall walk-mining.

out-of-phaze (approved):

PR #4862, opened 2025-02-15
Condensing sign macros/procs to fix OD issue.

MistakeNot4892:


code/__defines/math_physics.dm
```diff
@@ -34,5 +34,8 @@
 #define TICKS_IN_DAY 		24*60*60*10
 #define TICKS_IN_SECOND 	10
 
-#define SIMPLE_SIGN(X) ((X) < 0 ? -1 : 1)
-#define SIGN(X)        ((X) ? SIMPLE_SIGN(X) : 0)
+#if DM_VERSION < 516
+#define SIGN(X) ((X) < 0 ? -1 : 1)
```
out-of-phaze: this means SIGN(0) = 1 instead of 0. this might be a breaking change in some spots?

out-of-phaze (dismissed):

code/__defines/math_physics.dm
MistakeNot4892: Fixed.

out-of-phaze (approved):

MistakeNot4892:
Yolomerging this to get it to staging/dev

PR #4868, opened 2025-02-17
Cleaning up many R/L/W single-letter vars.

MistakeNot4892:
Opening now before I keep going and it becomes even more unreviewable.


code/_helpers/lists.dm
```diff
@@ -270,17 +269,17 @@ Checks if a list has the same entries and values as an element of big.
...
+	for (item in target_list)
+		if (!target_list[item])
+			target_list[item] = 1
+		total += target_list[item]
 
 	total = rand(1, total)
-	for (item in L)
-		total -=L [item]
+	for (item in target_list)
+		total -=target_list [item]
```
out-of-phaze:
```suggestion
		total -= target_list[item]
```
ewwwwwwww...

out-of-phaze (dismissed):

code/_helpers/lists.dm
MistakeNot4892: Oh god why does that compile

out-of-phaze (approved):

PR #4869, opened 2025-02-17
Generalising some bodybag code.

MistakeNot4892:


out-of-phaze (commented):
not super hyped about having this use default proc args, i'd prefer if it were a var on the item so subtypes could override it

MistakeNot4892:
> not super hyped about having this use default proc args, i'd prefer if it were a var on the item so subtypes could override it

Done.

out-of-phaze (approved):

PR #4870, opened 2025-02-18
Replaces all instances of W and WT as var names, audits attackby() params.

MistakeNot4892:
- `W` and `WT` are no longer used as var names.
- `attackby()` consistently uses `used_item` as the param name.

Depends on #4868 

Apologies for the diff.


out-of-phaze (approved):
um. i don't know how exactly i'm expected to review this tbh. this is why we have a dev branch i guess, and why we trust the compiler and linter

PR #4872, opened 2025-02-18
Makes Airlock Helpers for buttons actually work

Neerti:
Makes `/obj/abstract/airlock_helper/button` actually set buttons' id_tags instead of doing nothing.
Presumably when I tested the helpers the first time around I had forgotten to wipe `id_tag` off of the buttons on the airlock I was testing it on and assumed the buttons were being set.


MistakeNot4892:
The issue is presumably on staging too as the helpers are on that branch also, could you please target staging?

MistakeNot4892 (approved):

PR #4875, opened 2025-02-19
Optimize event registration and firing

out-of-phaze:
## Description of changes
Optimizes the cleanup of global events.
Dramatically optimizes some of the most expensive events by leveraging the no-global-registration flag, using a separate helper to avoid firing events if no one is listening.

Uh. We may need to manually carry over event_listeners in changeturf, I'm not 100% sure, do tell me one way or the other

Makes events faster... probably. Also removes the three-deep list on events and makes bookkeeping/cleanup a good bit faster.

### DIFF
--- a/code/__defines/observ.dm
+++ b/code/__defines/observ.dm
@@ -1,4 +1,1 @@
-// This also works, and removes the need for the _REPEAT variant, but the linter hates it:
-//   #define RAISE_EVENT(OBS, args...) (GET_DECL(OBS))?.raise_event(args);
-#define RAISE_EVENT(OBS, args...) var/decl/observ/__event = GET_DECL(OBS); __event?.raise_event(args);
-#define RAISE_EVENT_REPEAT(OBS, args...) __event = GET_DECL(OBS); __event?.raise_event(args);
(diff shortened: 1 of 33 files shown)


MistakeNot4892 (dismissed):
Willing to give this a go.

out-of-phaze:
> Willing to give this a go.

was actually waiting to be told if i need to manually copy the var over in changeturf or if it persists automatically, i'm not super sure how new()ing a turf works

MistakeNot4892:
> was actually waiting to be told if i need to manually copy the var over in changeturf or if it persists automatically, i'm not super sure how new()ing a turf works

I have absolutely no idea :D
I doubt it persists automatically though given how much state copying goes on in ChangeTurf

out-of-phaze:
will it reset the approval/merge timer if i push code to copy it? theoretically we don't need to inform anyone but the turf of the change since turf new() doesn't make current turf refs outdated afaik, so we'd just have to copy the turf's lists over.

MistakeNot4892:
I don't think it warrants a timer reset nah.

PR #4876, opened 2025-02-20
Remove or reorganize old global variables

out-of-phaze:
## Description of changes
Makes the global uplink singleton use a decl.
Breaks up global_lists.dm a decent amount, more to come later?
Consolidates or removes some old global variables, many of which were unused or redundant. Others were just moved to more fitting locations.

I kind of hate the global variables folder, it doesn't even make sense. We also have a bunch of them in `__defines` and `_helpers`. This is a step towards making global variables easier to find, and also just towards maintaining old code overall.


code/modules/client/preference_setup/general/02_body.dm
```diff
@@ -122,7 +122,7 @@
 
 	pref.blood_type = sanitize_text(pref.blood_type, initial(pref.blood_type))
 
-	if(!pref.species || !(pref.species in get_playable_species()))
+	if(mob_species.spawn_flags & SPECIES_IS_RESTRICTED)
```
MistakeNot4892: This doesn't seem equivalent? You've replaced a general validation check with a specific flag check.

MistakeNot4892: You also don't update `mob_species` which might be important to other stuff below.

out-of-phaze: oh i forgot the !species bit. but `get_playable_species()` is just all species without `SPECIES_IS_RESTRICTED`, plus the default map species, which we now check to make sure isn't restricted anyway.

out-of-phaze: moved it further up in the proc, good catch.

PR #4877, opened 2025-02-20
Remove a lot of unnecessary uses of initial(name)

out-of-phaze:
## Description of changes
Replaces a bunch of `initial(foo.name)` uses with `atom_info_repository.get_name_for(foo)`.
Replaces a bunch of `initial(reagent.name)` uses with `reagent.use_name`.
Makes category datums use abstract types rather than checking if name is non-null.
Removes the (now-)unused `SSmaterials.all_materials` var, since it's entirely redundant now.

Generally, I think `initial()` should be avoided where unnecessary, and these weren't necessary.

### DIFF
--- a/code/modules/materials/definitions/solids/materials_solid_mundane.dm
+++ b/code/modules/materials/definitions/solids/materials_solid_mundane.dm
@@ -5,7 +5,6 @@
 	ore_name = "slag"
 	ore_desc = "Someone messed up..."
 	ore_icon_overlay = "lump"
-	hidden_from_codex = TRUE
 	reflectiveness = MAT_VALUE_DULL
 	wall_support_value = MAT_VALUE_LIGHT
 	value = 0.1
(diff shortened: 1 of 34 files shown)


code/modules/materials/definitions/solids/materials_solid_mundane.dm
```diff
@@ -5,7 +5,6 @@
 	ore_name = "slag"
 	ore_desc = "Someone messed up..."
 	ore_icon_overlay = "lump"
-	hidden_from_codex = TRUE
```
MistakeNot4892, line 8: Why remove this?

MistakeNot4892 (dismissed):

code/modules/materials/definitions/solids/materials_solid_mundane.dm
out-of-phaze: this lets people see that you can process slag into other materials, + it also caused CI issues with links to dissolved reagent results (slag not having a codex page and all) until i added special handling for that.

i guess technically i could make it generate a codex page but not have it show up in the list, or something? idk

MistakeNot4892 (approved):

PR #4880, opened 2025-02-20
Reagent datum now uses material instances instead of types as keys.

MistakeNot4892:
## Description of changes
Instead of using type keys like `/decl/material/foo`, reagent lists will now use instances from the decls repository as keys.

Less overhead when iterating reagent lists.
Easier to VV reagents.

## Authorship
Myself.

## Changelog
Nothing player-facing.


MistakeNot4892:
Tentatively seems to be working? This _has_ to be broken somewhere fundamentally but wandering around on a local server eating soup and injecting drugs did not seem to break.

out-of-phaze (approved):

PR #4887, opened 2025-02-21
Fix shields becoming frames when damaged

### LINKED ISSUE #4865: Firing emiter at Shield Wall spawns free mashine frame
#### Description of issue

if you set up a wall shield and fire emiter at it, a mashine frame spawns in place of shield. after removing frame and restarting shield it works again.

Image

Image

https://github.com/user-attachments/assets/16e81b56-b2d3-4593-98e8-a784861c7e43

it should either pass trough or simply break the shield

#### Steps to reproduce

set up shield wall, fire emitter at it. Free wire and struts.

noticed today when playing with shields.

Client Version: 515
Server Revision: e3f580aae68ca5aa42d9a2b98ead6e99a7950597 - dev - 2025-02-15
Game ID: cyb-dt6B

#### Issue bingo

- [] Issue could be reproduced in multiple rounds

out-of-phaze:
## Description of changes
Overrides dismantle(), frame_type, and construct_state for both types of shields made by shield generators.
Cleans up related shield code overall, which will make it substantially easier to make them not machines at some point.

Fixes #4865.

### DIFF
--- a/code/modules/shieldgen/emergency_shield.dm
+++ b/code/modules/shieldgen/emergency_shield.dm
@@ -7,8 +7,28 @@
 	opacity = FALSE
 	anchored = TRUE
 	max_health = 200
+	frame_type = null
+	construct_state = /decl/machine_construction/noninteractive
 	var/shield_generate_power = 7500	//how much power we use when regenerating
 	var/shield_idle_power = 1500		//how much power we use when just being sustained.
+
+/obj/machinery/shield/take_damage(amount, damtype, silent)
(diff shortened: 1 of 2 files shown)


MistakeNot4892:
Why are we keeping them as /machinery?

out-of-phaze:
> Why are we keeping them as /machinery?

lazy, this is just a step towards changing it later. also mumble mumble Process()

MistakeNot4892 (approved):

PR #4892, opened 2025-02-21
Fix alien whitelist not filtering displayed species

out-of-phaze:
## Description of changes
Fix the `playables` list being generated but not used.
Fixes being able to select species you can't see via href hacking.
Untested please remind me to test it before you hit approve I'm making this in a rush

Fixes issues with the alien whitelist config option

## Changelog
:cl:
bugfix: Players can no longer view or select species in character creation that they are not whitelisted for, if species whitelists are enabled.
/:cl:

### DIFF
--- a/code/modules/client/preference_setup/background/01_species.dm
+++ b/code/modules/client/preference_setup/background/01_species.dm
@@ -89,6 +89,14 @@
 
 		var/choice = href_list["set_species"]
 		if(choice != pref.species)
+			if(!check_rights(R_ADMIN, 0) && get_config_value(/decl/config/toggle/use_alien_whitelist))
+				var/decl/species/new_species = get_species_by_key(choice)
+				if(!new_species)
+					return TOPIC_REFRESH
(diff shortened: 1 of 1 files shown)


code/modules/client/preference_setup/background/01_species.dm
```diff
@@ -89,6 +89,14 @@
 
 		var/choice = href_list["set_species"]
 		if(choice != pref.species)
+			if(!check_rights(R_ADMIN, 0) && get_config_value(/decl/config/toggle/use_alien_whitelist))
+				var/decl/species/new_species = get_species_by_key(choice)
+				if(!new_species)
+					return TOPIC_REFRESH
```
out-of-phaze, line 95: we use TOPIC_REFRESH here bc we're being charitable and assuming that if they pass invalid data, it's because their UI is stale and the option should no longer be available.

MistakeNot4892 (approved):

PR #4894, opened 2025-02-22
Refactor species to use UID instead of name

out-of-phaze:
## Description of changes
Makes species use uid instead of name.
I tested it in-game and also was pretty thorough with my code analysis, checking all the uses of `name` on species.
~~Aside from one instance where a skeleton stand wielding a surgical saw (no, really) showed up instead of my character in the character setup menu until I triggered a refresh, it was fine.~~ I had to revert the icon update queueing during init to fix that, it turns out it was just the most obvious manifestation of some pretty insidious ordering issues this brought to the surface. Oh well.

Also removes get_root_species_name and associated (unused) subspecies support. Whoever readds it, if they do, really needs to think through how it'd work with all the new systems and such and whether or not bodytypes would be more applicable.

Depends on #4876.

Decouples species name and species UID. Modernises species stuff overall, really. This is also a step towards things like runtime-created random/custom species, by choosing to use UID. (At some point, if we really want to commit to supporting that, we would need to replace all the species istypes with flags or dedicated vars, too.)

### DIFF
--- a/code/game/jobs/job/_job.dm
+++ b/code/game/jobs/job/_job.dm
@@ -235,13 +235,13 @@
 		to_chat(feedback, "<span class='boldannounce'>Wrong rank for [title]. Valid ranks in [prefs.branches[title]] are: [get_ranks(prefs.branches[title])].</span>")
 		return TRUE
 
-	var/decl/species/S = get_species_by_key(prefs.species)
+	var/decl/species/S = prefs.get_species_decl()
 	if(!is_species_allowed(S))
 		to_chat(feedback, "<span class='boldannounce'>Restricted species, [S], for [title].</span>")
 		return TRUE
 
-	if(LAZYACCESS(minimum_character_age, S.get_root_species_name()) && (prefs.get_character_age() < minimum_character_age[S.get_root_species_name()]))
-		to_chat(feedback, "<span class='boldannounce'>Not old enough. Minimum character age is [minimum_character_age[S.get_root_species_name()]].</span>")
+	if(LAZYACCESS(minimum_character_age, S.uid) && (prefs.get_character_age() < minimum_character_age[S.uid]))
+		to_chat(feedback, "<span class='boldannounce'>Not old enough. Minimum character age is [minimum_character_age[S.uid]].</span>")
 		return TRUE
 
 	if(!S.check_background(src, prefs))
@@ -379,17 +379,17 @@
 	if(!SSjobs.job_icons[title])
 		var/mob/living/human/dummy/mannequin/mannequin = get_mannequin("#job_icon")
 		if(mannequin)
-			var/decl/species/mannequin_species = get_species_by_key(global.using_map.default_species)
+			var/decl/species/mannequin_species = decls_repository.get_decl_by_id(global.using_map.default_species)
 			if(!is_species_allowed(mannequin_species))
 				// Don't just default to the first species allowed, pick one at random.
 				for(var/other_species in shuffle(get_playable_species()))
-					var/decl/species/other_species_decl = get_species_by_key(other_species)
+					var/decl/species/other_species_decl = decls_repository.get_decl_by_id(other_species)
 					if(is_species_allowed(other_species_decl))
 						mannequin_species = other_species_decl
 						break
 			if(!is_species_allowed(mannequin_species))
 				PRINT_STACK_TRACE("No allowed species allowed for job [title] ([type]), falling back to default!")
-			mannequin.change_species(mannequin_species.name)
+			mannequin.change_species(mannequin_species.uid)
 			dress_mannequin(mannequin)
 			mannequin.set_dir(SOUTH)
 			var/icon/preview_icon = getFlatIcon(mannequin)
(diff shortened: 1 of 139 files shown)


out-of-phaze:
image
I AM BEING HAUNTED BY AN EVIL SKELETON. NO, REALLY. WHY IS IT PICKING THESE TWO SPRITES FROM SURGERY.DMI. IT'S SO FUCKING MENACING. PLEASE HELP

out-of-phaze:
I fixed the skeleton, it turns out the icon update queue was fucking up ordering. probably wiser to just make the expensive calls queue instead. i fixed the skeleton issue by making the organ update icon call in `get_human_icon_cache_key()` skip the queueing, but then the mob was naked because i guess the worn items hadn't updated. and also the character was bald and blah blah yeah it sucks.

code/game/jobs/server_whitelist.dm
```diff
@@ -63,8 +63,8 @@ var/global/list/alien_whitelist = list()
 			alien_whitelist[row["ckey"]] = list(row["race"])
 	return TRUE
 
-/proc/is_species_whitelisted(mob/M, var/species_name)
-	var/decl/species/S = get_species_by_key(species_name)
+/proc/is_species_whitelisted(mob/M, var/species_uid)
+	var/decl/species/S = decls_repository.get_decl_by_id(species_uid)
```
MistakeNot4892, line 67: Would suggest id_or_var here so you don't break existing whitelists (we should pretend they exist)

code/game/jobs/server_whitelist.dm
```diff
@@ -95,7 +95,7 @@ var/global/list/alien_whitelist = list()
 			return FALSE
 		if(!get_config_value(/decl/config/toggle/use_alien_whitelist) || !(S.spawn_flags & SPECIES_IS_WHITELISTED))
 			return TRUE
-		return whitelist_lookup(S.get_root_species_name(M), M.ckey)
+		return whitelist_lookup(S.uid, M.ckey)
```
MistakeNot4892: This isn't equivalent - we have nothing using it currently, but get_root_species_name() is meant to allow for stuff like being vox whitelisted allowing you to use all vox subspecies. I can reimplement it in a less clunky way if I need it down the track though so not blocking.

code/game/jobs/job/_job.dm
```diff
@@ -235,13 +235,13 @@
...
 
-	var/decl/species/S = get_species_by_key(prefs.species)
+	var/decl/species/S = prefs.get_species_decl()
 	if(!is_species_allowed(S))
 		to_chat(feedback, "<span class='boldannounce'>Restricted species, [S], for [title].</span>")
 		return TRUE
 
-	if(LAZYACCESS(minimum_character_age, S.get_root_species_name()) && (prefs.get_character_age() < minimum_character_age[S.get_root_species_name()]))
-		to_chat(feedback, "<span class='boldannounce'>Not old enough. Minimum character age is [minimum_character_age[S.get_root_species_name()]].</span>")
+	if(LAZYACCESS(minimum_character_age, S.uid) && (prefs.get_character_age() < minimum_character_age[S.uid]))
```
MistakeNot4892, line 243: Similar to below, this was meant to allow setting up requirements on one species that also apply to all subspecies.

code/modules/mob/living/human/examine.dm
```diff
@@ -3,7 +3,7 @@
 	if(!(hideflags & HIDEJUMPSUIT) || !(hideflags & HIDEFACE))
 		var/species_name = "\improper "
 		if(isSynthetic() && species.cyborg_noun)
-			species_name += "[species.cyborg_noun] [species.get_root_species_name(src)]"
```
MistakeNot4892, line 6: This use of get_root_species_name() was meant to be for human subspecies to still show up as 'human' on examine.

mods/species/ascent/machines/ship_machines.dm
```diff
@@ -164,7 +164,7 @@ MANTIDIFY(/obj/item/chems/chem_disp_cartridge, "canister", "chemical storage")
 		return ..()
 	if(ishuman(user))
 		var/mob/living/human/H = user
-		if(!(H.species.name in ALL_ASCENT_SPECIES))
+		if(!istype(H.get_species(), /decl/species/mantid))
```
MistakeNot4892, line 167: Not a fan of this, the general list was there because of non-mantid species (serpentid). I appreciate there's no real reason for it now but this does feel like a regression. :(

mods/species/vox/organs_vox.dm
```diff
@@ -185,7 +185,7 @@
 
 /obj/item/organ/internal/voxstack/get_examine_strings(mob/user, distance, infix, suffix)
 	. = ..()
-	var/user_vox = user.get_species_name() == SPECIES_VOX // TODO use bodytype flags instead so subspecies are included
+	var/user_vox = istype(user.get_species(), /decl/species/vox)
```
MistakeNot4892, line 188: Similar :( to mantids. Maybe I'll add some kind of species category var for this kind of thing.

MistakeNot4892 (changes requested):
Would like a y/n on the whitelist id_or_var comment, otherwise looks fine.

code/game/jobs/server_whitelist.dm
out-of-phaze: oh true it didn't even occur to me. except that's not the correct solution, i think it should actually be `return whitelist_lookup(S.uid, M.ckey) || whitelist_lookup(S.name, M.ckey)` in `is_alien_whitelisted()`

MistakeNot4892 (approved):

PR #4906, opened 2025-02-25
Fixing some overlooked uses of abstract wood.

MistakeNot4892:


code/modules/banners/sign_post.dm
```diff
@@ -16,7 +16,7 @@
 	density             = FALSE
 
 /obj/structure/banner_frame/sign/random/Initialize(ml, _mat, _reinf_mat)
-	material = pick(typesof(/decl/material/solid/organic/wood))
+	material = pick(decls_repository.get_decls_of_subtype_unassociated(/decl/material/solid/organic/wood))
```
out-of-phaze: does this work as an instance rather than a path?
```dm
/obj/item/proc/set_material(var/new_material)
	if(new_material)
		material = GET_DECL(new_material)
```
looks to me like it won't because it'll try to pass a decl to GET_DCL.

out-of-phaze (changes requested):

code/modules/banners/sign_post.dm
MistakeNot4892: Oh I actually meant to not use _unassociated whoops.

out-of-phaze (approved):

PR #4898, opened 2025-02-23
Fix clothing modifier unit test and related issues

out-of-phaze:
## Description of changes
See commit messages for details.

## Why and what will this PR improve
Closes #4897.

## Authorship
Me for the kitbashed sprites, Loaf for the other ones that will hopefully be done soon.

## Changelog
:cl:
add: gave the letterman varsity jacket, NT black leather jacket, and brown suit vest unbuttoned sprites
imageadd: added world sprites for rolled jumpsuits
imageadd: added rolled sprites for jumpsuits that were missing them
bugfix: fixed missing markings on labcoats in certain states
/:cl:

PR #4912, opened 2025-02-27
Remove unused clothing and modpack Bay lore references

out-of-phaze:
## Description of changes
Removes unused clothing.
Moves a bunch of Bay lore stuff into the government modpack.
Moves the government/corporate lawsets into a compat module for the relevant modpacks.
Moves chaplain insignias into the modern Earth modpack to pair with the actual religions selectable in character creation.

Codebase organization, removes items missing sprites in #4898.


code/datums/trading/traders/goods.dm
```diff
@@ -179,7 +179,6 @@
 		/obj/item/clothing/neck                       = TRADER_ALL,
 		/obj/item/clothing/suit/jacket                = TRADER_ALL,
 		/obj/item/clothing/suit/robe                  = TRADER_ALL,
-		/obj/item/clothing/shoes/legbrace             = TRADER_ALL,
```
MistakeNot4892: Nooooo, I like the legbrace. If it's 'unused' then it really just needs a loadout entry.

code/modules/clothing/jumpsuits/misc.dm
```diff
@@ -3,25 +3,10 @@
...
-/obj/item/clothing/jumpsuit/johnny
-	name = "brown jumpsuit"
-	desc = "A label on the inside of the collar reads, 'johnny~~~'."
-	icon = 'icons/clothing/jumpsuits/jumpsuit_johnny.dmi'
-
 /obj/item/clothing/jumpsuit/rainbow
 	name = "rainbow"
 	icon = 'icons/clothing/jumpsuits/jumpsuit_rainbow.dmi'
 
-/obj/item/clothing/jumpsuit/psyche
```
MistakeNot4892: I would favour keeping this for chameleon jumpsuit reasons.

code/modules/clothing/misc/venter.dm
```diff
@@ -1,6 +0,0 @@
-/obj/item/clothing/venter
```
MistakeNot4892: Likewise I'd like to keep this with a loadout entry.

code/modules/clothing/neck/brace.dm
```diff
@@ -1,4 +0,0 @@
-/obj/item/clothing/neck/neckbrace
```
MistakeNot4892: Likewise venter/legbrace. These are fun props for spacer characters.

MistakeNot4892 (changes requested):
I would like to keep the space adapt props and the psychadelic jumpsuit.

out-of-phaze:
> I would like to keep the space adapt props and the psychadelic jumpsuit.

i'm not adding loadout entries for the space adapt props (in this PR), ideally they'd be modpacked so that they wouldn't have to be repeatedly defined on space maps (that or... i guess we could define a tech level for loadout items, but i feel like even if we do that it should be modpacked because i want to make tech level not global at some point)

MistakeNot4892 (approved):
Thanks for replacing those.

PR #4917, opened 2025-03-05
Backported citizenship from the Polaris Neb fork.

MistakeNot4892:
Adds a citizenship background category and detail, originally implemented by @Atermonera on Polaris' Neb fork.


code/modules/backgrounds/background_categories.dm
```diff
@@ -31,8 +31,16 @@
 	old_tag = "home_system"
 	uid = "heritage_cat_homeworld"
 	item_type = /decl/background_detail/location
+	// When citizenship decls are better implemented, remove BACKGROUND_FLAG_CITIZENSHIP from flags.
```
out-of-phaze, line 34: what would need to be changed for this to happen? i remember you mentioning it on discord but it's easy for that to get lost

out-of-phaze (approved):

code/modules/backgrounds/background_categories.dm
MistakeNot4892: Species that currently have lore included would need citizenship decls defined, that's all. BACKGROUND_FLAG_CITIZENSHIP is only used by dog tags and passports.

PR #4938, opened 2025-03-09
Change Tajaran species modpack back to old sprites

Cerebulon:
## Description of changes

- Tajaran species from the Tajaran modpack now use old-style sprites and body types.
- Adjusted old base sprites, markings, hairs etc. to use multiplacative blending and new (since removal) markings systems. Colouration may be somewhat different from before but should allow for a wider range of colours.
- Re-adjusted Tajaran voidsuit sprite overrides and removed equipment offsets to fit the 32x32 format again.
- Tajaran ears are now "ear" accessories so will be correctly hidden by the relevant equipment flags.

(If I could track down the origins of a lot of the voidsuits I would also add the helmet and world sprites but alas there is zero crossover with the current Polaris sprites.)

Longneck taj are now being used for Hnoll and providing the "original" Tajaran sprites/species as an option maximises compatibility with other species sprite sets, other codebase sprites etc.

## Authorship
Code and sprite edits by Cerebulon, ear marking code contributed by Loaf.

## Changelog
:cl:
imageadd: Changed modpack Tajara to use old-style humanoid sprites instead of a unique feline body shape which has since been used by Hnolls.

/:cl:


out-of-phaze (approved):

MistakeNot4892:
Sprite sheet issue is a weird one, works fine for avians. Will investigate post-merge.

PR #4945, opened 2025-03-11
A minor item.dm refactor and also pill bottle labeling.

martin509:
## Description of changes
While implementing pill bottle labeling, I noticed that the property `base_desc` (for storing what the original description is for items whose descriptions are frequently modified) had been implemented in multiple separate locations in the code, and decided I would rather refactor that into the base `_item.dm` class rather than continue to add redundant implementations of the same property, so I did.

You can now write on pill bottles with pens and such. It's an exact copy paste of the code from Polaris and works fine, and the relevant functionality for posters and rings should be unaffected as well. Future items that have descriptions that can change will be able to just reference `base_desc` like they can reference the `base_name` that currently exists.

## Authorship
Myself.

## Changelog
:cl:
add: Writing on pill bottles
refactor: Refactored base_desc into item base class
/:cl:

### DIFF
--- a/code/modules/posters/_poster.dm
+++ b/code/modules/posters/_poster.dm
@@ -130,9 +130,6 @@
 	_base_attack_force = 0
 	material = /decl/material/solid/organic/paper
 
-	///The description for the item/medium without any reference to the design.
-	//var/base_desc = "The poster comes with its own automatic adhesive mechanism, for easy pinning to any vertical surface."
-
 	///Type path to the /decl for the design on this poster. At runtime is changed for a reference to the decl
 	var/decl/poster_design/poster_design
 
--- a/code/modules/posters/_poster.dm
+++ b/code/modules/posters/_poster.dm
@@ -129,14 +129,17 @@
 	icon_state = "rolled_poster"
 	_base_attack_force = 0
 	material = /decl/material/solid/organic/paper
+
 	///The description for the item/medium without any reference to the design.
-	var/base_desc = "The poster comes with its own automatic adhesive mechanism, for easy pinning to any vertical surface."
+	//var/base_desc = "The poster comes with its own automatic adhesive mechanism, for easy pinning to any vertical surface."
+
 	///Type path to the /decl for the design on this poster. At runtime is changed for a reference to the decl
 	var/decl/poster_design/poster_design
 
 /obj/item/poster/Initialize(ml, material_key, var/given_poster_type = null)
 	//Init design
 	base_name ||= name
+	base_desc ||= desc
 	set_design(given_poster_type || poster_design || pick(decls_repository.get_decl_paths_of_subtype(/decl/poster_design)))
 	return ..(ml, material_key)
 
--- a/code/modules/posters/_poster.dm
+++ b/code/modules/posters/_poster.dm
@@ -135,10 +135,9 @@
 
 /obj/item/poster/Initialize(ml, material_key, var/given_poster_type = null)
 	//Init design
-	base_name ||= name
-	base_desc ||= desc
+	. = ..(ml, material_key)
 	set_design(given_poster_type || poster_design || pick(decls_repository.get_decl_paths_of_subtype(/decl/poster_design)))
-	return ..(ml, material_key)
+	return .
 
 /obj/item/poster/proc/set_design(var/decl/poster_design/_design_path)
 	if(ispath(_design_path, /decl))
--- a/code/modules/posters/_poster.dm
+++ b/code/modules/posters/_poster.dm
@@ -137,7 +137,6 @@
 	//Init design
 	. = ..(ml, material_key)
 	set_design(given_poster_type || poster_design || pick(decls_repository.get_decl_paths_of_subtype(/decl/poster_design)))
-	return .
 
 /obj/item/poster/proc/set_design(var/decl/poster_design/_design_path)
 	if(ispath(_design_path, /decl))
--- a/code/modules/reagents/storage/pill_bottle.dm
+++ b/code/modules/reagents/storage/pill_bottle.dm
@@ -66,24 +66,11 @@
 	if(wrapper_color)
 		add_overlay(overlay_image(icon, "pillbottle_wrap", wrapper_color, RESET_COLOR))
 
-/obj/item/pill_bottle/proc/update_name_label()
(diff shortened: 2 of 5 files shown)


code/modules/reagents/storage/pill_bottle.dm
```diff
@@ -64,4 +64,32 @@
...
+		return
+	else if(length(labeled_name) > 10)
+		var/short_label_name = copytext(labeled_name, 1, 11)
+		name = "[base_name] ([short_label_name]...)"
+	else
+		name = "[base_name] ([labeled_name])"
+	desc = "[base_desc] It is labeled \"[labeled_name]\"."
+
+/obj/item/pill_bottle/attackby(obj/item/W, mob/user)
+	if(istype(W, /obj/item/pen) || istype(W, /obj/item/flashlight/pen))
```
out-of-phaze:
```suggestion
/obj/item/pill_bottle/attackby(obj/item/used_item, mob/user)
	if(IS_PEN(used_item))
```
we don't allow single-letter variable names in new code, sorry. also `IS_PEN()` is preferable to this typecheck

code/modules/reagents/storage/pill_bottle.dm
```diff
@@ -64,4 +64,32 @@
...
+		if(length(tmp_label) > 50)
+			to_chat(user, "<span class='notice'>The label can be at most 50 characters long.</span>")
+		else if(length(tmp_label) > 10)
+			to_chat(user, "<span class='notice'>You set the label.</span>")
+			labeled_name = tmp_label
+			update_name_label()
+		else
+			to_chat(user, "<span class='notice'>You set the label to \"[tmp_label]\".</span>")
+			labeled_name = tmp_label
+			update_name_label()
```
out-of-phaze: this should probably adapt paper code a little to consume pen charges for ink quills and crayons. also, not a fan of the duplication here

code/modules/reagents/storage/pill_bottle.dm
```diff
@@ -64,4 +64,32 @@
...
+	if(!labeled_name)
+		name = base_name
+		desc = base_desc
+		return
+	else if(length(labeled_name) > 10)
+		var/short_label_name = copytext(labeled_name, 1, 11)
+		name = "[base_name] ([short_label_name]...)"
+	else
+		name = "[base_name] ([labeled_name])"
+	desc = "[base_desc] It is labeled \"[labeled_name]\"."
```
out-of-phaze: honestly i think this should probably use some kind of `/datum/extension/label` subtype. might even be able to get away with just giving pill bottles `/datum/extension/labels/single` and then having an attackby override that just sets the label text when you use a pen

out-of-phaze (changes requested):
some nitpicks on the pill bottle stuff—style guidelines and also i think a bespoke label var with aggressive truncation is less preferable than using the label extension. open to changing my mind on it though

code/modules/posters/_poster.dm
```diff
@@ -129,14 +129,14 @@
...
-	///The description for the item/medium without any reference to the design.
-	var/base_desc = "The poster comes with its own automatic adhesive mechanism, for easy pinning to any vertical surface."
+
 	///Type path to the /decl for the design on this poster. At runtime is changed for a reference to the decl
 	var/decl/poster_design/poster_design
 
 /obj/item/poster/Initialize(ml, material_key, var/given_poster_type = null)
 	//Init design
 	base_name ||= name
+	base_desc ||= desc
```
MistakeNot4892: Believe this and the above are redundant in view of the ..() call, no?

EDIT: If set_design() was moved after the parent call, I think it would be fine, but base_name and base_desc are both used in that proc, so feel free to ignore this comment.

code/modules/reagents/storage/pill_bottle.dm
```diff
@@ -64,4 +64,32 @@
...
+
+/obj/item/pill_bottle/attackby(obj/item/W, mob/user)
+	if(istype(W, /obj/item/pen) || istype(W, /obj/item/flashlight/pen))
+		var/tmp_label = sanitize_safe(input(user, "Enter a label for [name]", "Label", labeled_name), MAX_NAME_LEN)
+		if(length(tmp_label) > 50)
+			to_chat(user, "<span class='notice'>The label can be at most 50 characters long.</span>")
+		else if(length(tmp_label) > 10)
+			to_chat(user, "<span class='notice'>You set the label.</span>")
+			labeled_name = tmp_label
```
MistakeNot4892: As Penny has commented, I think you'd be better served with calling `attach_label(null, null, labeled_name)` like it does in Initialize() rather than a parallel implementation.

martin509:
Changed up a few things with the pen labeling, I decided against using `attach_label` since the normal behavior for pill labels is to just continually rewrite the same label. I've kept the label var stuff for now until I genericize it with `/datum/extension/label` but other than that have done my best to make it fit the style issues you brought up.

code/modules/posters/_poster.dm
```diff
@@ -129,16 +129,15 @@
...
 	///Type path to the /decl for the design on this poster. At runtime is changed for a reference to the decl
 	var/decl/poster_design/poster_design
 
 /obj/item/poster/Initialize(ml, material_key, var/given_poster_type = null)
 	//Init design
-	base_name ||= name
+	. = ..(ml, material_key)
 	set_design(given_poster_type || poster_design || pick(decls_repository.get_decl_paths_of_subtype(/decl/poster_design)))
-	return ..(ml, material_key)
+	return .
```
MistakeNot4892: This is redundant, setting `.` returns automatically at the end of the proc.

MistakeNot4892 (commented):
Not a fan of duplicating the labelling behavior over using the extension but I appreciate it would need some extra work to function as desired. If you could snip off the `return .` I'll approve this.

martin509:
> Not a fan of duplicating the labelling behavior over using the extension but I appreciate it would need some extra work to function as desired. If you could snip off the `return .` I'll approve this.

Done. I think the approach for using the extension would be to add a subtype of label extension that behaves similarly to `extension/labels/single` but allows replacing the label, from what I understand looking at the relevant code.

MistakeNot4892 (dismissed):

out-of-phaze (approved):

PR #4950, opened 2025-03-12
Fix savefile, footprints, and ambient lighting bugs

out-of-phaze:
## Description of changes
- Fixes a typo in legacy savefile loading that prevented species (saved as name) being loaded properly.
- Fix footprints breaking from being passed instances instead of typepaths.
- Partially reverts a change to lighting code done in commit e1a1d93d0594d26f21e6712a11beb28b1b121aec, and re-enables some commented code that existed prior to that commit. Now the code is on par with CitRP's ambient lights prior to ambience groups.

- Saves made prior to the species UID migration now load species properly.
- Muddy footprints no longer runtime out the ass.
- Random tiles on Shaded Hills are no longer missing lighting corners.

### DIFF
--- a/code/modules/lighting/lighting_turf.dm
+++ b/code/modules/lighting/lighting_turf.dm
@@ -64,15 +64,8 @@
 	if (abs(ambient_r + ambient_g + ambient_b) == 0)
 		return
 
-	// Unlit turfs will have corners if they have a lit neighbor -- don't generate corners for them, but do update them if they're there.
-	if (!corners)
-		var/force_build_corners = FALSE
-		for (var/turf/T as anything in RANGE_TURFS(src, 1))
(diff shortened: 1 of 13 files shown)


MistakeNot4892 (dismissed):

MistakeNot4892 (approved):

code/modules/lighting/lighting_turf.dm
```diff
@@ -48,7 +46,7 @@
 	var/ambient_g = 0
 	var/ambient_b = 0
 
-	if (ambient_light)
+	if (ambient_light && ambient_light_multiplier) // If either of these are false-y we can use the simplier path and avoid calculations
```
Lohikar, line 49: It's technically undefined behavior for ALM to be zero, but I don't have a problem with this change otherwise.

Issue #4953, opened 2025-03-13
Biogenerator (and maybe other machines?) can't be assembled/repaired


Typhin:
#### Description of issue
Biogenerators cannot be assembled because once the machine is constructed (Screwdriver on Circuitboard), the machine tries to accept any parts or tools as being fed into the plant hopper, instead of allowing interactions such as installing parts or closing the panel.  This might affect other machines as well, but I haven't dug into exactly which are affected.  (I think there was a similar issue with hydroponics trays demanding a beaker, but trying to install a beaker caused the player to pour the beaker contents into the tray instead?)

Similarly, the machine could not be repaired, as trying to open the panel would instead give an error about the machine not being able to accept a screwdriver.

#### Difference between expected and actual behavior
A machine with an open panel should be able to be maintained/repaired/built.  A machine with a closed panel should open the panel when a screwdriver is used on it.

A check to see if the panel is open or if the item being used is a screwdriver, and passing that interaction up to the base class to handle it, allows these interactions to proceed as expected.

#### Steps to reproduce
Start creating a biogenerator: Use struts to make a machine frame, wrench it, wire it, place a circuit board into it, and use a screwdriver.
Try to install the required parts (power, screen, input, manip, bin).
Be told the machine cannot accept them for processing.

Find an assembled biogenerator
Use a screwdriver on it to try to open the panel, such as to replace a damaged/destroyed component.
Be told the machine cannot accept the screwdriver for processing.

#### Specific information for locating

/code/game/machinery/biogenerator.dm:82

#### Length of time in which bug has been known to occur

I think I ran into this once forever ago when I was still new, but this was something another player ran into today, so I figured I would try to sort it out.

#### Client version, Server revision & Game ID

I forgot to grab this, oops.  Most current Scav Station revision, also verified it's on Nebula Testing.
Client Version: 516
Server Revision: 445f963bea45ed104f26af7e4d34bc76ff575a4e - staging - 2025-03-12
Game ID: cyA-dyvs
Current map: Tradepost Zebra

#### Issue bingo

- [x] Issue could be reproduced at least once
- [x] Issue could be reproduced by different players
- [x] Issue could be reproduced in multiple rounds
- [x] Issue happened in a recent (less than 7 days ago) round
- [x] [Couldn't find an existing issue about this](https://github.com/NebulaSS13/Nebula/issues)

PR #4954, opened 2025-03-13
Fix for Biogenerator issue #4953

Typhin:
Fixes biogenerators not allowing parts/tools to build/repair as mentioned in issue #4953

## Description of changes

Adds a check at the start of `/obj/machinery/biogenerator/attackby(var/obj/item/used_item, var/mob/user)` to allow normal maintenance to be done while the panel is open, and the screwdriver to open the panel when it is closed.

```
	if(panel_open || IS_SCREWDRIVER(used_item))
		return ..()
```

Currently, biogenerators cannot be built or repaired.  If a mapped one is damaged, it's dead for the rest of the round.  If a new one is desired, there is no way to construct it.  That is not great, so this makes biogenerators act properly.

## Authorship

Typhin

## Changelog
:cl:
bugfix: Allows parts/screwdrivers to be used on biogenerators
code: Adds a check for if panel is open, or if item used is a screwdriver
/:cl:

### DIFF
--- a/code/game/machinery/biogenerator.dm
+++ b/code/game/machinery/biogenerator.dm
@@ -81,6 +81,9 @@
 
 /obj/machinery/biogenerator/attackby(var/obj/item/used_item, var/mob/user)
 
+	if(panel_open || IS_SCREWDRIVER(used_item))
+		return ..()
+
 	if(processing)
 		if((. = component_attackby(used_item, user)))
 			return


MistakeNot4892 (approved):

out-of-phaze (approved):
a more thorough fix would probably move `if(!istype(used_item, /obj/item/food/grown))` or change the logic there, but yeah this is good for now

PR #4965, opened 2025-03-23
Add breath-holding inspired by the Scavstation feature

out-of-phaze:
## Description of changes
"Ports" breath-holding from Scavstation, except I completely rewrote it from scratch on accident.
Includes some foundational work that will need to be moved into other PRs on different branches.

TODO: Make holding your breath in vacuum a bad idea?

Prevents Scavstation merge skew.

## Authorship
Scavstation for the idea, but I rewrote almost every single line of code for the breath-holding feature in order to modpack it.

## Changelog
:cl:
add: Added a breath-holding modpack, which adds a 'Hold Breath' verb.
/:cl:

### DIFF
--- a/maps/tradeship/tradeship.dm
+++ b/maps/tradeship/tradeship.dm
@@ -11,11 +11,12 @@
 	#include "../../mods/content/government/away_sites/icarus/icarus.dm"
 	#include "../../mods/content/corporate/away_sites/lar_maria/lar_maria.dm"
 
-	#include "../../mods/content/beekeeping/_beekeeping.dme"
 	#include "../../mods/content/mundane.dm"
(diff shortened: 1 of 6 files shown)


out-of-phaze:
at some point I'd like to add automatic breath holding if you fall into water with high enough athletics (or some other) skill, but it's too fiddly to design for it to get into this PR

MistakeNot4892 (approved):
My main beefs with the Scavstation implementation other than the generally quite convoluted tracking var were that it didn't use the lungs organ, and that there was no UI element for it. This fixes the lungs issue and per discussion on Discord action buttons don't really work for the UI element.

I would still possibly like something under the internals button, or possibly a modifier that will cease holding breath when clicked, but I won't block over them.

PR #4971, opened 2025-03-26
Centralize flash behavior and add flash burn support.

martin509:
## Description of changes

While poking around for a port of Polaris' Zaddat race, I noticed that the race has a feature where they take mild burn damage to the head from being flashed. This lead me to find out that currently, the functionality for handheld flashes, mech-mounted flashes and static flashers is implemented separately (and slightly differently) for each one. This PR consolidates them all to call `mob/living/handle_flashed()` instead, and adds support for a bodytype-specific `eye_flash_burn` modifier (default 0). The new behavior isn't a 100% 1:1 match of how the flasher and mech-flash function currently, but I don't really see a reason for those behaviors to be significantly different from a more powerful regular flash anyways.
Also fixes portable flashers having the same sprite offset as wall flashers.

Future modification of flash behavior will be easily unified since it's all in one place now. In addition, bodytypes with eyes that take burn damage from flashes are made possible.

## Authorship

Myself.

## Changelog
:cl:
add: added support for burn damage from flashes
refactor: centralized flash functionality
bugfix: fixed portable flasher sprite offset
/:cl:

### DIFF
--- a/code/modules/bodytype/_bodytype.dm
+++ b/code/modules/bodytype/_bodytype.dm
@@ -239,6 +239,8 @@ var/global/list/bodytypes_by_category = list()
 	var/eye_blend                   = ICON_ADD
 	/// Stun from blindness modifier.
 	var/eye_flash_mod               = 1
+	//how much damage to take from being flashed (if any)
+	var/eye_flash_burn              = 0
 
 	// Bodytype temperature damage thresholds.
 	/// Cold damage level 1 below this point. -30 Celsium degrees
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -869,6 +869,14 @@ default behaviour is:
 			return I.get_flash_mod()
 	return get_bodytype()?.eye_flash_mod
 
+/mob/living/proc/get_flash_burn()
+	var/vision_organ_tag = get_vision_organ_tag()
+	if(vision_organ_tag)
+		var/obj/item/organ/internal/eyes/I = get_organ(vision_organ_tag, /obj/item/organ/internal/eyes)
+		if(I)
+			return I.get_flash_burn()
+	return get_bodytype()?.eye_flash_burn
+
 /mob/living/proc/eyecheck()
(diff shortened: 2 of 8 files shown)


code/modules/mob/living/living.dm
```diff
@@ -1689,20 +1697,28 @@ default behaviour is:
 		return range * range - 0.333
 	return range
 
-/mob/living/handle_flashed(var/flash_strength)
+/mob/living/handle_flashed(var/flash_strength, do_stun = TRUE)
 
 	var/safety = eyecheck()
-	if(safety >= FLASH_PROTECTION_MODERATE || flash_strength <= 0) // May be modified by human proc.
+	var/flash_burn = get_flash_burn()
+	var/flash_mod = get_flash_mod()
```
comma, line 1704: You get flash_mod but nothing is modified by it, is that intended?

martin509: Not intended, I originally was using it but realized that `mob/living/human/handle_flash()` was using it.

comma (approved):

code/modules/bodytype/_bodytype.dm
```diff
@@ -239,6 +239,8 @@ var/global/list/bodytypes_by_category = list()
 	var/eye_blend                   = ICON_ADD
 	/// Stun from blindness modifier.
 	var/eye_flash_mod               = 1
+	//how much damage to take from being flashed (if any)
+	var/eye_flash_burn              = 0
```
out-of-phaze, line 243: i kind of feel like this could be a bodytype on-flashed proc rather than a dedicated var, but i'm fine with this for now unless anyone else agrees enough to object. leaving this as a comment rather than an approval for discussion

martin509: My view is that you probably want it to have a tweakable multiplier anyways.

MistakeNot4892: Pretty easy to refactor into a proc down the track if needed/desired.

PR #4972, opened 2025-03-27
Add linting for each modpack separately

out-of-phaze:
## Description of changes
Adds a linting step that makes sure each modpack works with just the base game. Should prevent issues like one modpack tacitly requiring the other without that being a planned dependency.

Depends on #4973.

I think we ran into this a while back but I forget what it was for, something like the fantasy modpack not working without drakes included?

## Authorship
Me


out-of-phaze:
i tested and made sure both failures in modpacks and in the base DME cause dreamchecker to fail, and also we don't waste time running the modpack checks if the base DME fails. hooray

MistakeNot4892 (approved):

PR #4976, opened 2025-03-29
Fix a number of issues on the dev branch

liz-lavenza:
## Description of changes
The sharpened item effect now returns the base damage times the multiplier, rather than just the multiplier.
Sets default `icon_state` for various objects that lacked it.
Sets default `color` for objects that use material colour but didn't have a default colour set.
Wood and grass tiles now set their colour from material rather than an arbitrary value. (This will change how grass looks.)
Makes mob modifier icons take the color and alpha of the selected UI theme.
Makes the stamina bar use its preset icon rather than the UI theme icon.
Makes item attackby only return early for whetstones if sharpening actually takes place.
Resets offsets in initialisation for premapped laid-out doormats.
Fixes expend_attack_force returning null/zero if a sharpened item effect exists on the item.
Adds wattle walls to smoothing lists for wood and stone materials.
Adds a check if materials are forgeable before turning them into billets on the forge/anvil.
Rebuilds hand overlays after a new hand slot is added.
Unsets the default base colour for horses so that they properly take random colours.
Makes large cask racks use the name prefix system.
Makes banners/signs show their decals as standalone items.
Sets the reverb environment in the foresters' hut on Shaded Hills.

Fixes the trim of wooden chests disappearing/changing colour when you open them.
Fixes sharpened items returning their damage multiplier as a raw damage value rather than multiplying their base damage.
Makes various objects have mapping preview icon states and colours.
Makes wood and grass tiles use their material colour.
Makes mob modifier screen icons use the correct color and alpha.
Makes the stamina bar show up (before, it had the wrong icon and would not display).
Allows whetstones to be put into storage items.
Fixes premapped laid-out doormats having random offsets.
Fixes sharpened items dealing zero damage on use.
Prevents forging dirt bars or making dirt billets.
Fixes hand slots lacking overlays after a new hand slot is added.
Fixes horses all being the same colour.
Fixes the foresters' hut having way too much reverb.


code/game/objects/items/weapons/material/kitchen.dm
```diff
@@ -1,6 +1,8 @@
 /obj/item/kitchen
```
MistakeNot4892: Didn't this type get removed recently?

code/modules/materials/definitions/solids/materials_solid_stone.dm
```diff
@@ -14,7 +14,10 @@
 		'icons/turf/walls/solid.dmi' = TRUE,
 		'icons/turf/walls/wood.dmi' = TRUE,
 		'icons/turf/walls/brick.dmi' = TRUE,
+		'icons/turf/walls/brick.dmi' = TRUE,
```
MistakeNot4892: Duplicate entry?

MistakeNot4892 (changes requested):
Duplicate entry needs to be removed.

code/modules/banners/__banner.dm
```diff
@@ -84,6 +84,24 @@ var/global/list/banner_type_to_symbols = list()
 	if(.)
 		return english_list(.)
 
+/obj/item/banner/on_update_icon()
+	. = ..()
+	var/image/I
+	for(var/decl/banner_symbol/decal as anything in decals)
```
out-of-phaze: this doesn't make sense for banners, only for signs. also the sign overlays need to be offset on the item, `world-hanging` is like 8px lower than `world`

out-of-phaze (dismissed):

liz-lavenza:
I removed the banner decal commit because I didn't want to bother adding a system for per-state pixel offsets.

MistakeNot4892 (approved):

PR #4978, opened 2025-03-29
Fix issues interacting with target stakes/dummies

liz-lavenza:
## Description of changes
Makes target stakes/dummies use vis_contents for display. They... don't seem to inherit atom colour, I don't know what that was about. ~~If they did, then trays wouldn't work.~~ Oh, I guess items in trays have `RESET_COLOR`, so I should probably add that to target dummies...
Cleans up target stake attackby so that hitting it hits the dummy too. This was added before the vis_contents change and can probably be removed in light of that?
Makes setting `dummy` use a `set_dummy()` proc that centralises various updates.

I couldn't actually hit a target dummy when it was on a stake...

### DIFF
--- a/code/game/objects/structures/target_stake.dm
+++ b/code/game/objects/structures/target_stake.dm
@@ -11,15 +11,22 @@
 	structure_flags     = STRUCTURE_FLAG_THROWN_DAMAGE
 	var/obj/item/training_dummy/dummy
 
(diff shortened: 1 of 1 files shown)


MistakeNot4892 (dismissed):

out-of-phaze (dismissed):
target dummies need RESET_COLOR for vis_contents to not affect their color iirc

PR #4979, opened 2025-03-29
Update colour code to use modern helpers

liz-lavenza:
## Description of changes
Replaces some outdated icon/colour helper code with modern builtins and better-written code.
HSV colour strings are no more, instead they use HSV colour lists. This prevents unnecessary string tree operations and also allows it to work with the built-in colour procs.

Faster, better-written colour helper code.


liz-lavenza:
Dreamchecker needs to be updated to 1.9.

MistakeNot4892 (approved):

PR #4981, opened 2025-03-29
Only show character creation blurb expand buttons if the blurb was cut off

liz-lavenza:
## Description of changes
image
The expand buttons are only shown for background blurbs that get cut off.

It felt weird for them to be always-visible.


MistakeNot4892:
Seems to have the same commit twice?

liz-lavenza:
> Seems to have the same commit twice?

No. They're separate, one is for species, one is for culture.

MistakeNot4892 (approved):

PR #4989, opened 2025-04-20
reinf window fix

markoatonc:
WD.set_anchored(TRUE) // handles setting construction state for us // was orinigally FALSE but it caused bugs wth reinforced window which cannot then be pried out of frame

## Description of changes

Changed WD.set_anchored(FALSE) to WD.set_anchored(TRUE) because that state was causing issues with reinforced window not being able to be pried out of frame or deconstructed, being loked in wrong state. Since Window placed by crafting in hand is already achored normally this bring it into the same state and fixes issue. Tested it fixing both full tile and thin reinforced windows 

the bug has been plaguing our engineer players for long time now, forcing us to climb onto walls and built it there by hand, also made drones unable to install windows due to not being able to craft "in hand" this should solve it. 
## Authorship

myself
## Changelog
:cl:

bugfix: Fixed reinforced window locked in wrong state when built directly on turf
/:cl:

### DIFF
--- a/code/game/objects/structures/window.dm
+++ b/code/game/objects/structures/window.dm
@@ -638,8 +638,8 @@
 
 		if (ST.use(required_amount))
 			var/obj/structure/window/WD = new(loc, ST.material.type, ST.reinf_material?.type, dir_to_set, FALSE)
-			to_chat(user, SPAN_NOTICE("You place [WD]."))
-			WD.set_anchored(FALSE) // handles setting construction state for us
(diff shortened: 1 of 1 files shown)


markoatonc:
fixes 
https://github.com/ScavStation/ScavStation/issues/1265#issue-2986977099
the issue was not pushed upstream but it is that issue

markoatonc:
closed. wrong branch on my end

Issue #4929, opened 2025-03-05
Nascent Diona Gesalts running into other non-diona mobs causes player movement to cease for entire server


MistakeNot4892:
Upstreaming https://github.com/ScavStation/ScavStation/issues/1237

#### Description of issue

When a nascent diona gesalt forms and tries to move to a tile that is occupied by another mob that isn't a diona nymph, the server Master Controller will stop firing ticks.

#### Difference between expected and actual behavior

I would expect a nascent diona gesalt to simply pass the mob like how most mob movement works, instead of crawling the server to a halt.

#### Steps to reproduce

So, what I did:
1) Inhabit a diona nymph as a ghost
2) Click another diona nymph on harm intent to become a nascent diona gesalt
3) Run into another mob that is not a diona nymph

#### Specific information for locating

nascent diona gesalt, diona nymph

Notice: DEFCON 3. The Master Controller has not fired in the last 40 ticks.
Warning: DEFCON 2. The Master Controller has not fired in the last 60 ticks. Automatic restart in 20 ticks.
Warning: DEFCON 1. The Master Controller has still not fired within the last 80 ticks. Killing and restarting...
MC restarted successfully
Master starting processing

#### Length of time in which bug has been known to occur

As of February 17th, 2025

#### Client version, Server revision & Game ID

Client Version: 515
Server Revision: e3f580aae68ca5aa42d9a2b98ead6e99a7950597 - dev - 2025-02-17
Game ID: cyd-dsp7
Current map: Tradepost Mollusc

#### Issue bingo

- [x] Issue could be reproduced at least once
- [x] Issue could be reproduced by different players
- [x] Issue could be reproduced in multiple rounds
- [x] Issue happened in a recent (less than 7 days ago) round
- [x] [Couldn't find an existing issue about this](https://github.com/ScavStation/ScavStation/issues)

MistakeNot4892:
Guessing this is some kind of bump loop as the nascent gestalt tries to move into the target turf regardless of the success of rolling someone up.

PR #4990, opened 2025-04-20
Diona collision bug workaround

markoatonc:
## Description of changes

Changed the roll_up_atom() to return TRUE or FALSE depending of if it succeeded or not, and changed the Bump() proc to check for the success before trying to move the gestalt, so it doesn't get stuck in a loop when it runs into a mob it cannot absorb. It is a workaround until Mistake or someone properly addresses it. 

Fixing miscellaneous bugs
## Authorship

myself 
## Changelog
:cl:

bugfix: Fixed the issue causing a loop freezing up the game when running into mobs as Diona Gestalt. https://github.com/NebulaSS13/Nebula/issues/4929
/:cl:

### DIFF
--- a/code/game/objects/structures/window.dm
+++ b/code/game/objects/structures/window.dm
@@ -17,6 +17,7 @@
 	handle_generic_blending = TRUE
 	hitsound = 'sound/effects/Glasshit.ogg'
 	max_health = 100
+	material_alteration = MAT_FLAG_ALTERATION_ALL
 
 	var/damage_per_fire_tick = 2 		// Amount of damage per fire tick. Regular windows are not fireproof so they might as well break quickly.
 	var/const/CONSTRUCTION_STATE_NO_FRAME = 0
--- a/code/game/objects/structures/window.dm
+++ b/code/game/objects/structures/window.dm
@@ -640,7 +640,7 @@
 		if (ST.use(required_amount))
 			var/obj/structure/window/WD = new(loc, ST.material.type, ST.reinf_material?.type, dir_to_set, FALSE)
 			to_chat(user, SPAN_NOTICE("You place [WD]."))
-			WD.set_anchored(FALSE) // handles setting construction state for us
+			WD.set_anchored(TRUE) // handles setting construction state for us
 		else
 			to_chat(user, SPAN_NOTICE("You do not have enough sheets."))
 			return
--- a/code/game/objects/structures/window.dm
+++ b/code/game/objects/structures/window.dm
@@ -17,7 +17,6 @@
 	handle_generic_blending = TRUE
 	hitsound = 'sound/effects/Glasshit.ogg'
 	max_health = 100
-	material_alteration = MAT_FLAG_ALTERATION_ALL
 
 	var/damage_per_fire_tick = 2 		// Amount of damage per fire tick. Regular windows are not fireproof so they might as well break quickly.
 	var/const/CONSTRUCTION_STATE_NO_FRAME = 0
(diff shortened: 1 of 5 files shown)


markoatonc:
updated with additional fixes

code/modules/materials/definitions/solids/materials_solid_glass.dm
```diff
@@ -39,7 +39,7 @@
 	codex_name = null
 	uid = "solid_borosilicate_glass"
 	lore_text = "An extremely heat-resistant form of glass."
-	flags = MAT_FLAG_BRITTLE
+	flags = MAT_FLAG_BRITTLE | MAT_FLAG_ALTERATION_ALL
```
out-of-phaze: No, that's not where this flag is used.

code/modules/materials/definitions/solids/materials_solid_glass.dm
```diff
@@ -60,6 +60,7 @@
 		/decl/material/solid/organic/plastic = 0.3
 	)
 	color = COLOR_OFF_WHITE
+	flags = MAT_FLAG_ALTERATION_COLOR
```
out-of-phaze: This flag doesn't go here.

mods/mobs/dionaea/mob/gestalt/gestalt_movement.dm
```diff
@@ -6,8 +6,7 @@
 	. = ..()
 	if(AM && can_roll_up_atom(AM) && AM.Adjacent(src))
 		var/turf/stepping = AM.loc
-		roll_up_atom(AM)
-		if(stepping)
+		if(roll_up_atom(AM) && stepping)
```
out-of-phaze, line 9: I'm not convinced this is the correct fix but it *should* solve it at least.

out-of-phaze (changes requested):

code/modules/materials/definitions/solids/materials_solid_glass.dm
markoatonc: removed it

code/game/objects/structures/window.dm
```diff
@@ -639,7 +640,7 @@
 		if (ST.use(required_amount))
 			var/obj/structure/window/WD = new(loc, ST.material.type, ST.reinf_material?.type, dir_to_set, FALSE)
 			to_chat(user, SPAN_NOTICE("You place [WD]."))
-			WD.set_anchored(FALSE) // handles setting construction state for us
+			WD.set_anchored(TRUE) // handles setting construction state for us
```
MistakeNot4892: This isn't an appropriate fix for reinforced windows. If they're getting locked in the wrong state then there's a logic error in the construction code, this is just masking the issue. Deconstructing windows will run into the same problem if it's not fixed in the appropriate spot.

As far as I recall windows have always started unanchored when constructed and are then anchored with screwdriver. Not sure where the comment otherwise in the body came from.

code/modules/projectiles/ammunition/boxes.dm
```diff
@@ -59,6 +59,7 @@
 
 /obj/item/ammo_magazine/shotholder/on_update_icon()
 	..()
+	create_initial_contents()
```
MistakeNot4892: This isn't an appropriate fix for an issue with an interaction. If the issue is with `stored_ammo` then it should be swapped out for using the proc that includes abstract ammo. If that's not feasible then create_initial_contents() should be called prior to the interaction, not in a totally unrelated icon update proc.

MistakeNot4892:
Window construction and ammo icon issues fixed in #4997 and #4996, related changes will need to be dropped from this PR.

markoatonc:
> Window construction and ammo icon issues fixed in #4997 and #4996, related changes will need to be dropped from this PR.

done, and the diona movement is still only a workaround of sorts. but I do not really play or understand diona enough to really fix it.

out-of-phaze (approved):

PR #4995, opened 2025-05-08
Ambient Lighting: Multiple-source Ambience, Ambience Groups, and Reorg

Lohikar:
This updates the ambience code to the newest revision (with multiple-source support), as well as finally adding ambience groups.

A turf can be a member of multiple ambience groups at once, and will sum the light applied by all of them. There is a hard limit of 24 ambience groups in the world at once, but ambience groups can be deleted to free up their index.

The turf's own ambience is treated as a separate channel that will be blended with ambience groups. Clearing it out will not clear out ambience group applied lighting. It's intended that this channel is only used by the turf itself, but this is not enforced.

Other changes:
- There is now a hook to temporarily suspend instant lighting updates when doing bulk updates. This is stack based, multiple items can call this and updates will only resume when all locks are released.
- `LIGHTING_SOFT_THRESHOLD` is now mostly dead, it wasn't really doing anything in the first place.
- The global lighting corner has been removed since nothing really used it. The supermatter cascade is currently using a hack workaround as it _did_ use this list, but it should be changed to use ambient lighting at some point, likely a special global channel.

### DIFF
--- a/code/__defines/lighting.dm
+++ b/code/__defines/lighting.dm
@@ -10,7 +10,8 @@
 #define LIGHTING_DARKNESS_ICON_STATE "black"	// icon_state used for lighting overlays with no luminosity.
 #define LIGHTING_TRANSPARENT_ICON_STATE "blank"
 
-#define LIGHTING_SOFT_THRESHOLD 0.001 // If the max of the lighting lumcounts of each spectrum drops below this, disable luminosity on the lighting overlays.
+// This is purely used as a threshold for 'is this turf probably dark' to avoid floating point nonsense.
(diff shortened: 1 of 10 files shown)


Lohikar:
This is probably fine as-is, it just needs testing. I think it worked when @out-of-phaze tested it?

out-of-phaze:
> This is probably fine as-is, it just needs testing. I think it worked when @out-of-phaze tested it?

i forgot to actually make ambience use it, needs some tweaks for that to happen i think. will test sometime this week

comma (approved):

PR #4996, opened 2025-05-08
Fix ammo icon and interaction issues

out-of-phaze:
Fixes issues with shotholders and other magazines not using the lazy-ammo system properly.
Makes shotgun shells use the single-icon system.
Adds preview sprites for casings.
Sets a mapping preview color for casings.
Chem darts are no longer made of brass.


out-of-phaze:
Tested, confirmed to work.

MistakeNot4892 (approved):

PR #4997, opened 2025-05-08
Fix window color and construction

out-of-phaze:
Fixes reinforced windows being in an uninteractable state when constructed.
Fixes windows not using material color.
Fixes windows not respecting name/desc material alteration flags.


Lohikar (dismissed):

out-of-phaze:
Tested, confirmed to work.

MistakeNot4892 (approved):

PR #4999, opened 2025-05-09
Fix potential href exploit in stack crafting

out-of-phaze:
Fixes a potential href exploit in stack crafting. Also makes all products receive fingerprints, not just the first. `produce_recipe()` also no longer directly calls `list_recipe()`, eliminating an unnecessary argument and allowing it to be used without opening the recipe menu, which means we can have in-world interactions use stack recipes eventually.

This PR was tested with:
- creating more than 1 item per sheet (with clay lumps)
- batch-creating items (once more, clay lumps)
- creating items that require more than 1 sheet each (travois made of logs)
and it didn't run into any false positives.

### DIFF
--- a/code/game/objects/items/stacks/stack.dm
+++ b/code/game/objects/items/stacks/stack.dm
@@ -166,9 +166,13 @@
 	popup.open()
 
 
-/obj/item/stack/proc/produce_recipe(decl/stack_recipe/recipe, var/producing, var/expending, mob/user, paint_color, sublist = null)
+/obj/item/stack/proc/produce_recipe(decl/stack_recipe/recipe, var/producing, var/expending, mob/user, paint_color)
 
 	if(producing <= 0 || expending <= 0 || expending > get_amount())
+		return
+
+	if(expending > recipe.get_required_stack_amount(src, product_amount = producing))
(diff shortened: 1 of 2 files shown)


code/game/objects/items/stacks/stack.dm
```diff
@@ -166,11 +166,15 @@
...
 
 
-/obj/item/stack/proc/produce_recipe(decl/stack_recipe/recipe, var/producing, var/expending, mob/user, paint_color, sublist = null)
+/obj/item/stack/proc/produce_recipe(decl/stack_recipe/recipe, var/producing, var/expending, mob/user, paint_color)
 
 	if(producing <= 0 || expending <= 0 || expending > get_amount())
 		return
 
+	if(expending > recipe.get_required_stack_amount(src, product_amount = producing))
+		PRINT_STACK_TRACE("Possible HREF hacking attempt, recipe amount consumed and produced doesn't match!")
```
MistakeNot4892, line 175: Should this mention ckey explicitly?

MistakeNot4892 (approved):

code/game/objects/items/stacks/stack.dm
out-of-phaze: i think having it be implicit in the stacktrace is probably okay, the important thing is to just stop the exploit

PR #5002, opened 2025-05-09
Minor camera fixes and tweaks

Neerti:
## Description of changes

This was supposed to be a quick and easy fix for mapped in x-ray cameras not actually becoming x-ray cameras, but I stumbled into some more bugs, so I fixed them as well.
List of all changes:
- Changes the component rating threshold for cameras becoming x-ray cameras, or EMP proof, from `> 2` to `>= 2`, since the mapped in emp proof and x-ray cameras could not meet those thresholds with the parts they start with. They can now reach the thresholds with one adv component, which is how I remembered how it worked on other non-Neb codebases.
- Removed a few magic numbers, in favor of some consts.
- Somewhat refactors camera EMP code, to actually function, and use timers and callbacks instead of checking in `Process()`.
- A camera which becomes capable of being an x-ray camera by adding components will have its sprite changed to the x-ray camera one, instead of that sprite only being applied to mapped in x-ray cameras.
- Mapped in camera variants (emp proof, x-ray, motion, and the 'all' camera which has all three) changed to use uncreated components instead.

Mapped in special cameras will work as advertised, and cameras can be EMPed again.

## Authorship

Neerti

## Changelog
:cl:
tweak: X-ray cameras made in-round will show the same sprite as mapped in x-ray cameras.
/:cl:


out-of-phaze (dismissed):
I appreciate the use of constants for this rather than defines

EDIT: actually this might need to be sent to staging or stable depending on if it applies to them. not sure though

MistakeNot4892:
Looks like it could probably go to staging at least.

Neerti:
I did some testing and the bugs also seem to be on `stable`, so I'll point it there soonish.

comma:
Need to cherry pick the fix commits, 3.8k commits doesnt look right

PR #5005, opened 2025-05-10
Add a user agent to the BYOND install script

out-of-phaze:
Fixes the BYOND install script not working due to BYOND refusing downloads to Github Actions IPs if there's no user agent.


out-of-phaze:
This fixes CI across all branches, so I'm going to selfmerge so I can update staging and dev.

PR #5010, opened 2025-05-12
Fixes inability to sell to merchants due to overly eager telepads

Neerti:
## Description of changes

Merchants currently cannot be sold to, because the merchant pad tries to sell anything on top of it, including immobile machines mounted on the walls, and in Tradeship's case, a job spawn point.
The merchants don't want a fire alarm or a `/obj/abstract/landmark/start` and so you can't sell anything. This bug affects `stable`, `staging`, and `dev`, and this PR fixes it by excluding unsimulated objects and anchored objects, and removes a bit of copypasta on the side.

Merchants can be sold to once again.

## Authorship

Neerti

### DIFF
--- a/code/modules/merchant/merchant_machinery.dm
+++ b/code/modules/merchant/merchant_machinery.dm
@@ -25,10 +25,8 @@
 		. += a
 
 /obj/machinery/merchant_pad/proc/is_valid_target(atom/movable/thing)
+	if(!istype(thing))
+		return FALSE
 	if(thing == src)
 		return FALSE
-	if(!isobj(thing) && !isliving(thing))
-		return FALSE
-	if(thing.anchored || !thing.simulated)
-		return FALSE
-	return TRUE
+	return thing.is_valid_merchant_pad_target()
--- a/code/modules/merchant/merchant_machinery.dm
+++ b/code/modules/merchant/merchant_machinery.dm
@@ -12,7 +12,7 @@
 /obj/machinery/merchant_pad/proc/get_target()
 	var/turf/T = get_turf(src)
 	for(var/a in T)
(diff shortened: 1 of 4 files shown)


code/modules/merchant/merchant_machinery.dm
```diff
@@ -12,14 +12,23 @@
...
-		if(a == src || (!istype(a,/obj) && !isliving(a)) || istype(a,/obj/effect))
+		if(!is_valid_target(a))
 			continue
 		return a
 
 /obj/machinery/merchant_pad/proc/get_targets()
 	. = list()
 	var/turf/T = get_turf(src)
 	for(var/a in T)
-		if(a == src || (!istype(a,/obj) && !isliving(a)) || istype(a,/obj/effect))
```
out-of-phaze, line 23: oh that's funny, when we repathed the start landmarks from `/obj/effect/landmark` to `/obj/abstract/landmark` this must've stopped working, ha

out-of-phaze (dismissed):

MistakeNot4892:
Might want to add an exception to the anchored check for `/mob/living/exosuit` as that is anchored.

Neerti:
Changed the checks to be on the objects themselves, since adding the mecha exemption to the merchant pad itself seemed like a code smell, plus I thought of a potential issue arising if somebody tries to sell an exosuit while somebody else is still inside.

I can change the name of the proc if desired since it's a bit clunky.

out-of-phaze (approved):
I'd be fine with moving the procs from the base object files and into the `merchant_machinery` file, honestly, since that sort of thing promotes feature isolation and would allow for merchant to be made into a drop-in modpack in the future if desired. That said this works fine as-is

PR #5014, opened 2025-05-16
Fix RPD dispensing uncolored pipes

Typhin:
## Description of changes

Fixes two small problems that prevented pipes dispensed by an RPD from having their color set properly.
/code/game/objects/items/weapons/RPD.dm - Line 149 had a typo in "selected_color" that caused the color not to be seen by the build() function.
/code/modules/fabrication/designs/pipe/pipe_datum_base.dm - Lines 38 and 40 set new_item.color directly, but then the color was being overridden with null due to there not being any paint_color ever specified.  Changed to call .set_color() instead, which sets the paint_color variable.

The color setting on the RPD was useless, now it works correctly.  Before, all pipes would come out as regular gray/white pipes, regardless of type.  Now, Supply/Scrubber/Fuel pipes dispensed are their respective colors, and regular pipes respect the color selector.  

## Authorship

Typhin

## Changelog
:cl:
bugfix: RPD now dispenses colored pipes like it should
code: /code/game/objects/items/weapons/RPD.dm and /code/modules/fabrication/designs/pipe/pipe_datum_base.dm
/:cl:

### DIFF
--- a/code/game/atoms_movable.dm
+++ b/code/game/atoms_movable.dm
@@ -532,3 +532,5 @@
 		appearance_flags &= ~remove_flags
 	return old_appearance != appearance_flags
 
+/atom/movable/proc/is_valid_merchant_pad_target()
+	return simulated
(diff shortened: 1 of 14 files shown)


Typhin:
Wrong branch, supposed to be on Staging, gonna try again.

PR #5021, opened 2025-05-16
Refactor stimulants/antidepressants to reduce code duplication

out-of-phaze:
makes these two share a parent type and a more general framework.


MistakeNot4892 (approved):
Personally I would like this to be generalized on /decl/material but heigh ho.

PR #5023, opened 2025-05-16
Refactor syndie cigarettes to reduce code duplication

out-of-phaze:
they had a lot of duplicated code, this moves it to a shared type. includes a map migration. currently untested :)
honestly this could probably be removed but i think the codex or uplink or something depends on the names being like this. shrug


out-of-phaze:
so this is broken (the reagents are added to the carton instead of the cigarettes) but like. uh. i think that's how it was before. idk who broke it or when but the part this code touches is fine enough

out-of-phaze:
nvm it transfers the reagents to the cigarette on removal from the carton which feels. really janky? but works i guess, so there's no bug.

comma (approved):

PR #5039, opened 2025-05-26
Fix HIDETAIL not hiding tails

out-of-phaze:
missing return caused it to just. do the tail sprite anyway. lol


out-of-phaze:
stable PRs just don't work rn because it can't download a copy of BYOND 515 to cache

comma (approved):

PR #5040, opened 2025-05-26
Fix issues with ZM discovery and updates

### LINKED ISSUE #4998: Objects that fall down z-levels (sometimes?) aren't mimicked
I pushed a portable air pump down the elevator shaft and it wasn't mimicked by ZM even after moving it around some, until I pulled it to a turf without openspace above, then back into the elevator. That seemed to fix it.

This was on Tradeship staging, commit f539d763a66a14e69e7df205674c7455a8615aa6.

out-of-phaze:
Fixes #4998. Generally just does a correctness pass on movable discovery in ZM, with lots and lots of guidance from Lohikar. Also collapses some duplicate movable-level definitions of forceMove, one of which was already redundant.

### DIFF
--- a/code/controllers/subsystems/zcopy.dm
+++ b/code/controllers/subsystems/zcopy.dm
@@ -502,12 +502,13 @@ SUBSYSTEM_DEF(zcopy)
 
 	ZM_RECORD_START
 
+	var/above_needs_discovery = FALSE
 	if (!object.bound_overlay)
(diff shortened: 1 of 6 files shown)


Lohikar (dismissed):

Lohikar (approved):
Gating `discover_movable()` on init state doesn't feel right (probably should check explicitly for list existence), but it's realistically fine.

PR #5041, opened 2025-05-28
Fix has_trait not working for species traits

out-of-phaze:
thanks to @markoatonc for pointing this out, just needs to check get_traits instead of traits. adds a doc comment and renames the var, plus makes it private and adds a helper proc for resetting mob traits. renames a bunch of the UpperCamelCase procs to use snake_case instead, for consistency.


MistakeNot4892 (approved):

markoatonc:
is there a chance this can be done on DEV too? Scav pulls from dev and it seems most fixes are done to staging not dev, which is confusing me since I was taught the chain is dev>staging>release

PR #5045, opened 2025-06-03
Prevent suit cyclers from rotating with shuttles

### LINKED ISSUE #4927: onmob overlays of suit cyclers don't rotate when you fly a shuttle in rotation
#### Description of issue

Image

latest dev

out-of-phaze:
fixes #4927, I think. the suits were doing the right thing, the issue is just the cyclers were mapped in a weird direction and it wasn't visible because they only have south-facing sprites. this just disables rotation so they stay south-facing.


MistakeNot4892 (approved):

MistakeNot4892:
Overrding set_dir() on the cycler might be wise.

PR #5051, opened 2025-06-08
Fix issues with clothing icons

### LINKED ISSUE #4984: Suit cycler icons are immediately overwritten by update_icon code
#### Description of issue

I was investing #4939 (it's mostly unrelated actually!) and seeing if I could add world sprite overrides, but found that they weren't working despite world sprites being part of the checks done for species sprites. After adding a bunch of debug strings, I think I found the issue but don't think I can fix it myself just now, so documenting it here:

This is the bit of debug I added on proc/refit_for_bodytype:
```
	var/last_icon = icon
	var/species_icon = LAZYACCESS(sprite_sheets, target_bodytype)
	to_world("checking" + target_bodytype) //DEBUG REMOVE ME CEREBULON
	if(species_icon)
		to_world("There is a species icon!")//DEBUG REMOVE ME CEREBULON
	if(!species_icon)
		to_world("no species icon found!") //DEBUG REMOVE ME CEREBULON
	if(species_icon && (check_state_in_icon(ICON_STATE_INV, species_icon) || check_state_in_icon(ICON_STATE_WORLD, species_icon)))
		to_world("Changing state") //DEBUG REMOVE ME CEREBULON
		icon = species_icon

	if(!skip_rename)
		SetName("refitted [initial(name)]")

	if(last_icon != icon)
		reconsider_single_icon()
		update_clothing_icon()
```

All messages are displayed indicating that the icon = species_icon line was run, but no change to the world sprite is made. On-mob sprite is actually handled by a check on bodytype spritesheets when the item is equipped, all that is "functionally" happening here is the change of the bodytype equip flags, which allows them to appear to function from a player perspective.

I tried manually changing the icon var to the species spritesheet with the new world sprite (which icon = species_icon is doing here), which changed the icon successfully. Until I picked it up, or switched hands, or any other interaction, where it reverts to the base_clothing_icon value.

This bit of update_icon code seems to be the culprit:

```
/obj/item/clothing/on_update_icon()

...

	if(!base_clothing_icon)
		base_clothing_icon = initial(icon)
	set_icon(base_clothing_icon)
	if(!base_clothing_state)
		base_clothing_state = initial(icon_state)
	set_icon_state(base_clothing_state)
```

Which effectively reverts any changes made to the icon?

Unsure if changing the base_clothing_icon via the refit proc here is actually the correct approach since species spritesheets can in many cases "only" contain the onmob sprite and no world icon, it probably does need to be an "override" of some sort?

Suit cycler should apply change to "icon" var. This change gets reverted immediately by icon updates.

#### Steps to reproduce

Insert suit into cycler, refit to another bodytype, icon var is unchanged.

Tested with engineering voidsuit with placeholder 'world' icons added for avian and felinoid bodytypes.

Unsure if it began then, but update icon code seems to have been refactored in #4567 without accounting for refit code.

#### Issue bingo

- [X ] [Couldn't find an existing issue about this](https://github.com/NebulaSS13/Nebula/issues)

out-of-phaze:
Should hopefully fix #4984. I don't think any more stuff is going to get stuffed into this particular PR. Basically just removes some old/deprecated code. More stuff could be done but it's not worth blocking the staging->stable merge.

### DIFF
--- a/code/modules/clothing/_clothing.dm
+++ b/code/modules/clothing/_clothing.dm
@@ -296,9 +296,9 @@
 	var/last_icon = icon
 	var/species_icon = LAZYACCESS(sprite_sheets, target_bodytype)
 	if(species_icon && (check_state_in_icon(ICON_STATE_INV, species_icon) || check_state_in_icon(ICON_STATE_WORLD, species_icon)))
-		icon = species_icon
+		base_clothing_icon = species_icon
 
-	if(last_icon != icon)
+	if(last_icon != base_clothing_icon)
 		reconsider_single_icon()
 		update_clothing_icon()
 
--- a/code/modules/clothing/_clothing.dm
+++ b/code/modules/clothing/_clothing.dm
@@ -39,9 +39,6 @@
 	var/markings_color	// for things like colored parts of labcoats or shoes
 	var/should_display_id = TRUE
 	var/fallback_slot
-	// Used to track our icon, or custom icon, for resetting when accessories are added/removed
-	var/base_clothing_icon
-	var/base_clothing_state
 
 /obj/item/clothing/get_equipment_tint()
 	return tint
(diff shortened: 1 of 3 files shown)


code/modules/clothing/_clothing.dm
```diff
@@ -234,10 +234,10 @@
 
 	if(!base_clothing_icon)
 		base_clothing_icon = initial(icon)
-	set_icon(base_clothing_icon)
-	if(!base_clothing_state)
+	set_icon(base_clothing_icon) // todo: update use_single_icon and has_inventory_icon when we do this
+	if(!base_clothing_state) // todo: remove this var if unnecessary? we enforce that clothes use world/inventory states
```
MistakeNot4892: This predates all clothing using single icons, probably safe to remove.

MistakeNot4892 (approved):

PR #5053, opened 2025-06-08
Readded monitor head loadout option.

MistakeNot4892:
Cherrypicked from Polaris as part of incidental changes.

### DIFF
--- a/code/modules/client/preference_setup/loadout/lists/headwear.dm
+++ b/code/modules/client/preference_setup/loadout/lists/headwear.dm
@@ -168,3 +168,10 @@
 	name = "balaclava"
 	path = /obj/item/clothing/mask/balaclava
 	uid = "gear_head_balaclava"
+
+/decl/loadout_option/head/monitor_head
+	name = "monitor head (synthetics only)"
+	path = /obj/item/clothing/mask/monitor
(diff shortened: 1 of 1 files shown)


code/modules/client/preference_setup/loadout/lists/headwear.dm
```diff
@@ -168,3 +168,10 @@
 	name = "balaclava"
 	path = /obj/item/clothing/mask/balaclava
 	uid = "gear_head_balaclava"
+
+/decl/loadout_option/head/monitor_head
```
out-of-phaze, line 172: uhhh, this isn't actually restricted to utility frames only, which means people can just carry around monitor screens?

out-of-phaze (changes requested):

code/modules/client/preference_setup/loadout/lists/headwear.dm
MistakeNot4892: The only equip condition is having a robotic head and I didn't want to add support for checking that in this PR. If someone with a meat head buys it they won't be able to equip it.

EDIT: For clarity, what I mean is, it is not restricted to utility frames specifically, it's meant to work for any character with a robotic head.

out-of-phaze (approved):

PR #5056, opened 2025-06-13
Move custom circuit system into a modpack

out-of-phaze:
## Description of changes
Moves the custom circuit system into a modpack. Includes a few misc. changes to facilitate that. Currently untested, opening for CI.

- [x] Add map migration for `/obj/item/integrated_electronics` removal.
- [x] Add map migration for `/obj/prefab` -> `/obj/abstract/prefab` removal.

More modularization.

## Authorship
Me.


out-of-phaze:
Tested, hand tele prefab works fine.

code/game/objects/items/devices/paint_sprayer.dm
```diff
@@ -114,6 +114,11 @@
 		else
 			change_color(new_color, user)
 
+	else if(istype(A, /obj/item/card/data)) // TODO: un-hardcode this please. better yet redo how this entire proc is done
```
MistakeNot4892, line 117: Should be easy enough to put in a stub proc on /atom or such potentially.

MistakeNot4892 (approved):

PR #5057, opened 2025-06-14
Rename a var from `usr` to `user`

ike709:
## Description of changes

OpenDream is adding `usr` to the `SoftReservedKeyword` pragma: https://github.com/OpenDreamProject/OpenDream/pull/2307

This PR renames a `var/mob/usr` to `var/mob/user`.

This pragma is currently an error on Nebula, because using internal var names for declared var names is cringe.

## Authorship

I wrote the DM changes. See https://github.com/OpenDreamProject/OpenDream/pull/2307 for the OpenDream side of things.

## Changelog

it's a var rename


MistakeNot4892:
idk if we can trust this new contributor he seems kind shifty

MistakeNot4892 (approved):
Looks like a typo regardless of OD stuff, we typically block use of `usr`. Thanks for the commit.

PR #5069, opened 2025-07-08
Split preferences.copy_to across many, many procs

noelle-lavenza:
## Description of changes
Splits up `/datum/preferences/proc/copy_to()` into `copy_to_physical()` and `copy_to_nonphysical()`. Those have also been split across the various `player_setup_item`s. At some point I think the prefs vars those access should even be moved onto it, but that'd cause a lot of headaches I think because there's a ton of cross-dependencies.

I did some testing but it's very likely this is fundamentally broken in some non-obvious way I haven't been able to identify.

Oh and there's some commented-out code. I kept it in for review to make sure I didn't miss/overlook anything 🥲 

I wanted to be able to remove/replace the way characters are created and/or populated from prefs on a downstream.
Also, spawning in creates a character directly rather than populating a blank one, which lets us save some wasted time in init.

## Authorship
Me

### DIFF
--- a/code/modules/client/preference_setup/general/01_basic.dm
+++ b/code/modules/client/preference_setup/general/01_basic.dm
@@ -18,7 +18,7 @@
 	name = "Basic"
 	sort_order = 1
 
-/datum/category_item/player_setup_item/physical/basic/copy_to_nonphysical(mob/living/human/character, is_preview_copy = FALSE)
+/datum/category_item/player_setup_item/physical/basic/apply_post_snapshot_preferences(mob/living/human/character, is_preview_copy = FALSE)
 	character.set_gender(pref.gender)
 
 /datum/category_item/player_setup_item/physical/basic/preload_character(datum/pref_record_reader/R)
--- a/code/modules/client/preference_setup/general/01_basic.dm
+++ b/code/modules/client/preference_setup/general/01_basic.dm
@@ -18,6 +18,9 @@
 	name = "Basic"
 	sort_order = 1
 
+/datum/category_item/player_setup_item/physical/basic/copy_to_nonphysical(mob/living/human/character, is_preview_copy = FALSE)
+	character.set_gender(pref.gender)
+
 /datum/category_item/player_setup_item/physical/basic/preload_character(datum/pref_record_reader/R)
 	pref.gender =         R.read("gender")
 	pref.bodytype =       R.read("bodytype")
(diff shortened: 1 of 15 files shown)


code/modules/client/preference_setup/general/01_basic.dm
```diff
@@ -18,6 +18,9 @@
 	name = "Basic"
 	sort_order = 1
 
+/datum/category_item/player_setup_item/physical/basic/copy_to_nonphysical(mob/living/human/character, is_preview_copy = FALSE)
```
MistakeNot4892: Personally I would love it if we took this opportunity to make these procs apply to `/mob/living` but that is probably out of scope.

code/modules/client/preference_setup/general/03_traits.dm
```diff
@@ -39,6 +39,11 @@
 	else
 		pref.prune_invalid_traits()
 
+/datum/category_item/player_setup_item/traits/copy_to_nonphysical(mob/living/human/character, is_preview_copy = FALSE)
```
MistakeNot4892: Seems a bit odd to make traits applied as nonphysical things when they include missing limbs, robot organs, etc.

code/modules/client/preferences.dm
```diff
@@ -440,22 +398,7 @@ var/global/list/time_prefs_fixed = list()
...
-	character.flavor_texts["head"] = flavor_texts["head"]
-	character.flavor_texts["face"] = flavor_texts["face"]
-	character.flavor_texts["eyes"] = flavor_texts["eyes"]
-	character.flavor_texts["torso"] = flavor_texts["torso"]
-	character.flavor_texts["arms"] = flavor_texts["arms"]
-	character.flavor_texts["hands"] = flavor_texts["hands"]
-	character.flavor_texts["legs"] = flavor_texts["legs"]
-	character.flavor_texts["feet"] = flavor_texts["feet"]
-
+	// why is this in copy_to? jfc
```
MistakeNot4892, line 400: Believe so it only applies at roundstart.

noelle-lavenza: Well, it would apply to latejoiners too. I just don't like that it's applied in preference loading/character creation because that's... really silly, it should be handled in spawn code probably? But changing that is out of scope, I think.

code/modules/client/preference_setup/general/03_traits.dm
noelle-lavenza: Yeah it's a nomenclature issue, pushing a commit to fix that.

code/modules/client/preference_setup/general/01_basic.dm
noelle-lavenza: like specifically the nonphysical ones (renamed to `apply_post_snapshot_preferences` in a commit i've yet to push)? I could see about doing that (maybe in a future PR) but some things like character info are specifically defined on /mob/living/human, should I move those?

noelle-lavenza: okay to do that i'd have to:
- move `var/list/background_info` and `/mob/living/human/proc/set_background_value` onto `/mob/living`
- move `var/list/flavor_texts` onto `/mob/living`, or gate it behind an `ishuman` typecheck
- gate `appearance_descriptors` behind an `ishuman` typecheck

and i think a few other things. not sure what we'd gain from that at this point in time, since silicons don't go through this path, but i guess making it so that they do would be nice.

MistakeNot4892: Yeah no worries, that's more work than I expected.

MistakeNot4892 (approved):

PR #5070, opened 2025-07-16
Fix OpenDream reserved keyword errors

out-of-phaze:
this is holding up the staging->stable merge. technically `caller` isn't reserved in 515, which is the version OD should be using, but iirc staging should be running on 516 anyway? so once we merge staging->stable we should bump tests to 516 also


out-of-phaze:
this was apparently already done on https://github.com/NebulaSS13/Nebula/pull/5063 for dev, but this one is more thorough i think

MistakeNot4892 (approved):

PR #5097, opened 2025-07-25
Fix material instances being passed to GET_DECL, make holy water use reagent data fields

out-of-phaze:
see title/commit messages


MistakeNot4892 (approved):
Could this target staging/stable?

PR #5101, opened 2025-08-18
Fix stable branch issues from latest Dreamchecker update

out-of-phaze:
some issues appeared, this fixes them


code/modules/events/blob.dm
```diff
@@ -15,7 +15,7 @@
 
 	log_and_message_admins("Blob spawned in \the [get_area_name(T)]", location = T)
 	Blob = new /obj/effect/blob/core(T)
-	for(var/i = 1; i < rand(3, 4), i++)
+	for(var/i = 1; i < rand(3, 4); i++)
```
Lohikar: This should probably be converted into a range loop so it doesn't roll rand() multiple times.

Lohikar (approved):

PR #5112, opened 2025-08-24
Fix chat filters being broken

out-of-phaze:
oops, i forgot to make sure the build filter lists proc got called when it stopped being a startup hook

### DIFF
--- a/code/modules/chat_filter/_chat_filter.dm
+++ b/code/modules/chat_filter/_chat_filter.dm
@@ -1,21 +1,29 @@
 var/global/list/chat_blockers_in_use
 var/global/list/chat_modifiers_in_use
 
-/proc/build_filter_lists()
+/proc/_get_chat_filters_in_use()
 	global.chat_blockers_in_use = list()
 	global.chat_modifiers_in_use = list()
-	var/list/all_filters = decls_repository.get_decls_of_type(/decl/chat_filter)
-	for(var/filtertype in all_filters)
-		var/decl/chat_filter/chat_filter = all_filters[filtertype]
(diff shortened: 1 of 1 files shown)


code/modules/chat_filter/_chat_filter.dm
```diff
@@ -38,6 +25,15 @@ var/global/list/chat_modifiers_in_use
 	var/can_modify_message = FALSE
 	var/can_deny_message = FALSE
 
+/decl/chat_filter/Initialize()
```
MistakeNot4892: Is the type tree iterated somewhere else? If the chat code just uses those global lists, they're always going to be empty.

out-of-phaze: it's iterated yes. it worked in local testing. i can't actually figure out where? which is weird. but it worked

out-of-phaze: it didn't work, i think it only worked in testing due to some other local changes involving filters. new version does work
image

MistakeNot4892: Guessing maybe you looked at the help verb, since that may have iterated it? Unsure, I don't remember that code well.

MistakeNot4892 (approved):

code/modules/chat_filter/_chat_filter.dm
out-of-phaze: > Guessing maybe you looked at the help verb, since that may have iterated it? Unsure, I don't remember that code well.

nope, that didn't work either. i believe i just had other code that iterated it which wasn't included in the PR

PR #5114, opened 2025-09-05
ambience: Fix member turfs list corruption

Lohikar:
changes: 
- Adding a turf to an ambience group with a Z less than the local maximum will no longer erase the ambience group.

### DIFF
--- a/code/modules/lighting/ambient_group.dm
+++ b/code/modules/lighting/ambient_group.dm
@@ -44,7 +44,7 @@ var/global/ambience_group_map[BITWISE_MAX_BITS]
 	set waitfor = FALSE
 
 	UNTIL(!busy)
-	if (T.z > member_turfs_by_z)
+	if (T.z > member_turfs_by_z.len)
 		member_turfs_by_z.len = T.z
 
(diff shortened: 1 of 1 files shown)


code/modules/lighting/ambient_group.dm
```diff
@@ -44,7 +44,7 @@ var/global/ambience_group_map[BITWISE_MAX_BITS]
 	set waitfor = FALSE
 
 	UNTIL(!busy)
-	if (T.z > member_turfs_by_z)
+	if (T.z > member_turfs_by_z.len)
```
out-of-phaze, line 47:
```suggestion
	if (T.z > length(member_turfs_by_z))
```
i believe we prefer length() in checks and `.len` only for assignment

out-of-phaze (changes requested):

code/modules/lighting/ambient_group.dm
Lohikar: This file already uses `.len` everywhere.

out-of-phaze (approved):

PR #5126, opened 2025-09-15
Clean up unused antagonist code on special_role and mind

out-of-phaze:
## Description of changes
Cleans up and/or implements unused code related to antagonists.

`/datum/mind/proc/is_brigged()` is a legacy helper that is now unused and irrelevant to the present state of the game. It should be removed to improve maintainability and avoid confusion.
`/decl/special_role/proc/tick()` is unused and not called anywhere, and ticking special roles are no longer supported. To avoid issues with subtypes overriding this unused method and getting unexpected results, the method should be removed entirely.
`/decl/special_role/proc/is_latejoin_template()` exists but is unused. Several checks that are identical to its definition exist in the code. They have been replaced with `is_latejoin_template()` to reduce redundancy and make future extension easier.

### DIFF
--- a/code/datums/mind/mind.dm
+++ b/code/datums/mind/mind.dm
@@ -449,20 +449,6 @@
 	if(H)
 		qdel(H)
 
-
-// check whether this mind's mob has been brigged for the given duration
-// have to call this periodically for the duration to work properly
-/datum/mind/proc/is_brigged(duration)
-	var/area/A = get_area(current)
-	if(!isturf(current.loc) || !istype(A) || !(A.area_flags & AREA_FLAG_PRISON) || current.GetIdCard())
-		brigged_since = -1
-		return 0
-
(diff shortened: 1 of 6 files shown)


code/datums/mind/mind.dm
```diff
@@ -449,20 +449,6 @@
 	if(H)
 		qdel(H)
 
-
-// check whether this mind's mob has been brigged for the given duration
-// have to call this periodically for the duration to work properly
-/datum/mind/proc/is_brigged(duration)
-	var/area/A = get_area(current)
-	if(!isturf(current.loc) || !istype(A) || !(A.area_flags & AREA_FLAG_PRISON) || current.GetIdCard())
-		brigged_since = -1
```
comma, line 458: `brigged_since` seems to be mind var, can be removed too

comma (approved):

PR #5130, opened 2025-09-23
Fix missing populate_branches call

out-of-phaze:
This call was forgotten when stuff was moved off hooks.

- [ ] Tested

(We don't have a map with branches, so... uh...?)

### DIFF
--- a/code/game/world.dm
+++ b/code/game/world.dm
@@ -101,6 +101,7 @@ GLOBAL_PROTECTED_UNTYPED(game_id, null)
 	update_holiday() //Uncommenting ALLOW_HOLIDAYS in configuration will enable this.
 	try_load_alien_whitelist()
 	investigate_reset()
+	global.using_map.populate_branches()
 	// Precache/build trait trees.
 	for(var/decl/trait/trait in decls_repository.get_decls_of_type_unassociated(/decl/trait))
 		trait.build_references()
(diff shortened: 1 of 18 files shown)


code/game/world.dm
```diff
@@ -101,6 +101,7 @@ GLOBAL_PROTECTED_UNTYPED(game_id, null)
 	update_holiday() //Uncommenting ALLOW_HOLIDAYS in configuration will enable this.
 	try_load_alien_whitelist()
 	investigate_reset()
+	global.using_map.populate_branches()
```
MistakeNot4892:
```suggestion
	populate_branches()
```
It's a global proc, not on `/datum/map`.

MistakeNot4892 (changes requested):
See suggestion.

code/modules/submaps/submap_join.dm
```diff
@@ -68,9 +68,9 @@
 		var/mob/living/human/user_human
 		if(ishuman(character))
 			user_human = character
-			if(job.branch && mil_branches)
-				user_human.char_branch = mil_branches.get_branch(job.branch)
-				user_human.char_rank =   mil_branches.get_rank(job.branch, job.rank)
+			if(job.branch && global.using_map & MAP_HAS_BRANCH)
```
MistakeNot4892: Missing parens

MistakeNot4892 (approved):

PR #5132, opened 2025-09-24
Expanding mob controller handler-directing logic.

MistakeNot4892:
- Adds entrypoints for AIs to take direction from another mob, currently used by mounts.
- Loosens hunter AI target handling so they can potentially target non-mobs in the future (used by separate Pyrelight PR).


code/modules/mob/skills/skillset.dm
```diff
@@ -37,7 +37,7 @@ var/global/list/all_skill_verbs
 	for(var/datum/skill_buff/SB in skill_buffs)
 		. += SB.buffs[skill_path]
 
-/datum/skillset/proc/obtain_from_mob(mob/mob)
+/datum/skillset/proc/obtain_from_mob(mob/living/mob)
```
out-of-phaze, line 40: This change is duplicated from #5131.

out-of-phaze (changes requested):
Mind explaining how the `silent = FALSE` argument to `get_scooped` fits into this PR? Was it just a change you made while in the area?

MistakeNot4892:
> Mind explaining how the `silent = FALSE` argument to `get_scooped` fits into this PR? Was it just a change you made while in the area?

Snuck in when I separated the commits out, oops. I need it for birds to reenter the handler's inventory without the default message.

code/modules/mob/skills/skillset.dm
MistakeNot4892: Rebased, this is no longer in the diff.

MistakeNot4892:
This appears to have broken hunting AI, will need to rework it.

MistakeNot4892:
Tested, works fine, rabbits are just too big for foxes to hunt currently.

out-of-phaze (dismissed):

MistakeNot4892:
Pushed some additional commits from the bird branch - apologies. They are largely self contained in each commit.

code/game/objects/effects/fadein.dm
out-of-phaze, line 1: Why is this in `code/game/objects/items/devices`?

code/game/objects/effects/fadeout.dm
out-of-phaze, line 1: Why is this in `code/game/objects/items/devices`?

out-of-phaze (changes requested):

out-of-phaze (approved):

PR #5143, opened 2025-09-26
Rewrite shuttle computers for better code quality

out-of-phaze:
Rewrites shuttle computers. Now instead of displaying a warning to the user on interaction if the evac controller is the wrong type, it throws a runtime error. This will probably mean it needs to be excluded from Create and Destroy testing and codex dump generation.
Makes them use `emag_act`.
Makes `authorized` a lazylist.
Cleans up some really old decompiler artifacts like the use of `3.0`, and renames `auth_need` to `auths_needed`.
Eliminates one-line guard clauses because IIRC we don't like those.
Just overall cleans up the code.
- [ ] Tested

### DIFF
--- a/code/controllers/subsystems/initialization/codex_dump.dm
+++ b/code/controllers/subsystems/initialization/codex_dump.dm
@@ -131,7 +131,8 @@ TODO: work out how to implement an external search function.
 		/obj/item/organ/external/chest,
 		/obj/machinery/power/apc,
 		/obj/machinery/alarm,
-		/obj/structure/stairs
+		/obj/structure/stairs,
+		/obj/machinery/computer/shuttle,
 	)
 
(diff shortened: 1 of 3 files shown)


MistakeNot4892:
CI fail is genuine.

MistakeNot4892 (changes requested):
CI fail.

MistakeNot4892 (dismissed):

MistakeNot4892 (changes requested):
Im reviewer

out-of-phaze (commented):
hey so issue i can't actually test this; `/obj/machinery/computer/shuttle` isn't mapped in on any map ever. it uses `/obj/machinery/computer/shuttle_control/emergency` instead. i think this is just. dead code

out-of-phaze:
superseded by #5353

PR #5144, opened 2025-09-26
Improve implant code quality

out-of-phaze:
## Description of changes
Removes the unused `/obj/machinery/implantchair` type, its entire file, and the associated icon file.
Stylistically and logically improves implant-related code.
- All span strings in the same files as `/obj/item/implant` references have been replaced with `SPAN_FOO()` macros for readability and maintainability.
- Boolean variables in implant-related methods have had numerical values replaced with the `TRUE` and `FALSE` defines.
- Enum constants used in implant code have been made to use `const` variables rather than raw numbers.
- Variables in implant-related code have been renamed to meet modern style guidelines, and should have clear, unambiguous names longer than one- or two- characters.
- Several implant-related overrides of Topic() (implants or implant management devices or consoles) have been replaced with OnTopic() overrides, and `usr` has been replaced with `user`.
- Guards for HREF hacking have been added to certain OnTopic overrides, along with istype() checks.
- The prisoner implant management console's chem implant HREF handling in OnTopic() has been rewritten to avoid duplicate code by sending the implant and the amount as separate arguments.
- Loyalty implants have a new description, taken from their fluff text.
- Locators have had their UI code rewritten and brought in line with modern code standards for clarity, readability, and maintainability.
- One-line guard clauses have been changed to be across multiple lines, as one-line conditionals are stylistically dispreferred.
- `foo.len` has been replaced with `length(foo)` for better null-safety and performance.
- Imprinting implant strings have been tweaked for better readability: `jointext(., "<br>")` is now used in `get_instructions()` rather than hardcoded `<br>` insertion, and multi-line `to_chat()` strings have been broken up into multiple `to_chat()` calls where feasible.
- A hardcoded z==2 check has been removed from locators.
- Certain constant values have been rewritten (with the same end value) to make more sense and improve readability, e.g. an `exposed_volume` of 125 is now `CELL_VOLUME/20`, a duration of `6000` is now `10 MINUTES` and a duration of `10` is now `1 SECOND`, etc.
- Repeated strings like implant activation memos have been saved in a variable so they can be referenced twice, by `to_chat()` and `StoreMemory()`, reducing redundancy.
- Some comments and strings have had minor grammar adjustments.
- `AttemptAugmentation`, `implant_in_mob`, and `implanted` now take a `/mob/living` as their first argument rather than just a `/mob`.
- All calls to `implant_in_mob` now have all three arguments, as some were previously missing `user`.
- Some hardcoded `implanted = TRUE; imp_in = mob; part = bodypart` sets in implant code were replaced with `implant_in_mob()` where feasible (anywhere a user exists, because `implanted()` often blocks/requires user input).

I might pull this one out into its own PR later, but:
- Readds the codex implant and makes requiring it a config option.
- If the codex implant config option is enabled, the codex implant is available as a 0-point augmentation in the loadout.

TBH it might be worth deleting some of the entirely-unused implants like the adrenalin implant, but YMMV.

Improves code quality for implants, which are like... ancient.


MistakeNot4892:
UID failure on a codex trait decl in CI.

MistakeNot4892 (changes requested):
Missing UID. Will review this properly once I'm at my desktop again (24h)

MistakeNot4892 (approved):

PR #5145, opened 2025-09-26
Add support for saltwater turfs

out-of-phaze:
## Description of changes
Salvages an old, forgotten commit that adds support for contaminants to fluid-filled turfs, and uses that system to add saltwater turfs.

Proof-of-concept for contaminated fluid sources.

## Authorship
Me

### DIFF
--- a/code/game/turfs/unsimulated/mask.dm
+++ b/code/game/turfs/unsimulated/mask.dm
@@ -9,6 +9,11 @@
 	icon_state = "rockvault"
 	color = COLOR_SILVER
 
+// Why do these exist? Are they just for typechecks when generating random maps? Does the flooding code even run for unsim turfs?
 /turf/unsimulated/mask/flooded
 	flooded = /decl/material/liquid/water
 	color = COLOR_LIQUID_WATER
+
+/turf/unsimulated/mask/flooded/salt
+	contaminant_reagent_type = /decl/material/solid/sodiumchloride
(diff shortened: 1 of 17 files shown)


code/game/turfs/unsimulated/mask.dm
```diff
@@ -9,6 +9,11 @@
 	icon_state = "rockvault"
 	color = COLOR_SILVER
 
+// Why do these exist? Are they just for typechecks when generating random maps? Does the flooding code even run for unsim turfs?
```
MistakeNot4892, line 12: Flooded status is transferred in ChangeTurf, these were added for underwater cave gen. Not sure you need a saltwater one for this unless that is also transferred in ChangeTurf.

code/modules/fluids/fluid_flood.dm
```diff
@@ -1,5 +1,7 @@
 // Permaflood overlay.
 var/global/list/flood_type_overlay_cache = list()
+// TODO: does this need to also take contaminant type as an argument? flooding contaminants are totally untested
+// also, do flooded turfs even apply fluid_act and touch effects?
```
MistakeNot4892, line 4: It has been ages since I touched that code but I believe they apply them via touch_turf. Could be wrong about the ordering though.

MistakeNot4892 (approved):
May want to add transferring contaminant type and proportion to ChangeTurf() alongside `flooded`.

PR #5147, opened 2025-10-01
Adding trained birds to fantasy modpack.

MistakeNot4892:
Depends on #5132.

## Description of changes
- Adds pigeons, which can be given a piece of paper or tiny item, and used in-hand to return to their hutch.
- Adds hawks, which can be directed to fetch an item or attack a mob.
- Adds crows, which can be told a player name and given a piece of paper or tiny item to deliver.

New communication methods/functionality for the fantasy modpack.

## Authorship
Code is mine.
Icons kindly made by @PlayerDeer.

## Changelog
:cl:
add: Added trained birds to fantasy modpack.
/:cl:

# TODO
- [x] Perched icon for birds
- [x] Implement pigeon returning behavior.
- [x] Add crow and pigeon message/small item carrying behavior.
- [x] Implement hawk item fetching behavior.
- [x] Implement hawk hunting behavior.
- ~~Get hawk grabs to work.~~
- ~~Clean up hawk deployment to remove lengthy pauses before processing starts.~~
- [x] Implement crow target/message passing behavior.


MistakeNot4892:
Calling this good enough for an initial PR - they're admin spawn only anyway.

code/modules/butchery/butchery_products.dm
```diff
@@ -3,7 +3,7 @@
 	material_alteration = MAT_FLAG_ALTERATION_COLOR
 	icon_state          = ICON_STATE_WORLD
 	material            = /decl/material/solid/organic/meat
-	color               = /decl/material/solid/organic/meat::color
+//	color               = /decl/material/solid/organic/meat::color
```
MistakeNot4892: No idea why, but this keeps preventing compilation.

MistakeNot4892: BYOND subversion issue

mods/content/birds/bird.dm
```diff
@@ -0,0 +1,81 @@
...
+/obj/item/holder/bird/attack_hand(mob/user)
+	if(loc == user)
+		var/mob/bird = locate() in src
+		if(istype(bird) && length(bird.get_held_item_slots()))
+			var/list/equipped = bird.get_held_items()
+			if(length(equipped))
+				var/obj/item/removing = pick(equipped)
+				if(bird.try_unequip(removing))
+					user.put_in_hands(removing)
+					to_chat(bird, SPAN_NOTICE("\The [user] confiscates your [removing.name]."))
```
Lohikar: This might be better as `visible_message()`.

mods/content/birds/bird_pigeon.dm
```diff
@@ -0,0 +1,55 @@
...
+	. = ..()
+	LAZYADD(., SPAN_SUBTLE("Pigeons can be given a small item like a piece of paper to hold."))
+	LAZYADD(., SPAN_SUBTLE("While holding a pigeon, use it in hand to release it to return to its home hutch."))
+
+/datum/mob_controller/passive/pigeon
+	emote_speech   = list("Oo-ooo.","Oo-ooo?","Oo-ooo...")
+	emote_hear     = list("coos")
+	emote_see      = list("preens its feathers", "puffs out its neck", "ruffles its wings")
+
+/mob/living/simple_animal/passive/bird/crow/get_examine_hints(mob/user, distance, infix, suffix)
```
Lohikar: Looks like this got misplaced.

mods/content/birds/hutch.dm
```diff
@@ -0,0 +1,71 @@
+/datum/storage/hutch
+	can_hold      = list(/obj/item/holder)
+	max_w_class   = MOB_SIZE_SMALL
+	storage_slots = 10
+
+/datum/storage/hutch/open(mob/user)
+	. = ..()
+	var/atom/hutch = holder
+	if(istype(hutch))
+		hutch.queue_icon_update()
```
Lohikar: Why not update immediately? This shouldn't be happening in bulk.

Lohikar (approved):

mods/content/birds/bird.dm
```diff
@@ -0,0 +1,80 @@
...
+	if(istype(hutch))
+		var/obj/item/holder/bird_item = new holder_type
+		forceMove(bird_item)
+		bird_item.sync(src)
+		hutch.storage?.handle_item_insertion(null, bird_item)
+		if(bird_item.loc != target)
+			dropInto(target.loc)
+			qdel(bird_item)
+	else
+		dropInto(isturf(target) ? target : target.loc)
```
out-of-phaze, line 39: dropInto exists so we don't have to do this sort of ternary, we can probably just do `dropInto(target)`? it forcemoves it into the first non-movable loc, with exceptions for closed closets and that's basically it.

mods/content/birds/bird.dm
```diff
@@ -0,0 +1,80 @@
...
+
+	var/obj/structure/hutch/hutch = target
+	if(istype(hutch))
+		var/obj/item/holder/bird_item = new holder_type
+		forceMove(bird_item)
+		bird_item.sync(src)
+		hutch.storage?.handle_item_insertion(null, bird_item)
+		if(bird_item.loc != target)
+			dropInto(target.loc)
+			qdel(bird_item)
```
out-of-phaze, line 37: this seems kind of similar to `/datum/storage/proc/scoop_inside(mob/living/scooped, mob/living/user)`, just without the `do_after()`. creates the holder, syncs it, inserts it, and if it fails to insert the holder gets deleted. i am kind of surprised that `scoop_inside()` only works if `scooped == user`, i wonder if maybe it was meant to be the opposite, but it works in this case.

```suggestion
		hutch.storage?.scoop_inside(src, src)
```

mods/content/birds/bird.dm
```diff
@@ -0,0 +1,80 @@
...
+		var/mob/bird = locate() in src
+		if(istype(bird) && length(bird.get_held_item_slots()))
+			var/list/equipped = bird.get_held_items()
+			if(length(equipped))
+				var/obj/item/removing = pick(equipped)
+				if(bird.try_unequip(removing))
+					user.put_in_hands(removing)
+					user.visible_message(SPAN_NOTICE("\The [user] confiscates \the [bird]'s [removing.name]."))
+					return TRUE
+	return ..()
```
out-of-phaze, line 69: not a fan of this really deep indentation. also, why do we check for held item slots and then check held items? don't we get the held item slots as part of that? there's no difference in outcome here between "has no held item slots" and "has no held items".

mods/content/birds/bird.dm
```diff
@@ -0,0 +1,80 @@
...
+
+/obj/item/holder/bird/attackby(obj/item/used_item, mob/user)
+	var/mob/living/simple_animal/passive/bird/bird = locate() in src
+	if(istype(bird) && bird.can_be_handled_by(user) && bird.get_empty_hand_slot() && user.try_unequip(used_item))
+		bird.put_in_hands(used_item)
+		if(used_item.loc != bird)
+			user.put_in_hands(used_item)
+		else
+			user.visible_message(SPAN_NOTICE("\The [user] gives \the [bird] \a [used_item] to carry."))
```
out-of-phaze, line 78: could do something like `bird.put_in_active_hand(used_item) || bird.put_in_inactive_hand(used_item))` to avoid having to check the loc stuff, because this way it won't remove it from the user's inventory on failure

```suggestion
		if(bird.put_in_active_hand(used_item) || bird.put_in_inactive_hand(used_item)
			user.visible_message(SPAN_NOTICE("\The [user] gives \the [bird] \a [used_item] to carry."))
```

mods/content/birds/bird_crow.dm
```diff
@@ -0,0 +1,110 @@
...
+		return
+	message = lowertext(strip_html_properly(message))
+
+	var/command = trim(copytext(message, 1, 6))
+	if(command != "go to" && command != "find")
+		return
+
+	// TODO make this less stupid/magic, can be exploited for metagaming as written.
+	// Check Z-level?
+	// Check faction/knowledge?
```
out-of-phaze, line 43: maybe make it so the crow has to be shown the mob beforehand? could even make it use taught nicknames, so you say the teaching keyword and the nickname, point to the mob you want to teach the crow the name of, and reuse that nickname for the find/go-to command. i dunno, just an idea.

to make it a little easier, you could even make it so that crows will spread that knowledge to other crows in the same hutch either immediately or over time, because that's a thing crows do, right? or is it bees? the whole thing about crows passing down the knowledge of which humans they hate because they were wronged, or whatever. that'd mean you could have someone send your crow to you, then you'd teach them the name of someone nearby (maybe even stealthily, e.g. they can't see you but you can see them), then send them back.

also opens up the possibility of using scrying/remote-viewing spells or similar, if those ever get added, to teach a crow who someone farther away is

and then any remaining weirdness about "how do they know where this person is" can be explained with magic

mods/content/birds/bird_crow.dm
```diff
@@ -0,0 +1,110 @@
...
+
+	if(isnull(current_target))
+		for(var/mob/player in global.dead_mob_list_)
+			if(findtext(lowertext(player.name), target_string))
+				current_target = weakref(player)
+				break
+
+	var/decl/pronouns/bird_pronouns = get_pronouns()
+	var/mob/mob_target = current_target?.resolve()
+	if(!istype(mob_target) || QDELETED(mob_target))
```
out-of-phaze, line 70: `/Destroy()` unsets `var/tmp/weakref/weakref` which is used to ensure the weakref resolves to the right datum (e.g. there hasn't been a del-and-reallocate for the text ID used under the hood), so `current_target?.resolve()` should never fail the QDELETED check, and it should never be a type we don't explicitly set it to. so i think just 
```suggestion
	if(!mob_target)
```
is fine

mods/content/birds/bird_crow.dm
```diff
@@ -0,0 +1,110 @@
...
+	LAZYADD(., SPAN_SUBTLE("While holding a crow, say 'Go to \[target\]' or 'Find \[target\]', then use the crow in hand to release it to travel to the target."))
+
+/mob/living/simple_animal/passive/bird/crow/hear_say(message, verb, decl/language/language, italics, mob/speaker, sound/speech_sound, sound_vol)
+	. = ..()
+
+	if(get_recursive_loc_of_type(/mob) != speaker || !can_be_handled_by(speaker))
+		return
+	message = lowertext(strip_html_properly(message))
+
+	var/command = trim(copytext(message, 1, 6))
```
out-of-phaze, line 37: `6` is a twice-used magic number here, if it's based on the length of `"go to"` or `"find"` i would like to either do that or possibly even use a regex to get the command and target name

mods/content/birds/bird.dm
MistakeNot4892: Ah I think I got confused by onDropInto() on /movable returning loc and thought it did that for /atom too

mods/content/birds/bird.dm
MistakeNot4892: I -believe- scoop_inside() was intended to let people put themselves into bags and such on the ground, I don't think there's much use for it being called by anyone other than the one being scooped. Or the idea was to prevent someone scooping mice into their bag for whatever reason? 

I'll use it in this case anyway.

mods/content/birds/bird.dm
MistakeNot4892: It's removed from the user's inv by try_unequip() prior to this block

PR #5148, opened 2025-10-07
Recursive Ambient Z-Lighting

Lohikar:
changes:
- Ambient lights now propagate upwards recursively instead of a single level.
- Ambient light z-propagation updates are now always queued.
- Fixes an issue where ambient light delta updates would break (burn) for turfs that had a light initialize before ambience.
- Turf opacity changes now trigger lighting vis updates as you'd expect them to.

~~Briefly tested as working on O7, but completely untested here.~~ Ambient light burn has been tested fixed, propagation is not tested here but works elsewhere.


MistakeNot4892 (dismissed):
Assuming the 'untested' note no longer applies given discussion on Discord.

MistakeNot4892 (approved):

PR #5151, opened 2025-10-10
Make APCs not a subtype of /obj/machinery/power

out-of-phaze:
## Description of changes
Repaths APCs from `/obj/machinery/power/apc` to `/obj/machinery/apc` and removes some jank that was necessary to make them work as an `/obj/machinery/power` subtype anyway.
Also updates some APC code and moves it around.

- [x] Tested (VERY IMPORTANT, NEEDS TO BE TESTED)

There was no reason for it to be on that subtype. At all. This makes the code make more sense.

## Authorship
Me


out-of-phaze:
APCs work, stuff has power, the APCs charged off of SMES power, etc. Seems good from my end.

MistakeNot4892 (approved):

PR #5155, opened 2025-10-10
Added barrel rims and basic coopery for crafted barrels.

MistakeNot4892:
## Description of changes

If forging is ticked, crafted barrels will need a barrel rim crafted/forged and attached before they can be used to hold items or reagents.

## TODO
- [x] Test.
- ~~Maybe move barrel rims into general code and add an autolathe recipe? Seems very specific for a futuretech machine.~~ Out of scope for this PR.

Makes crafting storage a bit more involved and allows customising barrel materials. Also makes the metal component of barrels not as imaginary.

## Authorship

Myself for the code, @PlayerDeer for the barrel rim icon if I remember correctly.

## Changelog
:cl:
tweak: Crafted barrels will need barrel rims forged if the forging modpack is in use.
/:cl:


out-of-phaze (approved):
this is awesome

out-of-phaze:
oops just saw that it's marked WIP. i don't think we really need barrel rims in core code though. i might see about combining barrels/casks and the beer keg at some point in the future though lol

mods/content/blacksmithy/barrel_rim.dm
```diff
@@ -0,0 +1,52 @@
...
+
+// Barrels with no reinforcement break apart when things are put inside.
+/obj/structure/reagent_dispensers/barrel/Entered(atom/movable/atom, atom/old_loc)
+	. = ..()
+	if(!metal_material && length(contents))
+		physically_destroyed()
+
+/obj/structure/reagent_dispensers/barrel/on_reagent_change()
+	. = ..()
+	if(!metal_material && length(contents))
```
MistakeNot4892: This should check `reagents?.total_volume` not `length(contents)`

mods/content/blacksmithy/barrel_rim.dm
```diff
@@ -0,0 +1,52 @@
...
+/obj/structure/reagent_dispensers/barrel/create_dismantled_products(turf/target)
+	if(metal_material)
+		new /obj/item/barrel_rim(target, metal_material)
+		metal_material = null
+	return ..()
+
+// Barrels with no reinforcement break apart when things are put inside.
+/obj/structure/reagent_dispensers/barrel/Entered(atom/movable/atom, atom/old_loc)
+	. = ..()
+	if(!metal_material && length(contents))
```
out-of-phaze: i'm not sure why we need to check `length(contents)` here, since this is in `Entered()`? `Entered()` can't even return false to cancel it, so it will always be added to contents here

mods/content/blacksmithy/barrel_rim.dm
out-of-phaze:
```suggestion
	if(!metal_material && reagents?.total_volume)
```

out-of-phaze (changes requested):

MistakeNot4892:
Changes in, tested, working.

mods/content/blacksmithy/barrel_rim.dm
```diff
@@ -0,0 +1,52 @@
...
+		physically_destroyed()
+
+// Adding a rim to a crafted barrel.
+/obj/structure/reagent_dispensers/barrel/attackby(obj/item/used_item, mob/user)
+	if(isnull(metal_material) && istype(used_item.material) && istype(used_item, /obj/item/barrel_rim))
+		user.visible_message(SPAN_NOTICE("\The [user] begins securing \the [src] with \the [used_item]."))
+		if(user.do_skilled(5 SECONDS, work_skill, src, check_holding = TRUE) && user.try_unequip(used_item))
+			metal_material = used_item.material
+			update_icon()
+			user.visible_message(SPAN_NOTICE("\The [user] secures \the [src] with \the [used_item]."))
```
out-of-phaze, line 49: in theory this could use a second-person message but /shrug

out-of-phaze (approved):

mods/content/blacksmithy/barrel_rim.dm
MistakeNot4892: I never put in second person messages tbh.

PR #5156, opened 2025-10-10
Added nails to forging.

MistakeNot4892:
## Description of changes
Adds nails, like the things what you hammer into things.

I genuinely don't know. 

This is left over from the original forging implementation and I'm kind of opening it because I don't know what to do with it. I think there's room for nails to be a useful component for crafting, but I don't think it would add anything to have them actually required for crafting furniture or anything like that.

Some possible uses for nails:
- Ammo for pneumatic cannons or a nailgun.
- Payload for an IED or explosive trap.
- Hammer them into planks to make improvised caltrops.
- A way to permanently attach things to other things.

## Authorship
Code is mine, I believe the original icons were @PlayerDeer.

## Changelog
:cl:
add: Nails can be forged. Why? No idea.
/:cl:


PlayerDeer:
Can confirm that is my nail sprite :)

out-of-phaze:
honestly, i could see fasteners being like. a way to reduce the skill needed for crafting. because while you can do wood joinery without metal fasteners (i know this because someone constantly shows me cool videos of it) it's a lot more difficult. so printing off some fasteners that you have to hold in your offhand/combine with a hammer to make building wood furniture easier would be neat.

probably not in scope though. i mean hey, at least even as-is nails would have more of a use than struts am i right,

MistakeNot4892:
I guess we can just merge them as they are now and use them as a prop if nothing else.

out-of-phaze (approved):

PR #5158, opened 2025-10-10
Adds a flash effect on mob damage.

MistakeNot4892:
## Description of changes
- Adds a modpacked feature that causes mobs to briefly flash a solid colour when hit.
- Speeds SSfastprocessing up to 1ds.
- Moves T-ray scanners off SSfastprocessing.
- Removes outdated IC removal from SSfastprocessing.

Opening as a draft because I'm really not sure if it should go in - it can look quite weird and choppy due to the silhouette not lerping when the donor mob is moving between animation states. Would appreciate some feedback on this one.

## TODO
- [ ] Test T-ray scanners.

Immediate damage feedback is nice.

## Authorship
Myself.

## Changelog
:cl:
add: Added a modpack for mobs flashing a solid colour when damaged.
/:cl:

### DIFF
--- a/mods/content/silhouette/effect.dm
+++ b/mods/content/silhouette/effect.dm
@@ -1,0 +1,51 @@
+// A silhouette mask used to cause mobs to flash a solid colour when hit.
+/obj/effect/silhouette
+	icon              = 'icons/effects/32x32.dmi'
+	icon_state        = "FFF"
+	mouse_opacity     = MOUSE_OPACITY_UNCLICKABLE
+	layer             = ABOVE_LIGHTING_LAYER
+	plane             = ABOVE_LIGHTING_PLANE
+	appearance_flags  = RESET_ALPHA|RESET_COLOR|KEEP_APART
+	is_spawnable_type = FALSE
+	alpha             = 0
+	var/mob/living/owner
+
+/obj/effect/silhouette/Initialize(mapload)
+	. = ..()
+	owner = loc
+	if(!istype(owner))
+		return INITIALIZE_HINT_QDEL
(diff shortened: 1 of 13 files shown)


mods/content/silhouette/effect.dm
```diff
@@ -0,0 +1,51 @@
...
+	transform = donor.transform
+	pixel_x = donor.pixel_x
+	pixel_y = donor.pixel_y
+	pixel_z = donor.pixel_z
+	pixel_w = donor.pixel_w
+	glide_size = donor.glide_size
+
+/obj/effect/silhouette/proc/follow_owner()
+	glide_size = owner.glide_size
+	forceMove(owner.loc)
```
out-of-phaze, line 51: not sure how this would interact with a mob going into a locker or similar

mods/content/silhouette/effect.dm
```diff
@@ -0,0 +1,51 @@
...
+	alpha             = 0
+	var/mob/living/owner
+
+/obj/effect/silhouette/Initialize(mapload)
+	. = ..()
+	owner = loc
+	if(!istype(owner))
+		return INITIALIZE_HINT_QDEL
+	global.events_repository.register(/decl/observ/moved, owner, src, TYPE_PROC_REF(/obj/effect/silhouette, follow_owner))
+	add_filter("owner_mask", 1, list(type = "alpha", render_source = "render_\ref[owner]"))
```
out-of-phaze, line 19: this feels a little over-engineered, are we sure we can't just use `blend_mode = BLEND_INSET_OVERLAY` and an overlay?

MistakeNot4892: Yes. Breaks the rendering.

MistakeNot4892:
I think I am abandoning this one unless someone works out a better solution.

PR #5159, opened 2025-10-11
Remove the last vestiges of struts

out-of-phaze:
Struts were basically already soft-removed, except for a few places. Now they're entirely gone, which should hopefully reduce confusion. The base type, all defined subtypes, the icons, and the fabricator recipes are gone, as well as the one reference to them as a substitute for rods in crafting.

The strut fabricator recipes have been replaced with rod fabricator recipes. This does mean you can both hand-craft and print rods, but that's probably okay since we want to move away from handcrafting for this sort of thing.


MistakeNot4892:
Farewell struts... an idea whose time has yet to come.

MistakeNot4892 (approved):

PR #5161, opened 2025-10-11
Remove hardcoded /area/space checks

out-of-phaze:
## Description of changes
~~Renames `AREA_FLAG_EXTERNAL` to `AREA_FLAG_SPACE_EXPOSED` to make it clearer what it actually does. Removes it from areas where it makes no sense (exoplanet, fantasy, etc.).~~ This change was removed.
Replaces hardcoded `/area/space` checks in APC and area alarm code with an `always_unpowered` check.
Replaces a hardcoded (and redundant) `/area/space` check in level data code with a `world.area` check instead. Also shuffles that proc around a bit and renames it for better clarity, IMO.
Removes a hardcoded space check in gravity generator code by just using a variable for it.
Removes a hardcoded space check in buildmode area mode by using world.area instead.

Steps towards making it easier for downstreams to de-space the codebase, and maybe even a space modpack.

## Authorship
Me

### DIFF
--- a/code/modules/maps/template_types/random_exoplanet/random_planet_level_data.dm
+++ b/code/modules/maps/template_types/random_exoplanet/random_planet_level_data.dm
@@ -80,7 +80,7 @@
 	//Try to adopt our parent planet's ambient lighting preferences
 	apply_planet_ambient_lighting(parent_planetoid)
 	//Rename the surface area if we have one yet
-	adapt_location_name(parent_planetoid.name)
+	adapt_to_location_name(parent_planetoid.name)
 
 ///If we're getting atmos from our parent planet, apply it.
 /datum/level_data/planetoid/proc/apply_planet_atmosphere(var/datum/planetoid_data/P)
@@ -96,13 +96,12 @@
 	if(!ambient_light_color)
 		ambient_light_level = P.surface_light_color
 
-/datum/level_data/planetoid/adapt_location_name(location_name)
-	if(!(. = ..()))
-		return
+/datum/level_data/planetoid/adapt_to_location_name(location_name)
 	if(!ispath(base_area) || ispath(base_area, world.area))
 		return
 	var/area/A = get_base_area_instance()
 	//Make sure we're not going to rename the world's base area
 	if(!istype(A, world.area))
+		// vvvv This feels bad, should we be doing this? vvvv
 		global.using_map.area_purity_test_exempt_areas |= A.type //Make sure we add any of those, so unit tests calm down when we rename
 		A.SetName("[location_name]")
--- a/code/game/area/areas.dm
+++ b/code/game/area/areas.dm
@@ -45,6 +45,8 @@ var/global/list/areas = list()
 	var/oneoff_light   =      0
 	var/oneoff_environ =      0
 	var/has_gravity =         TRUE
+	/// If FALSE, this area is unable to have its gravity overridden by a gravity generator. Used on /area/space.
+	var/can_have_gravity =    TRUE
 	var/air_doors_activated = FALSE
 
 	var/obj/machinery/apc/apc
--- a/code/game/area/areas.dm
+++ b/code/game/area/areas.dm
@@ -31,7 +31,8 @@ var/global/list/areas = list()
 
 	var/lightswitch =         TRUE
 	var/requires_power =      TRUE
-	var/always_unpowered =    FALSE //this gets overriden to 1 for space in area/New()
+	/// Disables constructing or using APCs in this area.
+	var/always_unpowered =    FALSE
 
 	var/atmosalm =            0
 	var/power_equip =         1 // Status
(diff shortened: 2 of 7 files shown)


code/modules/maps/template_types/random_exoplanet/random_planet_areas.dm
```diff
@@ -10,7 +10,7 @@
 		'sound/effects/wind/wind_5_1.ogg'
 	)
 	always_unpowered = TRUE
-	area_flags = AREA_FLAG_IS_BACKGROUND | AREA_FLAG_EXTERNAL | AREA_FLAG_HIDE_FROM_HOLOMAP
```
MistakeNot4892, line 13: I feel like this should be kept - exoplanet areas are still sun/weather-exposed, so you'd expect to see bleaching on a crashed or landed spaceship for example.

code/modules/maps/template_types/random_exoplanet/random_planet_level_data.dm
```diff
@@ -96,13 +96,12 @@
...
-/datum/level_data/planetoid/adapt_location_name(location_name)
-	if(!(. = ..()))
-		return
+/datum/level_data/planetoid/adapt_to_location_name(location_name)
 	if(!ispath(base_area) || ispath(base_area, world.area))
 		return
 	var/area/A = get_base_area_instance()
 	//Make sure we're not going to rename the world's base area
 	if(!istype(A, world.area))
+		// vvvv This feels bad, should we be doing this? vvvv
```
MistakeNot4892, line 105: I don't think we have an option due to cases where multiple exoplanets with the same base area are created.

code/modules/maps/template_types/random_exoplanet/random_planet_areas.dm
MistakeNot4892: Likewise, exterior shield checks should reasonably be expected to treat exoplanet exteriors as the 'outside'.

maps/shaded_hills/areas/_areas.dm
```diff
@@ -37,4 +37,4 @@
 	description = "Birds and insects call from the grasses, and a cool wind gusts from across the river."
 	area_blurb_category = /area/shaded_hills/outside
 	interior_ambient_light_modifier = -0.4
-	area_flags = AREA_FLAG_EXTERNAL | AREA_FLAG_IS_BACKGROUND
```
MistakeNot4892, line 40: Kind of moot since shadyhills has no shields or spaceship hulls, but I feel like we should keep it flagged as exposed.

maps/~mapsystem/maps_areas.dm
```diff
@@ -1,15 +1,15 @@
 /datum/map
 	/*
 		Areas where crew members are considered to have safely left the station.
-		Defaults to all area types on the admin levels if left empty
+		Defaults to all non-space-exposed area types on the admin levels if left empty
```
MistakeNot4892: Not a blocker, just a thought, but this is such a weird predicate for 'safety' in a post-Centcom world.

mods/content/fantasy/submaps/_submaps.dm
```diff
@@ -44,7 +44,7 @@
 		'sound/effects/wind/wind_5_1.ogg'
 	)
 	interior_ambient_light_modifier = -0.4
-	area_flags = AREA_FLAG_EXTERNAL | AREA_FLAG_IS_BACKGROUND
```
MistakeNot4892, line 47: Same argument as exoplanet and shadyhills flagging.

maps/~mapsystem/maps_areas.dm
out-of-phaze: i figure that as a default it doesn't need to make sense

MistakeNot4892 (changes requested):
Would like a response to the comments even if it's just 'nah', I know we discussed on Discord but would be good to have a record.

out-of-phaze:
> Would like a response to the comments even if it's just 'nah', I know we discussed on Discord but would be good to have a record.

i have plans to do the changes requested or outright drop that commit, i just am too tired to do it rn

code/modules/maps/template_types/random_exoplanet/random_planet_areas.dm
PsyCommando, line 13: It might be good to differentiate an area being exposed to space in space and exposed to space on a planet without atmosphere. Cause, I could see people wanting to make exoplanets without atmosphere eventually.

code/modules/maps/template_types/random_exoplanet/random_planet_level_data.dm
PsyCommando: Yeah, you'd need to rewrite the area purity test to be a bit less strict for certain areas.  Cause that test assumes areas cannot be used as a "template" area. It's a relic from when the only shared area was /area/space afaik.
You could maybe try to create a "temporary" area subtype in here, like the byond map editor does in dmm map files. But, I never tried  it, just heard it was possible to do it in code.

code/modules/maps/template_types/random_exoplanet/random_planet_areas.dm
out-of-phaze: we already have barren exoplanets that randomly generate

MistakeNot4892 (approved):

PR #5166, opened 2025-10-13
Move light fixtures and bulbs/tubes into their own module

out-of-phaze:
Moves light fixtures and bulbs/tubes into their own module. Reorganizes how the lights are laid out, most light fixture/bulb related stuff is inside the lights module now, including light switches and fabricator recipes etc. Also switched it to use `const` instead of defines.


MistakeNot4892 (dismissed):

ike709:
OpenDream CI fixed by https://github.com/OpenDreamProject/OpenDream/pull/2417

MistakeNot4892 (approved):

PR #5168, opened 2025-10-19
Backporting animal ears/tails from Pyrelight as biomod modpack.

MistakeNot4892:
## Description of changes
- Moves elf ears into `biomods` modpack.
- Adds animal ears/tails to `biomods` modpack. 

Tested on tradeship.

Adds some generally popular character customisation as an optional modpack.

## Authorship
Sprites for ears/tails were heavily adapted from various ears/tails from Polaris and Virgo for use on Pyrelight.
Code is mine.

## Changelog
:cl:
add: Added biomods modpack with animal ears and tails for human customisation.
/:cl:


out-of-phaze (approved):

code/modules/mob/mob_snapshot.dm
```diff
@@ -78,6 +78,9 @@
 	target.set_skin_tone(skin_tone)
 
 	for(var/limb_data in extra_limbs)
+		// For whatever reason, we already have a limb in this slot. Creating a new one would cause limb overwrite runtimes.
+		if(target.get_organ(limb_data))
```
out-of-phaze: does this mean we can't overwrite the species tail now?

out-of-phaze (changes requested):

code/modules/mob/mob_snapshot.dm
MistakeNot4892: I have added some handling for removing and replacing non-matching tails which should allow for this. Can't test because we don't have any tails that are available to tailed species at the moment.

out-of-phaze (approved):

PR #5169, opened 2025-10-19
Backporting zombies/skellymans from Pyrelight.

MistakeNot4892:
image

## Description of changes
- Adds an `undead` modpack that includes a trait and many overrides for human mobs to implement zombies and skeletons.
- Moves some drake compat stuff around.

Tested on tradeship and shadyhills.

Useful mobs for populating dungeons etc.

## Authorship
Myself.

## Changelog
:cl:
add: Added zombie/skeleton modpack.
/:cl:


mods/content/undead/mods/undead.dm
```diff
@@ -0,0 +1,10 @@
+/mob/living/human/get_movement_delay(var/travel_dir)
+	. = ..()
+	if(has_trait(/decl/trait/undead))
+		set_moving_slowly()
+		if(istype(default_walk_intent))
+			. = max(., default_walk_intent.move_delay) // no runner zombies yet
```
out-of-phaze, line 6: not sure about this, would maybe prefer pulling out the movement intent initialization from `/mob/living/human/set_species()` and then having this override it so undead don't get any intents with the `MOVE_INTENT_QUICK` flag? or just remove them after `set_species()` is run, or in a helper called between `move_intents = species.move_intents.Copy()` and `set_move_intent(GET_DECL(move_intents[1]))`.

out-of-phaze (changes requested):

mods/content/undead/mods/undead.dm
MistakeNot4892: Per Discord, implementing this to be updated on trait set/removal would be quite painful and I would prefer not to have to mess with movement intent code for zombies.

out-of-phaze (approved):

PR #5170, opened 2025-10-19
Plant chemical composition rework, plant dissection port.

MistakeNot4892:
## Description of changes
- Plant chemicals now use getters/setters.
- Adds `plant_dissection` modpack.
- Adds plant dissection configuration to /datum/seed/flower for harvestable petals as an example implementation.

Good system for survivalist and botany stuff in the future.

## Authorship
Code is largely mine I believe, may have had some input from Penny at some point?

## Changelog
:cl:
add: Added plant dissection modpack.
/:cl:


mods/content/plant_dissection/plants.dm
```diff
@@ -0,0 +1,112 @@
+// TODO: unify with /obj/structure/flora/plant
```
out-of-phaze: this is newer, right? i'd honestly rather merge `flora/plant` into `flora/growing` instead of the other way around. i guess the main difference is the icons anyway

mods/content/plant_dissection/plants_fruit.dm
```diff
@@ -0,0 +1,113 @@
+/obj/item/food/fruit
```
out-of-phaze: possibly able to be mixed up with `/obj/item/food/grown`, unsure what to do about that. `/obj/item/food/grown/fruit`? this seems to be like... totally unrelated to seeds and stuff though, so i don't know what to do about that.

mods/content/plant_dissection/plants_fruit.dm
```diff
@@ -0,0 +1,113 @@
...
+		var/remaining = comp.dissect_amount - LAZYACCESS(removed_segments, comp)
+		if(remaining <= 0 || comp.dissect_tool)
+			continue
+		if(!comp.dissect_skill || !comp.dissect_skill_requirement || user.skill_check(comp.dissect_skill, comp.dissect_skill_requirement))
+			comp.on_harvest(null, user, src)
+			return TRUE
+	return ..()
+
+/obj/item/food/fruit/proc/get_composition()
+	return
```
out-of-phaze: without either comments explaining what this should be structured as, or an example implementation, this is going to be really opaque to people trying to use this modpack as a base.

mods/content/plant_dissection/plants.dm
MistakeNot4892: Main difference is that /growing doesn't have a seed datum associated, and we probably don't want /plant processing on a subsystem.

mods/content/plant_dissection/plants_fruit.dm
MistakeNot4892: Yeah we couldn't make it /grown without an associated seed datum.

MistakeNot4892:
~~I'll revisit this down the track, properly integrating it with plant genetics would be a bit hellish.~~ oops I integrated it with plant genetics

mods/content/plant_dissection/plants_fruit.dm
MistakeNot4892: Ended up doing a much broader merger of this into /datum/seed.

mods/content/plant_dissection/plants_fruit.dm
MistakeNot4892: Doing an example implementation for flowers (very simple) and leaving a comment for a more complex example.

mods/content/plant_dissection/seed_dissection.dm
```diff
@@ -0,0 +1,54 @@
...
+		LAZYADD(., _physical_composition[segment_type])
+
+/datum/seed/flower/New()
+	..()
+	// Adding this soley for the purposes of 'she loves me, she loves me not'
+	set_segment_data(
+		new /datum/plant_segment/petal(
+			"petal",
+			"A small petal from a flower.",
+			rand(3,5)
```
out-of-phaze: won't this mean every single flower will have exactly 3 to 5 petals, rather than having 3 to 5 rolled per plant?

out-of-phaze (changes requested):

mods/content/plant_dissection/plants.dm
out-of-phaze: think this should be revisited post merger?

out-of-phaze (dismissed):

mods/content/plant_dissection/seed_dissection.dm
MistakeNot4892: It'll be rolled per species, but, yes. I could add like a 'pre-remove some segments' step maybe.

MistakeNot4892: This is still WIP regardless, I need to sort put the icons stuff.

MistakeNot4892:
So fun thing I realised about /growing vs /plant - if I'm integrating all the dissection stuff into /datum/seed then the stuff implemented by /growing doesn't need to exist, I can just use /plant. The only addition would be making /plant process, which is already a TODO in that file. So I'll leave it for a future PR.

mods/content/plant_dissection/seed_dissection.dm
MistakeNot4892: Reworked the way growns track segments to allow for variable petals. Wouldn't work with nightweave (or anything that generates the grown icon from segments) but will be fine for this example code.

mods/content/plant_dissection/plants.dm
MistakeNot4892: Left a more detailed comment but yeah removed /growing from this PR.

MistakeNot4892:
Appears to be working. Tested with spawned ambrosiadeus and poppies. Petals got the expected colour/strings/reagents.

out-of-phaze (dismissed):

out-of-phaze (approved):
was approved for like 48h, conflicts resolved

PR #5171, opened 2025-10-19
Simple animal default colours can be retrieved with a helper.

MistakeNot4892:
## Description of changes
- Adds `get_default_animal_colours()` and `get_default_animal_colour()` to simple_animal.
- Moves wolf, rabbit, deer and fox initial colouration into the proc.

Not actually sure if we want this one. Opening it just because it's some general code from Pyrelight that might have utility elsewhere.

## Authorship
Myself.

## Changelog
Nothing player-facing.

### DIFF
--- a/code/modules/mob/living/simple_animal/_simple_animal.dm
+++ b/code/modules/mob/living/simple_animal/_simple_animal.dm
@@ -114,6 +114,10 @@
 /mob/living/simple_animal/Initialize()
 	. = ..()
 
+	if(isnull(draw_visible_overlays))
+		var/list/defaults = get_default_animal_colours()
+		draw_visible_overlays = defaults?.Copy() // do not mutate static list
+
 	if(length(ability_handlers))
 		for(var/handler in ability_handlers)
 			add_ability_handler(handler)
@@ -127,6 +131,9 @@
 		minbodytemp = 0
 
 	check_mob_icon_states(TRUE)
+	if(length(draw_visible_overlays))
+		update_icon()
+
 	if(isnull(base_animal_type))
 		base_animal_type = type
 	if(LAZYLEN(natural_armor))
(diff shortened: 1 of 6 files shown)


code/modules/mob/living/simple_animal/_simple_animal.dm
```diff
@@ -603,3 +607,10 @@ var/global/list/simplemob_icon_bitflag_cache = list()
 
 /mob/living/simple_animal/is_space_movement_permitted(allow_movement = FALSE)
 	return skip_spacemove ? SPACE_MOVE_PERMITTED : ..()
+
+/mob/living/simple_animal/proc/get_default_animal_colour(marking_type)
+	var/list/colors = get_default_animal_colours()
+	return LAZYACCESS(colors, marking_type) || COLOR_BLACK
```
out-of-phaze: Should this default to black, or should it be null to indicate that the marking type doesn't exist?

code/modules/mob/living/simple_animal/passive/horse.dm
```diff
@@ -15,7 +15,7 @@
 	max_rider_size        = MOB_SIZE_MEDIUM
 	ai                    = /datum/mob_controller/passive/horse
 	color                 = "#806146" // preview color
-	draw_visible_overlays = null // e.g. list("base" = "#806146")
+	draw_visible_overlays = list() // to avoid applying any defaults
```
out-of-phaze, line 18: this pairs with lines 35 and 36 further down in this file, where they pick a random color as default. does that still function with these changes?

code/modules/mob/living/simple_animal/_simple_animal.dm
MistakeNot4892: I would argue null

code/modules/mob/living/simple_animal/passive/horse.dm
MistakeNot4892: I think you've misread the horse colour code. Horses do not use the default overlay setup, they have their own weird bespoke version. They could probably be unified but I don't want to do that in this PR.

MistakeNot4892: I may have misread your comment actually.

TLDR: it does still work yes.

MistakeNot4892:
Tested. Rabbit was also in this screenshot but the fox ate it :(
image

out-of-phaze (approved):

PR #5174, opened 2025-10-20
Added octopus modpack.

MistakeNot4892:
## Description of changes
- Unifies wheelchair checking using cannot_stand().
- Adds a limb category for maniple limbs.
- Converts limb categories to use a JSON string to allow for a list of values returned to initial() (a la origin tech).
- Changes human movement delay tallying to use limb categories.
- Adds an octopus species originally ported from Europa.
  - Octopodes can blend in with their turf via gradual alpha changes, move faster in water, and have eight dual-purpose hands/feet.

- Example of a very divergent nonhuman bodyform.
- Example of multiple hand slots.

## TODO
- [X] Update port to modern code.
- [X] Update movement code to use organ categories instead of hardcoded BP lists.
- [X] Redo octopus limbs to be dual-purpose grippers/legs.
- [ ] Fix octopus movement - too slow in water.
- [ ] Add offsets for hats/glasses/etc.
- [x] Redo octopus limbs to have hand slots.
- [ ] Properly forbid octopodes from wearing standard uniforms etc.
- [ ] Add basic gear harness sprite.
- [ ] Add basic backpack sprite.
- [ ] Add basic emergency suit and emergency helmet sprites.

## Authorship
Code and icons are both mine.

## Changelog
:cl:
add: Added octopus species modpack.
/:cl:


code/__defines/damage_organs.dm
```diff
@@ -77,6 +77,8 @@
 #define ORGAN_CATEGORY_STANCE      "stance"
 /// Limb is considered the 'root' of a given stance limb (leg) - also counted for stance damage a la ORGAN_CATEGORY_STANCE
 #define ORGAN_CATEGORY_STANCE_ROOT "stance_root"
+// Limb is considered a manipulator, currently only used when trying to pilot a wheelchair.
+#define ORGAN_CATEGORY_MANIPLE     "maniple"
```
out-of-phaze: frankly i'm not sure why this is necessary? like... we have `get_manual_dexterity()`. i guess you wouldn't want a drake to be able to pilot a wheelchair with their mouth (lol) but... i really feel like it should be handled more elegantly/emergently if possible, like a mix of dexterity (to determine if the organ can even grip the wheel) and some other variable to check if the organ is flexible enough to reach the wheel.

MistakeNot4892:
Depends on #5198

code/__defines/damage_organs.dm
MistakeNot4892: I really, really don't want to commit to a wheelchair rewrite at this point. I can make this iterate inventory slots or something instead in the interim if you don't want me to use an organ category but I don't want to commit more effort than that.

out-of-phaze: it's good as-is, though i'm not super sure why that would require a wheelchair rewrite any more than checking for an organ category would?

MistakeNot4892: Well I've already done that one.

MistakeNot4892:
Depends on #5200

MistakeNot4892:
Every time I work on this PR I have to go listen to https://www.youtube.com/watch?v=iQ9oJslfRSo

PR #5176, opened 2025-10-20
Adds a per-level persistence system.

MistakeNot4892:
## Description of changes
- Adds `/atom/Serialize()` to return an assoc list of relevant values suitable for saving out.
- Adds `/atom/Preload()` and `/atom/PreloadData()` and hooks in SSatoms flush to deserialize atoms during flush.
- Adds `/datum/level_data/load_persistent_data()` and `/datum/level_data/save_persistent_data()` as entrypoints for per-level persistence.
- Adds a fire() override to SSpersistence to do a periodic level data save.
- Converts the innards of the previous persistence system to use the new save/load handlers and procs.

General flow of level persistence:
- Saving:
  - SSpersistence periodically iterates the z-level list, finds levels that want to serde, and calls save_persistent_data()
  - Levels return a list of instances to get_persistent_instances(), instances have Serialize() called and return a list of modified fields.
  - Fields are serialized (to JSON with the default handler) and written to disk.
- Loading:
  - SSmapping initializes and calls preload_persistent_data() and load_persistent_data() on relevant /datum/level_data z-level objects.
  - load_persistent_data() creates the base instances and (for atoms) sets __init_deserialisation_payload with the data loaded from tile.
  - SSatoms flush calls Preload() on all deserialized atoms which pre-populates vars on the atom.
  - Ssatoms proceeds to Initialize() atoms as normal.

## TODO
- ~~Implement area serde.~~
- [X] Implement datum serde.
- [x] Implement non-turf loc for serialized atoms.
- [x] Test with implementation on tradeship.
- [x] Test that wall construction serializes.
- [x] Test that wall removal serializes.
- [x] Test that removing flooring serializes.
- [x] Test that adding flooring serializes.
- [x] Test that roofing turf serializes.
- [X] Test that changing turf height serializes.
- [x] Get simplemob serde to work.
- [x] Test legacy persistence:
  - [x]  Graffiti
  - [x]  Filth
  - [x]  Trash
  - [x]  Paper
  - [x]  Books
- [x]  Implement some kind of grandfather check to convert old persistence format to new.
- [x] Test legacy migration.
  - [x]  Graffiti
  - [x]  Filth
  - [x]  Trash
  - [x]  Paper
  - [x]  Books
- [x] Implement first-run level generation (suspend level gen if we initialized a level via serde)
- ~~Test level generator reordering (split into 2nd PR?)~~ Reverted these changes.

## Future work
- ~~Completely integrate `/decl/persistence_handler` into this system instead of the bespoke serde on SSpersistence.~~ ended up largely doing this in this PR.
- Add a bespoke system for serializing area changes (blueprints etc) on a z-level.

Adds a framework for handling persistence in the future.

## Authorship
Myself.

## Changelog
Nothing player-facing.


MistakeNot4892:
Depends on #5175

code/controllers/subsystems/atoms.dm
```diff
@@ -29,6 +31,13 @@ SUBSYSTEM_DEF(atoms)
 
 	atom_init_stage = INITIALIZATION_INNEW_MAPLOAD
 
+	// Preload any atoms that have deserialized during the initial load process prior to flush.
+	while(length(deserialized_atoms))
+		var/atom/A = deserialized_atoms[deserialized_atoms.len]
+		deserialized_atoms.len--
```
out-of-phaze: for the same reason we don't modify length during the loop in the `created_atoms` loop, i'd suggest not doing so here. iterate from start to end and then do a cut afterwards.

code/controllers/subsystems/initialization/persistence.dm
```diff
@@ -30,9 +32,32 @@ SUBSYSTEM_DEF(persistence)
...
+	while(_persistent_save_running)
+		sleep(1)
+
+/datum/controller/subsystem/persistence/fire(resumed)
+	start_persistent_level_save()
+
+/datum/controller/subsystem/persistence/proc/start_persistent_level_save()
+	// Is this actually desirable? People moving around or modifying
+	// atoms across save could result in inconsistent data.
+	set waitfor = FALSE
```
out-of-phaze: i think you'd have to actually add `if(_persistent_save_running) return` hooks into move and other client-driven/verb things to make it block anyway, is the thing. i'm not sure `waitfor = TRUE` would be enough. so it'd only be a partial solution either way

out-of-phaze (changes requested):

code/controllers/subsystems/initialization/persistence.dm
MistakeNot4892: We could potentially suspend all subsystems and kick people back to the lobby during saves, maybe? Similar to how DMMS suspends fluids and such until the run finishes.

MistakeNot4892:
I'm going to call ~~datum and~~ area serde out of scope for this PR, I don't have any clue how to handle them in a clean way.

The atom side of this appears to be working fine, I've tested with a debug item and it all seems pretty good.

EDIT: Went ahead and just let it instance datums with the serde data as a list arg, in case it's needed in the future.

MistakeNot4892:
Slipped and accidentally added turf serde, limited testing on Shaded Hills but needs proper testing before I sign off on it.

MistakeNot4892:
Depends on #5182

MistakeNot4892:
Testing notes:
- ~~shutter state needs to be serialized on walls~~
- fluids need to be serialized on floors
- flooring is busted

PsyCommando:
Alright so, first, a few things. Mainly about some of the naive mistakes we've made before that I can spot in here.

I really really really recommend making sure whatever your approach is to make sure it scales well with a lot of things to save all of a sudden. Because, you can easily end up with hundred of thousands of atoms to save. Especially if you're saving in real time (for example, someone blows up the whole station, or everything vents to space). Querying over and over would get more and more costly.
It gets surprisingly slow quickly. Especially if you're using byond's IO to disk. 
So, definitely stress test this early on.

Next, error handling. I see you're just kind of erroring out on any exceptions. And it'll come to bite you in the ass later on.
Make sure early on to have a mechanism for tolerating some deserialization errors, and even some serialization errors. Ideally on a per atom basis. Strictly terminating the whole process on a single error is going to make you lose saves over really dumb minor things when running on a real server. 
Like, if for some reasons a single atom on the map that was edited by admins, had it's init code changed, or just bugged out during the game and caused an exception while serializing. And you lose the whole map if your admins aren't coders themselves and realize they can just delete that one thing breaking the save.

Another thing you probably want to do early is allow for runtime generated levels (for example, mining maps or ruins) to possibly be regenerated and not have their state loaded from save. But that will require handling things from the map template level rather than the level data level.

MistakeNot4892:
Thanks for your thoughts - will have a proper read later, but just offhand:

> Next, error handling. I see you're just kind of erroring out on any exceptions.

The exception handling specifically does not terminate the whole process on an error, that's why I've stuffed in so many try/catch blocks. An individual atom will be skipped but the overall process should complete regardless. Only issue I don't have a good solution for is an atom failing to deserialize and then being effectively removed from the save when the serialization pass runs, but that might be desirable.

> Querying over and over would get more and more costly.

I don't really follow what this is referring to. It queries once on roundstart, and then once each save run (which is once an hour on current code, but I'll probably make it a config).
 
> It gets surprisingly slow quickly. Especially if you're using byond's IO to disk.

Likewise, it's only doing I/O once per level per load/save run.

MistakeNot4892:
Depends on #5191, #5190, and #5189.

code/controllers/subsystems/mapping.dm
```diff
@@ -149,10 +149,35 @@ SUBSYSTEM_DEF(mapping)
...
+		stale_data.Cut()
+
+	global.level_persistence_ref_map.Cut()
+
 	// Generate turbolifts last, since away sites may have elevators to generate too.
 	for(var/obj/abstract/turbolift_spawner/turbolift as anything in turbolifts_to_initialize)
 		turbolift.build_turbolift()
 
+	// With levels set up and serde complete (and levels flagged) we can do any remaining level generation.
+	for(var/datum/level_data/level in levels_by_z)
```
PsyCommando: IMO turbolifts should ideally be the latest thing to happen in here. Especially if someone decides to generate stuff with elevators at some point.
We've had issues in the past with turbolifts building happening too soon.

PsyCommando:
> Thanks for your thoughts - will have a proper read later, but just offhand:
> 
No problems! I don't really have that much to say so far. I'll have to see what you do with it later to get a better idea and make some more helpful comments.

> > Next, error handling. I see you're just kind of erroring out on any exceptions.
> 
> The exception handling specifically does not terminate the whole process on an error, that's why I've stuffed in so many try/catch blocks. An individual atom will be skipped but the overall process should complete regardless. Only issue I don't have a good solution for is an atom failing to deserialize and then being effectively removed from the save when the serialization pass runs, but that might be desirable.

Yeah, exceptions don't terminate it, but IIRC, ``PRINT_STACK_TRACE`` terminates the current proc stack to unwind it and write a stack dump? It calls CRASH. 
Shouldn't it just use ``error()`` or rethrow/throw an exception upwards instead?
I remember it being a reason we didn't use ``PRINT_STACK_TRACE`` unless it was a completely unrecoverable problem on persistence cause it prematurely ended the save loop.

Skipping atoms that fail to deserialize is fine tbh. A partially loaded atom might cause issues anyways. As long as it doesn't prevents everything else being loaded it should be good enough.


> 
> > Querying over and over would get more and more costly.
> 
> I don't really follow what this is referring to. It queries once on roundstart, and then once each save run (which is once an hour on current code, but I'll probably make it a config).

Yeah, you're right about the querying. I didn't notice the subsystem would fire once every hour. My bad.

> 
> > It gets surprisingly slow quickly. Especially if you're using byond's IO to disk.
> 
> Likewise, it's only doing I/O once per level per load/save run.

About this, what I meant here is just, make sure writing to json isn't going to take 40 minutes to save several z-levels with stuff in them. Cause, we ran benchmarks on our old save system, and the bulk of the saving time was writing to file and string operations.
Since we began using SQL we bypassed that bottleneck and It can take anywhere from 2-3 minutes to 5 minutes to save a really big map that would take 30+ minutes to save if we had used byond's file IO.

MistakeNot4892:
> Yeah, exceptions don't terminate it, but IIRC, `PRINT_STACK_TRACE` terminates the current proc stack to unwind it and write a stack dump? It calls CRASH.

PRINT_STACK_TRACE is wrapped in a proc specifically so the CRASH doesn't end whatever context it's called it. If it was prematurely ending runs on Persistence then something else was wrong.

PsyCommando:
> > Yeah, exceptions don't terminate it, but IIRC, `PRINT_STACK_TRACE` terminates the current proc stack to unwind it and write a stack dump? It calls CRASH.
> 
> PRINT_STACK_TRACE is wrapped in a proc specifically so the CRASH doesn't end whatever context it's called it. If it was prematurely ending runs on Persistence then something else was wrong.

Hmm, never mind that then. I'd have to check in a live context again. But I remember it causing issues before, and not giving a stack trace most of the time.

code/game/turfs/turf_serde.dm
```diff
@@ -0,0 +1,60 @@
...
+
+/turf/proc/UnpackSerializableInstances()
+	// Get all recursively nested instances on this turf.
+	var/list/instances_to_unpack = list(src)
+	while(length(instances_to_unpack))
+		var/datum/instance = instances_to_unpack[1]
+		instances_to_unpack.Cut(1, 2)
+		if(instance in .)
+			continue
+		LAZYDISTINCTADD(., instance)
```
out-of-phaze: i don't think the `LAZYDISTINCTADD` is necessary here since we already do an existence-in-list check on the line before, this code will never run if it isn't distinct

code/modules/multiz/level_data.dm
```diff
@@ -231,16 +231,14 @@
...
-/datum/level_data/proc/setup_level_data(var/skip_gen = FALSE)
+/datum/level_data/proc/setup_level_data()
+
 	if(_level_setup_completed)
 		log_debug("level_data for [src], on z [level_z], had setup_level_data called more than once!")
 		return //Since we can defer setup, make sure we only setup once
 
 	setup_level_bounds()
 	setup_strata()
-	if(!skip_gen)
```
PsyCommando, line 241: I'd look into whether this breaks anything tbh.  It miiiiight mess with exoplanet gen iirc.  If anything is in there, it was probably timing sensitive.

code/modules/multiz/level_persistence_serialization.dm
```diff
@@ -0,0 +1,98 @@
...
+// - Ssatoms proceeds to Initialize() atoms as normal.
+
+var/global/list/level_persistence_ref_map = list()
+/datum/level_data
+	/// String pointing to a base directory on the filesystem for use in serde.
+	/// If set, will automatically suffix map path and level name.
+	/// Leave null to opt out of any persistence for this level.
+	// Example setting would be:
+	//   persistent_data_location = "data/level_data"
+	VAR_PROTECTED/persistent_data_location
```
PsyCommando, line 20: Shouldn't that be abstracted out? Like, the level data probably shouldn't store a file path, since that's implementation specific to the serializer?

code/modules/multiz/level_persistence_serialization.dm
```diff
@@ -0,0 +1,98 @@
...
+	/// Decl handler, mostly forcing myself to keep this general so it can be optimized with a DB or something down the track.
+	VAR_PRIVATE/persistence_handler = /decl/serialization_handler/json
```
PsyCommando, line 22: Similarly as my last comment, shouldn't the way the level is saved really be stored in the level_data? Because, it doesn't really make sense to save each levels with a different serializer. It should be something decided at the start of the serialization imo.

code/__defines/serde.dm
PsyCommando, line 1: A bit nitpicky, but serde is a bit ambiguous?
Shouldn't it be called SerDes (SerializationDeserialization)? Or anything else? Cause, that's the only form of the term serde I could find when doing a websearch. https://en.wikipedia.org/wiki/SerDes

Serde itself gets me to the serde rust library page. Which might imply that the library might be wrapped in there.
Again, that's just a nitpick.

code/modules/multiz/level_persistence_serialization.dm
MistakeNot4892: It isn't a file path, it's a general location. The serialization handler then determines what is actually done in that location. You could set it to a DB name if you wanted the handler to do something other than use the filesystem.

code/modules/multiz/level_persistence_serialization.dm
MistakeNot4892: The serialization logic doesn't really care where the data comes from, once it's in DM it's all in the same format. I don't really think it matters much. The alternative would be forcing the entire game to use the same handler, which to me limits the ability for people to mix and match or run their own local changes alongside general changes from upstream. 

What would the benefit be to forcing a single handler for all cases?

code/__defines/serde.dm
MistakeNot4892: I was actually under the impression it was a general term, probably because of Lohi using it as one on Neb way back when. Seems like serdes is also a specific term not related to what we're doing. A dilemma.

code/modules/multiz/level_data.dm
MistakeNot4892: It's on the testing list.

out-of-phaze (changes requested):

code/__defines/serde.dm
out-of-phaze: there's also serde packages for go and python that are unrelated. i believe it's just a generic term that's been used by a number of high profile serialization-deserialization packages. we could take a page out of python's playbook and call it pickling, but i think serde is fine, it gets the idea of ser(ialization)/de(serialization) across fine

code/modules/multiz/level_persistence_serialization.dm
out-of-phaze: i could see using a different handler for, specifically, custom ships vs a main map

code/modules/multiz/level_persistence_serialization.dm
MistakeNot4892: I'll update the comment.

code/game/atoms_serde.dm
```diff
@@ -0,0 +1,82 @@
...
+		SERIALIZE_IF_MODIFIED(temperature, /atom)
+	if(istype(reagents))
+		SERIALIZE_REAGENTS(reagents, /atom, "atom")
+	SERIALIZE_DECL_IF_MODIFIED(material, /atom)
+	SERIALIZE_DECL_IF_MODIFIED(reinf_material, /atom)
+	SERIALIZE_IF_MODIFIED(paint_color, /atom)
+	SERIALIZE_IF_MODIFIED(pixel_x, /atom)
+	SERIALIZE_IF_MODIFIED(pixel_y, /atom)
+	SERIALIZE_IF_MODIFIED(default_pixel_x, /atom)
+	SERIALIZE_IF_MODIFIED(default_pixel_x, /atom)
```
out-of-phaze:
```suggestion
	SERIALIZE_IF_MODIFIED(default_pixel_y, /atom)
```

code/game/atoms_serde.dm
```diff
@@ -0,0 +1,82 @@
+/atom
+	/// Var for holding serde information when this atom was loaded from a persistent source.
+	var/__init_deserialization_payload
+
+/atom/Serialize()
+	. = ..()
+	if(current_health != max_health)
```
out-of-phaze:
```suggestion
	if(current_health != get_max_health())
```

code/game/atoms_serde.dm
```diff
@@ -0,0 +1,82 @@
...
+		SERIALIZE(current_health, /atom)
+	SERIALIZE_IF_MODIFIED(max_health, /atom)
+	SERIALIZE_IF_MODIFIED(dir, /atom)
+	SERIALIZE_IF_MODIFIED(current_health, /atom)
```
out-of-phaze:
```suggestion
```

code/_helpers/serde.dm
```diff
@@ -0,0 +1,131 @@
...
+
+	// Now that everything is loaded and placed, clear out anything that should not be present on the turfs we've loaded.
+	for(var/uid in SSatoms.deserialized_atoms)
+		var/turf/turf = SSatoms.deserialized_atoms[uid]
+		if(!istype(turf))
+			continue
+		for(var/atom/thing in turf)
+			if(!thing.simulated)
+				continue
+			if(thing in SSatoms.deserialized_atoms)
```
out-of-phaze: for some reason this isn't working, that's why david hasselhoff is disappearing. i checked `thing.__init_deserialization_payload` instead which i guess won't work for stuff without a payload, but it should be good enough to stop your guy from being deleted

code/game/objects/__objs.dm
```diff
@@ -264,15 +267,21 @@
...
 	SHOULD_CALL_PARENT(TRUE)
-	if(reagents?.total_volume > 0)
+	// Check if this is getting called twice, or we created reagents somewhere in Initialize() (bad juju)
+	if(istype(reagents))
 		log_warning("\The [src] possibly is initializing its reagents more than once!")
-	create_or_update_reagents(chem_volume)
-	if(populate)
+	// If preloaded from serde, handle expected list structure.
+	// Returns if preload is successful to skip populate_reagents() call.
+	FINALIZE_REAGENTS_SERDE_AND_RETURN(reagents)
```
out-of-phaze, line 280: think occasionally `update_alpha_mask` on the fluid overlay can run before this does, which is a little concerning. led to runtimes locally

out-of-phaze (changes requested):

code/_helpers/serde.dm
MistakeNot4892: Per Discord, it's because `deserialized_atoms` has a string key. I think your solution works fine, a serde atom with no payload runtimes anyway.

code/game/objects/__objs.dm
MistakeNot4892: Should be fixable by istype(neighbor.reagents) in that proc

MistakeNot4892:
I'm not really sure what to do about randomly generated levels. I might pull the reordering logic out of this PR and just prevent placing more mobs if we deserialized. Otherwise we end up with mask turfs, etc. left on the map.

code/modules/multiz/level_data.dm
MistakeNot4892: As far as I can tell the order-sensitive stuff is fine with this separation/reorder. I haven't tested exhaustively though.

MistakeNot4892: Scratch that, it looks like ruins are kinda cooked.

out-of-phaze:
> I'm not really sure what to do about randomly generated levels. I might pull the reordering logic out of this PR and just prevent placing more mobs if we deserialized. Otherwise we end up with mask turfs, etc. left on the map.

should we not just automatically mark those turfs as serialized?

code/modules/multiz/level_data.dm
PsyCommando: Honestly, it's probably a good idea to move that gen stuff outside the level_data anyways. Personally, I kinda intended on doing that. It's getting a bit bloated in there for what's essentially just supposed to keep metadata about a z-level. And it would probably be a lot easier to manage the ordering sensitive stuff if it was all in the same place as the rest of the level gen stuff imo.

code/modules/multiz/level_persistence_serialization.dm
PsyCommando: Right, but, I just don't understand why you need to store a variable unique to each z-level? Especially since, you'd probably want a lot of them to share the same base general location path?

Wouldn't it make more sense for it to be figured out at runtime when saving/loading by the serialization code to keep a coherent structure, and limit the chances for errors and the need to mass replace values? I mean, you do what you want, but that feels like the kind of things that would come bite me in the ass later on.

code/modules/multiz/level_persistence_serialization.dm
PsyCommando: > The serialization logic doesn't really care where the data comes from, once it's in DM it's all in the same format. I don't really think it matters much. The alternative would be forcing the entire game to use the same handler, which to me limits the ability for people to mix and match or run their own local changes alongside general changes from upstream.

Yeah, but why should levels each individually chose to save in a different format? I don't really see a point to that? Because, you really want the same kind of data to be in the same format for coherence's sake and maintainability. 

Also, what if you host a server, and you used json to store the save, but now your setup changed and you wanna use a DB. You'd need to change every single levels to the db handler and recompile. While, if you'd use all the same handler for levels you could easily make it a config entry, or call the serialization proc with a different handler specified and not have to mess with the code or do mass replace and risk making mistakes.

Unless I really don't understand what you mean/intend to do here?

> What would the benefit be to forcing a single handler for all cases?

It would be decided at the time the serialization proc is called ideally. And it would be less error prone, much more predictable, easier to maintain, simpler to back up, etc.. It would also allow for consolidation of saved level data later on.

PsyCommando: > i could see using a different handler for, specifically, custom ships vs a main map

Yeah, but, depending on why you'd want a different handler, you probably would need extra metadata or extra processing for saving something to be used separately from the main save. Especially with anything multi-z.

Like, for example, I'll assume what you're thinking of is to move a ship and crew to another server via export or something else. 
You'd need to make substantial changes to do that, and probably need a separate mechanism called on demand specifically for saving and moving ships and sending the save. 
It would be less of a bother to just specify what handler to use when launching serialization, and what you want to serialize rather than making sure all ships being transferred have the right handler set and figure out what should even be done in the case they have the wrong handler?

code/__defines/serde.dm
PsyCommando: Honestly, the first time I heard the term was when you guys used it to talk about stuff about persistence. Buuut then again, I'm probably not the best reference for what's a common term or not.

code/modules/multiz/level_data.dm
MistakeNot4892: I mean, you're free to do that, but I can't see why you would put it somewhere else when levelwide gen is currently encapsulated per level. The ordering issue here is solely because exoplanet gen is overly complex/brittle. I'm not even certain the issues are due to my changes yet.

code/modules/multiz/level_persistence_serialization.dm
MistakeNot4892: I don't understand your argument. You set it to a root value and the handler then uses it as a spot to do whatever it likes, allowing you flexibility if you want to implement your own handler or group your save data by region. How does it cause errors or the need for mass value replacement? How would you even 'figure it out' at runtime?

code/modules/multiz/level_persistence_serialization.dm
MistakeNot4892: I am getting increasingly confused by some of these comments. Having the ability to specify handler per level really doesn't impose any overhead or constraints that you wouldn't have by having it as a param or such. If you decide to change your data format you have a lot more problems than changing a handful of compile time values (such as having to manually migrate all your data). 

Making it a config value or such just means there's no capacity whatsoever to have variance. If I wanted to save custom ships to JSON so players could save and load their layout privately, that wouldn't be possible. If I wanted to have large level serde handled by a DB for performance but some other aspect handled by JSON for some other reason, that wouldn't be possible.

If you don't want to use multiple handlers, don't change the value ever and you're golden. I really cannot see a benefit to centralising and restricting things like you seem to be asking for here and earlier.

code/modules/reagents/Chemistry-Holder.dm
out-of-phaze, line 1: it looks like the most recent commit actually un-privated vars on `/datum/reagents`

code/modules/reagents/Chemistry-Holder.dm
```diff
@@ -73,6 +75,11 @@ var/global/datum/reagents/sink/infinite_reagent_sink = new
 	var/atom/my_atom
 	var/cached_color
```
out-of-phaze:
```suggestion
	VAR_PRIVATE/list/tmp/reagent_volumes
	VAR_PRIVATE/list/liquid_volumes
	VAR_PRIVATE/list/solid_volumes // This should be taken as powders/flakes, rather than large solid pieces of material.
	VAR_PRIVATE/list/reagent_data
	VAR_PRIVATE/atom/my_atom
	VAR_PRIVATE/tmp/cached_color
	VAR_PRIVATE/tmp/primary_reagent
	VAR_PRIVATE/tmp/primary_solid
	VAR_PRIVATE/tmp/primary_liquid
	VAR_PRIVATE/tmp/total_volume = 0
	VAR_PRIVATE/tmp/total_liquid_volume // Used to determine when to create fluids in the world and the like.
	VAR_PRIVATE/maximum_volume = 120
	VAR_PRIVATE/tmp/max_solvent_power
```
it would be nice if temporary/non-serialized vars were marked as such

code/__defines/serde.dm
```diff
@@ -0,0 +1,63 @@
...
+	if((SERDE_REAGENT_LIST in V) && (SERDE_REAGENT_VOLUME in V)) {                         \
+		var/list/LV = V;                                                                   \
+		var/__serde_volume = LV[SERDE_REAGENT_VOLUME];                                     \
+		if(__serde_volume <= 0) {                                                          \
+			V = null;                                                                      \
+		} else {                                                                           \
+			var/list/__serde_reagents = LV[SERDE_REAGENT_LIST];                            \
+			V = new /datum/reagents(__serde_volume, src);                                  \
+			for(var/list/entry in __serde_reagents) {                                      \
+				V.add_reagent(RESOLVE_TO_DECL(entry[1]), entry[2], phase = entry[3]);      \
```
out-of-phaze: we may want to have `defer_update = TRUE` in this and then explicitly call handle_update after

out-of-phaze (changes requested):
needs rebase

MistakeNot4892:
Current outstanding issue with this one is that I cannot for the life of me seem to get map templates placed during init to flag the turfs as changed. Generating on Shaded Hills and restarting leaves template-shaped sections of mask turfs. Not sure what the issue is, it happened even after I made all ChangeTurf() calls and relevant sections of DMMS flag the turfs directly. I am assuming that something is newing /turf directly and not flagging.

out-of-phaze:
> Current outstanding issue with this one is that I cannot for the life of me seem to get map templates placed during init to flag the turfs as changed. Generating on Shaded Hills and restarting leaves template-shaped sections of mask turfs. Not sure what the issue is, it happened even after I made all ChangeTurf() calls and relevant sections of DMMS flag the turfs directly. I am assuming that something is newing /turf directly and not flagging.

does `/area/fantasy/outside/point_of_interest` have AREA_FLAG_ALLOW_LEVEL_PERSISTENCE? it doesn't seem so, that'd mean at least a few non-passthrough templates (chemistry shack etc) would fail to save

MistakeNot4892:
> does `/area/fantasy/outside/point_of_interest` have AREA_FLAG_ALLOW_LEVEL_PERSISTENCE? it doesn't seem so, that'd mean at least a few non-passthrough templates (chemistry shack etc) would fail to save

Genuinely no idea why I didn't think of this. Thank you :(

MistakeNot4892:
This is now working as far as I can tell. Main things I am worried about in live testing is accidentally cooking legacy persistence data, but it should at least write out a backup before it does any kind of migration of that data.

I am not committing any changes that actually use this system in full yet, I'm going to save that for after we've done playtesting (probably via Pyrelight or maybe Scav)

MistakeNot4892:
Realised belatedly that I need to get area serde in before this will work as intended, due to map templates setting area, but I will do that in a second PR, I don't want to fiddle with this one further currently.

out-of-phaze (dismissed):
i frankly really don't like that we still have the legacy persistence system, it really doesn't work for anything but papers and dirt (the generic 'filth' stuff was extremely annoying). but i guess some maps want a middle-ground re: persistence.

once this is in i'll try to make it faster

code/datums/datum.dm
```diff
@@ -9,6 +9,8 @@
 	var/list/active_timers
 	/// Used to avoid unnecessary refstring creation in Destroy().
 	var/tmp/has_state_machine = FALSE
+	/// Var for holding a unique-to-this-run identifier for a serialized datum.
+	VAR_PRIVATE/__run_uid
```
out-of-phaze: this could use `/tmp/` possibly

PsyCommando: Yes please. I really wish ``/tmp`` was used more. It helps make transient vars a whole lot more obvious.

out-of-phaze (dismissed):
was approved prior to conflict resolution, has conflicts again (my bad), if you fix conflicts i'll merge

out-of-phaze (approved):

PR #4969, opened 2025-03-24
Remove or implement unused vars

out-of-phaze:
## Description of changes
i will do a commit by commit explanation if needed but god i'm tired and lazy so not now

needs combing through to pull stuff out to stable/staging where necessary, e.g. ion thruster stuff probably needs to be pulled to stable, that was some ancient merge skew

## Why and what will this PR improve
code quality improvements, implement some stuff that was missing/broken/unfinished

PR #5180, opened 2025-10-20
Implement sprite accessory metadata sanitize_data proc

out-of-phaze:
Split out of #4969.
`metadata.sanitize_data` should now handle default values for sprite accessory metadata, as well as validation failures. This is basically just what some other code does but in its own helper.
This was tested when it was done in #4969 but it hasn't been tested over several months' worth of rebases, so I'll declare this one untested.


MistakeNot4892 (approved):
Looks pretty straightforward.

PR #5185, opened 2025-10-29
Drake and digging tweaks.

MistakeNot4892:
## Description of changes
Several drake fixes, a few general tweaks, and some changes to allow digging vertically down through turfs and digging ramps from above. Commit list should be fairly instructive. 

Nicer drake gameplay, nicer mining gameplay.

## Authorship
Myself.

## Changelog
:cl:
tweak: Drakes can no longer farm, but they can mine (in dirt).
add: Digging a maximum-depth trench on a multi-z level will allow you to dig an open space.
add: Alt-clicking on an open space over a wall will allow you to dig a ramp.
/:cl:

### DIFF
--- a/code/datums/extensions/abilities/abilities_predator.dm
+++ b/code/datums/extensions/abilities/abilities_predator.dm
@@ -56,7 +56,8 @@
 		to_chat(user, SPAN_WARNING("\The [victim] is too big for you to dismember."))
 		return TRUE
 
-	var/obj/item/organ/external/limb = victim.get_organ(user.get_target_zone())
+	var/target_zone = user.get_target_zone()
+	var/obj/item/organ/external/limb = victim.get_organ(target_zone)
 	if(!limb)
 		to_chat(user, SPAN_WARNING("\The [victim] is missing that limb!"))
 		return TRUE
@@ -65,13 +66,17 @@
 	if(!do_after(user, max(2 SECONDS, victim.get_object_size() * 5), victim) || QDELETED(victim) || !victim.butchery_data || victim.stat != DEAD)
 		return TRUE
 
+	// Changing zone means we cancel.
+	if(target_zone != user.get_target_zone())
+		return
+
 	var/list/external_organs = victim.get_external_organs()
 	if(length(external_organs) <= 1)
 		user.visible_message(SPAN_DANGER("\The [user] tears \the [victim] apart!"))
 		victim.gib()
 		return TRUE
 
-	limb = victim.get_organ(user.get_target_zone()) // In case we changed zone or such in the meantime.
+	limb = victim.get_organ(target_zone) // In case it was removed in the interim.
 	if(!limb)
 		to_chat(user, SPAN_WARNING("\The [victim] is missing that limb!"))
 		return TRUE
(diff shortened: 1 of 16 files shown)


code/datums/extensions/abilities/abilities_predator.dm
```diff
@@ -66,13 +71,24 @@
 		victim.gib()
 		return TRUE
 
-	var/obj/item/organ/external/limb = victim.get_organ(user.get_target_zone())
+	limb = victim.get_organ(user.get_target_zone()) // In case we changed zone or such in the meantime.
```
out-of-phaze: personally i'd rather make it fail if we change zone, and instead have this check be to make sure someone else didn't destroy/remove the limb while we were in the do_after. but this is probably fine as is

code/game/turfs/floors/floor_digging.dm
```diff
@@ -25,22 +25,25 @@
...
 	var/decl/flooring/flooring = get_topmost_flooring()
 	if(!flooring.handle_turf_digging(src))
 		return
 	// Only drop mats if we actually changed the turf height sufficiently.
 	var/old_height  = get_physical_height()
 	var/new_height  = max(old_height-TRENCH_DEPTH_PER_ACTION, -(FLUID_DEEP))
-	var/height_diff = abs(old_height-new_height)
-	if(height_diff >= TRENCH_DEPTH_PER_ACTION)
+	var/height_diff = round(abs(old_height-new_height))
+	if(height_diff == 0)
```
out-of-phaze: this confused me at first, i was expecting a `new_height == -(FLUID_DEEP)` check, but i think this is fine. at most i'd maybe suggest a comment explaining the check (we're at max depth so we can't dig any lower)

code/game/turfs/floors/floor_digging.dm
```diff
@@ -25,22 +25,25 @@
...
+		if(!HasBelow(z))
+			to_chat(user, SPAN_WARNING("You cannot dig any lower!"))
+			return
+		to_chat(user, SPAN_DANGER("You break through \the [src]!"))
+		drop_diggable_resources(user)
+		ChangeTurf(/turf/open)
```
out-of-phaze: i was looking into this yesterday, pretty sure this may need to be `area.open_turf || open_turf_type` or similar. although, `/turf/floor`'s `open_turf_type` is `/turf/open/airless` and i. don't know if that's right?

out-of-phaze (dismissed):

code/game/turfs/floors/floor_digging.dm
MistakeNot4892: It shouldn't really matter since ChangeTurf usually persists air contents of the turf as far as I know.

code/datums/extensions/abilities/abilities_predator.dm
MistakeNot4892: Changed.

code/game/turfs/floors/floor_digging.dm
MistakeNot4892: There's actually a check/message earlier than this that prints the user a message about being able to dig no further. I feel like the 'you can't dig any lower' message is sufficient for code readability to be honest.

code/game/turfs/floors/floor_digging.dm
MistakeNot4892: Added a new helper patterned after the base turf proc to grab the area or turf open type.

out-of-phaze (approved):

PR #5186, opened 2025-10-29
Update BYOND and SpacemanDMM versions in CI

out-of-phaze:
updates byond version to ~~latest~~ 516.1663, because 516.1664 adds a dependency on libcurl4(-dev?) which isn't included by default
updates spacemandmm version to latest


MistakeNot4892:
Seems to be unable to find a library on latest Ubuntu?

MistakeNot4892 (approved):

PR #5189, opened 2025-10-31
Disambiguates and unifies volume vars.

MistakeNot4892:
## Description of changes
- Various uses of `volume` changed to more specific names (like `gas_volume` and `total_volume`)
- `/obj/item/chem` volume var converted to `chem_volume` on `/obj`.
- Various bespoke volume vars outside of `/obj/item/chem` converted to `chem_volume`.
- `/obj/Initialize()` will call `initialize_reagents()` if `chem_volume` > 0.
- Many `initialize_reagents()` overrides condensed into `Initialize()` or simply setting `chem_volume`.
- Changes several uses of `volume` to properly refer to `reagents.maximum_volume`.
- Formalizes the 'dummy chem holder' idea as `/obj/effect/chem_holder`.
- Turns most uses of `create_reagents()` into `chem_volume` setting.
- Adds `create_or_update_reagents()` to replace most remaining uses of `create_reagents()`

- Unifies chem reagent holder creation on /obj.
- Simplifies reagent serde (future work).
- Makes it a lot less annoying to tell at a glance what `volume` is referring to in a given context.

## TODO
- [x] Test beakers in IVs and ChemMaster.
- [x] Fix pill name behavior to allow serde.

## Authorship
Myself.

## Changelog
Nothing player-facing.

### DIFF
--- a/code/datums/extensions/milkable/milkable.dm
+++ b/code/datums/extensions/milkable/milkable.dm
@@ -85,7 +85,7 @@
 			to_chat(user, SPAN_WARNING("Wait for \the [critter] to stop moving before you try milking it."))
 			return TRUE
 
-	if(container.reagents.total_volume >= container.reagents.maximum_volume)
+	if(REAGENTS_FREE_SPACE(container.reagents) <= 0)
 		to_chat(user, SPAN_WARNING("\The [container] is full."))
 		return TRUE
 
@@ -112,7 +112,7 @@
 		to_chat(user, SPAN_WARNING("\The [critter]'s udder is dry. Wait a little longer."))
 		return TRUE
 
-	if(container.reagents.total_volume >= container.reagents.maximum_volume)
+	if(REAGENTS_FREE_SPACE(container.reagents) <= 0)
 		to_chat(user, SPAN_NOTICE("\The [container] is full."))
 		return TRUE
 
--- a/code/datums/extensions/milkable/milkable.dm
+++ b/code/datums/extensions/milkable/milkable.dm
@@ -85,7 +85,7 @@
 			to_chat(user, SPAN_WARNING("Wait for \the [critter] to stop moving before you try milking it."))
 			return TRUE
 
-	if(container.reagents.total_volume >= container.volume)
+	if(container.reagents.total_volume >= container.reagents.maximum_volume)
 		to_chat(user, SPAN_WARNING("\The [container] is full."))
 		return TRUE
 
@@ -112,7 +112,7 @@
 		to_chat(user, SPAN_WARNING("\The [critter]'s udder is dry. Wait a little longer."))
 		return TRUE
 
-	if(container.reagents.total_volume >= container.volume)
+	if(container.reagents.total_volume >= container.reagents.maximum_volume)
 		to_chat(user, SPAN_NOTICE("\The [container] is full."))
 		return TRUE
 
--- a/code/game/machinery/alarm.dm
+++ b/code/game/machinery/alarm.dm
@@ -272,7 +272,7 @@
 			environment.merge(gas)
 
 /obj/machinery/alarm/proc/overall_danger_level(var/datum/gas_mixture/environment)
-	var/partial_pressure = R_IDEAL_GAS_EQUATION*environment.temperature/environment.air_volume
+	var/partial_pressure = R_IDEAL_GAS_EQUATION*environment.temperature/environment.total_volume
 	var/environment_pressure = environment.return_pressure()
 
 	var/other_moles = 0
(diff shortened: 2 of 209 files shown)


code/game/objects/items/weapons/storage/fancy/cigar.dm
```diff
@@ -15,11 +15,7 @@
 	storage = /datum/storage/box/cigar
 
 /obj/item/box/fancy/cigar/Initialize(ml, material_key)
-	. = ..()
-	initialize_reagents()
-
-/obj/item/box/fancy/cigar/initialize_reagents(populate)
-	create_reagents(10 * max(1, storage?.storage_slots))
+	chem_volume = 10 * max(1, /datum/storage/box/cigar::storage_slots)
```
out-of-phaze: i think this won't work for subtypes that use a larger or smaller storage subtype, idk if that's a concern. before `..()` in Initialize `storage`, if non-null, is a path, so we might be able to just do `storage ? storage::storage_slots : null` with a comment explaining `storage` is a path or null pre-init

code/game/objects/structures/fires.dm
```diff
@@ -374,12 +374,12 @@
 			if(reagent.accelerant_value <= FUEL_VALUE_SUPPRESSANT && reagent.phase_at_temperature(get_effective_burn_temperature(), ambient_pressure) == MAT_PHASE_GAS)
 				do_steam = TRUE
 
-			var/volume = NONUNIT_CEILING(REAGENT_VOLUME(reagents, reagent) / REAGENT_UNITS_PER_GAS_MOLE, 0.1)
-			var/list/waste_products = burn_material(reagent, volume)
+			var/result_volume = NONUNIT_CEILING(REAGENT_VOLUME(reagents, reagent) / REAGENT_UNITS_PER_GAS_MOLE, 0.1)
```
out-of-phaze: this is not volume, it's mols, since we have  `// We assume a normalized mole amount for 'amount'.` in `get_burn_products()` (which `burn_material()` passes to)

code/game/objects/structures/fires.dm
```diff
@@ -374,12 +374,12 @@
...
+			var/list/waste_products = burn_material(reagent, result_volume)
 			if(!isnull(waste_products))
 				for(var/product in waste_products)
 					waste[product] += waste_products[product]
-				reagents.remove_reagent(reagent.type, volume)
+				reagents.remove_reagent(reagent.type, result_volume)
```
out-of-phaze: it's technically not a change, but i don't think this is actually correct? `volume` isn't volume, it's a (supposed) mol amount, which is the volume divided by 10. that means the reagents will last 10x longer than they should. maybe remove `moles_used * REAGENT_UNITS_PER_GAS_MOLE` units?

out-of-phaze (changes requested):
this pr is a bitch to review but i think i caught some of the important stuff at least. i'll keep going later

code/datums/extensions/milkable/milkable.dm
```diff
@@ -85,7 +85,7 @@
 			to_chat(user, SPAN_WARNING("Wait for \the [critter] to stop moving before you try milking it."))
 			return TRUE
 
-	if(container.reagents.total_volume >= container.volume)
+	if(container.reagents.total_volume >= container.reagents.maximum_volume)
```
out-of-phaze: in theory this could probably just be `REAGENTS_FREE_SPACE(container.reagents) <= 0`? that might be more idiomatic/readable but i'm not sure it matters

code/game/objects/items/weapons/storage/fancy/cigar.dm
MistakeNot4892: `storage` is a datum at this point, it's initialized in `New()`, I just didn't think it mattered as no subtypes change `storage`.

code/game/objects/structures/fires.dm
MistakeNot4892: I don't want to start reworking actual values in this PR if that's alright.

code/game/machinery/alarm.dm
```diff
@@ -272,7 +272,7 @@
 			environment.merge(gas)
 
 /obj/machinery/alarm/proc/overall_danger_level(var/datum/gas_mixture/environment)
-	var/partial_pressure = R_IDEAL_GAS_EQUATION*environment.temperature/environment.volume
+	var/partial_pressure = R_IDEAL_GAS_EQUATION*environment.temperature/environment.air_volume
```
PsyCommando: You know, I feel like, calling this var ``air_volume`` is a bit weird when the datum's name is ``gas_mixture``? It's probably one of those things that kinda made sense to be kept as ``volume`` to make it less redundant.  Or at least call it ``gas_volume`` imo?

MistakeNot4892: Leaving it as `volume` defeats the purpose of disambiguating it (annoying to search and sometimes ambiguous). `total_volume` would at least be appropriate for gasmixes.

out-of-phaze (dismissed):

out-of-phaze (approved):

PR #5190, opened 2025-10-31
Reworking some really bad code in borg hypos.

MistakeNot4892:
Reworks some old crap code.


MistakeNot4892:
Tested, appears to be working.

out-of-phaze (approved):

PR #5191, opened 2025-10-31
Reworks sealant gun code to use reagents instead of bespoke tracking vars.

MistakeNot4892:


MistakeNot4892:
Depends on #5189

code/modules/sealant_gun/sealant_gun.dm
```diff
@@ -58,7 +58,7 @@
 	. = ..()
 	if(loc == user)
 		if(loaded_tank)
-			. += SPAN_NOTICE("The loaded tank has about [loaded_tank.foam_charges] liter\s of sealant left.")
+			. += SPAN_NOTICE("The loaded tank has about [REAGENT_VOLUME(loaded_tank.reagents, /decl/material/liquid/foam) || 0] liter\s of sealant left.")
```
out-of-phaze: i know it's mostly like, opaque and not exposed to the user, but 1u = 1L is a new level of unit nonsense

out-of-phaze (approved):

PR #5193, opened 2025-10-31
Flooring lists are now handled as JSON text.

MistakeNot4892:
This makes it possible to properly initial() flooring values and is useful for serde.


MistakeNot4892:
I don't understand why this compiles fine on the persistence branch but doesn't compile here.

MistakeNot4892:
Depends on #5182

out-of-phaze (changes requested):
`VAR_PROTECTED/list/decl/flooring/_flooring` is an incorrect var typing pls fix

out-of-phaze (approved):
disregard last review

PR #5195, opened 2025-10-31
Moves material references onto /atom

MistakeNot4892:
## Description of changes
- `material` and `reinf_material` are now on `/atom`. Actually initializing/using the vars is handled on the respective types, only the references are shared.
- Bunch of 'material' args and vars renamed for clarity.
- Replaces `floor_material` on /turf/floor with `material`.

Makes material serde a lot more straightforward and eases future work around unifying materials.

## Authorship
Myself.

## Changelog
Nothing player-facing.

### DIFF
--- a/code/game/machinery/doors/airlock.dm
+++ b/code/game/machinery/doors/airlock.dm
@@ -42,7 +42,7 @@
 	var/aiDisabledIdScanner = 0
 	var/aiHacking = 0
 	autoclose = 1
-	var/mineral = null
+	var/mineral = null // TODO: replace with material and get_material()
 	var/justzap = 0
 	var/safe = 1
 	var/speaker = 1
--- a/code/game/machinery/doors/airlock.dm
+++ b/code/game/machinery/doors/airlock.dm
@@ -15,14 +15,12 @@
 	icon_state = "preview"
 	power_channel = ENVIRON
 	interact_offline = FALSE
-
 	explosion_resistance = 10
-
 	base_type = /obj/machinery/door/airlock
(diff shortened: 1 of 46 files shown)


MistakeNot4892:
This one spiralled a bit, need to test the airlock changes in particular.

MistakeNot4892:
Tested, working.

code/game/machinery/doors/airlock.dm
```diff
@@ -324,15 +314,15 @@ About the new airlock wires panel:
...
+	if(door_color)
 		var/ikey = "[airlock_type]-[door_color]-color"
 		color_overlay = airlock_icon_cache["[ikey]"]
 		if(!color_overlay)
 			color_overlay = new(color_file)
 			color_overlay.Blend(door_color, ICON_MULTIPLY)
 			airlock_icon_cache["[ikey]"] = color_overlay
-	if(glass)
-		if (window_color && window_color != "none")
+	if(reinf_material)
```
out-of-phaze, line 324: eeeeeeesh, idk how to feel about this, feels like we're losing specificity and also that `reinf_material` is too overloaded. ~~at this point maybe `reinf_material` should be `secondary_material` or maybe even just `second_material`~~

out-of-phaze (dismissed):

code/game/machinery/doors/airlock.dm
MistakeNot4892: It's not the most indicative name, but this is generally the role it fills elsewhere, as a secondary inset or decoration material.

out-of-phaze (approved):

PR #5196, opened 2025-11-01
Fixes inverted ordinal logic.

MistakeNot4892:


code/_helpers/unsorted.dm
```diff
@@ -774,7 +774,7 @@ var/global/list/WALLITEMS = list(
 
 /**Returns a number string with its ordinal suffix th, st, nd, rd */
 /proc/get_ordinal_string(var/num)
-	if(num < 10 && num > 20) //11, 12, 13 are exceptions in english, and just get 'th' like everything else
+	if(num < 10 || num > 20) //11, 12, 13 are exceptions in english, and just get 'th' like everything else
```
out-of-phaze, line 779: will not work for 111th etc

out-of-phaze (changes requested):

code/_helpers/unsorted.dm
```diff
@@ -774,7 +774,8 @@ var/global/list/WALLITEMS = list(
 
 /**Returns a number string with its ordinal suffix th, st, nd, rd */
 /proc/get_ordinal_string(var/num)
-	if(num < 10 && num > 20) //11, 12, 13 are exceptions in english, and just get 'th' like everything else
+	num %= 100
```
out-of-phaze: this means if you input 100 it'll give you 0th

out-of-phaze (changes requested):

out-of-phaze (approved):

PR #5205, opened 2025-11-26
Add new grammar helpers

out-of-phaze:
Adds some grammar helpers that are currently unused. Separated out from a wider implementation of these helpers because I'm not sure what the best way to handle that is.

### DIFF
--- a/code/modules/mob/mob.dm
+++ b/code/modules/mob/mob.dm
@@ -123,6 +123,123 @@
 	if(bound_overlay)
 		bound_overlay.visible_message(message, self_message, blind_message)
 
+/mob/proc/get_action_string(is_self, var/using_verb, var/object_phrase, var/infix, var/postfix)
+	// use mob gender value because otherwise subject verb agreement gets messed up by disguised mobs
+	// todo: pseudoplural pronouns? we need "Unknown (as David Hasselhoff) spins around." and "You spin around." to both work
+	// without breaking "The bats spin around."
+	var/decl/pronouns/using_pronouns = is_self ? get_self_pronouns() : get_pronouns_by_gender(gender)
+	// A little kludgy/special-cased: we don't use the name for self messages.
+	var/actor_string = is_self ? using_pronouns.He : "\The [src]"
+	// this will hopefully handle is/does/has agreement properly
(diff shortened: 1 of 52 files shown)


code/modules/mob/mob.dm
```diff
@@ -123,6 +123,123 @@
 	if(bound_overlay)
 		bound_overlay.visible_message(message, self_message, blind_message)
 
+/mob/proc/get_action_string(is_self, var/using_verb, var/object_phrase, var/infix, var/postfix)
+	// use mob gender value because otherwise subject verb agreement gets messed up by disguised mobs
+	// todo: pseudoplural pronouns? we need "Unknown (as David Hasselhoff) spins around." and "You spin around." to both work
```
MistakeNot4892: I think technically this is what ey/em is for, but I can see that would be weird/jarring for The Common Player if it just started showing up instead of they/them.

code/modules/mob/mob.dm
```diff
@@ -123,6 +123,123 @@
...
+/// Uses SPAN_NOTICE for both self and other messages.
+var/global/const/ACTION_DANGER_NONE = 0
+/// Uses SPAN_DANGER for others and SPAN_WARNING for self.
+var/global/const/ACTION_DANGER_OTHERS = 1
+/// Uses SPAN_DANGER for both self and others.
+var/global/const/ACTION_DANGER_ALL = 2
+/**
+	Show an action message to all mobs and objects in sight of this mob.
+
+	Used for atoms performing visible actions. Handles basic self-messages automatically.
```
MistakeNot4892, line 171: An example usage would really help explicate these functions, I'm struggling to follow them by eye at the moment.

MistakeNot4892: Yeah after looking at the other PR this makes a lot more sense to read.

MistakeNot4892 (dismissed):
I don't fully understand what the adverb handling is actually doing, but this looks pretty solid to me. I like the way it's used in the other PR.

out-of-phaze:
> I don't fully understand what the adverb handling is actually doing, but this looks pretty solid to me. I like the way it's used in the other PR.

the adverb handling is incomplete, oops

anyway, this pr will not work as-is because it will use "Unknown have" for `gender = PLURAL` characters. unsure how to resolve

MistakeNot4892 (dismissed):

MistakeNot4892 (approved):

PR #5209, opened 2025-11-28
Replacing direct access to /datum/reagents vars with macros.

MistakeNot4892:
## Description of changes
Replaces direct access to reagent datum vars with macros.

With serde adding the potential for reagents to be a list, more rigorous typechecking is needed. Doing this with a macro allows for avoiding proc overhead while also enforcing typechecking.

## Authorship
Myself.

## Changelog
Nothing player-facing.

### DIFF
--- a/code/game/objects/items/weapons/ecigs.dm
+++ b/code/game/objects/items/weapons/ecigs.dm
@@ -39,7 +39,7 @@
 /obj/item/clothing/mask/smokable/ecig/simple/get_examine_strings(mob/user, distance, infix, suffix)
 	. = ..()
 	if(ec_cartridge)
-		. += SPAN_NOTICE("There are [round(ec_cartridge.reagents.total_volume, 1)] units of liquid remaining.")
+		. += SPAN_NOTICE("There are [round(REAGENT_TOTAL_VOLUME(ec_cartridge.reagents), 1)] units of liquid remaining.")
 	else
 		. += SPAN_NOTICE("There's no cartridge connected.")
 
@@ -58,7 +58,7 @@
 /obj/item/clothing/mask/smokable/ecig/util/get_examine_strings(mob/user, distance, infix, suffix)
 	. = ..()
 	if(ec_cartridge)
-		. += SPAN_NOTICE("There are [round(ec_cartridge.reagents.total_volume, 1)] units of liquid remaining.")
+		. += SPAN_NOTICE("There are [round(REAGENT_TOTAL_VOLUME(ec_cartridge.reagents), 1)] units of liquid remaining.")
 	else
 		. += SPAN_NOTICE("There's no cartridge connected.")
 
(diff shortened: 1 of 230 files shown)


MistakeNot4892:
This is mostly done but I'd like to check the macro usage to make sure I'm not passing a giant condition chain into any unnecessary spots (like LAZYLEN etc)

code/game/objects/items/weapons/storage/bible.dm
```diff
@@ -89,7 +89,7 @@
 	if(proximity && user?.mind?.assigned_job?.is_holy)
 		if(A.reagents && A.reagents.has_reagent(/decl/material/liquid/water)) //blesses all the water in the holder
 			to_chat(user, SPAN_NOTICE("You bless \the [A].")) // I wish it was this easy in nethack
-			LAZYSET(A.reagents.reagent_data, /decl/material/liquid/water, list(DATA_WATER_HOLINESS = TRUE))
+			REAGENT_SET_DATA(A.reagents, /decl/material/liquid/water, list(DATA_WATER_HOLINESS = TRUE))
```
out-of-phaze, line 92: come to think of it, doesn't this wipe all the other data on the water?

code/game/objects/items/weapons/ecigs.dm
```diff
@@ -102,8 +102,9 @@
 	if(ishuman(loc))
 		var/mob/living/human/user = loc
 
-		if (!lit || !ec_cartridge || !ec_cartridge.reagents.total_volume)//no cartridge
-			if(!ec_cartridge.reagents.total_volume)
+		var/cart_vol = REAGENT_TOTAL_VOLUME(ec_cartridge.reagents)
```
out-of-phaze, line 105: this will runtime if ec_cartridge is null because we try to access reagents on null

code/game/objects/items/weapons/towels.dm
```diff
@@ -93,17 +96,17 @@
 
 /obj/item/towel/proc/dry_mob(mob/living/target, mob/living/user)
 	user.setClickCooldown(DEFAULT_ATTACK_COOLDOWN)
-	var/reagent_space = reagents.maximum_volume - reagents.total_volume
+	var/reagent_space = REAGENT_MAXIMUM_VOLUME(reagents) - REAGENT_TOTAL_VOLUME(reagents)
```
out-of-phaze, line 99: not sure why i didn't use REAGENTS_FREE_SPACE here, but now that would make it faster i think, since it'd go from 2 typechecks to 1

out-of-phaze (changes requested):
frankly i'm not sure i 100% get why this is necessary

code/game/objects/items/weapons/storage/bible.dm
MistakeNot4892: Yes.

MistakeNot4892:
> frankly i'm not sure i 100% get why this is necessary

I keep running into cases on the persistence branch where a proc runs before serde has finished and results in runtime errors. Figured I'd be thorough with protecting against it.

out-of-phaze (approved):
i would like it if we made this unnecessary eventually but that can be something i work on later tbh. thanks for going through on this

also i like the UNLINT to avoid direct access to these 👍

PR #5210, opened 2025-11-28
Move brain-machine interfaces into a modpack

out-of-phaze:
## Description of changes
Moves MMIs and related content into a modpack. Cyborgs and flying cyborgs now require the modpack to be available.
Makes core code more brainmob-holder-agnostic when it comes to things like robot/AI construction.
AIs now take a brainmob rather than a brain organ in Initialize(), which changes basically nothing.
`mmi_types_by_title` is now `processor_types_by_title`.
Adds a `drop_equipped_items()` helper.
`Robotize()` now exists on `/mob/living`.

- [ ] Make sure FBPs are properly restricted?
- [ ] Tested
  - [ ] `/obj/item/projectile/change`
  - [x] FBPs
  - [ ] Constructing a borg with a posibrain
  - [ ] Constructing a borg with an MMI
  - [ ] Constructing an AI with a posibrain
  - [ ] Constructing an AI with an MMI
  - [ ] Admin-robotizing a mob

I don't like these much, and I was really worried about them worming their way into core code as a dependency (like when posibrains/etc. were a subtype of them). This should ensure that can't happen again.

As a bonus: Some code that was broken (no longer accepted posibrains/circuitbrains) when digital MMIs were repathed to `brain/robotic` should now work again.


out-of-phaze:
image
well fbps work at least

MistakeNot4892 (approved):
This hasn't been tested but it's a barely used feature and should be tested during staging (if we get to that).

PR #5213, opened 2025-11-28
Modpacks the sealant gun module

out-of-phaze:
## Description of changes
- Merges the human-level `handle_speech_problems` override into the living-level definition because no behavior in it is humantype-specific.
- Avoids bespoke typechecks in speech blocking checks by using a method instead.
- Removes a hardcoded sealant tank reference in the metal foam reaction with a typecheck, since we use the overflow reaction flag anyway. On a turf it will expand, otherwise it creates the foam reagent.
- You can now make sealant foam with iron instead of aluminum.
- The sealant gun module is now in a modpack.

- [x] Tested

Slowly clearing out `modules`. Also reduces bespoke typechecks and allows for better extensibility.

## Changelog
:cl:
tweak: Iron can now optionally be used instead of aluminum when filling sealant tanks for sealant guns.
/:cl:


MistakeNot4892:
Do foam grenades still work?
The foam reagent is useless at the moment, I honestly prefer the typecheck since it means nowhere else in the game can trip a footgun and end up with a ton of useless foam reagent.

out-of-phaze:
foam grenades do work. funnily enough they don't actually work on current dev

mods/content/sealant_gun/_sealant_gun.dm
```diff
@@ -0,0 +1,3 @@
+/decl/modpack/sealant_gun
```
MistakeNot4892, line 1: Could probably move decompiler nades in here and call it an Art Deco Scifi Thriller Modpack

out-of-phaze: hm, i think only if we get `skilljacks` that give you a skill boost and hostile aliens called `duplicates` that can turn into arbitrary objects....

MistakeNot4892 (approved):

PR #5215, opened 2025-12-20
Allow quick-equip hotkey to place items in backpacks and wallets

noelle-lavenza:
## Description of changes
Adds a new 'in wallet' abstract slot which is used to equip things to storage items in the ID slot, similar to the 'in backpack' abstract slot.

Quick-equip will now attempt to place items in backpacks/wallets if no other options work.

Passports will now try to equip inside wallets if possible.

Some equipping QoL functionality. It might be worth making it so that, when triggered by the quick-equip hotkey, it opens the storage menu as well so it doesn't just seem like it disappears? It might also be worth adding 'currently open storage' but I'm unsure how to handle that...

## Authorship
Me, but the concept exists on TG.

## Changelog
:cl:
tweak: The equip hotkey can now place items into backpacks and wallets.
/:cl:

### DIFF
--- a/code/game/objects/items/__item.dm
+++ b/code/game/objects/items/__item.dm
@@ -726,6 +726,9 @@
 	if(slot == slot_in_backpack_str)
 		var/obj/item/back = user.get_equipped_item(slot_back_str)
 		return back?.storage?.can_be_inserted(src, user, TRUE)
+	if(slot == slot_in_wallet_str)
+		var/obj/item/wallet = user.get_equipped_item(slot_wear_id_str)
+		return wallet?.storage?.can_be_inserted(src, user, TRUE)
 
 	var/datum/inventory_slot/inv_slot = user.get_inventory_slot_datum(slot)
 	if(!inv_slot)
--- a/code/modules/mob/living/inventory.dm
+++ b/code/modules/mob/living/inventory.dm
@@ -14,10 +14,16 @@
 		var/list/all_slots = list()
 		for(var/slot in get_inventory_slots())
 			all_slots += get_inventory_slot_datum(slot)
+		var/list/low_priority_slots // Slots that will always be added at the end, in the order of their parent slots' priority.
+		// This is sort of due to technical limitations but mostly due to laziness.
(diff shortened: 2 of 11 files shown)


code/modules/mob/living/inventory.dm
```diff
@@ -14,10 +14,16 @@
...
+		var/list/low_priority_slots // Slots that will always be added at the end, in the order of their parent slots' priority.
+		// This is sort of due to technical limitations but mostly due to laziness.
 		for(var/datum/inventory_slot/inv_slot as anything in sortTim(all_slots, /proc/cmp_inventory_slot_desc))
+			if(LAZYLEN(inv_slot.additional_quick_equip_slots))
+				for(var/extra_slot in inv_slot.additional_quick_equip_slots)
+					LAZYADD(low_priority_slots, extra_slot)
 			if(isnull(inv_slot.quick_equip_priority)) // Never quick-equip into some slots.
 				continue
 			_inventory_slot_priority += inv_slot.slot_id
+		_inventory_slot_priority += low_priority_slots
```
out-of-phaze: won't this result in a null entry if nothing's been added?

```suggestion
		if(LAZYLEN(low_priority_slots))
			_inventory_slot_priority += low_priority_slots
```
alternatively, make `low_priority_slots` a non-lazylist, since `list() += list()` is just `list()` (it tries to add zero items to the list)

out-of-phaze (dismissed):

code/game/objects/items/__item.dm
```diff
@@ -726,6 +726,9 @@
 	if(slot == slot_in_backpack_str)
 		var/obj/item/back = user.get_equipped_item(slot_back_str)
 		return back?.storage?.can_be_inserted(src, user, TRUE)
+	if(slot == slot_in_wallet_str)
```
MistakeNot4892, line 729: I feel like we might want to consider an abstract inventory slot type for handling this without blocks in this proc in the future. Might not be needed but seeing this expand is a bit hmmm.

MistakeNot4892 (approved):

PR #5216, opened 2025-12-26
Porting quadbikes from Polaris.

MistakeNot4892:
DNM, opening for diff. Currently this is just copypasted code and the diff will make it easier to review/refine.

- Repaths cargo trains to a general train and trolley type.
- Ports various tweaks and supporting changes from Polaris vehicle code and splits vehicle code into separate files for ease of review.
- Ports quadbikes and trailers.


MistakeNot4892:
Tested, appears to be working.

PlayerDeer:
Horses when?

code/__defines/misc.dm
```diff
@@ -402,3 +402,7 @@
 #define MM_ATTACK_RESULT_NONE      0
 #define MM_ATTACK_RESULT_DEFLECTED BITFLAG(0)
 #define MM_ATTACK_RESULT_BLOCKED   BITFLAG(1)
+
+#define VEHICLE_GENERIC    1
+#define VEHICLE_QUADBIKE   2
+#define VEHICLE_SNOWMOBILE 3
```
out-of-phaze: i would suggest making these consts on `/obj/vehicle`. we can still access them off that type using `/obj/vehicle::VEHICLE_TYPE_GENERIC` but it makes organization a lot nicer imo. you can even use a vehicle reference, so `vehicle::VEHICLE_TYPE_GENERIC`. look at light fixture/lightbulb code with `STATUS_OK` and etc. for an example of how this is done

code/game/turfs/flooring/flooring_snow.dm
```diff
@@ -54,8 +57,14 @@
...
+	return vehicle.vehicle_transit_type == VEHICLE_SNOWMOBILE ? 0.8 : ..()
+
 /decl/flooring/snow/fake
 	name            = "holosnow"
 	desc            = "Not quite the same as snow on an entertainment terminal, but close."
 	holographic     = TRUE
 	uid             = "floor_snow_fake"
+
+/decl/flooring/snow/fake/get_vehicle_transit_delay(obj/vehicle/vehicle)
+	return initial(vehicle.speed_mod)
```
out-of-phaze: in theory the parent override could just check `holographic`, unlike with fake grass which is just a totally different material. this is probably fine as-is though

code/game/turfs/flooring/_flooring.dm
```diff
@@ -391,3 +391,6 @@ var/global/list/flooring_cache = list()
 /// contaminant is, optionally, the material of the coating that wants to be added.
 /decl/flooring/proc/can_show_coating_footprints(turf/target, decl/material/contaminant)
 	return TRUE
+
+/decl/flooring/proc/get_vehicle_transit_delay(obj/vehicle/vehicle)
+	return initial(vehicle.speed_mod)
```
out-of-phaze: undecided on if this should be `vehicle::speed_mod` or not. i also kind of would prefer a separate `base_speed` variable instead of using initial, because relying on initial value tends to make future-proofing a little painful. not a required change though

out-of-phaze (dismissed):

MistakeNot4892:
> Horses when?

We have ridable horses already!

PlayerDeer:
> > Horses when?
> 
> We have ridable horses already!

Wagons when? Carts when? 🥺🥺🥺

code/game/turfs/flooring/_flooring.dm
MistakeNot4892: Ended up on redoing how base speed was calculated to clean it up a bit. Included a `base_speed` value now.

out-of-phaze (approved):

PR #5217, opened 2025-12-28
Removing assoc list syntax using numerical keys.

MistakeNot4892:
New BYOND does not permit numerical keys to be used like assoc keys in basic list().


deathride58 (approved):
this makes us wish byond had proper structs

PR #5220, opened 2025-12-28
Moves augments into their own modpack

out-of-phaze:
## Description of changes
- Separates `/obj/item/helping_hands` out from the augments modpack because they're only barely conceptually related and share no code at all.
- Removes unused organ tag defines.
- Moves modpack-only organ tag defines into their associated modpacks.
- Moves augments into their own modpack.

- [x] Tested

More modpacking. The current augment implementation is a little... opinionated? So this allows downstreams to replace it with their own entirely.

## Authorship
Me.


out-of-phaze:
frankly i hate BP_AUGMENT_L_ARM and similar so much that i think straight up adding a `var/obj/item/organ/internal/augment` var to external organs might be preferable. in fact then we might be able to make it so that augments aren't organs at all

out-of-phaze:
tested by giving myself an armor augment and making sure that worked. that was the most invasive change i made so everything else should still be fine

MistakeNot4892 (approved):

PR #5223, opened 2025-12-28
Remove provocateur/semi-antagonist role

out-of-phaze:
## Description of changes
Removes the provocateur special_role and the is_semi_antagonist job variable. Changes kharmaan nymphs to have their own roleban instead of using the provocateur one. Makes it marginally easier for modpacks to define new misc. jobban roles.

Reasoning given on Discord:
> the `is_semi_antagonist` var on submap jobs is entirely unused (always false), provocateur is similarly unused anywhere other than that, and as far as i can recall it originates entirely from the Baystation rules for non-antagonists, which doesn't really apply here
> they're a Bay thing derived from a specific approach to adminning and roleplay that doesn't make sense here. because they're so specific to one particular server, if a server needs something similar the onus to reimplement it should be on them
> also on a personal level, i don't like having cruft sitting around getting crusty in my codebase. removing it makes things easier on us, the maintainers.

## Authorship
Me

## Changelog
:cl:
admin: Kharmaan nymph jobbans now use the "Mantid Nymph" role in the jobban menu, rather than the Semi-Antagonist role.
/:cl:


MistakeNot4892 (dismissed):
It kinda sucked on Bay in practice anyway.

MistakeNot4892 (approved):
Conflict resolution

PR #5224, opened 2025-12-28
Use LAZYLEN macro where applicable

out-of-phaze:
## Description of changes
Replaces `(!foo || !foo.len)` and `(foo && foo.len)` with `!LAZYLEN(foo)` and `LAZYLEN(foo)` respectively. I tried to avoid using it in cases where it was on a string instead of a list, and in those cases just used `length(foo)` on its own because the nullcheck is redundant.

Should be marginally faster, and also easier to read. This is broadly a code-quality cleanup.

## Authorship
Me

### DIFF
--- a/code/modules/power/powernet.dm
+++ b/code/modules/power/powernet.dm
@@ -101,7 +101,7 @@
 	if(problem > 0)
 		problem = max(problem - 1, 0)
 
-	if(nodes && nodes.len) // Added to fix a bad list bug -- TLE
+	if(LAZYLEN(nodes)) // Added to fix a bad list bug -- TLE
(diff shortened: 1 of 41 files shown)


code/modules/power/powernet.dm
```diff
@@ -101,7 +101,7 @@
 	if(problem > 0)
 		problem = max(problem - 1, 0)
 
-	if(nodes && nodes.len) // Added to fix a bad list bug -- TLE
+	if(LAZYLEN(nodes)) // Added to fix a bad list bug -- TLE
```
MistakeNot4892, line 104: Comment is a bit pointless to keep

MistakeNot4892 (approved):

PR #5120, opened 2025-09-10
Reorganize the power module

out-of-phaze:
Splits up the power module. WIP.

PR #5229, opened 2025-12-28
Make breaker boxes no longer a power subtype

out-of-phaze:
## Description of changes
Split out of #5120.
Breaker boxes are no longer a `/obj/machinery/power` subtype because, well, they don't need to be.

- [x] Tested
- [x] Map migration added

Unnecessary subtyping is unnecessary.

## Authorship
Me


out-of-phaze:
image

works now, had a really bad bug that took literally 2 seconds to test, why did i avoid doing it for almost 7 months

MistakeNot4892 (approved):

PR #5230, opened 2025-12-28
Make power debug machines not use the power subtype

out-of-phaze:
## Description of changes
Split out of #5120.
Power debug objects are no longer a `/obj/machinery/power` subtype because, well, they don't need to be.

- [x] Map migration added

Unnecessary subtyping is unnecessary.

## Authorship
Me

### DIFF
--- a/tools/map_migrations/5320_power_debug.txt
+++ b/tools/map_migrations/5320_power_debug.txt
@@ -1,0 +1,2 @@
+/obj/machinery/power/debug_items/@SUBTYPES : /obj/machinery/debug_items/@SUBTYPES{@OLD}
+/obj/machinery/power/ascent_reactor/@SUBTYPES : /obj/machinery/ascent_reactor/@SUBTYPES{@OLD}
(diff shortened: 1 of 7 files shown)


tools/map_migrations/5320_power_debug.txt
```diff
@@ -0,0 +1,2 @@
+/obj/machinery/power/debug_items/@SUBTYPES : /obj/machinery/debug_items/@SUBTYPES{@OLD}
+/obj/machinery/power/ascent_reactor/@SUBTYPES : /obj/machinery/power/@SUBTYPES{@OLD}
```
MistakeNot4892: This doesn't seem correct.

MistakeNot4892 (changes requested):
Migration is using the wrong replacement path.

tools/map_migrations/5320_power_debug.txt
out-of-phaze: oops

MistakeNot4892 (approved):

PR #5232, opened 2025-12-29
Moves emergency response team into a modpack

out-of-phaze:
## Description of changes
- Moves ERT content into a modpack.

The modpack is only enabled on Exodus currently, since that's a NanoTrasen map.

- [x] Tested
  - [x] keycard auth
  - [x] admin verb
  - [x] disable ert via gamemode options verb

After the other PRs are merged, this will bring us down to 0 special roles in core code. :)
Also, it's lore stuff tied pretty strongly into megacorp/NanoTrasen stuff, and it was never really as extensible as people made it out to be, it only ever supported the one ERT type after all.

## Authorship
Me


out-of-phaze:
dependency on #5233

out-of-phaze:
tested, works, also fixed a long-standing runtime.

MistakeNot4892 (approved):

PR #5233, opened 2025-12-29
Make the keycard auth device use decls for event options

out-of-phaze:
## Description of changes
They aren't hardcoded anymore. Decls. Woo.
Also moves `maint_all_access` onto `/datum/map` in preparation for multi-map stuff, maybe...? Iunno.
todo: add priority so we can control ordering

- [x] Tested

allows modpacks to add their own entries

## Authorship
Me

### DIFF
--- a/code/game/jobs/access.dm
+++ b/code/game/jobs/access.dm
@@ -66,7 +66,7 @@
 	if(!istype(supplied_access, /list))
 		return FALSE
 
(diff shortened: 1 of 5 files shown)


out-of-phaze:
tested, should be working now.

MistakeNot4892 (approved):

PR #5238, opened 2025-12-31
Several fixes/tweaks from Pyrelight test.

MistakeNot4892:
species_allowed now uses UID, so this just breaks markings for drakes on fantasymap.

### DIFF
--- a/code/modules/materials/definitions/solids/materials_solid_stone.dm
+++ b/code/modules/materials/definitions/solids/materials_solid_stone.dm
@@ -30,6 +30,7 @@
 /decl/material/solid/stone/Initialize()
 	. = ..()
 	texture = image('icons/turf/wall_texture.dmi', "concrete")
+	texture.appearance_flags |= RESET_COLOR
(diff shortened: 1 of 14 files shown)


code/modules/materials/definitions/solids/materials_solid_stone.dm
```diff
@@ -30,6 +30,7 @@
 /decl/material/solid/stone/Initialize()
 	. = ..()
 	texture = image('icons/turf/wall_texture.dmi', "concrete")
+	texture.appearance_flags |= RESET_COLOR
```
out-of-phaze: in theory since it's just multiplicative, RESET_ALPHA would also be good to avoid it becoming imperceptible on stuff like glass?

out-of-phaze (dismissed):
not blocking

out-of-phaze (dismissed):
planet generators do set paint_color for rock weathering but those are really weird and underbaked so frankly it's fine if ramps don't fully respect it and just use material color

out-of-phaze (approved):

PR #5239, opened 2025-12-31
Changing /decl/material/gas out for a filterable material proc.

### LINKED ISSUE #5100: Scrubbers can't filter gases that aren't subtypes of /decl/material/gas
#### Description of issue
Scrubbers cannot selectively remove certain gases if those gases are not a subtype of `/decl/material/gas`, but are still in a gaseous state. Random exoplanet atmos is really likely to produce those kinds of gases.

I expected scrubbers to be able to remove all unwanted gases from their local atmosphere. They can scrub gases like hydrogen and chlorine, but they can't remove things like vaporized sulfuric acid.

#### Steps to reproduce
(I'm not sure if Neb has a verb to arbitrarily add gas of any material somewhere. If there is, that would probably be easier than this.)
1. Find a way to bring a functional scrubber pipe system to an exoplanet with weird gases (weird being defined as non-`/decl/material/gas`). This could be by moving a ship with a scrubber system over, or building one on the planet and powering it somehow.
2. Introduce the weird gas to the scrubbers and then make the room airtight.
3. If any regular `/decl/material/gas`es beside oxygen and nitrogen were there, they should slowly be getting pulled out of the air and into whatever pipes the scrubbers are attached to.
4. If any non-`/decl/material/gas` gases are in the air, observe how they are not being removed selectively.

https://github.com/NebulaSS13/Nebula/blob/dev/code/modules/atmospherics/components/unary/vent_scrubber.dm#L67
Changing it to filter all `/decl/material` would probably fix it but the filter list would probably become massive and waste valuable memory, so it might be better to change the filtering logic to exclude gases contained instead of filtering only the gases contained, but that probably requires changes in `scrub_gas()` as well to operate under that same logic.

I realized this bug while testing a new ship for Spacefarers (a downstream) but I assume this bug has existed for a long time.

(Local server for Spacefarers but this bug also exists on stock Neb on Dev)
Client Version: 514
Server Revision: 6e7978d06ef458aaa2e472be979688154e1bf3c2 - spacefarers_finishing_salvage_ship - 2025-07-25
Game ID: cAN-avmW
Current map: the Frontier

#### Issue bingo

MistakeNot4892:
Fixes #5100

### DIFF
--- a/code/_helpers/type2type.dm
+++ b/code/_helpers/type2type.dm
@@ -1,12 +1,19 @@
 /*
  * Holds procs designed to change one type of value, into another.
  * Contains:
- *			text2list & list2text
+ *          alist2list
  *			file2list
  *			angle2dir
  *			angle2text
- *			worldtime2text
  */
+
+// This proc does not support converting numerically indexed alists to assoc lists.
+/proc/alist2list(alist/input)
+	. = list()
+	for(var/k,v in input)
+		if(isnum(k))
+			CRASH("Numeric index passed to alist2list()!")
+		.[k] = v
 
 // Splits the text of a file at seperator and returns them in a list.
 /proc/file2list(filename, seperator = "\n")
--- a/code/modules/maps/template_types/random_exoplanet/fauna_generator.dm
+++ b/code/modules/maps/template_types/random_exoplanet/fauna_generator.dm
@@ -126,7 +126,7 @@
 
 /datum/fauna_generator/proc/generate_breathable_gases(var/datum/gas_mixture/atmosphere, var/list/breath_gases, var/list/toxic_gases)
 	//Set up gases for living things
-	var/list/all_gasses = decls_repository.get_decl_paths_of_subtype(/decl/material/gas)
+	var/list/all_gasses = get_filterable_material_types(as_list = TRUE)
 	if(!length(breath_gases))
 		var/list/goodgases = all_gasses.Copy()
 		var/gasnum = min(rand(1,3), goodgases.len)
--- a/code/game/machinery/alarm.dm
+++ b/code/game/machinery/alarm.dm
@@ -159,7 +159,7 @@
 	TLV["temperature"] =	list(T0C-26, T0C, T0C+40, T0C+66) // K
 
 	var/decl/environment_data/env_info = GET_DECL(environment_type)
-	for(var/g in decls_repository.get_decl_paths_of_subtype(/decl/material/gas))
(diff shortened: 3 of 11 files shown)


MistakeNot4892:
Seeeems to be working fine?

code/_helpers/type2type.dm
```diff
@@ -1,13 +1,17 @@
...
  *			file2list
  *			angle2dir
  *			angle2text
- *			worldtime2text
  */
 
+/proc/alist2list(alist/input)
+	. = list()
+	for(var/k,v in input)
+		.[k] = v
```
out-of-phaze, line 16: `alist2list(alist(1 = "a"))` will break here, you need to either enforce stringification of numerical indices or skip numerical indices or etc

code/modules/maps/template_types/random_exoplanet/fauna_generator.dm
```diff
@@ -126,7 +126,7 @@
 
 /datum/fauna_generator/proc/generate_breathable_gases(var/datum/gas_mixture/atmosphere, var/list/breath_gases, var/list/toxic_gases)
 	//Set up gases for living things
-	var/list/all_gasses = decls_repository.get_decl_paths_of_subtype(/decl/material/gas)
+	var/list/all_gasses = get_filterable_material_types(as_list = TRUE)
```
out-of-phaze, line 129: curious why the rest of this couldn't use alist to prevent needing the conversion? toxic_gases is already explicitly constructed as a list, as is `breath_gases`, you'd just have to retype badgases and goodgases to `alist` i think

out-of-phaze (changes requested):

code/modules/maps/template_types/random_exoplanet/fauna_generator.dm
MistakeNot4892: Can't pick() from alists. :(

code/_helpers/type2type.dm
MistakeNot4892: Added a comment saying it's unsupported, and had it crash if encountered.

code/modules/maps/template_types/random_exoplanet/fauna_generator.dm
out-of-phaze: some PR i have open has a `pick_alist()` method, i think, but i don't remember which. oh well

out-of-phaze (approved):
what if we pretended i hit approve when i left that last comment 😳

PR #5246, opened 2026-01-01
Porting various projectile impact effects and sounds from Polaris.

MistakeNot4892:
## Description of changes
- Projectiles have sound effects that play on when striking an atom, one for mobs, one for non-mobs.
- Projectiles will spawn a visual effect on impact.
- Projectiles now use static lists for their various SFX instead of duplicate lists. 

## Authorship
Sounds. icons and general concept ported from Polaris, implementation was fairly bespoke but more or less just a port.

## Changelog
:cl:
add: Projectiles have new visual effects and SFX for impacts.
/:cl:


code/game/sound.dm
```diff
@@ -141,29 +141,29 @@ var/global/const/FALLOFF_SOUNDS = 0.5
...
+			if ("light_bic")    soundin = pick(global.lighter_sound)
+			if ("keyboard")     soundin = pick(global.keyboard_sound)
+			if ("keystroke")    soundin = pick(global.keystroke_sound)
+			if ("switch")       soundin = pick(global.switch_sound)
+			if ("button")       soundin = pick(global.button_sound)
+			if ("chop")         soundin = pick(global.chop_sound)
+			if ("glasscrack")   soundin = pick(global.glasscrack_sound)
+			if ("tray_hit")     soundin = pick(global.tray_hit_sound)
+			if ("sweeping")     soundin = pick(global.sweeping_sound)
+			if("ricochet")      soundin = pick(global.ricochet_sound)
```
out-of-phaze, line 164:
```suggestion
			if ("ricochet")      soundin = pick(global.ricochet_sound)
```
my consistency...

code/modules/mob/living/living_defense.dm
```diff
@@ -42,7 +42,8 @@
 
 // For visuals and blood splatters etc
 /mob/living/proc/bullet_impact_visuals(var/obj/item/projectile/P, var/def_zone, var/damage)
-	var/list/impact_sounds = LAZYACCESS(P.impact_sounds, get_bullet_impact_effect_type(def_zone))
+	var/list/all_impact_sounds = P.get_impact_sounds()
+	var/list/impact_sounds = LAZYACCESS(all_impact_sounds, get_bullet_impact_effect_type(def_zone))
```
out-of-phaze, line 46: i am not a big fan of the visuals/sounds being linked like this but it seems to be mostly a proc naming thing rather than conceptual linking, so i think this is fine

out-of-phaze (approved):

code/modules/mob/living/living_defense.dm
MistakeNot4892: Yeah the proc should be 'bullet impact effect' or smt

PR #5248, opened 2026-01-01
Make gases use alist

out-of-phaze:
## Description of changes
gases use alist now. also includes some attempts at simplifying the math. this is probably extremely buggy

in theory this will be a good bit faster.

## Authorship
me


out-of-phaze:
cc @noelle-lavenza, i think some issues with my math were found on nuclear phase that need to be backported?

code/modules/xgm/xgm_gas_mixture.dm
```diff
@@ -493,10 +493,7 @@
 	return 1
 
 /datum/gas_mixture/proc/get_mass()
-	. = 0
-	for(var/g in gas)
-		var/decl/material/mat = GET_DECL(g)
-		. += gas[g] * mat.molar_mass * group_multiplier
+	return values_dot(gas, global.cached_molar_mass) * group_multiplier
```
MistakeNot4892, line 496: This could be inlined as a macro, looks like?

code/modules/power/singularity/collector.dm
```diff
@@ -85,7 +85,8 @@ var/global/list/rad_collectors = list()
 		toggle_power()
 		user.visible_message("[user.name] turns \the [src] [active? "on":"off"].", \
 		"You turn \the [src] [active? "on":"off"].")
-		investigate_log("turned [active?"<font color='green'>on</font>":"<font color='red'>off</font>"] by [user.key]. [loaded_tank?"Fuel: [round(loaded_tank.air_contents.gas[/decl/material/gas/hydrogen]/0.29)]%":"<font color='red'>It is empty</font>"].","singulo")
+		// WHY DOES THIS DIVIDE BY 0.29
```
MistakeNot4892, line 88: Old fuel consumption constant maybe? No clue

code/modules/materials/_materials.dm
```diff
@@ -373,6 +373,9 @@ var/global/list/materials_by_gas_symbol = list()
 		else if(isnull(ignition_point) || (new_temperature_damage_threshold > ignition_point))
 			temperature_damage_threshold = new_temperature_damage_threshold
 
+	global.cached_specific_heat[type] = gas_specific_heat
```
MistakeNot4892, line 394: Is this to avoid decl lookup overhead?

code/modules/maps/template_types/random_exoplanet/flora_generator.dm
```diff
@@ -71,9 +71,9 @@
 		S.exude_gasses -= exuded_gases_exclusions
 	if(length(atmos.gas))
 		if(S.consume_gasses)
-			S.consume_gasses = list(pick(atmos.gas)) // ensure that if the plant consumes a gas, the atmosphere will have it
-		for(var/g in atmos.gas)
-			var/decl/material/mat = GET_DECL(g)
+			S.consume_gasses = list(apick(atmos.gas)) // ensure that if the plant consumes a gas, the atmosphere will have it
+		for(var/gas_type in atmos.gas)
+			var/decl/material/mat = GET_DECL(gas_type)
```
MistakeNot4892, line 76: We need to reproduce the work done with `matter` to make gasses use decls as keys instead of types at some point.

out-of-phaze: i think with how often gas mixture datums are created and destroyed, and new keys are added, that would be an unacceptable amount of overhead, but maybe it's fine

MistakeNot4892: Destroyed shouldn't matter, should it?

Created is potentially an issue but given it's usually via splitting or transferring, there should be no additional overhead - the keys would already be decls. Adding gas to a mix would need resolution though, that is true (though we could just ensure adding is passed a decl instead of a type for cases like exhaling gas).

Not relevant to this PR in any case.

code/modules/materials/_materials.dm
out-of-phaze: that, and it lets us use it in the new `values_` procs, so we don't have to explicitly loop over stuff and can get a speed boost. we checked with lummox and it's as expensive as the shortest list, so using a big lookup table is fine

MistakeNot4892 (approved):

PR #5250, opened 2026-01-02
Fix some lighting correctness issues

out-of-phaze:
## Description of changes
Adds an `INIT_CORNER` macro which is `APPLY_CORNER` but very slightly faster.
Fixes lighting corners not being lit from space.
Adjusts actual_range in `recalc_corner()` to be the same as in `update_corners()`.
Some misc. optimizations.

Correctness and optimization


MistakeNot4892 (dismissed):

out-of-phaze:
would like lohikar to approve this too tbh

Lohikar (approved):

PR #5253, opened 2026-01-02
Renormalized .sh, .dme and .py files.

MistakeNot4892:
Not relevant to Neb main currently but noticed while rebasing PolNeb.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -12,6 +12,11 @@ html/changelog.html merge=union
 # Declare files that will always have CRLF line endings on checkout.
 *.dm text eol=crlf
 *.dmm text eol=crlf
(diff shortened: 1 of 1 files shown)


out-of-phaze:
did you actually renormalize the files as well? only .gitattributes is changed, and i know polneb changed cult.dme's line endings after adding the dme line, so

MistakeNot4892:
> did you actually renormalize the files as well? only .gitattributes is changed, and i know polneb changed cult.dme's line endings after adding the dme line, so

Yes. When I run it on polneb there's a bunch of DMEs that get renormalized but those changes are not on Neb main, so not reflected here.

out-of-phaze (approved):

PR #5255, opened 2026-01-03
Fixes for partial corner lists during shuttle transit.

MistakeNot4892:
Fixes on the advice of @Lohikar, space turfs can have incomplete or null corner lists and hence will break when shuttles traverse them.


out-of-phaze (approved):

Lohikar:
As mentioned on discord, the code I gave you was a temporary fix that needed revisiting -- on hold for a proper fix.

MistakeNot4892:
Will reopen this once I can repro, Cynosure branch is no longer working enough to show it.

PR #5219, opened 2025-12-28
Porting cliffs and mines from Polaris.

MistakeNot4892:
Porting a mine rewrite I did for Polaris, and cliff structures also from Polaris.

- Rewrites and cleans up `/obj/item/mine`.
- Adds a `silent` arg to can_climb() proc.
- Adds code to prevent walking into hazards when moving deliberately.
- Ports cliffs that will cause movers to fall down and hurt themselves if walking over them.

## TODO
- [X] Tested mines.
- [x] Tested cliffs.
- [x] Fix visual issues with cliffs on turfs.

PR #5262, opened 2026-01-08
Porting main code changes from Polaris rebase.

MistakeNot4892:
Keeping main code and Polaris code aligned as much as possible will make future merges less painful. I can split the commit up potentially but a list of changes is below.

### DIFF
--- a/code/datums/storage/subtypes_box.dm
+++ b/code/datums/storage/subtypes_box.dm
@@ -18,6 +18,9 @@
 /datum/storage/box/animal_cube
 	can_hold = list(/obj/item/food/animal_cube)
 
+/datum/storage/box/large/metal
+	use_sound = 'sound/effects/closet_open.ogg'
+
 /datum/storage/box/snappop
 	can_hold = list(/obj/item/toy/snappop)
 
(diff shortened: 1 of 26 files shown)


MistakeNot4892:
Pulled in a ton of new commits from the Cynosure branch, they should be fairly self-explanatory.

MistakeNot4892:
Depends on #5219

MistakeNot4892:
Rebased to remove all the removed commits/logic.

code/modules/emotes/emote_define.dm
```diff
@@ -18,7 +18,7 @@
 		. = replacetext(., "$USER_THEM$",  user_gender.him)
 		. = replacetext(., "$USER_THEIR$", user_gender.his)
 		. = replacetext(., "$USER_SELF$",  user_gender.self)
-		. = replacetext(., "$USER$",       "<b>\the [user]</b>")
+		. = replacetext(., "$USER$",       "<b>[user]</b>")
```
out-of-phaze: this change was made for a reason. capitalize_proper_html() is used to fix the capitalization of the start of the emote string. where are you using the replacements that that isn't used?

out-of-phaze (changes requested):

code/modules/emotes/emote_define.dm
MistakeNot4892: Cryopods, looks like. Dropping this commit and fixing that.

out-of-phaze (approved):

PR #5265, opened 2026-01-08
Ported/reimplemented salvage structures and items from Polaris.

MistakeNot4892:
## Description of changes
Adds structures and items that can be repaired or dismantled for salvaged materials/parts.
 
Good for mapping, POIs, novelty.

## Authorship
The idea and the icons of the structures are from Polaris but I wrote the code from scratch.

## Changelog
Nothing player-facing.


code/modules/salvage/salvage.dm
```diff
@@ -0,0 +1,128 @@
...
+	. = ..()
+
+/obj/item/salvage/get_examine_hints(mob/user, distance, infix, suffix)
+	. = ..()
+	. += SPAN_NOTICE("It requires the following items to be fully repaired:")
+	for(var/datum/salvage_repair_requirement/rep in repairs_required)
+		if(rep.repair_amount <= 0)
+			continue
+		var/obj/item/repair_type = rep.repair_type
+		var/repair_thing = repair_type::name
```
out-of-phaze: atom info repository may be preferable here

code/modules/salvage/salvage.dm
```diff
@@ -0,0 +1,128 @@
...
+		if(ispath(rep.repair_type, /obj/item/stack))
+			var/obj/item/stack/stack = rep.repair_type
+			repair_thing = rep.repair_amount == 1 ? stack::singular_name : stack::plural_name
+			if(ispath(rep.repair_type, /obj/item/stack/material) && rep.repair_material)
+				var/decl/material/repair_mat = GET_DECL(rep.repair_material)
+				repair_thing = "[repair_mat.solid_name] [repair_thing]"
+		else if(rep.repair_amount > 1)
+			repair_thing = "[repair_thing]s" // is there a better way to add plural suffix?
```
out-of-phaze:
```suggestion
			repair_thing = text_make_plural(repair_thing)
```

code/modules/salvage/salvage.dm
```diff
@@ -0,0 +1,128 @@
...
+
+	for(var/datum/salvage_repair_requirement/rep in repairs_required)
+		if(rep.repair_amount > 0)
+			if(user)
+				to_chat(user, SPAN_NOTICE("You mend some of the damage to \the [src], but further repair is required."))
+			return
+
+	var/obj/item/created = new salvaged_type(get_turf(src))
+	if(isitem(created))
+		created.name_prefix = pick("salvaged", "restored", "old", "worn")
```
out-of-phaze: in theory there might already have been a name prefix? so a pipe wrench would be "salvaged pipe wrench" instead of "enormous pipe wrench." which might be okay, but you could also prefix it to an existing prefix if it exists, or something. not mandatory though

code/modules/salvage/salvage.dm
```diff
@@ -0,0 +1,128 @@
...
+			repair_thing = rep.repair_amount == 1 ? stack::singular_name : stack::plural_name
+			if(ispath(rep.repair_type, /obj/item/stack/material) && rep.repair_material)
+				var/decl/material/repair_mat = GET_DECL(rep.repair_material)
+				repair_thing = "[repair_mat.solid_name] [repair_thing]"
+		else if(rep.repair_amount > 1)
+			repair_thing = "[repair_thing]s" // is there a better way to add plural suffix?
+		. += SPAN_NOTICE("- [rep.repair_amount] [repair_thing]")
+
+/obj/item/salvage/proc/check_repair_completion(mob/user)
+
```
out-of-phaze, line 88: you use a lot of blank lines in spots like this and i'm curious if that's polaris copypasta or personal preference

code/modules/salvage/salvage.dm
```diff
@@ -0,0 +1,128 @@
...
+	SHOULD_CALL_PARENT(FALSE)
+	if(!salvaged_type)
+		return
+	var/old_name = name
+	var/old_desc = desc
+	var/old_pixel_x = pixel_x
+	var/old_pixel_y = pixel_y
+	var/old_plane = plane
+	var/old_layer = layer
+	var/obj/item/reference = atom_info_repository.get_instance_of(salvaged_type)
```
out-of-phaze: this is invalid. in fact, `get_instance_of()` should be a private proc. it is not automatically cleaned up, this might just leak the memory. either we should include a cached mutable appearance on the atom info repository, which might not be valid given some stuff has random variation (wrench handles, etc), or this should use a different method, like creating it without using the atom info repository or using the `appearance = /obj/item/type` format (which uses the initial vars).

code/modules/salvage/salvage_repair_option.dm
```diff
@@ -0,0 +1,60 @@
+/decl/salvage_repair_option
+	abstract_type = /decl/salvage_repair_option
+	var/selection_prob = 0
+	var/list/selection_types
+	var/list/selection_materials
+	var/selection_min_amount = 1
+	var/selection_max_amount = 3
+
+/decl/salvage_repair_option/proc/create_salvage_requirement()
+	var/use_type = pick(selection_types)
```
out-of-phaze, line 24: this will runtime if `selection_types` is null, we should have `/decl/salvage_repair_option/validate()` check to make sure all non abstract repair options have selection types

code/modules/salvage/salvage_repair_option.dm
```diff
@@ -0,0 +1,60 @@
...
+	var/use_mat  = ispath(use_type, /obj/item/stack/material) ? pick(selection_materials) : null
```
out-of-phaze, line 25: validation should enforce `if(ispath(use_type, /obj/item/stack_material)) ASSERT(length(selection_materials))` to avoid a runtime

out-of-phaze (dismissed):
i like this a lot, would be nice to go through and replace stuff on away sites with it. could even use a landmark that has a chance to replace a machine with a corresponding salvage structure, for more variety (or that could just be the proc caller landmark if you added a 'convert to salvage structure' proc). it seems like a great way to improve salvaging gameplay and make recycling actually feasible.

code/modules/salvage/salvage.dm
MistakeNot4892: Why? I just need 'pistol' or such, I don't need any prefixes or material names or anything.

code/modules/salvage/salvage.dm
MistakeNot4892: Personal preference, I like to break stuff up into blocks a bit so I can skim read the file. I can remove it if it's a bother.

code/modules/salvage/salvage.dm
MistakeNot4892: Added a `get_appearance_of()` proc to the atom info repo, only caches appearance for that specific proc to avoid bloating the cache.

code/modules/salvage/salvage.dm
MistakeNot4892: Went ahead and swapped it out regardless.

out-of-phaze (approved):

PR #5270, opened 2026-01-09
Enlarges the stamina bar and moves it centrally.

MistakeNot4892:
Rationale for this change is that I would like to make stamina more relevant and the current tiny icon and position over the right-hand HUD make it unclear what it means, and make it hard to keep an eye on during action.

DNM until the democracy channel has a bit to vote.

image


MistakeNot4892:
Vote for moving it centrally over the hands (Y for yes, N for no):
image

Second vote for putting it high up on the screen instead (Y for Y, N for keep it on move intent selector, B for keep it above hands): 
image

Going to go ahead and keep it over the hands in this PR since the votes are fairly strongly in favour.

out-of-phaze (approved):
not sure about having it be solid white, that doesn't scream 'stamina' to me. that's just a style/presentation thing though

i will say, it would be nice if it did a smooth fade-in/fade-out via alpha instead of just vanishing entirely.

MistakeNot4892:
I would like to do a pass on progress bars in general, the state-based approach is kinda naff. I've done something similar using a transform on a box on other projects that should apply here.

PR #5271, opened 2026-01-11
Expanding language keys and adding mixed-language speech.

MistakeNot4892:
## Description of changes
This was just supposed to be the keys stuff but I kinda spiralled.
- Languages can have keys of any length.
- Language keys must be text.
- Language keys in speech must be followed by a space.
- Using a numerical key for speech will look for that language sequentially in your language list. ie. `,1` is your first known language.
- Partial keys will use the first matching language in your language list. `,h` will match `,human`.
- You can use a language key in the middle of a sentence to switch language.
- Rewrote holopads to handle message passing in a better/more coherent way with less code.

image

Mixing languages in speech is good fun for RP, and the backend changes have cleaned up, centralized a fair wodge of logic.

## TODO
- [X] Rewrite holopads to handle phrases properly. Oh god oh fuck.
- [X] Sort out why the MC is soft resetting after speech (probably a blocking loop somewhere).
- [X] Test basic speech (no language, one language, multiple languages)
- [X] Test basic speech over radio.
- [x] Test basic speech over broadcast (binary).
- [X] Test basic speech over holopad.
- [x] Test mixed broadcast (binary) and spoken.
- [X] Test mixed sign/nonvocal and spoken.

## Issues
- [x] Overhead text not offset high enough.
- ~~Animal noises are not scrambled.~~ Type understands type, intentional.
- [x] Sign language not transmitted over holopad.
- [x] Cannot speak plain binary (gives 'you cannot mix non-spoken and spoken' error).

## Authorship
Myself.

## Changelog
TBD


code/modules/mob/living/say.dm
```diff
@@ -5,12 +5,13 @@
...
-/mob/living/proc/get_default_language()
+/mob/living/get_default_language()
 	var/decl/language/lang = GET_DECL(default_language)
 	if(istype(lang) && can_speak(lang))
 		return lang
+	return get_any_good_language(set_default=TRUE)
 
-/mob/living/proc/get_any_good_language(set_default=FALSE)
+/mob/living/get_any_good_language(set_default=FALSE)
 	. = get_default_language()
```
out-of-phaze: this results in infinite recursion if `default_language` is unset.

out-of-phaze (dismissed):

code/modules/mob/living/say.dm
MistakeNot4892: Awesome, thank you for spotting that.

MistakeNot4892: Added a bool to prevent looping.

MistakeNot4892:
I need to do some more exhaustive testing before merge, but I think this is good for review now.

MistakeNot4892:
Binary radio headset does allow you to send different languages over binary, but that is a niche enough case that I don't want to fix it for this PR.

MistakeNot4892:
Last issue is that overhead text flickers and is out of place for 1 tick immediately after creation. I have no idea why, logic looks straightforward.

code/_helpers/game.dm
```diff
@@ -204,19 +204,22 @@
 	var/list/hearturfs = list()
 	FOR_DVIEW(var/turf/T, range, center, INVISIBILITY_MAXIMUM)
 		hearturfs[T] = TRUE
-		for(var/mob/M in T)
-			mobs += M
+		if(islist(mobs))
+			for(var/mob/M in T)
+				mobs += M
 	END_FOR_DVIEW
```
out-of-phaze: a bit of code duplication but for things as hot as this it might be worth it to move the `islist(mobs)` check all the way out of the loop and instead have two different ones
```
	if(islist(mobs))
		FOR_DVIEW(var/turf/T, range, center, INVISIBILITY_MAXIMUM)
			hearturfs[T] = TRUE
			for(var/mob/M in T)
				mobs += M
		END_FOR_DVIEW
	if(else)
		FOR_DVIEW(var/turf/T, range, center, INVISIBILITY_MAXIMUM)
			hearturfs[T] = TRUE
		END_FOR_DVIEW
```

code/_helpers/game.dm
```diff
@@ -204,19 +204,22 @@
...
-
-	for(var/obj/O in global.listening_objects)
-		if(hearturfs[get_turf(O)])
-			objs += O
+	if(islist(mobs))
+		for(var/mob/M in global.player_list)
+			if(check_ghosts && M.stat == DEAD && M.get_preference_value(check_ghosts) != PREF_NEARBY)
+				mobs |= M
+			else if(hearturfs[get_turf(M)])
+				mobs |= M
```
out-of-phaze: hold on, won't this always be redundant because we added the mobs earlier when we set hearturfs? or is this for recursive contents given the get_turf? because in that case the virtual mobs system may work better for that kind of thing, idk

code/_helpers/game.dm
```diff
@@ -204,19 +204,22 @@
...
+
+	if(islist(objs))
+		for(var/obj/O in global.listening_objects)
```
out-of-phaze, line 224: note to self: `global.listening_objects_by_z` could be a really good optimization here

code/_helpers/text.dm
```diff
@@ -365,6 +365,15 @@ var/global/regex/starts_lowercase_regex = regex(@"^[a-z]")
 /proc/trim(text)
 	return trim_left(trim_right(text))
 
+/// Adds punctuation to an emote or speech message automatically.
+/proc/handle_autopunctuation(message)
```
out-of-phaze, line 369: note to self: moved from `code/modules/mob/say.dm`

out-of-phaze (dismissed):
stopped reviewing at holopad because that's gonna take a while to get through. will look into the text over head issue in the meantime

code/_helpers/game.dm
MistakeNot4892: Why? This will mean we're doing two full separate dviews.

out-of-phaze: i must have been sleep deprived when i wrote that, `else` instead of `if(else)`. only one will run at a time, we're just hoisting the check out of the loop

code/_helpers/game.dm
MistakeNot4892: It does seem to be completely redundant since any mob on a turf in `hearturfs` has already been added.

code/_helpers/text.dm
```diff
@@ -365,6 +365,15 @@ var/global/regex/starts_lowercase_regex = regex(@"^[a-z]")
...
+	if(!message)
+		return
+	var/end_char = copytext_char(trim_right(strip_html_properly(message)), -1)
+	if(!(end_char in list(".", "?", "!", "-", "~")))
```
Lohikar: Is this called much? Using a switch here would avoid the list allocation, but the list allocation is probably not significant.

code/game/atoms.dm
```diff
@@ -642,22 +642,22 @@
 	- `range?`: The number of tiles away the message will be visible from. Default: world.view
 	- `check_ghosts?`: Set to `TRUE` if ghosts should see the message if their preferences allow
 */
-/atom/proc/visible_message(var/message, var/self_message, var/blind_message, var/range = world.view, var/check_ghosts = null)
+/atom/proc/visible_message(var/message, var/self_message, var/blind_message, var/range = world.view, var/check_ghosts = null, atom/source)
```
out-of-phaze:
```suggestion
/atom/proc/visible_message(var/message, var/self_message, var/blind_message, var/range = world.view, var/check_ghosts = null, atom/source = null)
```
i don't think you can mix keyword/non-keyword args like this? maybe i'm thinking of python or C# though, but it's also a style thing

code/game/atoms.dm
```diff
@@ -670,18 +670,18 @@
 	- `check_ghosts?`: TRUE if ghosts should hear the message if their preferences allow
 	- `radio_message?`: The string to send over radios
 */
-/atom/proc/audible_message(var/message, var/deaf_message, var/hearing_distance = world.view, var/check_ghosts = null, var/radio_message)
+/atom/proc/audible_message(var/message, var/deaf_message, var/hearing_distance = world.view, var/check_ghosts = null, var/radio_message, atom/source)
```
out-of-phaze:
```suggestion
/atom/proc/audible_message(var/message, var/deaf_message, var/hearing_distance = world.view, var/check_ghosts = null, var/radio_message, atom/source = null)
```
see above. would also like consistency, `var/` to match the rest of the args or removing it from the rest. not like, a strict thing though i'd just like it

code/modules/fabrication/fabricator_food.dm
```diff
@@ -16,22 +16,27 @@
 	global.listening_objects -= src
 	return ..()
 
-/obj/machinery/fabricator/replicator/hear_talk(var/mob/M, var/text, var/verb, var/decl/language/speaking)
-	if(speaking && !speaking.machine_understands)
+/obj/machinery/fabricator/replicator/hear_talk(mob/living/speaker, datum/speech/phrases, verb, stars = FALSE, decl/language/force_language)
```
out-of-phaze:
```suggestion
/obj/machinery/fabricator/replicator/hear_talk(mob/living/speaker, datum/speech/phrases, verb, stars = FALSE, decl/language/force_language = null)
```
see above about mixing kwargs and positional args, i think we prefer an explicit null at the very least

code/modules/implants/implant_types/explosive.dm
```diff
@@ -91,13 +91,13 @@
 	frequency = new_frequency
 	radio_connection = radio_controller.add_object(src, frequency, RADIO_CHAT)
 
-/obj/item/implant/explosive/hear_talk(mob/M, msg)
-	hear(msg)
+/obj/item/implant/explosive/hear_talk(mob/living/speaker, datum/speech/phrases, verb, stars = FALSE, decl/language/force_language)
```
out-of-phaze:
```suggestion
/obj/item/implant/explosive/hear_talk(mob/living/speaker, datum/speech/phrases, verb, stars = FALSE, decl/language/force_language = null)
```
again explicit `= null` once we've started kwargs/things

out-of-phaze (dismissed):
i don't actually feel strongly enough about these to request they all be changed, so shrug

code/modules/fabrication/fabricator_food.dm
MistakeNot4892: I went through and removed the `= FALSE` earlier

MistakeNot4892:
Selfmerging this after partial review and discussion, per Discord.

PR #5275, opened 2026-01-15
Various fixes/tweaks for unit tests.

MistakeNot4892:
- Allows bodytypes to skip organ validation (needed for bodytypes that don't implement a full set of organs, like specific prosthetic models).
- Makes category generation more explicit about failures in fusion reaction info.
- Fixes a copypaste error.

### DIFF
--- a/code/modules/bodytype/_bodytype.dm
+++ b/code/modules/bodytype/_bodytype.dm
@@ -89,6 +89,8 @@ var/global/list/bodytypes_by_category = list()
 	var/z_flags              = 0
 	/// Amount to shift overlays when lying. TODO: check if this is still needed with KEEP_TOGETHER
 	var/list/prone_overlay_offset
+	/// Set to TRUE to skip unit testing as a primary bodytype in a human. Generally for partial prosthetic models.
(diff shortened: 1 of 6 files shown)


code/modules/bodytype/_bodytype.dm
```diff
@@ -552,7 +554,7 @@ var/global/list/bodytypes_by_category = list()
 				. += "non-list default metadata for [acc_decl.type]: [accessories[accessory] || "NULL"]"
 
 	var/list/tail_data = has_limbs[BP_TAIL]
-	if(tail_data)
+	if(tail_data && !TYPE_IS_ABSTRACT(src))
```
out-of-phaze: i don't think abstract bodytypes can even be instantiated or validated, can they? did this fix a real practical issue? because afaict it should never be triggered.

out-of-phaze (dismissed):

code/modules/bodytype/_bodytype.dm
MistakeNot4892: Oh, no, I just chucked it in there in case.

out-of-phaze (approved):

PR #5276, opened 2026-01-15
Misc Polneb changes

MistakeNot4892:
Subset of the changes from #5262.


code/game/objects/random/subtypes/tech.dm
```diff
@@ -120,3 +120,20 @@
 		/obj/item/oxycandle
 	)
 	return spawnable_choices
+
+/obj/random/hardsuit
```
out-of-phaze: `/obj/random/hardsuit` exists in `code/game/objects/random/subtypes/suits.dm`

code/game/objects/random/subtypes/food.dm
```diff
@@ -194,3 +194,13 @@
 		/obj/item/food/plumphelmetbiscuit
 	)
 	return spawnable_choices
+
+/obj/random/mug
```
out-of-phaze: in theory this could be better placed next to `/obj/random/drinkingglass` in misc.dm, or that could be moved here. not blocking though

code/game/turfs/floors/subtypes/floor_tiled.dm
```diff
@@ -68,6 +72,10 @@
 	icon_state    = "techfloor_gray"
 	_flooring     = /decl/flooring/tiling/tech
 
+/turf/floor/tiled/techfloor/cryo
+	initial_gas = list(/decl/material/gas/nitrogen = MOLES_CELLSTANDARD)
+	temperature = 73
```
out-of-phaze: i sort of wonder if it might be better to write this as `-200 CELSIUS`, since `73` seems arbitrary whereas it's obvious when written with the celsius macro why it was chosen

code/modules/clothing/masks/smokable.dm
```diff
@@ -135,9 +135,10 @@
 			return
 		lit = TRUE
 		atom_damage_type = BURN
-		if(REAGENT_VOLUME(reagents, /decl/material/liquid/fuel)) // the fuel explodes
+		var/explosive_power = round(REAGENT_VOLUME(reagents, /decl/material/liquid/fuel) / 5, 1)
```
out-of-phaze: this should maybe be generalised to a reagent var? not sure. because like, i remember doing this on lighthouse to allow phoron to burn in cigarettes, and it'd be nice if readding phoron required as few core edits as possible.

code/modules/reagents/chems/chems_medicines.dm
```diff
@@ -30,10 +30,11 @@
 	value = 1.5
 	exoplanet_rarity_gas = MAT_RARITY_EXOTIC
 	uid = "chem_antirads"
+	var/antirad_power = 30
```
out-of-phaze, line 33: this was to readd arithrazine in the baychems modpack, right? makes sense, i think toxins have a similar var for toxin strength. might need to do the same for oxygel so that we can adjust what level of `CE_OXYGENATED` it gives, we support 1-3 but it only ever gives 1.

out-of-phaze (changes requested):

code/modules/reagents/chems/chems_medicines.dm
MistakeNot4892: Yeah, that's from removing dexplus I think.

out-of-phaze (approved):

PR #5277, opened 2026-01-16
Various other Polaris/Cynosure changes.

MistakeNot4892:
Many small atomic commits, should be self-documenting.


code/game/objects/structures/flaps.dm
```diff
@@ -32,7 +33,7 @@
 
 	var/mob/living/M = A
 	if(istype(M))
-		if(M.current_posture.prone && !M.buckled)
+		if(M.current_posture.prone && can_pass_lying)
```
out-of-phaze, line 36: does this mean you can shove a rollerbed through with someone on it now? i figure the buckled check was to avoid exactly that

code/game/objects/items/candelabra.dm
```diff
@@ -49,3 +49,6 @@
 			if(i > length(candles_storage.candle_offsets))
 				break
 	compile_overlays()
+
+/obj/item/candelabra/infinite/WillContain()
```
out-of-phaze, line 53: not sure how to feel about infinite candle/candelabra subtypes. i guess it's fine but it's still Weird to me for stuff that's mapped in to have weird exceptions. (this isn't blocking i just included it with my other comments)

out-of-phaze (changes requested):

code/game/objects/structures/flaps.dm
MistakeNot4892: That seems fine to me particularly considering grabs and crawling are both easier ways to move under that aren't prevented by the check. I dunno, I can readd it if you really want it. Just seems pointless in a post-crawling world.

out-of-phaze (approved):

code/game/objects/items/candelabra.dm
MistakeNot4892: Yeah I'm not totally sure about them either. I could just map unlit candles in instead.

out-of-phaze: not a huge deal but i would personally prefer that

Issue #5011, opened 2025-05-13
Heaters and freezers need to be unified


out-of-phaze:
11 year old `TODO:` comment:
`//TODO: Put this under a common parent type with freezers to cut down on the copypasta`
on `/obj/machinery/atmospherics/unary/freezer` and `/obj/machinery/atmospherics/unary/heater`.

PR #5281, opened 2026-01-23
Unify unary freezers/heaters under a shared type

out-of-phaze:
## Description of changes
Unifies `/obj/machinery/atmospherics/unary/freezer` and `/obj/machinery/atmospherics/unary/heater` under `/obj/machinery/atmospherics/unary/temperature` to reduce code duplication. Most of their code was identical already, so it only required minor changes.
- [x] Tested

Completes #5011.

## Authorship
Me


out-of-phaze:
Noticed some weirdness with the heater overshooting the target temperature by a lot, but that might have already existed and is just an issue when dealing with small gas amounts...?

out-of-phaze:
fixed that issue too

MistakeNot4892 (approved):

Issue #5288, opened 2026-02-04
Fashioning splints/torches (cloth + stick) is not working


Typhin:
#### Description of issue
Using cloth bolts or other items on sticks to create torches or splints does not work.  It brings up the "Do you want to make a torch or a splint?" menu, but then nothing happens.

#### Difference between expected and actual behavior
After making a selection, the item should be crafted, or an error message should explain why it could not be crafted.

#### Steps to reproduce
Make a stick (in-hand crafting from wooden plank)
Have a stack of cloth bolts or other material (at least 3 for Torch, 5 for Splint)
Use material on stick
Select desired item
...Nothing happens.

#### Specific information for locating

`code/game/objects/items/weapons/material/stick.dm:30` has an error in the initial test to determine if a selection has been made and everything is ready to proceed.  When it's testing to ensure the relevant objects still exist, it checks `!QDELETED(src)` but it should be testing without the negation.  As it is, it only attempts to proceed if the stick has been deleted between the click to select a desired product and the selection of a desired product.  Removing the negation allows the creation of items as expected.

#### Length of time in which bug has been known to occur

Shortly after Shaded Hills first appeared in Scav Station, I had discovered the interaction but could not get items to craft.  I then forgot about it until splints were mentioned due to the travois issue and I mentioned splinting broken limbs and stopping bleeding was a way to prevent dragging damage.  Another player mentioned needing a way to craft splints, and that's when I remembered the interaction and began my investigation.

Verified on Nebula/dev.

#### Client version, Server revision & Game ID

Client Version: 516
Server Revision: 1ffd13449104ed6fb8a0b06773fdbbf74c7e545c - dev - 2026-02-04
Game ID: cDT-dAjx
Current map: Tradeship Ocelot Alpha

#### Issue bingo

- [x] Issue could be reproduced at least once
- [x] Issue could be reproduced by different players
- [x] Issue could be reproduced in multiple rounds
- [x] Issue happened in a recent (less than 7 days ago) round
- [x] [Couldn't find an existing issue about this](https://github.com/NebulaSS13/Nebula/issues)

out-of-phaze:
Fixed in #5290.

PR #5289, opened 2026-02-04
Fix Issue #5288 - crafting torch/splint with stick

Typhin:
Fixes the ability to craft torches/splints by using materials on sticks.

## Description of changes

Fixes the one-character issue in `code/game/objects/items/weapons/material/stick.dm:30` to allow crafting torches and splints to proceed correctly.

Allows the crafting of torches and splints as intended.

## Authorship

Typhin

## Changelog
:cl:
bugfix: fixed the conditional preventing crafting
/:cl:


MistakeNot4892:
This will need to target `stable` for it to go into Scav in a reasonable timeframe

MistakeNot4892:
Obviated by #5290

PR #5292, opened 2026-02-08
Assorted changes from Cynosure branch.

MistakeNot4892:
Commit list should be descriptive.


code/_helpers/visual_filters.dm
```diff
@@ -63,16 +63,19 @@
...
-	monkeypatched_params[1] = filters[index]
+	var/list/monkeypatched_params = params.Copy()
+	monkeypatched_params.Insert(1, filters[index])
 	animate(arglist(monkeypatched_params))
 
 	// If we're being copied by Z-Mimic, update mimics too.
 	if (bound_overlay)
 		for (var/atom/movable/AM as anything in get_above_oo())
+			if(index > length(AM.filters))
+				continue // presumably there is a z-mimic flush pending - TODO check if this would fuck with the wrong filter on the shadower
```
out-of-phaze, line 79: is this a "before it's ready to review/merge" TODO or a long-term "when i get around to it" TODO?

code/controllers/evacuation/evacuation_lifepods.dm
```diff
@@ -17,7 +17,9 @@
 	if(waiting_to_leave())
 		return
 	state = EVAC_IN_TRANSIT
-	priority_announcement.Announce(replacetext(replacetext(global.using_map.emergency_shuttle_leaving_dock, "%dock_name%", "[global.using_map.dock_name]"),  "%ETA%", "[round(get_eta()/60,1)] minute\s"))
+	var/msg = global.using_map.emergency_shuttle_leaving_dock
+	if(msg)
+		priority_announcement.Announce(replacetext(replacetext(msg, "%dock_name%", "[global.using_map.dock_name]"),  "%ETA%", "[round(get_eta()/60,1)] minute\s"))
```
out-of-phaze, line 22: note to self: we should centralize map message token replacement in a proc somewhere so it can be reused.

code/game/area/areas.dm
```diff
@@ -13,6 +13,9 @@ var/global/list/areas = list()
 	luminosity =    0
 	mouse_opacity = MOUSE_OPACITY_UNCLICKABLE
 
+	// If true, this area will force light switches on during init.
+	var/start_lit = FALSE
```
out-of-phaze: hm, this is *okay* but what would you think about using the `lightswitch` var for this? where `null` -> use config ('lights start on' setting), FALSE/TRUE -> start as that value?

code/game/objects/structures/watercloset.dm
```diff
@@ -92,9 +92,8 @@ var/global/list/hygiene_props = list()
 				visible_message("\The [src] gurgles and overflows!")
 				next_gurgle = world.time + 80
 				playsound(T, pick(SSfluids.gurgles), 50, 1)
-			var/adding = min(flood_amt - REAGENT_TOTAL_VOLUME(T?.reagents), rand(30,50)*clogged)
-			if(adding > 0)
-				T.add_to_reagents(/decl/material/liquid/water, adding)
+			if(loc.get_fluid_depth() < flood_amt)
+				T.add_to_reagents(/decl/material/liquid/water, round(rand(10,20)*clogged))
```
out-of-phaze, line 96: notably this will actually allow it to flood the turf up to `20*clogged-1` above `flood_amt`, i think. it's not a huge deal but it IS technically different than the previous one, which would add only enough to get the total level in the turf up to `flood_amt`.

code/game/turfs/turf_height.dm
```diff
@@ -1,5 +1,6 @@
 /turf/proc/set_physical_height(var/new_height)
 	return
 
+// Open turfs should count as as low as possible for the purposes of fluid flows, etc.
```
out-of-phaze, line 4: oh huh, i think this is a change i did locally and forgot to mention anywhere. nice. except this is a slightly better implementation than what i did

code/modules/lights/light_switch.dm
```diff
@@ -44,6 +44,11 @@
 	. = ..()
 	if(isnull(on))
 		on = get_config_value(/decl/config/toggle/lights_start_on)
+	if(!on)
+		var/area/area = get_area(src)
+		if(area?.start_lit)
```
out-of-phaze: should use `connected_area` instead

code/modules/mining/mine_items.dm
out-of-phaze, line 31: you know it still messes with me that these are called `/obj/item/stack/flag`. like i know why they are, historically speaking, but dang

code/modules/mob/living/bot/bot.dm
```diff
@@ -42,6 +42,8 @@
 /mob/living/bot/Initialize()
 	. = ..()
 	update_icon()
+	add_language(/decl/language/human/common)
+	set_default_language(GET_DECL(/decl/language/human/common))
```
out-of-phaze, line 46: wonder if we should have a `/datum/map/var/default_language` or `lingua_franca` so that downstreams won't have to change it if their default language is different than `/decl/language/human/common`. not blocking obvs that'd be a larger change

code/modules/multiz/level_data.dm
```diff
@@ -354,6 +354,22 @@
...
+		var/turf/floor/above = GetAbove(below)
+		if(!istype(above))
+			continue
+		var/area/area = get_area(below)
+		if(!istype(area) || !(area.area_flags & AREA_FLAG_CONSTRUCTED))
+			continue
+		var/decl/flooring/base_flooring = above.get_base_flooring()
+		if(!base_flooring || base_flooring.constructed)
+			continue
+		above.set_base_flooring(/decl/flooring/plating) // TODO: check if we can skip update here.
```
out-of-phaze, line 371: kind of feel like we should have some kind of method for inserting flooring at the bottom of a turf, like "replace base flooring and add the old base flooring to the bottom of the normal flooring list" or something. not blocking obvs

code/modules/reagents/reagent_containers/hypospray.dm
```diff
@@ -197,6 +197,10 @@
 	if(REAGENT_TOTAL_VOLUME(reagents) > 0 && autolabel && !label_text) // don't override preset labels
 		label_text = "[reagents.get_primary_reagent_name()], [REAGENT_TOTAL_VOLUME(reagents)]u"
 
+/obj/item/chems/hypospray/autoinjector/used/Initialize()
+	. = ..()
+	atom_flags &= ~ATOM_FLAG_OPEN_CONTAINER
+
```
out-of-phaze: already exists a few lines above

code/controllers/subsystems/shuttle.dm
```diff
@@ -42,7 +42,7 @@ SUBSYSTEM_DEF(shuttle)
 	while (working_shuttles.len)
 		var/datum/shuttle/shuttle = working_shuttles[working_shuttles.len]
 		working_shuttles.len--
-		if(shuttle.process_state && (shuttle.Process(wait, times_fired, src) == PROCESS_KILL))
+		if((shuttle.always_process || shuttle.process_state) && (shuttle.Process(wait, times_fired, src) == PROCESS_KILL) && !shuttle.always_process)
```
out-of-phaze, line 45: i don't know why this is necessary, shouldn't you just make your shuttle not return `PROCESS_KILL` in `/Process`? also this whole thing is kind of odd even prior to your changes, this is all autodock-shuttle-specific but it's on the base type.

code/modules/shuttles/shuttle.dm
```diff
@@ -453,4 +461,19 @@
...
+		var/area/mob_area = get_area(C.mob)
+		if(!istype(mob_area))
+			continue
+		// I don't know how much of this is actually needed.
+		if(ispath(shuttle_area) && istype(mob_area, shuttle_area))
+			C.mob.visible_message(message)
+		else if(istype(shuttle_area, /area) && mob_area == shuttle_area)
+			C.mob.visible_message(message)
+		else if(islist(shuttle_area) && ((mob_area in shuttle_area) || (mob_area.type in shuttle_area)))
+			C.mob.visible_message(message)
```
out-of-phaze: uh... i don't think this is good practice, no
```dm
/datum/shuttle/autodock/overmap/fuel_check()
	if(!try_consume_fuel()) //insufficient fuel
		for(var/mob/hearer in global.living_mob_list_ + global.ghost_mob_list)
			if(is_type_in_list(get_area(hearer), shuttle_area))
				hearer.show_message(SPAN_WARNING("You hear the shuttle engines sputter... perhaps it doesn't have enough fuel?"), AUDIBLE_MESSAGE,
				SPAN_WARNING("The shuttle shakes but fails to take off."), VISIBLE_MESSAGE)
		return FALSE //failure
	return TRUE //sucess, continue with launch
```
this is what other code does and i believe it works. regardless `visible_message` here is incorrect regardless, it should use `to_chat` or at the very least `show_message`.

mods/content/response_team/ert_rig.dm
```diff
@@ -217,3 +217,50 @@
...
+
+/obj/item/rig/ia/equipped
+	req_access = list(access_lawyer)
+	initial_modules = list(
+		/obj/item/rig_module/ai_container,
+		/obj/item/rig_module/device/flash,
+		/obj/item/rig_module/device/paperdispenser,
+		/obj/item/rig_module/device/pen,
+		/obj/item/rig_module/device/stamp
+		)
```
out-of-phaze: `/obj/item/rig/light/internalaffairs` already exists, it's just a different typepath.

out-of-phaze (changes requested):

code/_helpers/visual_filters.dm
MistakeNot4892: I don't have the knowledge to do it, I intended it as a long term TODO.

code/game/objects/structures/watercloset.dm
MistakeNot4892: The previous one was broken in some way and would flood until max depth. I'm alright with a bit of imprecision but I can make it more refined if you want.

code/controllers/subsystems/shuttle.dm
MistakeNot4892: It needs to process, and skip removing from the processing list, regardless of process_state. It could certainly be implemented better, this is cribbed from Polaris.

out-of-phaze: what shuttle type is it even for, though? i can't see a case where it would be removed from the processing list, unless it straight up didn't have a Process() override, at which point... why does it need to process?

code/game/objects/structures/watercloset.dm
out-of-phaze: nah, this one's probably fine.

code/controllers/subsystems/shuttle.dm
MistakeNot4892: It's supporting code specifically for the arrivals shuttle, which isn't on main code. It's on map/cynosure if you wanted to check it and suggest a rewrite. All I recall is that it was not ticking when idle until I re-added always_process

code/game/area/areas.dm
MistakeNot4892: I would rather not do this because it would require setting `lightswitch` on a bunch of areas, not really up for it currently. Wouldn't be a bad idea for its own PR though.

code/modules/mob/living/bot/bot.dm
MistakeNot4892: Yeah it'd be good considering `default_species` is map-level.

code/modules/shuttles/shuttle.dm
MistakeNot4892: Generalized the existing proc.

out-of-phaze (approved):

PR #5295, opened 2026-02-12
Final Cynosure branch changes for sync.

MistakeNot4892:
Last one, I promise.


out-of-phaze (approved):
not sure about putting a new fluff cat in core code but it's honestly whatever

MistakeNot4892:
> not sure about putting a new fluff cat in core code but it's honestly whatever

I can move him into Cynosure, main change for main code was the tagging override.

PR #5300, opened 2026-02-16
Shotgun icon fix

l3lb0t:
## Description of changes
I noticed that pump action shotguns check for shells in the tube, not shells in the chamber when decided whether to display as empty. This doesn't reflect whether the shotgun is actually "loaded", as you can still have a shell in the chamber.

You can now tell if a shotgun is loaded by looking at it, rather than guessing on whether or not there's a shell in the chamber. This is better from a gameplay standpoint, more immersive, and more realistic.

## Authorship
It's me, Karl Johansson

## Changelog
:cl: Karl Johansson
bugfix: Fixes pump-action shotgun icons incorrectly displaying the state of the weapon
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/projectile/shotgun.dm
+++ b/code/modules/projectiles/guns/projectile/shotgun.dm
@@ -24,6 +24,12 @@
 	else
 		icon_state = "[get_world_inventory_state()]-empty"
 
+/obj/item/gun/projectile/shotgun/pump/update_base_icon_state()
+	if(chambered)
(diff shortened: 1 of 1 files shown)


MistakeNot4892:
Should this be going to `stable`? Fixes typically go there first rather than dev.

l3lb0t:
is that so

l3lb0t:
Moved to [#5303](https://github.com/NebulaSS13/Nebula/pull/5303)

PR #5302, opened 2026-02-16
Fixes rejuvenate not preserving skin color

l3lb0t:
## Description of changes
I noticed in testing that if you chop your arm off, and regrow it with the Rejuvenate verb (something I often find myself doing), the arm you grow back is inexplicably white, regardless of your character's actual skin color.
image

Now, if you regrow a limb, it will actually match your body, rather than looking like you stole someone else's arm. This is probably a good thing.

## Authorship
The one, the only, Karl Johansson

## Changelog
:cl: Karl Johansson
bugfix: Regrown limbs now match the skin color of the rest of your body
/:cl:


l3lb0t:
ah christ this ain't it

PR #5303, opened 2026-02-16
Shotgun icon fix (Take Two)

l3lb0t:
## Description of changes
I noticed that pump action shotguns check for shells in the tube, not shells in the chamber when decided whether to display as empty. This doesn't reflect whether the shotgun is actually "loaded", as you can still have a shell in the chamber.

You can now tell if a shotgun is loaded by looking at it, rather than guessing on whether or not there's a shell in the chamber. This is better from a gameplay standpoint, more immersive, and more realistic.

## Authorship
It's me, Karl Johansson

## Changelog
:cl: Karl Johansson
bugfix: Fixes pump-action shotgun icons incorrectly displaying the state of the weapon
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/projectile/shotgun.dm
+++ b/code/modules/projectiles/guns/projectile/shotgun.dm
@@ -24,6 +24,12 @@
 	else
 		icon_state = "[get_world_inventory_state()]-empty"
 
+/obj/item/gun/projectile/shotgun/pump/update_base_icon_state()
+	if(chambered)
+		icon_state = get_world_inventory_state()
+	else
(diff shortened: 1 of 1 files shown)


code/modules/projectiles/guns/projectile/shotgun.dm
out-of-phaze, line 21: you know, this is actually the only thing that directly references `/obj/item/gun/projectile/shotgun`. while it's a shared type used for both the pump shotgun and double-barreled shotgun, they don't actually share any behavior aside from this override, which isn't even correct for the pump shotgun. i should fix that

out-of-phaze (approved):

PR #5314, opened 2026-03-22
Prevent Garlic Oil from being poison

Typhin:
## Description of changes

Tags garlic oil as `injectable_nutrition`.  Because this causes `adjust_mob_nutrition()` to get called twice, nutriment/hydration factors are cut in half to compensate.

Garlic Oil now can be used as an antibiotic without dealing 1.2 TOX damage per unit ingested (due to a food substance being present in the bloodstream).

## Changelog
:cl:
tweak: Prevented Garlic Oil from dealing TOX damage
/:cl:

### DIFF
--- a/code/modules/reagents/chems/chems_drinks.dm
+++ b/code/modules/reagents/chems/chems_drinks.dm
@@ -18,7 +18,8 @@
 
 /decl/material/liquid/drink/affect_blood(var/mob/living/M, var/removed, var/datum/reagents/holder)
 	. = ..()
-	M.take_damage(removed, TOX) // Probably not a good idea; not very deadly though
+	if(!injectable_nutrition)
+		M.take_damage(removed, TOX) // Probably not a good idea; not very deadly though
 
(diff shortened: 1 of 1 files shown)


MistakeNot4892:
Hmmm I don't much like having it process twice, it means immunity and such will be adjusted twice. Not sure of a better fix short of duplicating the logic into affect_ingest on garlic juice specifically, which is gross too.

Typhin:
The only other thing I can think of to do is have `affect_ingest()` basically skip all processing if it has both `injectable_nutrition` and `affect_blood_on_ingest` so it only processes once.  (And since `affect_blood_on_ingest` is a ratio value rather than simple true/false, it could do something like `digest_amount = (1 - affect_blood_on_ingest)` and then use that to multiply the effects of digestion.  Basically it only processes the total amount but in two passes, but an ingest portion and an inject portion.)

The "duplicating logic into affect_ingest()" unpleasantness is why I went with the `affect_blood_on_ingest` route, only to run into the "food in bloodstream deals TOX damage" problem.

MistakeNot4892 (approved):
Going to approve this for now but I might revisit it with another solution in the future.

PR #5315, opened 2026-03-30
Fix accessories vanishing when clothing moves, hoodie bugs

### LINKED ISSUE #5199: Clothing Sprites Disappear when in Backpack or Other Storage Item
#### Description of issue

Certain clothing sprites that are worn in the uniform slot have their sprites not render correctly when placed into a storage item like a backpack. This makes them difficult to remove and in some events, impossible as they cannot be detected properly, leading to a large chunk of space taken up by an item that cannot be removed. This only seems to occur when the uniform item has things attached to it. Normal baggy pants would render fine in a container, but would fail to render if a suit jacket was attached to it.

Seeing how the sprites render when equipped in the Uniform slot, and in the two hand slots, I feel like the sprites should also render when placed into a backpack or other storage item.

#### Steps to reproduce

(How I did it)
1. I equipped baggy pants with a suit jacket and a necklace attached in my character loadout. (Depending on the role. Jobs with have loadout items in the suit slot tend to attach the items automatically, while ones that don't will have them equipped separately in different slots.)
2. I unequipped the uniform and placed it in my backpack (which in this scenario was tailbags, but also worked when placed in a red toolbox.)
3. The sprite would not render when in the container.

/obj/item/clothing/pants/baggy 
/obj/item/clothing/suit/jacket 
"_Sorry, the following is not valid: baggy-pants.0x200c028
usage: Delete mob|obj|turf_" (When trying to admin delete the item with a missing sprite in a container)

At least since the 1st of November, 2025.

Client Version: 516
Server Revision: 871324d5b3d9824e9889ec06e55944651656f3ae - main - 2025-11-01
Game ID: cCm-dpqC

#### Issue bingo

out-of-phaze:
## Description of changes
Accessory display no longer uses appearance copy, and instead sets icon, icon_state, underlays, and (managed) overlays. This will hopefully make it less brittle. This also means compile_overlays won't clobber the overlays.

Fixes #5199.
Fixes an issue where hoodie hoods had the wrong material and color, and showed up in the hoodie's storage.

## Authorship
Me

### DIFF
--- a/code/modules/clothing/suits/_suit_hood.dm
+++ b/code/modules/clothing/suits/_suit_hood.dm
@@ -16,9 +16,18 @@
 	if(. && istype(hood))
 		hood.set_color(new_color)
 
+// i don't think this needs to be a get_stored_inventory override instead since it should never be separated currently
+// but that may change in the future
+/obj/item/clothing/suit/get_contained_external_atoms()
+	. = ..()
+	if(istype(hood)) // this is considered a component rather than a contained item
+		. -= hood
+
 /obj/item/clothing/suit/Initialize()
 	if(ispath(hood))
-		hood = new hood(src)
+		hood = new hood(src, material)
+		if(paint_color)
+			hood.set_color(paint_color)
 	return ..()
 
 /obj/item/clothing/suit/Destroy()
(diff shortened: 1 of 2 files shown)


code/modules/clothing/suits/_suit_hood.dm
```diff
@@ -16,9 +16,18 @@
...
+	. = ..()
+	if(istype(hood)) // this is considered a component rather than a contained item
+		. -= hood
+
 /obj/item/clothing/suit/Initialize()
 	if(ispath(hood))
-		hood = new hood(src)
+		hood = new hood(src, material)
+		if(paint_color)
+			hood.set_color(paint_color)
```
MistakeNot4892, line 30: I swear this should already be being done somewhere. I wonder if I never cherrypicked it out of the cynosure stuff.

MistakeNot4892: Oh, might be fixed on dev.

MistakeNot4892 (approved):

PR #5316, opened 2026-04-02
Fixes runtime with hood override.

MistakeNot4892:
Also duplicates some changes from dev to make a merge easier later.

### DIFF
--- a/code/modules/clothing/suits/_suit_hood.dm
+++ b/code/modules/clothing/suits/_suit_hood.dm
@@ -20,7 +20,7 @@
 // but that may change in the future
 /obj/item/clothing/suit/get_contained_external_atoms()
 	. = ..()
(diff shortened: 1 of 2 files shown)


out-of-phaze (approved):

code/modules/clothing/suits/_suit_hood.dm
```diff
@@ -20,7 +20,7 @@
 // but that may change in the future
 /obj/item/clothing/suit/get_contained_external_atoms()
 	. = ..()
-	if(istype(hood)) // this is considered a component rather than a contained item
+	if(islist(.) && istype(hood)) // this is considered a component rather than a contained item
 		. -= hood
```
out-of-phaze, line 24: i probably should've used lazyremove here

PR #5334, opened 2026-04-09
Mark some unused variables

out-of-phaze:
## Description of changes
Marks some unused variables I don't feel comfortable implementing myself, mostly because they're other people's code.

Leaves a visible note that these variables are unused and need to be implemented or removed.

### DIFF
--- a/code/_onclick/hud/screen/screen_intent.dm
+++ b/code/_onclick/hud/screen/screen_intent.dm
@@ -70,7 +70,7 @@
 	requires_ui_style    = FALSE
 	apply_screen_overlay = FALSE
 	var/intent_width     = 16
-	var/intent_height    = 16
+	var/intent_height    = 16 // currently unused
 	var/list/intent_selectors
 
 /obj/screen/intent/Initialize(mapload, mob/_owner, decl/ui_style/ui_style, ui_color, ui_alpha, ui_cat)
--- a/code/modules/backgrounds/citizenship/_citizenship.dm
+++ b/code/modules/backgrounds/citizenship/_citizenship.dm
@@ -1,7 +1,7 @@
 /decl/background_detail/citizenship
 	abstract_type = /decl/background_detail/citizenship
 	category = /decl/background_category/citizenship
-	var/ruling_body = "Other Faction"
+	var/ruling_body = "Other Faction" // currently unused? todo: display this or remove it? nothing seems to actually set it...
 	var/capital
 	var/size_heading = "Systems"
 	var/size_value
--- a/code/modules/projectiles/projectile.dm
+++ b/code/modules/projectiles/projectile.dm
@@ -78,9 +78,9 @@
 	var/fire_sound
 	var/fire_sound_vol = 50
 	var/fire_sound_vol_silenced = 10
-	var/miss_sounds
-	var/ricochet_sounds
-	var/list/impact_sounds	//for different categories, IMPACT_MEAT etc
+	var/miss_sounds // CURRENTLY UNUSED
+	var/ricochet_sounds // CURRENTLY UNUSED
+	var/list/impact_sounds	//for different categories, IMPACT_MEAT etc. CURRENTLY UNUSED
 	var/shrapnel_type = /obj/item/shard/shrapnel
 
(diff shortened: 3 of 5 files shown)


code/_onclick/hud/screen/screen_intent.dm
```diff
@@ -70,7 +70,7 @@
 	requires_ui_style    = FALSE
 	apply_screen_overlay = FALSE
 	var/intent_width     = 16
-	var/intent_height    = 16
+	var/intent_height    = 16 // currently unused
```
MistakeNot4892, line 73: This can be removed, it was for when intents were offset vertically but they are only ever offset horizontally now.

code/modules/projectiles/projectile.dm
```diff
@@ -78,9 +78,9 @@
 	var/fire_sound
 	var/fire_sound_vol = 50
 	var/fire_sound_vol_silenced = 10
-	var/miss_sounds
-	var/ricochet_sounds
-	var/list/impact_sounds	//for different categories, IMPACT_MEAT etc
+	var/miss_sounds // CURRENTLY UNUSED
+	var/ricochet_sounds // CURRENTLY UNUSED
+	var/list/impact_sounds	//for different categories, IMPACT_MEAT etc. CURRENTLY UNUSED
```
MistakeNot4892, line 83: This is weird, these should have been replaced by static getters I believe.

code/modules/salvage/structure.dm
```diff
@@ -4,7 +4,7 @@
 	tool_interaction_flags = TOOL_INTERACTION_DECONSTRUCT
 	material = /decl/material/solid/metal/steel
 	var/frame_type = /obj/machinery/constructable_frame/machine_frame
-	var/work_skill = SKILL_CONSTRUCTION
+	var/work_skill = SKILL_CONSTRUCTION // currently unused
```
MistakeNot4892, line 7: Unneeded since I made this use structure dismantling for dropping products, can be removed.

code/_onclick/hud/screen/screen_intent.dm
out-of-phaze: for some reason i was under the impression that this was for, like, robot/silicon UIs or something where they were laid out differently

code/modules/projectiles/projectile.dm
MistakeNot4892: My dev was out of date, these can all be removed in lieu of `get_ricochet_sounds()` and co.

code/modules/backgrounds/citizenship/_citizenship.dm
```diff
@@ -1,7 +1,7 @@
 /decl/background_detail/citizenship
 	abstract_type = /decl/background_detail/citizenship
 	category = /decl/background_category/citizenship
-	var/ruling_body = "Other Faction"
+	var/ruling_body = "Other Faction" // currently unused? todo: display this or remove it? nothing seems to actually set it...
```
MistakeNot4892, line 4: I think this might have been a partial port from the Polaris sync, it should be in the info box on chargen. I can do this if I remember.

MistakeNot4892 (changes requested):
Per comments some of these can just be removed.

code/_onclick/hud/screen/screen_intent.dm
MistakeNot4892: Not to my knowledge.

code/modules/backgrounds/citizenship/_citizenship.dm
out-of-phaze: noted, think that's the only one still standing then? EDIT: that and `tallness`

MistakeNot4892: image

Does not appear to be unused, seven year old code.

MistakeNot4892: Tallness is also in use.

MistakeNot4892: I'm stupid, there are multiple vars named that.

MistakeNot4892:
Remaining two are removed/cleaned up in #5378

PR #5342, opened 2026-04-09
Move fission power into its own module

out-of-phaze:
## Description of changes
Moves the fission core and associated content into its own module. _**Wait, since when did we have a fission reactor????????**_

Improves code quality by breaking up the `power` supermodule bit by bit for better organization.
No really since when did this thing even exist. Huh???


MistakeNot4892 (dismissed):

MistakeNot4892 (approved):
Conflict resolution

Issue #5285, opened 2026-02-04
low tech map content: wooden fences call for struts and struts are uncraftable in lowtech


sharah-za-fluffy:
#### Description of issue
this is an actual issue I know is a issue and not me having skill issues: wooden fences (railings) are uncraftable in ironscav (and presumably pyrelight) due to calling for struts, that cannot be crafted by low tech means

#### Difference between expected and actual behavior
I should not have to murder everything for bones to craft railings out of (for some reason bone does not require to be made into struts to be made into tables and railings but wood does because elves coded this apperently)

#### Steps to reproduce
chop down tree, make logs into planks, look fruitlessly for railing option, be told it calls for struts that cannot be found in hand crafting menus, be sad.

#### Specific information for locating

railings, struts

#### Length of time in which bug has been known to occur
who knows

#### Client version, Server revision & Game ID

Revision: ScavStation
Date
2026-02-03
Branch
main
Game ID
cDS-dtYb
Runtime Version
515.1643
Compiler Version
515.1643
Revision
777792621db4fba0a950245212a162d9f61e4c46

#### Issue bingo

- [x] Issue could be reproduced at least once
- [x] Issue could be reproduced by different players
- [x] Issue could be reproduced in multiple rounds
- [x] Issue happened in a recent (less than 7 days ago) round
- [x] [Couldn't find an existing issue about this](https://github.com/NebulaSS13/Nebula/issues)

MistakeNot4892:
Probably just need to cherrypick the strut removal PR from dev.

PR #5345, opened 2026-04-09
Removes struts on stable.

MistakeNot4892:
Struts are pretty busted and it's a low impact change, so redoing on stable to address #5285.


out-of-phaze (changes requested):
needs a map migration :(

tools/map_migrations/5345_strut_removal.txt
```diff
@@ -0,0 +1,2 @@
+/obj/item/stack/material/rods/ten : /obj/item/stack/material/rods/mapped/steel/ten{@OLD}
```
out-of-phaze, line 1: do we need to replace `/obj/item/stack/material/rods` with `/obj/item/stack/material/rods/mapped/steel`?

out-of-phaze (dismissed):

tools/map_migrations/5345_strut_removal.txt
MistakeNot4892: No, `/obj/item/stack/material/rods/cyborg` exists.

MistakeNot4892: I guess it wouldn't ever be mapped, though.

out-of-phaze: > No, `/obj/item/stack/material/rods/cyborg` exists.

it doesn't include subtypes unless you add `/@SUBTYPES`

out-of-phaze (approved):

PR #5347, opened 2026-04-11
Avoid queueing turfs for ambience during init

out-of-phaze:
## Description of changes
Instead of queueing turfs on SSambience during init, we loop over turfs in world instead.
Also, the SSambience queue is flushed during init now. I actually don't know why I added that given that nothing should be queueing during init...

In my local testing this provided a modest performance improvement, probably due to avoiding the creation of a huge list?? either way it also means ambience is set up prior to roundstart so there's no pop-in ambience issues where you start off in total darkness


out-of-phaze:
image

ambience appears to work fine still. chopping down that tree also updates the ambience around it, so post-init updates are fine too. daycycle updates also work

MistakeNot4892 (approved):

PR #5349, opened 2026-04-17
Open Turf Fix

l3lb0t:
## Description of changes
There's a bug where if you step onto a catwalk or ladder tile, to get back onto a normal floor tile you must pull yourself out, as if you were in a deep hole. This doesn't make a lot of sense, you would expect the catwalk to be roughly the same height as the floor. Turns out it's because you're technically standing on an open space tile whenever you stand on a catwalk or ladder tile.

You can now seamlessly walk through the ship without randomly being trapped on catwalks and ladders.

## Authorship
Me, myself, and Karl

## Changelog
:cl:
bugfix: You no longer need to physically pull yourself off of catwalks or ladder tiles.
/:cl:

### DIFF
--- a/code/game/turfs/turf.dm
+++ b/code/game/turfs/turf.dm
@@ -351,7 +351,7 @@
 			var/old_height     = old_turf.get_physical_height() + old_turf.reagents?.total_volume
 			var/current_height = get_physical_height() + reagents?.total_volume
 			if(abs(current_height - old_height) > FLUID_SHALLOW)
-				if(current_height > old_height)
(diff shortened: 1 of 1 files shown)


MistakeNot4892:
This was fixed in a slightly different way on `dev` - generally type checks like that aren't desirable as the height logic uses `is_open()` instead.

https://github.com/NebulaSS13/Nebula/pull/5308

MistakeNot4892 (changes requested):
Would prefer it if you could replicate the changes from the linked dev PR rather than adding the istype. Happy to cherrypick this myself if you'd prefer.

l3lb0t:
Done.

MistakeNot4892 (approved):
Much obliged!

PR #5353, opened 2026-05-06
Remove or implement defunct types

out-of-phaze:
## Description of changes
Removes a number of types that were unused both on maps or in code.
Fixes some incorrect/nonexistent organ typepaths.
Fixes oak floor tile recipes.
Fixes a typo in the suspension field generator description.
Adds typehints to some procs that are used with `foo().bar`-type expressions, which means that `bar` will be counted as used.

I feel kind of bad for removing the Runescape reference whip but it's unused and Nebula is supposed to be a template codebase. If it did anything neat I'd keep it like the KSBD reference items, but it's straight-up just a whip with a sprite and description referencing OSRS.

Fixes and code quality improvements.


code/modules/projectiles/guns/energy/lasertag.dm
```diff
@@ -7,21 +7,21 @@
 	origin_tech = @'{"combat":1,"magnets":2}'
 	self_recharge = 1
 	material = /decl/material/solid/metal/steel
-	projectile_type = /obj/item/projectile/beam/lastertag/blue
+	projectile_type = /obj/item/projectile/beam/lasertag/omni
```
MistakeNot4892, line 10: Blue to omni?

MistakeNot4892: Ah I see.

MistakeNot4892 (dismissed):

MistakeNot4892 (approved):

PR #5356, opened 2026-05-07
Fix turret controllers unable to find areas

Typhin:
## Description of changes

Turret control panels (`/obj/machinery/turretid`) were unable to find areas based on name, due to areas being initialized after the turret controller and thus not populating `global.areas` yet.
Turret control panel initialization has been moved to `LateInitialize()`, and an extra branch has been added to the search to be able to find by path.  Turret control panels can now have their area of control specified in three ways:
* If `control_area` is null, it will take control of the area in which it is located.
* If `control_area` is set to the name of an area (such as `"\improper AI Core"`), it will search for an area of the same name and take control of the first match found.
* If `control_area` is set to the path of an area (such as `/area/ministation/ai_core`), it will take control of that area.

ministation-0.dmm has been updated to fix the AI Core turret control panel, as it was incorrectly set to `"\improper AI Upload Chamber"`, and to clear the AI Upload Chamber turret control panel's `control_area` as it is already within the area and does not need to specify an area by name.  These two turret control panels now function as intended.

Previously, turret control panels were unable to find an area by name.  Now, they can.  This enables a turret control panel to take control of an area other than the one in which they reside.
Previously, turret control panels were unable to specify an area by path.  Now they can.  This enables a turret control panel to specify which area to control, should there be multiple areas with identical names for some reason.

## Authorship

Typhin

## Changelog
:cl:
add: Turret control panels can now specify an area in `control_area` by path.
bugfix: Turret control panels can now find the specified area name in `control_path` and take control of it.
/:cl:


maps/ministation/ministation-0.dmm
out-of-phaze: I'm not sure about using a string like this, it could very easily be brittle. I would suggest making it use the area path, lookup-by-name is generally not preferred anymore.

code/game/machinery/turret_control.dm
```diff
@@ -48,13 +49,19 @@
 	. = ..()
 
 /obj/machinery/turretid/Initialize()
+	. = ..()
+	return max(., INITIALIZE_HINT_LATELOAD) // Because areas initialize AFTER these!
```
out-of-phaze, line 53: you know, i was going to say this isn't necessary, but i guess we might make machinery return the qdel hint at some point, so it's technically fine? but mostly we don't bother with this and just unconditionally return the lateload hint in situations like this.

out-of-phaze (changes requested):

maps/ministation/ministation-0.dmm
Typhin: Lookup by name was the previous behavior and how existing turret control panels were attempting to use.  I don't mind removing that, but I wasn't going to assume its removal was desired.

out-of-phaze: you can leave support for it, maybe, but i meant more changing the map to use the typepath since you added the behavior

code/game/machinery/turret_control.dm
Typhin: I figured there was the potential for something to change down the line, some situation arising where the parent needed to return qdel, and in that case I didn't want this to make it so it discarded the qdel.  It was actually something that struck me at the last second as future-proofing.  If you're returning qdel, I'm assuming you have a good reason for it, I guess?

PR #5358, opened 2026-05-09
Fix turret control panels

Typhin:
## Description of changes

Fixes Turret Control Panels to initialize late, so it can search areas after they initialize and populate `global.areas`.
Previously, when trying to search by name, the search happened before areas ran their `Initialize()` and populated `global.areas`, so the search would fail.  Additionally, turret control panels can now specify an area by ID instead of by name, removing ambiguity and reducing maintenance.

Turret control panels can now specify their control area in three ways:
* If `control_area` is null, it will take control of the area in which it resides.
* If `control_area` is a text string, such as `"\improper AI Core"`, it will search through areas until it finds an area whose name matches and take control of it.
* If `control_area` is a path to an area, such as `/area/ministation/ai_core`, it will get that area and take control of it.
Additionally, the turret control panels in Ministation's AI Satellite have been updated.  The Upload Chamber is set to null because it is already within that area, and the AI Core has been set to `/area/ministation/ai_core` rather than rely on a name string.

## Authorship

Typhin

## Changelog
:cl:
add: Turret control panels can now specify their controlled area by path in addition to implicitly and by name.
bugfix: Turret control panels searching for an area by name works as intended now.
/:cl:

### DIFF
--- a/maps/ministation/ministation-0.dmm
+++ b/maps/ministation/ministation-0.dmm
@@ -10255,7 +10255,7 @@
 	dir = 1
 	},
 /obj/machinery/turretid/stun{
-	control_area = "\improper AI Upload Chamber";
+	control_area = /area/ministation/ai_core;
 	id_tag = "aihome";
 	pixel_y = -24;
 	dir = 1
@@ -13371,7 +13371,6 @@
 /obj/machinery/turretid/stun{
 	name = "AI Upload turret control";
 	id_tag = "upload";
-	control_area = "\improper AI Upload Chamber";
 	pixel_y = -24;
 	dir = 1
 	},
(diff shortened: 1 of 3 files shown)


maps/ministation/ministation-0.dmm
MistakeNot4892, line 13374: Might need to copy the control_area set from above here?

Typhin: Because this control is within the Upload Chamber area, leaving it null allows it to take control of the turrets properly.  I could have left it as-is or changed it to `/area/ministation/ai_upload` and it would also work.  I set it to null because I figured it would be less fragile.  If anyone were to change the map to make that room not part of the same area, they're more likely to think to check the control area than they are if they change the area name/path.

MistakeNot4892 (dismissed):

MistakeNot4892:
Additional changes would generally go in a separate PR (per Discord, just noting for the record), but this is tiny and the changes look fine, so I'll reapprove.

MistakeNot4892 (approved):

PR #5361, opened 2026-05-11
Convers hnoll and kobaloi ears and horns to correct accessory subtypes.

MistakeNot4892:
- Converts hnoll ears, kobaloi ears and kobaloi horns into their respective accessory types.
- Adds an 'extra' accessory metadata decl (used for tufts on Hnoll ears).
- Refactors accessory overlay gen to allow a list of metadata types to apply.
- Renames `Default Ears` and `Default Tail` to `No Custom Ears` and `No Custom Tail`.
- Fixes grafadreka spines masking.


mods/content/fantasy/datum/kobaloi/ears.dm
out-of-phaze, line 1: in theory this change means you can't have one floppy ear and one stubby ear, which i think you could before, but that's... probably fine...?

mods/content/fantasy/datum/kobaloi/horns.dm
out-of-phaze, line 1: same as with ears, this means you can't have one broken horn and one curved horn. that's probably fine, just noting the regression

out-of-phaze (approved):

mods/content/fantasy/datum/kobaloi/horns.dm
MistakeNot4892: Horns didn't get collapsed, you can pick multiple horns (like markings).

out-of-phaze: ahhh okay, so it's just ears then

mods/content/fantasy/datum/kobaloi/ears.dm
MistakeNot4892: Other species have ears as single units so I figured it was better to be consistent - I could add stub ears for the floppy variants too though. The alternative is keeping ears separate and letting multiple ears be selected, but that would need some work on the hnoll, tajaran and teshari sides as well.

out-of-phaze: maybe make a tracking issue for readding it, converting them is more important for now

PR #5362, opened 2026-05-12
Fix issues with traits in character setup

out-of-phaze:
## Description of changes
'Heritable' traits (everything but amputations and prostheses) are now applied in character setup during the mob snapshot step rather than after the snapshot is applied. This means downstream (Pyrelight) doesn't run into issues with the traits not being set. It also makes sense for at least some traits to be copied in the snapshot; this just allows some granular control over which traits are included.

TODO: May need to make get_mob_snapshot copy heritable traits, this was mostly just done as a fix for the issues and not a full-fledged feature.

Also fixes some correctness issues with character creation: the supplied mob snapshot is now threaded all the way through setup_human -> create_missing_organs, which should avoid a bit more organ/limb churn.

Also-also avoids duplicate `update_preview_mob()` calls in `/datum/preferences/Topic()` and allows bgstate changes to skip re-dressing the mob.

Also-also-also makes antlers have the right typepath. No migration needed because of UIDs. Yay.

Fixes wyrdlings on Pyrelight not getting their ears.

## Authorship
Me for the fix, Loaf for pointing out the issue.

### DIFF
--- a/code/datums/traits/_traits.dm
+++ b/code/datums/traits/_traits.dm
@@ -106,6 +106,8 @@
 	var/available_at_map_tech = MAP_TECH_LEVEL_ANY
 	/// Whether or not a rejuvenation should apply this aspect.
 	var/reapply_on_rejuvenation = FALSE
+	/// Whether this trait should be copied and applied by mob snapshots.
+	var/is_heritable = TRUE
 	/// What species can select this trait in chargen?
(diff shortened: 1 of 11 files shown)


MistakeNot4892 (approved):
Mob snapshots need some work around organ information and traits but this is fine as a fix per Disco chat

PR #5363, opened 2026-05-13
Minor hood fixes.

MistakeNot4892:
- Duplicate suit init/destroy removed.
- Suit hoods take suit material.
- Suit hoods are unremovable.


out-of-phaze (approved):
huh, i swear there wasn't a duplicate `hood = new hood(src)` when they had material added. i guess it worked in testing because the material one is a later-defined override? glad to see that sorted out

PR #5364, opened 2026-05-13
Adding an Interact hotkey.

MistakeNot4892:
- Adds an Interact hotkey that attempts to click on the turf you are facing.
- Adds `interaction_priority` to mark atoms that we are most interested in interacting with (so that lights above airlocks do not block interacting with the airlock).

### DIFF
--- a/code/modules/keybindings/binds/mob.dm
+++ b/code/modules/keybindings/binds/mob.dm
@@ -183,3 +183,30 @@
 /datum/keybinding/mob/minimal_hud/down(client/user)
 	user.mob.minimize_hud()
 	return TRUE
+
+/datum/keybinding/mob/interact
+	hotkey_keys = list("Enter")
+	name = "interact"
+	full_name = "Interact"
+	description = "Interact with the turf directly in front of you."
+
+/datum/keybinding/mob/interact/down(client/user)
+	user.mob.interact_with_facing()
+	return TRUE
+
+/mob/proc/interact_with_facing()
+
+	var/turf/facing = get_step(get_turf(src), dir)
+	if(!istype(facing) || !facing.simulated)
+		return
+
+	var/list/atoms = facing.get_contained_external_atoms()
+	// Move non-prioritised atoms to the end of the list (to avoid burning our hands on lights when trying to open a closet)
+	if(length(atoms) > 1)
+		atoms = sortTim(atoms, /proc/cmp_planelayer_interact_priority)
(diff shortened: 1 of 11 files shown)


code/modules/keybindings/binds/mob.dm
```diff
@@ -183,3 +183,42 @@
...
+
+	var/atom/click_on
+	if(length(facing.contents))
+		var/list/atoms = sortTim(facing.contents.Copy(), /proc/cmp_planelayer)
+
+		// Move non-prioritised atoms to the end of the list (to avoid burning our hands on lights when trying to open a closet)
+		for(var/atom/clickable as anything in atoms)
+			if(!clickable.interaction_priority)
+				atoms -= clickable
+				atoms += clickable
```
out-of-phaze, line 211: wonder if it might be faster to create a custom comparator that's cmp_planelayer but anything with `interaction_priority` is sorted lower. that'd lower the number of times we need to loop over the contents list at least. not blocking but i'd at least like to consider it

code/modules/keybindings/binds/mob.dm
```diff
@@ -183,3 +183,42 @@
...
+
+		// Try to click something.
+		for(var/atom/clickable as anything in atoms)
+			if(!clickable.simulated)
```
out-of-phaze, line 215: if we use `facing.get_contained_external_atoms()` instead of just `facing.contents` we can skip the simulated check and probably some other things we want to avoid here too

out-of-phaze (dismissed):
seems good except for some nitpicks

code/modules/keybindings/binds/mob.dm
MistakeNot4892: Something like this? Not completely confident it will sort as expected.
````
/proc/cmp_planelayer_interact_priority(atom/A, atom/B)
	if(A.interaction_priority != B.interaction_priority)
		return (B.interaction_priority - A.interaction_priority)
	return (B.plane - A.plane) || (B.layer - A.layer)
````

code/modules/keybindings/binds/mob.dm
MistakeNot4892: Simplified the block a fair bit.

code/modules/keybindings/binds/mob.dm
out-of-phaze: i think so? because that means anything with a higher interaction priority will be sorted later, and within the interaction priorities theyll be sorted by plane and layer.

out-of-phaze (approved):

PR #5368, opened 2026-05-13
ChangeArea() is now a /turf proc.

MistakeNot4892:
This is needed for overrides downstream. Should be equivalent in overhead, maybe microscopically slightly faster due to less params.


MistakeNot4892:
Not clear on what I have broken, reviewing this PR I can't spot any busted changes. :(

MistakeNot4892:
Never mind, it was an empty param on /turf/space

out-of-phaze (approved):

PR #5369, opened 2026-05-13
Background helpers are now usable on /mob not /mob/living.

MistakeNot4892:
Needed for downstream, but useful in general for making use of culture decls.


out-of-phaze (approved):
not totally sure why a non-`/mob/living` mob would ever have a background, but this is probably fine

PR #5371, opened 2026-05-13
Fences.

MistakeNot4892:
Rewrites and expands on fences.

- Fences now autosmooth based on neighboring fences.
- Fences use the structure lock system (hence can have keys, be picked, etc.).
- Fences can be climbed again.
- Fences can do four and three way junctions.
- Log, brick, stick and plank fences have been added (thank you @PlayerDeer for the icons that I butchered)
- Fences can be constructed.

# TODO
- [X] Brick fence icons.
- [X] Chain fence icons.
- [x] Palisade icons.
- [x] Plank fence icons.
- [x] Stick fence icons.
- [X] Convert locks to use structure logic.
- [X] General icon pass to make sure they all smooth correctly.
- [X] Testing.


code/game/objects/structures/__structure.dm
```diff
@@ -63,7 +63,7 @@
 	. = ..()
 	update_materials()
 	paint_verb ||= "painted" // fallback for the case of no material
-	if(lock && !istype(loc))
+	if(lock && !istype(lock))
```
out-of-phaze, line 66: oops

out-of-phaze (approved):

PR #5372, opened 2026-05-15
Adding channelled melee attacks.

MistakeNot4892:
Opening for visibility/diff.

## Description of changes
Adds melee charge attacks.

More involved melee system.

## TODO
- ~~Update clickdrag system to tick periodically instead of only on atom change (so that cursor state can be changed properly).~~ Split into separate PR.
- [ ] Replace temporary icons.
- [ ] Other stuff probably

## Authorship
Myself and Penny to date I believe. Will update this closer to merge.

## Changelog
Will update closer to merge.


MistakeNot4892:
image

Also now depends on #5373 

EDIT: 'solved' by having the mouseup proc set waitfor = FALSE. In any other language I would be worried about a thread safety violation since it's now diddling the SSclickdrag lists in parallel, but... should be fine I guess.

MistakeNot4892:
Depends on #5373.

PR #5373, opened 2026-05-15
Clickdrag subsystem and autofire enhancements.

MistakeNot4892:
## Description of changes
- Adds a subsystem to handle mobs clicking and dragging the mouse.
- Clients now track a var of the last atom they moused over on their mob.
- Generalises autofiring procs from guns to `/obj/item`.
- Full auto is now FULL AUTO, guns will fire ten times a second.
- Mounted plasma cutter is full auto now.

Makes click and drag more consistent and simplifies the relevant code.

## Authorship
Myself.

## Changelog
:cl:
tweak: Mech plasmacutter is full auto and mech AR full auto should now work.
tweak: Autofire should be more responsive in general.
/:cl:


MistakeNot4892:
Going to split some of these changes out.

MistakeNot4892:
Depends on #5375.

code/controllers/subsystems/clickdrag.dm
```diff
@@ -0,0 +1,53 @@
+SUBSYSTEM_DEF(clickdrag)
+	name = "Clickdrag"
+	wait = 1
```
out-of-phaze, line 3: should this be a ticker subsystem instead of just a wait 1 subsystem? given that this handles user input i feel that would be desirable

out-of-phaze (changes requested):

code/controllers/subsystems/clickdrag.dm
MistakeNot4892: I don't know how to do this. The ticker we have in code is a 1 second wait.

MistakeNot4892: Nvm I found SS_TICKER

code/controllers/subsystems/clickdrag.dm
```diff
@@ -0,0 +1,54 @@
...
+	flags = SS_TICKER
```
out-of-phaze: probably also needs SS_NO_INIT since there's no initialize override for it?

out-of-phaze (changes requested):
should be good to go after SS_NO_INIT is added?

code/controllers/subsystems/clickdrag.dm
MistakeNot4892: Donezo

out-of-phaze (approved):

PR #5379, opened 2026-05-17
Temporary commit to make girders craftable with metal sheets.

MistakeNot4892:


code/modules/crafting/stack_recipes/recipes_steel.dm
```diff
@@ -108,3 +108,10 @@
 
 /decl/stack_recipe/steel/furniture/drill_brace
 	result_type       = /obj/structure/drill_brace
+
+// Temporary recipe until dev merge.
+/decl/stack_recipe/steel/girder
+	result_type                 = /obj/structure/girder
+	required_wall_support_value = 10
+	req_amount                  = 5 * SHEET_MATERIAL_AMOUNT // Arbitrary value since girders return weird matter values.
```
out-of-phaze: dismantling them produces 2 sheets, did you always lose 3 sheets from dismantling?

out-of-phaze (dismissed):

code/modules/crafting/stack_recipes/recipes_steel.dm
MistakeNot4892: I copypasted from the rods recipe.

MistakeNot4892: Girders have 5 sheets of matter in them so lowering the recipe amount causes CI fails. Them giving 2 sheets on dismantle is due to using parts_amount I guess.

MistakeNot4892: With tweaked code, they drop 3 sheets and cost 6 to craft. I'm happy with that for the moment.

out-of-phaze (approved):

code/modules/crafting/stack_recipes/recipes_struts.dm
```diff
@@ -36,7 +36,6 @@
 /decl/stack_recipe/rods/girder
 	result_type                 = /obj/structure/girder
 	required_wall_support_value = 10
-	req_amount                  = 5 * SHEET_MATERIAL_AMOUNT // Arbitrary value since girders return weird matter values.
```
out-of-phaze, line 39: without this CI explodes :(

out-of-phaze (changes requested):

code/modules/crafting/stack_recipes/recipes_struts.dm
MistakeNot4892: I can't work out -why-, girders aren't do anything weird compared to any other structure.

MistakeNot4892: Issue seems to be null `parts_type`, I'm not certain why that breaks it but heigh ho.

out-of-phaze (approved):

PR #5387, opened 2026-06-15
Multiplanar ZM

Lohikar:
This generally reworks how ZM renders: depths now consist of multiple render slices, which are summed into 'zsum' objects for purposes of alpha masking. Adding new render slices is significantly easier than it used to be, and it is now possible to support things like darksight/ghostsight, emissives, and planar lighting effects in ZM. MZ checks have also generally been standardized around the ZM helpers instead of bespoke (slightly different) logic in a few places.

changes:
- ZM now renders atoms in groups of render slices, including a lighting render slice.
- Shadower objects are now masked using the atoms on their associated slice, allowing space to shine through at full brightness.
- Mimic planes are now managed by plane master objects for clarity and ease of definition.
- Z-AO rendering now uses explicitly planed overlays for layering.
- Elevator mimic is now reliable.
- Atoms becoming visible when not having been under a z-turf in the last 10 seconds now correctly glide.
- ZM has been given more debug logging, and the analyzer now checks more internal state (like z-stack integrity).
- ZM now has a concept of persistent Z-flags that survive ChangeTurf, used for ~~BOUNDARY and~~ OVERRIDE.
- ZM now ticks 4 times per tick check for improved efficiency.
- Mimics occluded by non-z turfs (but not yet destroyed) are now hidden from right-click.
- Opaque z-turfs can now opt into hiding below atoms from right-click.
- Examining a mimic now also shows how deep that atom is.
- ZM now validates that the specified ZM plane range is valid at compile-time.
- General correctness fixes.
- `ZM_MIMIC_OVERWRITE` has been renamed to `ZM_MIMIC_REPLACE` for clarity vs OVERRIDE.
- 'Midspan' AO (AO between the observed turf and the root turf) is now rendered.
- Large turfs that are below boundary ZM turfs are now rendered, except when the boundary is itself large.
- Synchronous overlay operations are now available.
- Synchronous AO operations are now available.
- Using Z-Mimic on `/turf/space` is now supported and performant.
- Shadowers are no longer used to dim the below level, and are now optional.
- The layering logic is now mostly free of magic numbers.

The way the plane master objects are being initialized here is not necessarily ideal -- it's largely a result of O7 just not really having a graceful way to do this. While I do want to keep the mimic master code as close as possible to the O7 version of it, I'm open to better approaches than CreatePlanesFor().

~~This PR is incomplete (but fully functional), further optimizations to shadowers are being tested.~~


code/modules/admin/view_variables/topic.dm
```diff
@@ -4,6 +4,8 @@
 		return
 	if(href_list["Vars"])
 		debug_variables(locate(href_list["Vars"]))
+	if(href_list["zm_analyze"])
+		analyze_openturf(locate(href_list["zm_analyze"]))
```
Lohikar, line 8: This feels ugly, open to better places to put it (though this is safe as-is, it already checks rights).

Lohikar:
Depends on #5393.

code/modules/mob/login.dm
```diff
@@ -90,6 +91,7 @@
 	RAISE_EVENT(/decl/observ/logged_in, src)
 
 	hud_reset(TRUE)
+	SSzcopy.CreatePlanesFor(client)
```
MistakeNot4892, line 94: A slightly more systematic approach to this might be nice, even it it's just a create_plane_masters() proc or such. Could move refresh_lightning_master() in there.

code/modules/admin/view_variables/topic.dm
MistakeNot4892: It's already pretty ugly, should be fine imo. At least the debug analysis stuff is somewhat centralised.

code/controllers/subsystems/overlays.dm
```diff
@@ -112,14 +118,12 @@ SUBSYSTEM_DEF(overlays)
...
-	if(islist(new_overlays))
-		var/list/new_overlays_list = new_overlays
-		if(null in new_overlays_list)
-			new_overlays_list -= new /list(length(new_overlays_list)) // Clears nulls from the list prior to appearancifying.
-		for(var/i in 1 to length(new_overlays_list))
-			var/image/cached_overlay = new_overlays_list[i]
-			APPEARANCEIFY(cached_overlay, new_overlays_list[i])
-		return new_overlays_list
+	if (islist(new_overlays))
+		new_overlays:RemoveAll(null)
```
MistakeNot4892: Need to cast to `/list` to placate the linter, which would also allow for using `.` instead of `:`.

MistakeNot4892:
Per discussion on Discord, going to pull this in for staging so it is tested alongside all the other changes going to stable.

PR #5388, opened 2026-06-27
Lighting: Corner Z-Bleed & Z-ambience rewrite

Lohikar:
I got tired of this logic breaking.

This version avoids piles of bespoke manual z-stack iteration with above/below references stored on the corner itself, built during corner init (and ChangeTurf). The stack will forcibly generate corners above/below it as required, but only the corner directly above/below rather than all corners on that master. It also supports delta updates for removing/adding connections.

Turf self-ambience init has also been moved: it now only initializes the turf it was directly applied to, with the general stack building logic handling propagation instead. This should avoid issues with double application of ambience during init.

### DIFF
--- a/code/__defines/lighting.dm
+++ b/code/__defines/lighting.dm
@@ -25,6 +25,11 @@
 #define TURF_IS_AMBIENT_LIT_UNSAFE(T) (T:ambient_active)
 #define TURF_IS_AMBIENT_LIT(T) (isturf(T) && TURF_IS_AMBIENT_LIT_UNSAFE(T))
 
+// The relation of these is important.
+#define LIGHTING_CORNER_GENERATE_UP    1
+#define LIGHTING_CORNER_GENERATE_BOTH  0
+#define LIGHTING_CORNER_GENERATE_DOWN -1
+
 // If I were you I'd leave this alone.
 #define LIGHTING_BASE_MATRIX \
 	list            \
--- a/code/__defines/lighting.dm
+++ b/code/__defines/lighting.dm
@@ -25,7 +25,7 @@
 #define TURF_IS_AMBIENT_LIT_UNSAFE(T) (T:ambient_active)
 #define TURF_IS_AMBIENT_LIT(T) (isturf(T) && TURF_IS_AMBIENT_LIT_UNSAFE(T))
 
-// The relation of these is important.
+// These are centered around zero to simplify logic; 'up' is 'is greater than -1', 'down' is 'is less than 0'. 0 matches both conditions.
 #define LIGHTING_CORNER_GENERATE_UP    1
 #define LIGHTING_CORNER_GENERATE_BOTH  0
 #define LIGHTING_CORNER_GENERATE_DOWN -1
--- a/code/modules/lighting/ambient_turf.dm
+++ b/code/modules/lighting/ambient_turf.dm
@@ -18,7 +18,11 @@
 	ambient_light = isnull(color) ? ambient_light : color
 	ambient_light_multiplier = isnull(multiplier) ? ambient_light_multiplier : multiplier
 
-	update_ambient_light()
+	// If we haven't initialized our corners yet, do that instead to avoid ambience double-init.
+	if (!corners || !lighting_corners_initialised)
+		generate_missing_corners()
+	else
+		update_ambient_light()
 
 /// Replace one ambient light with another. This is effectively a delta update, but it can be used to pretend that our one channel is doing color blending.
 /turf/proc/replace_ambient_light(old_color, new_color, old_multiplier, new_multiplier = 0)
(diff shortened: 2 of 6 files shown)


MistakeNot4892 (commented):
Gotta be honest I didn't review this at all beyond a look over for obvious errors. You're the SME for this section, happy to trust you tested.

code/__defines/lighting.dm
```diff
@@ -25,6 +25,11 @@
 #define TURF_IS_AMBIENT_LIT_UNSAFE(T) (T:ambient_active)
 #define TURF_IS_AMBIENT_LIT(T) (isturf(T) && TURF_IS_AMBIENT_LIT_UNSAFE(T))
 
+// The relation of these is important.
```
out-of-phaze: nitpick but i don't like this comment. what relation? how is it important? do you mean the ordering? do you mean that `-LIGHTING_CORNER_GENERATE_DOWN == LIGHTING_CORNER_GENERATE_UP` and vice versa? or do you mean that `LIGHTING_CORNER_GENERATE_UP + LIGHTING_CORNER_GENERATE_DOWN == `LIGHTING_CORNER_GENERATE_BOTH`? cryptic comments with rationales only understood by the original developer, if that, are the reason dirlights was so hard for me to understand at first.

code/modules/lighting/ambient_turf.dm
```diff
@@ -79,13 +83,12 @@
...
-		if (!corners || !lighting_corners_initialised)
-			generate_missing_corners()
+	if (TURF_IS_DYNAMICALLY_LIT_UNSAFE(src) && (!corners || !lighting_corners_initialised))
+		generate_missing_corners()
 
-		// This list can contain nulls on things like space turfs -- they only have their neighbors' corners.
-		for (var/datum/lighting_corner/C in corners)
-			C.update_ambient_lumcount(lr, lg, lb, !update)
+	// This list can contain nulls on things like space turfs -- they only have their neighbors' corners.
+	for (var/datum/lighting_corner/C in corners)
```
out-of-phaze, line 90: this now runs even if the turf is not dynamically lit. i guess that's correct, because it should work on space?

code/modules/lighting/lighting_corner.dm
```diff
@@ -171,7 +178,182 @@ var/global/list/REVERSE_LIGHTING_CORNER_DIAGONAL = list(0, 0, 0, 0, 3, 4, 0, 0,
...
+		This brick is responsible for finding the corner that's directly above us, and forcibly generating the corner if it doesn't exist yet.
+		It's just the same block of code repeated four times (for each master), plus the case of there now being no above corner, but previously having had one.
+		We also only initialize the one corner we need rather than all four since there's no benefit to initializing them all -- if a true light needs them, it'll make them itself.
+	*/
+	if      (t1 && (T = t1.above || GET_ABOVE(t1)) && (T.z_flags & ZM_ALLOW_LIGHTING) && TURF_IS_DYNAMICALLY_LIT_UNSAFE(T))
+		if (!(above_corner = T.corners?[t1i]) && GOING_UP)
+			if (!T.corners)
+				T.corners = new(4)
+			T.corners[t1i] = new/datum/lighting_corner(T, LIGHTING_CORNER_DIAGONAL[t1i], t1i, LIGHTING_CORNER_GENERATE_UP)
+			above_corner = T.corners[t1i]
```
out-of-phaze, line 226: ~~is it not possible for this to cause burn-in? i don't see where we remove `old_above_corner`'s contribution if we go from it to a different corner~~ oh it is fundamentally impossible for them to be different, lol, this is on the corner and not the turf. nvm

code/modules/lighting/lighting_source.dm
```diff
@@ -358,8 +358,8 @@
 				continue
 
 		Tcorners = T.corners
-		// These checks are inlined from generate_missing_corners. They must be kept in sync.
-		if (TURF_IS_DYNAMICALLY_LIT_UNSAFE(T) || T.light_source_solo || T.light_source_multi || (T.z_flags & ZM_ALLOW_LIGHTING))
```
out-of-phaze, line 362: i'll take your word for it these clauses are no longer necessary, i think they were just hacks anyway?

out-of-phaze (dismissed):
maybe expand that one comment to point to the relevant proc that explains it more in detail, but aside from that this looks ok

code/modules/lighting/ambient_turf.dm
Lohikar: It is intended that a static turf like space can cast ambient light onto a dynamic turf, they'll just only apply to the corners that the neighboring dynamic turf initialized itself.

code/modules/lighting/lighting_source.dm
Lohikar: I'm not actually sure why they were there in the first place, only TIDLU should matter given that this is just handling creating the dynamic overlay. A static turf with a light on it still can cast light onto adjacent dynamic turfs even if it isn't affecting itself.

code/__defines/lighting.dm
Lohikar: To be fair, I didn't really understand dirlights when I wrote it either. Comment does need fixing though.

code/modules/lighting/ambient_turf.dm
out-of-phaze: isnt there a comment somewhere that says the opposite, like "statically lit turfs can't ever allow lights through" or something? or is that onyl for non ambient lights?

code/modules/lighting/lighting_source.dm
out-of-phaze: it was something to do with flashlights in space, now that i think about it

code/modules/lighting/ambient_turf.dm
Lohikar: Statically lit turfs do not allow passage of light up or down, but they don't block transmission of light on the x/y axis due to DVIEW. A statically lit turf itself being a light source is separate, but intended.

code/modules/lighting/lighting_source.dm
Lohikar: Quick check says those are indeed busted, looking into it.

out-of-phaze (dismissed):

Lohikar:
As-is this breaks flashlights on space, fix is pending.

out-of-phaze (approved):

PR #5395, opened 2026-07-06
Resprites sand, converts it to a greyscale palette

triplezeta:
## Description of changes
Resprites sand and converts it to a greyscale palette

Nicer sprites good. Also possibility for strange colored sands for biome variety, in case that's someone's thing.
## Authorship

me, for the new sprites 
## Changelog
:cl:
imageadd: sand resprited and converted to greyscale
/:cl:


code/game/turfs/flooring/flooring_sand.dm
```diff
@@ -5,6 +5,7 @@
 	icon            = 'icons/turf/flooring/sand.dmi'
 	icon_base       = "sand"
 	icon_edge_layer = FLOOR_EDGE_SAND
+	color			= "#ffd076"
```
MistakeNot4892: I think this might prevent sand taking material colour from the turf. It might be better to set this color on the material and leave the flooring colour null, I think?

out-of-phaze:
```suggestion
	color			= null // autoset from material
```
as explained on discord, the default on `/decl/flooring` is COLOR_WHITE which means don't recolor, and this means the turf ends up with `color = null`. but setting it to null *on the flooring* means we want to auto-set it from the material color.

out-of-phaze (changes requested):

out-of-phaze (approved):

PR #5396, opened 2026-07-09
Mark a bunch of areas as abstract, skip abstract areas in area usage test

noelle-lavenza:
## Description of changes
Marks a bunch of abstract areas as abstract. Makes the area usage test skip abstract areas. Also makes it so only the base turbolift area is skipped, because the subtypes actually SHOULD be tested for.

Code cleanup, prevents use of abstract areas, also reshuffles vars so we actually take advantage of inheritance

## Authorship
Me

### DIFF
--- a/code/game/area/area_abstract.dm
+++ b/code/game/area/area_abstract.dm
@@ -1,9 +1,11 @@
 /area/hallway
+	abstract_type = /area/hallway
 	name = "hallway"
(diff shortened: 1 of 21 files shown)


MistakeNot4892 (approved):
Might be worth adding something to area/Initialize to runtime on creation of an abstract instance, if that isn't handled at /atom.

PR #5398, opened 2026-07-15
Fix gitignore not specifying repo root

noelle-lavenza:
the codex code folder was getting excluded. as far as i know, this is only needed for entries with `/` in it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,10 +6,10 @@
 *.lk
 *.backup
 *.before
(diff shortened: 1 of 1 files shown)


MistakeNot4892:
Shouldn't this go to `stable`?
EDIT: can't, change is not on stable

MistakeNot4892 (approved):

PR #5402, opened 2026-08-09
void suit accepts portable coolers into tank slot.

markoatonc:
## Description of changes

Added handling for adding portable cooler units to the voidsuit tank slot. It appears that a change of methods of handling suit storage about 10 years ago broke it and it has been broken since. It used to take flashlight and the cooler but now it can't. Added handling of attaching the cooling unit.
according to configuration still left in the file it should also take flashlight but I decided to leave that one as it was.

it could be helpful for synthetic characters. 
## Changelog
:cl:
add: Added  handling of adding portable suit cooler to the suit tank slot. 
/:cl:

### DIFF
--- a/code/modules/clothing/spacesuits/void/void.dm
+++ b/code/modules/clothing/spacesuits/void/void.dm
@@ -274,14 +274,14 @@ else if(##equipment_var) {\
 			playsound(loc, 'sound/items/Deconstruct.ogg', 50, 1)
 		return TRUE
 
-	if(istype(W,/obj/item/suit_cooling_unit))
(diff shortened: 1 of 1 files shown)


MistakeNot4892:
You have pointed this code at dev rather than stable. It does not compile on dev.

If you want it on scav, it needs to go to stable.

markoatonc:
> You have pointed this code at dev rather than stable. It does not compile on dev.
> 
> If you want it on scav, it needs to go to stable.

it works after this small tweak, but ok, I made separate PR for that.

MistakeNot4892:
Obviated by #5403

PR #5403, opened 2026-08-09
tanks commit for stable branch

markoatonc:


MistakeNot4892:
Your local branch was out of date when you authored the commit, and there is a conflict.

markoatonc:
https://github.com/NebulaSS13/Nebula/pull/5405 is the correct version because git broke my local branches.

PR #5407, opened 2026-08-13
Add helper for internal pressure difference checks

out-of-phaze:
## Description of changes
Replaces this duplicated pattern with a helper (actually three helpers, only one is used). I thought about using the get_internal_pressure_difference method in the two other helpers but it's small and this is atmos code so avoiding extra proc call overhead might actually be worth it.

Reduces code duplication.

### DIFF
--- a/code/game/objects/__objs.dm
+++ b/code/game/objects/__objs.dm
@@ -73,6 +73,23 @@
 /obj/return_air()
 	return loc?.return_air()
 
+/obj/proc/get_internal_pressure_difference()
+	var/datum/gas_mixture/int_air = return_air()
+	var/datum/gas_mixture/env_air = loc.return_air()
+	return int_air.return_pressure()-env_air.return_pressure()
+
+/// Return TRUE if the internal pressure difference is over `limit`.
+/obj/proc/check_internal_pressure_difference_over(limit)
+	var/datum/gas_mixture/int_air = return_air()
+	var/datum/gas_mixture/env_air = loc.return_air()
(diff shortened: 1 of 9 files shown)


code/game/objects/__objs.dm
```diff
@@ -73,6 +73,23 @@
...
+/obj/proc/get_internal_pressure_difference()
+	var/datum/gas_mixture/int_air = return_air()
+	var/datum/gas_mixture/env_air = loc.return_air()
+	return int_air.return_pressure()-env_air.return_pressure()
+
+/// Return TRUE if the internal pressure difference is over `limit`.
+/obj/proc/check_internal_pressure_difference_over(limit)
+	var/datum/gas_mixture/int_air = return_air()
+	var/datum/gas_mixture/env_air = loc.return_air()
+	return (int_air.return_pressure()-env_air.return_pressure()) > limit
```
MistakeNot4892, line 85: Couldn't these blocks just be `return get_internal_pressure_difference() > limit` and such

out-of-phaze: from the pr body
> I thought about using the get_internal_pressure_difference method in the two other helpers but it's small and this is atmos code so avoiding extra proc call overhead might actually be worth it.

fine with changing it though

MistakeNot4892: Making them macros would get both benefits I think, not really fussed either way though.

MistakeNot4892 (approved):

PR #5412, opened 2026-08-13
Reduce code duplication in sushi attackby

out-of-phaze:
## Description of changes
moves some of the duplicated sushi code into a proc that we reuse, so only one spot actually `new`s the sushi.

- [x] tested

better code and also leaves a comment to help with anyone else's confusion about raw chicken sushi


MistakeNot4892 (dismissed):

out-of-phaze:
the sushi also copies the offset from the item on the table. layering is a bit odd, not sure how to make it more intuitive, but this is better than it just defaulting to the middle

MistakeNot4892 (approved):

PR #5417, opened 2026-08-24
Adding supply packs to exploration modpack.

MistakeNot4892:
Grabbed from the Cynosure map branch - overlooked during phase guns port.


MistakeNot4892:
This is tiny so I'm taking off the frozen tag.

out-of-phaze (approved):

PR #5421, opened 2026-08-27
Add UIDs to a lot of decl types

out-of-phaze:
## Description of changes
Split out of #5420.
Adds UIDs to a lot of decl types.

This will let us replace some uses of `ref()`/`\ref\[]` and `locate()` with UIDs, which avoids string operation overhead and is overall more stable.

## Authorship
me


out-of-phaze:
depends on https://github.com/NebulaSS13/Nebula/pull/5425

PR #5423, opened 2026-08-27
Fix unmarked abstract decl types

out-of-phaze:
## Description of changes
some decl types were abstract but not marked as such, so they got instantiated when they didn't need to be or, worse, should not have been

avoid instantiating stuff that shouldn't be, and also make sure stuff doesn't break with unit test improvements. this is part 1, part 2 goes to dev

## Authorship
me

### DIFF
--- a/code/datums/ai_holo.dm
+++ b/code/datums/ai_holo.dm
@@ -3,6 +3,7 @@
 * bypass_colorize: if false, the hologram will be decolorized.
 */
 /decl/ai_holo
+	abstract_type = /decl/ai_holo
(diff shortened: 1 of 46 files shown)


MistakeNot4892 (dismissed):

out-of-phaze:
breaks curtains, need to wait on https://github.com/NebulaSS13/Nebula/pull/5425 to fix them properly

out-of-phaze:
dropped the curtain changes but otherwise its the same as what was approved 4 days ago, so i think this is probably fine to go in now

PR #5432, opened 2026-09-01
Readded unsim rock wall for natural maps.

MistakeNot4892:
The icon for this still existed, just lost the definition at some point. Useful for contained maps with rock walls.


MistakeNot4892:
Tiny PR, no reviews at present, going ahead and selfmerging.

PR #5434, opened 2026-09-01
Added special role checking to aggressive AI for merc viscerator.

MistakeNot4892:
- Adds a special role check to aggressive AI (overrides faction checking).
- Renames manhack props to viscerators as manhack is the name from Half-Life and was being used inconsistently.
- Adds an IFF-enabled viscerator spawner grenade for mercs.


MistakeNot4892:
Small PR, tested fairly extensively on vox branch. Going to self-merge.

PR #5436, opened 2026-09-02
Porting/reimplementing Polaris chainsaw.

MistakeNot4892:
## Description of changes
- Repaths welding tools to `/obj/item/fuelled_tool/welding`
- Repaths welding tanks to `/obj/item/chem/fuel_tank`
- Repaths welding packs to  `/obj/item/fuelled_tool/welding/pack`.
- Adds a chainsaw - high damage and ability to destroy obstacles while fuelled and wielded. Adminspawn only for the moment.

Chainsaw idle sound cut from https://freesound.org/people/Stefan021/sounds/431741/

Splits the base fuel tank and fuel behavior away from also working as a welding tool. 
Adds a neat, powerful item.

## Authorship
Myself.

## Changelog
Nothing player-facing.


MistakeNot4892:
IDB seems a bit cooked, CI is passing otherwise.

code/game/objects/items/tools/fuelled/fuel_tank.dm
```diff
@@ -1,10 +1,10 @@
 //////////////////////////////////////////////////////////////////
-// Welding tool tanks
+// Tool fuel tanks
 //////////////////////////////////////////////////////////////////
-/obj/item/chems/welder_tank
+/obj/item/chems/fuel_tank
 	name               = "welding tank"
 	base_name          = "welding tank"
```
out-of-phaze: these weren't changed even though the description and path were?

code/game/turfs/turf.dm
```diff
@@ -976,3 +976,6 @@
 			if(istype(thing) && !QDELETED(thing) && !thing.is_safe_to_step(stepper))
 				return FALSE
 	return TRUE
+
+/turf/get_object_size()
+	return ITEM_SIZE_LARGE_STRUCTURE
```
out-of-phaze, line 981: that's interesting

out-of-phaze (approved):
I think this could be split into two PRs but it's probably fine as is? I don't know, a little opposed to `/obj/item/fuelled_tool`, I feel like it will go the way of `/obj/item/weapon/device` down the line and end up being moved onto its own `/datum/extension/fuelled_tool` or something.

code/game/turfs/turf.dm
MistakeNot4892: Oh I forgot I did that, I wanted walls to take awhile to cut through. I don't know if it will have impacts anywhere else, most stuff checking `get_object_size()` is behind typechecks and such anyway.

code/game/objects/items/tools/fuelled/fuel_tank.dm
MistakeNot4892: I probably changed them in one of the later commits and commit --amend'd them like a fool.

MistakeNot4892:
````I think this could be split into two PRs but it's probably fine as is? I don't know, a little opposed to /obj/item/fuelled_tool, I feel like it will go the way of /obj/item/weapon/device down the line and end up being moved onto its own /datum/extension/fuelled_tool or something.````

That would be fine, just not something I wanted to do in this PR. I would personally make it a set of helpers on `/datum/extension/tool` which would be a fair bit more involved than this PR.

As to being two PRs - I could make the actual chainsaw a separate PR if you want, but I'm not sure that one extra file benefits much from being split into a second PR.

out-of-phaze (approved):
I mean, I already reviewed it so no point in splitting it now.

PR #5437, opened 2026-09-02
Adds a merc intercom and ability for radios to decrypt without keys.

MistakeNot4892:


MistakeNot4892:
Tiny PR, minimal impact. Selfmerging after CI.

PR #5438, opened 2026-09-02
Generalising/expanding on crystal structures.

MistakeNot4892:
- Ports and greyscales crystal icons from Polaris.
- Adds basic crystal structure, ice crystals and magma crystals.
- Lightly reworks xenocrystals.


code/modules/tools/subtypes/hammers.dm
```diff
@@ -3,12 +3,12 @@
 	desc                = "A simple hammer. Ancient technology once thought lost."
 	icon                = 'icons/obj/items/tool/hammers/hammer.dmi'
 	attack_verb         = list(
-		"bludgeons",
-		"slaps",
-		"beats",
-		"strikes",
-		"bashes",
-		"hammers"
```
out-of-phaze, line 11: i know it's inconsistent with the rest of the game atm but iirc at some point i was going to change all the attack verbs to present tense plural, e.g. `bludgeon`, `slap`, etc., for ease of use with the grammar helpers i added. bleh, it can be done later though

code/game/objects/structures/crystal.dm
```diff
@@ -0,0 +1,89 @@
...
+	. = ..()
+	if(base_state)
+		icon_state = base_state
+	if(has_shadow)
+		add_overlay(overlay_image(icon, "[icon_state]_shadow", COLOR_WHITE, RESET_COLOR))
+
+/obj/structure/crystal/create_dismantled_products(var/turf/T)
+	SHOULD_CALL_PARENT(FALSE)
+	visible_message(SPAN_DANGER("\The [src] shatters!"))
+	. = list(new /obj/item/shard(T, material.type))
```
out-of-phaze: originally we got 3 borosilicate and 3 glass shards (which would probably be the crystal material now), now we only get 3. that's fine i just wanted to note the change. i was honestly assuming the xeno crystals were more like crystals with borosilicate running through it to explain that

out-of-phaze (dismissed):

code/game/objects/structures/crystal.dm
MistakeNot4892: I can bump it up or add reinf_material maybe to drop additional shards.

out-of-phaze: It's an okay change, but maybe it should be based on matter or something? That'd be kind of a pain but we do have helpers, just maybe add the ability for a multiplier so we don't get perfectly-efficient crystal material extraction from breaking it?

code/modules/tools/subtypes/hammers.dm
MistakeNot4892: Yeah I was more interested in consistency with the other attack strings for this PR, it popped up when I was testing smashing crystals with a hammer.

code/game/objects/structures/crystal.dm
MistakeNot4892: It's now based on matter.

code/__defines/maths.dm
```diff
@@ -26,3 +26,5 @@
 
 // Linear interpolation between two values
 #define LERP(a, b, amount) ( amount ? ((a) + ((b) - (a)) * (amount)) : a )
+
+#define RAND(BASE, DOMAIN) (BASE + (rand() * DOMAIN))
```
out-of-phaze: This feels a bit hard to understand intuitively compared to the existing `RAND_F(LOW, HIGH)` macro.. which I forgot to mention because I forgot about it until now, lol

out-of-phaze (dismissed):

code/__defines/maths.dm
MistakeNot4892: Replaced with RAND_F

out-of-phaze (approved):

PR #5439, opened 2026-09-03
Fix holsters breaking alt-interactions on clothing

out-of-phaze:
## Description of changes
`get_alt_interactions()` should not use `GET_DECL()` on any of its entries, the doc comment has been updated to reflect this

fixes a runtime where a value returned from `get_alt_interactions()` would be null after `GET_DECL()` was run on it


MistakeNot4892 (approved):
I'll do the RESOLVE_TO_DECL tweak on dev

PR #5440, opened 2026-09-03
Loot piles now handle /obj/random without runtimes.

MistakeNot4892:
Fixes a runtime when /obj/random is used in loot lists.


out-of-phaze (dismissed):
i would like to revisit this and do it a different way on dev down the line, but for now this will do

out-of-phaze (approved):
awesome

PR #5442, opened 2026-09-03
Interaction retrieval is slightly more robust.

MistakeNot4892:
Should allow a little more leeway with things like accidentally using GET_DECL() in an interaction proc.


MistakeNot4892:
Tiny PR, minimal impact. Self-merging to get it off the list.

PR #5443, opened 2026-09-03
Porting giant spider castes from Polaris.

MistakeNot4892:
## Description of changes
Various new mutant spiders with associated abilities and AI ported from Polaris.

More variety in spider enemies.
Enhancements to enemy AI.

## Authorship
I believe @Neerti wrote most of the fancy spider logic on Polaris.

## Changelog
:cl:
add: Added more spiders from Polaris.
/:cl:


MistakeNot4892:
Definitely going to break this up into several PRs before merge.

MistakeNot4892:
Depends on #5445 #5448 #5449 and #5447

MistakeNot4892:
I may still break the AI stuff (ambusher and social distancing) into a separate PR.

mods/mobs/spiders/mobs/_giant_spider_ranged.dm
```diff
@@ -0,0 +1,24 @@
+/mob/living/simple_animal/hostile/giant_spider/ranged
+	abstract_type = /mob/living/simple_animal/hostile/giant_spider/ranged
+	var/ranged_charges = 10
+	var/max_ranged_charges = 15
+	var/ranged_charge_regen_prob = 20
```
out-of-phaze, line 5: i feel kind of weird using per-tick probabilities like this, i sort of feel like we should be passing delta-time to handle_regular_status_updates and doing some math to set up an effective/expected regen rate. but this isn't a request for changes, just sort of a "thing we should maybe do going forward"

mods/mobs/spiders/mobs/ai_tunneller.dm
```diff
@@ -0,0 +1,97 @@
...
+	var/tunnel_max_range  = 6
+	var/tunnel_warning    = 0.5 SECONDS	// How long the dig telegraphing is.
+	var/tunnel_tile_speed = 2			// How long to wait between each tile. Higher numbers result in an easier to dodge tunnel attack.
+
+/decl/maneuver/tunnel/perform(var/mob/living/user, var/atom/target, var/strength, var/reflexively = FALSE)
+	if(!(. = ..()) || !target)
+		return
+
+	user.do_windup_animation(target, windup_time = tunnel_warning)
+	sleep(tunnel_warning)
```
out-of-phaze: i would probably make everything after this sleep a `finish_tunneling()` proc that we use a timer for (as lohikar said, sleeping procs don't get interrupted/cancelled by deletion). but it's not urgent, it's probably fine. maybe need to add a `if(QDELETED(user) || QDELETED(target))` guard after the sleep though?

mods/mobs/spiders/mobs/pepper.dm
```diff
@@ -0,0 +1,12 @@
...
+	icon = 'mods/mobs/spiders/icons/spider_pepper.dmi'
+	eye_color = "#ff0000"
+	max_health = 210
+	poison_chance = 20
+	poison_per_bite = 5
+	poison_type = /decl/material/liquid/capsaicin/condensed
+
+/mob/living/simple_animal/hostile/giant_spider/pepper/Initialize(mapload, atom/parent)
+	. = ..()
+	set_scale(1.1)
```
out-of-phaze:
```suggestion
	icon_scale_x = 1.1
	icon_scale_y = 1.1
```
this is how hivebots do it

mods/mobs/spiders/mobs/volatile.dm
```diff
@@ -0,0 +1,52 @@
...
+	var/explosion_heavy_range	= 2
+	var/explosion_light_range	= 4
+	var/explosion_flash_range	= 6
+
+	var/explosion_delay_lower	= 1 SECOND	// Lower bound for explosion delay.
+	var/explosion_delay_upper	= 2 SECONDS	// Upper bound.
+
+/mob/living/simple_animal/hostile/giant_spider/volatile/Initialize(mapload, atom/parent)
+	. = ..()
+	set_scale(1.25)
```
out-of-phaze: this is how hivebots do it
```suggestion
	icon_scale_x = 1.1
	icon_scale_y = 1.1

	var/exploded = FALSE
	var/explosion_dev_range		= 1
	var/explosion_heavy_range	= 2
	var/explosion_light_range	= 4
	var/explosion_flash_range	= 6

	var/explosion_delay_lower	= 1 SECOND	// Lower bound for explosion delay.
	var/explosion_delay_upper	= 2 SECONDS	// Upper bound.
```

mods/mobs/spiders/mobs/volatile.dm
```diff
@@ -0,0 +1,52 @@
...
+	set waitfor = FALSE
+
+	// Flash black and red as a warning.
+	var/delay = rand(explosion_delay_lower, explosion_delay_upper)
+	for(var/i = 1 to delay)
+		if(i % 2 == 0)
+			color = "#000000"
+		else
+			color = "#ff0000"
+		sleep(1)
```
out-of-phaze, line 47: Noooooooooooo. We use animate() for this I'm pretty sure? This should be equivalent but I haven't tested. Jump easing so it flashes instead of smoothly animating the color, but if you *want* it to be smooth you can change the easing. loops for delay/2 because it's a 2ds sequence and we want it to run for `delay` deciseconds. set to parallel just in case i guess

```suggestion
	// Flash black and red as a warning.
	var/delay = rand(explosion_delay_lower, explosion_delay_upper)
	animate(src, color = "#000000", time = 1, loop = delay/2, flags = ANIMATION_PARALLEL, easing = JUMP_EASING)
	animate(color = "#ff0000", time = 1, easing = JUMP_EASING)
	sleep(delay)
	color = initial(color)
	animate(src) // end animation if it's still ongoing
```

out-of-phaze (dismissed):

mods/mobs/spiders/mobs/_giant_spider_ranged.dm
MistakeNot4892: I'm going to generalize this simple ammo system to simple_animal in the future and change it to a time-based regen then.

mods/mobs/spiders/mobs/volatile.dm
MistakeNot4892: This does not loop, it goes smoothly from black to red once.

out-of-phaze: The new code? It should loop `delay/2` times and last `delay` deciseconds. It also shouldn't be hanging smoothly because it uses jump easing...

MistakeNot4892: Well, I copied it in verbatim and it didn't, so. Not sure how I could have messed it up as I just replaced the current sleep-based block.

out-of-phaze: it seems to be because of the cloaking code animating mobs every tick

MistakeNot4892: Fix merged and PR updated

MistakeNot4892:
As review was addressed and animation fix was put in place I'm going to selfmerge this.

PR #5446, opened 2026-09-03
SSInput Improvement: ANY is the work of the devil

francinum:
## Description of changes

SSInput's macro set handling has been completely rewritten. The macrosets are now built per-player to allow avoiding the use of the `ANY` special macro, which breaks the ability to copy-paste out of `output` elements (like the chat). As well as playing havoc with the practically-undocumented ability for `command` elements to be prioritized for text input, assuming a compatible macro set is applied.

Focus Chat control mode is now actually usable.

## Authorship

This code is inspired by the original implementation of SSInput's keybinding systems, but has been practically completely rewritten at this point.

## Changelog
:cl:
refactor: The way hotkeys are bound has been rewritten completely.
add: Hotkey state is now saved to your preferences.
tweak: Default hotkeys for 'Focus Chat'/'Classic' mode have been audited.
/:cl:


francinum:
Still need to do:
Warning nag preference (And spans, They're directly imported from Daedalus and don't seem to exist in the old stylesheet)

code/modules/keybindings/setup.dm
```diff
@@ -14,30 +14,85 @@
...
+	for(var/k in 1 to length(m_macro_set))
+		var/key = m_macro_set[k]
+		var/command = m_macro_set[key]
+		winset(src, "modifier-\ref[key]", "parent=default;name=[key];command=[command]")
+	var/list/printables = list()
 
-	if(prefs?.hotkeys)
-		winset(src, null, "outputwindow.input.focus=true input.background-color=[COLOR_INPUT_ENABLED]")
+	for(var/key in personal_macro_set) //We don't care about the bound key, just the key itself
+		var/keycode = replacetext(key, regex("(Alt|Shift|Ctrl)", "g"), "")
```
MistakeNot4892: Would it be faster to compile this regex as a static reference somewhere?

Lohikar: It should at least be pulled out of the loop.

francinum: Pulled both regexes out of their loop.

code/_helpers/global_lists.dm
```diff
@@ -50,6 +50,7 @@ var/global/list/child_stance_limbs = list(
 
 // TODO: Replace keybinding datums with keybinding decls to make this unnecessary.
 var/global/list/hotkey_keybinding_list_by_key = list() // Replace this with just looping over all keybinding decls (as below) in a 'reset hotkeys' proc.
+var/global/list/hotkey_keybinding_list_by_key_fc = list()
```
out-of-phaze: `fc`? "for classic"? can we rename this or add a comment explaining it
EDIT: `focus chat`, which is a better term for what we used to call ~~"hotkey mode."~~ *non*-hotkey mode, okay. i would like this to be renamed `hotkey_keybinding_list_by_key_focused` or have a doc comment explaining that it's for focus chat

code/modules/client/preference_setup/controls/01_keybindings.dm
```diff
@@ -1,6 +1,8 @@
 /datum/preferences
 	/// Whether or not this client has standard hotkeys enabled
 	var/hotkeys = TRUE
+	/// Focus Chat masked-out hotkey nag.
+	var/fc_hotkey_nag = TRUE
```
out-of-phaze: we aren't exactly billed by the character, we could probably call it `focus_chat_hotkey_nag` or `focus_chat_nag`. since there's a doc comment this one doesn't *need* to be changed though

out-of-phaze (changes requested):
partial review

MistakeNot4892 (approved):
Willing to give this a shot considering any issues will likely be exposed by playtesting.

MistakeNot4892:
Going to merge this in as review was addressed and I want to see how it works during staging testing.

PR #5448, opened 2026-09-03
Allowed simple_animals to specify damage types that they are immune to.

MistakeNot4892:


MistakeNot4892:
Tiny PR with minimal impact, going to self-merge it when CI has run.

PR #5449, opened 2026-09-03
Implements simple_animal cloaking and integrates it into aggressive AI.

MistakeNot4892:
- Simple mobs can be given the ability to toggle an alpha-based cloaking effect.
- Mob controllers can have `ambusher` or `socially_distancing` flags set to keep a minimum distance from their target (unless cloaked, in the case of ambushers).

This standalone PR is difficult to test due to nothing using the new flags, but I have tested it in #5443 fairly extensively.


code/datums/ai/aggressive.dm
```diff
@@ -3,7 +3,10 @@
 	stop_wander_when_pulled = FALSE
 	try_destroy_surroundings = TRUE
 	target_scan_distance = 10
+
 	var/decl/special_role/friendly_to_role = null
+	var/socially_distancing = FALSE // If set, will try to stay 3 tiles away from the target.
```
out-of-phaze: feels like this could easily be a num rather than a bool, to specify the distance

out-of-phaze (dismissed):

out-of-phaze (approved):

PR #5452, opened 2026-09-04
Fixes from Cynosure playtesting.

MistakeNot4892:
- Fixes inverted open/close state for fence gates.
- Fixes inverted density checking for some fence logic.
- Fixes fence cutting.
- Fixes fences not blending with walls.
- Fixes level data ceilings replacing dirt paths with plating.
- Fixes headlamp lighting overlays being offset twice on nonstandard humanmobs.
- Fixes cloak overlays/underlays being offset twice on nonstandard humanmobs.
- Fixes pizzas being banished to the Shadow Realm if stacked and thrown.


MistakeNot4892:
Pointed this to staging as it's a fix PR so will go back to dev eventually anyway.

PR #5453, opened 2026-09-05
Ports radiation emitter effect from Polaris.

MistakeNot4892:
Useful for map hazards.


MistakeNot4892:
Minimal impact small PR. Self-merging.

PR #5455, opened 2026-09-05
Ported various remains from Polaris.

MistakeNot4892:
Useful for mapping variety. Also organises the icons a bit better.


MistakeNot4892:
This PR does not touch any logic systems or add the new spawners to any maps, going to self-merge it.

PR #5458, opened 2026-09-05
Added a blue parka preset.

MistakeNot4892:


MistakeNot4892:
Tiny PR, selfmerging.

PR #5463, opened 2026-09-05
Clothing markings can now be emissive.

MistakeNot4892:
Clothing markings can now be made emissive, for things like vox armour running lights.


MistakeNot4892:
Tiny PR tested extensively on vox branch, self-merging.

PR #5464, opened 2026-09-05
Reorder outfit equip procs to avoid voidsuits blocking other gear.

MistakeNot4892:
Fixes voidsuits with attached helmets blocking mask equip due to the helmet deploying.


MistakeNot4892:
Tiny PR, reviews not available, selfmerging.

PR #5468, opened 2026-09-06
Ported blob projectile from Polaris.

MistakeNot4892:
- Ports a reagent cloud projectile type from Polaris.
- Projectiles can apply mob modifiers.
- Mob modifiers can impact click cooldowns and movement speed.


code/modules/mob/mob.dm
```diff
@@ -316,6 +317,9 @@ var/global/const/ACTION_DANGER_ALL = 2
 		. += move_intent.move_delay
 	else
 		. += _automove_delay
+	for(var/modifier_type in get_mob_modifiers())
+		var/decl/mob_modifier/modifier = RESOLVE_TO_DECL(modifier_type)
+		. += modifier.movement_slowdown
```
out-of-phaze: it may be better to eventually have this read a cached slowdown var that's modified when mob modifiers change, but for now this is fine

code/modules/mob_modifiers/modifiers_helpers.dm
```diff
@@ -68,11 +68,17 @@
 			modifier = existing_modifier
 			break
 
+	var/already_present = TRUE
 	if(!istype(modifier))
 		modifier = new archetype.modifier_type(archetype, src, source)
+		already_present = FALSE
```
out-of-phaze:
```suggestion
	var/already_present = istype(modifier)
	if(!already_present)
		modifier = new archetype.modifier_type(archetype, src, source)
```
not necessary but IMO this is easier to parse at a glance

code/modules/projectiles/projectile/blob.dm
```diff
@@ -0,0 +1,35 @@
+/obj/item/projectile/energy/blob //Not super strong.
```
out-of-phaze: curious why this descends from `/energy`? is that a holdover from polaris or is it needed here for functionality rather than stat/etc inheritance?

out-of-phaze (dismissed):
would like to hear back about the `/obj/item/projectile/energy` descendance but aside from that this is fine

code/modules/projectiles/projectile/blob.dm
MistakeNot4892: No specific reason other than Polaris using it. I guess SFX and basic stats.

out-of-phaze (dismissed):

code/modules/mob/mob.dm
```diff
@@ -316,8 +317,13 @@ var/global/const/ACTION_DANGER_ALL = 2
 		. += move_intent.move_delay
 	else
 		. += _automove_delay
-	. = max(. + (ENCUMBERANCE_MOVEMENT_MOD * encumbrance()), 1)
 
+	if(isnull(modifier_movement_slowdown))
+		modifier_movement_slowdown = 0
+		for(var/modifier_type in get_mob_modifiers())
+			var/decl/mob_modifier/modifier = RESOLVE_TO_DECL(modifier_type)
+			modifier_movement_slowdown += modifier.movement_slowdown
```
out-of-phaze, line 325: don't we already run code when modifiers are added and removed? would it not be preferable to just adjust the cached values there?

out-of-phaze (changes requested):
aside from this i thhhhink it's good to go

code/modules/mob/mob.dm
MistakeNot4892: Two reasons to do it here that I can see 
- You can VV the mob and null the cache to do a full rebuild independently of the modifier decls
- Hypothetically having several movement modifiers applied at once will rebuild the cached value repeatedly whereas this one will null it and then it will only rebuild when the mob next calls get_movement_delay()

out-of-phaze (approved):

PR #5469, opened 2026-09-06
Ported perch from Polaris.

MistakeNot4892:
Missed a fish in the original fish port PR.


MistakeNot4892:
Low-impact fish, reviews unavailable, self-merging.

PR #5472, opened 2026-09-06
Fix issues with ambient light on staging

out-of-phaze:
## Description of changes
ambient light set prior to corner init wasn't working properly because corners checked `ambient_active`, but that wasn't set. i don't know how to make it properly set in the right order and lohikar said a proper fix will take a while, so here's a stopgap

fixes issues with ambient light on staging, avoids issues with the partial revert of non-queueing ambience in #5470

## Authorship
some of this stuff is adapted from lohi talking about fixes on O7 but the code here was written by me


MistakeNot4892 (approved):

Lohikar (approved):
Yeah, sorry about the jank, some of that looks like it might be merge artifacts from trying to replay Neb-local changes onto O7 ambience during updates.

PR #5473, opened 2026-09-07
Staging fixes, Drake Edition

MistakeNot4892:
Fixes various oversights and problems noted during a playtest as a drake on the Shaded Hills map. Commit list should be atomic/exhaustive for the most part though I know a few of the commits are mixed up/mangled in terms of actual changeset.


MistakeNot4892:
In the absence of anyone to do timely reviews due to life stuff and whatnot, I'm going to do a self-review and self-merge, like an evil villain.

MistakeNot4892:
A fair chunk of this arguably should go to dev, but as it's smallish QOL stuff I'm going to abuse my merge powers and just push it. Heigh ho.

PR #5485, opened 2026-09-08
AI mobs can be tethered to a home atom.

MistakeNot4892:
Useful for mapping random mob spawners into POIs so they don't sprint off and get lost in the woods immediately.


code/datums/ai/_ai.dm
```diff
@@ -126,15 +126,21 @@
...
-			// Recheck in case we walked into lava or something during wandering.
-			return get_stance() != STANCE_BUSY && !QDELETED(body) && !QDELETED(src)
-		return TRUE
-	return FALSE
+
+	if(QDELETED(body) || QDELETED(src) || get_stance() == STANCE_BUSY)
+		return FALSE
+
+	if(!isnull(home) && get_dist(body, home) > home_wander_distance)
+		body.start_automove(home)
```
out-of-phaze: in theory we don't have to automove all the way to the home turf, we just have to move enough to be comfortably inside the wander distance. could maybe do this by randomly setting an automove_metadata datum's `_acceptable_distance` to `rand(0, home_wander_distance)` or something. not blocking though

code/game/objects/random/subtypes/mobs.dm
```diff
@@ -60,6 +60,9 @@
 	icon_state = "standing"
 	spawn_nothing_percentage = 80
 
+	mob_returns_home = 1
```
out-of-phaze:
```suggestion
	mob_returns_home = TRUE
```

code/datums/ai/_ai.dm
```diff
@@ -126,15 +126,21 @@
...
-			try_bark()
-			// Recheck in case we walked into lava or something during wandering.
-			return get_stance() != STANCE_BUSY && !QDELETED(body) && !QDELETED(src)
-		return TRUE
-	return FALSE
+
+	if(QDELETED(body) || QDELETED(src) || get_stance() == STANCE_BUSY)
+		return FALSE
+
+	if(!isnull(home) && get_dist(body, home) > home_wander_distance)
```
out-of-phaze, line 133: this might make dragging a mob outside of their home radius + getting them stuck there an issue? since this might make them stop doing combat and trying to move instead. unsure though, maybe they can automove and fight

out-of-phaze (dismissed):

code/datums/ai/_ai.dm
MistakeNot4892: They do try to run back to their home radius in the middle of combat, but the alternative would be letting people kite hostile mobs indefinitely which I think kinda defeats the purpose of having a home tether.

code/game/objects/random/subtypes/mobs.dm
MistakeNot4892: Actually meant to delete that after testing.

code/datums/ai/_ai.dm
MistakeNot4892: Don't even need to be comfortably inside, really, just need to be at the border. Using automove metadata was a good thought, slapped that in there.

out-of-phaze (approved):

PR #5493, opened 2026-09-09
Adds a mob_mimic simple_animal type for complex mob NPCs.

MistakeNot4892:
## Description of changes
- Adds `/mob/living/simple_animal/mob_mimic` which caches appearance, weapons, etc. from a more complex mob and applies them to spawns.
- Adds NPC grafadreka.
- Adds NPC utility frame variants.
- Adds NPC autonomous exosuit variants.

The AI is currently extremely lacklustre but with the other AI changes sitting in dev PRs I will be able to revisit it later and give them more interesting behavior (drakes spitting until the target is stunned then switching to claws, for example).

Expands our ability to make NPCs for exoplanets, ruins, etc. without needing to produce bespoke sprites. 

## Authorship
Myself.

## Changelog
Nothing player-facing.

### DIFF
--- a/code/datums/repositories/atom_info.dm
+++ b/code/datums/repositories/atom_info.dm
@@ -44,6 +44,7 @@ var/global/repository/atom_info/atom_info_repository = new()
 		description_cache[key] = instance.desc
 	if(cache_appearance && !appearance_cache[key])
 		instance = instance || get_instance_of(_path, _mat, _amount)
+		instance.compile_overlays()
 		appearance_cache[key] = instance.appearance
 	if(!matter_mult_cache[key] && ispath(_path, /obj))
 		var/obj/obj_instance = instance || get_instance_of(_path, _mat, _amount)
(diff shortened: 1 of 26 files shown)


out-of-phaze:
really wondering why we can't just do what pyrelight does and give complex mobs an AI datum

MistakeNot4892:
> really wondering why we can't just do what pyrelight does and give complex mobs an AI datum

Equipment is a huge pain in the ass to handle for ranged attacks. Reloading, drawing bows, swapping hands when in melee, picking up their weapon after being disarmed, etc. 

`/mob/living/simple_animal` has the projectile behavior baked in,`/mob/living/human` does not.

out-of-phaze (approved):
i would like to eventually replace anything using mob mimics with properly-written ai handling (noelle esp. has experience working with that sort of thing from roguetown and caves of qud, so i might bat it to her) but i see no problem with having them until then (and as a fallback for any new/unsupported behaviors/mobs you might want to mimic)

MistakeNot4892:
Yeah I have Plans:tm: to add stuff like iterating your equipped items to find a suitable held item, but it will be awhile. I want to rewrite AI in general before I start implementing potentially quite intensive stuff like that. Once all the current PRs are in (including the WIP AI stance one) I can look at doing it properly.

PR #5495, opened 2026-09-10
Overlay groups, ZM automangling

Lohikar:
This PR replaces priority overlays with overlay groups: what amounts to partitioning the overlays lists into named groups of overlays that can be manipulated as a unit without disturbing other groups.

All overlay procs that had a `priority` argument now have a `group` argument in its place. Overlay operations with a group specified will only apply to that group, such as only replacing overlays in that group, or only removing overlays from that group even if the same appearance exists elsewhere. This largely eliminates the need for manually tracking priority overlays for removal.
`TRUE` is a valid group name, so existing priority overlay logic works unmodified as long as it doesn't refer to the `priority` arg by name.

example:
```dm
var/list/my_overlays
if (some_condition)
    LAZYADD(my_overlays, foo)
if (some_other_condition)
    LAZYADD(my_overlays, bar)
set_overlays(my_overlays, "my thing")    // should probably be a macro, but not required
```

This also brings over ZM automatic plane mangling: SSoverlays will inspect overlays during overlay operations to automatically detect if an atom requires ZM mangling. Atoms that either explicitly mark themselves as requiring mangling or mark themselves as not requiring mangling will skip this scan.

caveats:
- It is currently not possible to define the order of overlay groups relative to each other. You can use `FLOAT_LAYER` offsets from BYOND if this is important. They're always above ungrouped overlays.
- Automangling does not look at overlays-of-overlays: atoms that are doing this should manually set the mangling flag.

other changes:
- `compile_overlays()` is now a no-op if called without the atom being queued. Pass `force = TRUE` to override (such as to restore broken overlay lists).
- `copy_overlays()` has been split into `copy_overlays()` and `replace_overlays()`.
- `set_overlays()` now accepts null as an overlay argument, which will cause it to act like `cut_overlays()`. This simplifies logic using lazylists.
- SSoverlays now has some statistics.
- ~~`z_flags` has been moved to `/atom` to simplify logic, though its meaning is still different between movables and turfs. No defined behavior for area.~~

### DIFF
--- a/code/__defines/lists.dm
+++ b/code/__defines/lists.dm
@@ -9,7 +9,6 @@
 #define SAFEPICK(L) DEFAULTPICK(L, null)
 // Ensures L is initailized after this point
 #define LAZYINITLIST(L) if (!L) L = list()
-#define LAZYINITALIST(L) if (!L) L = alist()
 // Sets a L back to null iff it is empty
 #define UNSETEMPTY(L) if (L && !L.len) L = null
 // Removes I from list L, and sets I to null if it is now empty
--- a/code/__defines/lists.dm
+++ b/code/__defines/lists.dm
@@ -9,6 +9,7 @@
 #define SAFEPICK(L) DEFAULTPICK(L, null)
 // Ensures L is initailized after this point
 #define LAZYINITLIST(L) if (!L) L = list()
+#define LAZYINITALIST(L) if (!L) L = alist()
 // Sets a L back to null iff it is empty
 #define UNSETEMPTY(L) if (L && !L.len) L = null
 // Removes I from list L, and sets I to null if it is now empty
--- a/code/__defines/overlays.dm
+++ b/code/__defines/overlays.dm
@@ -1,0 +1,4 @@
+/// Does this atom currently have managed overlays?
+#define HAS_MANAGED_OVERLAYS(atom) (atom.simple_overlays || atom.grouped_overlays)
+#define HAS_OVERLAY_GROUP(TGT, NAME) ((NAME) in (TGT):grouped_overlays)
+#define GET_OVERLAY_GROUP(TGT, NAME) ((TGT):grouped_overlays?[NAME])
--- a/code/controllers/subsystems/overlays.dm
+++ b/code/controllers/subsystems/overlays.dm
@@ -11,6 +11,18 @@ SUBSYSTEM_DEF(overlays)
 	var/list/overlay_icon_state_caches = list()
 	var/list/overlay_icon_cache = list()
 
+	var/compiles = 0
+	var/compiles_with_simple = 0
+	var/compiles_with_groups = 0
+	var/compiles_with_none = 0
+	var/compiles_async = 0
+	var/automangled = 0
+
+	// If the overlay set currently being considered contains a manglable overlay.
+	// This is only safe because SSoverlays can only ever consider one overlay list at a time with no interior sleeps. Professional on closed course, do not attempt.
+	// Setting this to TRUE on a non-movable will explode.
+	var/context_needs_automangle
+
 /// How many items should we process before we check for yield? Increasing this increases efficiency, but also raises risk of overrun.
 #define OVR_PUMP_RATIO 4
 /// Initialize state required for OVR_MC_TRY_YIELD.
@@ -20,7 +32,15 @@ SUBSYSTEM_DEF(overlays)
 #define OVR_MC_TRY_YIELD if ((++__yield) >= OVR_PUMP_RATIO) { __yield = 0; if (no_mc_tick) { CHECK_TICK; } else if (MC_TICK_CHECK) { break; } }
 
 /datum/controller/subsystem/overlays/stat_entry()
-	..("Ov: [processing.len - (idex - 1)]")
+	var/sync = compiles - compiles_async
+	var/ratio = compiles ? (sync / compiles) : 1
+	var/list/entries = list(
+		"Q: [processing.len - (idex - 1)]",
+		"Type: { Any: [compiles] | Simple: [compiles_with_simple] | Grouped: [compiles_with_groups] | Empty: [compiles_with_none] | Mangled: [automangled] }",
+		"Chrony: { Sync: [sync] | Async: [compiles_async] | SR: [round(ratio*100, 0.1)]% }",
+		"Caches: { Icon: [overlay_icon_cache.len] | Text: [overlay_icon_state_caches.len] }"
+	)
+	..(entries.Join("\n\t"))
 
 /datum/controller/subsystem/overlays/Initialize()
 	Flush()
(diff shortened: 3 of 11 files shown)


code/modules/mob_modifiers/definitions/modifiers_mech_shields.dm
```diff
@@ -108,7 +108,7 @@
 		object.icon_state = "mech_shield_[hardpoint]"
 		var/image/mech_overlay = image(object.icon, "[object.icon_state]_over")
 		mech_overlay.layer = ABOVE_HUMAN_LAYER
-		object.add_overlay(mech_overlay, priority = TRUE)
+		object.add_overlay(mech_overlay, TRUE)
```
MistakeNot4892, line 111: Is it valid to have `group` == TRUE?

Lohikar: As noted on Discord, anything other than the three falsey values is valid as a group name.

MistakeNot4892: I can't read

code/__defines/lists.dm
```diff
@@ -9,6 +9,7 @@
 #define SAFEPICK(L) DEFAULTPICK(L, null)
 // Ensures L is initailized after this point
 #define LAZYINITLIST(L) if (!L) L = list()
+#define LAZYINITALIST(L) if (!L) L = alist()
```
out-of-phaze: line 36/37, `A_LAZYINITLIST` already exists
```suggestion
```

code/controllers/subsystems/overlays.dm
```diff
@@ -128,102 +173,270 @@ SUBSYSTEM_DEF(overlays)
...
+	ASSERT(other != null)
+
+	z_flags |= other.z_flags & ZMM_AUTOMANGLE
+
+	if (other.simple_overlays)
+		LAZYINITLIST(simple_overlays)
+		simple_overlays += other.simple_overlays
+
+	if (also_grouped && other.grouped_overlays)
+		LAZYINITALIST(grouped_overlays)
```
out-of-phaze:
```suggestion
		A_LAZYINITLIST(grouped_overlays)
```

code/controllers/subsystems/overlays.dm
```diff
@@ -128,102 +173,270 @@ SUBSYSTEM_DEF(overlays)
...
-			cut_overlays()
-		return
+/// Copy overlays from another atom. If `also_grouped` is set, also copy grouped overlays. This is synchronous by default.
+/atom/proc/copy_overlays(atom/other, also_grouped = FALSE, now = TRUE)
+	ASSERT(other != null)
+
+	z_flags |= other.z_flags & ZMM_AUTOMANGLE
+
+	if (other.simple_overlays)
+		LAZYINITLIST(simple_overlays)
```
out-of-phaze, line 402:
```suggestion
		A_LAZYINITLIST(simple_overlays)
```

code/__defines/overlays.dm
```diff
@@ -0,0 +1,4 @@
+/// Does this atom currently have managed overlays?
+#define HAS_MANAGED_OVERLAYS(atom) (atom.simple_overlays || atom.grouped_overlays)
+#define HAS_OVERLAY_GROUP(TGT, NAME) ((NAME) in (TGT):grouped_overlays)
+#define GET_OVERLAY_GROUP(TGT, NAME) ((TGT):grouped_overlays?[NAME])
```
out-of-phaze, line 4: is it necessary to use `:` here? can we not just cast `TGT` before using the macro? as far as i can tell these two are unused so i can't see any example uses to tell if casting/etc would be too inconvenient

code/controllers/subsystems/overlays.dm
```diff
@@ -11,6 +11,18 @@ SUBSYSTEM_DEF(overlays)
...
+	var/compiles_with_simple = 0
+	var/compiles_with_groups = 0
+	var/compiles_with_none = 0
+	var/compiles_async = 0
+	var/automangled = 0
+
+	// If the overlay set currently being considered contains a manglable overlay.
+	// This is only safe because SSoverlays can only ever consider one overlay list at a time with no interior sleeps. Professional on closed course, do not attempt.
+	// Setting this to TRUE on a non-movable will explode.
+	var/context_needs_automangle
```
out-of-phaze: may be the rare case where marking this as `var/tmp/` would be useful to indicate the value is junk outside of its normal context. not blocking though

out-of-phaze (changes requested):
preliminary review

code/__defines/overlays.dm
Lohikar: Requiring manual casts before using a macro is bad UX imo, and this is consistent with how Z-Mimic's macros work. There's little benefit to changing this to use a typechecked operator. Probably could do with an 'isatom' check like the ZM macros have, though.

It's used in the changeset that replaces the manual AO turf state tracking with overlay group usage, but that's not in this PR.

code/controllers/subsystems/overlays.dm
Lohikar: This is a normal list.

out-of-phaze: i was wondering at the time why my ctrl-f didn't find it and i had to add it manually. that would explain that

code/__defines/overlays.dm
out-of-phaze:
```suggestion
#define HAS_OVERLAY_GROUP(TGT, NAME) (isatom(TGT) && ((NAME) in (TGT):grouped_overlays))
#define GET_OVERLAY_GROUP(TGT, NAME) (astype((TGT), /atom)?.grouped_overlays?[NAME])
```
i think an isatom check or astype is reasonable then. we could also have HAS_OVERLAY_GROUP_UNSAFE if that ever becomes necessary. not blocking though

code/controllers/subsystems/overlays.dm
```diff
@@ -101,7 +101,7 @@ SUBSYSTEM_DEF(overlays)
 		overlays.Cut()
 		SSoverlays.compiles_with_none++
 
-	if (istype(src, /atom/movable) && (z_flags & ZMM_AUTOMANGLE))
+	if (astype(src, /atom/movable)?.z_flags & ZMM_AUTOMANGLE)
```
out-of-phaze, line 104: +1 for astype usage

code/controllers/subsystems/overlays.dm
```diff
@@ -321,7 +326,7 @@ SUBSYSTEM_DEF(overlays)
 	overlays = build_appearance_list(overlays)
 
 	if (SSoverlays.context_needs_automangle)	// this will only ever be true on movables
```
out-of-phaze, line 328: wonder if we could make build_appearance_list rely on mutating its argument rather than returning a new list, and have it return automangle status instead. and if there would be any point to doing this

out-of-phaze (approved):

PR #5503, opened 2026-09-16
Fix (nearly-)every-tick animation spam for non-cloaked mobs

out-of-phaze:
Fixes an issue introduced by #5449. Non-cloaked mobs were animating their alpha every time update_icon was called, which was very often. Now we only animate it if the alpha value is the cloaked one. Should also possibly use ANIMATE_RELATIVE and ANIMATE_PARALLEL to avoid interrupting other animations but I'm less sure about that.


MistakeNot4892:
Ah, whoops. Thank you for identifying that.

Would suggest a slightly broader fix:
````
	if(can_use_cloak)
		var/cloak_alpha = (stat == CONSCIOUS && is_cloaked()) ? cloaked_alpha : initial(alpha)
		if(alpha != cloak_alpha)
			animate(src, alpha = cloak_alpha, time = cloak_anim_time, flags = ANIMATION_PARALLEL)
````
I'm a bit perplexed about why simplemobs are calling update_icon() so often though.

MistakeNot4892:
Covered by #5509

PR #5504, opened 2026-09-16
Converting everything to use the set_anchored() helper.

MistakeNot4892:
Finally getting around to `// TODO: make everything use this.` and swapped direct setting of `anchored` for the /atom/movable helper. Think I got 99% of uses but there's ~500 hits for `anchored\s*=\s*` and almost all of them are definitions on types so I likely missed a few.


MistakeNot4892:
Straightforward PR, selfmerging.

PR #5509, opened 2026-09-20
Fixes simple_animal cloak animation happening every update_icon().

MistakeNot4892:
Alternative to #5503 with an extra guard.


MistakeNot4892:
Going to selfmerge this as it's basically the same as Penny's fix but with an extra guard.

